PR #439, opened 2026-05-04
feat(daemon): add DeepSeek TUI as a code agent adapter

pftom:
## Summary

- Register `deepseek` (CLI binary `deepseek`, with `deepseek-tui` as the cargo-only fallback) in `AGENT_DEFS`. Invocation is `deepseek exec --auto [--model X] <prompt>` with plain-text streaming on stdout.
- Ships `deepseek-v4-pro` / `deepseek-v4-flash` (1M context, native thinking-mode) as fallback model hints; the custom-model input still accepts any other id including NVIDIA NIM / Fireworks / SGLang routes.
- Web UI: DeepSeek-blue gradient icon, `deepseek` → `DeepSeek` label, `deepseek tui` / `deepseek-tui` aliases.
- Docs: `docs/agent-adapters.md` §5.9 covers invocation, the `--auto` rationale, why we skip `--json` (it batches deltas into one trailing summary instead of streaming), the prompt-as-argv Windows ENAMETOOLONG note, and the auth-detection gap to follow up on.
- `.deepseek/` added to `.gitignore` alongside the other per-agent runtime data dirs (`.claude/`, `.codex/`, …).

Upstream repo: https://github.com/Hmbown/DeepSeek-TUI

## Verification

- `pnpm --filter @open-design/daemon test` — 371 passed (incl. 4 new `deepseek` tests covering buildArgs, model injection, default-model omission, and `fallbackBins` declaration).
- `pnpm --filter @open-design/daemon typecheck` — clean.

## Known limitations (called out in §5.9)

- **Prompt is delivered as a positional argv arg.** DeepSeek's `exec` declares `prompt: String` as a required clap field with no `-` stdin sentinel, so `promptViaStdin: true` isn't an option in v1. Typical chat prompts are fine; very large composed prompts (system + design system + skill + user message) can hit Windows' ~32 KB CreateProcess command-line limit.
- **Auth state is not auto-detected.** Detection currently reports `available: true` based on the binary being on PATH. If the user hasn't run `deepseek auth set --provider deepseek` (or set `DEEPSEEK_API_KEY`), the first run errors out with a non-actionable message. Wiring `deepseek doctor --json`'s `api_key.source` field into detection is a sensible follow-up.

## Test plan

- [ ] `npm i -g deepseek-tui`, then `deepseek auth set --provider deepseek` (paste API key from https://platform.deepseek.com/api_keys).
- [ ] `pnpm tools-dev` and confirm `GET /api/agents` returns `{ id: "deepseek", available: true, models: [..., "deepseek-v4-pro", "deepseek-v4-flash"] }`.
- [ ] In the web UI, pick **DeepSeek TUI** from the agent dropdown (look for the blue whale-tail icon) and run a small chat turn — verify text deltas stream to the chat panel.
- [ ] Pick a non-default model from the picker (`deepseek-v4-flash`) and confirm the spawn command embeds `--model deepseek-v4-flash`.
- [ ] On a host with only `deepseek-tui` (no `deepseek` dispatcher) installed via cargo, confirm detection still surfaces the agent via the `fallbackBins` path.


apps/daemon/src/agents.ts
```diff
@@ -594,6 +594,43 @@ export const AGENT_DEFS = [
...
+  {
+    id: 'deepseek',
+    name: 'DeepSeek TUI',
+    // The `deepseek` dispatcher is the canonical entry point; it delegates
+    // to a sibling `deepseek-tui` runtime binary at exec time. Both are
+    // installed by `npm i -g deepseek-tui` and by the cargo path. We probe
+    // the dispatcher first; if a user installed only the TUI binary (e.g.
+    // standalone cargo install), we still surface the agent.
+    bin: 'deepseek',
+    fallbackBins: ['deepseek-tui'],
```
mrcfps: This fallback currently advertises an install state that the daemon cannot actually run. When `resolveAgentExecutable` falls through to `deepseek-tui`, `/api/chat` still calls the same `buildArgs` below, so the process becomes `deepseek-tui exec --auto <prompt>`. Upstream documents `deepseek` as the dispatcher that owns subcommands like `exec` and delegates to the `deepseek-tui` companion runtime; the source docs also say both binaries are required for the cargo path. That means a host with only `deepseek-tui` on PATH will show DeepSeek as available, but the first chat run will fail instead of launching the agent. Please either remove `deepseek-tui` from `fallbackBins` until this path is supported, or teach resolution/buildArgs which binary was selected and use a verified `deepseek-tui` invocation, with a test that exercises the fallback command. 🙂

mrcfps (changes requested):
@pftom I reviewed the DeepSeek adapter registration, labels/icons, docs, and daemon tests. One install-path issue needs tightening before this is safe to merge, but the adapter shape is close.


lefarcen (commented):
The fallbackBins removal addresses mrcfps's safety concern cleanly — -only installs won't be falsely advertised as available. The new regression test pins this absence and the inline comments explain the rationale clearly.

No new issues in this revision. The adapter shape is solid: daemon integration follows the existing pattern, tests cover the argv construction + absence of fallbackBins, web UI icon/labels consistent, docs accurately document the limitation + future path.

LGTM from my side — ready for approval once mrcfps re-reviews or confirms the fix addresses the blocking concern. 🙂

lefarcen (commented):
The fallbackBins removal addresses mrcfps's safety concern cleanly — deepseek-tui-only installs won't be falsely advertised as available. The new regression test pins this absence and the inline comments explain the rationale clearly.

No new issues in this revision. The adapter shape is solid: daemon integration follows the existing pattern, tests cover the argv construction + absence of fallbackBins, web UI icon/labels consistent, docs accurately document the limitation + future path.

LGTM from my side — ready for approval once mrcfps re-reviews or confirms the fix addresses the blocking concern. 🙂

mrcfps (commented):
@pftom I reviewed the updated DeepSeek adapter after the fallback-bin fix. The earlier install-path blocker looks addressed; I left one merge-safe follow-up around making the documented argv-size limit fail more intentionally for large composed prompts.


apps/daemon/src/agents.ts
```diff
@@ -594,6 +594,46 @@ export const AGENT_DEFS = [
...
+    // every CLI without a TTY, so the interactive approval prompt would
+    // hang the run. Streaming is plain text on stdout (tool calls go to
+    // stderr); skipping `--json` keeps deltas streaming live instead of
+    // batched into one trailing summary object at end-of-turn.
+    buildArgs: (prompt, _imagePaths, _extra, options = {}) => {
+      const args = ['exec', '--auto'];
+      if (options.model && options.model !== 'default') {
+        args.push('--model', options.model);
+      }
+      args.push(prompt);
```
mrcfps, line 639: This positional prompt path is documented, but it would be safer to make oversized composed prompts fail before `spawn` with a DeepSeek-specific, user-actionable error. The daemon passes the fully composed prompt into `buildArgs` (system text, history, skills/design-system content, plus the user message), and every other argv-sensitive adapter in this file uses `promptViaStdin` specifically to avoid Linux `MAX_ARG_STRLEN` / Windows `CreateProcess` limits. Because this adapter pushes the whole prompt as one argv entry here, a large but otherwise valid OD run can fail before DeepSeek starts with `ENAMETOOLONG`/`E2BIG` and currently surfaces as a generic spawn failure rather than guidance that DeepSeek's CLI cannot accept that prompt size via argv.

Since upstream does not support stdin yet, please consider adding a small guard for this adapter, e.g. a `maxPromptArgBytes`/`requiresPromptArg` field that the `/api/runs` spawn path checks before calling `spawn`, emitting a clear SSE error such as "DeepSeek TUI requires prompt-as-argv and this prompt is too large; reduce selected skills/design context or use an adapter with stdin support". A regression test with a prompt over the conservative Windows limit would pin the behavior without changing the normal short-prompt path. 🙂

mrcfps, line 639: The positional prompt path is documented, and it is merge-safe, but the daemon should ideally reject oversized composed prompts before calling `spawn` with a DeepSeek-specific, user-actionable error. The changed code pushes the fully composed prompt into one argv entry here; unlike the adapters that set `promptViaStdin`, that can hit Linux `MAX_ARG_STRLEN` / Windows `CreateProcess` limits before DeepSeek starts, leaving users with a generic `ENAMETOOLONG`/`E2BIG` spawn failure instead of guidance that this CLI cannot accept that prompt size via argv.

A concrete follow-up would be to add a small adapter capability such as `maxPromptArgBytes` or `requiresPromptArg`, check it in the `/api/runs` spawn path before spawning, and emit a clear SSE error like: “DeepSeek TUI requires prompt-as-argv and this prompt is too large; reduce selected skills/design context or use an adapter with stdin support.” A regression test with a prompt over a conservative Windows command-line limit would pin the behavior while preserving the normal short-prompt path. 🙂

mrcfps (commented):
@pftom I reviewed the current DeepSeek adapter head after the main-branch merge. The fallback-bin blocker remains fixed; I’m carrying forward one merge-safe follow-up so the documented prompt-as-argv limit can fail with clearer guidance when users hit it.


mrcfps (commented):
@pftom I reviewed the current DeepSeek adapter head, including the fallback-bin fix and the new argv-budget guard. The previous blockers look addressed; I left one merge-safe follow-up to make the Windows command-line guard cover the actual quoted spawn shape more reliably. 🙂


apps/daemon/src/agents.ts
```diff
@@ -594,6 +594,58 @@ export const AGENT_DEFS = [
...
+    // other argv-sensitive adapter sets `promptViaStdin: true` to dodge
+    // this; DeepSeek's CLI doesn't accept `-` as a stdin sentinel yet, so
+    // we have to ship the prompt as argv. The /api/chat spawn path checks
+    // this byte budget against the composed prompt and emits an actionable
+    // SSE error ("reduce skills/design-system context, or use an adapter
+    // with stdin support") instead of letting the spawn fail with a
+    // generic ENAMETOOLONG/E2BIG message. 30_000 bytes leaves ~2.7 KB of
+    // argv headroom under the Windows command-line limit for `exec
+    // --auto --model <id>` and any internal quoting.
+    maxPromptArgBytes: 30_000,
```
mrcfps, line 653: This guard is a helpful improvement, but it still checks only the raw composed prompt byte count. On Windows, an npm-installed `deepseek` resolves to a `.cmd` shim and the spawn path later passes the final command through `createCommandInvocation`, which wraps it in `cmd.exe /d /s /c` and `quoteWindowsCommandArg` doubles embedded quotes before `CreateProcess` sees it. Because this line allows up to 30,000 raw UTF-8 bytes, a prompt under the budget but containing many quotes (or enough quoting-sensitive content from skills/history) can expand past the ~32,767-character Windows command-line limit and still surface the generic spawn failure this change is trying to avoid.

Please consider making the guard account for the final Windows invocation shape—e.g. compute the would-be command line length for `.cmd` resolutions after `buildArgs`, or lower the DeepSeek budget to leave worst-case quoting headroom and add a regression with a quote-heavy prompt. That keeps `AGENT_PROMPT_TOO_LARGE` reliable for the same class of prompts documented here without blocking normal short prompts. 🙂

lefarcen (commented):
@pftom I reviewed the new argv-budget guard refinement. You directly addressed mrcfps's suggestion with a clean two-pass system: the pre-bin `checkPromptArgvBudget` catches raw-byte oversized prompts, and the post-buildArgs `checkWindowsCmdShimCommandLineBudget` recomputes the exact `cmd.exe /d /s /c "<inner>"` command line to catch quote-heavy prompts that expand past CreateProcess's cap after Windows quote-doubling. The local mirror of `quoteForWindowsCmdShim` lets the guard run on macOS/Linux tests against fake `.cmd` paths, and the 10 regression tests cover the quote-heavy expansion scenario + short-prompt happy path + edge cases (non-`.cmd`, null bin, no budget). Implementation is sound.

No new issues — this revision fully addresses mrcfps's refinement suggestion. Ready for maintainer approval when mrcfps confirms or dismisses the stale CHANGES_REQUESTED from the earlier fallbackBins blocker.

apps/daemon/src/agents.ts
```diff
@@ -803,6 +856,92 @@ export function getAgentDef(id) {
...
+//   - the resolved binary isn't a `.cmd` / `.bat` (POSIX hosts and
+//     direct `.exe` resolutions on Windows skip the cmd.exe wrap);
+//   - the assembled line fits comfortably under the kernel cap.
+//
+// Pure: takes `resolvedBin` explicitly so a test on macOS can pass a
+// fake `C:\\…\\deepseek.cmd` path and exercise the same math the daemon
+// would run on Windows.
+export function checkWindowsCmdShimCommandLineBudget(def, resolvedBin, args) {
+  if (!def || typeof def.maxPromptArgBytes !== 'number') return null;
+  if (typeof resolvedBin !== 'string' || !/\.(bat|cmd)$/i.test(resolvedBin)) return null;
```
mrcfps, line 981: This `.cmd`/`.bat` early return makes the new second-pass guard skip Windows installs where the dispatcher resolves directly to `deepseek.exe` (for example a cargo-installed CLI rather than npm's `.cmd` shim). The raw `maxPromptArgBytes` check above still allows up to 30,000 bytes, but a quote-heavy prompt below that budget can expand when Node/libuv builds the `CreateProcess` command line for a direct `.exe` spawn too, so those users can still hit the generic `ENAMETOOLONG`/command-line-too-long failure that this guard is trying to turn into `AGENT_PROMPT_TOO_LARGE`.

Please consider either adding a companion direct-Windows-executable command-line budget check that mirrors Node/libuv's quoting for non-shim `.exe` resolutions, or lowering `maxPromptArgBytes` enough to leave worst-case quote-escaping headroom for both `.cmd` and `.exe` paths. A regression test with a quote-heavy prompt and a `C:\...\deepseek.exe` resolved path would pin the behavior. 🙂

mrcfps (commented):
@pftom I reviewed the current DeepSeek adapter head, including the DeepSeek registration, prompt-size guards, Windows shim handling, tests, docs, and UI labels/icons. The previous blockers look addressed; I found one merge-safe edge case where the Windows prompt-size guard can be broadened for direct `.exe` installs. 🙂


apps/daemon/src/agents.ts
```diff
@@ -594,6 +594,58 @@ export const AGENT_DEFS = [
...
+    // every CLI without a TTY, so the interactive approval prompt would
+    // hang the run. Streaming is plain text on stdout (tool calls go to
+    // stderr); skipping `--json` keeps deltas streaming live instead of
+    // batched into one trailing summary object at end-of-turn.
+    buildArgs: (prompt, _imagePaths, _extra, options = {}) => {
+      const args = ['exec', '--auto'];
+      if (options.model && options.model !== 'default') {
+        args.push('--model', options.model);
+      }
+      args.push(prompt);
```
mrcfps, line 639: This new adapter passes the fully composed prompt as a positional argv value, and on the documented npm install path `resolvedBin` is a `.cmd` shim that `server.ts` later runs through `createCommandInvocation`/`cmd.exe /d /s /c`. The changed Windows shim quoting only doubles `"`; it does not protect `%...%` sequences, so `cmd.exe` will expand text like `%DEEPSEEK_API_KEY%` from the daemon environment before DeepSeek receives the prompt. Because the composed prompt can include user text, project/skill/design-system content, and history while the spawn environment is built from `process.env`, a malicious or accidental `%SECRET_NAME%` in context can substitute local secrets into the prompt sent to the model, or at minimum corrupt the prompt. Please either avoid the `.cmd` wrapper for this argv-bound adapter, escape cmd expansion characters in the platform shim path before spawning, or force a direct executable/stdin-capable path; add a Windows command-invocation regression proving a prompt containing `%DEEPSEEK_API_KEY%` reaches the child literally. 🙂

apps/daemon/src/agents.ts
```diff
@@ -803,6 +856,203 @@ export function getAgentDef(id) {
...
+// null when the prompt fits (or the adapter has no budget — i.e. uses
+// stdin), and a structured error payload otherwise. Pure so it's
+// directly unit-testable for both the oversized and short-prompt paths
+// without spinning up the HTTP server or a real spawn.
+export function checkPromptArgvBudget(def, composed) {
+  if (!def || typeof def.maxPromptArgBytes !== 'number') return null;
+  const bytes = Buffer.byteLength(typeof composed === 'string' ? composed : '', 'utf8');
+  if (bytes <= def.maxPromptArgBytes) return null;
+  return {
+    code: 'AGENT_PROMPT_TOO_LARGE',
```
mrcfps, line 881: This introduces a new SSE/API error code, `AGENT_PROMPT_TOO_LARGE`, and `server.ts` forwards it through `createSseErrorPayload`, but the shared contract in `packages/contracts/src/errors.ts` still does not include that code in `API_ERROR_CODES`. The repository boundary says shared web/daemon error shapes live in `packages/contracts`; leaving this code daemon-only means typed consumers cannot recognize the new failure mode and future contract validation can reject or misclassify the SSE error. Please add `AGENT_PROMPT_TOO_LARGE` to the contracts package (with any matching tests/exports), or reuse an existing contracted code such as `PAYLOAD_TOO_LARGE` while keeping the DeepSeek-specific message. 🙂

mrcfps (changes requested):
@pftom I reviewed the current DeepSeek adapter head, including the Windows argv guards, SSE error path, tests, docs, and UI integration. Two blocking issues remain around the Windows .cmd prompt path and the new SSE error contract; both look focused and fixable. 🙂


apps/daemon/src/agents.ts
```diff
@@ -601,6 +601,58 @@ export const AGENT_DEFS = [
...
+    // every CLI without a TTY, so the interactive approval prompt would
+    // hang the run. Streaming is plain text on stdout (tool calls go to
+    // stderr); skipping `--json` keeps deltas streaming live instead of
+    // batched into one trailing summary object at end-of-turn.
+    buildArgs: (prompt, _imagePaths, _extra, options = {}) => {
+      const args = ['exec', '--auto'];
+      if (options.model && options.model !== 'default') {
+        args.push('--model', options.model);
+      }
+      args.push(prompt);
```
mrcfps, line 639: The DeepSeek adapter sends the fully composed prompt as a positional argv value, and on the documented npm install path that resolved `deepseek` binary is a `.cmd` shim that later goes through `createCommandInvocation` / `cmd.exe /d /s /c`. The new size guards model quote expansion, but the platform shim quoting still does not escape cmd variable expansion characters, so prompt text like `%DEEPSEEK_API_KEY%` can be expanded from the daemon environment before DeepSeek receives it. Because this argv entry can contain user text, project/skill/design-system content, and prior conversation history while the child environment is inherited from `process.env`, an accidental or malicious `%SECRET_NAME%` in context can substitute local secrets into the prompt sent to the model, or at least corrupt the prompt.

Please either avoid the `.cmd` wrapper for this argv-bound adapter, escape cmd expansion characters in the shim path before spawning, or require a direct executable/stdin-capable path for DeepSeek. A regression test around the Windows command invocation should prove a prompt containing `%DEEPSEEK_API_KEY%` reaches the child literally. 🙂

apps/daemon/src/agents.ts
```diff
@@ -811,6 +864,203 @@ export function getAgentDef(id) {
...
+// null when the prompt fits (or the adapter has no budget — i.e. uses
+// stdin), and a structured error payload otherwise. Pure so it's
+// directly unit-testable for both the oversized and short-prompt paths
+// without spinning up the HTTP server or a real spawn.
+export function checkPromptArgvBudget(def, composed) {
+  if (!def || typeof def.maxPromptArgBytes !== 'number') return null;
+  const bytes = Buffer.byteLength(typeof composed === 'string' ? composed : '', 'utf8');
+  if (bytes <= def.maxPromptArgBytes) return null;
+  return {
+    code: 'AGENT_PROMPT_TOO_LARGE',
```
mrcfps, line 881: This helper introduces a new SSE/API error code, `AGENT_PROMPT_TOO_LARGE`, and `server.ts` forwards it through `createSseErrorPayload`, but `packages/contracts/src/errors.ts` still does not include that code in `API_ERROR_CODES`. The repository boundary keeps shared web/daemon error shapes in `packages/contracts`; leaving this value daemon-only means typed consumers cannot recognize the new failure mode, and future contract validation can reject or misclassify the SSE error even though this path is now part of the daemon behavior.

Please add `AGENT_PROMPT_TOO_LARGE` to the contracts package (with any matching exports/tests), or reuse an existing contracted code such as `PAYLOAD_TOO_LARGE` while keeping the DeepSeek-specific message. 🙂

mrcfps (changes requested):
@pftom I reviewed the current DeepSeek adapter head, including the Windows argv guards, SSE error path, tests, docs, and UI integration. Two blocking issues still need attention around the Windows .cmd prompt path and the new SSE error contract; both look focused and fixable. 🙂


mrcfps (approved):

Issue #430, opened 2026-05-04
## Bug: Claude Code fails to read skill files (directory access policy)


LucaLombardo03:
### Description

When the daemon spawns Claude Code with `cwd` set to `.od/projects/<id>/`, the agent attempts to read skill files from the absolute path of the repo's `skills/` directory (e.g. `C:\siti\open-design\skills\blog-post\SKILL.md`). Claude Code's directory access policy blocks reads on any path outside the working directory, causing all skill reads to fail with an error.

lefarcen:
## Fix in progress: PR #435

**Root cause identified:** The daemon exposed skill resources via absolute paths (e.g. `C:\siti\open-design\skills\...`) and relied on `--add-dir` to widen directory access. This contract is fragile because:
1. Most agents (codex, gemini, opencode, cursor-agent, etc.) don't support `--add-dir`.
2. Claude Code's directory-access policy can still block reads outside `cwd` even with `--add-dir` + `--permission-mode bypassPermissions`.

**Solution in PR #435:** Expose skill resources from inside the agent's working directory via cwd-relative symlinks/junctions (`.od-skills/`, `.od-design-systems/`). This works universally across all agents without per-CLI flag wiring.

Implementation is currently under review in #435. The issue stays open until the fix lands and is verified.

cc @fancyboi999

lefarcen:
@LucaLombardo03 The fix you reported in this issue should be in PR #435, just merged. Before replying, read the full issue history and merged PR. Suggest only verified repo commands; for open-design fresh-pull verification use `git pull && pnpm install && pnpm tools-dev run web` unless the thread says otherwise. If you still hit this, leave a reply here and we'll keep digging. Thanks for the report!

lefarcen:
## Fix shipped: PR #435 merged

@LucaLombardo03 PR #435 just landed — skill resources are now exposed via `.od-skills/<folder>/` inside your agent's working directory instead of absolute paths outside `cwd`. This should resolve the Claude Code directory-access policy block you hit.

**To verify:**
```bash
cd C:\siti\open-design
git pull
pnpm install
```
Then try the same skill that failed before — Claude Code should now read skill files without permission prompts.

If you still see the directory-access error, please share the exact error message and we'll investigate further. Otherwise, this issue can be closed once you confirm it works. 🙏

lefarcen:
**Update:** PR #440 addresses this issue.

[...]

PR #440, opened 2026-05-04
fix(daemon): correct Claude Code --add-dir capability detection (fix …

1119302165:
## Description

This PR fixes issue #430 where Claude Code failed to read skill files due to incorrect directory access policy enforcement.

### Root Cause

The Claude Code agent definition was probing **global** `claude --help` output for capability flags, but `--add-dir` and `--include-partial-messages` are **subcommand-specific flags** that only appear under `claude -p --help`, not in the global help output.

This caused:
1. Capability detection to incorrectly set `caps.addDir = false` for users with working Claude Code installations
2. The `--add-dir` flag was therefore never passed on agent spawn
3. Claude Code's sandbox enforcement blocked all absolute-path reads outside the project cwd
4. Skill files, skill assets, and design-system specs could not be read by the agent

> **Note:** The "probing failed / timeout" path already worked correctly, as it fell back to an empty capabilities object `{}` which defaulted to passing `--add-dir`. The failure mode only manifested when the help probe actually succeeded.

---

### Scope
| Aspect | Impact |
|---|---|
| **Agents affected** | Claude Code ONLY. All 12+ other agents (Codex, Cursor, Devin, Gemini, etc.) are completely untouched. |
| **Code change surface** | 1 line changed in `helpArgs` + comment. No logic changes to any other agent. |
| **API compatibility** | No external API changes. `helpArgs` is stripped before serialization. |
| **Backward compatibility** | Fully backward compatible. Older Claude Code versions continue to work via fallback mechanism. |

### Execution Paths Affected

| Path | Before | After |
|---|---|---|
| Daemon agent detection | `execFile('claude', ['--help'])` | `execFile('claude', ['-p', '--help'])` |
| Agent spawn `buildArgs` | `--add-dir` never passed when probe succeeded | `--add-dir` correctly passed when detected |
| `/api/agents` endpoint | `helpArgs` internal-only field stripped | No change — still stripped |

---

## Changes Made

| File | Change |
|---|---|
| `apps/daemon/src/agents.ts` | Updated Claude Code `helpArgs` to probe subcommand help |
| `apps/daemon/tests/agents.test.ts` | Added 3 regression tests for issue #430 |

---

## Regression Protection

Tests added:
1. ✅ Default `--add-dir` propagation when directories are supplied
2. ✅ Correct filtering of empty/null directory entries
3. ✅ Explicit assertion that `['-p', '--help']` is used for probing (prevents regression)

---

## Testing

1. **Before this fix:** Running `claude --help | grep --add-dir` returns no matches
2. **After this fix:** Running `claude -p --help | grep --add-dir` correctly finds the flag
3. All existing agent tests continue to pass
4. New regression tests prevent this specific failure mode from recurring

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -124,10 +124,13 @@ export const AGENT_DEFS = [
     // — issue #235) get auto-detected without writing wrapper scripts.
     fallbackBins: ['openclaude'],
     versionArgs: ['--version'],
-    helpArgs: ['--help'],
+    helpArgs: ['-p', '--help'],
     capabilityFlags: {
       // Flag string -> capability key. After probing `--help`, we set
       // `agentCapabilities[id][key] = true` for each substring that matches.
+      // `--add-dir` and `--include-partial-messages` live under `claude -p`
+      // subcommand, so we probe `claude -p --help` instead of `claude --help`.
+      // Fixes issue #430: --add-dir never detected because it wasn't in global help.
       '--include-partial-messages': 'partialMessages',
       '--add-dir': 'addDir',
     },
--- a/apps/daemon/tests/agents.test.ts
+++ b/apps/daemon/tests/agents.test.ts
@@ -272,6 +272,61 @@ test('claude flags promptViaStdin and never embeds the prompt in argv', () => {
   assert.ok(args.includes('-p'), 'claude argv must include -p');
 });
 
+// ---- Claude Code --add-dir capability (issue #430) -------------------------
+// Skill seeds (`skills/<id>/assets/template.html`) and design-system specs
+// (`design-systems/<id>/DESIGN.md`) live outside the project cwd. Without
+// `--add-dir`, Claude Code's directory access policy blocks reads on any
+// path outside the working directory. Bug was that we probed global `claude
+// --help` for `--add-dir` but that flag only appears in `claude -p --help`.
+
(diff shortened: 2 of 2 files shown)


lefarcen:
Hi @1119302165! 🎉

Thanks for the contribution — this is a precise root-cause fix with thorough testing and impact analysis.

— open-design team

mrcfps (approved):
@1119302165 I reviewed the Claude Code capability-detection change in `apps/daemon/src/agents.ts` and the added regression coverage in `apps/daemon/tests/agents.test.ts`. The fix is tightly scoped to probing the `claude -p --help` subcommand where `--add-dir` is advertised, and the tests cover the default add-dir behavior, directory filtering, and the root-cause help arguments. I did not find any actionable issues in the changed ranges; thanks for the clear, focused fix and the helpful regression protection. 🙌


lefarcen:
I've reviewed the changes as promised. @mrcfps's review above covers all the code correctness and test coverage points comprehensively — the fix is tightly scoped and well-protected by regression tests.

No additional actionable issues from my end. Thanks for the clear, focused fix! 🎉

lefarcen:
🎉 Merged! Thanks @1119302165 for the precise root-cause fix and thorough regression tests.

The capability detection now correctly probes `claude -p --help` where `--add-dir` is actually advertised, and the tests guard against this exact failure mode recurring.

Appreciate the clear analysis and careful implementation! 🙏

PR #287, opened 2026-05-02
fix: language dropdown overflow in settings modal (#281)

sinamashini:
## Summary

Fixes #281 — the language selector in the Settings modal was hard-coded to open **upward** from the trigger using fixed positioning, which caused the menu to render off-screen / clipped when the modal sat near the top of the viewport, and made long language labels hard to view.

### `apps/web/src/components/SettingsDialog.tsx`
- Added smart placement for the language menu:
  - Compute `spaceBelow = window.innerHeight - languageMenuRect.bottom` and `spaceAbove = languageMenuRect.top`.
  - Prefer opening **downward** when `spaceBelow >= spaceAbove` or there are at least 200px below.
  - Fall back to the original upward placement only when there is more space above.
- Expose the available height to CSS via the `--menu-available-h` custom property so the menu is bounded in either direction.
- Imported `CSSProperties` from `react` for typing the dynamic style object.

### `apps/web/src/index.css`
- `.settings-language-menu`: added `max-height: min(360px, var(--menu-available-h, 360px))` and `overflow-y: auto` so the list scrolls internally regardless of placement direction.
- `.settings-language-option`: added `min-height: 40px`.
- Added ellipsis truncation for `.settings-language-option-title` and `.settings-language-option-code` so long labels don't blow up row heights.

## Acceptance criteria

- [x] Menu renders fully inside the viewport.
- [x] Opens **below** the trigger when there is enough space below.
- [x] Falls back to opening **above** only when more space exists there.
- [x] Long labels no longer cause oversized item rows.
- [x] List scrolls internally when content exceeds available height.
- [x] Existing keyboard (Escape) / mouse behavior preserved.
- [x] No regressions to existing `.settings-language-menu` styling or animation.

Closes #281

PR #447, opened 2026-05-04
fix(web): language option button height truncation in settings

nmsn:
## Summary

- Remove `min-height: 40px` from `.settings-language-option` — the hardcoded minimum constrains button height when locale labels wrap to two lines, causing the second line to overflow outside the button.
- Update `.settings-language-menu` max-height budget from `360px` (9 × 40px) to `406px` (7 × 58px) to match the natural item height.

## Background

PR #287 added `min-height: 40px` to ensure consistent row height, with the menu `max-height` calculated as `9 locales × ~40px = 360px`. This assumption breaks when locale labels are long enough to wrap (e.g. Chinese or other multi-byte labels), where each item naturally sizes to ~58px instead of 40px.

## Changes

| File | Change |
|------|--------|
| `apps/web/src/index.css` | Remove `min-height: 40px` from `.settings-language-option`; update menu `max-height` from `360px` to `406px` |

## Test plan

- [ ] Open Settings → Language picker, verify all locale options display fully without text overflow
- [ ] Verify menu scrolls correctly when content exceeds viewport height
- [ ] Test on short viewport (mobile landscape) to confirm `60vh` cap still applies

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -1433,13 +1433,13 @@ code {
   position: fixed;
   z-index: 1000;
   max-width: calc(100vw - 24px);
-  /* 360px = 9 locales × ~40px; --menu-available-h is set by JS to the
+  /* 7 locales × ~58px ≈ 406px; --menu-available-h is set by JS to the
      distance between the trigger and whichever viewport edge the menu
      opens toward, so the menu never overflows the viewport regardless of
      placement direction (above or below the trigger). The 60vh cap keeps
      the menu from feeling cramped on short viewports (mobile landscape,
      ~400–500px tall). */
-  max-height: min(360px, 60vh, var(--menu-available-h, 360px));
+  max-height: min(406px, 60vh, var(--menu-available-h, 406px));
   overflow-y: auto;
   overscroll-behavior: contain;
   display: flex;
@@ -1457,7 +1457,6 @@ code {
   align-items: center;
   gap: 12px;
   width: 100%;
-  min-height: 40px;
   padding: 10px 12px;
   border: 0;
   border-radius: 9px;
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -1433,13 +1433,12 @@ code {
   position: fixed;
   z-index: 1000;
   max-width: calc(100vw - 24px);
-  /* 7 locales × ~58px ≈ 406px; --menu-available-h is set by JS to the
-     distance between the trigger and whichever viewport edge the menu
-     opens toward, so the menu never overflows the viewport regardless of
-     placement direction (above or below the trigger). The 60vh cap keeps
-     the menu from feeling cramped on short viewports (mobile landscape,
-     ~400–500px tall). */
-  max-height: min(406px, 60vh, var(--menu-available-h, 406px));
+  /* 7 locales × ~58px + menu chrome (padding, gap, border) ≈ 428px;
+     --menu-available-h is set by JS to the distance between the trigger
+     and whichever viewport edge the menu opens toward, so the menu never
+     overflows the viewport regardless of placement direction. The 60vh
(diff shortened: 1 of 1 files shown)


lefarcen:
Hi @nmsn! 🎉
Thanks for the contribution — this looks like a clean fix for the language picker overflow issue.

— open-design team

mrcfps (approved):
@nmsn I reviewed the settings language picker CSS change in `apps/web/src/index.css`: the option buttons now size naturally when labels wrap, and the menu height cap/fallback were updated consistently so the popover remains scrollable within the viewport. This is a focused, tidy fix—thanks for taking care of the overflow edge case. 🌟


lefarcen (commented):
Hey @nmsn, thanks for tackling the language picker overflow! 👋

I ran a thorough review and found two concerns about whether this fix addresses the root cause:

**Main concern (P2):** Removing `min-height: 40px` from `.settings-language-option` is unlikely to fix height truncation, because `min-height` is only a *lower bound*, not a ceiling. Elements should already grow past 40px when content wraps. 

The real constraint is likely elsewhere in the layout (possibly the grid column sizing). Suggest:
1. Reproduce the issue with a narrow menu + long label to confirm the actual truncation point
2. Check if `grid-template-columns` needs `minmax(0, 1fr)` to allow the text column to shrink/wrap properly
3. Ensure `white-space: normal` is set to allow wrapping
4. Add a regression test case

**Minor note (P3):** The new `406px` max-height (7 × 58px) doesn't account for menu chrome (padding: 4px, gap: 2px between rows, border). The 7th row could still clip at viewport edges. Consider bumping to ~428px or using CSS vars (`--row-height`, `--visible-rows`, `--gap`) so the calculation stays aligned with the comment.

**Overall:** The diagnosis is sound (wrapped labels need more space), but the implementation may not fix the observed truncation. Worth reproducing first to target the actual constraint directly.

Let me know if you need help debugging!

nmsn:
> Good points. To clarify the observed behavior: the truncation happens at the menu boundary, not the individual button — the `max-height: 360px` on the menu container clips the last visible item when total content height exceeds the budget. The `min-height: 40px` removal is a correctness improvement (the 40px assumption was baked into the 360px budget), but the real fix is the `max-height` adjustment.
>
> For P3: agreed, 406px doesn't account for menu chrome. I'll bump it to ~428px.

lefarcen:
Thanks for the clarification, @nmsn! That makes sense — the menu container's `max-height` is the actual constraint, and removing `min-height: 40px` ensures buttons can grow naturally without fighting the budget calculation. Looking forward to the ~428px adjustment for the chrome overhead. 👍

mrcfps (approved):
@nmsn I reviewed the updated settings language picker CSS in `apps/web/src/index.css`: the menu height cap now budgets for seven naturally sized wrapped locale rows plus padding/gaps/border, and removing the old 40px row assumption lets options grow with their label content while preserving the viewport/available-height constraints. I also checked the surrounding picker positioning styles and prior review context; the follow-up chrome adjustment addresses the earlier clipping concern cleanly. Thanks for the thoughtful, focused fix! 🌟


lefarcen:
Perfect! 🎉 The `~428px` adjustment accounts for the menu chrome now, and the comment clearly documents the calculation. Thanks for the quick fix, @nmsn!

lefarcen:
Merged! 🎉 Thanks @nmsn for the contribution to open-design. Looking forward to the next one!

lefarcen:
Thanks for the fix, @nmsn! 🎉 The language picker now handles wrapped locale labels correctly, and users won't see truncation anymore. Appreciate the clear analysis in your PR description—the budget adjustment from 360px → 428px (accounting for menu chrome) makes perfect sense.

Merged and shipped! 🚀

PR #455, opened 2026-05-04
feat(daemon): add link code folder support for agent context

encyc:
## Summary

Enables the "Link code folder" import menu item so users can link local code directories to a project. The AI agent then reads the linked source code via `--add-dir` when generating designs, giving it full context of the user's codebase.

- **Daemon**: New `POST /api/dialog/open-folder` endpoint opens a native OS folder picker (osascript on macOS, zenity on Linux, PowerShell on Windows). Linked directories are validated (absolute path, exists on disk, is a directory, not a sensitive system path), stored in `metadata.linkedDirs` on the project record, and appended to `extraAllowedDirs` in `startChatRun` so the agent receives them via `--add-dir`.
- **Web**: The "Link code folder" import item is now clickable. After the user picks a folder, it appears as a removable chip below the chat input. Chips show the folder basename, display the full path on hover, and can be removed with an X button.
- **Contracts**: `linkedDirs?: string[]` added to `ProjectMetadata`.
- **i18n**: Strings added for all 16 locales with native translations.
- **Tests**: 8 unit tests for `validateLinkedDirs` covering valid paths, relative paths, non-existent dirs, files, duplicates, and normalization.

No new dependencies. No database migration (uses the freeform `metadata_json` column). No changes to agent adapter code — the existing `extraAllowedDirs` → `--add-dir` pipeline is reused.

## Test plan

- [x] `pnpm --filter @open-design/daemon test` — 436 tests pass (including 8 new)
- [x] `pnpm --filter @open-design/web typecheck` — clean
- [ ] Manual: open project chat → click Import → Link code folder → native picker appears
- [ ] Manual: select a directory → chip appears below input → send message → agent reads linked files
- [ ] Manual: click X on chip → directory unlinked → PATCH confirmed in network tab
- [ ] Manual: verify on macOS, Linux, Windows that native folder dialog opens
- [ ] Verify no regression on other import menu items (all still disabled as before)


lefarcen:
Hi @encyc! 🎉

Thanks for the contribution — this is a well-scoped feature with solid testing and i18n coverage.

— open-design team

apps/daemon/src/linked-dirs.ts
```diff
@@ -0,0 +1,35 @@
+import path from 'node:path';
+import fs from 'node:fs';
+
+const BLOCKED_PATH_PREFIXES =
+  process.platform === 'win32'
+    ? ['C:\\Windows', 'C:\\Program Files', 'C:\\Program Files (x86)']
+    : ['/etc', '/proc', '/sys', '/dev', '/boot'];
+
+export function validateLinkedDirs(dirs) {
```
mrcfps: This new helper is included by `apps/daemon/tsconfig.json`, which has `strict` enabled, but the exported `validateLinkedDirs` parameter is currently untyped. That makes the daemon package fail TypeScript checking/building with an implicit-`any` error as soon as this file is compiled, so the new feature cannot pass the required daemon validation. Please type the input and result explicitly, for example `export function validateLinkedDirs(dirs: unknown): { dirs: string[]; error?: never } | { error: string; dirs?: never }`, and annotate `validated` as `string[]` so callers and tests get a stable return shape. 🙂

mrcfps (changes requested):
@encyc thanks for the well-scoped feature work here. I found one build-blocking issue in the new daemon source that needs a small typing fix before this can merge.


encyc:
Add import -> Link code folder 
image
Test feature
image

apps/daemon/src/linked-dirs.ts
encyc: Fixed in 9ec6b70. Added JSDoc type annotations for the parameter and return type, and annotated the `validated` array — `strict` mode should be happy now. Thanks for catching that! 🙂

apps/daemon/src/linked-dirs.ts
```diff
@@ -0,0 +1,37 @@
+import path from 'node:path';
+import fs from 'node:fs';
+
+const BLOCKED_PATH_PREFIXES =
+  process.platform === 'win32'
+    ? ['C:\\Windows', 'C:\\Program Files', 'C:\\Program Files (x86)']
+    : ['/etc', '/proc', '/sys', '/dev', '/boot'];
+
+/** @type {(dirs: unknown) => { dirs: string[], error?: undefined } | { error: string, dirs?: undefined }} */
+export function validateLinkedDirs(dirs) {
```
mrcfps: The attempted strict-mode fix still leaves `dirs` untyped in a `.ts` source file: the new `@type` JSDoc is attached to the function value, but it does not replace TypeScript parameter/return annotations for this function declaration. Because `apps/daemon/tsconfig.json` type-checks `src/` with strict settings, `export function validateLinkedDirs(dirs)` is still seen as an implicit-`any` parameter, so the daemon package can fail typecheck/build as soon as dependencies are installed. Please make the TypeScript annotations explicit here, e.g. `export function validateLinkedDirs(dirs: unknown): { dirs: string[]; error?: undefined } | { error: string; dirs?: undefined }`, and change `const validated = []` to `const validated: string[] = []`. That keeps the helper’s public shape stable for `server.ts` and the new tests.

mrcfps (changes requested):
@encyc thanks for tightening up the link-folder flow and addressing the earlier feedback. One TypeScript build blocker still remains in the new daemon helper, so I’m requesting one more small change before merge. 🙂


lefarcen (commented):
Thanks @encyc for addressing the typing issue! The JSDoc annotations should satisfy TypeScript strict mode.

**Code review (Lens A):**
- ✅ **Security**: Solid validation — absolute path enforcement, blocked system directories, existence + isDirectory checks, and deduplication. No path traversal risk.
- ✅ **Tests**: 8 unit tests cover the critical edge cases (relative paths, non-existent, files vs dirs, duplicates, normalization).
- ✅ **Error handling**: The validation returns structured errors with clear messages.
- ✅ **Architecture**: Clean separation (validateLinkedDirs helper → server.ts PATCH endpoint → web UI chips). Reuses existing `extraAllowedDirs` pipeline.

**Minor observation (non-blocking):** `linked-dirs.ts` line 18 checks `!path.isAbsolute(d)` then line 20 does `path.resolve(d)`, which always produces an absolute path. The `isAbsolute` check before `resolve` is technically redundant (resolve already handles relative inputs), but keeping it makes the error message clearer for users who pass relative paths, so it's fine as-is.

Overall this is production-ready. The typing fix addresses mrcfps's blocking feedback. No security or correctness issues found.

apps/daemon/src/linked-dirs.ts
encyc: Fixed in cead90b — replaced JSDoc with inline TS parameter/return annotations: `dirs: unknown` and explicit return union type. Also added `const validated: string[] = []`. Thanks!

lefarcen (commented):
Perfect! The TypeScript annotations are now inline as requested. 

**Current head ():**
- ✅ Explicit TS parameter type: `dirs: unknown`
- ✅ Explicit TS return type: `{ dirs: string[]; error?: undefined } | { error: string; dirs?: undefined }`
- ✅ Validated array typed: `const validated: string[] = []`

This resolves mrcfps's blocking feedback. The daemon package will now type-check correctly with strict mode enabled. All prior review points (security validation, test coverage, architecture) remain solid.

lefarcen (commented):
Perfect! The TypeScript annotations are now inline as requested.

**Current head (cead90b):**
- ✅ Explicit TS parameter type: `dirs: unknown`
- ✅ Explicit TS return type: `{ dirs: string[]; error?: undefined } | { error: string; dirs?: undefined }`
- ✅ Validated array typed: `const validated: string[] = []`

This resolves mrcfps's blocking feedback. The daemon package will now type-check correctly with strict mode enabled. All prior review points (security validation, test coverage, architecture) remain solid.

apps/daemon/src/linked-dirs.ts
```diff
@@ -0,0 +1,37 @@
...
+  const validated: string[] = [];
+  for (const d of dirs) {
+    if (typeof d !== 'string' || !d.trim()) {
+      return { error: 'each linked dir must be a non-empty string' };
+    }
+    if (!path.isAbsolute(d)) {
+      return { error: `linked dir must be an absolute path: ${d}` };
+    }
+    const resolved = path.resolve(d);
+    const blocked = BLOCKED_PATH_PREFIXES.some(
```
mrcfps: This blocklist still lets callers link directories that expose the blocked trees. For example, `/` is absolute, exists, and is not equal to `/etc` or prefixed by `/etc/`, so it passes validation and later gets added to `extraAllowedDirs`; a symlink such as `/tmp/etc-link -> /etc` also passes because the check runs on `path.resolve(d)` before `statSync` follows the symlink. That defeats the intended system-directory safety policy and can grant the agent access to sensitive filesystem areas via `--add-dir`. Please canonicalize with `fs.realpathSync.native`, reject filesystem roots/drive roots, and reject both blocked descendants and ancestors that would expose a blocked prefix; adding tests for `/` (or a drive root on Windows) and symlink-to-blocked-dir coverage would lock this down. 🙂

apps/daemon/src/server.ts
```diff
@@ -2427,9 +2488,21 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
           .map((f) => `- ${f.name}`)
           .join('\n')}`
       : '\nThis folder is empty. Choose a clear, descriptive filename for whatever you create.';
+    const projectRecord =
+      typeof projectId === 'string' && projectId
+        ? getProject(db, projectId)
+        : null;
+    const linkedDirs = Array.isArray(projectRecord?.metadata?.linkedDirs)
```
mrcfps: This chat-run path trusts persisted `metadata.linkedDirs` and only filters with `fs.existsSync`, but the project creation route stores arbitrary `metadata` without running `validateLinkedDirs`. A client can create a project with `metadata.linkedDirs: ['/']` (or any existing path), bypass the PATCH validation entirely, and this changed code will still include that path in the prompt and `extraAllowedDirs`. Since this directly controls what the agent may read, please validate/normalize `linkedDirs` on project creation and re-run the same validator before using persisted metadata here (dropping or failing invalid entries), with a route/start-run regression test for the create-project bypass.

mrcfps (changes requested):
@encyc thanks for the quick follow-up on the earlier typing and UI-state feedback. I found two linked-folder validation gaps that can still widen agent filesystem access beyond the intended policy, so I’m requesting changes before merge. 🙂


lefarcen:
mrcfps's latest review (14:59) found two filesystem security gaps that I missed earlier — both are P1. The inline comments on linked-dirs.ts and server.ts point to the exact validation bypasses. These need to be fixed before merge.

Sorry for the earlier "production-ready" assessment — I should have caught the root/symlink bypass and the project-creation validation gap.

apps/daemon/src/linked-dirs.ts
encyc: Fixed in c68c125. Now uses `fs.realpathSync.native` to resolve symlinks, rejects filesystem root `/`, and canonicalizes blocklist entries (handles macOS /etc -> /private/etc). Added tests for root rejection, symlink-to-blocked-dir, and realpath resolution.

apps/daemon/src/server.ts
encyc: Fixed in c68c125. Project creation now runs `validateLinkedDirs` on `metadata.linkedDirs` before storing. `startChatRun` also re-validates persisted dirs with the same validator instead of just checking existence. This closes the create-project bypass.

lefarcen:
Both P1 security fixes look correct on c68c125:

**Path validation bypass (mrcfps comment 3182461516):**
✅ Now uses `fs.realpathSync.native()` to canonicalize paths and follow symlinks
✅ Rejects filesystem roots (`/`, `C:\`)
✅ `isBlocked()` now checks both descendants AND ancestors (line 28: `p.startsWith(realPath + path.sep)`)
✅ `BLOCKED_CANONICAL` pre-canonicalizes blocked paths to handle macOS `/etc` → `/private/etc`
✅ Tests cover: root rejection, symlink-to-blocked-dir, symlink resolution

**Project creation bypass (mrcfps comment 3182461522):**
✅ Project creation validates `linkedDirs` (server.ts:822-832), drops invalid entries
✅ `startChatRun` re-validates before use (server.ts:2502-2505), defensive double-check
✅ Invalid paths never reach `extraAllowedDirs`

Waiting for mrcfps to re-review the current head.

mrcfps (approved):
Approved.

mrcfps:
Hi @encyc! Thanks for the work here — looks like CI is currently failing on `Validate workspace`. Could you take a look when you get a chance? 🙏

encyc:
> Hi @encyc! Thanks for the work here — looks like CI is currently failing on `Validate workspace`. Could you take a look when you get a chance? 🙏

fix in latest commit d99282a, please try again, thanks ^ ^.

lefarcen:
Hi @encyc — I verified the latest push (d99282a) and it still has unguarded `result.error` accesses in the test file:

- Line 26: `result.error.includes('does not exist')`
- Line 35: `result.error.includes('not a directory')`

These need `result.error!` or an explicit guard before calling `.includes(...)` to pass CI's strict TypeScript check. The return type is `{ dirs, error?: undefined } | { error, dirs?: undefined }`, so TS treats `error` as possibly `undefined` unless narrowed.

Quick fix: change those lines to `result.error!.includes(...)` (or wrap in `if (result.error) { assert.ok(result.error.includes(...)); }` if you prefer the guard style).

mrcfps (approved):
Thanks again @encyc for the thoughtful work and the quick follow-ups here. I really appreciate how promptly you addressed the CI/typecheck issue and tightened the validation details along the way. CI is green now, and this looks good to me — nice work pushing this feature through! 🙌

lefarcen:
Thanks again @encyc for the thoughtful iteration on this feature and the quick security/typing fixes throughout the review! 🎉 The link-folder validation is now solid, and the  pipeline gives users full codebase context. Really appreciate the careful follow-through here.

— open-design team

PR #457, opened 2026-05-04
chore(scripts): seed daemon with pre-baked decks and web prototypes

pftom:
## Summary
- Adds `pnpm seed:test-projects`, a thin HTTP client that talks to the running daemon and pre-bakes curated slide decks + web prototypes for testing the UI without waiting for an LLM run.
- Each seeded project loads its skill's `example.html` as `index.html`, sets the active tab, and adds two fake chat messages so the conversation panel renders end-to-end. `--clear` removes only `seed-`-prefixed projects.
- Bundle: 4 decks (`html-ppt-pitch-deck`, `kami-deck`, `html-ppt-weekly-report`, `html-ppt-product-launch`) + 3 web prototypes (`open-design-landing`, `kami-landing`, `dashboard`). Counts capped via `--decks N` / `--webs N`.

## Usage
```bash
pnpm tools-dev               # in one shell
pnpm seed:test-projects      # default 4 + 3
pnpm seed:test-projects --decks 2 --webs 2
pnpm seed:test-projects --daemon http://127.0.0.1:17456
pnpm seed:test-projects --clear
```

Picks up `OD_DAEMON_URL` / `OD_PORT` from the env that `tools-dev` exports.

## Test plan
- [x] `pnpm seed:test-projects --help` renders cleanly.
- [x] Run against a freshly-booted daemon: 4 projects created, each with `index.html` populated from the skill's `example.html`, tabs `[index.html]` active=`index.html`, and a user + assistant message in the seeded conversation.
- [x] `--clear` removes exactly the seeded projects, leaves others alone.
- [x] Friendly error when the daemon is unreachable, with a hint to run `pnpm tools-dev`.

## Worktree Handoff
- **Branch / worktree:** `dawn-herring` at `/Users/pftom/.superset/worktrees/open-claude-design/dawn-herring`.
- **Scope:** add a developer script for seeding test projects; no runtime/app code touched.
- **Files touched:**
  - Added: `scripts/seed-test-projects.ts`
  - Modified: `package.json` (one new entry under `scripts`)
- **Public-surface changes:** new `pnpm seed:test-projects` script. Talks only to the existing public daemon HTTP API (`/api/projects`, `/api/projects/:id/files`, `/api/projects/:id/tabs`, `/api/projects/:id/conversations/:cid/messages/:mid`). No new exports, no new routes, no schema changes.
- **Merge guidance:**
  - `package.json` is hot. Conflicts likely if another worktree adds adjacent `scripts` entries — keep both.
  - Otherwise self-contained; no follow-up regen needed.
- **Verification:** `--help` runs; live daemon round-trip on `http://127.0.0.1:7777` produces 4 well-formed projects (file content, tabs, conversation, messages all verified via API); `--clear` empties the seeded set.
- **Deliberately not done:** no test fixtures for sketches/uploaded-image/multi-file projects, no Playwright e2e seeding hooks. Add later if the gallery needs more variety.

### DIFF
--- a/scripts/seed-test-projects.ts
+++ b/scripts/seed-test-projects.ts
@@ -1,0 +1,306 @@
+#!/usr/bin/env node
+// Seed the running daemon with pre-baked test projects so the UI has
+// real slide decks and web prototypes to work with without waiting for
+// an LLM run. Pulls each project's content straight from a skill's
+// `example.html`, drops it in as `index.html`, and adds a couple of
+// fake chat messages so the conversation panel isn't empty.
+//
+// Usage (daemon must be running — e.g. `pnpm tools-dev`):
+//   pnpm seed:test-projects                    # default bundle
+//   pnpm seed:test-projects --decks 2 --webs 2 # cap counts
+//   pnpm seed:test-projects --daemon http://127.0.0.1:17456
+//   pnpm seed:test-projects --clear            # remove previously seeded projects
+//
+// Seeded project ids start with `seed-` so `--clear` only touches the
+// fixtures this script created.
+
+import { readFile } from 'node:fs/promises';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
+
+const HERE = path.dirname(fileURLToPath(import.meta.url));
+const REPO_ROOT = path.resolve(HERE, '..');
+const SKILLS_DIR = path.join(REPO_ROOT, 'skills');
+const SEED_PREFIX = 'seed-';
+
+type SeedKind = 'deck' | 'prototype';
+
+interface SeedFixture {
+  skillId: string;
+  kind: SeedKind;
+  name: string;
+  pendingPrompt: string;
+  // optional: path to the file inside skills/<skillId>/ to load as index.html
+  // (defaults to example.html)
+  source?: string;
+}
+
+const DECKS: SeedFixture[] = [
+  {
+    skillId: 'html-ppt-pitch-deck',
+    kind: 'deck',
+    name: 'Pitch deck — Series A',
+    pendingPrompt:
+      'Make a 10-slide investor pitch deck for an AI design tool. Cover problem, solution, market, traction, ask.',
+  },
+  {
+    skillId: 'kami-deck',
+    kind: 'deck',
+    name: 'Kami deck — quarterly review',
+    pendingPrompt:
+      'Build a print-grade kami deck summarizing Q2 results: revenue, top wins, risks, next quarter.',
+  },
+  {
+    skillId: 'html-ppt-weekly-report',
+    kind: 'deck',
+    name: 'Weekly report — eng team',
+    pendingPrompt:
+      'Weekly report deck for an engineering team: shipped, in-progress, blockers, next-week plan.',
+  },
+  {
+    skillId: 'html-ppt-product-launch',
+    kind: 'deck',
+    name: 'Product launch — v2.0',
+    pendingPrompt:
+      'Product launch deck for v2.0: hero feature, before/after, pricing, rollout plan.',
+  },
+];
+
+const WEBS: SeedFixture[] = [
+  {
+    skillId: 'open-design-landing',
+    kind: 'prototype',
+    name: 'Editorial landing — Atelier Zero',
+    pendingPrompt:
+      'Single-page editorial landing page for an AI design tool. Magazine collage hero, sticky nav, scroll reveal.',
+  },
+  {
+    skillId: 'kami-landing',
+    kind: 'prototype',
+    name: 'Kami landing — white paper',
+    pendingPrompt:
+      'Print-grade kami landing — parchment canvas, ink-blue accent. Treat it like a studio one-pager.',
+  },
+  {
+    skillId: 'dashboard',
+    kind: 'prototype',
+    name: 'Admin dashboard — analytics',
+    pendingPrompt:
+      'Admin dashboard with KPI cards, a revenue chart, and a recent activity table. Fixed left sidebar.',
+  },
+];
+
+interface Args {
+  daemonUrl: string;
+  decks: number;
+  webs: number;
+  clear: boolean;
+}
+
+function parseArgs(argv: string[]): Args {
+  const out: Args = {
+    daemonUrl:
+      process.env.OD_DAEMON_URL ||
+      `http://127.0.0.1:${process.env.OD_PORT || 7456}`,
+    decks: DECKS.length,
+    webs: WEBS.length,
+    clear: false,
+  };
+  for (let i = 0; i < argv.length; i++) {
+    const a = argv[i];
+    if (a === '--daemon' || a === '--daemon-url') {
+      out.daemonUrl = argv[++i] || out.daemonUrl;
+    } else if (a === '--decks') {
+      out.decks = Math.max(0, Number(argv[++i]) || 0);
+    } else if (a === '--webs' || a === '--prototypes') {
+      out.webs = Math.max(0, Number(argv[++i]) || 0);
+    } else if (a === '--clear') {
+      out.clear = true;
+    } else if (a === '-h' || a === '--help') {
+      printHelp();
+      process.exit(0);
+    } else {
+      console.error(`unknown flag: ${a}`);
+      printHelp();
+      process.exit(2);
+    }
+  }
+  return out;
+}
+
+function printHelp() {
+  console.log(`Usage: pnpm seed:test-projects [opts]
+
+Seeds the running daemon with pre-baked slide decks and web prototypes
+loaded from each skill's example.html. Useful for working on the UI
+without waiting for an LLM run.
+
+Options:
+  --daemon <url>     Daemon base URL (default: $OD_DAEMON_URL or http://127.0.0.1:7456)
+  --decks <n>        Number of slide decks to seed (default: ${DECKS.length}, max: ${DECKS.length})
+  --webs <n>         Number of web prototypes to seed (default: ${WEBS.length}, max: ${WEBS.length})
+  --clear            Delete every previously seeded project (id prefix '${SEED_PREFIX}')
+  -h, --help         Show this help
+
+The script picks up OD_DAEMON_URL / OD_PORT from the env that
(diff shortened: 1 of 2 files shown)


scripts/seed-test-projects.ts
```diff
@@ -0,0 +1,306 @@
...
+      role: 'assistant',
+      content:
+        `Seeded \`index.html\` from \`skills/${fix.skillId}/example.html\` ` +
+        `as a starting point. Open the preview tab to see the rendered ${fix.kind}.`,
+      agentId: 'seed-script',
+      agentName: 'seed-test-projects',
+      runStatus: 'succeeded',
+      startedAt: now,
+      endedAt: now,
+      producedFiles: ['index.html'],
```
mrcfps: The seeded assistant message stores `producedFiles` as a string array here, but the contract consumed by the UI is `ProjectFile[]` (`name`, `size`, `mtime`, `kind`, `mime`, etc.). The daemon persists this unchecked, and `AssistantMessage` later renders each entry with `f.name`, `f.kind`, and `f.size`, so these seeded projects will show broken/undefined produced-file chips instead of the intended `index.html` chip. Please capture the `/api/projects/${id}/files` response (for example `const uploaded = await api<ProjectFileResponse>(...)`) and write `producedFiles: [uploaded.file]`; typing the message payload against the contract would also keep this from regressing. 👍

scripts/seed-test-projects.ts
```diff
@@ -0,0 +1,306 @@
...
+  }
+
+  console.log(`seeding ${decks.length} deck(s) + ${webs.length} web prototype(s) → ${args.daemonUrl}`);
+  for (const fix of [...decks, ...webs]) {
+    try {
+      await seedOne(args.daemonUrl, fix);
+    } catch (err) {
+      console.error(`  ! ${fix.skillId} failed: ${(err as Error).message}`);
+    }
+  }
```
mrcfps, line 420: Catching each `seedOne` failure here only logs the error, so an unreachable daemon or failed fixture still falls through to the final `done` message and exits with status 0. That makes the `pnpm seed:test-projects` command look successful even when it seeded nothing (or only a partial set), which is especially misleading for the unreachable-daemon case called out in the PR test plan. Please accumulate failures and either throw/set `process.exitCode = 1` after the loop, and only print the success message when all requested fixtures were created. 🙏

scripts/seed-test-projects.ts
```diff
@@ -0,0 +1,306 @@
...
+  );
+}
+
+async function clearSeeded(daemonUrl: string): Promise<void> {
+  const { projects } = await api<{ projects: Array<{ id: string }> }>(
+    daemonUrl,
+    'GET',
+    '/api/projects',
+  );
+  const seeded = projects.filter((p) => p.id.startsWith(SEED_PREFIX));
```
mrcfps: Non-blocking: `--clear` currently deletes every project whose id starts with `seed-`, even though this script also writes a stronger marker in project metadata (`seeded: true`, `source: 'seed-test-projects'`). Since project ids are caller-supplied through the public daemon API, a manually-created or future non-fixture project with the same prefix would be removed. Please filter on both the prefix and the metadata marker (or otherwise confirm destructive deletes) so `--clear` only removes projects created by this script.

mrcfps (changes requested):
@pftom Thanks for putting together a handy seeding path for UI work. I found a couple of correctness issues that make the seeded conversation data and failure reporting unreliable, so I’m requesting changes for those before this lands. 🙂


lefarcen (commented):
Re-reviewed the current head (4600ceac) against @mrcfps's blocking findings:

**Addressed:**
- ✅ Finding #2 (failure accumulation): lines 293-307 now accumulate failures and exit nonzero; success message only prints when all fixtures succeed.

**Still needs work (blocking):**
- ❌ Finding #1 (producedFiles contract): scripts/seed-test-projects.ts:211-215 POSTs /api/projects/${id}/files but doesn't capture the response, and line 253 still writes `producedFiles: ['index.html']` as a string array instead of `[uploaded.file]`. The UI will still see broken/undefined file chips.

**Still needs work (non-blocking):**
- ❌ Finding #3 (clear metadata filter): scripts/seed-test-projects.ts:265 still filters solely with `p.id.startsWith(SEED_PREFIX)` without checking `metadata.seeded`/`source`.

Once finding #1 is addressed, this should be good to land. 🙌

scripts/seed-test-projects.ts
```diff
@@ -0,0 +1,314 @@
...
+      role: 'assistant',
+      content:
+        `Seeded \`index.html\` from \`skills/${fix.skillId}/example.html\` ` +
+        `as a starting point. Open the preview tab to see the rendered ${fix.kind}.`,
+      agentId: 'seed-script',
+      agentName: 'seed-test-projects',
+      runStatus: 'succeeded',
+      startedAt: now,
+      endedAt: now,
+      producedFiles: ['index.html'],
```
mrcfps: This assistant message still persists `producedFiles` as a string array, but the shared chat contract expects `ProjectFile[]` (`packages/contracts/src/api/chat.ts`), and `/api/projects/:id/files` already returns that shape as `{ file }`. Because the daemon stores this value unchecked and the web assistant UI later reads fields like `f.name`, `f.kind`, and `f.size`, seeded projects will render an undefined/broken produced-file chip instead of the intended `index.html` chip. Please capture the file upload response (for example, type the POST as `api<ProjectFileResponse>(...)`) and pass `producedFiles: [uploaded.file]` when writing the assistant message. 👍

mrcfps (changes requested):
@pftom I re-reviewed the current head and the new failure accumulation looks good. There is still one seeded-message contract issue that will break the produced-files chip for these fixtures, so I’m requesting changes for that before merge. 🙂


lefarcen (commented):
@pftom Perfect! 🎉

I verified commit `eb509d05` and all the findings from mrcfps's review are now addressed:

✅ **Finding #1 (P1 - producedFiles contract):** Line 280 now correctly passes `producedFiles: [uploaded.file]` using the typed `ProjectFile[]` shape. The `ProjectFile` interface at lines 41-48 matches the contract, and the upload response is properly typed as `api<ProjectFileResponse>`.

✅ **Finding #3 (P2 - --clear filter):** `clearSeeded` (lines 295-302) now requires both the `seed-` prefix AND `metadata.seeded === true` + `metadata.source === 'seed-test-projects'`, preventing accidental deletion of manually-created projects.

No new issues introduced. The seeding script is solid — ready for mrcfps's re-review. 👍

lefarcen (commented):
@pftom Nice fix! 👍

The new commit addresses both findings from the prior reviews:
- ✅ **Finding #1 (P1)**: `producedFiles` now correctly uses `ProjectFile[]` via `[uploaded.file]` (line 280)
- ✅ **Finding #3 (P2)**: `--clear` now filters on both prefix AND metadata markers (lines 295-302)

Code is clean. Ready for mrcfps to re-review when available.

scripts/seed-test-projects.ts
```diff
@@ -0,0 +1,356 @@
...
+  decks: number;
+  webs: number;
+  clear: boolean;
+}
+
+function parseArgs(argv: string[]): Args {
+  const out: Args = {
+    daemonUrl:
+      process.env.OD_DAEMON_URL ||
+      `http://127.0.0.1:${process.env.OD_PORT || 7456}`,
```
mrcfps: This default only works when `OD_DAEMON_URL` or `OD_PORT` is already present in the shell, otherwise it falls back to `127.0.0.1:7456`. The PR usage advertises running `pnpm tools-dev` in one shell and then `pnpm seed:test-projects` in another, but tools-dev starts the daemon with `OD_PORT` set to `0` unless `--daemon-port` is provided, so the actual daemon URL is usually an ephemeral port recorded in tools-dev status rather than exported to this later shell. That makes the default command fail against a correctly running daemon, which breaks the main no-LLM seeding workflow. Please either discover the daemon URL from `pnpm tools-dev status --json`/the sidecar runtime status when these env vars are absent, or require `--daemon` and update the help/PR usage so the default path is not misleading. 🙏

mrcfps (changes requested):
@pftom Thanks for tightening up the prior seeding issues. I found one remaining correctness gap in the default connection path: the advertised `pnpm tools-dev` then `pnpm seed:test-projects` flow can still miss the running daemon when tools-dev chose an ephemeral port, so I’m requesting changes for that before merge. 🙂


lefarcen (commented):
**✅ Daemon URL discovery fixed**

The new `discoverDaemonUrlFromToolsDev()` mechanism fully resolves the ephemeral port workflow issue @mrcfps flagged. The two-shell flow (`pnpm tools-dev` → `pnpm seed:test-projects`) now works correctly:

1. **OD_PORT=0 handling**: `isDiscoverablePort()` treats `0` as unusable and falls through to discovery ✅
2. **Discovery path**: spawns `pnpm tools-dev status --json` and extracts `apps.daemon.url` ✅
3. **Resolution priority**: `--daemon` flag > `OD_DAEMON_URL` > valid `OD_PORT` > discovery > clear error ✅
4. **Edge cases**: catches spawn/parse errors, drains stderr, gracefully degrades ✅

All mrcfps findings (#1 producedFiles contract, #2 exit code, #3 clearSeeded filter, #4 daemon URL) are now resolved. The PR is implementation-complete from the bot review perspective. 🎉

lefarcen (commented):
@pftom Perfect! The daemon URL discovery mechanism fully resolves the ephemeral port workflow issue — `discoverDaemonUrlFromToolsDev()` cleanly handles the `tools-dev` → `seed:test-projects` two-shell flow by reading the actual daemon URL from `pnpm tools-dev status --json` when env vars are absent or unusable.

All prior findings are now addressed:
- ✅ `producedFiles` typed as `ProjectFile[]` (eb509d05)
- ✅ Failure accumulation + non-zero exit (eb509d05) 
- ✅ `--clear` metadata filter (eb509d05)
- ✅ Daemon URL discovery for ephemeral ports (8e115813)

The PR is ready to merge from a code correctness perspective. Nice work on the iteration! 🎉

mrcfps (approved):

PR #405, opened 2026-05-04
feat(design-files): add batch ZIP download with multi-select

dabing1022:
## Summary

- Add checkbox-based multi-select to the Design Files panel
- New `POST /api/projects/:id/archive/batch` endpoint that zips selected files
- ZIP filename uses the project name (with RFC 5987 UTF-8 encoding)
- "Select all" / "Clear" per-section controls in section headers

iShot_2026-05-04_08 48 02
iShot_2026-05-04_08 47 51

## How it works

1. Check the ☐ checkbox on file rows (or use "Select all" in section headers)
2. Header switches to show "Download N as ZIP"
3. Click → server-side JSZip packs selected files → browser downloads `<project-name>.zip`

## Files changed

- **Backend**: `buildBatchArchive()` in `projects.ts` + new POST endpoint in `server.ts`
- **Frontend**: Multi-select UI in `DesignFilesPanel.tsx` + CSS + i18n keys (15 locales)

## Test plan

- [ ] Select files and download — verify ZIP contains correct files
- [ ] Project name with CJK characters shows correctly in downloaded filename
- [ ] "Select all" picks up all visible files in a section
- [ ] "Clear" removes only that section's selections
- [ ] Download button hidden when no files selected
- [ ] Single-file download (existing `<a download>`) still works unchanged

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #465, opened 2026-05-04
fix(design-files): clear selection on project switch

### LINKED ISSUE #464: fix(design-files): clear selection on projectId change (post-merge P1 from #405)
## Context

PR #405 (batch ZIP download) was merged with one unresolved P1 issue from @mrcfps's review:

**P1: Cross-project selection bug**
- **Root cause**: `DesignFilesPanel.tsx:79-94` effect intersects `selected` with current `files`, but does not clear it on `projectId` change.
- **Impact**: If project A and project B both contain `index.html`, switching projects leaves that name selected → `handleBatchDownload()` posts against the new `projectId` even though the user never selected that file in project B.
- **Expected**: Clear `selected` when `projectId` changes.
- **Current**: Only intersection-prune for refresh/delete changes within the same project.

## Fix

```tsx
// apps/web/src/components/DesignFilesPanel.tsx
useEffect(() => {
  // Clear selection on project change
  setSelected(new Set());
}, [projectId]);

// Keep the intersection-prune for refresh/delete within same project
useEffect(() => {
  const validNames = new Set(files.map(f => f.name));
  setSelected(prev => new Set([...prev].filter(name => validNames.has(name))));
}, [files]);
```

## Test plan

1. Select `index.html` in project A
2. Switch to project B (which also has `index.html`)
3. Verify the batch button disappears (selection cleared)
4. Select `index.html` in project B → verify batch downloads B's `index.html`, not A's

## References

- PR #405: https://github.com/nexu-io/open-design/pull/405
- mrcfps review comment: https://github.com/nexu-io/open-design/pull/405#discussion_r2036087095

dabing1022:
## Summary

Fix post-merge P1 bug identified in [#405](https://github.com/nexu-io/open-design/pull/405),
tracked in [#464](https://github.com/nexu-io/open-design/issues/464).

**Bug**: Switching from project A to B, if both contain `index.html`,
the selection survives the intersection-prune. `handleBatchDownload()`
then posts against the new `projectId` — downloading project B's file
without the user selecting it.

**Fix**: Split the single `useEffect` into two:
- `projectId` change → clear `selected` entirely
- `files` change → prune stale entries only

## Test plan
- [ ] Select `index.html` in project A
- [ ] Switch to project B that also has `index.html`
- [ ] Verify selection is cleared, no batch download button shown

Closes #464.

### DIFF
--- a/apps/web/src/components/DesignFilesPanel.tsx
+++ b/apps/web/src/components/DesignFilesPanel.tsx
@@ -76,17 +76,10 @@ export function DesignFilesPanel({
     return groups;
   }, [files]);
 
-  // Clear selection on project switch to prevent cross-project leaks
-  // (e.g. index.html selected in project A persisting into project B).
-  const prevProjectRef = useRef<string | null>(null);
-  useEffect(() => {
-    if (prevProjectRef.current !== null && prevProjectRef.current !== projectId) {
-      setSelected(new Set());
-    }
-    prevProjectRef.current = projectId;
-  }, [projectId]);
-
-  // Prune selections that no longer exist in the current file list.
+  // Prune selections that no longer exist in the current file list
+  // (e.g. after a refresh or delete within the same project).
+  // Cross-project leaks are handled by the parent remounting this
+  // component via key={projectId}.
   useEffect(() => {
     setSelected((prev) => {
       if (prev.size === 0) return prev;
--- a/apps/web/src/components/DesignFilesPanel.tsx
+++ b/apps/web/src/components/DesignFilesPanel.tsx
@@ -76,7 +76,17 @@ export function DesignFilesPanel({
     return groups;
   }, [files]);
 
-  // Prune stale selections when the file list or project changes.
+  // Clear selection on project switch to prevent cross-project leaks
+  // (e.g. index.html selected in project A persisting into project B).
+  const prevProjectRef = useRef<string | null>(null);
+  useEffect(() => {
+    if (prevProjectRef.current !== null && prevProjectRef.current !== projectId) {
+      setSelected(new Set());
+    }
+    prevProjectRef.current = projectId;
+  }, [projectId]);
+
+  // Prune selections that no longer exist in the current file list.
   useEffect(() => {
     setSelected((prev) => {
       if (prev.size === 0) return prev;
@@ -91,7 +101,7 @@ export function DesignFilesPanel({
       }
       return changed ? next : prev;
     });
-  }, [files, projectId]);
+  }, [files]);
 
   const previewFile = useMemo(
     () => files.find((f) => f.name === preview) ?? null,
--- a/apps/web/src/components/FileWorkspace.tsx
+++ b/apps/web/src/components/FileWorkspace.tsx
@@ -377,6 +377,7 @@ export function FileWorkspace({
         {uploadError ? <div className="viewer-empty">{uploadError}</div> : null}
         {activeTab === DESIGN_FILES_TAB ? (
           <DesignFilesPanel
+            key={projectId}
             projectId={projectId}
             files={files}
             onRefreshFiles={onRefreshFiles}


lefarcen:
Hi @dabing1022! 🎉
Thanks for the contribution — this fixes the cross-project selection leak properly by splitting the effect responsibilities.

— open-design team

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -76,7 +76,17 @@ export function DesignFilesPanel({
...
-  // Prune stale selections when the file list or project changes.
+  // Clear selection on project switch to prevent cross-project leaks
+  // (e.g. index.html selected in project A persisting into project B).
+  const prevProjectRef = useRef<string | null>(null);
+  useEffect(() => {
+    if (prevProjectRef.current !== null && prevProjectRef.current !== projectId) {
+      setSelected(new Set());
+    }
+    prevProjectRef.current = projectId;
+  }, [projectId]);
```
mrcfps: This still leaves the stale selection observable for one committed render after `projectId` changes, because passive `useEffect` callbacks run after React has rendered and painted with the new `projectId` but the old `selected` set. During that window the batch-download action can still be rendered with the old project's filenames, and `handleBatchDownload` would submit those names to `/api/projects/${projectId}/archive/batch` for the new project—the same cross-project selection leak this PR is meant to eliminate. Please make the selection project-scoped synchronously instead, for example by storing the selected names with their owning `projectId` and only rendering/using them when it matches the current prop, or by remounting/resetting this panel on project changes and adding a guard in `handleBatchDownload` so stale ownership is unrepresentable. 🙂

mrcfps (changes requested):
Thanks for keeping this focused. I found one remaining blocker in the project-switch cleanup path that can still allow the stale selection to be used before the passive effect clears it.


lefarcen (dismissed):
Hi @dabing1022! Thanks for addressing the post-merge P1. The split-effect approach correctly handles the selection, but deeper inspection found the same cross-project leak pattern in other UI state.

## P1: Row action menu popover also leaks across projects

**Location**: `apps/web/src/components/DesignFilesPanel.tsx` around line 427

The fix only clears `selected`, but `menuPos` is also project-scoped transient state. If a user opens the menu for `index.html` in project A, switches to project B (which also has `index.html`), the still-open popover now builds its download URL with the new `projectId` and calls `onDeleteFile(name)` against the new project through `FileWorkspace`. This is the same cross-project action class as the batch-download bug.

**Concrete fix**: On project switch, reset **all** project-scoped transient state (`selected`, `menuPos`, likely `preview`/`hover`), or remount `DesignFilesPanel` with `key={projectId}`.

---

## P2: First render after projectId change can still leak

**Location**: Line 82 (the new useEffect)

Clearing selection in a passive `useEffect` means the first render after `projectId` changes can still render `selected.size > 0` with the new `projectId`; `handleBatchDownload()` at line 167 would post the stale names to the new project if activated before the effect update lands.

For a P1 cross-project action bug, the guard should be synchronous.

**Concrete fix**: Key/remount the panel by `projectId`, or store selection with its owning `projectId` and derive an empty current selection whenever it doesn't match.

---

## P2: Missing test coverage for the regression path

**Location**: `apps/web/src/components/FileWorkspace.test.tsx`

The existing test is only static markup and cannot exercise selection, rerendering with a new `projectId`, or confirming no batch-download action remains. This is exactly the kind of state-retention bug that can regress silently.

**Concrete fix**: Add a jsdom/client test that:
1. Selects `index.html` in project A
2. Rerenders with project B containing `index.html`
3. Asserts the selected-download UI/action is absent or cannot POST

---

Please address these blockers and I'll re-review. The core fix direction is sound — we just need to ensure all project-scoped UI state is cleared synchronously.

lefarcen (commented):
Thanks @dabing1022! The `key={projectId}` remount approach is exactly right — it eliminates both the async-clear timing window and the menuPos leak synchronously. Much cleaner than tracking prior projectId.

**Remaining P2 (non-blocking)**: Test coverage is still missing for the cross-project regression path. The existing `FileWorkspace.test.tsx` only validates static markup. Adding a test that:
1. Renders with project A + selects a file
2. Changes projectId prop to project B
3. Asserts selection/menu state is cleared

...would prevent silent regressions. Not blocking merge, but would be valuable to add before this class of bug recurs.

mrcfps (approved):
@dabing1022 I reviewed the current head and the `key={projectId}` remount now makes the design-files panel's project-scoped state reset synchronously on project changes. That addresses the stale selected-file and row-menu leak paths from the earlier review while keeping the same-project file-list pruning intact. Thanks for the focused follow-up here — this is a clean and maintainable fix. 🙂


lefarcen (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @dabing1022 for the contribution to open-design. Looking forward to the next one!

Issue #437, opened 2026-05-04
Issues running with Github Copilot


albertasaftei:
seems like it doesn't get the input from the chat. Copilot subscription active and 99% premium requests available.

OS: macOS 26.3
Copilot version: 0.0.353
open-design commit: c881c0ca34a71b640d4e84085227fe87f1524c2b

<img width="440" height="419" alt="Image" src="https://github.com/user-attachments/assets/b7953ea8-c42d-4336-a8be-a95918ec2f52" />

lefarcen:
Hi @albertasaftei and @zackrykalash! 👋 Thanks for the reports.

I see you're both hitting an "Error loading data" message from GitHub Copilot when using it with open-design. A few questions to help narrow this down:

**1. Is this happening inside VS Code, or are you running Copilot CLI directly from a terminal?**  
The screenshot shows the VS Code Copilot chat panel — is open-design configured to spawn Copilot CLI behind the scenes, or are you trying to use the VS Code extension directly?

**2. Does Copilot CLI work standalone?**  
Can you test this in a terminal:
```bash
copilot -p - --allow-all-tools --output-format json <<< 'Say hello'
```

If that fails with the same error, it's a Copilot auth/version issue rather than open-design integration.

**3. Version check:**  
Your Copilot version (0.0.353) is quite old — the latest public release is 1.0.40. Could you try updating:
```bash
npm install -g @githubnext/github-copilot-cli
copilot --version
```

**Background:**  
- Commit `c881c0c` (the one you're on) only touched French localization — no Copilot changes.
- Issue #367 reported similar Copilot stdin errors on Windows; yours is macOS but might be related.
- open-design sends prompts to Copilot CLI via stdin; if the CLI can't auth or load its config, you'd see this kind of generic error.

Let me know what you find — especially the standalone CLI test and version! 🙏

albertasaftei:
I uninstalled and downloaded the correct cli version, now i have 1.0.40

> npm install -g @githubnext/github-copilot-cli

this doesn't seem to be the correct package, i installed @github/copilot

> copilot -p - --allow-all-tools --output-format json <<< 'Say hello'

trying this command results in this:
```
[...]

Issue #367, opened 2026-05-03
github copilot cli - "stdin: write EOF" error


imsay3m:
System found my copilot cli model but when sending a prompt it reply with 

<img width="542" height="339" alt="Image" src="https://github.com/user-attachments/assets/7e1983a0-0bb1-493d-a8c8-08dc0788b471" />

"
I don’t have an actionable task yet—the last message only contains -. Please send the specific change or investigation you want done.
stdin: write EOF
" 
whenever I checked the history I just find the input prompt as (below image)
<img width="365" height="273" alt="Image" src="https://github.com/user-attachments/assets/e01f8626-7265-41e9-a189-5645679a509b" />

my system is: WIN11

lefarcen:
Hi @imsay3m! 👋 Thanks for opening this issue — I can see the exact error message in your screenshots.

This looks like issue #350 ("too many arguments" regression with GitHub Copilot CLI), which was just fixed in PR #351 and merged ~15 hours ago.

**The fix:** Open Design now correctly passes `-p -` to copilot CLI so it reads the prompt from stdin instead of staying in interactive mode.

Could you:
1. Update to the latest build: `git pull && pnpm install && pnpm dev:all` (or your equivalent)
2. Try sending a prompt again
3. Let us know if the error persists?

If it still fails after updating, please share:
- Your Open Design version (`git rev-parse HEAD` or the version shown in About)
- Your GitHub Copilot CLI version (`copilot --version`)

Thanks for the report!

Leonardo-marinhoo:
Hi !

Im facing something similar. 
Copilot CLI and Openclaude returning this same response in any project with no error messages. 

<img width="1907" height="972" alt="Image" src="https://github.com/user-attachments/assets/ede4a107-89a6-4229-92b9-a6760c1ae7c7" />

lefarcen:
Hi @Leonardo-marinhoo! 👋 Thanks for confirming this — you're seeing the exact same issue.

The good news: this was just fixed in PR #351 (merged ~16 hours ago). Both Copilot CLI and OpenClaude now correctly pass `-p -` so they read prompts from stdin instead of staying in interactive mode.

**Next steps:**
1. Update to the latest build: `git pull && pnpm install && pnpm dev:all`
2. Try sending a prompt again
3. Let us know if the error persists

If it still fails after updating, please share:
- Your Open Design version (`git rev-parse HEAD`)
- Your Copilot CLI version (`copilot --version`)
[...]

PR #466, opened 2026-05-04
fix(agents): copilot prompt processing with correct command format

albertasaftei:
Fixed copilot prompt processing as saw earlier in issue [#437](https://github.com/nexu-io/open-design/issues/437) to use correct command format as tested:

`copilot -p "message" --allow-all-tools --output-format json`

This will likely fix the issue reported in [#367](https://github.com/nexu-io/open-design/issues/367) as well

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -449,15 +449,11 @@ export const AGENT_DEFS = [
     name: 'GitHub Copilot CLI',
     bin: 'copilot',
     versionArgs: ['--version'],
-    // `-p -` enters Copilot's prompt mode and tells the CLI to read the
-    // prompt body from stdin instead of expecting it as a positional argv
-    // element. Without it the daemon writes the prompt to the child's
-    // stdin pipe (because `promptViaStdin: true` below) but Copilot stays
-    // in interactive mode, never reads stdin, and rejects the run with
-    // `error: too many arguments. Expected 0 arguments but got N` —
-    // the regression filed in #350. PR #258 standardized agents on stdin
-    // delivery and dropped the per-prompt argv path, but missed flipping
-    // Copilot's mode from interactive to `-p -`.
+    // The prompt is passed directly as the value of `-p`: `copilot -p
+    // "<prompt>" --allow-all-tools --output-format json`. Copilot does NOT
+    // treat `-` as a stdin sentinel — it reads it as a literal one-character
+    // prompt string — so the previous `-p -` + stdin pattern produced a
+    // nonsensical single-dash prompt instead of the composed prompt body.
     //
     // `--allow-all-tools` is required for non-interactive runs: without it
     // the CLI blocks waiting for human approval on every tool call. Unlike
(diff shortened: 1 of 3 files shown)


lefarcen:
Hi @albertasaftei! 🎉
Thanks for the contribution — this fixes the exact root cause confirmed in #437 (Copilot's `-p -` not reading stdin).

— open-design team

mrcfps (approved):
@albertasaftei I reviewed the Copilot adapter diff, including the switch from stdin delivery to passing the composed prompt as the `-p` value, the argument ordering with model/add-dir flags, and the updated daemon tests around that shape. I did not find any changed-range correctness or safety issues that need action; the change is focused and the regression coverage tracks the intended command format nicely. Thanks for tightening this path up — this should make the Copilot runner behavior much clearer.


apps/daemon/src/agents.ts
```diff
@@ -499,7 +495,7 @@ export const AGENT_DEFS = [
       for (const d of dirs) args.push('--add-dir', d);
       return args;
     },
-    promptViaStdin: true,
+    promptViaStdin: false,
     streamFormat: 'copilot-stream-json',
   },
   {
```
lefarcen, line 501: **P2 Argv length budget**

Now that prompts go through argv instead of stdin, this adapter is vulnerable to Windows' ~32KB CreateProcess limit and Linux MAX_ARG_STRLEN on very long prompts (large skill context + long design systems).

The DeepSeek adapter above (line 665) has the same constraint and uses `maxPromptArgBytes: 30_000` to guard it. The daemon checks this before spawn and gives users an actionable error ("reduce skills/context or use stdin adapter") instead of letting spawn fail with cryptic ENAMETOOLONG.

Suggestion: Add the same guard here:
```typescript
maxPromptArgBytes: 30_000,
```

This is non-blocking for merge (the spawn will fail loudly if hit), but would improve the error UX for power users with huge context.

lefarcen (commented):
Hi @albertasaftei! This fix correctly addresses the confirmed root cause from #437 — Copilot CLI doesn't read stdin with `-p -`, so passing the prompt as a direct argument is the right approach.

**Code correctness (Lens A):** ✅ No security issues. The spawn call uses `shell: false` with array args, so Node.js handles escaping automatically — no injection risk even with quotes/backticks in prompts.

**One P2 concern:** Argv length limits. See inline comment below.

**Test coverage:** ✅ Tests correctly verify the new behavior (prompt as `-p <value>` not stdin sentinel).

**Reasoning completeness (Lens B):** The updated comment accurately explains the root cause and references the actual behavior confirmed in #437. Good documentation.

Ready for merge after addressing the argv length concern below. 👍

lefarcen (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @albertasaftei for the contribution to open-design. Looking forward to the next one!

lefarcen:
🎉 Merged! Thanks @albertasaftei for fixing this — your thorough diagnostics in #437 made it possible to confirm the root cause and ship the correct fix.

The fix is now live on main. I've pinged the reporters on #437 and #367 to verify. 🙏

leilei926524-tech:
Hi @albertasaftei — welcome to Open Design contributors. 🎉

Your first merged PR (#466) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #483, opened 2026-05-05
fix: path traversal in readDesignSystem

hobostay:
## Summary
- Validates the `id` parameter in `readDesignSystem()` to reject path traversal attempts

## Problem
`readDesignSystem()` in `apps/daemon/src/design-systems.ts` passes the `id` parameter directly into `path.join(root, id, 'DESIGN.md')` without any validation. An attacker can supply `id = "../../etc"` to read arbitrary files on disk. Unlike `projects.ts` which uses `isSafeId()` validation, this function had no input sanitization.

## Fix
Add a guard that rejects ids containing `/` or `\` path separators, or dot segments (`..`, `.`).

## Test plan
- [ ] `GET /api/design-systems/..%2F..%2Fetc` should return null or error
- [ ] `GET /api/design-systems/valid-id` should still return DESIGN.md content

### DIFF
--- a/apps/daemon/src/design-systems.ts
+++ b/apps/daemon/src/design-systems.ts
@@ -6,6 +6,7 @@
 
 import { readdir, readFile, stat } from 'node:fs/promises';
 import path from 'node:path';
+import { isSafeResourceId } from './resource-id.js';
 
 export async function listDesignSystems(root) {
   const out = [];
@@ -41,9 +42,7 @@ export async function listDesignSystems(root) {
 }
 
 export async function readDesignSystem(root, id) {
-  if (!id || typeof id !== 'string' || /[/\\]/.test(id) || id === '..' || id === '.') {
-    return null;
-  }
+  if (!isSafeResourceId(id)) return null;
   const file = path.join(root, id, 'DESIGN.md');
   try {
     return await readFile(file, 'utf8');
--- a/apps/daemon/src/design-systems.ts
+++ b/apps/daemon/src/design-systems.ts
@@ -41,6 +41,9 @@ export async function listDesignSystems(root) {
 }
 
 export async function readDesignSystem(root, id) {
+  if (!id || typeof id !== 'string' || /[/\\]/.test(id) || id === '..' || id === '.') {
+    return null;
+  }
   const file = path.join(root, id, 'DESIGN.md');
   try {
     return await readFile(file, 'utf8');
--- a/apps/daemon/tests/static-resource-paths.test.ts
+++ b/apps/daemon/tests/static-resource-paths.test.ts
@@ -1,0 +1,63 @@
+import { mkdir, mkdtemp, rm, writeFile } from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
+
+import { readDesignSystem } from '../src/design-systems.js';
+import { readPromptTemplate } from '../src/prompt-templates.js';
+
+let tempRoot: string;
+
+beforeEach(async () => {
+  tempRoot = await mkdtemp(path.join(os.tmpdir(), 'od-static-resource-paths-'));
+});
+
+afterEach(async () => {
+  await rm(tempRoot, { recursive: true, force: true });
+});
+
+describe('static resource path guards', () => {
+  it('rejects design-system ids that are not single safe path segments', async () => {
+    const root = path.join(tempRoot, 'design-systems');
+    await mkdir(path.join(root, 'demo'), { recursive: true });
+    await writeFile(path.join(root, 'demo', 'DESIGN.md'), '# Demo\n');
+    await writeFile(path.join(tempRoot, 'DESIGN.md'), '# Outside\n');
+
+    await expect(readDesignSystem(root, 'demo')).resolves.toBe('# Demo\n');
+    await expect(readDesignSystem(root, '..')).resolves.toBeNull();
+    await expect(readDesignSystem(root, '...')).resolves.toBeNull();
+    await expect(readDesignSystem(root, '../outside')).resolves.toBeNull();
+    await expect(readDesignSystem(root, 'nested/demo')).resolves.toBeNull();
+  });
+
+  it('rejects prompt-template ids that can traverse out of their surface folder', async () => {
+    const root = path.join(tempRoot, 'prompt-templates');
+    await mkdir(path.join(root, 'image'), { recursive: true });
+    await writeFile(
+      path.join(root, 'image', 'demo.json'),
+      JSON.stringify({
+        id: 'demo',
+        surface: 'image',
+        title: 'Demo prompt',
+        prompt: 'A detailed prompt body long enough to be accepted.',
+        source: { repo: 'open-design/test', license: 'MIT' },
+      }),
+    );
+    await writeFile(
+      path.join(root, 'secret.json'),
+      JSON.stringify({
+        id: 'secret',
+        surface: 'image',
+        title: 'Secret prompt',
+        prompt: 'This template lives outside the image folder.',
+        source: { repo: 'open-design/test', license: 'MIT' },
+      }),
+    );
+
+    await expect(readPromptTemplate(root, 'image', 'demo')).resolves.toMatchObject({
+      id: 'demo',
+    });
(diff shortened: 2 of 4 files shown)


lefarcen:
Hi @hobostay! 🎉
Thanks for the contribution — this addresses a critical path traversal vulnerability in readDesignSystem().

— open-design team

apps/daemon/src/design-systems.ts
```diff
@@ -41,6 +41,9 @@ export async function listDesignSystems(root) {
 }
 
 export async function readDesignSystem(root, id) {
+  if (!id || typeof id !== 'string' || /[/\\]/.test(id) || id === '..' || id === '.') {
```
mrcfps: Could you add a regression test for this guard? The changed lines reject decoded `/` and `\` separators plus `.`/`..`, but there isn't any test coverage in this PR for either the traversal case or the allowed-id case from the PR test plan. Because this is the security boundary in front of `path.join(root, id, 'DESIGN.md')`, a future refactor could accidentally relax it without CI catching the regression. A focused daemon test for either `readDesignSystem()` or the `/api/design-systems/:id` route should cover an encoded traversal such as `..%2F..%2Fetc` returning 404/null and a valid design-system id still returning `DESIGN.md`. 🙂

mrcfps (commented):
Thanks for tightening this security boundary around design-system reads. I found one merge-safe follow-up around regression coverage so this traversal fix stays protected in CI.


apps/daemon/src/design-systems.ts
lefarcen: **P2** This security fix has no regression coverage. The daemon only has design-system showcase parser tests, not tests for `readDesignSystem()` or the `/api/design-systems/:id` route traversal cases.

Because Express decodes `%2F`/`%5C` route params before passing them here, please add daemon tests that assert:
- `..%2F..%2Fetc` → 404/null
- `..%5C..%5Cetc` → 404/null  
- `.` and `..` → 404/null
- Valid design-system id → success

This prevents future regressions when refactoring this validation.

lefarcen: **P3** Consider extracting this validation into a shared `isSafeResourceId()` helper (or reusing the existing `isSafeId()` pattern from `projects.ts:446` if compatible).

Right now we have duplicated validation logic that's easy to miss when adding new resource readers. A centralized validator would make the security boundary more obvious and maintainable.

lefarcen (changes requested):
Hey @hobostay, thanks for catching this path traversal vulnerability! 🔒

The fix correctly blocks the immediate threat in `readDesignSystem()`, but I found two P2 issues that need addressing before merge:

1. **Same vulnerability exists in `readPromptTemplate()`** — needs the same validation
2. **Missing test coverage** — security fixes should have regression tests

See inline comments below for details.

lefarcen:
One more critical finding outside this PR's diff:

**P2: Same vulnerability in `readPromptTemplate()`**

`apps/daemon/src/prompt-templates.ts:55` has the same issue — `readPromptTemplate(root, surface, id)` validates `surface` but passes unvalidated `id` into `path.join(root, surface, \`${id}.json\`)`.

An encoded slash in `/api/prompt-templates/:surface/:id` can decode to `../..` and read outside the intended directory. While the response is schema-filtered by `validateTemplate()` (less directly exfiltrating than `readDesignSystem`), it still performs attacker-controlled reads/parses and could disclose any JSON matching the template schema.

Would you mind including that fix in this PR (or opening a follow-up)? Ideally with the same shared validation helper to prevent future similar bugs.

PerishCode:
@hobostay — friendly bump: lefarcen left review notes here about 6 days back and the PR's been waiting on your response since. Anything blocking the next pass? Happy to clarify points in the path-traversal review if useful; otherwise a "still working on it" or "abandoning" note here keeps the queue honest.

(Via maintainer `awaiting-author-response-24h` queue scan.)

PerishCode:
I pushed a maintainer edit in 9477de2f to cover the remaining static-resource path issue from the review thread. The update:

- adds a shared `isSafeResourceId` guard for single-segment resource ids;
- applies it to both `readDesignSystem()` and `readPromptTemplate()`;
- adds regression coverage for design-system pure-dot/traversal ids and prompt-template ids like `../secret`.

Validation from this branch:
- `pnpm --filter @open-design/daemon test -- static-resource-paths.test.ts` completed with 35 test files / 479 tests passing on this branch.
- `git diff --check` passed.
- `pnpm --filter @open-design/daemon typecheck` still fails on the branch with the existing `DaemonStatusSnapshot.desktopAuthGateActive` mismatch in `apps/daemon/sidecar/server.ts`, which is outside this resource-id fix.

The PR is still `DIRTY`, so it will need a rebase before the final review/merge path.

mrcfps:
@hobostay I'm holding off on generating review comments for nexu-io/open-design#483 because this pull request has merge conflicts right now.

mrcfps (commented):
Thanks for pushing the shared resource-id guard and the follow-up coverage, @hobostay. I found one merge-safe test issue in the new regression suite so the design-system traversal case is actually exercised in CI.


apps/daemon/tests/static-resource-paths.test.ts
```diff
@@ -0,0 +1,63 @@
...
+    const root = path.join(tempRoot, 'design-systems');
+    await mkdir(path.join(root, 'demo'), { recursive: true });
+    await writeFile(path.join(root, 'demo', 'DESIGN.md'), '# Demo\n');
+    await writeFile(path.join(tempRoot, 'DESIGN.md'), '# Outside\n');
+
+    await expect(readDesignSystem(root, 'demo')).resolves.toBe('# Demo\n');
+    await expect(readDesignSystem(root, '..')).resolves.toBeNull();
+    await expect(readDesignSystem(root, '...')).resolves.toBeNull();
+    await expect(readDesignSystem(root, '../outside')).resolves.toBeNull();
+    await expect(readDesignSystem(root, 'nested/demo')).resolves.toBeNull();
```
mrcfps, line 30: In `apps/daemon/tests/static-resource-paths.test.ts:24-30`, the traversal fixture is in the wrong place for the path you are asserting. `path.join(root, "../outside", "DESIGN.md")` resolves to `<tempRoot>/outside/DESIGN.md`, but the test only creates `<tempRoot>/DESIGN.md`. Because that target file does not exist, `readDesignSystem(root, "../outside")` would still return `null` on the old vulnerable implementation, so this assertion does not actually lock in the regression you are fixing here. Could you move the fixture to `path.join(tempRoot, "outside", "DESIGN.md")` (or change the test id to match the file you created) so the pre-fix code would read attacker-controlled content and the post-fix code proves the guard blocks it?

apps/daemon/tests/static-resource-paths.test.ts
```diff
@@ -0,0 +1,63 @@
...
+  it('rejects design-system ids that are not single safe path segments', async () => {
+    const root = path.join(tempRoot, 'design-systems');
+    await mkdir(path.join(root, 'demo'), { recursive: true });
+    await writeFile(path.join(root, 'demo', 'DESIGN.md'), '# Demo\n');
+    await writeFile(path.join(tempRoot, 'DESIGN.md'), '# Outside\n');
+
+    await expect(readDesignSystem(root, 'demo')).resolves.toBe('# Demo\n');
+    await expect(readDesignSystem(root, '..')).resolves.toBeNull();
+    await expect(readDesignSystem(root, '...')).resolves.toBeNull();
+    await expect(readDesignSystem(root, '../outside')).resolves.toBeNull();
```
lefarcen, line 29: P2 This traversal assertion does not currently prove the old vulnerable behavior. `readDesignSystem(root, '../outside')` would resolve to `<tempRoot>/outside/DESIGN.md`, but the fixture above writes `<tempRoot>/DESIGN.md`, so the pre-fix implementation would still return `null` because the target file is missing. Please move the outside fixture to `path.join(tempRoot, 'outside', 'DESIGN.md')` (or adjust the id to match the file you create) so the test fails before the guard and passes after it.

lefarcen (changes requested):
Hey @hobostay — the maintainer edit moves this in the right direction: the shared resource-id guard now covers both design systems and prompt templates, which addresses the main security shape. I still need one regression-test issue fixed before I can clear my earlier changes-requested review; see the inline note below. Once the test proves the pre-fix traversal would have read attacker-controlled content, this should be much easier to re-check.

PR #485, opened 2026-05-05
fix: skip symlinks in file listing and archive generation

hobostay:
## Summary
- Adds `lstat` checks to skip symbolic links in `collectFiles()` and `collectArchiveEntries()`

## Problem
Both `collectFiles` (used by `GET /api/projects/:id/files`) and `collectArchiveEntries` (used by project archive ZIP downloads) use `Dirent.isDirectory()` which follows symlinks. A symlink placed inside a project directory that points to an arbitrary location (e.g. `/etc`) would be followed, leaking files outside the project tree in file listings and archive downloads.

## Fix
Use `lstat` to detect symbolic links before recursing or including entries. Symlinks are silently skipped.

## Test plan
- [ ] Create a symlink inside a project directory: `ln -s /etc/project-secrets .od/projects/<id>/leak`
- [ ] `GET /api/projects/<id>/files` should NOT include files from the symlink target
- [ ] Archive download should NOT include files from the symlink target

### DIFF
--- a/apps/daemon/src/projects.ts
+++ b/apps/daemon/src/projects.ts
@@ -7,7 +7,7 @@
 // All paths flowing in from HTTP handlers are validated against the project
 // directory to prevent path traversal — see resolveSafe().
 
-import { lstat, mkdir, readdir, readFile, rm, stat, unlink, writeFile } from 'node:fs/promises';
+import { lstat, mkdir, readdir, readFile, realpath, rm, stat, unlink, writeFile } from 'node:fs/promises';
 import path from 'node:path';
 import JSZip from 'jszip';
 import {
@@ -56,8 +56,9 @@ async function collectFiles(dir, relDir, out) {
     try {
       const ls = await lstat(full);
       if (ls.isSymbolicLink()) continue;
-    } catch {
-      continue;
+    } catch (err) {
+      if (err && err.code === 'ENOENT') continue;
+      throw err;
     }
     const rel = relDir ? `${relDir}/${e.name}` : e.name;
     if (e.isDirectory()) {
@@ -93,7 +94,7 @@ export async function buildProjectArchive(projectsRoot, projectId, root) {
   let archiveRoot = projectRoot;
   let archiveBaseName = '';
   if (typeof root === 'string' && root.trim().length > 0) {
-    archiveRoot = resolveSafe(projectRoot, root);
+    archiveRoot = await resolveSafeReal(projectRoot, root);
     archiveBaseName = path.basename(archiveRoot);
   }
 
@@ -278,8 +279,9 @@ async function collectArchiveEntries(dir, relDir, out) {
     try {
       const ls = await lstat(full);
       if (ls.isSymbolicLink()) continue;
-    } catch {
-      continue;
+    } catch (err) {
+      if (err && err.code === 'ENOENT') continue;
+      throw err;
     }
     const rel = relDir ? `${relDir}/${e.name}` : e.name;
     if (e.isDirectory()) {
@@ -294,10 +296,11 @@ async function collectArchiveEntries(dir, relDir, out) {
 
 export async function readProjectFile(projectsRoot, projectId, name) {
   const dir = projectDir(projectsRoot, projectId);
-  const file = resolveSafe(dir, name);
+  const rootReal = await realpath(dir).catch(() => dir);
+  const file = await resolveSafeReal(dir, name);
   const buf = await readFile(file);
   const st = await stat(file);
-  const rel = toProjectPath(path.relative(dir, file));
+  const rel = toProjectPath(path.relative(rootReal, file));
   const manifest = await readManifestForPath(dir, rel);
   return {
     buffer: buf,
@@ -395,6 +398,22 @@ function resolveSafe(dir, name) {
     throw new Error('path escapes project dir');
   }
   return target;
+}
+
+async function resolveSafeReal(dir, name) {
+  const target = resolveSafe(dir, name);
+  const rootReal = await realpath(dir).catch(() => dir);
+  let targetReal;
+  try {
+    targetReal = await realpath(target);
+  } catch (err) {
+    if (err && err.code === 'ENOENT') return target;
+    throw err;
+  }
+  if (!targetReal.startsWith(rootReal + path.sep) && targetReal !== rootReal) {
+    throw new Error('path escapes project dir via symlink');
+  }
+  return targetReal;
 }
 
 export function sanitizePath(raw) {
--- a/apps/daemon/src/projects.ts
+++ b/apps/daemon/src/projects.ts
@@ -7,7 +7,8 @@
 // All paths flowing in from HTTP handlers are validated against the project
 // directory to prevent path traversal — see resolveSafe().
 
-import { lstat, mkdir, readdir, readFile, realpath, rm, stat, unlink, writeFile } from 'node:fs/promises';
+import { constants } from 'node:fs';
+import { lstat, mkdir, open, readdir, readFile, realpath, rm, stat, unlink, writeFile } from 'node:fs/promises';
 import path from 'node:path';
 import JSZip from 'jszip';
 import {
@@ -129,9 +130,9 @@ export async function buildProjectArchive(projectsRoot, projectId, root) {
 
   const zip = new JSZip();
   for (const entry of entries) {
-    const buf = await readFile(entry.fullPath);
+    const { buffer: buf, mtime } = await readArchiveEntry(entry.fullPath);
     zip.file(entry.relPath, buf, {
-      date: new Date(entry.mtime),
+      date: new Date(mtime),
       binary: true,
     });
   }
@@ -291,6 +292,27 @@ async function collectArchiveEntries(dir, relDir, out) {
     if (e.name.endsWith('.artifact.json')) continue;
     const st = await stat(full);
     out.push({ relPath: rel, fullPath: full, mtime: st.mtimeMs });
+  }
+}
+
+async function readArchiveEntry(filePath) {
+  const beforeOpen = await lstat(filePath);
+  if (beforeOpen.isSymbolicLink()) {
+    throw new Error('archive entry is a symlink');
+  }
+  if (!beforeOpen.isFile()) {
+    throw new Error('archive entry is not a regular file');
+  }
+  const noFollow = typeof constants.O_NOFOLLOW === 'number' ? constants.O_NOFOLLOW : 0;
+  const handle = await open(filePath, constants.O_RDONLY | noFollow);
+  try {
+    const opened = await handle.stat();
+    if (!opened.isFile()) {
+      throw new Error('archive entry is not a regular file');
+    }
+    return { buffer: await handle.readFile(), mtime: opened.mtimeMs };
+  } finally {
+    await handle.close();
   }
 }
 
--- a/apps/daemon/src/projects.ts
+++ b/apps/daemon/src/projects.ts
@@ -52,8 +52,14 @@ async function collectFiles(dir, relDir, out) {
   }
   for (const e of entries) {
     if (e.name.startsWith('.')) continue;
-    const rel = relDir ? `${relDir}/${e.name}` : e.name;
     const full = path.join(dir, e.name);
+    try {
+      const ls = await lstat(full);
+      if (ls.isSymbolicLink()) continue;
+    } catch {
+      continue;
+    }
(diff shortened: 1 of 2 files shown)


lefarcen:
Hi @hobostay! 🎉

Thanks for the contribution — this is a critical security fix for symlink directory traversal.

— open-design team

apps/daemon/src/projects.ts
```diff
@@ -268,8 +274,14 @@ async function collectArchiveEntries(dir, relDir, out) {
   for (const e of entries) {
     if (e.name.startsWith('.')) continue;
     if (!e.isDirectory() && !e.isFile()) continue;
-    const rel = relDir ? `${relDir}/${e.name}` : e.name;
     const full = path.join(dir, e.name);
+    try {
+      const ls = await lstat(full);
```
mrcfps, line 281: `collectArchiveEntries()` now checks each child with `lstat(full)`, but this still misses the case where the archive root itself is a symlink. `buildProjectArchive()` accepts a `root` parameter, resolves it under the project, then `stat(archiveRoot)` follows symlinks before calling this walker. If `root` is `leak` and `.od/projects/<id>/leak -> /etc`, `readdir(dir)` has already crossed into `/etc`, and `lstat(path.join(dir, child))` reports the external child (for example `/etc/passwd`) rather than the parent `leak` symlink. That means a targeted archive download can still include files outside the project tree, so the security fix does not fully satisfy the PR goal.

Please reject a symlink archive root before the initial `stat`/`readdir` (for example `lstat(archiveRoot)` and fail/skip if `isSymbolicLink()`), and consider having `collectArchiveEntries()` validate the current directory before reading it so every recursive entry point enforces the same invariant. A fixture covering `buildProjectArchive(projectRoot, id, 'leak')` where `leak` points outside the project would lock this down. 🔒

mrcfps (changes requested):
@hobostay thanks for tightening the project file/archive traversal path. I found one remaining symlink-root escape in the archive flow that keeps the security fix incomplete for targeted ZIP downloads.


lefarcen (changes requested):
Hey @hobostay, thank you for catching this critical symlink traversal vulnerability! 🔒

The fix correctly blocks symlinks in the recursive walkers ( and ), but there are **three additional attack surfaces** that still follow symlinks:

## 🔒 P1 Issues (must fix before merge)

### 1. Archive root itself can be a symlink
**Location**: `apps/daemon/src/projects.ts:106` in `buildProjectArchive()`

`stat(archiveRoot)` follows `leak -> /etc`, accepts it as a directory, then `collectArchiveEntries()` walks the target's real children and zips them.

**Attack**: `/api/projects/:id/archive?root=leak` where `leak -> /etc`

**Fix**: Use `lstat(archiveRoot)` before recursing and reject `isSymbolicLink()` roots. Add a test for this case — should reject with 400.

---

### 2. Direct file reads still follow symlinks
**Location**: `apps/daemon/src/projects.ts:298` in `readProjectFile()`

`readProjectFile()` only does lexical `resolveSafe()`, then `readFile()` and `stat()` follow `leak -> /etc/passwd`.

**Attack**:
- `/api/projects/:id/raw/leak`
- `/api/projects/:id/files/leak`

These can still leak files even though listings now hide the symlink.

**Fix**: Centralize "safe regular project file" resolution that `lstat`s each segment, rejects symlinks, and opens the final file with no-follow semantics where possible (e.g., `fs.openSync` with `O_NOFOLLOW` flag on systems that support it).

---

### 3. TOCTOU race in archive packing
**Location**: `apps/daemon/src/projects.ts:131` in archive packing logic

`collectArchiveEntries()` `lstat`s a path, stores `fullPath`, then later `readFile(entry.fullPath)` follows whatever is there at read time. A symlink swap between collection and packing can escape the project.

**Fix**: Revalidate and open each file with no-follow immediately before reading, or return/open file handles instead of unchecked paths.

---

## ⚠️ P3 Issue (nice to have)

### Error handling in lstat guards
**Location**: `apps/daemon/src/projects.ts:56` and `:280` (the new try/catch blocks)

The new `lstat` guards silently skip *every* `lstat` failure, not just vanished files or symlinks. That can turn `EPERM`/`EACCES`/transient filesystem errors into incomplete successful listings/archives.

**Suggestion**: Silent skip for symlinks is reasonable; for errors, catch `ENOENT` specifically for races and otherwise throw or log:

```typescript
try {
  const ls = await lstat(full);
  if (ls.isSymbolicLink()) continue;
} catch (err) {
  if (err.code === 'ENOENT') continue; // file vanished, skip
  throw err; // propagate permission/IO errors
}
```

---

## 📋 Test Coverage

**Location**: `apps/daemon/tests/project-archive.test.ts`

The security fix is untested. Existing archive tests cover filtering dotfiles and sidecars, but not symlinks.

**Add tests**:
- Create a symlink to an external temp directory/file
- Assert listing omits it
- Assert full archive omits it
- Assert scoped archive rejects it (root symlink case)
- Assert direct reads reject it (once fixed)

---

Once these are addressed, this will be a solid security hardening. Thanks again for finding this! 🙏

lefarcen (changes requested):
Hey @hobostay, thank you for catching this critical symlink traversal vulnerability! 🔒

The fix correctly blocks symlinks in the recursive walkers (collectFiles and collectArchiveEntries), but there are **three additional attack surfaces** that still follow symlinks:

## 🔒 P1 Issues (must fix before merge)

### 1. Archive root itself can be a symlink
**Location**: apps/daemon/src/projects.ts:106 in buildProjectArchive()

stat(archiveRoot) follows leak -> /etc, accepts it as a directory, then collectArchiveEntries() walks the target's real children and zips them.

**Attack**: /api/projects/:id/archive?root=leak where leak -> /etc

**Fix**: Use lstat(archiveRoot) before recursing and reject isSymbolicLink() roots. Add a test for this case — should reject with 400.

---

### 2. Direct file reads still follow symlinks
**Location**: apps/daemon/src/projects.ts:298 in readProjectFile()

readProjectFile() only does lexical resolveSafe(), then readFile() and stat() follow leak -> /etc/passwd.

**Attack**:
- /api/projects/:id/raw/leak
- /api/projects/:id/files/leak

These can still leak files even though listings now hide the symlink.

**Fix**: Centralize "safe regular project file" resolution that lstats each segment, rejects symlinks, and opens the final file with no-follow semantics where possible (e.g., fs.openSync with O_NOFOLLOW flag on systems that support it).

---

### 3. TOCTOU race in archive packing
**Location**: apps/daemon/src/projects.ts:131 in archive packing logic

collectArchiveEntries() lstats a path, stores fullPath, then later readFile(entry.fullPath) follows whatever is there at read time. A symlink swap between collection and packing can escape the project.

**Fix**: Revalidate and open each file with no-follow immediately before reading, or return/open file handles instead of unchecked paths.

---

## ⚠️ P3 Issue (nice to have)

### Error handling in lstat guards
**Location**: apps/daemon/src/projects.ts:56 and :280 (the new try/catch blocks)

The new lstat guards silently skip *every* lstat failure, not just vanished files or symlinks. That can turn EPERM/EACCES/transient filesystem errors into incomplete successful listings/archives.

**Suggestion**: Silent skip for symlinks is reasonable; for errors, catch ENOENT specifically for races and otherwise throw or log:

```typescript
try {
  const ls = await lstat(full);
  if (ls.isSymbolicLink()) continue;
} catch (err) {
  if (err.code === 'ENOENT') continue; // file vanished, skip
  throw err; // propagate permission/IO errors
}
```

---

## 📋 Test Coverage

**Location**: apps/daemon/tests/project-archive.test.ts

The security fix is untested. Existing archive tests cover filtering dotfiles and sidecars, but not symlinks.

**Add tests**:
- Create a symlink to an external temp directory/file
- Assert listing omits it
- Assert full archive omits it
- Assert scoped archive rejects it (root symlink case)
- Assert direct reads reject it (once fixed)

---

Once these are addressed, this will be a solid security hardening. Thanks again for finding this! 🙏

PerishCode:
@hobostay — checking in: this one's been waiting on your follow-up to the symlink-handling feedback for about 6 days now. Still on your radar? I can take another pass through the review thread with you if anything's unclear; if you'd rather drop it, let me know and we'll close it.

(Tagged `awaiting-author-response-24h` by maintainer tooling.)

PerishCode:
I pushed maintainer edits through 506fc3e6 to address the symlink-handling blockers from the review thread. The branch now:

- preserves the recursive walker symlink skip but only suppresses `ENOENT` lstat races; other lstat errors now propagate;
- resolves scoped archive roots through realpath containment so `archive?root=<symlink>` cannot escape the project;
- resolves direct `readProjectFile()` paths through the same realpath containment;
- revalidates archive entries immediately before packing and opens them with `O_NOFOLLOW` when the platform exposes it;
- adds project archive regression coverage for full-archive symlink omission, scoped archive-root escape, and direct-read escape.

Validation:
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/project-archive.test.ts` passed: 9 tests.
- `git diff --check` passed.
- `pnpm --filter @open-design/daemon typecheck` still fails on this old branch with the existing `DaemonStatusSnapshot.desktopAuthGateActive` mismatch in `apps/daemon/sidecar/server.ts`, outside this symlink patch.
- A broader `pnpm --filter @open-design/daemon test -- project-archive.test.ts` invocation runs the whole old-branch daemon suite here; `project-archive.test.ts` passed, but `project-watchers.test.ts` hit a `waitFor timeout`. I treated that as unrelated to this archive/read-path change and verified the target suite directly above.

The PR is now `DIRTY`, so it still needs a rebase before final review.

mrcfps:
@hobostay I'm holding off on generating review comments for nexu-io/open-design#485 because this pull request has merge conflicts right now.

apps/daemon/src/projects.ts
```diff
@@ -268,8 +276,15 @@ async function collectArchiveEntries(dir, relDir, out) {
...
+      if (ls.isSymbolicLink()) continue;
+    } catch (err) {
+      if (err && err.code === 'ENOENT') continue;
+      throw err;
+    }
+    const rel = relDir ? `${relDir}/${e.name}` : e.name;
     if (e.isDirectory()) {
       await collectArchiveEntries(full, rel, out);
```
mrcfps, line 289: The new `lstat(full)` guard still leaves a symlink-swap window in both `collectArchiveEntries()` here and `collectFiles()` above. After this check passes, we recurse with `collectArchiveEntries(full, rel, out)` / `collectFiles(full, rel, out)` and later reopen children by path again. If an attacker replaces that directory with a symlink between the `lstat()` and the recursive `readdir(full)`, Node will follow the swapped link and walk the external tree; `readArchiveEntry()` does not catch it because `lstat('project/link/secret.txt')` already resolves the intermediate `link` segment and reports a regular file. That keeps full-project ZIPs and file listings vulnerable to the same escape this PR is trying to close. Please pin each descended directory to the validated object instead of re-resolving `full` by path (for example via a descriptor-based walk / openat-style helper), or otherwise make the recursion reject when the object you recurse into is no longer the one you just validated.

apps/daemon/src/projects.ts
```diff
@@ -280,12 +295,34 @@ async function collectArchiveEntries(dir, relDir, out) {
...
+  } finally {
+    await handle.close();
+  }
+}
+
 export async function readProjectFile(projectsRoot, projectId, name) {
   const dir = projectDir(projectsRoot, projectId);
-  const file = resolveSafe(dir, name);
+  const rootReal = await realpath(dir).catch(() => dir);
+  const file = await resolveSafeReal(dir, name);
```
mrcfps, line 322: `resolveSafeReal()` only proves that `name` points inside the project *at check time*. `readProjectFile()` then immediately reopens the same path with `readFile(file)` / `stat(file)`, so a concurrent rename+symlink swap after `resolveSafeReal()` still lets this endpoint read an arbitrary file outside the project. That means the direct file-read surface remains exploitable even though the static symlink case is covered by the new test. Please switch this path to a descriptor-pinned read (for example `open(..., O_NOFOLLOW)` plus `handle.readFile()`/`handle.stat()` on supported platforms) or fail closed when you cannot guarantee a no-follow open, instead of validating with `realpath()` and then reopening by path.

mrcfps (changes requested):
@hobostay thanks for pushing the follow-up hardening here. The root-symlink case is fixed, but I found two remaining TOCTOU escapes in the changed paths: the recursive walkers still descend/read by path after the `lstat()` check, and `readProjectFile()` still validates with `realpath()` before reopening the file by path. Those races can still expose files outside the project root, so I’m keeping this blocked until the reads are pinned to the validated object.

PR #486, opened 2026-05-05
perf: cache allowed origins set instead of rebuilding per request

hobostay:
## Summary
- Caches the `Set` of allowed origins after first build instead of creating a new `Set` on every API request

## Problem
`buildAllowedOrigins()` was called inside the `/api` middleware on every single request, creating a new `Set` with ~12-18 entries each time. The allowed origins depend only on `resolvedPort` and `OD_WEB_PORT`, which don't change at runtime.

## Test plan
- [ ] API requests still enforce origin checks correctly
- [ ] No behavioral change — the cached set contains the same entries


lefarcen:
Hi @hobostay! 🎉

Thanks for the contribution — this is a smart optimization that eliminates repeated Set construction on every request.

— open-design team

apps/daemon/src/server.ts
```diff
@@ -659,6 +661,7 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
     );
+    return _cachedAllowedOrigins;
   }
```
lefarcen, line 1848: **P3** Stale comment reference — this line still mentions `buildAllowedOrigins` after the rename to `getAllowedOrigins`. Not a functional issue, but in security-sensitive origin-checking code it's worth keeping comments accurate to avoid future confusion.

Suggested fix: update the comment to reference `getAllowedOrigins()`.

lefarcen (commented):
Hey @hobostay, thanks for the optimization! This eliminates ~1200-1800 Set constructions per second on a busy daemon (12-18 origins × 100 req/s). I verified the caching is safe — `resolvedPort`, `host`, and `OD_WEB_PORT` are all immutable after `startServer()` completes.

Two minor suggestions below (non-blocking):

lefarcen:
**Additional suggestion — test coverage gap:**

The existing origin tests in `apps/daemon/tests/origin-validation.test.ts` copy the middleware logic instead of exercising the real `startServer()` and the new cached `getAllowedOrigins()` closure. That means they would still pass if production caching captured a stale or missing `OD_WEB_PORT`.

**Suggestion:** Add a focused integration test against real `startServer({ port: 0, returnServer: true })` with `process.env.OD_WEB_PORT` set before the first browser-origin request, asserting:
- The web port origin is allowed
- An unrelated port is rejected

This would catch any future cache lifetime issues. Not blocking, but worth considering for future robustness.

apps/daemon/src/server.ts
```diff
@@ -644,13 +644,15 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
   // Shared by the global origin middleware and isLocalSameOrigin() so
   // both use the same policy (loopback + explicit bind host, HTTP + HTTPS,
   // OD_WEB_PORT support).
-  function buildAllowedOrigins() {
+  let _cachedAllowedOrigins = null;
```
mrcfps: This cache needs an explicit `Set<string> | null` type before it can merge. With `apps/daemon/tsconfig.json` enabling `strict`, a `let` initialized only with `null` is not safely typed as the later `Set`, but the changed code assigns `new Set(...)` and then returns it for `.has(...)` checks in the middleware. That makes the daemon typecheck fail (or leaves the cache without the intended `Set<string>` contract), blocking this otherwise small optimization.

Please type the cache and return value explicitly, for example:

```ts
let _cachedAllowedOrigins: Set<string> | null = null;
function getAllowedOrigins(): Set<string> {
  if (_cachedAllowedOrigins) return _cachedAllowedOrigins;
  // ...
}
```

That keeps the cached origin set type aligned with the middleware usage and avoids a strict-mode compile break. 🙂

mrcfps (changes requested):
@hobostay I reviewed the daemon origin-cache change and found one merge-blocking TypeScript issue in the new cache variable. Once the cache is explicitly typed, the optimization should preserve the intended origin policy while avoiding the repeated Set allocation. Thanks for the focused performance cleanup 🙂


lefarcen:
Hi @hobostay! 👋

Just checking in — this PR has been waiting on the TypeScript type fix that @mrcfps mentioned. The change is straightforward:

```ts
let _cachedAllowedOrigins: Set<string> | null = null;
function getAllowedOrigins(): Set<string> {
  if (_cachedAllowedOrigins) return _cachedAllowedOrigins;
  // ... rest stays the same
}
```

Once you push that type annotation, we can move forward with merging this nice performance optimization. Let us know if you have any questions!

— open-design team

lefarcen:
Hi @hobostay — I rebased this branch onto current `main` and pushed the result, but heads up: while resolving the conflict I noticed the optimization this PR was targeting has been overtaken by an unrelated refactor on `main`, so the merge ends up being a no-op. Wanted to flag it before you spend any more time here.

Specifically: the `buildAllowedOrigins()` function this PR was caching no longer exists on `main`. The CORS origin logic was moved out of `apps/daemon/src/server.ts` into a dedicated module, `apps/daemon/src/origin-validation.ts`. The new shape:

1. `extraAllowedOrigins = configuredAllowedOrigins()` is computed **once** at `startServer` setup (`server.ts:1838`) — that's already the cache your PR was after for the env-driven list.
2. The per-request hot path is now `allowedBrowserPorts(resolvedPort)` + `isAllowedBrowserOrigin(...)`. `allowedBrowserPorts` is a 5-line function that returns at most 2 numbers; allocating a 2-element array per request is cheaper than the indirection a cache would add, so caching it doesn't pencil out.

Net effect: `git diff origin/main..fix/cache-allowed-origins` is now empty for `apps/daemon/src/server.ts`. The PR can technically merge — the CI should go green — but it won't produce any code change beyond a merge commit. Given that, I'm going to close this rather than land a no-op.

If there's still a hot-path origin check you want to optimize, the remaining candidate is inside `isAllowedBrowserOrigin` itself in `origin-validation.ts` — that one walks the allowlist on every request. Happy to look at a follow-up there if you spot something.

Thanks for the patch, sorry the timing didn't work out.

PR #488, opened 2026-05-05
fix: cap htmlPreviewSlideState Map to prevent memory leak

hobostay:
## Summary
- Adds a 64-entry cap with FIFO eviction to the `htmlPreviewSlideState` module-level Map

## Problem
`htmlPreviewSlideState` in `apps/web/src/components/FileViewer.tsx` (line 45) is a module-level `Map<string, SlideState>` keyed by `${projectId}:${fileName}`. Entries are set but never deleted. In long-lived sessions where the user opens many projects and files, this Map grows without bound, leaking memory.

## Fix
Wrap the `.set()` call in a `setSlideStateCached()` function that evicts the oldest entry when the Map exceeds 64 entries. The `.get()` calls remain unchanged — they just return `undefined` for evicted keys, which is handled gracefully (falls back to slide 0).

## Test plan
- [ ] Open many files with deck previews — slide state should still be preserved for recent files
- [ ] Verify old entries are evicted after opening >64 unique files

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -42,7 +42,16 @@ import type { PreviewComment, PreviewCommentTarget } from '../types';
 type TranslateFn = (key: keyof Dict, vars?: Record<string, string | number>) => string;
 type SlideState = { active: number; count: number };
 
+const MAX_CACHED_SLIDE_STATES = 64;
 const htmlPreviewSlideState = new Map<string, SlideState>();
+
+function setSlideStateCached(key: string, state: SlideState) {
+  htmlPreviewSlideState.set(key, state);
+  if (htmlPreviewSlideState.size > MAX_CACHED_SLIDE_STATES) {
+    const oldest = htmlPreviewSlideState.keys().next().value;
+    if (oldest != null) htmlPreviewSlideState.delete(oldest);
+  }
+}
 
 interface Props {
   projectId: string;
@@ -739,7 +748,7 @@ function HtmlViewer({
       if (!data || data.type !== 'od:slide-state') return;
       if (typeof data.active !== 'number' || typeof data.count !== 'number') return;
       const next = { active: data.active, count: data.count };
-      htmlPreviewSlideState.set(previewStateKey, next);
+      setSlideStateCached(previewStateKey, next);
       setSlideState(next);
     }
     window.addEventListener('message', onMessage);


lefarcen:
Hi @hobostay! 🎉
Thanks for the contribution — solid catch on the memory leak.

— open-design team

mrcfps (approved):
@hobostay I reviewed the FileViewer slide-state cache update and verified the new bounded helper caps the module-level deck state cache while preserving the existing fallback behavior for evicted entries. This is a focused, safe fix for the long-lived-session memory growth path — thanks for keeping the change small and easy to reason about. 😊


lefarcen (commented):
Hi @hobostay! Solid fix for the memory leak — the bounded cache approach is the right direction.

Two refinements worth considering (non-blocking):

**1. FIFO vs LRU semantics** — Currently the cache evicts by first insertion order, not by recent use. When you re-visit a deck (send a slide-state message for an existing key), the `Map.set()` updates the value but doesn't move that key to the end of insertion order. This means a deck you just interacted with could still be among the oldest 64 and get evicted next.

If the goal is preserving *recently used* decks (as the test plan suggests), the cache should be LRU. Small fix: delete before setting to refresh the key's position:

```ts
function setSlideStateCached(key: string, state: SlideState) {
  if (htmlPreviewSlideState.has(key)) htmlPreviewSlideState.delete(key);
  htmlPreviewSlideState.set(key, state);
  if (htmlPreviewSlideState.size > MAX_CACHED_SLIDE_STATES) {
    const oldest = htmlPreviewSlideState.keys().next().value;
    if (oldest != null) htmlPreviewSlideState.delete(oldest);
  }
}
```

**2. Test coverage** — Existing `FileViewer.test.tsx` covers viewer routing/srcdoc rendering but not the `od:slide-state` message caching or eviction behavior. Without tests, off-by-one errors and FIFO/LRU edge cases can regress silently. Consider adding a focused test for:
- Size stays at 64 after 65th unique key
- 65th entry evicts the expected key (oldest or least-recently-used)
- size === MAX does not trigger eviction

**What's good:**
- ✅ The guard `if (oldest != null)` handles the (unreachable) undefined case safely
- ✅ Performance is O(1)-ish and not on a hot path
- ✅ 64 entries is reasonable for tiny `{active, count}` objects
- ✅ Falling back to slide 0 for evicted keys is acceptable for an ephemeral cache

**Verdict:** Ship as-is if you're comfortable with FIFO semantics and want to add tests later. Otherwise the LRU tweak is a quick win — up to you!

lefarcen (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @hobostay for the contribution to open-design. Looking forward to the next one!

lefarcen:
Thanks for the fix, @hobostay! 🎉 The bounded cache is now merged and will prevent the memory leak in long-lived sessions.

If you're using open-design and notice any regressions with slide state preservation, feel free to open a follow-up issue.

PR #492, opened 2026-05-05
fix(daemon): handle ACP `-32603` errors gracefully in `session/set_model`

### LINKED ISSUE #443: ACP `session/set_model` returns `-32603 Internal error` then `ACP response timed out after 180000ms`
## Environment

- Open Design v0.3.0 (tag `open-design-v0.3.0`, running via `tools-dev run web`)
- Agent: Hermes Agent v0.12.0 (ACP JSON-RPC, `streamFormat: acp-json-rpc`)
- Node.js v24.15.0, pnpm via corepack
- OS: Ubuntu 24.04

## Summary

When a non-default model is selected (e.g., `ollama-cloud:kimi-k2.6`), Open Design sends `session/set_model` to the agent. Hermes ACP returns `{"jsonrpc":"2.0","id":3,"error":{"code":-32603,"message":"Internal error"}}`. Open Design's `acp.ts` parser treats this as a fatal error via `fail()`, which kills the child process. Since the prompt is never sent, the 180-second stage timer fires and the UI shows "ACP response timed out after 180000ms".

When the default model is selected, `session/set_model` is skipped and everything works correctly in ~30 seconds.

## Root Cause

In `apps/daemon/src/acp.ts`, the `attachAcpSession` parser routes ALL JSON-RPC errors through `fail()`. For `-32603 Internal error` specifically:

1. Hermes successfully completes `initialize` (id=1) and `session/new` (id=2)
2. Open Design sends `session/set_model` (id=3) with the user-selected model
3. Hermes returns `{"id":3,"error":{"code":-32603,"message":"Internal error"}}` for `set_model`
4. `rpcErrorMessage()` formats this as `"json-rpc id 3: Internal error"`
5. `fail()` is called → `child.kill('SIGTERM')` → prompt is never sent → 180s timeout

The `Internal error` from Hermes appears to be related to model switching within an ACP session — the `set_session_model` handler in Hermes may not support switching to a cloud-provider model after session creation, or there may be a provider initialization failure.

## Fix (Workaround on Open Design Side)

Since modifying Hermes is not an option, the fix handles `-32603` gracefully in Open Design so the conversation can proceed with the default model:

### Three patches in `apps/daemon/src/acp.ts`:

**Patch 1 — `detectAcpModels` parser (~L167):** Suppress `-32603` during model detection (harmless noise):
```typescript
if (rpcErr) {
  if (raw.error?.code === -32603) return;  // non-fatal race condition
  fail(rpcErr);
  return;
}
```

**Patch 2 — `attachAcpSession` parser (~L314):** Distinguish between `-32603` on the expected response ID (recoverable) vs. unexpected IDs (suppress as noise):
```typescript
if (rpcErr) {
  if (finished) return;
  if (raw.error?.code === -32603 && raw.id !== expectedId) {
    // Late-arriving cleanup noise — suppress
    return;
  }
  if (raw.error?.code === -32603 && raw.id === expectedId) {
    // Expected response failed — fall through to id-matching logic for recovery
  } else {
    fail(rpcErr);
    return;
  }
}
```

**Patch 3 — Recovery block (~L374):** When `session/set_model` fails, skip model switching and proceed with the default model:
```typescript
if (
  raw.error?.code === -32603 &&
  raw.id === expectedId &&
  sessionId &&
  model &&
  model !== 'default'
) {
  console.error('[ACP DEBUG] session/set_model failed with -32603, falling back to default model');
  activeModel = activeModel || 'default';
  send('agent', { type: 'status', label: 'model', model: activeModel });
  sendPrompt();
  return;
}
```

## Verified Debug Log

```
[ACP DEBUG] writeRpc: id=1 method=initialize timeoutLabel=initialize
[ACP DEBUG] [ACP RECV RESULT] id=1 keys=agentCapabilities,agentInfo,authMethods,protocolVersion
[ACP DEBUG] writeRpc: id=2 method=session/new timeoutLabel=session/new
[ACP DEBUG] [ACP RECV RESULT] id=2 keys=models,sessionId
[ACP DEBUG] writeRpc: id=3 method=session/set_model timeoutLabel=session/set_model
[ACP DEBUG] [ACP RECV] method=session/update id=null
[ACP DEBUG] [ACP RECV ERROR] id=3 code=-32603 msg=Internal error
[ACP DEBUG] -32603 on expected id=3 — falling through to recovery logic
[ACP DEBUG] session/set_model failed with -32603, falling back to default model and sending prompt
[ACP DEBUG] sendPrompt: id=4 sessionId=... model=ollama-cloud:kimi-k2.6
[ACP DEBUG] writeRpc: id=4 method=session/prompt timeoutLabel=session/prompt
[ACP DEBUG] [ACP RECV] method=session/update id=null  (streaming starts)
[ACP DEBUG] [ACP RECV RESULT] id=4 keys=stopReason,usage  (completed in ~30s)
```

## Suggested Upstream Fix

The ideal fix would be in Hermes ACP's `set_session_model` handler to not throw `-32603` when switching to a valid cloud-provider model, or at minimum to return a proper error response that Open Design can handle gracefully. However, since not all agents support `session/set_model`, Open Design should also gracefully degrade when the agent rejects it.

## Additional Context

- The default model (whatever the agent uses at session creation) works fine — the issue only manifests when a non-default model is explicitly selected via `session/set_model`
- With the default model, the full flow (initialize → session/new → sendPrompt) completes without error
- The same `-32603` error also appears after prompt completion as a late-arriving cleanup message (race condition on stdin pipe close), which should also be suppressed — this is the cosmetic `"json-rpc id 3: Internal error"` users see in the UI

goemonwanwan:
## Summary

When a non-default model is selected (e.g. `ollama-cloud:kimi-k2.6`), Hermes ACP returns `{"jsonrpc":"2.0","id":3,"error":{"code":-32603,"message":"Internal error"}}` for the `session/set_model` request. Open Design's `acp.ts` parser routes **all** JSON-RPC errors through `fail()`, which kills the child process. Since the prompt is never sent, the 180-second stage timer fires and the UI shows "ACP response timed out after 180000ms".

Closes #443

## Changes

Three patches in `apps/daemon/src/acp.ts`:

### Patch 1 — `detectAcpModels` parser (~L167)
Suppress `-32603` during model detection. These are harmless noise from agent cleanup.

```typescript
if (raw.error?.code === -32603) return;
```

### Patch 2 — `attachAcpSession` parser (~L299)
Distinguish between `-32603` on the expected response ID vs unexpected IDs:
- **Unexpected ID**: suppress as cleanup race-condition noise
- **Expected ID** (`session/set_model` response): fall through to the id-matching logic for recovery

Also adds a `finished` guard to suppress late-arriving errors after response completion.

### Patch 3 — Recovery block (~L351)
When `session/set_model` fails with `-32603`, skip model switching and proceed with the default model by calling `sendPrompt()`.

## Tested

Verified with Hermes Agent v0.12.0 (ACP JSON-RPC, `streamFormat: acp-json-rpc`):

```
[ACP DEBUG] writeRpc: id=1 method=initialize timeoutLabel=initialize
[ACP DEBUG] [ACP RECV RESULT] id=1 keys=agentCapabilities,agentInfo,authMethods,protocolVersion
[ACP DEBUG] writeRpc: id=2 method=session/new timeoutLabel=session/new
[ACP DEBUG] [ACP RECV RESULT] id=2 keys=models,sessionId
[ACP DEBUG] writeRpc: id=3 method=session/set_model timeoutLabel=session/set_model
[ACP DEBUG] [ACP RECV] method=session/update id=null
[ACP DEBUG] [ACP RECV ERROR] id=3 code=-32603 msg=Internal error
[ACP DEBUG] -32603 on expected id=3 — falling through to recovery logic
[ACP DEBUG] session/set_model failed with -32603, falling back to default model and sending prompt
[ACP DEBUG] sendPrompt: id=4 sessionId=... model=ollama-cloud:kimi-k2.6
[ACP DEBUG] writeRpc: id=4 method=session/prompt timeoutLabel=session/prompt
[ACP DEBUG] [ACP RECV] method=session/update id=null (streaming starts)
[ACP DEBUG] [ACP RECV RESULT] id=4 keys=stopReason,usage (completed in ~30s)
```

## Design rationale

Per `specs/current/runtime-adapter.md`, `session/set_model` is **optional**. Agents that don't support it (or reject specific models) should not break the conversation flow. This PR degrades gracefully by falling back to the default model.

### DIFF
--- a/apps/daemon/src/acp.ts
+++ b/apps/daemon/src/acp.ts
@@ -165,6 +165,11 @@ export async function detectAcpModels({
     const parser = createJsonLineStream((raw) => {
       const rpcErr = rpcErrorMessage(raw);
       if (rpcErr) {
+        // JSON-RPC -32603 "Internal error" during model detection:
+        // If this is for the current expected-id (initialize/session/new),
+        // it's a real probe failure — reject immediately.
+        // Otherwise it's cleanup noise — suppress it.
+        if (raw.error?.code === -32603 && raw.id !== expectedId) return;
         fail(rpcErr);
         return;
       }
@@ -223,6 +228,7 @@ export function attachAcpSession({
   let expectedId = 1;
   let nextId = 2;
   let promptRequestId = null;
+  let setModelRequestId = null;
   let sessionId = null;
   let activeModel = null;
   let emittedThinkingStart = false;
@@ -296,8 +302,28 @@ export function attachAcpSession({
     resetStageTimer('response');
     const rpcErr = rpcErrorMessage(raw);
     if (rpcErr) {
-      fail(rpcErr);
-      return;
+      // After response completion, any late-arriving errors from the agent
+      // (pipe-broken, cleanup race conditions, etc.) are safe to ignore.
+      if (finished) return;
+      // JSON-RPC -32603 "Internal error" handling:
+      // Unexpected-id -32603 errors are cleanup noise — suppress them.
+      // Expected-id -32603 errors for session/set_model fall through to the
+      // recovery block below. All other expected-id -32603 errors (initialize,
+      // session/new, session/prompt) are real failures — call fail().
+      if (raw.error?.code === -32603 && raw.id !== expectedId) {
+        return;
+      }
+      if (raw.error?.code === -32603 && raw.id === expectedId) {
+        if (raw.id === setModelRequestId) {
+          // Fall through — the recovery block will handle this
+        } else {
+          fail(rpcErr);
+          return;
+        }
+      } else {
+        fail(rpcErr);
+        return;
+      }
     }
     if (raw.method === 'session/request_permission') {
       replyPermission(raw);
@@ -333,6 +359,21 @@ export function attachAcpSession({
       }
       return;
     }
+    // Recovery: if session/set_model failed with -32603, fall back to
+    // sending the prompt with the default (already-active) model.
+    // This is scoped to the exact set_model request id to avoid
+    // triggering on prompt or other request failures.
+    if (
+      raw.error?.code === -32603 &&
+      raw.id === setModelRequestId &&
+      promptRequestId === null
+    ) {
+      setModelRequestId = null;
+      activeModel = activeModel || 'default';
+      send('agent', { type: 'status', label: 'model', model: activeModel });
+      sendPrompt();
+      return;
+    }
     if (raw.id !== expectedId || !raw.result || typeof raw.result !== 'object') {
       return;
     }
@@ -360,6 +401,7 @@ export function attachAcpSession({
         send('agent', { type: 'status', label: 'model', model: activeModel });
       }
       if (sessionId && model && model !== 'default') {
+        setModelRequestId = nextId;
         expectedId = nextId;
         writeRpc(
           nextId,


lefarcen:
Hi @goemonwanwan! 🎉

Thanks for the contribution — excellent diagnostic work on the `-32603` error handling and graceful degradation approach.

— open-design team

apps/daemon/src/acp.ts
```diff
@@ -333,6 +351,20 @@ export function attachAcpSession({
       }
       return;
     }
+    // Recovery: if session/set_model failed with -32603, fall back to
```
lefarcen, line 362: **P1 Duplicate-prompt risk**

This recovery branch is not tied to `session/set_model`. If `session/prompt` returns `-32603` after a non-default model was requested, all three conditions (`sessionId`, `model`, `model !== 'default'`) are still true → the daemon sends the *same prompt again* with the default model.

That can:
- Duplicate user actions (e.g., two `git commit` runs, two file writes, two API calls)
- Loop on repeated prompt failures

**Fix:** Track the exact `session/set_model` request id and only fallback when `raw.id === setModelRequestId`:

```typescript
let setModelRequestId: number | undefined;

// Before calling sendModelRequest():
setModelRequestId = nextId;

// In the recovery block:
if (
  raw.error?.code === -32603 &&
  raw.id === setModelRequestId &&
  sessionId &&
  model &&
  model !== 'default'
) {
  setModelRequestId = undefined; // one-shot
  activeModel = activeModel || 'default';
  send('agent', { type: 'status', label: 'model', model: activeModel });
  sendPrompt();
  return;
}
```

apps/daemon/src/acp.ts
```diff
@@ -296,8 +299,23 @@ export function attachAcpSession({
...
+      if (finished) return;
+      // JSON-RPC -32603 "Internal error" handling:
+      // If this error is for the expected response ID, fall through so the
+      // id-matching logic below can attempt recovery (e.g. session/set_model
+      // fallback). Otherwise, suppress it — these are typically race-condition
+      // errors during cleanup that don't affect the conversation.
+      if (raw.error?.code === -32603 && raw.id !== expectedId) {
+        return;
+      }
+      if (raw.error?.code === -32603 && raw.id === expectedId) {
```
lefarcen, line 316: **P2 Silent timeout for other expected-id errors**

Expected `-32603` errors for `initialize`, `session/new`, or `session/prompt` (when model was already default) fall through this block, fail the result-shape check below, and wait for timeout instead of surfacing the actual JSON-RPC error.

**Fix:** Fail all expected-id errors unless the id matches the tracked `session/set_model` request id:

```typescript
if (raw.error?.code === -32603 && raw.id !== expectedId) {
  return; // cleanup noise
}
if (raw.error?.code === -32603 && raw.id === expectedId && raw.id !== setModelRequestId) {
  fail(rpcErr); // real failure for initialize/new/prompt
  return;
}
if (raw.error?.code === -32603 && raw.id === setModelRequestId) {
  // fall through for recovery
} else if (rpcErr) {
  fail(rpcErr);
  return;
}
```

apps/daemon/src/acp.ts
```diff
@@ -165,6 +165,9 @@ export async function detectAcpModels({
     const parser = createJsonLineStream((raw) => {
       const rpcErr = rpcErrorMessage(raw);
       if (rpcErr) {
+        // JSON-RPC -32603 "Internal error" is non-fatal during model detection.
+        // Some agents (e.g. Hermes) send this during cleanup; silently ignore.
+        if (raw.error?.code === -32603) return;
```
lefarcen: **P2 Model detection ignores all `-32603`, including real probe failures**

`detectAcpModels()` suppresses *every* `-32603`, including errors for the active `initialize` or `session/new` request. That converts real probe failures into a 15s timeout and can slow `/api/agents`.

**Fix:** Ignore only post-settlement or cleanup errors. Expected-id `-32603` should reject immediately:

```typescript
let probeExpectedId: number | undefined;

// When sending initialize or session/new:
probeExpectedId = id;

// In the error block:
if (raw.error?.code === -32603) {
  if (raw.id !== probeExpectedId) return; // cleanup noise
  fail(rpcErr); // real probe failure
  return;
}
```

lefarcen (changes requested):
Hi @goemonwanwan! 👋

Excellent diagnostic work on the root cause — the state-machine approach is sound. However, the recovery logic has a critical correctness bug that could duplicate user actions, plus two edge cases that convert real failures into timeouts.

**Summary:** The PR solves the immediate `-32603` crash but the recovery branch is too broad — it triggers for *any* `-32603` after a non-default model was requested, not just `session/set_model` failures. This can duplicate prompts or loop.

See inline comments for specifics. All fixable with surgical scoping.

apps/daemon/src/acp.ts
```diff
@@ -333,6 +351,20 @@ export function attachAcpSession({
...
     }
+    // Recovery: if session/set_model failed with -32603, fall back to
+    // sending the prompt with the default (already-active) model.
+    if (
+      raw.error?.code === -32603 &&
+      raw.id === expectedId &&
+      sessionId &&
+      model &&
+      model !== 'default'
+    ) {
```
mrcfps, line 370: This fallback needs to be tied to the specific `session/set_model` request id. Right now the condition only checks for `-32603`, `raw.id === expectedId`, `sessionId`, and a non-default `model`. Those same facts are still true after `sendPrompt()` changes `expectedId` to the prompt request id, so a real `session/prompt` `-32603` would be treated as a set-model failure and `sendPrompt()` would run again. That can duplicate the user’s prompt, spend tokens twice, or keep retrying on a persistent prompt failure.

Please track the pending `session/set_model` id/method explicitly (for example, `setModelRequestId`) and only take this recovery path when `raw.id === setModelRequestId` and no prompt has been sent yet (`promptRequestId === null`). Other expected-id `-32603` responses, including initialize/session-new/prompt failures, should continue through `fail(rpcErr)` so they surface as real agent errors instead of being swallowed. 😊

mrcfps (changes requested):
Thanks for the focused ACP recovery patch. I found one state-machine issue that can turn real prompt failures into repeated prompt sends, so I’m requesting a small adjustment before merge.


lefarcen (commented):
Hi @goemonwanwan! ✅

Excellent work addressing all the review feedback:

**P1 Fixed:** The recovery block is now correctly scoped to `setModelRequestId` and checks `promptRequestId === null` — no more duplicate-prompt risk.

**P2 Fixed:** Expected-id `-32603` errors for `initialize`/`session/new`/`session/prompt` now fail immediately instead of being swallowed.

**P2 Fixed:** `detectAcpModels` now fails expected-id `-32603` errors (real probe failures) while suppressing cleanup noise.

The state-machine logic is clean and surgical. This PR is ready from my side — deferring final approval to a maintainer. 🎉

mrcfps (approved):
@goemonwanwan I reviewed the updated ACP recovery diff in `apps/daemon/src/acp.ts`, including the `setModelRequestId` scoping, `promptRequestId === null` guard, and expected-vs-cleanup `-32603` handling in both session attachment and model detection. The earlier duplicate-prompt and swallowed-error paths are addressed cleanly, and I didn’t find any remaining actionable issues in the changed ranges. Nice, focused follow-up on a tricky state-machine edge case. 😊


lefarcen (approved):
Approving to unblock — all P1/P2 issues confirmed addressed in d9c62d4 (see prior comment). LGTM.

PR #513, opened 2026-05-05
Add Tweaks mode for HTML previews with picker, pod selection, and batched chat attachments

Martin-Atrin:
## Summary
- add a `Tweaks` workflow for rendered HTML previews
- support precise DOM-backed picking and freehand pod selection inside the preview iframe
- let users queue multiple notes and send them to chat as one structured attachment batch

## What changed
- extend preview comment contracts and persistence so pod selections can carry grouped member context
- upgrade the HTML preview bridge to expose nested DOM targets plus pod-stroke events
- replace the old single comment toggle with `Tweaks`, `Picker`, and `Pods` in the HTML viewer
- add helper/runtime/daemon tests for batched board attachments and the richer iframe bridge

## Validation
- `npm exec pnpm -- --filter @open-design/web typecheck`
- `npm exec pnpm -- --filter @open-design/web exec vitest run tests/comments.test.ts tests/runtime/srcdoc.test.ts`
- `npm exec pnpm -- --filter @open-design/daemon exec vitest run tests/comment-attachments.test.ts`
- `npm exec pnpm -- --filter @open-design/web build`

### DIFF
--- a/apps/daemon/src/db.ts
+++ b/apps/daemon/src/db.ts
@@ -9,7 +9,6 @@ import Database from 'better-sqlite3';
 import path from 'node:path';
 import fs from 'node:fs';
 import { randomUUID } from 'node:crypto';
-import { migrateCritique } from './critique/persistence.js';
 
 let dbInstance = null;
 let dbFile = null;
@@ -171,6 +170,17 @@ function migrate(db) {
   if (!messageCols.some((c) => c.name === 'comment_attachments_json')) {
     db.exec(`ALTER TABLE messages ADD COLUMN comment_attachments_json TEXT`);
   }
+
+  const previewCommentCols = db.prepare(`PRAGMA table_info(preview_comments)`).all();
+  if (!previewCommentCols.some((c) => c.name === 'selection_kind')) {
+    db.exec(`ALTER TABLE preview_comments ADD COLUMN selection_kind TEXT`);
+  }
+  if (!previewCommentCols.some((c) => c.name === 'member_count')) {
+    db.exec(`ALTER TABLE preview_comments ADD COLUMN member_count INTEGER`);
+  }
+  if (!previewCommentCols.some((c) => c.name === 'pod_members_json')) {
+    db.exec(`ALTER TABLE preview_comments ADD COLUMN pod_members_json TEXT`);
+  }
   const deploymentCols = db.prepare(`PRAGMA table_info(deployments)`).all();
   if (!deploymentCols.some((c) => c.name === 'status')) {
     db.exec(`ALTER TABLE deployments ADD COLUMN status TEXT NOT NULL DEFAULT 'ready'`);
@@ -181,7 +191,6 @@ function migrate(db) {
   if (!deploymentCols.some((c) => c.name === 'reachable_at')) {
     db.exec(`ALTER TABLE deployments ADD COLUMN reachable_at INTEGER`);
   }
-  migrateCritique(db);
 }
 
 // ---------- deployments ----------
@@ -738,6 +747,8 @@ export function listPreviewComments(db, projectId, conversationId) {
       `SELECT id, project_id AS projectId, conversation_id AS conversationId,
               file_path AS filePath, element_id AS elementId, selector, label,
               text, position_json AS positionJson, html_hint AS htmlHint,
+              selection_kind AS selectionKind, member_count AS memberCount,
+              pod_members_json AS podMembersJson,
               note, status, created_at AS createdAt, updated_at AS updatedAt
          FROM preview_comments
         WHERE project_id = ? AND conversation_id = ?
@@ -758,6 +769,11 @@ export function upsertPreviewComment(db, projectId, conversationId, input) {
   const text = typeof target.text === 'string' ? compactWhitespace(target.text).slice(0, 160) : '';
   const htmlHint = typeof target.htmlHint === 'string' ? compactWhitespace(target.htmlHint).slice(0, 180) : '';
   const position = normalizePosition(target.position);
+  const selectionKind = target.selectionKind === 'pod' ? 'pod' : 'element';
+  const podMembers = selectionKind === 'pod' ? normalizePodMembers(target.podMembers) : [];
+  const memberCount = selectionKind === 'pod'
+    ? Math.max(podMembers.length, Number.isFinite(target.memberCount) ? Math.round(target.memberCount) : 0)
+    : 0;
   const now = Date.now();
   const existing = db
     .prepare(
@@ -771,14 +787,18 @@ export function upsertPreviewComment(db, projectId, conversationId, input) {
   db.prepare(
     `INSERT INTO preview_comments
        (id, project_id, conversation_id, file_path, element_id, selector, label,
-        text, position_json, html_hint, note, status, created_at, updated_at)
-     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+        text, position_json, html_hint, selection_kind, member_count, pod_members_json,
+        note, status, created_at, updated_at)
+     VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
      ON CONFLICT(project_id, conversation_id, file_path, element_id) DO UPDATE SET
        selector = excluded.selector,
        label = excluded.label,
        text = excluded.text,
        position_json = excluded.position_json,
        html_hint = excluded.html_hint,
+       selection_kind = excluded.selection_kind,
+       member_count = excluded.member_count,
+       pod_members_json = excluded.pod_members_json,
        note = excluded.note,
        status = 'open',
        updated_at = excluded.updated_at`,
@@ -793,6 +813,9 @@ export function upsertPreviewComment(db, projectId, conversationId, input) {
     text,
     JSON.stringify(position),
     htmlHint,
+    selectionKind,
+    selectionKind === 'pod' ? memberCount : null,
+    selectionKind === 'pod' ? JSON.stringify(podMembers) : null,
     note,
     'open',
     createdAt,
@@ -828,6 +851,8 @@ function getPreviewComment(db, projectId, conversationId, id) {
       `SELECT id, project_id AS projectId, conversation_id AS conversationId,
               file_path AS filePath, element_id AS elementId, selector, label,
               text, position_json AS positionJson, html_hint AS htmlHint,
+              selection_kind AS selectionKind, member_count AS memberCount,
+              pod_members_json AS podMembersJson,
               note, status, created_at AS createdAt, updated_at AS updatedAt
          FROM preview_comments
         WHERE id = ? AND project_id = ? AND conversation_id = ?`,
@@ -837,6 +862,7 @@ function getPreviewComment(db, projectId, conversationId, id) {
 }
 
 function normalizePreviewComment(row) {
+  const podMembers = parseJsonOrUndef(row.podMembersJson);
   return {
     id: row.id,
     projectId: row.projectId,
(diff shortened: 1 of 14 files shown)


apps/daemon/src/db.ts
```diff
@@ -181,7 +191,6 @@ function migrate(db) {
   if (!deploymentCols.some((c) => c.name === 'reachable_at')) {
     db.exec(`ALTER TABLE deployments ADD COLUMN reachable_at INTEGER`);
   }
-  migrateCritique(db);
```
mrcfps, line 184: Blocking: please keep the critique migration wired into `migrate(db)` and restore the corresponding `migrateCritique` import. This deleted call is the only place that creates the `critique_runs` table, but daemon startup still calls `reconcileStaleRuns(db, ...)` immediately after `openDatabase()` in `apps/daemon/src/server.ts`, and critique runs still call `insertCritiqueRun`/`updateCritiqueRun`. On a fresh database this means the daemon can hit `no such table: critique_runs` during boot or the first critique flow, which breaks the main runtime path. The safe fix is to add the preview-comment column migration without removing `migrateCritique(db)`, and keep/extend a daemon DB migration test that opens a fresh database and verifies both the new preview comment columns and the critique table exist.

mrcfps (changes requested):
Found one blocking issue in the daemon migration path: the PR removes the critique table migration while server startup still relies on that table being present.


lefarcen (changes requested):
Hi @Martin-Atrin! Thanks for the contribution — this is an ambitious feature that brings precise DOM picking and pod-based batch commenting to HTML previews. 🙌

I've found several correctness gaps that need addressing before merge, particularly around coordinate space handling at different zoom levels and selector generation for modern CSS class names. The core picker/pod workflow has solid structure but needs fixes in the geometry and data validation paths.

**P1 findings (must fix):**

1. **apps/web/src/components/FileViewer.tsx:1613** — Pod hit-testing mixes coordinate spaces when preview zoom is not 100%. The stroke points come from the iframe viewport, but `selectionHitsSnapshot` compares them against `overlayBoundsFromSnapshot(..., scale)`, so at 50%/200% zoom the lasso selects the wrong elements or misses them. **Fix**: Keep geometry in iframe coordinates for selection, and only apply `previewScale` when rendering host overlays/strokes.

2. **apps/web/src/components/FileViewer.tsx:1361** — The pod stroke overlay is rendered with raw iframe coordinates while target overlays are scaled by `previewScale`. At any zoom other than 100%, the drawn stroke visibly drifts from the pointer/targets. **Fix**: Scale the SVG polyline points for display, or store a separate host-coordinate stroke for rendering.

**P2 findings (should fix):**

3. **apps/web/src/runtime/srcdoc.ts:143** — `cssPath` appends raw class names into selectors. Common generated/Tailwind classes like `hover:bg-blue-500`, `w-1/2`, or `md:grid` become invalid or semantically different selectors, so the picker sends unusable DOM selectors to chat. **Fix**: Escape each class token with `CSS.escape` before joining them into `.class` segments.

4. **apps/web/src/components/FileViewer.tsx:2187** — The parent accepts iframe-supplied positions with `Number(value) || 0`, which lets `Infinity` through. A generated preview can post `od:comment-target(s)` or pod points containing `Infinity`, producing invalid React styles and NaN/Infinity geometry. **Fix**: Use a finite-number helper for all message payload coordinates and clamp very large values.

5. **apps/web/src/runtime/srcdoc.ts:299** — The comment bridge observes all subtree mutations and reposts a full `querySelectorAll('*')` target snapshot on the next animation frame. Animated/generated pages that update classes, text, or attributes every frame will force full-DOM layout/style reads and large postMessage payloads while Tweaks is enabled. **Fix**: Debounce more aggressively, cap target counts, or only rescan on explicit host request/settled idle.

6. **apps/daemon/src/db.ts:774** — `memberCount` can drift from `podMembers`: the server persists `Math.max(podMembers.length, target.memberCount)`, so a client can store `memberCount: 100` with two members and the agent prompt will report a different group size than the actual member list. **Fix**: Derive `memberCount` from normalized `podMembers.length` when members are present, or define one field as authoritative and validate the other.

**P3 findings (nice to fix):**

7. **apps/web/src/components/FileViewer.tsx:1207** — The board composer popover is rendered as a plain div with no dialog/popover role, focus handoff, Escape handling, or labelled textarea, and Picker/Pods only expose visual active state. Keyboard and screen-reader users can toggle Tweaks but cannot reliably discover or operate the resulting composer. **Fix**: Give the popover dialog semantics/focus management and add `aria-pressed`/labels to mode buttons.

Let me know if you have questions on any of these. Happy to clarify!

Martin-Atrin:
Yeah @lefarcen will fix that. Thanks for review.

lefarcen:
Quick clarification: I'm the automated review bot, not a human maintainer! 😄

The 7 findings in my review (2 P1 coordinate issues, 4 P2 validation/performance, 1 P3 accessibility) need to be addressed by you (the PR author) in new commits. Once you push fixes, I'll automatically re-review.

mrcfps's blocking critique migration issue also needs your fix.

Happy to clarify any of the findings if helpful!

apps/daemon/src/db.ts
Martin-Atrin: Fixed in `66fc6d5`.

Changes in this follow-up:
- restored the `migrateCritique` import and `migrateCritique(db)` call inside `migrate(db)` so fresh databases still create `critique_runs`
- added a fresh-database migration assertion in `apps/daemon/tests/comment-attachments.test.ts` that verifies both the new preview comment pod columns and the `critique_runs` table exist
- also addressed the preview-review follow-ups on this branch: iframe-space pod hit testing, scaled stroke rendering at non-100% zoom, escaped class tokens in generated selectors, finite/clamped bridge coordinates, throttled/capped target rescans, and authoritative pod member counts when members are present

Revalidated:
- `npm exec pnpm -- --filter @open-design/daemon exec vitest run tests/comment-attachments.test.ts`
- `npm exec pnpm -- --filter @open-design/web exec vitest run tests/comments.test.ts tests/runtime/srcdoc.test.ts tests/components/FileViewer.test.tsx`
- `npm exec pnpm -- --filter @open-design/web typecheck`
- `npm exec pnpm -- --filter @open-design/web build`

lefarcen (approved):
Looks good! 🎉

Verified the fixes at 66fc6d5:
- ✅ mrcfps's blocking finding: `migrateCritique` import and call restored
- ✅ All 7 prior bot findings addressed in this iteration

The coordinate space handling, stroke rendering, selector escaping, validation, throttling, and member count logic all look clean now. Nice work on the comprehensive fix iteration!

Deferring final merge decision to a maintainer.

apps/web/src/components/FileViewer.tsx
```diff
@@ -1160,39 +1173,90 @@ function FileActions({
...
+  const pendingCount = notes.length + (draft.trim() ? 1 : 0);
+  const podMembers = target.podMembers ?? [];
   return (
-    <div className="comment-popover" data-testid="comment-popover">
+    <div
+      className="comment-popover"
+      data-testid="comment-popover"
+      role="dialog"
+      aria-modal="false"
+      aria-labelledby={`board-composer-title-${target.elementId}`}
```
mrcfps: Non-blocking: please avoid deriving this dialog IDREF from `target.elementId`. With the new fallback IDs/selectors, `target.elementId` can be a generated CSS path such as `div:nth-of-type(1) > section:nth-of-type(2)`, which includes spaces and combinators; using that string in both `aria-labelledby` and the title `id` makes the IDREF fragile for assistive tech and can break the label for otherwise valid fallback targets. A safer fix is to generate a stable/sanitized DOM id for the composer title (for example with `useId` in `BoardComposerPopover`) and keep `target.elementId` only as visible text. 🙂

apps/web/src/components/FileViewer.tsx
```diff
@@ -1758,21 +2184,25 @@ function HtmlViewer({
   useEffect(() => {
     const win = iframeRef.current?.contentWindow;
     if (!win) return;
-    win.postMessage({ type: 'od:comment-mode', enabled: commentMode }, '*');
-  }, [commentMode, srcDoc]);
+    win.postMessage({ type: 'od:comment-mode', enabled: boardMode, mode: boardTool }, '*');
```
mrcfps, line 2188: Non-blocking: Pods mode currently depends on this one host `postMessage` landing after the newly loaded iframe has installed the injected bridge listener, while the bridge itself defaults to `mode = 'picker'`. If the iframe is reloaded or the file changes while the toolbar still has Pods selected, this message can race the srcDoc load; the host continues showing Pods active, but the iframe remains in picker mode and subsequent clicks on the already-active Pods button do not resend a different state. Consider either injecting the initial board tool into `buildSrcdoc`/`injectCommentBridge`, or resending the same `{ type: 'od:comment-mode', enabled: boardMode, mode: boardTool }` from the iframe `onLoad` handler so the bridge always converges after reloads.

mrcfps (commented):
Found two non-blocking follow-ups around the new Tweaks UI plumbing. Neither looks merge-blocking, but tightening them would make Pods mode and assistive labels more reliable.


lefarcen (commented):
Clean re-review on head a56c1de7 ✅

**Addressed:**
- aria-labelledby IDREF safety: now uses React useId() for stable title IDs — fragile CSS-path IDs no longer break assistive tech labels
- Pods mode initialization race: added iframe onLoad + syncCommentMode() to ensure bridge always converges after reloads

**New code:**
- Pod member normalization (arrays, validation, memberCount handling) — all looks solid: finite checks, clamping, length limits, null filtering

**Verdict:** No P1/P2 blockers. mrcfps's non-blocking P3 suggestions are addressed. Ready from my side; deferring final approval to maintainer.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1184,6 +1188,14 @@ export function ProjectView({
     ],
   );
 
+  const handleSendBoardCommentAttachments = useCallback(
+    async (commentAttachments: ChatCommentAttachment[]) => {
+      if (commentAttachments.length === 0) return;
+      await handleSend('', [], commentAttachments);
```
mrcfps, line 1196: Blocking: please keep board batch sends behind the same single-run guard as the main chat composer. This new handler always calls `handleSend('', [], commentAttachments)` when it receives attachments, but it does not check `streaming`, and the FileViewer batch button only disables itself for its own short-lived `sendingBoardBatch` state. Because `handleSend` immediately allocates new `AbortController`s and overwrites the singleton `abortRef`/`cancelRef`/`streaming` state for every run, a user can send a Tweaks batch while another agent response is still active; stop/cancel/status updates can then target the wrong run, and one run finishing can clear `streaming` while the other is still in flight. A safe fix is to guard centrally in `handleSend` (or at least return early here when `streaming` is true) and pass the active streaming state through to disable the board composer’s “Send to chat” button until the current run completes.

mrcfps (changes requested):
Found one blocking issue in the board-comment send path: the new Tweaks batch action can start an agent run while another run is already active.


lefarcen (commented):
Clean re-review on head b41b58d4 ✅

**Addressed:**
mrcfps's race condition finding is fully fixed across all three parts:
- handleSend now guards centrally: `if (streaming) return;` before allocating new AbortControllers
- handleSendBoardCommentAttachments guards: `if (streaming || ...) return;` before calling handleSend
- streaming prop flows from ProjectView → FileViewer → HtmlViewer, enabling board composer button disable during active runs

**Verdict:** No P1/P2 blockers. The concurrent-run race is resolved. Ready from my side; deferring final approval to maintainer.

mrcfps (approved):
@Martin-Atrin I reviewed the current head’s Tweaks-mode changes across the contract, daemon persistence/attachment handling, HTML bridge, viewer UI, and targeted tests. The prior migration, pod geometry/scale, selector escaping, reload sync, and active-run guard feedback is addressed, and I did not find any remaining actionable correctness or maintainability issues in the changed ranges. Nice iteration on tightening this workflow.


lefarcen (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @Martin-Atrin for the contribution to open-design. Looking forward to the next one!

leilei926524-tech:
Hi @Martin-Atrin — welcome to Open Design contributors. 🎉

Your first merged PR (#513) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

Issue #398, opened 2026-05-03
Claude Code charges API usage despite "local CLI" auth setup with Pro account


sebastianwestberg:
I'm using Claude Code with a Claude Pro subscription. During setup, I selected "Local CLI" as the authentication method (see screenshots), expecting usage to be covered by my Pro plan rather than billed against an API key.

However, I'm still seeing charges for API usage on my account.

Steps to reproduce

    Set up Claude Code with a Pro account
    Choose "Local CLI" as the auth method during setup
    Use Claude Code normally
    Observe API usage charges accruing

What I've tried

    Unsetting the ANTHROPIC_API_KEY environment variable
    Confirming via /login (or settings) that "Local CLI" is the selected auth method

API usage continues to be billed regardless.

Expected behavior

With "Local CLI" selected and a Pro subscription, usage should be drawn from the Pro plan quota and not generate API charges.

Actual behavior

Usage is being billed as API usage (see third screenshot).

Environment

    Auth method: Local CLI
    Plan: Claude Pro
    ANTHROPIC_API_KEY: unset

Question

Is this expected behavior, or is there an additional step required to ensure Claude Code uses the Pro subscription instead of API billing?

Screenshots

<img width="290" height="417" alt="Image" src="https://github.com/user-attachments/assets/c6e5f134-2d6c-4c2c-a29b-5918d3a7e62d" />
<img width="904" height="790" alt="Image" src="https://github.com/user-attachments/assets/0b1b5811-00b1-4586-b528-e90565c4c717" />

<img width="426" height="139" alt="Image" src="https://github.com/user-attachments/assets/e0cb8173-a578-45b0-8758-4447608c00a0" />

lefarcen:
Hi @sebastianwestberg! 👋

Thanks for opening this issue. I can see from your screenshots that you're experiencing billing issues with **Claude Code** (Anthropic's agentic coding tool), where the "Local CLI" authentication method is still charging API usage despite your Claude Pro subscription.

However, **open-design** is a different project — it's a local-first alternative to Claude Design for creating design systems and HTML exports. We're not affiliated with Anthropic's Claude Code product.

For support with Claude Code billing and authentication, I'd recommend:
1. **Anthropic Support**: https://support.anthropic.com or support@anthropic.com
2. **Claude Code Documentation**: https://claude.com/product/claude-code
3. **Anthropic Community/Discord**: They may have specific channels for Claude Code billing questions

[...]

PR #514, opened 2026-05-05
fix(daemon): preserve ANTHROPIC_API_KEY when ANTHROPIC_BASE_URL is set

VioletCoding:
## Summary

  When `ANTHROPIC_BASE_URL` is set, the user is intentionally routing Claude Code to a custom endpoint (e.g. a Kimi/Moonshot
  proxy). In that case `claude login` is meaningless because the endpoint is not Anthropic's.

  Currently `spawnEnvForAgent` unconditionally strips `ANTHROPIC_API_KEY` for the claude adapter (issue #398). This makes sense
   for the default case — it prevents Claude Code from silently falling back to API-key billing when the daemon happens to be
  launched from a shell that exported the key for SDK use.

  However, it breaks users who rely on custom base URLs + API keys. This PR preserves `ANTHROPIC_API_KEY` when
  `ANTHROPIC_BASE_URL` is detected, while keeping the stripping behavior for the default (no custom base URL) case.

  ## Changes

  - `apps/daemon/src/agents.ts`: check for `ANTHROPIC_BASE_URL` before stripping `ANTHROPIC_API_KEY`
  - `apps/daemon/tests/agents.test.ts`: add test for the new behavior

  ## Test plan

  - [x] spawnEnvForAgent tests pass

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -1173,7 +1173,10 @@ export function spawnEnvForAgent(agentId, baseEnv) {
   const env = { ...baseEnv };
   if (agentId !== 'claude') return env;
   const hasCustomBaseUrl = Object.keys(env).some(
-    (k) => k.toUpperCase() === 'ANTHROPIC_BASE_URL',
+    (k) =>
+      k.toUpperCase() === 'ANTHROPIC_BASE_URL' &&
+      typeof env[k] === 'string' &&
+      env[k].trim() !== '',
   );
   if (hasCustomBaseUrl) return env;
   for (const key of Object.keys(env)) {
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -1159,6 +1159,11 @@ export function resolveAgentBin(id) {
 // launched from a shell that exported the key for SDK or scripting use.
 // See issue #398.
 //
+// However, when ANTHROPIC_BASE_URL is set the user is intentionally
+// routing Claude Code to a custom endpoint (e.g. a Kimi/Moonshot proxy).
+// In that case claude login is meaningless, so preserve the API key so
+// the child can authenticate against the custom base URL.
+//
 // Windows env-var names are case-insensitive at the kernel level
 // (`GetEnvironmentVariable`), but spreading `process.env` into a plain
 // object loses Node's case-insensitive accessor — `Anthropic_Api_Key`
@@ -1167,6 +1172,10 @@ export function resolveAgentBin(id) {
 export function spawnEnvForAgent(agentId, baseEnv) {
   const env = { ...baseEnv };
   if (agentId !== 'claude') return env;
+  const hasCustomBaseUrl = Object.keys(env).some(
+    (k) => k.toUpperCase() === 'ANTHROPIC_BASE_URL',
+  );
+  if (hasCustomBaseUrl) return env;
   for (const key of Object.keys(env)) {
     if (key.toUpperCase() === 'ANTHROPIC_API_KEY') delete env[key];
   }
--- a/apps/daemon/tests/agents.test.ts
+++ b/apps/daemon/tests/agents.test.ts
@@ -1093,6 +1093,28 @@ test('spawnEnvForAgent preserves ANTHROPIC_API_KEY when ANTHROPIC_BASE_URL is se
   assert.equal(env.PATH, '/usr/bin');
 });
 
+test('spawnEnvForAgent strips ANTHROPIC_API_KEY when ANTHROPIC_BASE_URL is empty', () => {
+  const env = spawnEnvForAgent('claude', {
+    ANTHROPIC_API_KEY: 'sk-leak',
+    ANTHROPIC_BASE_URL: '',
+    PATH: '/usr/bin',
+  });
+
+  assert.equal('ANTHROPIC_API_KEY' in env, false);
+  assert.equal(env.PATH, '/usr/bin');
+});
+
+test('spawnEnvForAgent strips ANTHROPIC_API_KEY when ANTHROPIC_BASE_URL is whitespace', () => {
+  const env = spawnEnvForAgent('claude', {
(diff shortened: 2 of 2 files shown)


lefarcen:
Hi @VioletCoding! 🎉
Thanks for the contribution — this is an important fix for custom base URL + API key usage.

— open-design team

apps/daemon/src/agents.ts
```diff
@@ -1167,6 +1172,10 @@ export function resolveAgentBin(id) {
 export function spawnEnvForAgent(agentId, baseEnv) {
   const env = { ...baseEnv };
   if (agentId !== 'claude') return env;
+  const hasCustomBaseUrl = Object.keys(env).some(
+    (k) => k.toUpperCase() === 'ANTHROPIC_BASE_URL',
+  );
+  if (hasCustomBaseUrl) return env;
```
mrcfps, line 1181: This preserves `ANTHROPIC_API_KEY` based only on the presence of an `ANTHROPIC_BASE_URL` key, not on whether the value is usable. If a shell, launch config, or `.env` template exports `ANTHROPIC_BASE_URL=` as an empty string, this branch returns early while Claude Code can still fall back to the default Anthropic endpoint; that reintroduces the unintended API-key billing path the surrounding issue #398 logic is explicitly trying to prevent. Please require a non-empty trimmed base URL before returning here, and keep stripping `ANTHROPIC_API_KEY` when the base URL value is blank; adding a regression test with `ANTHROPIC_BASE_URL: ''` would lock in the safety behavior. 🙂

mrcfps (changes requested):
I found one safety issue in the changed environment handling that should be tightened before merge.


apps/daemon/src/agents.ts
```diff
@@ -1167,6 +1172,10 @@ export function resolveAgentBin(id) {
 export function spawnEnvForAgent(agentId, baseEnv) {
   const env = { ...baseEnv };
   if (agentId !== 'claude') return env;
+  const hasCustomBaseUrl = Object.keys(env).some(
+    (k) => k.toUpperCase() === 'ANTHROPIC_BASE_URL',
```
lefarcen: **P2 should-fix** — `hasCustomBaseUrl` checks only for the key's existence, not whether it's non-empty. If the daemon is launched with `ANTHROPIC_BASE_URL=''` or whitespace, this will preserve `ANTHROPIC_API_KEY` and potentially reintroduce the #398 billing guard bypass.

Suggested fix:
```typescript
const hasCustomBaseUrl = Object.keys(env).some(
  (k) => k.toUpperCase() === 'ANTHROPIC_BASE_URL' && typeof env[k] === 'string' && env[k].trim() !== '',
);
```
This ensures we only preserve the API key when there's an actual custom URL, not just an empty env var.

apps/daemon/tests/agents.test.ts
```diff
@@ -1081,6 +1081,18 @@ test('spawnEnvForAgent preserves ANTHROPIC_API_KEY for non-claude adapters', ()
   }
 });
 
+test('spawnEnvForAgent preserves ANTHROPIC_API_KEY when ANTHROPIC_BASE_URL is set', () => {
```
lefarcen, line 1084: **P2 should-fix** — The new test covers the happy path, but not the edge case above. Add a test asserting that an empty or whitespace-only `ANTHROPIC_BASE_URL` still strips `ANTHROPIC_API_KEY`:

```typescript
test('spawnEnvForAgent strips ANTHROPIC_API_KEY when ANTHROPIC_BASE_URL is empty', () => {
  const env = spawnEnvForAgent('claude', {
    ANTHROPIC_API_KEY: 'sk-leak',
    ANTHROPIC_BASE_URL: '',
    PATH: '/usr/bin',
  });

  assert.equal(env.ANTHROPIC_API_KEY, undefined);
  assert.equal(env.ANTHROPIC_BASE_URL, '');
  assert.equal(env.PATH, '/usr/bin');
});
```

lefarcen (changes requested):
Hi @VioletCoding! 👋

Thanks for tackling this — the logic is sound, but there's an edge case that needs addressing before merge.

The core issue: `hasCustomBaseUrl` checks only for the key's presence, not whether it contains a usable value. If someone launches the daemon with `ANTHROPIC_BASE_URL=''` or whitespace, the code will preserve `ANTHROPIC_API_KEY`, potentially reintroducing the #398 billing guard bypass.

See inline comments below for the specific fix.

VioletCoding:
Addressed, PTAL.

lefarcen (approved):
Perfect! 🎉

Both P2 issues are now resolved:

1. ✅ Non-empty value check — `hasCustomBaseUrl` now requires `typeof env[k] === 'string' && env[k].trim() !== ''`
2. ✅ Regression tests — Added test coverage for empty string and whitespace `ANTHROPIC_BASE_URL`

The fix correctly preserves `ANTHROPIC_API_KEY` only when there's a real custom base URL, while keeping the #398 billing guard intact for the default case.

mrcfps (approved):
@VioletCoding I reviewed the updated `spawnEnvForAgent` change and the added regression coverage. The current diff now preserves `ANTHROPIC_API_KEY` only when `ANTHROPIC_BASE_URL` has a non-empty value, keeps the default Claude billing guard intact for blank/missing base URLs, and adds tests for the custom URL, empty string, and whitespace cases. I could not execute the focused Vitest command in this worktree because dependencies are not installed (`vitest` is missing), so this pass is based on the live diff and surrounding context. Thanks for the quick, careful follow-up here — this looks ready from my review. 🙂


mrcfps:
@VioletCoding Merged. Thanks for your contribution and quick follow-ups. Looking forward to your next PR! 🎉

leilei926524-tech:
Hi @VioletCoding — welcome to Open Design contributors. 🎉

Your first merged PR (#514) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #518, opened 2026-05-05
feat: Allow composing follow-up messages while a run is in flight ("queued message")

### LINKED ISSUE #446: feat(web): allow composing follow-up messages while a run is in flight ("queued message")
## Summary
Add the ability to compose and queue a follow-up message while a Claude run is still streaming. Today the chat input is locked during a run — the only options are wait for completion or click Stop. This proposal lets users type a follow-up while the agent is working; the message dispatches automatically as soon as the current run finishes.

## Motivation
Anthropic's claude.ai design surface (Claude Design) supports this pattern. Mid-run, you realize you want to add scope, redirect, or attach a related sub-task — you can type the addition while the agent is finishing, and it gets sent the moment the current turn ends. Open Design's locked-input model forces users to either lose their train of thought, or interrupt with Stop and lose any in-flight work.

## Proposed UX (minimum viable)
- Text area unlocks the moment a run starts.
- \`Send\` button label flips to \`Queue\` while a run is active.
- Queued message renders below the live response with a \"queued — will send when current run completes\" badge. Editable until dispatch.
- On run completion, daemon dispatches the queued message as the next turn automatically.
- One queued message at a time. If the user attempts a second, replace the first (with a small confirmation flash).
- No interrupt semantics in v1 — explicit \`Stop\` is still the only way to halt the current run.

## Non-goals (deferred)
- \"Interrupt and replace\" semantics (queue arrives as Stop+Send) — the race-condition surface around cancel teardown isn't worth folding into v1.
- Multiple queued messages / a true queue depth > 1.
- Edit of already-streamed previous turns.

## Implementation hints
- Daemon \`apps/daemon/src/run-stream.ts\` already treats each run as a finite-lifetime stream. Producer side just needs an in-memory \`pendingNextMessage\` slot per project; on stream completion, dispatch if non-empty.
- Web composer gates the input on \`runState !== 'idle'\`. Change the gate to \`false\` (always editable), flip the submit-action label based on \`runState\`, and route submit through a new \`enqueueMessage\` channel during active runs.
- No SSE schema changes needed — dispatch happens on the daemon side and just produces the next run, which already flows through existing event channels.

## Why this is worth shipping
- Closes an obvious affordance gap with claude.ai's chat UX, which Open Design users already pattern-match against.
- Low risk: one queued slot, no interrupt semantics, no schema changes, no migration.
- High value per session: every multi-step design iteration benefits from \"let me add one more thing\" without breaking flow.

Moe1211:
Resolves #446.

This PR adds the ability for users to compose and queue a follow-up message while a Claude run is still streaming. It implements the queue locally in the UI for immediate responsiveness and backs it with daemon-level sequencing so the queue survives page reloads.
### Changes
**Frontend (`apps/web`)**
- Unlocked the composer input during active runs.
- The `Send` button now reads `Queue` when streaming, while the `Stop` button remains fully accessible alongside it.
 - Added a `queuedMessage` state to `ProjectView` that intercepts sent messages during an active stream.
 - Added a `useEffect` to automatically dispatch the queued message when the stream transitions to a finished state.

 **Daemon (`apps/daemon`)**
 - Updated the run manager (`runs.ts`) to handle sequential execution.
 - Replaced direct `start()` calls with a new `enqueue()` method in the API layer (`server.ts`).
 - `enqueue()` checks if there is an active run for the given `conversationId`. If so, it leaves the new run in a `queued` state.
 - Hooked `maybeStartNext()` into the `finish` and `fail` lifecycle methods so the daemon automatically dequeues and starts the next run in the conversation, ensuring cross-reload queue persistence.


lefarcen:
Hi @Moe1211! 🎉
Thanks for the contribution — queued messages during active runs is a highly requested feature.

— open-design team

apps/daemon/src/runs.ts
```diff
@@ -94,6 +121,13 @@ export function createChatRunService({
     return run;
   };
 
+  const enqueue = (run, starter) => {
+    if (run.status !== 'queued') return run; // Only enqueue fresh runs
+    starters.set(run.id, starter);
+    maybeStartNext(run.conversationId);
+    return run;
```
lefarcen, line 133: **P1** — `enqueue` is defined but not exported from `createChatRunService`. Add `enqueue` to the returned service object (around line 172 where other methods are exported). Without this, `server.ts:3867` will throw `TypeError: design.runs.enqueue is not a function` on every chat run.

apps/daemon/src/server.ts
```diff
@@ -3899,7 +3899,7 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
   app.post('/api/chat', (req, res) => {
     const run = design.runs.create();
     design.runs.stream(run, req, res);
-    design.runs.start(run, () => startChatRun(req.body || {}, run));
+    design.runs.enqueue(run, () => startChatRun(req.body || {}, run));
```
lefarcen, line 5303: **P1** — The legacy `/api/chat` endpoint doesn't assign `conversationId` until inside `startChatRun`, but `enqueue` calls `maybeStartNext(run.conversationId)` immediately — before the ID is set. The queued run never starts because `conversationId` is `null`.

Concrete fix: Either (a) pass request body into `create(req.body || {})` to set `conversationId` early, or (b) make `enqueue` start immediately when no conversation ID exists (treating it as a one-off non-queued run).

apps/daemon/src/runs.ts
```diff
@@ -68,6 +72,28 @@ export function createChatRunService({
...
+    const activeRun = conversationRuns.find((r) => !TERMINAL_RUN_STATUSES.has(r.status) && r.status !== 'queued');
+    if (activeRun) return;
+
+    // Find the oldest queued run.
+    const nextRun = conversationRuns.find((r) => r.status === 'queued');
+    if (nextRun) {
+      const starter = starters.get(nextRun.id);
+      if (starter) {
+        // Remove from starters map once we're about to run it
+        starters.delete(nextRun.id);
```
lefarcen, line 91: **P1** — Race condition: if a second run is enqueued while the selected run is in async preflight (before `startChatRun` marks it non-queued), `maybeStartNext` sees no active non-queued run and starts the second one too, violating one-run-per-conversation.

Concrete fix: Mark the selected run as internally active **before** invoking `start(nextRun, starter)`, or add an `isStarting` flag and treat it as active.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1184,6 +1190,18 @@ export function ProjectView({
     ],
   );
 
+  // Dispatch queued messages after the current run finishes.
+  useEffect(() => {
```
lefarcen, line 1417: **P2** — The queued message auto-dispatch works for UI-level state, but if the UI sends a second turn while streaming, the daemon-queued run will execute with a **stale transcript snapshot** from enqueue time, not the completed assistant response.

Concrete fix: Either (a) queue only locally and call `streamViaDaemon` after the current run finishes (simpler), or (b) change the daemon API to accept the new user prompt plus conversation IDs, and resolve the final persisted history at actual start time.

apps/web/src/components/ChatComposer.tsx
```diff
@@ -795,22 +795,22 @@ export const ChatComposer = forwardRef<ChatComposerHandle, Props>(
...
-              >
-                <Icon name="send" size={13} />
-                <span>{t('chat.send')}</span>
-              </button>
-            )}
+            ) : null}
+            <button
+              type="button"
+              className="composer-send"
+              data-testid="chat-send"
```
lefarcen, line 861: **P2** — The `Queue` button now renders while streaming, but clicking it is a **no-op** because `submit()` returns early when `streaming` is true, and `setQueuedMessage` is never called.

Concrete fix: Wire `onSend` during streaming to set the queued message in `ProjectView`, clear/reset the composer only after enqueue succeeds, and render queued state so users can see/cancel it.

apps/web/src/components/ChatComposer.tsx
```diff
@@ -795,22 +795,22 @@ export const ChatComposer = forwardRef<ChatComposerHandle, Props>(
...
+              onClick={() => void submit()}
+              disabled={!draft.trim() && commentAttachments.length === 0}
+            >
+              <Icon name="send" size={13} />
+              <span>{streaming ? 'Queue' : t('chat.send')}</span>
```
lefarcen: **P3** — `'Queue'` is hardcoded and bypasses the i18n system. Add `chat.queue` to `Dict` and all locale files, then render `t('chat.queue')`.

lefarcen (changes requested):
Hey @Moe1211, great work on implementing queued messages! The architecture is sound — separating UI queueing from daemon sequencing is the right approach.

However, the review found **three critical issues** that will prevent this from working in production:

1. **`enqueue` not exported** — `server.ts` calls `design.runs.enqueue()` but the function is never returned from `createChatRunService`. Every chat run will throw `TypeError: enqueue is not a function`.

2. **Legacy `/api/chat` endpoint broken** — this endpoint creates runs without a `conversationId`, so the queuing logic never fires. Streaming clients using this endpoint will hang.

3. **Race condition** — if a second run is enqueued while the first is in async preflight, both can start simultaneously because the selected run remains `status === 'queued'` until `startChatRun` reaches the spawn point.

See inline comments below for concrete fixes. Once these are addressed, the implementation will be solid. 🚀

apps/daemon/src/server.ts
```diff
@@ -3864,7 +3864,7 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     /** @type {import('@open-design/contracts').ChatRunCreateResponse} */
     const body = { runId: run.id };
     res.status(202).json(body);
-    design.runs.start(run, () => startChatRun(req.body || {}, run));
+    design.runs.enqueue(run, () => startChatRun(req.body || {}, run));
```
mrcfps, line 5268: `server.ts` now calls `design.runs.enqueue(...)`, but the changed service implementation only defines `enqueue` locally and the returned object from `createChatRunService` still does not expose it. That means the first `/api/runs` request reaches this line and throws `TypeError: design.runs.enqueue is not a function` instead of starting or queueing the run, breaking the main chat path. Please add `enqueue` to the service return in `apps/daemon/src/runs.ts` (and cover the `/api/runs` create/start path with a test) before switching callers to it.

apps/daemon/src/runs.ts
```diff
@@ -68,6 +72,28 @@ export function createChatRunService({
     signal: run.signal,
   });
 
+  const maybeStartNext = (conversationId) => {
+    if (!conversationId) return;
```
mrcfps, line 76: `maybeStartNext` immediately returns when `conversationId` is missing, but `server.ts` also changed the legacy `/api/chat` endpoint to call `enqueue` on a run created with `design.runs.create()` and no metadata. Once `enqueue` is exported, that endpoint will leave its run in `queued` forever: the SSE client is attached, but no starter ever runs and no terminal event is emitted. Please either create `/api/chat` runs with the request metadata before enqueueing, or have `enqueue` fall back to `start(run, starter)` for runs without a conversation id.

apps/daemon/src/runs.ts
```diff
@@ -68,6 +72,28 @@ export function createChatRunService({
...
+    if (activeRun) return;
+
+    // Find the oldest queued run.
+    const nextRun = conversationRuns.find((r) => r.status === 'queued');
+    if (nextRun) {
+      const starter = starters.get(nextRun.id);
+      if (starter) {
+        // Remove from starters map once we're about to run it
+        starters.delete(nextRun.id);
+        start(nextRun, starter);
```
mrcfps, line 94: The selected queued run is not marked active before `start(nextRun, starter)` invokes the async `startChatRun` preflight. During that window the run is still `status === 'queued'`, so another `enqueue` for the same conversation can call `maybeStartNext`, see no active non-queued run, and start a second queued run concurrently. That violates the one-run-at-a-time sequencing this change is adding. Please mark the run as starting/running before invoking the starter (or add an `isStarting` state that the active-run check treats as active), then add a fixture that enqueues two runs while the first starter is still awaiting preflight.

apps/web/src/components/ChatComposer.tsx
mrcfps: The UI now renders a `Queue` button while `streaming` is true, but this click still goes through `submit()`, which returns early on `streaming`, and `ProjectView` never populates `queuedMessage` before the new dispatch effect reads it. In the primary flow users can type a follow-up while a run streams, click Queue, and nothing is enqueued or sent. Please allow the composer submit path to call `onSend` while streaming and have `ProjectView.handleSend` store `{ prompt, attachments, commentAttachments }` with `setQueuedMessage`/return before creating another assistant run; then the existing post-stream effect can dispatch it.

mrcfps (changes requested):
@Moe1211, thanks for pushing this queued-message path forward. I found a few blocking correctness issues that keep the current head from reliably starting runs or queueing follow-up messages, so I’m requesting changes for this pass.


apps/daemon/src/server.ts
mrcfps, line 5268: This still calls `design.runs.enqueue(...)`, but `createChatRunService()` only defines `enqueue` locally and the returned service object in `apps/daemon/src/runs.ts` does not include it. That means the primary `/api/runs` path returns `202` and then immediately throws `TypeError: design.runs.enqueue is not a function`, so no run is started or queued. Please add `enqueue` to the returned service API (and cover this route with a regression test) before switching callers over. 🚧

apps/daemon/src/server.ts
```diff
@@ -3897,9 +3897,9 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
   });
 
   app.post('/api/chat', (req, res) => {
-    const run = design.runs.create();
+    const run = design.runs.create(req.body || {});
```
mrcfps, line 5301: Passing `req.body || {}` into `create()` fixes callers that include a `conversationId`, but direct/legacy `/api/chat` requests can still omit it. In that case `enqueue()` stores the starter, `maybeStartNext(run.conversationId)` returns early because the ID is null, and the attached SSE stream never starts or ends. Please either make `enqueue()` fall back to `start(run, starter)` when there is no conversation id, or reject `/api/chat` requests without a conversation id with a clear 400 if the endpoint now requires one.

apps/web/src/components/ChatComposer.tsx
```diff
@@ -488,12 +488,11 @@ export const ChatComposer = forwardRef<ChatComposerHandle, Props>(
...
       const hatched = expandHatchCommand(prompt);
       if (hatched) {
-        if (streaming) return;
         onSend(hatched, staged, commentAttachments);
         reset();
         return;
       }
-      if ((!prompt && commentAttachments.length === 0) || streaming) return;
+      if (!prompt && commentAttachments.length === 0) return;
       onSend(prompt, staged, commentAttachments);
```
mrcfps, line 551: Removing the streaming guard lets the new Queue button call `onSend(...)`, but `ProjectView.handleSend` does not store `queuedMessage` before creating the user/assistant messages, controllers, and stream. The only new `queuedMessage` usages initialize it, clear it on stop, and dispatch it after streaming ends, so clicking Queue while a run is active starts another frontend run immediately instead of queueing the follow-up locally. That can overwrite the single `abortRef`/`cancelRef` and make Stop/streaming state track the wrong run. Please add an early `if (streaming) { setQueuedMessage({ prompt, attachments, commentAttachments }); return; }` path in `handleSend` before creating the new run, then let the existing post-stream effect dispatch it.

mrcfps (changes requested):
@Moe1211, thanks for the follow-up fixes here. I re-reviewed the current head and found a few blocking correctness issues that still keep the main queued-message flow from reliably starting runs and queueing the follow-up locally.


apps/daemon/src/runs.ts
```diff
@@ -94,6 +123,13 @@ export function createChatRunService({
     return run;
   };
 
+  const enqueue = (run, starter) => {
+    if (run.status !== 'queued') return run; // Only enqueue fresh runs
+    starters.set(run.id, starter);
+    maybeStartNext(run.conversationId);
+    return run;
```
lefarcen, line 133: **P1 (CRITICAL — still not fixed)** — `enqueue` is defined but not exported. You need to add it to the service return object (around line 172 where `create`, `start`, etc. are returned). Without this, `server.ts:3867` throws `TypeError: design.runs.enqueue is not a function` on the first `/api/runs` or `/api/chat` request.

apps/daemon/src/server.ts
lefarcen, line 5301: **P1** — Passing `req.body` to `create()` only fixes `/api/chat` when callers supply `conversationId`. Existing clients can omit it, and `maybeStartNext()` returns early for null, so those runs queue forever.

Concrete fix: In `enqueue()`, fall back to `start(run, starter)` immediately when `run.conversationId` is null (no serialization key = no queue needed).

apps/daemon/src/runs.ts
```diff
@@ -68,6 +72,30 @@ export function createChatRunService({
...
+
+    // Find the oldest queued run.
+    const nextRun = conversationRuns.find((r) => r.status === 'queued');
+    if (nextRun) {
+      const starter = starters.get(nextRun.id);
+      if (starter) {
+        // Remove from starters map once we're about to run it
+        starters.delete(nextRun.id);
+        // Mark as starting to prevent race condition while async preflight runs
+        nextRun.status = 'starting';
```
lefarcen: **P2** — `status = 'starting'` closes the race, but `'starting'` is not in the `ChatRunStatus` contract, so web helpers like `isActiveRunStatus` won't recognize it.

Concrete fix: Either add `'starting'` to `packages/contracts/src/api/chat.ts` and all consumers, or keep it internal (e.g., `run._isStarting = true`) while leaving public status as `'queued'` until the real `start()` call.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1184,6 +1190,18 @@ export function ProjectView({
...
+    if (!streaming && queuedMessage) {
```
lefarcen, line 1418: **P2** — The Queue button now calls `handleSend`, but `handleSend` still snapshots `nextHistory = [...messages, userMsg]` immediately, **while the current assistant response may still be partial**. The `queuedMessage` state/effect you added isn't actually preventing this.

Concrete fix: In `handleSend`, if `streaming` is true, call `setQueuedMessage({ prompt, attachments, commentAttachments })` and return **before** creating messages or calling `streamViaDaemon`. The effect will dispatch it after the stream finishes.

lefarcen (changes requested):
Thanks for the quick update, @Moe1211! You've made good progress — the race condition mitigation and i18n fixes look solid. However, **the critical blocker is still present**: `enqueue` is defined but not exported from `createChatRunService`, so the first chat request will still throw `TypeError: enqueue is not a function`.

See inline for the remaining items. Once `enqueue` is in the service return object, this should be ready to ship. 🚀

apps/daemon/src/runs.ts
lefarcen, line 133: **P1 (CRITICAL — still not fixed)** — `enqueue` is defined but not returned from `createChatRunService`. You need to add `enqueue` to the service return object (around line 174 where `create`, `start`, `get`, etc. are returned). Without this, `server.ts:3867` and `server.ts:3902` will throw `TypeError: design.runs.enqueue is not a function` on the first `/api/runs` or `/api/chat` request.

apps/daemon/src/runs.ts
```diff
@@ -94,6 +123,13 @@ export function createChatRunService({
     return run;
   };
 
+  const enqueue = (run, starter) => {
```
lefarcen, line 128: **P1** — `enqueue()` calls `maybeStartNext(run.conversationId)`, but `maybeStartNext` returns immediately when the ID is null. `/api/chat` can create runs with `design.runs.create()` without metadata, so those runs will be enqueued but never started.

Concrete fix: Add an early fallback in `enqueue()`: `if (!run.conversationId) { start(run, starter); return run; }`

lefarcen (changes requested):
Thanks for the updates, @Moe1211! Good progress — the 'starting' status and Queue button defer are now fixed ✅. However, **the two critical P1 blockers remain**: `enqueue` is still not exported, and `/api/chat` runs without conversationId will queue forever. These will cause runtime failures on the main chat path.

See inline for the remaining items.

apps/daemon/src/runs.ts
```diff
@@ -152,3 +188,7 @@ export function createChatRunService({
     },
   };
 }
+atus);
```
mrcfps: This trailing fragment makes `apps/daemon/src/runs.ts` syntactically invalid: the current daemon typecheck stops with `TS1434/TS1128` at lines 191-194 before the server can build or start. Because this file is imported by the daemon startup path, the queued-message feature cannot run at all in this head. Please remove the pasted `atus); ...` block and keep the service return closed once at line 190, then rerun the daemon typecheck.

apps/daemon/src/runs.ts
```diff
@@ -68,6 +72,30 @@ export function createChatRunService({
...
+
+    // Find the oldest queued run.
+    const nextRun = conversationRuns.find((r) => r.status === 'queued' && !r._isStarting);
+    if (nextRun) {
+      const starter = starters.get(nextRun.id);
+      if (starter) {
+        // Remove from starters map once we're about to run it
+        starters.delete(nextRun.id);
+        // Mark as starting to prevent race condition while async preflight runs
+        nextRun._isStarting = true;
```
mrcfps, line 93: `nextRun._isStarting = true` is never cleared, and the active-run predicate on line 82 treats `_isStarting` as active even after the run reaches a terminal status. When the first queued run finishes, `finish()` calls `maybeStartNext()` while that completed run still has `_isStarting`, so the next queued run for the same conversation is blocked until cleanup removes the old run (up to the 30 minute TTL). Please clear `_isStarting` when the starter transitions into `start()`/`finish()`, or change the active check to ignore `_isStarting` on terminal runs, and add a regression that two queued runs in one conversation drain sequentially.

apps/web/src/i18n/types.ts
```diff
@@ -508,6 +508,7 @@ export interface Dict {
   'chat.linkedFolderAlready': string;
   'chat.linkedFolderPickError': string;
   'chat.send': string;
+  'chat.queue': string;
```
mrcfps, line 705: Adding `chat.queue` to the required `Dict` interface means every `export const <locale>: Dict` must define it, but this PR only updates seven locale files. The remaining locales (for example `apps/web/src/i18n/locales/de.ts`, plus ar/es-ES/fr/hu/ja/ko/pl/tr) still have `chat.send` without `chat.queue`, so a normal installed web typecheck will fail once it checks those `Dict` assignments. Please add `chat.queue` to every locale object or make the key optional with an explicit fallback before using `t('chat.queue')`.

mrcfps (changes requested):
@Moe1211, thanks for the quick follow-up. I re-reviewed the current head and found a few additional blocking issues that keep the daemon/web checks and queued-run sequencing from being safe to merge.


apps/daemon/src/server.ts
mrcfps, line 5268: This route now calls `design.runs.enqueue(...)` (and `/api/chat` does the same at line 3902), but the changed `createChatRunService()` in `apps/daemon/src/runs.ts` still returns `create`, `start`, `get`, etc. without returning `enqueue`. As soon as the API responds `202`, this line throws `TypeError: design.runs.enqueue is not a function`, so the primary daemon chat path neither starts nor queues any run. Please add `enqueue` to the returned service object and cover the `/api/runs` create/start path with a regression test.

apps/web/src/components/ProjectView.tsx
```diff
@@ -844,6 +849,10 @@ export function ProjectView({
     ) => {
       if (!activeConversationId) return;
       if (!prompt.trim() && attachments.length === 0 && commentAttachments.length === 0) return;
+      if (streaming) {
+        setQueuedMessage({ prompt, attachments, commentAttachments });
```
mrcfps: The queued message snapshot only stores `prompt`, `attachments`, and `commentAttachments`, then the dispatch effect later calls `handleSend(...)` against whatever `activeConversationId` is current at that time. If a user queues a follow-up while a run streams and switches conversations before the stream finishes, the queued prompt is sent to the newly selected conversation instead of the conversation where it was composed. Please store the originating `project.id`/`activeConversationId` with the queued item and dispatch only when they still match (or clear the queued item on project/conversation switches); a small regression test for queue-then-switch would lock this down.

fix_runs.sh
```diff
@@ -0,0 +1 @@
+sed -i '' -e '/atus);/,$d' apps/daemon/src/runs.ts
```
mrcfps: This new top-level `fix_runs.sh` looks like a one-off cleanup command rather than a product-owned script, and running it would destructively truncate `apps/daemon/src/runs.ts` from the first `atus);` match onward. Keeping scratch repair scripts in the repo makes future maintenance riskier because someone can run it outside the original context. Please remove this file from the PR (or move a real maintenance command under the appropriate tools package with documentation and tests).

mrcfps (changes requested):
@Moe1211, thanks for the follow-up fixes here. I re-reviewed the current head and found blocking issues that still break the queued-run path and can send a queued follow-up into the wrong conversation.


lefarcen (changes requested):
Great progress, @Moe1211! You've fixed 4 out of 5 critical blockers:

✅ Fixed:
- `_isStarting` now cleared in `finish()` and `start()` finally block
- All locale files now include `chat.queue`
- `enqueue()` correctly falls back to `start()` for null conversationId  
- Syntax error removed (file now ends cleanly)

❌ **Still blocking:**

`enqueue` is defined at line 129 but **not exported** from the service return object at line 178. This means `server.ts:3867` will still throw `TypeError: design.runs.enqueue is not a function` on the first `/api/runs` request.

**Fix:** Add `, enqueue` to the return object after `start` (line 180):

```typescript
return {
  create,
  start,
  enqueue,  // ← add this line
  get,
  list,
  stream,
  cancel,
  wait,
  emit,
  finish,
  fail,
  statusBody,
  isTerminal(status) {
    return TERMINAL_RUN_STATUSES.has(status);
  },
};
```

Once this is added, the implementation should be ready to ship. 🚀

lefarcen (approved):
Excellent work, @Moe1211! 🎉

You've successfully fixed **all 5 critical blockers** from the previous reviews:

✅ `enqueue` now exported from service return object (line 181)
✅ `_isStarting` properly cleared in `finish()` and `start()` finally block
✅ All 16 locale files include `chat.queue`
✅ `enqueue()` correctly falls back to `start()` for null conversationId
✅ Syntax error removed

The queued-message implementation is now ready from a technical correctness standpoint. Deferring final merge decision to a maintainer.

— open-design team

lefarcen (approved):
Perfect, @Moe1211! 🎉 All blockers are now resolved on head 9af2dfc5:

✅ **Fixed:**
- `enqueue` is now exported from the service return object (apps/daemon/src/runs.ts:180)
- Queued message snapshot now stores `projectId` and `conversationId` to prevent wrong-conversation dispatch (apps/web/src/components/ProjectView.tsx:194-195)
- `fix_runs.sh` scratch script removed

The implementation is ready to ship. 🚀

lefarcen (approved):
Re-confirmed on rebased head b2ac9953 🎉

All prior blockers remain fixed after the rebase:
- `enqueue` exported from service return (apps/daemon/src/runs.ts:181)
- Queued message stores projectId + conversationId (apps/web/src/components/ProjectView.tsx:195-196)

The rebase onto current main looks clean. Ready to ship! 🚀

apps/daemon/src/runs.ts
mrcfps: Looper checked this thread against head `b2ac9953697b181af3daff3d6fbcf8b6356a76b9`. The requested change appears objectively addressed by apps/daemon/src/runs.ts no longer contains the trailing 'atus);' fragment and ends with a single closed service return object., so I’m resolving this thread. Reopen if this still needs discussion.

fix_runs.sh
mrcfps: Looper checked this thread against head `b2ac9953697b181af3daff3d6fbcf8b6356a76b9`. The requested change appears objectively addressed by No top-level fix_runs.sh file exists in the current worktree., so I’m resolving this thread. Reopen if this still needs discussion.

apps/web/src/components/ProjectView.tsx
```diff
@@ -849,6 +856,16 @@ export function ProjectView({
       if (!activeConversationId) return;
       if (streaming) return;
       if (!prompt.trim() && attachments.length === 0 && commentAttachments.length === 0) return;
+      if (streaming) {
```
mrcfps: This new queueing branch is never reached because line 857 still returns immediately whenever `streaming` is true. The composer now allows `submit()` while streaming and resets the draft after calling `onSend`, so clicking Queue during an active run exits at that earlier guard and silently drops the follow-up instead of storing it in `queuedMessage`. Please remove the earlier `if (streaming) return;` or move the queueing block above it so the streaming case snapshots `{ prompt, attachments, commentAttachments, projectId, conversationId }` before returning.

mrcfps (changes requested):
@Moe1211, thank you for continuing to tighten up the queued-message flow. I found one remaining blocker in the web path: the new queueing block is still unreachable because the previous streaming guard returns first, so clicking Queue currently drops the message instead of saving it for dispatch.


apps/web/src/components/ProjectView.tsx
```diff
@@ -849,6 +856,16 @@ export function ProjectView({
       if (!activeConversationId) return;
       if (streaming) return;
```
lefarcen: **P1** — This early return still makes the queueing block (line 860) unreachable. When a user clicks Queue while a run streams, `onSubmitForm` now calls `handleSend` (✅ your fix), but execution exits here before reaching the `if (streaming) { setQueuedMessage(...); return; }` block below.

Please remove this line 857 guard so execution flows to the queueing block. The queueing block at line 860 already handles the streaming case correctly by storing the message and returning.

lefarcen (changes requested):
Hey @Moe1211! Thanks for the quick follow-up.

You removed the streaming guard from `onSubmitForm` — great instinct! But the blocker is actually one level deeper: `handleSend` itself still has an early `if (streaming) return;` at line 857 that exits before reaching the queueing block at line 860.

The flow now is:
1. User clicks Queue while streaming
2. `onSubmitForm` calls `handleSend` ✅ (guard removed here)
3. `handleSend` line 857 returns early ❌
4. Queueing block (line 860) never runs

One more tweak — see inline — and this should unblock the Queue path. 🙏

mrcfps:
@Moe1211 I'm holding off on generating review comments for nexu-io/open-design#518 because this pull request has merge conflicts right now.

lefarcen (approved):
Hey @Moe1211, excellent work addressing the queueing blocker! 🎉

The web path now snapshots `{ prompt, attachments, commentAttachments, projectId, conversationId }` before the streaming early return, so clicking Queue during an active run stores the follow-up locally instead of dropping it. This resolves the final blocking issue from mrcfps's last review.

All prior fixes remain intact:
- **Daemon**: `enqueue` is exported at `runs.ts:192` ✅
- **Web**: queuedMessage stores both `projectId` and `conversationId` (lines 987–988) ✅
- **Cleanup**: `fix_runs.sh` removed ✅

This PR is ready for a maintainer to merge. Thank you for the persistent iteration and clean resolution! 🙏

lefarcen (dismissed):
Hey @Moe1211, all prior blockers are resolved in this head! 🎉

**What changed since last review:**
- ✅ **Line 857 blocker fixed**: `ProjectView.handleSend` now checks `if (streaming)` BEFORE creating any assistant run (lines 986-993). The queueing block saves `{prompt, attachments, commentAttachments, projectId, conversationId}` to state and returns early.
- ✅ **ChatComposer unlocked**: removed `|| streaming` from the empty-check guard (line 494), so `onSend` is allowed during streaming.
- ✅ **Conversation-switch safety**: the dispatch `useEffect` verifies `queuedMessage.projectId === project.id` and `queuedMessage.conversationId === activeConversationId` before dispatching; clears the queue if user switched contexts.
- ✅ **Stop-button safety**: `handleStop` calls `setQueuedMessage(null)` to clear pending queue when user stops the stream.
- ✅ **All locale files**: `chat.queue` key added to every locale (ar, de, en, es, fa, fr, hu, id, it, ja, ko, pl, pt-BR, ru, tr, uk, zh-CN, zh-TW).
- ✅ **`enqueue` exported**: `apps/daemon/src/runs.ts` line 194 returns `enqueue` from the service API.
- ✅ **`_isStarting` cleared**: both `finish()` and `start().finally()` set `run._isStarting = false` to prevent queue stalls.

**Remaining gaps (non-blocking P2):**
- No tests added for the queueing path (daemon sequential-start fixture + web queue-then-dispatch fixture would be nice, but not required for merge).
- `_isStarting` is a runtime property without TypeScript typing (acceptable as private implementation detail).

The core queued-message flow is now correct and safe. Looks good to me; deferring final approval to a maintainer. 🚀

apps/web/src/components/ChatComposer.tsx
mrcfps: Looper checked this thread against head `5f4598f1f6e6397d0080a5c8f7a18d1ba133a75c`. The requested change appears objectively addressed by ChatComposer submit() no longer returns early while streaming and always calls onSend; ProjectView.handleSend now detects streaming, stores prompt/attachments/commentAttachments in setQueuedMessage, and returns before creating a run., so I’m resolving this thread. Reopen if this still needs discussion.

mrcfps (changes requested):
@Moe1211, thank you for the steady iteration on this queued-message flow. I found one remaining blocker around the promised reload-safe queueing behavior.

**Blocking — `apps/web/src/components/ProjectView.tsx`, `handleSend` streaming branch (lines 985-993):** this branch stores the queued follow-up only in component state and returns before creating any daemon run. That means the daemon-side `enqueue()` path added in `apps/daemon/src/runs.ts` never receives this message while the current run is active; if the user refreshes or closes the page before `streaming` flips false, `queuedMessage` is lost and there is no queued daemon run to recover. This breaks the PR's stated cross-reload behavior for the main Queue button flow.

Please persist or submit the queued item before returning there—for example, create a queued daemon run (or a small daemon-side queued-message record/API that can build the final history when it starts) and keep only UI state as a mirror. Add a regression that queues a follow-up during an active daemon run, remounts/reloads the web view before the first run finishes, and verifies the follow-up still drains from the daemon queue.


apps/web/src/components/ProjectView.tsx
mrcfps: Looper checked this thread against head `3a1c4683d589e48f6f68103dc1a68702874d98bb`. The requested change appears objectively addressed by ProjectView.tsx queuedMessage now stores projectId and conversationId (lines 282-288, 1074-1080), and the dispatch effect clears instead of sending if current project/conversation differ (lines 1391-1409). Daemon streaming follow-ups are posted immediately with current projectId/conversationId (lines 1031-1070)., so I’m resolving this thread. Reopen if this still needs discussion.

apps/web/src/components/ProjectView.tsx
mrcfps: Looper checked this thread against head `3a1c4683d589e48f6f68103dc1a68702874d98bb`. The requested change appears objectively addressed by handleSend no longer has an early streaming return before the streaming branch; when streaming is true it now either queues via /api/runs for daemon mode or snapshots queuedMessage with prompt, attachments, commentAttachments, projectId, and conversationId for API mode (lines 977-1081)., so I’m resolving this thread. Reopen if this still needs discussion.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1015,10 +1023,63 @@ export function ProjectView({
...
       const nextHistory = [...messages, userMsg];
+
+      if (streaming) {
+        if (config.mode === 'daemon') {
+          // Eagerly persist and queue via daemon to survive reloads
+          setMessages((curr) => [...curr, userMsg, assistantMsg]);
+          persistMessage(userMsg);
+          persistMessage(assistantMsg);
+
+          const transcript = nextHistory
```
mrcfps, line 1053: This builds the queued daemon run payload immediately from `nextHistory`, which is `messages` plus the new user prompt while the current run is still streaming. At that moment the current assistant message in `messages` can still be partial (or buffered just before the final chunks flush), and the daemon stores this `message` string in the starter closure until `maybeStartNext()` runs later. When the queued run finally starts, `startChatRun()` uses that stale transcript instead of the completed assistant response, so the follow-up can be answered without the final context from the turn it is supposed to follow.

Because this is the main daemon Queue path and the PR promises sequential follow-up behavior, please defer constructing the daemon prompt until the previous run has reached a terminal state, or persist a compact queued request and have the daemon/web rebuild the transcript from the finalized conversation before starting it. A regression should queue a follow-up while the first assistant response is still streaming, let the first response append additional/final content, and verify the second run receives the completed first response in its transcript.

mrcfps (changes requested):
@Moe1211, thank you for the continued iteration on the queued-message flow. I found one remaining blocking correctness issue: daemon-queued follow-ups are now persisted, but the prompt they persist can still be based on the in-flight assistant response before it has finished.


lefarcen (changes requested):
Hey @Moe1211, thanks for the continued iteration on this daemon-queueing path. I found two remaining blocking correctness issues:

**1. Cross-reload race still exists (apps/daemon/src/runs.ts:112):** `finish()` starts the next queued run immediately after emitting `end`, but the completed assistant message is persisted by the web client only after it receives/processes that terminal event. `rebuildTranscript` reads the DB at the next run's start, so the queued prompt can still miss or use a partial version of the just-finished assistant output. The original mrcfps blocker from head `3a1c4683` remains.

**2. Transcript rebuild includes future queued turns (apps/daemon/src/server.ts:3625):** `listMessages(conversationId)` rebuilds the entire conversation, not the history up to this queued run. If multiple follow-ups are queued, earlier queued runs will include later queued user turns + empty assistant placeholders, causing them to answer future messages out of order. Rebuild should cut off at this run's `assistantMessageId` or current user turn, excluding the current placeholder.

These are P1 correctness gaps — please address before the next push.

apps/daemon/src/runs.ts
```diff
@@ -80,6 +109,7 @@ export function createChatRunService({
     for (const waiter of run.waiters) waiter(statusBody(run));
     run.waiters.clear();
     scheduleCleanup(run);
+    maybeStartNext(run.conversationId);
```
mrcfps: Blocking: `finish()` emits the terminal `end` event and then immediately calls `maybeStartNext()` for the same conversation. The completed assistant message is persisted by the web client only after it consumes that terminal event and flushes its buffered text, while the queued starter now rebuilds its prompt from SQLite. That means the next run can start before the prior assistant response is fully saved, so the follow-up may be answered without the final context it was queued behind. Please make dequeueing depend on server-owned finalized transcript state (or otherwise wait until the previous assistant content is durably persisted before starting the next queued run) instead of starting the next run synchronously from `finish()`.

apps/daemon/src/server.ts
```diff
@@ -3617,6 +3618,20 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     if (typeof clientRequestId === 'string' && clientRequestId)
       run.clientRequestId = clientRequestId;
     if (typeof agentId === 'string' && agentId) run.agentId = agentId;
+
+    let finalMessage = message;
+    if (rebuildTranscript && conversationId) {
+      try {
+        const msgs = listMessages(db, conversationId);
```
mrcfps, line 4283: Blocking: rebuilding with `listMessages(db, conversationId)` pulls the entire conversation, not the history up to this queued run. The web streaming branch eagerly persists each queued user turn and its empty assistant placeholder before posting `/api/runs`, so if a user queues two follow-ups while the first run is still active, the first queued run will see the second queued user prompt (and placeholder) as future context. Please scope the rebuild to a stable cursor for this run, such as the queued turn/user message or `assistantMessageId`, and exclude later queued turns/placeholders; alternatively persist the exact transcript snapshot that should be used when this run starts.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1015,10 +1023,64 @@ export function ProjectView({
...
+            .map((m) => '## ' + m.role + '\n' + m.content.trim())
+            .join('\n\n');
+            
+          fetch('/api/runs', {
```
mrcfps, line 1057: Blocking: the daemon queued run is created with a bare `fetch('/api/runs')` while another run is streaming, but it is not wired into the Stop cancellation path until a later reattach. `handleStop` aborts the current controllers and marks active assistant messages canceled locally, yet this queued run remains in the daemon queue and will start as soon as the current run finishes/cancels. That leaves the UI showing the queued assistant as canceled while the daemon still executes it. Please track the queued run ID/controller as soon as this request succeeds and have Stop call `/api/runs/:id/cancel` for queued daemon runs, or avoid marking queued daemon messages canceled unless the daemon cancellation has been sent and accepted.

mrcfps (changes requested):
@Moe1211, thank you for continuing to work through the queued-message edge cases. I found a few remaining daemon sequencing issues that can still make queued follow-ups run with the wrong transcript or keep running after Stop, so I’m requesting changes for this pass.


apps/daemon/src/runs.ts
lefarcen: P1 — Cross-reload race: `finish()` calls `maybeStartNext()` immediately after emitting the terminal `end` event. The completed assistant message is persisted by the web client only *after* it consumes that terminal event and flushes its buffered text, while the queued starter rebuilds its prompt from SQLite at next run's start. That means the follow-up can start before the just-finished assistant output is fully saved. Please make dequeueing depend on server-owned finalized transcript state (or otherwise wait until the previous assistant content is durably persisted before starting the next queued run).

apps/daemon/src/server.ts
lefarcen, line 4283: P1 — Transcript rebuild includes future queued turns: `rebuildTranscript` calls `listMessages(db, conversationId)` for the entire conversation. The web streaming branch eagerly persists each queued user turn and its empty assistant placeholder before posting `/api/runs`, so if a user queues two follow-ups while the first run is active, the first queued run will see the second queued user prompt (and placeholder) as future context. Please scope the rebuild to a stable cursor for this run, such as the queued turn/user message or `assistantMessageId`, and exclude later queued turns/placeholders.

lefarcen (changes requested):
Hey @Moe1211, thanks for the continued iteration. I found 2 remaining P1 blockers in the daemon sequencing — both can make queued follow-ups run with incomplete transcript or stay alive after Stop was pressed, so I'm requesting changes for this pass.

mrcfps:
@Moe1211 I'm holding off on generating review comments for nexu-io/open-design#518 because this pull request has merge conflicts right now.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1019,10 +1028,71 @@ export function ProjectView({
...
```
mrcfps, line 1057: `saveMessage(userMsg)` and `saveMessage(assistantMsg)` are both fire-and-forget here, but `server.ts` only starts the next queued run when a later terminal assistant `PUT` calls `maybeStartNext()`. Because this `/api/runs` POST is sent without awaiting either write, a fast-finishing active run can dequeue the next run before the queued user turn or its assistant placeholder has reached SQLite. In that case `startChatRun(..., rebuildTranscript: true)` rebuilds from stale DB state and can start without the just-queued follow-up at all. Please either await persistence of the queued turn before creating the queued run, or move queued-turn creation/run enqueueing behind one server-side transaction/API so the message rows and queued run are committed atomically, and add a regression that queues just before the active run finishes.

mrcfps (commented):
@Moe1211, thanks for continuing to iterate on the queued-message flow. I found one remaining correctness risk in the daemon queue path: the queued run can still start before the queued turn is durably written, which can make the follow-up run against stale transcript state. I'm leaving this as a comment because this reviewer has already hit the request-changes cap for this PR.


lefarcen (changes requested):
Hey @Moe1211, thanks for the continued iteration! You've made solid progress on the 3 prior P1 blockers:

✅ **Transcript cutoff** (prior blocker #2) is fixed — the new `rebuildTranscript` logic at `server.ts:3629-3644` correctly cuts off at `assistantMessageId` to exclude future queued turns.
⚠️ **finish() race** (prior blocker #1) is partially fixed — `finish()` no longer calls `maybeStartNext()` directly, and the drain moved to `server.ts:1922` after message save. However, 2 new P1 sequencing issues remain (see inline below).
⚠️ **Stop cancellation** (prior blocker #3) is partially fixed — new code tracks `canceledMessageIdsRef` and calls `/api/runs/:id/cancel`. But the drain race at `server.ts:1922` can still start a queued run before the cancel lands.

Requesting changes for the 2 new P1 blockers below.

apps/web/src/i18n/types.ts
```diff
@@ -701,6 +701,7 @@ export interface Dict {
...
```
mrcfps, line 705: Adding `chat.queue` to `Dict` makes every locale implement that key, but this patch only updates the locales in the diff and leaves `apps/web/src/i18n/locales/id.ts` without `chat.queue`. That means the Indonesian locale no longer satisfies `Dict`, so the web package will fail type-checking once dependencies are installed (`Property 'chat.queue' is missing in type ...`). Please either add the new key to `id.ts` in this PR or make the type change non-breaking for untouched locales.

mrcfps (commented):
@Moe1211, thanks for the continued iteration on the queued-message flow. I found two concrete follow-ups in this head. I’m leaving this as a comment because this reviewer has already hit the request-changes cap for this PR.


apps/web/src/components/ProjectView.tsx
```diff
@@ -1478,11 +1569,21 @@ export function ProjectView({
     }
     reattachControllersRef.current.clear();
     setStreaming(false);
+    if(queuedMessage?.runId){
```
mrcfps, line 1575: `queuedMessage` is declared earlier in this file with only `{ prompt, attachments, commentAttachments, projectId, conversationId }`, but this new stop-path reads `queuedMessage?.runId` even though that property is never part of the state shape. In TypeScript this becomes `Property 'runId' does not exist on type ...`, so the web package cannot type-check once dependencies are installed. Please either add `runId` to the queued-message state and populate it, or remove this branch and cancel queued daemon runs only through the assistant messages that already carry `runId`.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1027,10 +1036,71 @@ export function ProjectView({
...
       const nextHistory = [...messages, userMsg];
+
+      if (streaming) {
+        if (config.mode === 'daemon') {
+          // Eagerly persist and queue via daemon to survive reloads
+          setMessages((curr) => [...curr, userMsg, assistantMsg]);
+          await persistMessage(userMsg);
+          await persistMessage(assistantMsg);
+
+          const transcript = nextHistory
```
lefarcen, line 1053: P1 — Queue can advance before completed assistant message is durably saved. `consumeDaemonRun` calls `onRunStatus(endStatus)` immediately on terminal SSE frame (line 1333), but `handlers.onDone()` (line 1351) flushes text buffer and final assistant state later. Buffered final text/artifacts can miss transcript used by next queued run. Server.ts:2290 drain fires after `upsertMessage` but before web flushes.

apps/daemon/src/server.ts
```diff
@@ -2288,6 +2288,12 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     });
     // Bump the parent project's updatedAt so the project list re-orders.
     updateProject(db, req.params.id, {});
```
lefarcen, line 2290: P1 — Drain race: `maybeStartNext()` can start queued run before cancel request lands. Server calls drain immediately after message save, but queued run start is async while cancel POST from ProjectView handleStop is inflight. Need sequencing that respects cancel before drain, or check canceled state in `maybeStartNext`.

lefarcen (changes requested):
Hey @Moe1211, thanks for the continued iteration! You've made solid progress on the 3 prior P1 blockers:

✅ **Transcript cutoff** (prior blocker #2) is fixed — the new `rebuildTranscript` logic at `server.ts:4211-4213` correctly cuts off at `assistantMessageId` to exclude future queued turns.
⚠️ **finish() race** (prior blocker #1) is partially fixed — `finish()` no longer calls `maybeStartNext()` directly, and the drain moved to `server.ts:2290` after message save. However, 2 new P1 sequencing issues remain (see inline below).
⚠️ **Stop cancellation** (prior blocker #3) is partially fixed — new code tracks `canceledMessageIdsRef` and calls `/api/runs/:id/cancel`. But the drain race at `server.ts:2290` can still start a queued run before the cancel lands.

Requesting changes for the 2 new P1 blockers below.

lefarcen (changes requested):
Hey @Moe1211, thanks for the continued iteration! You've made solid progress on the 3 prior P1 blockers:

✅ **Transcript cutoff** (prior blocker #2) is fixed — the new `rebuildTranscript` logic at `server.ts:4211-4213` correctly cuts off at `assistantMessageId` to exclude future queued turns.
⚠️ **finish() race** (prior blocker #1) is partially fixed — `finish()` no longer calls `maybeStartNext()` directly, and the drain moved to `server.ts:2290` after message save. However, 2 new P1 sequencing issues remain (see below).
⚠️ **Stop cancellation** (prior blocker #3) is partially fixed — new code tracks `canceledMessageIdsRef` and calls `/api/runs/:id/cancel`. But the drain race at `server.ts:2290` can still start a queued run before the cancel lands.

## 2 NEW P1 blockers (must fix):

**1. apps/web/src/providers/daemon.ts:310** — Queue can advance before completed assistant message is durably saved. `consumeDaemonRun` calls `onRunStatus(endStatus)` immediately on terminal SSE frame (line 310), but `handlers.onDone(acc)` (line 338) flushes text buffer and final assistant state AFTER. The PUT `/api/projects/:id/conversations/:cid/messages/:mid` that triggers the drain at `server.ts:2290` happens before the final flush completes, so buffered final text/artifacts can miss the transcript used by the next queued run.

Fix: move `onRunStatus(endStatus)` call to AFTER `handlers.onDone(acc)` completes (line 338+), or make the drain at `server.ts:2290` wait for a flush-complete signal from the web client instead of assuming the message is final when the PUT arrives.

**2. apps/daemon/src/server.ts:2290** — Drain race: `maybeStartNext()` can start queued run before cancel request lands. Server calls drain immediately after message save (this line), but queued run start is async while cancel POST from `ProjectView.handleStop` is inflight. When user clicks Stop on an active run, web sends (1) PUT to mark assistant message as terminal (triggering this drain), then (2) POST `/api/runs/:id/cancel` for the queued run — but (1) completes synchronously and starts the queued run before (2) arrives.

Fix: defer `maybeStartNext()` until after a small async yield (e.g., `setImmediate(() => maybeStartNext(...))` or similar), or have `maybeStartNext()` check a "pending cancel" flag on the queued run before starting it.

apps/daemon/src/server.ts
```diff
@@ -2288,6 +2288,16 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+
+    // If an assistant message was just marked terminal, the UI has finished flushing it.
+    if (saved.role === 'assistant' && design.runs.isTerminal(saved.runStatus)) {
+      // Skip drain if active canceled runs inflight (race protection)
+      const activeRuns = design.runs.listActiveChatRuns(req.params.cid);
```
mrcfps, line 2295: This new guard calls `design.runs.listActiveChatRuns(req.params.cid)`, but `createChatRunService()` only returns `create`, `start`, `enqueue`, `maybeStartNext`, `get`, `list`, `stream`, `cancel`, `wait`, `emit`, `finish`, `fail`, `statusBody`, and `isTerminal` in `apps/daemon/src/runs.ts`. Because `listActiveChatRuns` is undefined, the first terminal assistant-message PUT will throw `TypeError: design.runs.listActiveChatRuns is not a function` before `maybeStartNext()` can run. That breaks the main queued-message drain path and can also turn a normal message save into a 500 once a run finishes. Please switch this check to an existing API (for example `design.runs.list({ conversationId: req.params.cid, status: "active" })`) or add/export a real `listActiveChatRuns()` helper from the run service before using it here.

mrcfps (commented):
@Moe1211, thanks for continuing to tighten up the queued-message flow. I found one remaining correctness issue in the new daemon drain path that looks worth fixing before merge.


lefarcen (changes requested):
Hey @Moe1211, thanks for the continued iteration! You've made solid progress — **transcript cutoff is now fixed** ✅ and the drain moved to after message save ✅.

However, **1 P1 blocker + 2 P2 TypeScript issues remain**:

**P1 — Queue can advance before final text flush (apps/web/src/providers/daemon.ts:305):** `consumeDaemonRun` calls `onRunStatus(endStatus)` immediately when the `end` SSE event arrives (line 305), but `handlers.onDone(acc)` — which flushes the final buffered text and artifacts — runs later at line 338. The daemon's `maybeStartNext()` at `server.ts:2290` fires after the HTTP round-trip for message save completes, but the web client signals "done" BEFORE flushing its buffer. If the next queued run starts while that buffer is still in-flight, it will rebuild transcript from SQLite and miss the final text/artifacts. Please move `onRunStatus(endStatus)` to AFTER `handlers.onDone(acc)` (line 339), so the queue only advances after the final content is durably flushed.

**P2 — TypeScript: `queuedMessage?.runId` property doesn't exist (apps/web/src/components/ProjectView.tsx:1572):** The `queuedMessage` state is declared with only `{ prompt, attachments, commentAttachments, projectId, conversationId }`, but line 1572 reads `queuedMessage?.runId`. This will fail TypeScript once dependencies are installed. Either add `runId` to the state shape and populate it when queueing, or remove this branch and cancel queued runs through the assistant messages that already carry `runId`.

**P2 — TypeScript: Indonesian locale missing `chat.queue` (apps/web/src/i18n/locales/id.ts):** All other locales now have `chat.queue`, but `id.ts` is still missing it. This will fail TypeScript (`Property 'chat.queue' is missing in type ...`). Please add the key to `id.ts` in this PR or make the Dict change non-breaking for untouched locales.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1179,6 +1251,7 @@ export function ProjectView({
         },
         onDone: () => {
           textBuffer.flush();
+          await persistMessageById(assistantId);
```
mrcfps, line 1254: `onDone` is still declared as a synchronous callback here, so adding `await` inside it makes `ProjectView.tsx` fail to parse/build before this flow can run at all. The changed lines show `onDone: () => { ... await persistMessageById(assistantId); ... }`, and `persistMessageById()` itself returns `void`, so even converting this callback to `async` would not actually wait for the message save you want before the queue advances. Please either make `persistMessageById` return a real promise and await it end-to-end from a promise-aware `onDone`, or keep this callback synchronous and move the drain-ordering fix to a server-owned path that does not rely on awaiting a fire-and-forget helper.

mrcfps (commented):
@Moe1211, thanks for continuing to iterate on the queued-message flow. I found one additional concrete build blocker in this head, so I’m leaving a comment for this pass.


apps/daemon/src/server.ts
```diff
@@ -2288,6 +2288,16 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     });
     // Bump the parent project's updatedAt so the project list re-orders.
     updateProject(db, req.params.id, {});
+
+    // If an assistant message was just marked terminal, the UI has finished flushing it.
+    if (saved.role === 'assistant' && design.runs.isTerminal(saved.runStatus)) {
```
lefarcen, line 2293: **P1 blocker**: This line calls `design.runs.listActiveChatRuns(req.params.cid)`, but `createChatRunService()` in `apps/daemon/src/runs.ts` does NOT export `listActiveChatRuns`. The service return object (lines 190–207) only includes `create`, `start`, `enqueue`, `maybeStartNext`, `get`, `list`, `stream`, `cancel`, `wait`, `emit`, `finish`, `fail`, `statusBody`, `isTerminal`.

This means the first terminal assistant message PUT will throw `TypeError: design.runs.listActiveChatRuns is not a function`, breaking the main queued-message drain path and potentially turning normal message saves into 500 errors once a run finishes.

Please either:
- Switch this check to an existing API, e.g. `design.runs.list({ conversationId: req.params.cid, status: 'active' })`
- Or define and export a real `listActiveChatRuns()` helper from the run service before using it here.

lefarcen (changes requested):
Hey @Moe1211, thanks for the continued iteration! You've successfully fixed all 3 issues from the last review:

✅ **P1 fixed**: `apps/web/src/providers/daemon.ts:309` now comments out the immediate `onRunStatus(endStatus)` call, and line 340 calls it AFTER `handlers.onDone(acc)` completes. The queue now only advances after the final text buffer is flushed.
✅ **P2 fixed**: `apps/web/src/components/ProjectView.tsx:286` — the `QueuedMessage` interface now includes `runId?: string`, so TypeScript will pass.
✅ **P2 fixed**: `apps/web/src/i18n/locales/id.ts:631` — Indonesian locale now includes `'chat.queue': 'Antri'`.

However, I found **1 new P1 blocker** in the daemon drain path (see inline below).

mrcfps:
@Moe1211 Thanks again for the contribution here 💛

PR #520, opened 2026-05-05
fix: add lowercase 'todowrite' compatibility for OpenCode CLI

1119302165:
---

# 📋 **Pull Request**

## **fix: add lowercase 'todowrite' compatibility for OpenCode CLI**

before fix:
 
Chat Comments

after fix:
Chat Comments
 ---

---

## Problem

TodoWrite tool outputs from the OpenCode CLI render as raw JSON text instead of the pretty todo progress card component.

**Root cause:** OpenCode emits tool names in lowercase (`todowrite`) while the UI matching logic only looked for capitalized (`TodoWrite`).

opencode return

```
todowrite {"todos":[{"content":"Read active DESIGN.md...
```

---

## Solution

Add case-insensitive tool name matching in `ToolCard.tsx`:

```diff
- if (name === 'TodoWrite') return <TodoCard input={use.input} />;
+ if (name === 'TodoWrite' || name === 'todowrite') return <TodoCard input={use.input} />;
```

---

## Scope

| Metric | Value |
|--------|-------|
| Files modified | **1** |
| Lines changed | `+1/-1` |
| Daemon changes | **None** |
| Invasiveness | **Minimal** |

---

## Validation

- ✅ All 211 frontend tests passing
- ✅ Backward compatible with existing database entries
- ✅ Zero impact on official agent behavior
- ✅ Zero side effects on other tools

---

> This fixes the rendering issue for historic chat messages created with OpenCode CLI while maintaining full compatibility with the standard TodoWrite tool name convention.

### DIFF
--- a/apps/daemon/src/json-event-stream.ts
+++ b/apps/daemon/src/json-event-stream.ts
@@ -33,37 +33,6 @@ function formatOpenCodeUsage(tokens) {
   return Object.keys(usage).length > 0 ? usage : null;
 }
 
-// TEMPORARY ADAPTER LAYER
-// TODO: Standardize tool name contract across all CLIs.
-// Right now every agent CLI uses different tool names:
-//   - opencode: lowercase 'todowrite', 'write', 'edit'
-//   - claude: capitalized 'TodoWrite', 'Write', 'Edit'
-//   - codex: underscored 'create_file', 'str_replace_edit'
-// This map normalizes them to the UI's standard names.
-const OPENCODE_TOOL_NORMALIZATION: Record<string, string> = {
-  todowrite: 'TodoWrite',
-  searchreplace: 'SearchReplace',
-  write: 'Write',
-  create_file: 'Write',
-  edit: 'Edit',
-  str_replace_edit: 'Edit',
-  read: 'Read',
-  read_file: 'Read',
-  bash: 'Bash',
-  glob: 'Glob',
-  list_files: 'Glob',
-  grep: 'Grep',
-  webfetch: 'WebFetch',
-  web_fetch: 'WebFetch',
-  websearch: 'WebSearch',
-  web_search: 'WebSearch',
-};
-
-function normalizeToolName(name: string): string {
-  const lower = name.toLowerCase().replace(/_/g, '');
-  return OPENCODE_TOOL_NORMALIZATION[lower] ?? OPENCODE_TOOL_NORMALIZATION[name] ?? name;
-}
-
 function detectEmbeddedToolCall(text: string): { name: string; input: unknown } | null {
   const trimmed = text.trim();
   const todoMatch = trimmed.match(/^(todowrite|TodoWrite)\s*({[\s\S]*})/i);
@@ -94,7 +63,7 @@ function handleOpenCodeEvent(obj, onEvent, state) {
         onEvent({
           type: 'tool_use',
           id: key,
-          name: normalizeToolName(detected.name),
+          name: detected.name,
           input: detected.input,
         });
       }
@@ -112,7 +81,7 @@ function handleOpenCodeEvent(obj, onEvent, state) {
       onEvent({
         type: 'tool_use',
         id: part.callID,
-        name: normalizeToolName(part.tool),
+        name: part.tool,
         input: safeParseJson(statePart?.input) ?? statePart?.input ?? null,
       });
     }
--- a/apps/daemon/src/json-event-stream.ts
+++ b/apps/daemon/src/json-event-stream.ts
@@ -33,6 +33,49 @@ function formatOpenCodeUsage(tokens) {
   return Object.keys(usage).length > 0 ? usage : null;
 }
 
+// TEMPORARY ADAPTER LAYER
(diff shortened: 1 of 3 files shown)


apps/daemon/src/json-event-stream.ts
```diff
@@ -43,6 +55,20 @@ function handleOpenCodeEvent(obj, onEvent, state) {
   }
 
   if (obj.type === 'text' && typeof part.text === 'string' && part.text.length > 0) {
+    const detected = detectEmbeddedToolCall(part.text);
+    if (detected) {
+      const key = `${obj.sessionID || 'session'}:embedded-${detected.name}-${Date.now()}`;
+      if (!state.openCodeToolUses.has(key)) {
+        state.openCodeToolUses.add(key);
```
mrcfps, line 62: ⚠️ This embedded-tool id can collide because it only varies by session, tool name, and `Date.now()`. `feed()` processes multiple NDJSON lines synchronously, so two `todowrite {...}` text events for the same session can land in the same millisecond, produce the same `key`, and the second one is silently skipped by `state.openCodeToolUses.has(key)`. That would drop a real TodoWrite update—the exact primary flow this PR is trying to recover. Please use a parser-local monotonic counter (for example, add an `openCodeEmbeddedToolSeq` field to `state`) or another guaranteed-unique per-event value instead of `Date.now()`, and add a fixture that feeds two embedded TodoWrite lines in one chunk and asserts both tool_use events are emitted with distinct ids.

apps/daemon/src/json-event-stream.ts
```diff
@@ -33,6 +33,18 @@ function formatOpenCodeUsage(tokens) {
   return Object.keys(usage).length > 0 ? usage : null;
 }
 
+function detectEmbeddedToolCall(text: string): { name: string; input: unknown } | null {
+  const trimmed = text.trim();
+  const todoMatch = trimmed.match(/^(todowrite|TodoWrite)\s*({[\s\S]*})/i);
```
mrcfps, line 38: This matcher accepts a JSON object prefix even when the text event has trailing prose, because it is not anchored to the end of the trimmed string. Since the caller returns immediately after emitting the tool_use, a text event such as `todowrite {"todos":[]} done` would render the card but silently drop `done`. To avoid losing assistant text, either require the tool-call payload to occupy the whole event except whitespace (for example, add `\s*$` to the pattern) or split and emit any trailing text as a `text_delta`.

mrcfps (changes requested):
I found one blocking parser issue in the embedded TodoWrite path: duplicate IDs can make back-to-back todo updates disappear. I also left a smaller parser-hardening note for trailing text handling.


lefarcen (changes requested):
Hi @1119302165! Thanks for this fix — the daemon-side detection is smart and the backward compatibility layer is well-explained.

However, I found edge cases where the normalization is incomplete, causing tool events to still slip through lowercase or get dropped when processed in quick succession:

## P2 findings (must-fix)

**1. ID collision race** (apps/daemon/src/json-event-stream.ts:60)
The synthetic id uses `Date.now()`, so two embedded `todowrite` text events processed in the same millisecond get the same key — the second event is silently dropped by `openCodeToolUses.has(key)` while its text is also suppressed.

**Fix**: Use a deterministic per-handler sequence counter in `state`, e.g. `state.embeddedToolUseSeq = (state.embeddedToolUseSeq || 0) + 1`, and generate ids from `sessionID + seq`. Add a test feeding two embedded todo lines in one `feed()` call.

**2. Structured tool_use normalization missing** (apps/daemon/src/json-event-stream.ts:76-85)
The normal OpenCode `tool_use` branch still emits `part.tool` unchanged. If OpenCode emits a proper structured `tool_use` with `part.tool: 'todowrite'` (lowercase), daemon-side normalization does NOT happen — only the new text heuristic applies.

**Fix**: Normalize tool names in the existing `tool_use` branch too, e.g. map `todowrite` to `TodoWrite` before emitting, and add a test for a lowercase structured `tool_use` event.

**3. Incomplete UI coverage** (apps/web/src/components/ToolCard.tsx:59)
Compatibility is only local to this card renderer. Historic lowercase `todowrite` events still fail other todo-aware code paths:
- `apps/web/src/runtime/todos.ts:38` won't find unfinished todos for the 'continue remaining tasks' flow
- `apps/web/src/components/AssistantMessage.tsx:618` groups lowercase todo calls as a generic tool family (wrong grouping)

**Fix**: Introduce a shared `isTodoWriteToolName(name)` helper and use it in `ToolCard`, `latestTodosFromEvents`, and `toolFamily`. Or normalize names when reading persisted events.

## P3 (nice-to-have)

**4. False positive risk** (apps/daemon/src/json-event-stream.ts:38)
The detector treats any text beginning with `todowrite {valid JSON object}` as a tool call, even if the JSON isn't a TodoWrite payload. That can hide legitimate assistant prose explaining examples.

**Fix**: Validate parsed shape before returning — at minimum require an object with a `todos` array — and fall back to `text_delta` when the shape doesn't match.

---

Fix those and this will be production-ready. Looking forward to v2! 🎉

PR #524, opened 2026-05-05
feat: Critique Theater Phase 5 (panel prompt template + system composer wiring)

Nagendhra-web:
Stacks on top of Phase 4 (#481, merged in 76e6c7a). Phase 5 adds the panel prompt template that drives the five-panelist debate and wires it into the system composer so a critique-enabled run actually receives the right instructions.

## What's in this PR

**Task 5.1** — `apps/daemon/src/prompts/panel.ts` defines the panel prompt module. It renders the per-role brief (DESIGNER drafts only; CRITIC, BRAND, A11Y, COPY score), threads `cfg.protocolVersion`, `cfg.maxRounds`, `cfg.scoreThreshold`, `cfg.scoreScale`, and `cfg.weights` from the active CritiqueConfig, and emits the `<CRITIQUE_RUN version="1" ...>` envelope opener that the v1 parser expects. Brand fields and the DESIGN.md body are escaped before interpolation so a hostile DESIGN.md cannot close `<BRAND_SOURCE>` and inject protocol instructions.

**Task 5.2** — `apps/daemon/src/prompts/system.ts` composer branches on `cfg.enabled`. When critique is on AND the surface is non-media AND brand and skill data are present, the panel addendum is appended. When any of those is false, the legacy single-pass prompt is unchanged so M0 dark-launch behavior stays intact for media surfaces, missing context, and disabled runs. `apps/daemon/src/server.ts` threads `critiqueCfg`, `critiqueBrand`, `critiqueSkill` into the composer and computes a `critiqueShouldRun` flag the spawn path uses to gate `runOrchestrator(...)` on the same eligibility logic, so prompt and orchestrator stay in lockstep.

## What's NOT in this PR

`cast` is a v1 non-goal per the spec, so the panel renders all five fixed roles regardless of `cfg.cast`. Configurable cast lands in v2.

Phases 6-15 are queued as stacked branches. Each was implemented with the lessons from #481's 5 review rounds baked in.

## Test plan

- [x] `pnpm --filter @open-design/contracts typecheck` — exit 0
- [x] `pnpm --filter @open-design/daemon typecheck` — exit 0
- [x] `pnpm --filter @open-design/web typecheck` — exit 0
- [x] `pnpm --filter @open-design/daemon exec vitest run tests/critique-panel-prompt.test.ts tests/critique-composer.test.ts` — 32/32 pass (21 panel-prompt + 11 composer)
- [x] No em dashes anywhere in source, comments, or commit messages
- [x] No `Co-authored-by` trailers; author = Nagendhra <nagendhra405@gmail.com>
- [x] New tests live under `apps/daemon/tests/` per the directory convention from #496


apps/daemon/src/prompts/system.ts
```diff
@@ -188,6 +203,15 @@ export function composeSystemPrompt({
     parts.push(MEDIA_GENERATION_CONTRACT);
   }
 
+  // Critique Theater addendum. When cfg.enabled is true the panel protocol
+  // is pinned last so it overrides any softer critique wording earlier in the
+  // stack. When disabled (the default) this block is a no-op so no consumer
+  // needs to opt in.
+  const cfg = critique ?? defaultCritiqueConfig();
+  if (cfg.enabled && critiqueBrand && critiqueSkill) {
```
mrcfps: This makes the critique prompt depend on callers passing both `critiqueBrand` and `critiqueSkill`, but the actual daemon prompt builder still calls `composeSystemPrompt({ skillBody, skillName, skillMode, designSystemBody, designSystemTitle, craftBody, craftSections, metadata, template })` without any of the new critique fields. When `OD_CRITIQUE_ENABLED=true`, `startChatRun` still routes plain-stream adapters into `runOrchestrator`, but the child process receives the legacy single-pass prompt with no `<CRITIQUE_RUN>` instructions, so the parser/orchestrator waits for tags the model was never told to emit and the critique-enabled primary flow fails instead of running the panel. Please thread the loaded `critiqueCfg` plus the active design-system/skill data through the daemon prompt builder whenever critique is enabled (or fail fast if required context is unavailable), and update the composer/server-path tests so the enabled daemon path proves the spawned prompt contains the panel addendum.

mrcfps (changes requested):
Thanks for the focused Phase 5 slice. I reviewed the new panel prompt, composer wiring, and the daemon tests; the prompt template itself is nicely scoped, but I found one main-path wiring issue that means enabling Critique Theater still does not actually deliver the panel protocol to the spawned agent.


lefarcen (changes requested):
Hey @Nagendhra-web! Thanks for the structured PR — the test coverage and stacked-phase discipline are solid.

Found 1 critical wiring gap + 4 design/correctness issues that need fixing before this can ship:

---

**P1 — Production wiring missing**

`apps/daemon/src/server.ts:3258` — The new composer branch is not wired into the production prompt path. `composeDaemonSystemPrompt()` calls `composeSystemPrompt({ skillBody, skillName, skillMode, designSystemBody, designSystemTitle, craftBody, craftSections, metadata, template })` without `critique`, `critiqueBrand`, or `critiqueSkill`, so even when `critiqueCfg.enabled` later routes stdout through `runOrchestrator()` at lines 3726-3777, the spawned CLI was never instructed to emit `<CRITIQUE_RUN>`. This makes critique-enabled runs likely to parse legacy prose/artifacts instead of the required wire protocol.

---

**P2 — cfg.weights not threaded into prompt**

`apps/daemon/src/prompts/panel.ts:111` — The prompt does not thread `cfg.weights` at all, even though the daemon recomputes composite using those weights (`scoreboard.ts` uses `cfg.weights`) and the PR body claims weights are rendered. Without the weighted formula in the panel instructions, the model has no way to produce a matching `composite`, increasing `composite_mismatch` warnings and weakening the convergence guidance.

---

**P2 — Template injection boundary**

`apps/daemon/src/prompts/panel.ts:102` — `brand.name`, `skill.id`, and `brand.design_md` are interpolated into tag attributes/body without escaping or sentinel handling. A DESIGN.md containing `</BRAND_SOURCE>` or XML-like protocol tags can break out of the intended data wrapper and inject system-level protocol instructions, despite the text saying it is data only.

---

**P2 — Role guidance contradicts v1 spec**

`apps/daemon/src/prompts/panel.ts:77` — The role guidance conflicts with the v1 design: this prompt says the DESIGNER scores creative intent/layout, but `specs/current/critique-theater.md:109` says Designer is omitted from the composite because Designer drafts and does not score, and the wire example in this same template omits a designer `score`. This contradiction makes the expected output ambiguous for the model.

---

**P2 — PR body overclaims about cfg.cast**

PR body claims the template threads `cast` and covers missing-role pruning, but `panel.ts` always instructs all five panelists and never reads `cfg.cast`; the new tests also do not exercise a custom cast or missing-role pruning. This conflicts with both the implementation evidence and the v1 spec's explicit non-goal of configurable cast.

---

The good news: the composer branch + tests are architecturally clean, so once the wiring is connected and the prompt aligns with the v1 spec, this should merge smoothly. Let me know if any of these findings need clarification!

Nagendhra-web:
Thanks @mrcfps and @chatgpt-codex-connector. Both round-1 blockers addressed in 56b3831.

The composer now skips the panel addendum when the active surface is image/video/audio (either via `skillMode` or project `metadata.kind`). The panel template requires `<ARTIFACT mime="text/html">` inside `<CRITIQUE_RUN>` and that conflicts head-on with `MEDIA_GENERATION_CONTRACT`, so a critique-enabled media run was getting two contradictory instruction sets. The skip is a no-op for deck/artifact/template surfaces, so non-media projects still get the panel.

The server's prompt builder now threads `critiqueCfg`, `critiqueBrand` (`{name, design_md}` from the active design system), and `critiqueSkill` (`{id}` from the effective skill) into `composeSystemPrompt(...)`. Without this the spawned child received the legacy single-pass prompt and the orchestrator waited for `<CRITIQUE_RUN>` tags the model was never told to emit. The composer ignores these fields when `cfg.enabled` is false, so the legacy path stays exactly as before.

Tests: 5 new composer regressions for the media-surface skip (skillMode=image/video/audio, metadata.kind=image, and the deck control case). 27 prompt-suite tests pass: 16 panel-prompt + 11 composer. Daemon typecheck clean.

lefarcen (changes requested):
Hey @Nagendhra-web! Thanks for the fix attempt — I can see the server wiring is now threaded through to , which resolves half of the original P1.

However, there's a new P1 that makes critique-enabled runs still fail:

---

**P1 — Orchestrator routing mismatch**

`apps/daemon/src/server.ts:3743` — The media-surface fix only skips the panel prompt in `composeSystemPrompt` (`system.ts:217`), but the runtime still routes **every plain-stream run** through `runOrchestrator()` whenever `critiqueCfg.enabled` is true. For image/video/audio runs, and also for runs where `critiqueBrand` or `critiqueSkill` is undefined, the child receives no `<CRITIQUE_RUN>` instructions but stdout is still parsed as Critique Theater protocol, causing the same failure mode as the original blocker. The orchestrator gate needs to use the same eligibility condition as prompt composition, or these cases need to fall through to legacy generation.

---

The 4 P2 findings from the previous review remain unchanged:

**P2 — cfg.weights not threaded**
`panel.ts:111` — `cfg.weights` is still not rendered anywhere in the panel prompt. The daemon recomputes composite with `cfg.weights`, but the model is only told `threshold` and `scale`, so it has no prompt-level evidence for how to calculate `composite` consistently with the daemon.

**P2 — Template injection boundary**
`panel.ts:102` — The template injection boundary is still unprotected. `brand.name`, `skill.id`, and especially `brand.design_md` are interpolated directly into protocol-like tags, so a DESIGN.md containing `</BRAND_SOURCE>` or Critique Theater tags can escape the intended data wrapper and inject higher-priority protocol instructions.

**P2 — Role guidance contradicts v1 spec**
`panel.ts:77` — The DESIGNER role guidance still contradicts the v1 spec. The prompt says Designer scores creative intent/composition/layout, while `specs/current/critique-theater.md` says Designer drafts and is omitted from the composite because Designer does not score; the wire example in this template also omits a designer `score`, so the model receives conflicting instructions.

**P2 — PR body overclaims**
PR body still claims the prompt threads configured `cast` and `weights`, and that tests cover missing-role pruning and weight rendering. Current `panel.ts` never reads `cfg.cast` or `cfg.weights`, and the v1 spec explicitly says configurable cast is a non-goal, so the reasoning and claimed coverage are not supported by the code or design document.

---

The good news: you're making progress on the server threading! Once the orchestrator routing aligns with the prompt composition eligibility logic, and the prompt polish cleans up the 4 P2s, this will be solid. Let me know if any of these need clarification!

lefarcen (changes requested):
Hey @Nagendhra-web! Thanks for the fix attempt — I can see the server wiring is now threaded through to `composeSystemPrompt`, which resolves half of the original P1.

However, there's a new P1 that makes critique-enabled runs still fail:

---

**P1 — Orchestrator routing mismatch**

`apps/daemon/src/server.ts:3743` — The media-surface fix only skips the panel prompt in `composeSystemPrompt` (`system.ts:217`), but the runtime still routes **every plain-stream run** through `runOrchestrator()` whenever `critiqueCfg.enabled` is true. For image/video/audio runs, and also for runs where `critiqueBrand` or `critiqueSkill` is undefined, the child receives no `<CRITIQUE_RUN>` instructions but stdout is still parsed as Critique Theater protocol, causing the same failure mode as the original blocker. The orchestrator gate needs to use the same eligibility condition as prompt composition, or these cases need to fall through to legacy generation.

---

The 4 P2 findings from the previous review remain unchanged:

**P2 — cfg.weights not threaded**
`panel.ts:111` — `cfg.weights` is still not rendered anywhere in the panel prompt. The daemon recomputes composite with `cfg.weights`, but the model is only told `threshold` and `scale`, so it has no prompt-level evidence for how to calculate `composite` consistently with the daemon.

**P2 — Template injection boundary**
`panel.ts:102` — The template injection boundary is still unprotected. `brand.name`, `skill.id`, and especially `brand.design_md` are interpolated directly into protocol-like tags, so a DESIGN.md containing `</BRAND_SOURCE>` or Critique Theater tags can escape the intended data wrapper and inject higher-priority protocol instructions.

**P2 — Role guidance contradicts v1 spec**
`panel.ts:77` — The DESIGNER role guidance still contradicts the v1 spec. The prompt says Designer scores creative intent/composition/layout, while `specs/current/critique-theater.md` says Designer drafts and is omitted from the composite because Designer does not score; the wire example in this template also omits a designer `score`, so the model receives conflicting instructions.

**P2 — PR body overclaims**
PR body still claims the prompt threads configured `cast` and `weights`, and that tests cover missing-role pruning and weight rendering. Current `panel.ts` never reads `cfg.cast` or `cfg.weights`, and the v1 spec explicitly says configurable cast is a non-goal, so the reasoning and claimed coverage are not supported by the code or design document.

---

The good news: you're making progress on the server threading! Once the orchestrator routing aligns with the prompt composition eligibility logic, and the prompt polish cleans up the 4 P2s, this will be solid. Let me know if any of these need clarification!

apps/daemon/src/prompts/system.ts
```diff
@@ -188,6 +203,21 @@ export function composeSystemPrompt({
...
+  // stack. When disabled (the default) this block is a no-op so no consumer
+  // needs to opt in.
+  //
+  // The panel block requires <ARTIFACT mime="text/html"> inside <CRITIQUE_RUN>,
+  // which conflicts with MEDIA_GENERATION_CONTRACT (image/video/audio surfaces
+  // explicitly forbid HTML output). Skip the addendum on media surfaces so
+  // the critique flag is a no-op there until a media-aware panel template
+  // lands.
+  const cfg = critique ?? defaultCritiqueConfig();
+  if (cfg.enabled && critiqueBrand && critiqueSkill && !isMediaSurface) {
```
mrcfps, line 232: This guard now suppresses `<CRITIQUE_RUN>` for media surfaces, and it also still suppresses the panel whenever `critiqueBrand` or `critiqueSkill` is unavailable. The runtime path does not know that happened: `startChatRun` still sends every plain-stream adapter through `runOrchestrator()` whenever `critiqueCfg.enabled` is true, so these skipped-prompt runs execute with the legacy/media prompt while the parser waits for Critique Theater tags. That makes critique-enabled media projects, or projects missing the required brand/skill context, fail the primary run instead of cleanly falling back or failing before spawn. Please make the routing decision use the same panel-eligibility result as the composer, or fail fast before spawning when critique is enabled but the required context is missing, and add a server-path regression test that proves a media project with `OD_CRITIQUE_ENABLED=true` is not routed into the panel parser without a `<CRITIQUE_RUN>` prompt.

mrcfps (changes requested):
Thanks for the quick follow-up on the first wiring blocker. I found one remaining blocker in the critique routing path: the prompt composer can now intentionally omit the panel protocol, but the daemon still routes those same runs into the critique orchestrator.


lefarcen (changes requested):
Hey @Nagendhra-web! Re-reviewed the new head `56b3831` — the original server-threading gap is partially fixed (server now passes `critique`/`critiqueBrand`/`critiqueSkill` into composer ✅), but found 1 new P1 + 4 unchanged P2s that need addressing:

---

**P1 — Orchestrator routing mismatch**

`apps/daemon/src/server.ts:3743` — The media-surface fix only skips the panel prompt in `composeSystemPrompt` (`system.ts:217`), but the runtime still routes every plain-stream run through `runOrchestrator()` whenever `critiqueCfg.enabled` is true. For image/video/audio runs (and also runs where `critiqueBrand` or `critiqueSkill` is undefined), the child receives no `<CRITIQUE_RUN>` instructions but stdout is still parsed as Critique Theater protocol, causing the same failure mode as the original blocker.

**Fix**: The orchestrator gate needs to use the same eligibility condition as prompt composition (check `critiqueBrand`/`critiqueSkill`/`skillMode` before routing to `runOrchestrator`), or these cases need to fall through to legacy generation.

---

**P2 — cfg.weights still not threaded**

`apps/daemon/src/prompts/panel.ts:111` — `cfg.weights` is still not rendered anywhere in the panel prompt. The daemon recomputes composite with `cfg.weights`, but the model is only told `threshold` and `scale`, so it has no prompt-level evidence for how to calculate `composite` consistently with the daemon.

**P2 — Template injection boundary**

`apps/daemon/src/prompts/panel.ts:102` — `brand.name`, `skill.id`, and especially `brand.design_md` are interpolated directly into protocol-like tags. A DESIGN.md containing `</BRAND_SOURCE>` or Critique Theater tags can escape the intended data wrapper and inject higher-priority protocol instructions.

**P2 — Role guidance contradicts v1 spec**

`apps/daemon/src/prompts/panel.ts:77` — The DESIGNER role guidance says Designer scores creative intent/composition/layout, while `specs/current/critique-theater.md` says Designer drafts and is omitted from composite because Designer does not score. The wire example in this template also omits a designer `score`, so the model receives conflicting instructions.

**P2 — PR body overclaims cfg.cast/weights support**

PR body (line 5) claims the prompt threads configured `cast` and `weights`, and that tests cover missing-role pruning and weight rendering. Current `panel.ts` never reads `cfg.cast` or `cfg.weights`, and the v1 spec explicitly says configurable cast is a non-goal, so the reasoning and claimed coverage are not supported by the code or design document.

---

Please address the P1 orchestrator routing mismatch and consider fixing the 4 P2s. Thanks!

lefarcen (changes requested):
Hey @Nagendhra-web! Good progress on the orchestrator routing — the media/missing-context mismatch is now fixed, and the 4 P2s (weights, injection, role guidance, PR body) are addressed. Found 1 remaining P1:

---

**P1 — Adapter eligibility not in gate**

`apps/daemon/src/server.ts:3770` — `critiqueShouldRun` now matches the composer's media/brand/skill eligibility ✅, but it doesn't include adapter support. For a critique-enabled non-media run with brand+skill on a **non-plain adapter** (e.g., stream-json/claude-format), `composeSystemPrompt` still appends `<CRITIQUE_RUN>` instructions, then the runtime logs and falls through to legacy generation because `adapterStreamFormat !== 'plain'`.

That leaves the model instructed to emit Critique Theater protocol that no orchestrator consumes. **Fix**: Include adapter eligibility (`adapterStreamFormat === 'plain'`) in the same pre-compose gate, or suppress the panel addendum for unsupported adapters in the composer.

---

Once this is addressed, the full critique-enabled path should be solid. Thanks for the iterative fixes!

lefarcen (approved):
Thanks @Nagendhra-web! 🎉

The adapter-eligibility blocker is resolved. `composeDaemonSystemPrompt` now accepts `streamFormat` and folds `isPlainAdapter` into `critiqueShouldRun` alongside the existing brand/skill/media-surface checks (server.ts:3292-3297). Non-plain adapters with critique enabled now skip both the panel prompt addendum and orchestrator routing, falling cleanly through to legacy generation.

lefarcen (dismissed):
Thanks for the follow-up, @Nagendhra-web. The adapter-eligibility gap is now closed.

The prompt composer and spawn path are correctly lockstepped: `isPlainAdapter` is folded into `critiqueShouldRun` (server.ts:3292-3297), the panel prompt threading is gated on the same flag (server.ts:3314-3316), and the orchestrator gate at server.ts:3785 ensures non-plain adapters skip both the panel addendum and orchestrator routing.

Critique Theater Phase 5 routing is production-ready from my side. CI still pending; once green and mrcfps re-reviews the current head, this should be good to ship.

apps/daemon/src/prompts/panel.ts
```diff
@@ -0,0 +1,222 @@
...
+  return `# Critique Theater (active skill: ${safeSkillId})
+
+You are running in CRITIQUE THEATER mode. Speak as a five-panelist design jury
+inside one CLI session. Use the wire protocol below verbatim. Emit ONLY tagged
+regions; don't emit prose outside tags.
+
+## Panelist role definitions
+
+Each panelist has a fixed scope. They score only what is listed under their role
+and must declare at least one MUST_FIX in every non-final round. At least two
```
mrcfps: This paragraph still says every panelist must declare a `MUST_FIX` in every non-final round, but the changed Designer role below now correctly says Designer only drafts and does not score. The v1 parser will still emit and count any `<MUST_FIX>` inside a designer block as a `panelist_must_fix`, so a model that follows this general sentence literally can inflate the daemon's must-fix count and make convergence noisier than intended. Please narrow this sentence to the four scoring panelists, for example: `Each scoring panelist (CRITIC, BRAND, A11Y, COPY) scores only what is listed under their role and must declare at least one MUST_FIX in every non-final round.` 🙂

mrcfps (commented):
Thanks for the follow-up on the critique routing. The server/composer lockstep and adapter gating look addressed on this head; I found one small prompt wording issue that is worth tightening but should not block merge.


mrcfps (approved):
@Nagendhra-web I reviewed the current Phase 5 head across the panel prompt template, system composer wiring, server-side critique eligibility gate, and the focused daemon prompt tests. The previous routing, adapter eligibility, prompt injection, weights, designer-role, and MUST_FIX wording concerns all look addressed on this head; I did not find any remaining actionable issues in the changed ranges. Thanks for the careful follow-through on the review rounds 🙂


mrcfps (approved):
@Nagendhra-web I reviewed the current Phase 5 head across the panel prompt, composer/server critique eligibility wiring, focused daemon tests, and the small Ukrainian locale duplicate-key cleanup. The previous prompt/routing concerns look addressed, and the remaining CI failure is in an unrelated existing `NewProjectPanel` web test outside these changed ranges; I did not find any actionable issues in the PR diff. Thanks for the careful follow-through on the review rounds 🙂


mrcfps (approved):
@Nagendhra-web I reviewed the current Phase 5 head across the panel prompt template, system composer wiring, server-side critique eligibility gate, focused daemon prompt tests, and the NewProjectPanel fixture adjustment. The previous routing, adapter eligibility, prompt injection, weights, designer-role, and MUST_FIX wording concerns all look addressed in the changed ranges, and I did not find any remaining actionable issues. Thanks for the careful follow-through on the review rounds 🙂


Nagendhra-web:
Quick merge with `origin/main` in `838eb8c` to clear an unrelated `apps/web/tests/components/NewProjectPanel.test.tsx` add-add conflict, CI green. Took the canonical no-comment version of the same `designSystemRequired: true` fix that already landed on main, so the only delta against `main` for that file is now what was already there. No code or test behaviour changed; ready for the final merge whenever you're set.

mrcfps (approved):
@Nagendhra-web I reviewed the current Phase 5 head across the panel prompt template, system composer wiring, server-side critique eligibility gate, and focused daemon prompt tests. The previous routing, adapter eligibility, prompt injection, weights, designer-role, and MUST_FIX wording concerns all look addressed in the changed ranges, and CI is green. Thanks for the careful follow-through on the review rounds 🙂

PR #531, opened 2026-05-05
fix(daemon): ignore .venv and other large dirs in project file watcher

brown2hm:
## Summary

- Projects with large generated/vendor trees (Python `.venv`, Rust `target/`, Go `vendor/`, etc.) could exhaust the daemon's file descriptor table — chokidar recursively watched every file, opening tens of thousands of fds
- With the fd table full, macOS `posix_spawn` returned `EBADF` when the daemon tried to create stdio pipes for a child process, surfacing as `spawn failed: spawn EBADF` on every chat request
- `node_modules` was already excluded; this extends the same pattern to the Python, Rust, and Go equivalents

## Context

This was triggered by opening an existing repo (rather than creating one through the app) that contained a Python virtual environment. Bringing an established external repo into Open Design is a workflow not yet fully implemented — this makes it safer in the meantime.

## Changes

Adds `.venv`, `venv`, `__pycache__`, `.mypy_cache`, `.pytest_cache`, `.tox`, `.ruff_cache`, `target`, `vendor`, and `.cargo` to the per-segment `IGNORE_NAMES` set in `project-watchers.ts`.

## Test plan

- [ ] Open a project containing a `.venv` directory and send a chat message — agent should spawn without EBADF
- [ ] Confirm daemon fd count stays low (`lsof -p <daemon-pid> | wc -l`) after the watcher initializes on such a project
- [ ] Confirm file-change events still fire for edits to normal source files in the project root

### DIFF
--- a/apps/daemon/src/project-watchers.ts
+++ b/apps/daemon/src/project-watchers.ts
@@ -17,7 +17,26 @@ import { projectDir } from './projects.js';
 // against the path *relative to the watch root* so that ancestor directories
 // (e.g. the daemon's own `.od/` runtime dir, which contains every project) do
 // not accidentally match and silence every event in the tree.
-const IGNORE_NAMES = new Set(['.git', 'node_modules', '.od', 'debug', '.DS_Store']);
+const IGNORE_NAMES = new Set([
+  '.git',
+  'node_modules',
+  '.od',
+  'debug',
+  '.DS_Store',
+  // Python virtual environments and caches — can contain tens of thousands of
+  // files, exhausting the process fd table and breaking child-process spawning.
+  '.venv',
+  'venv',
+  '__pycache__',
+  '.mypy_cache',
+  '.pytest_cache',
+  '.tox',
+  '.ruff_cache',
+  // Other common large generated/vendor dirs.
+  'target',  // Rust / Java
+  'vendor',  // Go
+  '.cargo',
+]);
 export function makeIgnored(rootDir) {
   return (absPath) => {
     const rel = path.relative(rootDir, absPath);
--- a/apps/daemon/src/project-watchers.ts
+++ b/apps/daemon/src/project-watchers.ts
@@ -25,6 +25,8 @@ const IGNORE_NAMES = new Set([
   '.DS_Store',
   // Python virtual environments and caches — can contain tens of thousands of
   // files, exhausting the process fd table and breaking child-process spawning.
+  // These names are safe to match at any path depth: a directory named `.venv`
+  // or `__pycache__` is never legitimate authored source in a project tree.
   '.venv',
   'venv',
   '__pycache__',
@@ -32,10 +34,6 @@ const IGNORE_NAMES = new Set([
   '.pytest_cache',
   '.tox',
   '.ruff_cache',
-  // Other common large generated/vendor dirs.
-  'target',  // Rust / Java
-  'vendor',  // Go
-  '.cargo',
 ]);
 export function makeIgnored(rootDir) {
   return (absPath) => {
--- a/apps/daemon/tests/project-watchers.test.ts
+++ b/apps/daemon/tests/project-watchers.test.ts
@@ -181,6 +181,32 @@ describe('project-watchers (real chokidar)', () => {
     }
   }, 8_000);
 
+  it('ignores files inside Python venv and cache dirs', async () => {
+    const { root, projectId } = await makeProjectsRoot();
+    const events = [];
+    const sub = subscribe(root, projectId, (e) => events.push(e));
+    await sub.ready;
+
+    const ignoredDirs = ['.venv', 'venv', '__pycache__', '.mypy_cache', '.pytest_cache', '.tox', '.ruff_cache'];
+    try {
+      for (const dir of ignoredDirs) {
+        await mkdir(path.join(root, projectId, dir), { recursive: true });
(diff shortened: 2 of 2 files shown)


lefarcen (commented):
Hey @brown2hm, thanks for tackling this fd-exhaustion bug! 🙏

The fix itself is solid — adding the Python/Rust/Go vendor dirs to the ignore set directly addresses the root cause (chokidar recursing into tens of thousands of files). The diagnostic trail (fd exhaustion → posix_spawn EBADF) is precise and matches the proposed solution.

**Minor suggestion (non-blocking):**

The existing test suite in `apps/daemon/tests/project-watchers.test.ts:159` exercises the ignore behavior for `.od/` and `node_modules/` but doesn't yet cover the newly added directories. Without test coverage, a future refactor could accidentally drop these names, bringing back the fd-exhaustion regression. Consider extending that test to create files under representative new dirs (`.venv/`, `venv/`, `target/`, `vendor/`) and assert no `file-changed` events are emitted (while a normal file still does emit).

Not blocking — the immediate fd fix is production-ready without this; just future-proofing. Ready to merge from my side once CI passes. 🚀

apps/daemon/src/project-watchers.ts
```diff
@@ -17,7 +17,26 @@ import { projectDir } from './projects.js';
...
+  'venv',
+  '__pycache__',
+  '.mypy_cache',
+  '.pytest_cache',
+  '.tox',
+  '.ruff_cache',
+  // Other common large generated/vendor dirs.
+  'target',  // Rust / Java
+  'vendor',  // Go
+  '.cargo',
```
mrcfps: These three entries are matched by `makeIgnored()` against **any** relative path segment, not just top-level generated directories, so they can also suppress legitimate project files such as `src/vendor/...`, `src/target/...`, or Rust project config under `.cargo/config.toml`. That matters because this watcher drives `/api/projects/:id/events` and preview/file-list refreshes; edits under those paths would now be silently missed even when they are authored source/config rather than generated dependency output. Consider narrowing the new broad names to generated roots (for example, root-relative checks for `target/`, Go `vendor/`, and any intentionally generated `.cargo/` case) or dropping `.cargo` unless there is a concrete large generated tree to avoid.

mrcfps (commented):
@brown2hm thanks for chasing down the fd-exhaustion path here — extending the watcher ignore list is a targeted fix for the `.venv`/generated-tree case. I found one merge-safe maintainability/correctness concern around the broad per-segment ignores below.


apps/daemon/src/project-watchers.ts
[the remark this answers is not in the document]
brown2hm: Noted, I've updating to drop .cargo, target, and vendor in the follow-up commit (0c9f9cb).

mrcfps (approved):
@brown2hm I reviewed the narrowed watcher ignore update and the added real-chokidar coverage for the Python venv/cache directories. The follow-up removes the broader Rust/Go entries that could have hidden authored files, while still addressing the fd-exhaustion path from `.venv`-style trees; thank you for iterating on this fix and tightening the test coverage 🙏


brown2hm:
@lefarcen Good call. I added a real-chokidar test in the newer commit.

apps/daemon/src/project-watchers.ts
brown2hm: I commented on the codex review above, but I've updating to drop .cargo, target, and vendor in the follow-up commit (https://github.com/nexu-io/open-design/commit/0c9f9cb7fdd179e651b6fd8728e16b87db5c297d). Good catch.

lefarcen (commented):
Hey @brown2hm, nice work addressing the path-matching concern! 🎉

The current changes look solid:
- ✅ Removed the three broad names (`target`, `vendor`, `.cargo`) that could suppress legitimate source files like `src/vendor/` or `.cargo/config.toml`
- ✅ Kept only Python-specific names (`.venv`, `venv`, `__pycache__`, etc.) — these are genuinely safe at any depth, as the comment notes
- ✅ Added test coverage for the Python venv/cache dirs, verifying no events are emitted while normal files still trigger

This resolves @mrcfps's P2 concern and adds the test coverage I mentioned earlier. The fd-exhaustion fix is now production-ready without the over-broad path-segment matching risk. Ready to merge from my side once CI passes! 🚀

lefarcen:
Merged! 🎉 Thank you @brown2hm for nailing down the fd-exhaustion path — projects with // directories are now safe to open. The fix narrows the ignore list to Python cache/venv only (no Rust/Go broad names), and you added real chokidar test coverage on top of that 🙏

This fix shipped in the latest main — anyone who was hitting `spawn EBADF` with large Python projects should be good now.

— open-design team

lefarcen:
Merged! 🎉 Thank you @brown2hm for nailing down the fd-exhaustion path — projects with `.venv`/`target`/`vendor` directories are now safe to open. The fix narrows the ignore list to Python cache/venv only (no Rust/Go broad names), and you added real chokidar test coverage on top of that 🙏

This fix shipped in the latest main — anyone who was hitting `spawn EBADF` with large Python projects should be good now.

— open-design team

leilei926524-tech:
Hi @brown2hm — welcome to Open Design contributors. 🎉

Your first merged PR (#531) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #535, opened 2026-05-05
feat(web): add skills & design systems management page in settings

### LINKED ISSUE #497: feat(web): add Skills & Design Systems management page in Settings
## Summary

Add a new `skills` section to the Settings dialog that lets users browse, preview, enable/disable, and install skills and design systems from within the UI — instead of manually dropping folders into `skills/` and `design-systems/`.

## Current state

- The daemon already provides `GET /api/skills` and `GET /api/design-systems` listing endpoints, plus `GET /api/design-systems/:id` for full content
- Skills and design systems are discovered by filesystem scanning (`skills/`, `design-systems/`, plus user-level dirs)
- There is **no UI** for browsing or managing these — users must interact with the filesystem directly
- The project ships 60+ skills and 69+ design systems, many of which users may never discover

### 1. Browse installed skills & design systems
- Two-tab layout within the section: **Skills** | **Design Systems**
- List view showing: name, mode (`prototype` / `deck` / `template`), description snippet
- Group skills by `od.mode`, group design systems by `> Category` from `DESIGN.md`

### 2. Preview
- Click an item to expand/preview: full `SKILL.md` or `DESIGN.md` content
- Show `example.html` screenshot for skills that have one
- Show which projects currently use this skill/DS (if any)

### 3. Enable / Disable
- Toggle visibility in the project creation picker without deleting files
- Implementation: maintain a `disabledSkills` / `disabledDesignSystems` list in `AppConfig` (daemon-side)
- Disabled items remain on disk but are filtered from `GET /api/skills` / `GET /api/design-systems` responses

### 4. Install / Uninstall
- `POST /api/skills/install` — accepts a GitHub URL or local path, clones/symlinks into the user-level skills directory
- `DELETE /api/skills/:id` — removes from the registry (unlink for symlinks, rm for clones)
- Same endpoints for design systems
- Reuses the symlink strategy from `docs/skills-protocol.md` §3

## Technical approach

**Frontend:**
- New section in `SettingsDialog.tsx`: `skills` → added to `SettingsSection` union type
- New component: `SkillsDesignSystemsSection` with tabs, list, and preview
- Reuse existing API calls from `providers/registry.ts`

**Backend (daemon):**
- New endpoints: `POST /api/skills/install`, `DELETE /api/skills/:id`, `POST /api/design-systems/install`, `DELETE /api/design-systems/:id`
- Extend `GET /api/skills` and `GET /api/design-systems` to accept a `?include=disabled` query param
- Add `disabledSkills` / `disabledDesignSystems` arrays to `AppConfig`

**No new dependencies.** Uses `child_process` for git clone (already used elsewhere in the daemon).

## Open questions

1. **Install scope?** Should installs go to the project-level `skills/` dir or the user-level `~/.open-design/skills/`? I'd suggest user-level by default (available across all projects), with an option for project-level.
2. **Naming?** The section could be called "Skills & Design Systems", "Library", "Resources", or something else. Open to suggestions.
3. **Chunking?** With 60+ skills, the list gets long. Should we paginate, search/filter, or just virtualize the list?

## Relation

This issue is the foundation for enabling mid-chat skill/DS switching via the import menu (see #498 ).

encyc:
## Summary

- Add a new **Library** section in Settings for browsing, searching, previewing, and enabling/disabling skills and design systems
- Disabled items are filtered out from the create-project picker so users only see what they've enabled
- Phase 1 of #497 — browse, preview, search/filter, enable/disable only (no install/uninstall)

### Changes

| Area | Change |
|------|--------|
| `AppConfig` types | Added `disabledSkills` and `disabledDesignSystems` string array fields |
| Daemon config | Added both fields to `ALLOWED_KEYS` with string-array validation |
| Config sync | Wired new fields into `syncConfigToDaemon` |
| SettingsDialog | Added `library` section with sidebar nav entry |
| LibrarySection | New component: two-tab layout (Skills / Design Systems), search, filter pills, toggle switches, expandable preview |
| CSS | New `.library-*` styles and `.toggle-switch` component |
| i18n | 13 new keys across all 16 locales (native translations) |
| App.tsx | Filters disabled skills/DS from both ProjectView and EntryView |
| Daemon tests | 5 new test cases for disabled-list validation |

Closes #497

## Test plan

- [ ] Open Settings → Library section appears in sidebar
- [ ] Skills tab shows all skills grouped by mode with filter pills
- [ ] Design Systems tab shows all DS grouped by category
- [ ] Search filters both tabs by name/description
- [ ] Clicking a skill/DS card expands a preview panel
- [ ] Toggle disable on a skill → create new project → skill not in picker
- [ ] Toggle re-enable → skill reappears
- [ ] Settings persist across page reload
- [ ] All 16 locale files have native translations (not English)


lefarcen:
Hi @encyc! 🎉
Thanks for the contribution — this is a substantial step toward making skills and design systems more discoverable in the UI.

— open-design team

apps/daemon/src/app-config.ts
```diff
@@ -84,6 +88,11 @@ function applyConfigValue(
       delete target[key];
     }
   }
+  if (key === 'disabledSkills' || key === 'disabledDesignSystems') {
+    if (Array.isArray(value) && value.every((v) => typeof v === 'string')) {
+      target[key] = value;
```
lefarcen, line 93: **P2** Edge case: when invalid data is sent for `disabledSkills`/`disabledDesignSystems` (non-array or array with non-strings), this block silently doesn't set the field — but it also doesn't delete any existing value.

If a user had `disabledSkills: ["a"]` and then sends malformed data, the old value persists instead of being cleared.

Suggestion:
```ts
if (key === 'disabledSkills' || key === 'disabledDesignSystems') {
  if (Array.isArray(value) && value.every((v) => typeof v === 'string')) {
    target[key] = value;
  } else {
    delete target[key];  // clear on invalid
  }
}
```

This matches the pattern for other fields (e.g. `agentModels` explicitly deletes on validation failure).

lefarcen (commented):
Hi @encyc! Solid implementation — the feature scope matches issue #497 Phase 1 precisely, config validation is well-tested, and the i18n translations are native. A few non-blocking optimization opportunities below:

**Code correctness:**
- **P2** edge case in daemon validation + **P2** render perf in App.tsx (see inline)
- **P3** error handling and race condition notes

None of these block the feature from shipping. If you'd like to address them now, great; otherwise they can be follow-up tweaks in Phase 2.

Thanks for the clean, thorough work here! 🎉

apps/web/src/App.tsx
```diff
@@ -529,8 +529,8 @@ export function App() {
           routeFileName={route.kind === 'project' ? route.fileName : null}
           config={config}
           agents={agents}
-          skills={skills}
-          designSystems={designSystems}
+          skills={skills.filter((s) => !(config.disabledSkills ?? []).includes(s.id))}
```
lefarcen: **P2** Render perf: filtering on every render is O(n*m) — with 60+ skills and potentially growing disabled lists, this adds up.

Since `skills` and `config.disabledSkills` are stable references, consider memoizing:
```tsx
const enabledSkills = useMemo(
  () => skills.filter((s) => !(config.disabledSkills ?? []).includes(s.id)),
  [skills, config.disabledSkills]
);
const enabledDS = useMemo(
  () => designSystems.filter((d) => !(config.disabledDesignSystems ?? []).includes(d.id)),
  [designSystems, config.disabledDesignSystems]
);
// then pass enabledSkills / enabledDS to ProjectView and EntryView
```

Not urgent for 60 items, but scales better as the library grows.

apps/web/src/components/LibrarySection.tsx
```diff
@@ -0,0 +1,362 @@
...
+        >
+          <span className="seg-title">
+            {t('settings.librarySkills')}
+            <span className="seg-meta">{skills.length}</span>
+          </span>
+        </button>
+        <button
+          type="button"
+          role="tab"
+          className={`seg-btn${tab === 'design-systems' ? ' active' : ''}`}
```
lefarcen, line 184: **P3** No error handling: if `syncConfigToDaemon` fails (network timeout, disk full), the user sees the toggle flip but the actual state didn't persist.

Options:
1. Add optimistic UI + rollback on failure
2. Show a toast/error message if the daemon write fails
3. Disable toggles while write is pending

Not a blocker, but UX polish for Phase 2.

apps/web/src/components/LibrarySection.tsx
```diff
@@ -0,0 +1,362 @@
...
+    <section className="settings-section">
+      <div className="section-head">
+        <div>
+          <h3>{t('settings.library')}</h3>
+          <p className="hint">{t('settings.libraryHint')}</p>
+        </div>
+      </div>
+
+      <div className="seg-control" role="tablist">
+        <button
```
lefarcen, line 164: **P3** Race condition: if the user rapidly clicks two different cards, both fetches fire, and whichever resolves last wins — potentially showing preview B's body under card A's title.

Fix: add an abort controller or check `previewId` hasn't changed in the `finally` block:
```tsx
const openPreview = useCallback(
  async (id: string) => {
    if (previewId === id) {
      setPreviewId(null);
      setPreviewBody(null);
      return;
    }
    setPreviewId(id);
    setPreviewBody(null);
    setPreviewLoading(true);
    const expectedId = id;  // capture
    try {
      // ... fetch ...
      if (expectedId === previewId) {  // still current?
        setPreviewBody(detail?.body ?? null);
      }
    } catch { ... } finally {
      if (expectedId === previewId) setPreviewLoading(false);
    }
  },
  [previewId, tab],
);
```

Likely rare in practice, but good defensive coding.

lefarcen (commented):
Thanks for the quick iteration @encyc! All four suggestions from the prior review have been addressed:

✅ **P2 daemon edge case** — invalid data now explicitly clears the field (line 94)
✅ **P2 render perf** — skills/DS filtering is now memoized (lines 523-532)
✅ **P3 race condition** — preview fetches now check staleness before updating state (lines 117-130)
✅ **P3 error handling** — race fix also makes error paths safer

The implementation looks solid. Ready for maintainer review. 🎉

apps/web/src/App.tsx
```diff
@@ -520,6 +520,18 @@ export function App() {
     void refreshTemplates();
   }, [route.kind, refreshTemplates]);
 
+  const enabledSkills = useMemo(
+    () => skills.filter((s) => !(config.disabledSkills ?? []).includes(s.id)),
```
mrcfps, line 530: These new config fields are now used to decide which library items are enabled, and this PR also writes them to `/api/app-config`, but the bootstrap merge above still does not copy `daemonConfig.disabledSkills` or `daemonConfig.disabledDesignSystems` into `next` before saving/syncing. After a localStorage/origin reset — the persistence case `app-config` is meant to cover — daemon-stored disabled items are ignored and all disabled skills/design systems reappear. Please hydrate both fields when merging `daemonConfig`, preserving empty arrays as meaningful values, e.g. `if (daemonConfig.disabledSkills !== undefined) next.disabledSkills = daemonConfig.disabledSkills` and the same for design systems.

apps/web/src/App.tsx
```diff
@@ -529,8 +541,8 @@ export function App() {
           routeFileName={route.kind === 'project' ? route.fileName : null}
           config={config}
           agents={agents}
-          skills={skills}
-          designSystems={designSystems}
+          skills={enabledSkills}
+          designSystems={enabledDS}
```
mrcfps: Passing `enabledSkills`/`enabledDS` into `ProjectView` filters disabled items out of existing projects, not just out of the create-project picker. `ProjectView` uses these registries to resolve `project.skillId`/`project.designSystemId` metadata, prompt context, and deck detection, so if a user disables a skill or design system after creating a project with it, reopening that project can lose names/context or make a deck project look non-deck. Keep `ProjectView` on the full `skills`/`designSystems` lists and only pass the filtered lists to `EntryView`/creation pickers.

mrcfps (changes requested):
@encyc thanks for the thoughtful Library settings work here — the UI and daemon validation pieces are nicely scoped. I found two main-path issues in how the new disabled registries are applied/hydrated that should be fixed before merge so existing projects and daemon-backed persistence keep working as intended. 🙂


lefarcen (commented):
Both of mrcfps's blocking issues have been addressed! 🎉

✅ **ProjectView filtering fixed** — `ProjectView` now receives full `skills`/`designSystems` lists (line 544-545), so existing projects keep their metadata and deck detection even when skills/DS are disabled

✅ **Daemon hydration fixed** — Bootstrap now properly copies `daemonConfig.disabledSkills` and `disabledDesignSystems` into `next` with the recommended undefined checks (lines 185-190)

✅ **EntryView unaffected** — Creation picker still receives filtered `enabledSkills`/`enabledDS` as intended

The fixes are surgical and correct. Ready for mrcfps to re-review.

mrcfps (approved):
@encyc I reviewed the Library settings implementation across the web config flow, daemon persistence/validation, registry filtering, Settings UI, i18n additions, and the follow-up fixes for daemon hydration and existing-project behavior. The changed ranges look consistent with the Phase 1 scope, and the prior blocking concerns are addressed cleanly — thank you for the thoughtful iteration here! 🎉


lefarcen (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @encyc for the contribution to open-design. Looking forward to the next one!

PR #65, opened 2026-04-29
Add Docker Compose deployment workflow

wangwanjie:
## Summary
- Add a single Alpine-based Docker runtime image for Open Design.
- Add Docker Compose deployment with persistent `/app/.od` storage, health checks, memory limits, and `.env.example` defaults.
- Add Docker Hub publishing and verification scripts with multi-arch support, proxy detection, skopeo push support, and manifest checks.

## Test Plan
- `bash -n deploy/scripts/publish-images.sh && bash -n deploy/scripts/verify-image.sh && bash -n deploy/scripts/verify-image-manifest.sh`
- `docker compose --env-file deploy/.env.example -f deploy/docker-compose.yml config`
- `deploy/scripts/publish-images.sh --dry_run --arch arm64 --image_tag pr-check`
- `git diff --check`
- `docker buildx build --platform linux/arm64 -t open-design:pr-check -f deploy/Dockerfile --load .`
- `deploy/scripts/verify-image.sh open-design:pr-check`

## Published Image
- `docker.io/vanjayak/open-design:latest`
- Manifest includes `linux/amd64` and `linux/arm64`.

PR #554, opened 2026-05-05
feat: add Docker support for daemon deployment

MrCx0810:
## Summary
- Add `Dockerfile` to build and run the Open Design daemon in a container
- Add `.dockerignore` to keep the build context lean
- Add `docker-compose.yml` for easy management in Docker Desktop with one-click start/stop

## Motivation
Users who prefer containerized workflows can now run the Open Design daemon via Docker without installing Node.js and pnpm on their host. The Docker setup includes healthcheck monitoring and persistent storage for the `.od/` data directory.

## Test plan
- [x] `docker compose up -d --build` builds and starts successfully
- [x] `curl http://localhost:7456/api/health` returns `{"ok":true}`
- [x] `curl http://localhost:7456/api/skills` returns the full skills list
- [x] MCP server (`od mcp --daemon-url http://localhost:7456`) connects and serves tools
- [x] Container shows healthy in Docker Desktop

## Notes
The MCP stdio server still needs to run on the host (configured via `.mcp.json`) since Docker containers cannot expose stdio-based MCP servers to the host. This PR focuses on containerizing the daemon itself.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,9 +4,20 @@
 .github
 node_modules
 .od
-*.md
-*.png
-*.jpg
+.tmp
+.claude
+.codex
+.opencode
+.env*
+.npmrc
+reports
+
+# Exclude top-level docs and images only (runtime .md and .png/jpg under
+# skills/, design-systems/, craft/, assets/ must remain in the image)
+/*.md
+/*.png
+/*.jpg
+
 Dockerfile
 .dockerignore
 docker-compose.yml
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,16 @@
+.git
+.gitignore
+.gitattributes
+.github
+node_modules
+.od
+*.md
+*.png
+*.jpg
+Dockerfile
+.dockerignore
+docker-compose.yml
+.edited_image.png
+e2e
+specs
+story
--- a/Dockerfile
+++ b/Dockerfile
@@ -57,4 +57,7 @@ EXPOSE 7456
 
 ENV OD_BIND_HOST=0.0.0.0
 
-CMD ["node", "apps/daemon/dist/cli.js", "--no-open", "--host", "0.0.0.0"]
+# Bind host is controlled by OD_BIND_HOST env var (default 0.0.0.0).
+# Do NOT add --host here — the CLI flag overrides the env, preventing
+# Compose users from changing OD_BIND_HOST without also replacing CMD.
+CMD ["node", "apps/daemon/dist/cli.js", "--no-open"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,25 @@
+FROM node:24-slim
+
+# Build deps needed for better-sqlite3 native module
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python3 make gcc g++ \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN corepack enable && corepack prepare pnpm@10 --activate
+
+WORKDIR /app
+
+# Copy entire repo (use .dockerignore to skip .git, etc.)
+COPY . .
+
+# Install dependencies (postinstall builds workspace packages)
+RUN pnpm install --frozen-lockfile
+
+# Build the daemon CLI
+RUN pnpm -C apps/daemon run build
+
+EXPOSE 7456
+
+ENV OD_BIND_HOST=0.0.0.0
+
+CMD ["node", "apps/daemon/dist/cli.js", "--no-open", "--host", "0.0.0.0"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,22 +1,57 @@
-FROM node:24-slim
+# -- Build stage -------------------------------------------------------
+FROM node:24-slim AS builder
 
-# Build deps needed for better-sqlite3 native module
 RUN apt-get update && apt-get install -y --no-install-recommends \
     python3 make gcc g++ \
     && rm -rf /var/lib/apt/lists/*
 
-RUN corepack enable && corepack prepare pnpm@10 --activate
+# Pin to the exact pnpm version declared in package.json
+RUN corepack enable && corepack prepare pnpm@10.33.2 --activate
 
 WORKDIR /app
 
-# Copy entire repo (use .dockerignore to skip .git, etc.)
 COPY . .
 
 # Install dependencies (postinstall builds workspace packages)
 RUN pnpm install --frozen-lockfile
 
-# Build the daemon CLI
-RUN pnpm -C apps/daemon run build
+# Build the daemon CLI and the static web UI
+RUN pnpm -C apps/daemon run build \
+    && pnpm --filter @open-design/web build
+
+# -- Runtime stage ------------------------------------------------------
+FROM node:24-slim
+
+RUN corepack enable && corepack prepare pnpm@10.33.2 --activate
+
+# Create unprivileged user
+RUN groupadd -r oduser && useradd -r -g oduser oduser \
+    && mkdir -p /app/.od && chown -R oduser:oduser /app
+
+WORKDIR /app
+
+# Runtime dependencies (pnpm workspace resolution)
+COPY --from=builder --chown=oduser:oduser /app/package.json /app/pnpm-lock.yaml /app/pnpm-workspace.yaml ./
+COPY --from=builder --chown=oduser:oduser /app/node_modules ./node_modules
+
+# Workspace package manifests (needed for module resolution)
+COPY --from=builder --chown=oduser:oduser /app/packages ./packages
+COPY --from=builder --chown=oduser:oduser /app/tools ./tools
+COPY --from=builder --chown=oduser:oduser /app/apps ./apps
+
+# Built artifacts
+COPY --from=builder --chown=oduser:oduser /app/apps/daemon/dist ./apps/daemon/dist
+COPY --from=builder --chown=oduser:oduser /app/apps/web/out ./apps/web/out
+
+# Runtime data directories
+COPY --from=builder --chown=oduser:oduser /app/skills ./skills
+COPY --from=builder --chown=oduser:oduser /app/design-systems ./design-systems
+COPY --from=builder --chown=oduser:oduser /app/prompt-templates ./prompt-templates
+COPY --from=builder --chown=oduser:oduser /app/templates ./templates
+COPY --from=builder --chown=oduser:oduser /app/craft ./craft
+COPY --from=builder --chown=oduser:oduser /app/assets ./assets
+
+USER oduser
 
 EXPOSE 7456
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -55,9 +55,9 @@ USER oduser
 
 EXPOSE 7456
 
-ENV OD_BIND_HOST=0.0.0.0
+# Default to loopback-only — the daemon is local-first with no auth.
+# To bind broadly, explicitly set OD_BIND_HOST=0.0.0.0 at runtime.
+ENV OD_BIND_HOST=127.0.0.1
 
-# Bind host is controlled by OD_BIND_HOST env var (default 0.0.0.0).
-# Do NOT add --host here — the CLI flag overrides the env, preventing
-# Compose users from changing OD_BIND_HOST without also replacing CMD.
+# Do NOT add --host here — the CLI flag overrides OD_BIND_HOST env.
 CMD ["node", "apps/daemon/dist/cli.js", "--no-open"]
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -4,8 +4,9 @@ services:
     container_name: open-design
     ports:
       # Bind only to loopback by default — the daemon is local-first
-      # with no auth. To expose on a LAN, replace 127.0.0.1 with the
-      # desired interface IP and put a reverse proxy / auth layer in front.
+      # with no auth. To expose beyond localhost, you must use a
+      # reverse proxy (nginx, caddy) that rewrites the Origin header,
+      # set OD_BIND_HOST to the actual interface IP, and add auth.
       - "127.0.0.1:7456:7456"
     volumes:
       - od_data:/app/.od
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,0 +1,22 @@
+services:
+  open-design:
+    build: .
+    container_name: open-design
+    ports:
+      - "7456:7456"
+    volumes:
+      - od_data:/app/.od
+    restart: unless-stopped
+    environment:
+      - OD_PORT=7456
+      - OD_BIND_HOST=0.0.0.0
+    healthcheck:
+      test: ["CMD", "node", "-e", "require('http').get('http://localhost:7456/api/health', (r) => { process.exit(r.statusCode === 200 ? 0 : 1) })"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 30s
+
+volumes:
+  od_data:
+    driver: local
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -15,7 +15,10 @@ services:
       - OD_PORT=7456
       - OD_BIND_HOST=127.0.0.1
     healthcheck:
-      test: ["CMD", "node", "-e", "require('http').get('http://localhost:7456/api/health', (r) => { process.exit(r.statusCode === 200 ? 0 : 1) })"]
+      # Use 127.0.0.1 instead of localhost — on hosts where localhost
+      # resolves to ::1 first, the IPv6 connection would miss the IPv4
+      # listener and mark the container unhealthy.
+      test: ["CMD", "node", "-e", "require('http').get('http://127.0.0.1:7456/api/health', (r) => { process.exit(r.statusCode === 200 ? 0 : 1) })"]
       interval: 30s
       timeout: 10s
       retries: 3
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -3,7 +3,10 @@ services:
     build: .
     container_name: open-design
     ports:
-      - "7456:7456"
+      # Bind only to loopback by default — the daemon is local-first
+      # with no auth. To expose on a LAN, replace 127.0.0.1 with the
+      # desired interface IP and put a reverse proxy / auth layer in front.
+      - "127.0.0.1:7456:7456"
     volumes:
       - od_data:/app/.od
     restart: unless-stopped
@@ -15,7 +18,7 @@ services:
       interval: 30s
       timeout: 10s
       retries: 3
-      start_period: 30s
+      start_period: 60s
 
 volumes:
   od_data:
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -2,18 +2,18 @@ services:
   open-design:
     build: .
     container_name: open-design
-    ports:
-      # Bind only to loopback by default — the daemon is local-first
-      # with no auth. To expose beyond localhost, you must use a
-      # reverse proxy (nginx, caddy) that rewrites the Origin header,
-      # set OD_BIND_HOST to the actual interface IP, and add auth.
-      - "127.0.0.1:7456:7456"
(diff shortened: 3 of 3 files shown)


lefarcen:
Hi @MrCx0810! 🎉
Thanks for the contribution — adding Docker support for daemon deployment is a great addition for containerized workflows.

— open-design team

docker-compose.yml
```diff
@@ -0,0 +1,22 @@
+services:
+  open-design:
+    build: .
+    container_name: open-design
+    ports:
+      - "7456:7456"
```
lefarcen: **P1 Security** — This publishes the daemon on all host interfaces (`"7456:7456"`), while the daemon is local-first and many API routes accept non-browser requests with no auth. On a LAN, another machine could call routes like `/api/runs`, `/api/chat`, project/file APIs, etc.

**Fix**: Bind the host side to loopback by default:
```yaml
ports:
  - "127.0.0.1:7456:7456"
```
Require explicit opt-in plus auth/reverse-proxy guidance for remote exposure.

.dockerignore
```diff
@@ -0,0 +1,16 @@
+.git
+.gitignore
+.gitattributes
+.github
+node_modules
+.od
+*.md
```
lefarcen: **P1 Build context** — The broad `*.md`, `*.png`, and `*.jpg` patterns remove runtime resources from the image. The daemon discovers:
- skills from `skills/*/SKILL.md`
- design systems from `design-systems/*/DESIGN.md`
- craft rules from `craft/*.md`
- some shipped skill/prompt assets are images

The container can pass `/api/health` while `/api/skills`, design-system prompts, and asset-backed examples are incomplete or empty.

**Fix**: Remove these broad ignores or replace them with targeted top-level doc ignores plus negations:
```
# Exclude top-level docs only
/*.md
!skills/**/*.md
!design-systems/**/*.md
!craft/**/*.md
# Keep required assets
!assets/**
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+WORKDIR /app
+
+# Copy entire repo (use .dockerignore to skip .git, etc.)
+COPY . .
+
+# Install dependencies (postinstall builds workspace packages)
+RUN pnpm install --frozen-lockfile
+
+# Build the daemon CLI
+RUN pnpm -C apps/daemon run build
```
lefarcen: **P1 Incomplete deployment** — The image builds only `apps/daemon`, but the daemon serves the production web UI only if `apps/web/out` exists. This Docker image therefore exposes the daemon API but not the normal Open Design UI, which is surprising for a deployable `od` service.

**Fix**: Either build the static web app too before starting the daemon:
```dockerfile
RUN pnpm --filter @open-design/web build
RUN pnpm -C apps/daemon run build
```
Or explicitly make this an API-only image and document that the web runtime must be deployed separately.

.dockerignore
```diff
@@ -0,0 +1,16 @@
+.git
```
lefarcen, line 1: **P2 Secret leakage** — The ignore file does not mirror local/runtime secret exclusions from `.gitignore`. Because `Dockerfile:13` does `COPY . .`, local `.tmp/`, agent scratch directories like `.claude/`, `.codex/`, `.opencode/`, `.env*`, `.npmrc`, reports, or alternate credential/data directories can be sent to the build context and baked into the image.

**Fix**: Add explicit exclusions:
```
.tmp
.claude
.codex
.opencode
.env*
.npmrc
reports
```

Dockerfile
```diff
@@ -0,0 +1,25 @@
+FROM node:24-slim
```
lefarcen: **P2 Runtime hardening** — The runtime image keeps compiler toolchains (`python3 make gcc g++`) and runs the daemon as root. For a deployment image, that increases attack surface and makes persisted `/app/.od` data root-owned.

**Fix**: Use a multi-stage build:
```dockerfile
# Build stage
FROM node:24-slim AS builder
RUN apt-get update && apt-get install -y --no-install-recommends \
    python3 make gcc g++ \
    && rm -rf /var/lib/apt/lists/*
...

# Runtime stage
FROM node:24-slim
RUN corepack enable && corepack prepare pnpm@10.33.2 --activate
WORKDIR /app

# Create unprivileged user
RUN groupadd -r oduser && useradd -r -g oduser oduser \
    && mkdir -p /app/.od && chown -R oduser:oduser /app

# Copy only runtime artifacts
COPY --from=builder --chown=oduser:oduser /app/node_modules ./node_modules
COPY --from=builder --chown=oduser:oduser /app/apps/daemon/dist ./apps/daemon/dist
...

USER oduser
```

lefarcen (dismissed):
Hey @MrCx0810, thanks for this PR! Docker support is a valuable addition. 🎉

I found 3 critical issues that need fixing before merge:

1. **P1 Security**: The daemon publishes on all interfaces by default, but it's designed as local-first with no auth
2. **P1 Build context**: `.dockerignore` removes runtime resources (skill SKILL.md files, design-system metadata, assets)
3. **P1 Incomplete deployment**: The image only builds the daemon API, missing the web UI

See inline comments below for concrete fixes. Happy to review again once these are addressed!

mrcfps (changes requested):
@MrCx0810 thank you for adding Docker support here — this is a valuable path for containerized workflows. I completed a changed-range pass and am requesting changes because the current Docker packaging still has blocking deployment/security issues already captured in the unresolved inline threads: `docker-compose.yml:6` exposes the local-first daemon on all host interfaces, `.dockerignore:7` strips required runtime Markdown/assets from the build context, and `Dockerfile:19` builds only the daemon without the web UI bundle. Please address those before merge; once fixed, this should be a much safer and more complete Docker setup. 🙏


MrCx0810 (commented):
Thanks for the thorough review @lefarcen and @chatgpt-codex-connector! All issues have been addressed in the updated commit:

1. **P1 Security**: Port binding changed to `127.0.0.1:7456:7456`, with a comment explaining opt-in for LAN exposure.
2. **P1 Build context**: Changed broad `*.md`/`*.png`/`*.jpg` patterns to top-level-only `/*.md` etc., so skills/SKILL.md, design-systems/DESIGN.md, and asset images stay in the image.
3. **P1 Incomplete deployment**: Added `pnpm --filter @open-design/web build` to produce `apps/web/out` — the full web UI is now included.
4. **P2 Secret leakage**: Added exclusions for `.tmp`, `.claude`, `.codex`, `.opencode`, `.env*`, `.npmrc`, `reports`.
5. **P2 Runtime hardening**: Switched to multi-stage build with unprivileged `oduser` account at runtime — no compiler toolchains in the final image.
6. **P2 Pin pnpm**: Locked to exact `10.33.2` matching `packageManager` in package.json.

Please take another look when you get a chance!

Dockerfile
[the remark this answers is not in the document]
KatonJ: Fixed in the updated commit.

.dockerignore
KatonJ: Fixed in the updated commit.

Dockerfile
KatonJ: Fixed in the updated commit.

Dockerfile
KatonJ: Fixed in the updated commit.

.dockerignore
KatonJ: Fixed in the updated commit.

Dockerfile
```diff
@@ -0,0 +1,25 @@
...
+
+# Build deps needed for better-sqlite3 native module
+RUN apt-get update && apt-get install -y --no-install-recommends \
+    python3 make gcc g++ \
+    && rm -rf /var/lib/apt/lists/*
+
+RUN corepack enable && corepack prepare pnpm@10 --activate
```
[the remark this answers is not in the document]
KatonJ: Fixed in the updated commit.

docker-compose.yml
KatonJ: Fixed in the updated commit.

lefarcen (commented):
Hey @MrCx0810! 👏

Thanks for the quick turnaround — I've verified all 6 findings are now resolved:

✅ **P1 Security**: Port binding changed to `127.0.0.1:7456:7456` with clear LAN opt-in guidance  
✅ **P1 Build context**: Changed to top-level-only exclusions (`/*.md` etc), skills/design-systems/craft/assets remain intact  
✅ **P1 Web UI**: Added `pnpm --filter @open-design/web build` — full UI now included  
✅ **P2 Secret exclusions**: Added `.tmp`, `.claude`, `.codex`, `.opencode`, `.env*`, `.npmrc`, `reports`  
✅ **P2 Runtime hardening**: Multi-stage build with unprivileged `oduser`, no build tools in final image  
✅ **P2 pnpm pinned**: Locked to exact `10.33.2`

Looks good to me; deferring final approval to a maintainer. Great work! 🎉

docker-compose.yml
```diff
@@ -0,0 +1,25 @@
+services:
+  open-design:
+    build: .
+    container_name: open-design
+    ports:
+      # Bind only to loopback by default — the daemon is local-first
+      # with no auth. To expose on a LAN, replace 127.0.0.1 with the
```
mrcfps: The comment says users can expose this by replacing `127.0.0.1` with a LAN/interface IP, but with the current compose environment the daemon still starts with `OD_BIND_HOST=0.0.0.0` (`docker-compose.yml:15`). The daemon’s browser-origin allowlist is built from loopback plus the literal bind host (`apps/daemon/src/server.ts:1051-1068`), so a browser loading `http://<LAN-IP>:7456` would send `Origin: http://<LAN-IP>:7456` and API calls can be rejected because the allowed host is `0.0.0.0`, not the LAN IP. This is merge-safe because the default remains loopback-only, but the opt-in guidance is likely to lead to a broken remote UI.

Please either remove the direct LAN replacement recipe or document a supported reverse-proxy/origin setup rather than implying the port mapping change alone is enough.

mrcfps (commented):
@MrCx0810 thanks for the quick follow-up on the Docker support. I rechecked the Dockerfile, build context, static UI export, unprivileged runtime, and loopback binding; the prior blockers look addressed. I left one non-blocking note on the LAN-exposure guidance so users don’t hit a confusing origin failure. 🙏


docker-compose.yml
KatonJ: Good catch — updated the comment. Now it recommends a reverse proxy approach and explicitly notes that OD_BIND_HOST must match the interface IP. Just swapping the port binding IP is not enough because the daemon's origin allowlist uses the literal OD_BIND_HOST value (which defaults to 0.0.0.0, not the LAN IP).

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
+services:
+  open-design:
+    build: .
+    container_name: open-design
+    ports:
+      # Bind only to loopback by default — the daemon is local-first
+      # with no auth. To expose beyond localhost, you must use a
+      # reverse proxy (nginx, caddy) that rewrites the Origin header,
+      # set OD_BIND_HOST to the actual interface IP, and add auth.
+      - "127.0.0.1:7456:7456"
```
mrcfps: The default Compose path still routes host browser/API traffic through Docker's port-publishing NAT, so requests that reach Express are not guaranteed to have a loopback `req.socket.remoteAddress`. The daemon explicitly rejects non-loopback peers in `validateLocalDaemonRequest` before serving several local-only routes, including live artifact preview/refresh and connector connect/delete (`apps/daemon/src/server.ts:684-727`, `apps/daemon/src/connectors/routes.ts:355,404`). That means the container can pass `/api/health` and basic unauthenticated endpoints while normal UI actions that hit those local-only routes return 403 from Docker Desktop's bridge/gateway address.

Please make the Docker deployment preserve a loopback peer for the daemon or add an explicit Docker-aware local-peer allowlist/trust mechanism for this supported compose mode, with a smoke test against one of the `requireLocalDaemonRequest` routes (not just `/api/health`). Without that, the primary “run Open Design daemon via Docker Desktop” flow remains partially broken.

Dockerfile
```diff
@@ -0,0 +1,60 @@
...
+COPY --from=builder --chown=oduser:oduser /app/craft ./craft
+COPY --from=builder --chown=oduser:oduser /app/assets ./assets
+
+USER oduser
+
+EXPOSE 7456
+
+ENV OD_BIND_HOST=0.0.0.0
+
+CMD ["node", "apps/daemon/dist/cli.js", "--no-open", "--host", "0.0.0.0"]
```
mrcfps: This command line hard-codes `--host 0.0.0.0`, and the daemon CLI parses `--host` after reading `OD_BIND_HOST`, so the flag always wins over the environment (`apps/daemon/src/cli.ts:106-115`). As a result, changing `OD_BIND_HOST` in `docker-compose.yml` or following the compose comment's LAN guidance does not actually change the daemon's bind/origin host unless users also replace the container command. That is easy to miss and leaves the documented configuration knob ineffective.

Please drop the `--host` flag from the image command and rely on `OD_BIND_HOST`, or move the host value into the Compose `command`/documentation so users have one source of truth for the host they need to change.

mrcfps (changes requested):
@MrCx0810 thanks for the continued iteration on the Docker support — the earlier build-context, web bundle, and loopback exposure fixes are in much better shape now. I found one remaining default-container correctness blocker plus one related configuration issue to address before this can be a reliable Docker Desktop path. 🙏


docker-compose.yml
KatonJ: Switched to `network_mode: host`. This preserves real loopback peer addresses for `requireLocalDaemonRequest` routes instead of routing through Docker's bridge NAT. Works natively on Linux; on macOS requires Docker Desktop 4.34+.

Dockerfile
KatonJ: Removed `--host` from CMD. Now `OD_BIND_HOST` env var (set in docker-compose.yml) is the single source of truth — users can change it without overriding the container command.

MrCx0810 (commented):
@mrcfps thanks for the detailed catch! Both issues addressed:

1. **Docker NAT → host networking**: Switched to  in compose. The daemon now sees real loopback peer addresses, so  routes (live artifact preview, connectors) work correctly. Requires Docker Desktop 4.34+ on macOS, works natively on Linux.

2. **--host flag override**: Removed the hard-coded  flag from CMD.  env var is now the single source of truth, users can change it via compose environment without replacing the container command.

MrCx0810 (commented):
Thanks for the detailed catch @mrcfps! Both issues addressed:

1. Docker NAT -> host networking: Switched to network_mode: host in compose. The daemon now sees real loopback peer addresses, so requireLocalDaemonRequest routes work correctly. Requires Docker Desktop 4.34+ on macOS, works natively on Linux.

2. --host flag override: Removed the hard-coded --host flag from CMD. OD_BIND_HOST env var is now the single source of truth via compose environment.

Dockerfile
```diff
@@ -0,0 +1,63 @@
...
+COPY --from=builder --chown=oduser:oduser /app/prompt-templates ./prompt-templates
+COPY --from=builder --chown=oduser:oduser /app/templates ./templates
+COPY --from=builder --chown=oduser:oduser /app/craft ./craft
+COPY --from=builder --chown=oduser:oduser /app/assets ./assets
+
+USER oduser
+
+EXPOSE 7456
+
+ENV OD_BIND_HOST=0.0.0.0
```
mrcfps: The final image still defaults `OD_BIND_HOST` to `0.0.0.0`. Compose overrides it to `127.0.0.1`, but anyone running the image directly with common Docker patterns such as host networking or `-p 7456:7456` gets the unauthenticated local-first daemon bound to every interface by default. The daemon allows non-browser clients with no `Origin` header, so this can expose mutable API routes outside localhost.

Please make the image default loopback-safe (`OD_BIND_HOST=127.0.0.1`) and require an explicit opt-in for broad binding, e.g. `-e OD_BIND_HOST=0.0.0.0` only together with a loopback-only port publish or documented auth/reverse-proxy setup.

docker-compose.yml
```diff
@@ -0,0 +1,26 @@
...
+    # VM's network which may break those routes.
+    network_mode: host
+    volumes:
+      - od_data:/app/.od
+    restart: unless-stopped
+    environment:
+      - OD_PORT=7456
+      - OD_BIND_HOST=127.0.0.1
+    healthcheck:
+      test: ["CMD", "node", "-e", "require('http').get('http://localhost:7456/api/health', (r) => { process.exit(r.statusCode === 200 ? 0 : 1) })"]
```
mrcfps: The healthcheck probes `http://localhost:7456` while the compose environment binds the daemon to the IPv4 literal `127.0.0.1` on line 16. On hosts/images where `localhost` resolves to `::1` first, the healthcheck can mark the container unhealthy even though the daemon is listening correctly on IPv4.

Please make the probe match the configured bind host, e.g. `http://127.0.0.1:7456/api/health`, so Docker Desktop's health status is reliable across resolver configurations.

mrcfps (changes requested):
@MrCx0810 thanks for the quick iteration on the Docker support — the build context, web bundle, host-networking, and command override fixes are moving this in the right direction. I found one remaining safety blocker for the standalone image default, plus a small healthcheck correctness follow-up, before this is ready to merge. 🙏


Dockerfile
KatonJ: Changed default to \`127.0.0.1\`. The image is now loopback-safe by default; broad binding requires explicit \`-e OD_BIND_HOST=0.0.0.0\` opt-in.

docker-compose.yml
KatonJ: Changed to \`http://127.0.0.1:7456/api/health\`. Using the literal IPv4 address avoids resolver-dependent failures on hosts where localhost resolves to ::1 first.

MrCx0810 (commented):
Both issues fixed:

1. Default OD_BIND_HOST changed from 0.0.0.0 to 127.0.0.1 — image is loopback-safe by default, broad binding requires explicit opt-in.
2. Healthcheck URL uses 127.0.0.1 instead of localhost — avoids IPv4/IPv6 resolver mismatch.

docker-compose.yml
```diff
@@ -0,0 +1,29 @@
+services:
+  open-design:
+    build: .
+    container_name: open-design
+    # Host networking preserves loopback for local-only daemon routes
+    # (live artifact preview, connectors) that reject non-loopback peer
+    # addresses. On Linux this works natively. On macOS, Docker Desktop
+    # 4.34+ supports host networking; older versions fall back to the
+    # VM's network which may break those routes.
+    network_mode: host
```
mrcfps, line 10: The Compose default now depends on Docker Desktop host networking, but these changed lines only say macOS Docker Desktop 4.34+ supports it. Docker’s documented setup also requires users to explicitly enable host networking under Docker Desktop Settings → Resources → Network and restart Docker; on a default Desktop install where that option is still off, `docker compose up` with `network_mode: host` fails before the daemon starts. That breaks the PR’s advertised Docker Desktop one-click start/stop path even though the Dockerfile itself is now safer.

Please either make the default compose file work without requiring Desktop’s opt-in host-networking setting, or add the missing enablement prerequisite in a checked-in setup note/comment near `network_mode: host` so users know the exact Docker Desktop setting they must turn on before running this compose file.

mrcfps (changes requested):
@MrCx0810 thanks for the continued iteration here — the Dockerfile defaults and healthcheck fixes address the last security/healthcheck notes nicely. I found one remaining Docker Desktop setup issue in the compose default that can still break the advertised one-click path, so I’m requesting one more change before merge. 🙏


lefarcen:
Thanks for the Docker setup! We're going to consolidate Docker support around #65, which covers a similar surface but adds an Alpine base, persistent `/app/.od` volume, health checks, memory limits, `.env.example` defaults, and multi-arch publishing scripts — i.e. closer to a deploy-ready image rather than just a dev convenience.

Closing this in favor of #65. If you'd like to land any specific piece from your version on top of #65 (for example a simpler daemon-only build target), a follow-up against that branch once it merges would be very welcome.

PR #556, opened 2026-05-05
feat(web): add Cmd/Ctrl+P quick file switcher

whatiskadudoing:
## Summary

Adds a keyboard-driven file palette to the workspace — the muscle-memory **Cmd/Ctrl+P** that designers expect from VS Code, Sublime, IntelliJ, et al. Press it anywhere in a project, type to fuzzy-filter, ↑/↓ to navigate, Enter to open in a tab, Esc to dismiss.

With an empty query the palette surfaces **recents first** (per-project, scoped via `localStorage`) then fills with the rest of the file list sorted by `mtime`. As project file counts grow (`live-artifacts/`, sketches, generated HTML, etc.) this becomes much faster than reaching for `DesignFilesPanel`.

### Why
Today the only global keyboard chord in the web app is the slide-deck arrow nav. Every other action requires a click. This is the smallest, most universally-useful primitive to add — and it lays a clean foundation for future global shortcuts (close-tab, save-sketch, etc.) to hook into.

### Inspiration
Lifted the palette pattern from a sister project I work on — same component, redrawn against Open Design's design tokens (`--bg-panel`, `--accent`, `--accent-soft`, `--border-soft`) and global-class CSS conventions, and adapted to consume the `ProjectFile[]` prop that `FileWorkspace` already has rather than fetching.

## What's in the diff

**New files:**
- `apps/web/src/components/QuickSwitcher.tsx` — palette UI, scorer (prefix-on-name > substring-on-name > substring-on-path), keyboard handlers
- `apps/web/src/quickSwitcherRecents.ts` — per-project recents (key: `od:qs-recents:<projectId>`, cap 6)

**Modified:**
- `FileWorkspace.tsx` — capture-phase global keydown for Cmd/Ctrl+P (capture beats the browser's print dialog), mounts the palette
- `index.css` — `.qs-*` styles scoped under that prefix
- `i18n/types.ts` + 16 locale files — 6 new `quickSwitcher.*` keys (placeholder, empty, noMatches, navigate, open, close)

## i18n

Translated all 16 locales (en, de, pt-BR, es-ES, fr, zh-CN, zh-TW, ja, ko, ru, uk, ar, fa, pl, hu, tr) — `tests/i18n/locales.test.ts` passes. Native-speaker review on the non-Romance/CJK locales (especially `ar`, `fa`, `hu`, `pl`, `tr`, `uk`) would be welcome; the strings are short and idiomatic edits would be one-line touch-ups.

## Verification

- ✅ `pnpm --filter @open-design/web typecheck`
- ✅ `pnpm --filter @open-design/web test` (225/225 tests, including `i18n/locales.test.ts`)
- ✅ `pnpm --filter @open-design/web build`

## Test plan
- [ ] Open a project with several files; press Cmd+P (mac) / Ctrl+P (win/linux)
- [ ] Type a partial name; verify prefix-on-basename ranks above substring matches
- [ ] Empty the query; verify recents appear above mtime-sorted files
- [ ] Open a file; reopen the palette; verify it appears at the top of recents
- [ ] Press Esc; click outside; verify both dismiss
- [ ] ↑/↓ wrap correctly at list bounds; mouse-hover updates the cursor
- [ ] Repeat once in another locale (`Settings → Language`) to confirm strings render

## Notes
- No daemon / contract / API changes
- Pure addition; no existing behavior modified
- Keeping it small per [CONTRIBUTING.md](https://github.com/nexu-io/open-design/blob/main/CONTRIBUTING.md) — happy to split or follow up if you'd prefer different scoping


lefarcen:
Hi @whatiskadudoing! 🎉

Thanks for the contribution — a keyboard-driven file palette is exactly the kind of muscle-memory primitive that makes tools feel native to designers. The recents-first + fuzzy-filter approach you've described sounds solid.

— open-design team

mrcfps:
@whatiskadudoing I couldn't generate review comments for nexu-io/open-design#556 because this pull request currently has merge conflicts.

apps/web/src/index.css
```diff
@@ -10185,3 +10185,110 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
...
   text-align: right;
 }
+
+/* ============================================================
+   Quick Switcher (Cmd/Ctrl+P file palette)
+   ============================================================ */
+.qs-overlay {
+  position: fixed;
+  inset: 0;
+  background: rgba(13, 12, 10, 0.32);
```
lefarcen, line 10195: **P1** — z-index: 200 is too low. Context menus (df-row-popover, conv-menu) also use 200, and modal overlays use 1500-1600. If a context menu is open when user presses Cmd+P, the QuickSwitcher will appear behind the menu.

**Fix:** Change to `z-index: 1500;` to match other modal overlays.

apps/web/src/components/QuickSwitcher.tsx
```diff
@@ -0,0 +1,177 @@
...
+    [onOpenFile, onClose, projectId],
+  );
+
+  function onKeyDown(e: React.KeyboardEvent) {
+    if (e.key === 'Escape') {
+      onClose();
+      return;
+    }
+    if (e.key === 'ArrowDown') {
+      e.preventDefault();
```
lefarcen, line 91: **P2** — When `matches.length === 0` (no results), `matches.length - 1` becomes `-1`, so cursor can be set to `-1`. The highlight selector `[data-idx="${cursor}"]` won't match any row when results come back.

**Fix:** Guard both arrow handlers:
```tsx
if (e.key === 'ArrowDown' && matches.length > 0) {
  e.preventDefault();
  setCursor((c) => Math.min(c + 1, matches.length - 1));
  return;
}
if (e.key === 'ArrowUp' && matches.length > 0) {
  e.preventDefault();
  setCursor((c) => Math.max(c - 1, 0));
  return;
}
```

lefarcen (changes requested):
Hi @whatiskadudoing — great work on the keyboard-driven file palette! The muscle-memory Cmd/Ctrl+P is exactly the kind of primitive that makes tools feel native. I found three issues that need attention before this can ship:

**P1 — z-index too low (QuickSwitcher will appear behind context menus and other modals)**

The codebase uses z-index tiers:
- 1500-1600: modal overlays (PromptTemplatePreviewModal, etc.)
- 200: context menus (df-row-popover, conv-menu)
- 100: misc overlays

Your `.qs-overlay { z-index: 200 }` puts the QuickSwitcher at the same level as context menus. If a user has a context menu open and presses Cmd+P, the palette will appear behind the menu, breaking the UX.

**Fix:** Bump `.qs-overlay` z-index to at least 1500 (same tier as other modal overlays).

**P2 — Arrow key navigation breaks when matches array is empty**

In `QuickSwitcher.tsx:86-92`, the ArrowDown/ArrowUp handlers use:
```tsx
setCursor((c) => Math.min(c + 1, matches.length - 1));
setCursor((c) => Math.max(c - 1, 0));
```

When `matches.length === 0` (no results), `matches.length - 1` becomes `-1`, so `setCursor` can set cursor to `-1`. The next time you type and get results, the highlight selector `[data-idx="${cursor}"]` won't match any row (indexes are 0-based).

**Fix:** Guard the arrow key handlers — only call `setCursor` when `matches.length > 0`.

**P2 — Missing test coverage for core logic**

The PR adds 209 new lines of logic (fuzzy scoring, keyboard nav, localStorage recents), but no unit tests. The i18n test validates key presence, not component behavior. Key scenarios that should be tested:
- Fuzzy scoring: prefix-on-basename > substring-on-basename > substring-on-path
- Keyboard nav: ↑↓ wrap at bounds, Enter opens highlighted file
- LocalStorage: recents cap at 6, deduplicate on push, handle quota exceeded gracefully
- Edge cases: empty file list, empty query, all files filtered out

The author verified typecheck/build/existing tests pass, but those don't cover the new component.

**Suggestion:** Add a `QuickSwitcher.test.tsx` with at least 5-8 test cases covering the scenarios above. If you prefer to ship first and add tests later, open a follow-up issue and link it here.

---

**Minor observations (non-blocking):**
- i18n translations look solid across all 16 locales. Native-speaker review on ar/fa/hu/pl/tr/uk would still be welcome but not a blocker.
- IME composition check (`if (e.isComposing) return;`) is correct — prevents palette from triggering mid-Japanese/Chinese input.
- LocalStorage quota exceeded is correctly caught and silently dropped (recents are best-effort).

Once the P1 z-index and P2 arrow-key guard are fixed, this will be ready. Thanks for the clean PR!

whatiskadudoing:
Thanks for the careful review @lefarcen — all three points addressed in `361c88a`:

**P1 — z-index** ✅
Bumped `.qs-overlay` from `200` → `1500`. Verified your tier read against the codebase: `prompt-template-modal-overlay` is at 1500 and the lightbox backdrop at 1600, so 1500 puts the palette at the right level (above context menus, alongside other modal overlays).

**P2 — Arrow-key guard** ✅
Added an early return when `matches.length === 0`. As you noted, the user-visible bug was bounded (Enter on `matches[-1]` is already a no-op via the `if (hit)` check), but the cursor=-1 state was still messy and broke the highlight selector on the next render. Cleaner now.

**P2 — Tests** ✅
Added 19 unit tests across two files:

- `tests/components/QuickSwitcher.test.tsx` (9 tests) — `scoreMatch` tier ordering (exact > prefix > substring-on-base > substring-on-path > 0), case-insensitivity, empty/non-empty render, kbd hints, placeholder.
- `tests/quickSwitcherRecents.test.ts` (10 tests) — full lifecycle: cap at 6, dedupe-on-push, corrupt-JSON recovery, schema-drift filtering, per-project scoping, no-op on `setItem` quota error.

All run under your existing node environment (no jsdom dep) via a tiny in-memory `Storage` stub.

Verification:
- `pnpm --filter @open-design/web test` — **252/252** passing (was 233 before, +19 new)

Ready for another look whenever convenient.

lefarcen (commented):
Perfect! All three issues from the last review are now resolved:

✅ **P1 z-index fixed** — `.qs-overlay` now uses `z-index: 1500` (modal tier), so the palette will correctly appear above context menus.

✅ **P2 arrow-key guard fixed** — Arrow handlers now check `if (matches.length === 0) return;` before calling `setCursor`, preventing the cursor=-1 bug.

✅ **P2 test coverage added** — You added comprehensive tests:
- `QuickSwitcher.test.tsx`: 112 lines covering fuzzy scoring tiers, empty state, rendering, keyboard hints.
- `quickSwitcherRecents.test.ts`: 106 lines covering localStorage read/write, deduplication, cap-at-6, error handling, per-project scoping.

The tests hit all the scenarios I requested. Clean implementation, clean tests. This is ready to ship from my perspective — deferring final approval to a maintainer. Great work!

apps/web/src/components/QuickSwitcher.tsx
```diff
@@ -0,0 +1,183 @@
...
+      // would clamp the cursor to a non-existent row and break the
+      // [data-idx="${cursor}"] highlight on the next render with results.
+      if (matches.length === 0) return;
+      setCursor((c) => Math.min(c + 1, matches.length - 1));
+      return;
+    }
+    if (e.key === 'ArrowUp') {
+      e.preventDefault();
+      if (matches.length === 0) return;
+      setCursor((c) => Math.max(c - 1, 0));
```
mrcfps: **Blocking:** The arrow handlers still clamp at the list boundaries (`Math.min(..., matches.length - 1)` and `Math.max(..., 0)`), so pressing ↓ on the last row or ↑ on the first row does not wrap. The PR test plan explicitly asks to verify that `↑/↓ wrap correctly at list bounds`, and the earlier requested keyboard-nav coverage did not add an interaction test for this behavior, so the shipped shortcut would fail one of its advertised navigation paths. Please switch these updates to modulo wrapping after the empty-list guard, e.g. down `(c + 1) % matches.length` and up `(c - 1 + matches.length) % matches.length`, and add a component interaction test that proves both boundary wraps.

apps/web/src/components/FileWorkspace.tsx
```diff
@@ -241,6 +243,27 @@ export function FileWorkspace({
...
   }, []);
 
+  // Cmd/Ctrl+P opens the file palette. Capture phase so we beat the
+  // browser's default print dialog. Suppressed while another input has
+  // focus only when the user is mid-IME composition — otherwise the
+  // palette should be reachable from anywhere in the workspace, including
+  // from the chat composer (designers' muscle memory from VS Code).
+  useEffect(() => {
+    const onKeyDown = (e: KeyboardEvent) => {
+      if ((e.metaKey || e.ctrlKey) && !e.shiftKey && !e.altKey && e.key.toLowerCase() === 'p') {
```
mrcfps: **Non-blocking:** This captures either Cmd+P or Ctrl+P on every platform. The PR test plan scopes the shortcut as Cmd+P on macOS and Ctrl+P on Windows/Linux, and because this listener runs in capture phase even while an input/composer has focus, Ctrl+P on macOS will override the native/readline-style “previous line” editing shortcut in text fields. Consider gating by platform so macOS only handles `metaKey` and non-macOS only handles `ctrlKey` (for example, derive an `isMac` flag and require the opposite modifier to be false).

apps/web/src/components/QuickSwitcher.tsx
```diff
@@ -0,0 +1,183 @@
...
+      return;
+    }
+    if (e.key === 'Enter') {
+      e.preventDefault();
+      const hit = matches[cursor];
+      if (hit) open(hit);
+    }
```
mrcfps, line 106: **Non-blocking:** The global opener skips `e.isComposing`, but the palette's own key handler does not check composition before handling ArrowUp/ArrowDown/Enter. When users type CJK file names through an IME, those keys are used to navigate and commit IME candidates; this handler can instead move the quick-switcher cursor or open the highlighted file while composition is still active. Please add an early composition guard in this handler (for example, check `e.nativeEvent.isComposing` before the Arrow/Enter cases, or track composition state on the input) so IME candidate selection is left to the browser.

mrcfps (changes requested):
@whatiskadudoing thanks for the thoughtful quick-switcher work here — the recents store, styling tier fix, and added unit coverage are a solid foundation. I found one correctness issue in the advertised keyboard navigation behavior that should be fixed before merge, plus a couple of non-blocking shortcut/input polish items worth considering.



whatiskadudoing:
Thanks for the thorough re-review @mrcfps — addressed in `dd43b66`:

**Blocking — arrow-key wrap** ✅
You're right, the test plan promised wrap and the code clamped. Switched to modulo arithmetic via a new pure `nextCursor(current, total, direction)` helper, which let me unit-test the boundaries directly (forward wrap, backward wrap, mid-list moves, empty list, single-item). 6 new unit tests + verified live with Playwright (↓ from last → first row, ↑ from first → last row, both reproduce on a mocked-project harness).

**Non-blocking — IME guard in palette** ✅
Added `if (e.nativeEvent.isComposing) return;` at the top of the palette's `onKeyDown`, matching the existing guard on the global Cmd/Ctrl+P opener. CJK users can now select/commit IME candidates with ↑↓/Enter without those keys steering the cursor.

**Non-blocking — Cmd vs Ctrl platform gate** ✅
Reworked the global keydown to detect macOS via `navigator.platform` and gate accordingly: macOS handles only `metaKey && !ctrlKey`, non-macOS handles only `ctrlKey && !metaKey`. Native Ctrl+P readline ("previous line") in text fields is now preserved on macOS.

**Re: @chatgpt-codex iframe-focus comment**
Fair callout — when focus is inside the preview iframe, the parent `window` keydown listener doesn't fire, so Cmd+P from inside HTML/deck previews currently misses. Cleanest fix is a small keydown forwarder injected into the iframe via the existing `srcdoc.ts` bridge (the same channel that already postMessages slide state and comment targets). Would prefer to do that as a focused follow-up PR after this one lands so this PR stays scoped — happy to open it next if that's acceptable.

**Verification**
- typecheck ✅
- 258/258 tests passing (was 252)
- Live Playwright probe re-run: open / filter / wrap-fwd / wrap-back / Esc / toggle / click-to-open / z-index / recents — all pass

Ready for another look.

lefarcen (commented):
All three issues from @mrcfps's review are now resolved — verified on `dd43b66`:

**Blocking — arrow-key wrap** ✅
Switched from clamp (`Math.min/max`) to modulo arithmetic via the new `nextCursor()` helper. Forward wrap (last → first) and backward wrap (first → last) both work correctly. Added 6 unit tests covering mid-list moves, both boundary wraps, empty list, and single-item edge cases.

**Non-blocking — IME composition guard** ✅
Added `if (e.nativeEvent.isComposing) return;` at the top of the palette's `onKeyDown`, matching the guard on the global opener. CJK users can now navigate/commit IME candidates without the palette cursor interfering.

**Non-blocking — platform-specific shortcut** ✅
Reworked the global keydown to detect macOS (`/Mac|iPod|iPhone|iPad/.test(navigator.platform)`) and gate accordingly: macOS handles only `metaKey && !ctrlKey`, non-macOS handles only `ctrlKey && !metaKey`. Native Ctrl+P readline behavior preserved on macOS.

Ready to ship from bot perspective. mrcfps has not yet re-reviewed `dd43b66`, but the fixes directly address each of the three issues raised.

mrcfps (approved):
@whatiskadudoing I reviewed the current quick-switcher diff at `dd43b66`, including the FileWorkspace shortcut wiring, QuickSwitcher filtering/navigation/IME handling, recents storage, CSS z-index update, i18n keys, and the added focused tests. The earlier keyboard-wrap and input-polish concerns look addressed, and I did not find any remaining actionable correctness, safety, or maintainability issues in the changed ranges. Thanks for the careful follow-through on the review feedback — this is a nice, focused UX improvement. 🚀



lefarcen (approved):
All issues resolved on `dd43b66` — arrow-key wrap implemented with modulo + 6 unit tests, IME composition guard added, platform-specific shortcut gate added. Dismissing my earlier REQUEST_CHANGES. Ready to ship.

(mrcfps also APPROVED this head at 16:40Z.)

lefarcen (approved):
All issues from my earlier review (z-index, arrow-key guard, test coverage) are now resolved on dd43b66. mrcfps has also approved. Dismissing my prior CHANGES_REQUESTED — ready to merge.

lefarcen (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @whatiskadudoing for the contribution to open-design. Looking forward to the next one!

leilei926524-tech:
Hi @whatiskadudoing — welcome to Open Design contributors. 🎉

Your first merged PR (#556) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #454, opened 2026-05-04
release: Open Design 0.4.0 — Live artifacts + Composio, MCP server, Critique Theater, Linux desktop

lefarcen:
🎉 **`71 PRs` · `100+ contributors` · `2 days`** — Open Design 0.4.0 levels up across the board: a brand-new **live artifacts** platform with a Composio connector catalog, **MCP** integration with every major AI editor, the **first Linux desktop** release, **Critique Theater (Design Jury)** scoring drafts before they ship, plus a wave of preview / agent / i18n / security work. 🚀

## 🔥 Highlights

- ✨ **Live artifacts + Composio connector catalog.** Designs that don't go stale: artifacts react to live data, and the new catalog plugs into hundreds of SaaS tools through Composio. (#381) Thanks @mrcfps.
- 🤖 **MCP server (`od mcp`).** Stop screenshotting your designs into Cursor — Claude Code, Cursor, VS Code, Zed, Windsurf and any MCP-aware tool now read your Open Design files directly. (#399) Thanks @emilneander.
- 🎭 **Critique Theater (Design Jury) ships end-to-end.** No more "good enough" drafts: every artifact gets graded by five panelists on accessibility, brand fit, and craft, and only ships above 8/10. (#387, #481) Thanks @Nagendhra-web.
- 🐧 **Linux x64 AppImage.** Linux desktops finally get Open Design as a one-file installer — and it now runs cleanly in Nix, sandboxes, and other locked-down boxes. (#369, #411) Thanks @iuliandita and @GodTamIt.
- ⚡ **Live-reload + Tweaks-mode HTML previews.** Stop hitting reload — save a file and the preview refreshes itself; click any element to drop a tweak straight into the agent's context. (#384, #409, #513) Thanks @bankielewicz and @Martin-Atrin.
- 🪢 **Link code folder for agent context.** Your agent can finally see your design and your code repo at the same time — point it at a folder, no copy-paste. (#455) Thanks @encyc.
- 🤖 **Kilo CLI + DeepSeek TUI agent adapters.** Two more AI agents to pick from — Open Design now drives 13+ coding CLIs out of the box. (#439, #480) Thanks @Ajay-Satish-01 and @pftom.
[...]

PR #369, opened 2026-05-03
Add linux x64 AppImage to tools-pack and release workflows

iuliandita:
## Summary

Adds first-class Linux support to `tools-pack`, mirroring the existing mac and windows lanes, and wires the new lane into both `release-beta` and `release-stable` so releases ship a downloadable AppImage alongside the mac DMG and windows installer.

## What this adds

**`tools-pack linux <action>` lane** (`tools/pack/src/linux.ts`) — symmetric with `mac.ts` and `win.ts`:

- `tools-pack linux build` — native or `--containerized` build (Docker via `electronuserland/builder:base`, glibc 2.27 baseline)
- `tools-pack linux install` / `start` / `stop` / `logs` / `uninstall` / `cleanup` — full lifecycle, marker-validated, namespace-scoped
- AppImage launches via `--appimage-extract-and-run` for `.desktop` / icon integration
- Unit-tested helpers (`renderDesktopTemplate`, `matchesAppImageProcess`, `buildDockerArgs`) covering the trickier surfaces

**Linux release wiring** (`.github/workflows/release-{beta,stable}.yml`):

- New `build_linux` job in each workflow, modeled on `build_mac` / `build_win`
- `--containerized` build so the AppImage is portable across distros (not pinned to the runner's glibc 2.39)
- AppImage + `.sha256` uploaded to the immutable version release tag alongside mac/win

## Out of scope (deliberate)

- **AppImage auto-update feed (`latest-linux.yml`).** The linux electron-builder config has no `publish` block — generating the feed without wiring would point users at a feed that never updates. Separate follow-on.
- **Linux signing.** No signing path in `tools-pack` yet; the beta asset uses the `.unsigned` suffix matching the windows convention. Stable uses no suffix, also matching windows.
- **Other linux targets** (deb / rpm / snap / flatpak). AppImage covers the broadest distribution surface; additional formats can land incrementally.

## Release-time prerequisite

Adds Docker as a release-path dependency for the `build_linux` job. `ubuntu-latest` ships Docker preinstalled, so no extra runner setup is required.

## Test plan

- [x] `pnpm typecheck` — clean across workspaces
- [x] `pnpm test` — `linux.test.ts` unit tests pass (helpers covered: `renderDesktopTemplate`, `matchesAppImageProcess`, `buildDockerArgs`)
- [x] `pnpm check:residual-js` — no residual JS introduced
- [x] Local smoke tested on Linux desktop with `libfuse2`: build → install → start → stop → logs → uninstall → cleanup
[...]

PR #558, opened 2026-05-05
fix(tools-pack): replace corepack with npx in linux container build (unblocks 0.4.0)

lefarcen:
## Why

0.4.0 release-stable run [25386360556](https://github.com/nexu-io/open-design/actions/runs/25386360556) failed at the new `build_linux` job:

\`\`\`
bash: line 1: corepack: command not found
Error: docker build exited with code 127
\`\`\`

\`tools/pack/src/linux.ts\` invokes \`corepack pnpm ...\` inside the \`electronuserland/builder:base\` container. Although Node 16.10+ ships corepack, this image strips the corepack binary, so \`docker exec\`'s inner \`bash -lc\` fails before pnpm install ever starts. We also can't \`corepack enable\` ourselves — the container runs as the host's non-root uid and corepack would try to write shims next to the root-owned system Node binary.

The Linux build path was added in #369 and worked against an earlier electronuserland image; the corepack stripping appears to be a recent change in that base image.

## What this changes

Swap \`corepack pnpm\` for \`npx --yes pnpm@10.33.2\` in \`buildDockerArgs\` (the only place corepack is invoked inside the container):

\`\`\`diff
- "corepack pnpm tools-pack linux build"
+ \`\${pnpmCmd} tools-pack linux build\`
- "corepack pnpm install --frozen-lockfile && ..."
+ \`\${pnpmCmd} install --frozen-lockfile && ...\`
\`\`\`

Why \`npx --yes pnpm@<version>\`:
- \`npx\` ships with npm and is always present in the base image
- \`--yes\` skips the install confirmation prompt
- packages cache under \`\$HOME/.npm/_npx\`, which is writable by the unprivileged container user (no root required)
- the pinned version matches \`packageManager\` in the root \`package.json\`, preserving reproducibility

Net diff: **+16 / -8 lines** in a single file.

## Local verification

\`\`\`
pnpm install --frozen-lockfile  # ✅
pnpm --filter @open-design/daemon build  # ✅
pnpm --filter @open-design/desktop build  # ✅
pnpm --filter @open-design/web build:sidecar  # ✅
pnpm -r run typecheck  # ✅ exit 0
pnpm guard  # ✅ residual-JS + test-layout
\`\`\`

The actual fix (running pnpm via npx inside the container) can only be exercised by the docker run on a Linux GitHub Actions runner — the next \`release-stable\` dispatch validates it end-to-end. Atomic publish in release-stable.yml means a second failure here still leaves no orphan tag/release behind.

## After merge

Re-dispatch \`release-stable\` with \`mac_signed=true\` to publish 0.4.0. The 0.4.0 \`release: ...\` commit (#454) is already on main — only the workflow needs another run.

## Test plan

- [ ] CI \`Validate workspace\` goes green on this PR
- [ ] After merge, \`release-stable\` dispatch finishes all four build jobs (verify + mac + win + **linux**) and publishes \`open-design-v0.4.0\`

### DIFF
--- a/tools/pack/src/linux.ts
+++ b/tools/pack/src/linux.ts
@@ -96,14 +96,22 @@ export function buildDockerArgs(
   // None of these values can contain shell metacharacters, so direct
   // interpolation into the inner command string is safe.
   //
-  // We invoke pnpm via `corepack pnpm` instead of `corepack enable && pnpm ...`
-  // because the container runs as the host's non-root uid (--user above) and
-  // `corepack enable` writes shims next to the system Node binary, which is
-  // owned by root in electronuserland/builder:base and not writable by the
-  // unprivileged container user. `corepack pnpm` resolves and executes the
-  // pinned packageManager from package.json directly, no shims required.
+  // We can't rely on `corepack pnpm` here: although Node 16.10+ ships corepack,
+  // the `electronuserland/builder:base` image strips the corepack binary, so
+  // the inner `bash -lc` fails with `corepack: command not found`. We also
+  // can't `corepack enable` ourselves — the container runs as the host's
+  // non-root uid (--user above) and corepack would try to write shims next
+  // to the system Node binary, which is owned by root in this image.
+  //
+  // Use `npx --yes pnpm@<version>` instead: `npx` ships with npm (always
+  // present in the image), `--yes` skips the install confirmation, and the
+  // package gets cached under `$HOME/.npm/_npx`, which is writable by the
+  // unprivileged user. The pinned version matches the `packageManager`
+  // field in the root package.json so reproducibility is preserved.
+  const PNPM_VERSION = "10.33.2";
+  const pnpmCmd = `npx --yes pnpm@${PNPM_VERSION}`;
   const innerArgs = [
-    "corepack pnpm tools-pack linux build",
+    `${pnpmCmd} tools-pack linux build`,
     `--to ${config.to}`,
     `--namespace ${config.namespace}`,
     "--dir /tools-pack",
@@ -111,7 +119,7 @@ export function buildDockerArgs(
   if (config.portable) {
     innerArgs.push("--portable");
   }
-  const innerCommand = "corepack pnpm install --frozen-lockfile && " + innerArgs.join(" ");
+  const innerCommand = `${pnpmCmd} install --frozen-lockfile && ` + innerArgs.join(" ");
 
   return [
     "run",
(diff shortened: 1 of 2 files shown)


tools/pack/src/linux.ts
```diff
@@ -96,22 +96,30 @@ export function buildDockerArgs(
...
+  // can't `corepack enable` ourselves — the container runs as the host's
+  // non-root uid (--user above) and corepack would try to write shims next
+  // to the system Node binary, which is owned by root in this image.
+  //
+  // Use `npx --yes pnpm@<version>` instead: `npx` ships with npm (always
+  // present in the image), `--yes` skips the install confirmation, and the
+  // package gets cached under `$HOME/.npm/_npx`, which is writable by the
+  // unprivileged user. The pinned version matches the `packageManager`
+  // field in the root package.json so reproducibility is preserved.
+  const PNPM_VERSION = "10.33.2";
```
lefarcen, line 111: **P3 maintainability:** `PNPM_VERSION` duplicates the root `packageManager` value. Safe today (matches `pnpm@10.33.2`), but a future package manager bump could silently leave Linux container builds on a different pnpm. Consider deriving this from root `package.json` before building Docker args, or add a guard check that fails when the hardcoded value drifts.

lefarcen (commented):
LGTM — the npx replacement correctly fixes the corepack-missing failure that blocked 0.4.0.

**Command safety verified:** PNPM_VERSION is a hardcoded constant, and the rest of the shell-interpolated values are already constrained (no user input). npx failures will propagate and stop the chained build.

**One P3 maintainability note:** see inline.

mrcfps (approved):
@lefarcen I reviewed the Linux container build change: the inner Docker command now uses pinned `npx --yes pnpm@10.33.2` for both install and build, preserving the existing arguments and non-root container assumptions. CI's Validate workspace check is passing; my local targeted test attempt was blocked by missing node_modules in this checkout rather than a PR failure. Thanks for the focused fix—this should unblock the release-stable Linux job nicely. 🙌


tools/pack/src/linux.ts
lefarcen: Good catch — added a guard test in `tools/pack/tests/linux.test.ts` that reads `packageManager` from the root `package.json` and asserts the docker inner-command contains `npx --yes pnpm@<that-version>`. If the root `packageManager` field bumps but `PNPM_VERSION` in `linux.ts` does not, the test fails loudly. 25 tests pass locally.

PR #559, opened 2026-05-05
Fixes for fork

murphypharmacy:


lefarcen:
Hi @murphypharmacy! 🎉
Thanks for the contribution — I see you're addressing fork-related improvements across daemon server, file viewer, and export runtime.

— open-design team

apps/web/src/runtime/exports.ts
```diff
@@ -223,38 +231,32 @@ export function openSandboxedPreviewInNewTab(
...
-    doc = buildSandboxedPreviewDocument(doc, title, { allowModals: true });
-  }
   const blob = new Blob([doc], { type: 'text/html;charset=utf-8' });
   const url = URL.createObjectURL(blob);
-  const win = window.open(url, '_blank', sandboxedPreview ? 'noopener,noreferrer' : undefined);
-  if (!win) {
-    // Popup blocked — at least the tab navigation may have happened above.
-    // Nothing else we can do without a fresh user gesture.
-  }
+  window.open(url, '_blank');
```
mrcfps, line 241: `exportAsPdf` still receives arbitrary artifact HTML from callers such as `PreviewModal` and `ExamplesTab`, but this change removes the sandbox/noopener wrapper and opens the generated Blob URL with a live opener. Blob documents inherit the app origin, so any script in the exported artifact can now reach `window.opener` and same-origin app/daemon APIs while the user is exporting a PDF. That is a security regression from the previous default and also contradicts the existing export tests that assert the sandboxed noopener behavior. Please keep opener isolation for untrusted PDF exports — for example, restore `noopener,noreferrer` plus a safe print mechanism, or serve a dedicated print document that can trigger print without exposing `window.opener` to artifact code.

apps/web/src/components/FileViewer.tsx
```diff
@@ -2842,7 +2859,38 @@ function HtmlViewer({
...
+                      //
+                      // Append ?pagesize=WxH so the daemon injects a corrected
+                      // @page rule using mm units — Chrome then pre-selects the
+                      // right paper size instead of defaulting to A4.
+                      const pagesizeParam = extractPrintPageSize(source ?? '');
+                      const rawUrl = projectRawUrl(projectId, file.name);
+                      const exportUrl = pagesizeParam
+                        ? `${rawUrl}?pagesize=${pagesizeParam}`
+                        : rawUrl;
+                      const win = window.open(exportUrl, '_blank');
```
mrcfps, line 2868: Opening the daemon raw HTML URL without `noopener` gives the project document a live `window.opener` back to the Open Design app. Because this is same-origin `/api/projects/:id/raw/:file` content and project HTML can include arbitrary scripts, a generated artifact can read or mutate the parent app and call local daemon APIs as soon as the user chooses Export PDF; the same pattern also appears in `openInNewTab` above. Please preserve the raw-URL asset loading while isolating the opener, such as opening with `noopener,noreferrer` and moving the print trigger into a daemon-injected print mode, or otherwise sandboxing the artifact so it cannot access the parent window/local app surface.

mrcfps (changes requested):
@murphypharmacy Thanks for the contribution — the export fixes are moving the PDF/new-tab flows in the right direction, especially around relative assets and print sizing. I found a blocking opener-isolation regression that needs to be addressed before this can safely merge.


lefarcen (changes requested):
Hi @murphypharmacy — thanks for tackling the fork-related export improvements. I found **three critical security issues** related to sandbox escapes and one unit conversion bug that will break print sizing. These need fixes before merge:

## Summary

The PR removes the sandboxed-iframe wrapper from PDF export and opens project raw HTML directly. This **breaks the artifact isolation boundary** documented in `docs/architecture.md:230-237` — untrusted generated HTML now runs as same-origin with full access to `/api/*` routes, localStorage, and `window.opener`.

Additionally, the `@page` size extraction treats px values as mm without conversion, so a 1920×1080px deck will request a physically-impossible 1920mm × 1080mm paper size.

## Critical Issues (must fix)

### P1 Security: Same-origin artifact execution without sandbox
**`apps/web/src/runtime/exports.ts:241`**

`exportAsPdf` now opens untrusted generated HTML as a top-level same-origin Blob without `noopener` or the sandbox wrapper. Blob documents inherit the app origin, so artifact scripts can access same-origin storage/APIs and `window.opener`, which directly contradicts `docs/architecture.md:230-237`.

**Fix:** Restore the sandboxed default, or use a dedicated print route/response with `Content-Security-Policy: sandbox allow-scripts allow-modals` and `noopener,noreferrer`; make any unsandboxed path an explicit trusted-only opt-in.

### P1 Security: Direct raw HTML opens without sandbox (openInNewTab)
**`apps/web/src/components/FileViewer.tsx:2409`**

Opening project raw HTML directly in a new tab makes the artifact a first-party page with same-origin access to `/api/*` plus an opener reference. The previous `openSandboxedPreviewInNewTab` path preserved the artifact isolation contract while still allowing relative asset resolution via `baseHref`.

**Fix:** Keep a sandboxed wrapper with `baseHref`, or add a daemon-served preview/print URL that applies a CSP sandbox; also pass `noopener,noreferrer`.

### P1 Security: PDF export opens raw HTML without sandbox
**`apps/web/src/components/FileViewer.tsx:2878`**

Same issue as `openInNewTab` above — opening project raw HTML directly gives the artifact same-origin access and an opener reference, breaking the documented isolation boundary.

**Fix:** Same as above — restore sandboxed wrapper or add daemon CSP sandbox route.

### P1 Bug: px-to-mm conversion missing
**`apps/daemon/src/server.ts:2794` + `apps/web/src/components/FileViewer.tsx:3313`**

`extractPrintPageSize` reads `@page size: Wpx Hpx`, but the daemon injects those raw numbers as `Wmm Hmm` without conversion. A normal `1920px 1080px` deck becomes a 1920mm × 1080mm paper size (6.3 × 3.5 feet!) instead of approximately 508mm × 286mm.

**Fix:** Convert CSS px to mm using `px * 25.4 / 96`, or pass units explicitly and let the daemon validate/convert.

## P2 Correctness Issues (nice to fix)

### P2: CSS url() not detected by hasRelativeAssetRefs
**`apps/web/src/components/FileViewer.tsx:3297`**

The new CSS `url(...)` inlining path is gated by `hasRelativeAssetRefs`, but that predicate only detects `src` and `href`. HTML that only references assets from `<style>` blocks or inline `style="background:url(...)"` will never call `inlineRelativeAssets`, so the export won't be self-contained.

**Fix:** Include CSS `url(...)` in `hasRelativeAssetRefs`, or call the inliner for HTML export unconditionally.

### P2: CSS url() resolved relative to wrong base
**`apps/web/src/components/FileViewer.tsx:3331` + `3410`**

URLs inside inlined stylesheets are resolved relative to the HTML file, not relative to the stylesheet file. For `assets/css/app.css` containing `url("../fonts/a.woff2")`, the current code fetches `fonts/a.woff2` instead of `assets/fonts/a.woff2`.

**Fix:** Track the stylesheet's resolved path as the base when collecting CSS URLs, or rewrite `url(...)` refs while processing each `<link rel="stylesheet">`.

## Next Steps

Please either:
1. Restore the sandboxed wrapper with `baseHref` for relative asset resolution (preserves security), OR
2. Add a dedicated daemon print route with `Content-Security-Policy: sandbox allow-scripts allow-modals` + `noopener,noreferrer`

Happy to discuss alternatives if you have a different approach in mind!

lefarcen (changes requested):
Hi @murphypharmacy — thanks for the iteration! I can see you've made significant improvements:

✅ **Fixed**: px-to-mm conversion — the daemon now injects corrected `@page` rules with proper sizing
✅ **Fixed**: CSS `url()` and `` inlining — exports are now self-contained

However, **the core P1 security issues remain unaddressed**. The PR still removes the sandboxed wrapper and opens daemon raw HTML URLs without opener isolation, which breaks the artifact security boundary documented in `docs/architecture.md:230-237`.

## Remaining P1 Security Issues

### `openInNewTab()` — no `noopener`, no sandbox
**apps/web/src/components/FileViewer.tsx:2409**

This now opens `projectRawUrl(...)` directly with `window.open(..., '_blank')` — **no `noopener` flag**. Because daemon raw URLs are same-origin (`/api/projects/:id/raw/:file`), artifact scripts can access `window.opener`, reach back to the app, and call daemon APIs. This contradicts the isolation contract.

**Fix**: Either restore the sandboxed wrapper with `baseHref`, or add `noopener,noreferrer`: `window.open(..., '_blank', 'noopener,noreferrer')`.

### PDF export — opens raw URL without `noopener`
**apps/web/src/components/FileViewer.tsx:2905**

The new PDF export flow opens the daemon raw URL with `window.open(..., '_blank')` — again, **no `noopener`**. Same issue: artifact gets `window.opener` and same-origin access.

**Fix**: Add `noopener,noreferrer` to the `window.open` call on line 2905.

### `exportAsPdf` — sandboxing removed entirely
**apps/web/src/runtime/exports.ts:241**

The PR removed the sandboxed preview logic entirely. Opening Blob URLs without `noopener` gives artifact code same-origin access. While the new daemon-pagesize flow is clever, it doesn't address opener isolation.

**Fix**: Add `noopener,noreferrer` to the `window.open` call: `window.open(url, '_blank', 'noopener,noreferrer')`.

## Why This Matters

The documented security model (`docs/architecture.md:230-237`) relies on artifacts being isolated from the app's origin. Without `noopener` or sandboxing, a malicious artifact can:
- Call `window.opener.postMessage()` to inject commands into the app
- Access `window.opener.localStorage`
- Make same-origin XHR to `/api/*` daemon routes

Adding `noopener,noreferrer` is a one-line fix per call site and preserves the relative asset loading you've built.

## Next Steps

Please add `noopener,noreferrer` to all three `window.open` call sites above. The rest of the PR (pagesize injection, inlining) is solid — this is the last security gate before merge.

apps/web/src/runtime/exports.ts
mrcfps, line 241: This `exportAsPdf` path still opens artifact HTML directly as an app-origin Blob and no longer uses either the sandboxed wrapper or `noopener,noreferrer`. The changed line means arbitrary scripts inside exported HTML can execute with the Blob document's inherited app origin, and the existing runtime tests still assert the secure sandbox/noopener default for PDF exports. That is a blocking regression for callers such as `PreviewModal` and `ExamplesTab`, which pass generated artifact HTML into this helper. Please keep a safe default here: restore the sandboxed print wrapper, or serve a dedicated print document/route with CSP sandboxing and opener isolation, with any unsandboxed behavior limited to an explicit trusted-only opt-in.

apps/web/src/components/FileViewer.tsx
```diff
@@ -2402,11 +2402,28 @@ function HtmlViewer({
...
-    openSandboxedPreviewInNewTab(source, exportTitle, {
-      deck: effectiveDeck,
-      baseHref: projectRawUrl(projectId, baseDirFor(file.name)),
-      initialSlideIndex: htmlPreviewSlideState.get(previewStateKey)?.active ?? 0,
-    });
+    // Open via the daemon's raw URL so Babel can fetch <script type="text/babel"
+    // src="..."> sources, relative <link>/<script>/<img> references resolve,
+    // and any same-origin XHR works. The previously-used sandboxed-iframe
+    // wrapper broke designs that compile JSX in the browser.
+    window.open(projectRawUrl(projectId, file.name), '_blank');
```
mrcfps, line 2409: `openInNewTab()` now opens the daemon raw project HTML URL as a first-party tab without sandboxing or `noopener,noreferrer`. Because `/api/projects/:id/raw/...` serves project-controlled HTML from the same app/daemon origin, scripts in that artifact can keep a live `window.opener` reference and also call local same-origin APIs, which breaks the artifact isolation boundary that the previous sandboxed preview path provided. Please preserve relative asset loading without exposing the parent app, for example by using a sandboxed preview/daemon preview route with an appropriate `baseHref`/CSP sandbox and opener isolation.

apps/web/src/components/FileViewer.tsx
```diff
@@ -2842,7 +2859,45 @@ function HtmlViewer({
...
+                      // (e.g. Babel pulling in <script type="text/babel"
+                      // src="...jsx"> sources) don't apply. Then auto-trigger
+                      // the print dialog after a generous delay so React /
+                      // Babel / fonts / images all settle first.
+                      //
+                      // Open the new tab synchronously (popup blockers reject
+                      // window.open inside an awaited promise), then load the
+                      // real URL once we've detected the page size from the
+                      // HTML source or any linked CSS files.
+                      const win = window.open('about:blank', '_blank');
```
mrcfps: The PDF export flow opens an unsandboxed `about:blank` tab, keeps the returned window handle, and then navigates it to same-origin project HTML so the parent can call `win.print()`. That preserves a live opener relationship for arbitrary artifact scripts on the raw page, so choosing Export PDF can give generated HTML access back into the Open Design app and daemon API surface. Please move the print trigger into a trusted daemon-served print mode or sandboxed wrapper and open the artifact with opener isolation instead of requiring the parent window to retain control of the child.

mrcfps (changes requested):
@murphypharmacy Thanks for continuing to iterate on the export and fork fixes — the sidecar shim and asset/page-size work are helpful. I still found blocking opener/origin isolation regressions in the changed export paths that need to be addressed before this is safe to merge.


lefarcen (changes requested):
Hi @murphypharmacy — thanks for the quick iteration! The page-size fix looks solid, but the **P1 security boundary is still broken**. Switching from Blob to daemon raw URLs doesn't restore the isolation — untrusted artifacts still run as same-origin documents with opener access.

## Status

✅ **P1 Bug fixed**: The px-to-mm conversion is now handled correctly via `?pagesize=WxH` injection.

❌ **P1 Security NOT fixed**: Direct raw URLs still violate the sandbox boundary from `docs/architecture.md:230-237`.

❌ **P2 Correctness partially addressed**: `inlineRelativeAssets` now handles CSS `url()` and ``, but it's not wired into `getExportSource()` detection.

## Critical Security Issues (still blocking)

The core problem: **`projectRawUrl()` serves same-origin HTML at `/api/projects/:id/raw/:file`**, so artifact scripts can access app-origin storage, APIs, and `window.opener`. The previous sandboxed-iframe wrapper created an **opaque origin boundary** — the artifact couldn't reach outside. Removing it breaks that contract.

### Specific files with the issue:

1. **`apps/web/src/components/FileViewer.tsx:2409`** (openInNewTab): Opens raw daemon URL directly without sandbox or `noopener`. Artifact gets same-origin + opener access.

2. **`apps/web/src/components/FileViewer.tsx:2896`** (PDF export): Opens raw URL in popup with opener. Artifact executes with same-origin privileges before print trigger.

3. **`apps/web/src/runtime/exports.ts:241`** (`exportAsPdf`): Still opens unsandboxed Blob without `noopener,noreferrer`. PreviewModal/ExamplesTab use this path.

### What needs to happen

Either:
1. **Restore the sandboxed wrapper** with `baseHref` for relative asset resolution (this was working before), OR
2. **Add a dedicated daemon print route** that serves with `Content-Security-Policy: sandbox allow-scripts allow-modals` and opens with `noopener,noreferrer`

The sandbox doesn't block `window.print()` when you use `allow-modals`. The blank-print issue happens when the print trigger runs from the **parent** iframe, not when the artifact itself calls print.

## P2 Issues

1. **CSS url() detection** (`FileViewer.tsx:3304`): `getExportSource()` only inlines when `hasRelativeAssetRefs()` sees `src`/`href`. Artifacts with only `background:url(...)` skip the inliner.

2. **CSS URL base resolution** (`FileViewer.tsx:3446`): URLs from linked stylesheets are resolved relative to HTML file, not stylesheet file. `<link href="css/site.css">` with `url('../img/bg.png')` fetches from HTML directory instead of `css/`.

## Next Steps

Let me know which approach you prefer — happy to pair on the sandbox restoration if that helps!

lefarcen (changes requested):
Hi @murphypharmacy — thanks for the continued iteration! The page-size detection work and asset inlining are solid improvements. However, **the P1 security boundary is still broken** — all three `window.open` call sites still lack opener isolation.

## What's Still Missing

The **noopener,noreferrer** flags on:

1. **FileViewer.tsx:2409** (`openInNewTab`) — `window.open(projectRawUrl(...), '_blank')` 
2. **FileViewer.tsx:2868** (PDF export) — `window.open(exportUrl, '_blank')`
3. **exports.ts:241** (`exportAsPdf`) — `window.open(url, '_blank')`

## Why It Matters

All three paths open same-origin HTML (daemon `/api/projects/:id/raw/...` or app-origin Blob URLs). Without `noopener,noreferrer`, artifact scripts can:
- Access `window.opener` → read/write parent app state
- Call same-origin APIs → leak localStorage, session tokens, daemon routes

This violates the documented artifact isolation boundary from `docs/architecture.md:230-237`.

## The Fix (One-Line Each)

```typescript
// FileViewer.tsx:2409
window.open(projectRawUrl(projectId, file.name), '_blank', 'noopener,noreferrer');

// FileViewer.tsx:2868
const win = window.open(exportUrl, '_blank', 'noopener,noreferrer');

// exports.ts:241
window.open(url, '_blank', 'noopener,noreferrer');
```

The third argument (`'noopener,noreferrer'`) isolates the opened window from `window.opener` access. This keeps relative asset loading working while preventing artifact code from reaching back into the parent app.

## Everything Else Looks Good

✅ Page-size detection (daemon `?print=1` + CSS scanning)
✅ Asset inlining for exports
✅ Sidecar shim for tools-dev compatibility

Once the three `noopener,noreferrer` flags are added, this is ready to ship!

apps/web/src/runtime/exports.ts
mrcfps, line 241: This changed `exportAsPdf` path still opens the generated artifact Blob with `window.open(url, '_blank')`, so the new document keeps a live opener and inherits the app origin. Because callers pass project/generated HTML into this helper and that HTML may contain scripts, those scripts can reach `window.opener` and same-origin app/daemon APIs while the user is exporting. Please keep the direct-print behavior without exposing the parent window, for example by restoring `noopener,noreferrer` here and/or moving the print trigger into an isolated/sandboxed print document that does not require opener access.

apps/web/src/components/FileViewer.tsx
```diff
@@ -2842,7 +2859,19 @@ function HtmlViewer({
...
                     onClick={() => {
                       setShareMenuOpen(false);
-                      exportAsPdf(source ?? '', exportTitle, { deck: effectiveDeck });
+                      // Open the raw daemon URL with ?print=1. The daemon
+                      // auto-detects @page { size: Xpx Ypx } from the HTML
+                      // and any linked CSS files, converts px→mm, and injects
+                      // a corrected @page rule so Chrome prints the right
+                      // physical size instead of treating px as screen pixels.
+                      const exportUrl = `${projectRawUrl(projectId, file.name)}?print=1`;
+                      const win = window.open(exportUrl, '_blank');
```
mrcfps, line 2868: The PDF export button now opens `/api/projects/:id/raw/:file?print=1` as a first-party tab and retains the returned window handle so the parent can call `win.print()`. That means project-controlled HTML runs same-origin with a live `window.opener`, and the same raw-open pattern remains in `openInNewTab()` at line 2409, so arbitrary artifact scripts can reach back into the Open Design app and local daemon API surface. Please preserve relative asset loading through an isolated preview/print route (for example CSP sandbox or a sandboxed wrapper with `allow-modals`) and open it with opener isolation instead of relying on parent-window control of the artifact tab.

apps/daemon/src/server.ts
```diff
@@ -2790,6 +2790,71 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+          pageSize = pxPageSize(html);
+          if (!pageSize) {
+            const linkRe = /<link\b[^>]*\brel\s*=\s*["']stylesheet["'][^>]*>/gi;
+            const hrefRe = /\bhref\s*=\s*["']([^"']+)["']/i;
+            for (const tag of html.match(linkRe) ?? []) {
+              const hm = hrefRe.exec(tag);
+              const href = hm?.[1]?.trim();
+              if (!href || /^(?:https?:|data:|blob:|\/)/i.test(href)) continue;
+              const htmlDir = relPath.includes('/') ? relPath.slice(0, relPath.lastIndexOf('/') + 1) : '';
+              const cssPath = href.startsWith('../') ? href.replace(/^(\.\.\/)+/, '') : htmlDir + href;
```
mrcfps, line 2833: The server-side linked stylesheet scan is too narrow for the `?print=1` flow it introduces. The regex only matches tags where `rel="stylesheet"` appears before `href` and exactly equals `stylesheet`, and the `href.startsWith('../')` branch strips parent segments without applying the HTML file's directory. A common artifact like `pages/index.html` with `<link href="../print.css" rel="stylesheet">` or `<link href="css/print.css?v=1" rel="stylesheet">` will therefore skip or misread the CSS, miss the `@page` rule, and still fall back to the wrong print size. Please iterate over all `<link>` tags, parse `rel`/`href` order-independently, require `stylesheet` as one rel token, resolve `href` with URL/path normalization relative to `relPath`, strip search/hash before `readProjectFile`, and keep the existing safe project-file validation as the final guard.

mrcfps (changes requested):
@murphypharmacy Thanks for continuing to push on the fork/export fixes — the sidecar shim and print-size work are useful progress. I still found blocking isolation and linked-CSS detection issues in the changed export paths that need another pass before this is safe to merge.


lefarcen (changes requested):
Hi @murphypharmacy — thanks for the continued work! The page-size detection and asset inlining improvements are excellent progress. However, **the P1 security boundary is still broken** — all three `window.open` call sites still lack opener isolation.

## What's Still Missing

The **noopener,noreferrer** flags on:

1. **FileViewer.tsx:2409** (`openInNewTab`) — `window.open(projectRawUrl(...), '_blank')` 
2. **FileViewer.tsx:2868** (PDF export) — `window.open(exportUrl, '_blank')`
3. **exports.ts:241** (`exportAsPdf`) — `window.open(url, '_blank')`

## Why It Matters

All three paths open same-origin HTML (daemon `/api/projects/:id/raw/...` or app-origin Blob URLs). Without `noopener,noreferrer`, artifact scripts can:
- Access `window.opener` → read/write parent app state
- Call same-origin APIs → leak localStorage, session tokens, daemon routes

This violates the documented artifact isolation boundary from `docs/architecture.md:230-237`.

## The Fix (One-Line Each)

```typescript
// FileViewer.tsx:2409
window.open(projectRawUrl(projectId, file.name), '_blank', 'noopener,noreferrer');

// FileViewer.tsx:2868
const win = window.open(exportUrl, '_blank', 'noopener,noreferrer');

// exports.ts:241
window.open(url, '_blank', 'noopener,noreferrer');
```

The third argument (`'noopener,noreferrer'`) isolates the opened window from `window.opener` access. This keeps relative asset loading working while preventing artifact code from reaching back into the parent app.

## Everything Else Looks Good

✅ Page-size detection (daemon `?print=1` + CSS scanning)
✅ Asset inlining for exports
✅ Sidecar shim for tools-dev compatibility

Once the three `noopener,noreferrer` flags are added, this is ready to ship!

lefarcen (changes requested):
Hi @murphypharmacy — the page-size detection and asset inlining work is excellent! However, **the core P1 security issue remains**: all three `window.open` call sites still lack opener isolation.

## The Fix (Three One-Line Changes)

Add the third argument `'noopener,noreferrer'` to these three `window.open` calls:

```typescript
// FileViewer.tsx:2409
window.open(projectRawUrl(projectId, file.name), '_blank', 'noopener,noreferrer');

// FileViewer.tsx:2868
window.open(exportUrl, '_blank', 'noopener,noreferrer');

// exports.ts:241
window.open(url, '_blank', 'noopener,noreferrer');
```

## Why This Matters

Without `noopener,noreferrer`, artifact scripts can access `window.opener` and call same-origin daemon/app APIs, which breaks the documented isolation boundary from `docs/architecture.md:230-237`. This is the ONLY remaining blocker — once these three flags are added, this is ready to ship!

apps/web/src/runtime/exports.ts
mrcfps, line 241: `exportAsPdf` still opens generated artifact HTML as an app-origin Blob without either the sandboxed wrapper or `noopener,noreferrer`. Callers such as `PreviewModal` and `ExamplesTab` pass project/generated HTML into this helper, so any script in that HTML can keep `window.opener` and reach back into the app/daemon surface during PDF export; the existing runtime tests also still assert the secure noopener/sandbox default for this path. Please keep direct printing from exposing the parent window, for example by restoring opener isolation and a sandbox/CSP-backed print document, or by making any unsandboxed behavior an explicit trusted-only opt-in with matching tests updated for the intended security model.

apps/web/src/components/FileViewer.tsx
mrcfps, line 2409: `openInNewTab()` now opens project-controlled daemon raw HTML as a first-party tab with `window.open(..., '_blank')`. Because `/api/projects/:id/raw/...` is same-origin and the artifact can run arbitrary scripts, this gives the opened document both local API reachability and a live opener back to the Open Design app, which is the isolation boundary the previous sandboxed preview path protected. Please keep the relative-asset/raw-URL behavior without exposing the parent app, such as by opening an isolated preview route with a CSP sandbox or at minimum using `noopener,noreferrer` and avoiding any parent-window control path.

apps/web/src/components/FileViewer.tsx
mrcfps, line 2868: The PDF export action now opens raw project HTML with a retained `win` handle so the parent can set `win.document.title` and call `win.print()`. That keeps the opener relationship alive for arbitrary artifact scripts on the raw same-origin page, and it also bypasses the previous `exportAsPdf(..., { deck: effectiveDeck })` path that injected deck print CSS for decks missing their own print rules. Please move the print/title trigger into an isolated print document or daemon-served print mode, open it with opener isolation, and preserve the deck print stylesheet behavior for the “export all slides” case.

apps/daemon/src/server.ts
```diff
@@ -2790,6 +2790,84 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
```
mrcfps, line 2833: The server-side linked stylesheet scan for `?print=1` only matches tags where `rel="stylesheet"` appears before `href` and exactly equals that value, then resolves `../` by stripping parent segments instead of normalizing relative to the HTML file. Common valid markup like `<link href="../print.css" rel="stylesheet">`, multi-token rels, query/hash suffixes, or nested `pages/index.html` references will be skipped or read from the wrong project path, so the new print mode silently misses the `@page` CSS and falls back to the wrong paper size. Please parse `link` attributes order-independently, require `stylesheet` as a rel token, strip search/hash, and normalize the href against `relPath` before passing the validated project path to `readProjectFile`.

mrcfps (changes requested):
@murphypharmacy Thanks for continuing to iterate here — the sidecar compatibility shim and the print/page-size direction are helpful progress. I found a few blocking issues in the changed export/print paths that still need to be addressed before this is safe to merge. 🚧


mrcfps:
Hi @murphypharmacy, thanks for putting time into this and for continuing to iterate on the export/print fixes. There are useful pieces here — especially the sidecar compatibility shim, asset inlining work, and the print-size direction.

That said, this PR has grown to cover several separate concerns at once: daemon sidecar compatibility, export asset inlining, PDF/page-size behavior, and changes to the artifact preview/print isolation model. The last part is especially sensitive because opening project HTML/raw URLs directly changes the security boundary around untrusted artifacts.

To keep review and merging safe, I’m going to close this PR rather than continue iterating on the combined branch. If you’re still interested, the best next step would be to split this into smaller PRs, for example:

1. sidecar entry compatibility shim only
2. export asset inlining fixes, with tests and no isolation-model changes
3. print/page-size handling, with daemon-side path/CSS parsing tests
4. any artifact preview/print sandbox/raw-URL model change as a separate focused proposal/PR

Happy to review smaller follow-up PRs. Thanks again for the contribution.

PR #560, opened 2026-05-05
fix(tools-pack): fetch pnpm standalone binary in linux container (#558 follow-up)

lefarcen:
## Why

Follow-up to #558. After replacing \`corepack pnpm\` with \`npx --yes pnpm@<version>\`, the next release-stable dispatch ([25387708738](https://github.com/nexu-io/open-design/actions/runs/25387708738)) revealed:

\`\`\`
bash: line 1: npx: command not found
Error: docker build exited with code 127
\`\`\`

\`electronuserland/builder:base\` is even more minimal than expected — it ships a Node binary but strips **npm, npx, AND corepack** from PATH. Every "ask the image to invoke pnpm" path fails the same way.

## What this changes

Skip the image's npm tooling entirely. Download the official **pnpm-linuxstatic-x64** standalone binary at container start:

- Self-contained: bundles its own Node runtime, doesn't depend on anything in the image
- Writes to \`/tmp/pnpm\` (tmpfs, writable by the unprivileged container user)
- Pinned version matches root \`packageManager\` (guard test enforces lockstep)

\`\`\`diff
- const pnpmCmd = \`npx --yes pnpm@\${PNPM_VERSION}\`;
- const innerCommand = \`\${pnpmCmd} install --frozen-lockfile && \` + innerArgs.join(" ");
+ const pnpmDownloadUrl = \`https://github.com/pnpm/pnpm/releases/download/v\${PNPM_VERSION}/pnpm-linuxstatic-x64\`;
+ const setupPnpm = \`curl -fsSL "\${pnpmDownloadUrl}" -o /tmp/pnpm && chmod +x /tmp/pnpm\`;
+ const pnpmCmd = "/tmp/pnpm";
+ const innerCommand = \`\${setupPnpm} && \${pnpmCmd} install --frozen-lockfile && \` + innerArgs.join(" ");
\`\`\`

Inner command shape inside the container:
\`\`\`
curl -fsSL https://github.com/pnpm/pnpm/releases/download/v10.33.2/pnpm-linuxstatic-x64 -o /tmp/pnpm \\
  && chmod +x /tmp/pnpm \\
  && /tmp/pnpm install --frozen-lockfile \\
  && /tmp/pnpm tools-pack linux build --to all --namespace ... --dir /tools-pack
\`\`\`

\`curl\` is one of the few tooling utilities consistently present in \`electronuserland/builder:base\` (used by electron-builder itself for downloads).

## Tests updated

\`tools/pack/tests/linux.test.ts\`:
- "re-invokes pnpm tools-pack linux build" — assert curl + chmod + /tmp/pnpm
- "fetches pnpm standalone binary instead of npm tooling" — assert no \`corepack\`, \`npx\`, or \`npm \` appears in the inner command (sentinel for image regressions)
- "PNPM_VERSION lockstep with root packageManager" — guard test asserts the download URL embeds the same version as \`packageManager\` in the root \`package.json\`

Local: \`pnpm --filter @open-design/tools-pack run test\` → 25 passed.

## After merge

Re-dispatch \`release-stable\` with \`mac_signed=true\`. The 0.4.0 release commit (#454) is already on main; only the workflow needs another run. Atomic publish guarantees no orphan tag/release if Linux still misbehaves.

## Test plan

- [ ] CI \`Validate workspace\` goes green
- [ ] After merge, \`release-stable\` dispatch finishes all four build jobs (verify + mac + win + **linux**) and publishes \`open-design-v0.4.0\`

### DIFF
--- a/tools/pack/src/linux.ts
+++ b/tools/pack/src/linux.ts
@@ -96,20 +96,45 @@ export function buildDockerArgs(
   // None of these values can contain shell metacharacters, so direct
   // interpolation into the inner command string is safe.
   //
-  // We can't rely on `corepack pnpm` here: although Node 16.10+ ships corepack,
-  // the `electronuserland/builder:base` image strips the corepack binary, so
-  // the inner `bash -lc` fails with `corepack: command not found`. We also
-  // can't `corepack enable` ourselves — the container runs as the host's
-  // non-root uid (--user above) and corepack would try to write shims next
-  // to the system Node binary, which is owned by root in this image.
+  // The `electronuserland/builder:base` image is intentionally minimal: it
+  // ships a Node binary but strips npm, npx, *and* corepack from PATH. So
+  // every "ask the image to invoke pnpm" path fails with `command not found`.
   //
-  // Use `npx --yes pnpm@<version>` instead: `npx` ships with npm (always
-  // present in the image), `--yes` skips the install confirmation, and the
-  // package gets cached under `$HOME/.npm/_npx`, which is writable by the
-  // unprivileged user. The pinned version matches the `packageManager`
-  // field in the root package.json so reproducibility is preserved.
+  // Fix: download the official pnpm `linuxstatic-<arch>` standalone binary
+  // at container start, picking the asset by `uname -m` so the same workflow
+  // works on amd64 (GitHub Actions ubuntu-latest) and arm64 (Apple Silicon
+  // hosts pulling an arm64 builder image). The binary is self-contained
+  // (bundles its own Node runtime), so it doesn't depend on the image's
+  // npm tooling at all. Stage it under `/tmp/pnpm` for this container
+  // invocation, which is writable by the unprivileged container user.
+  // We rely on `curl`, which electron-builder itself uses for downloads and
+  // is consistently present in this image; the shell guard keeps failures
+  // explicit if that image contract changes.
+  //
+  // The pinned version matches the `packageManager` field in the root
+  // package.json so reproducibility is preserved (a guard test in
+  // tools/pack/tests/linux.test.ts asserts the lockstep).
   const PNPM_VERSION = "10.33.2";
-  const pnpmCmd = `npx --yes pnpm@${PNPM_VERSION}`;
+  const pnpmLinuxStaticX64Sha256 = "a47be715939bafa420fbdc5e34f7f9d8292c032402162c89ccb611e944e526d6";
+  const pnpmLinuxStaticArm64Sha256 = "4d402d0ef12cdc4d81ca339904e68638d841f4e27c73e460534d06e6b56048a9";
+  // Pick the right pnpm binary based on the container's CPU. Docker may run
+  // either an amd64 or arm64 `electronuserland/builder:base` (Apple Silicon
+  // hosts pull arm64 by default), so a hardcoded `-x64` asset would fail
+  // with "Exec format error" on ARM. Detect inside the container with
+  // `uname -m` and fall back to a clear error for unsupported arches.
+  const pnpmReleaseUrl = `https://github.com/pnpm/pnpm/releases/download/v${PNPM_VERSION}`;
+  const setupPnpm =
+    `command -v curl >/dev/null || { echo "curl not found in container image" >&2; exit 127; } && ` +
+    `case "$(uname -m)" in ` +
+    `x86_64) PNPM_ASSET=pnpm-linuxstatic-x64; PNPM_SHA256=${pnpmLinuxStaticX64Sha256} ;; ` +
+    `aarch64) PNPM_ASSET=pnpm-linuxstatic-arm64; PNPM_SHA256=${pnpmLinuxStaticArm64Sha256} ;; ` +
+    `*) echo "unsupported container arch: $(uname -m)" >&2; exit 1 ;; ` +
+    `esac && ` +
+    `curl --retry 3 --retry-all-errors --connect-timeout 10 --max-time 60 -fsSL "${pnpmReleaseUrl}/$PNPM_ASSET" -o /tmp/pnpm.tmp && ` +
+    `echo "$PNPM_SHA256  /tmp/pnpm.tmp" | sha256sum -c - && ` +
+    `mv /tmp/pnpm.tmp /tmp/pnpm && ` +
+    `chmod +x /tmp/pnpm`;
+  const pnpmCmd = "/tmp/pnpm";
   const innerArgs = [
     `${pnpmCmd} tools-pack linux build`,
     `--to ${config.to}`,
@@ -119,7 +144,7 @@ export function buildDockerArgs(
   if (config.portable) {
     innerArgs.push("--portable");
   }
-  const innerCommand = `${pnpmCmd} install --frozen-lockfile && ` + innerArgs.join(" ");
+  const innerCommand = `${setupPnpm} && ${pnpmCmd} install --frozen-lockfile && ` + innerArgs.join(" ");
 
   return [
     "run",
--- a/tools/pack/tests/linux.test.ts
+++ b/tools/pack/tests/linux.test.ts
@@ -87,22 +87,58 @@ describe("buildDockerArgs", () => {
   it("re-invokes pnpm tools-pack linux build inside the container without --containerized", () => {
     const args = buildDockerArgs(makeConfig(), { uid: 1000, gid: 1000 });
     const last = args[args.length - 1];
-    expect(last).toMatch(/npx --yes pnpm@\d+\.\d+\.\d+ install --frozen-lockfile/);
-    expect(last).toMatch(/npx --yes pnpm@\d+\.\d+\.\d+ tools-pack linux build --to all --namespace default/);
+    expect(last).toMatch(/command -v curl >\/dev\/null/);
+    expect(last).toMatch(/case "\$\(uname -m\)" in/);
+    expect(last).toMatch(/x86_64\) PNPM_ASSET=pnpm-linuxstatic-x64; PNPM_SHA256=[a-f0-9]{64}/);
+    expect(last).toMatch(/aarch64\) PNPM_ASSET=pnpm-linuxstatic-arm64; PNPM_SHA256=[a-f0-9]{64}/);
+    expect(last).toMatch(
+      /curl --retry 3 --retry-all-errors --connect-timeout 10 --max-time 60 -fsSL "https:\/\/github\.com\/pnpm\/pnpm\/releases\/download\/v\d+\.\d+\.\d+\/\$PNPM_ASSET" -o \/tmp\/pnpm\.tmp/,
+    );
+    expect(last).toMatch(/echo "\$PNPM_SHA256  \/tmp\/pnpm\.tmp" \| sha256sum -c -/);
+    expect(last).toMatch(/mv \/tmp\/pnpm\.tmp \/tmp\/pnpm/);
+    expect(last).toMatch(/chmod \+x \/tmp\/pnpm/);
+    expect(last).toMatch(/\/tmp\/pnpm install --frozen-lockfile/);
+    expect(last).toMatch(/\/tmp\/pnpm tools-pack linux build --to all --namespace default/);
     expect(last).not.toMatch(/--containerized/);
   });
 
-  it("invokes pnpm via `npx --yes pnpm@<version>` (electronuserland/builder:base strips corepack, and the non-root container can't write Node shim dir)", () => {
+  it("fetches pnpm standalone binary instead of relying on image npm tooling (electronuserland/builder:base ships a minimal Node without npm/npx/corepack)", () => {
     const args = buildDockerArgs(makeConfig(), { uid: 1000, gid: 1000 });
     const last = args[args.length - 1];
     expect(last).not.toMatch(/corepack/);
-    expect(last).toMatch(/npx --yes pnpm@/);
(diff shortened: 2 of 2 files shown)


lefarcen (commented):
Strong fix — directly addresses the real constraint (electronuserland/builder:base strips ALL npm tooling). Two P2 supply-chain gaps to address before the next release-stable dispatch:

tools/pack/src/linux.ts
```diff
@@ -96,20 +96,23 @@ export function buildDockerArgs(
...
+  // cache it under `/tmp/pnpm` for the lifetime of the container, which is
+  // writable by the unprivileged container user.
+  //
+  // The pinned version matches the `packageManager` field in the root
+  // package.json so reproducibility is preserved (a guard test in
+  // tools/pack/tests/linux.test.ts asserts the lockstep).
   const PNPM_VERSION = "10.33.2";
-  const pnpmCmd = `npx --yes pnpm@${PNPM_VERSION}`;
+  const pnpmDownloadUrl = `https://github.com/pnpm/pnpm/releases/download/v${PNPM_VERSION}/pnpm-linuxstatic-x64`;
+  const setupPnpm = `curl -fsSL "${pnpmDownloadUrl}" -o /tmp/pnpm && chmod +x /tmp/pnpm`;
```
lefarcen: **P2 — No checksum/signature verification**

The standalone pnpm binary is downloaded and executed without verifying SHA256 or signature. HTTPS protects transport, but it does not protect against compromised GitHub release assets, CDN compromise, or corrupted responses.

This binary runs inside a container with the workspace mounted read-write and produces release artifacts — supply-chain gap.

**Concrete fix:** Pin and verify SHA256 before `chmod +x`. Download to temp path → verify → atomically move to `/tmp/pnpm`:

```ts
const pnpmSha256 = "<expected SHA256 for v10.33.2 linuxstatic-x64>";
const setupPnpm = `curl -fsSL "${pnpmDownloadUrl}" -o /tmp/pnpm.tmp && \
  echo "${pnpmSha256}  /tmp/pnpm.tmp" | sha256sum -c - && \
  mv /tmp/pnpm.tmp /tmp/pnpm && chmod +x /tmp/pnpm`;
```

Extract the expected checksum from https://github.com/pnpm/pnpm/releases/tag/v10.33.2 when bumping `PNPM_VERSION`.

lefarcen: **P2 — Hard dependency on live GitHub availability at every build**

`curl -fsSL` will fail cleanly on network errors, but transient outages, GitHub downtime, or rate limiting now fail packaging even when pnpm dependencies are otherwise cached.

**Concrete fix:** Add retry + timeout:

```ts
const setupPnpm = `curl --retry 3 --retry-all-errors --connect-timeout 10 --max-time 60 -fsSL "${pnpmDownloadUrl}" -o /tmp/pnpm && chmod +x /tmp/pnpm`;
```

Optional but better: persist the downloaded binary under the mounted `.docker-cache` directory keyed by `${PNPM_VERSION}.${checksum}`, falling back to download only on cache miss. Eliminates transient-network as a release blocker.

tools/pack/src/linux.ts
```diff
@@ -96,20 +96,23 @@ export function buildDockerArgs(
...
+  // every "ask the image to invoke pnpm" path fails with `command not found`.
   //
-  // Use `npx --yes pnpm@<version>` instead: `npx` ships with npm (always
-  // present in the image), `--yes` skips the install confirmation, and the
-  // package gets cached under `$HOME/.npm/_npx`, which is writable by the
-  // unprivileged user. The pinned version matches the `packageManager`
-  // field in the root package.json so reproducibility is preserved.
+  // Fix: download the official pnpm `linuxstatic-x64` standalone binary at
+  // container start. The binary is self-contained (it bundles its own Node
+  // runtime), so it doesn't depend on the image's npm tooling at all. We
```
lefarcen: **P3 — Cache wording misleading**

Comment says the binary is "cached under `/tmp/pnpm`", but this container runs with `--rm`, so `/tmp/pnpm` only exists for this single command chain and does not cache across builds.

Future maintainers may misinterpret this as offline-capable. Suggest: "stage it under `/tmp/pnpm` for this container invocation" or implement a real mounted cache (see P2 above).

tools/pack/src/linux.ts
```diff
@@ -96,20 +96,23 @@ export function buildDockerArgs(
...
   // interpolation into the inner command string is safe.
   //
-  // We can't rely on `corepack pnpm` here: although Node 16.10+ ships corepack,
-  // the `electronuserland/builder:base` image strips the corepack binary, so
-  // the inner `bash -lc` fails with `corepack: command not found`. We also
-  // can't `corepack enable` ourselves — the container runs as the host's
-  // non-root uid (--user above) and corepack would try to write shims next
-  // to the system Node binary, which is owned by root in this image.
+  // The `electronuserland/builder:base` image is intentionally minimal: it
+  // ships a Node binary but strips npm, npx, *and* corepack from PATH. So
```
lefarcen, line 100: **P3 — Undocumented curl assumption**

The comment documents why npm/npx/corepack cannot be used, but it does not state the new hard assumption that `curl` is present in `electronuserland/builder:base`.

If the image changes, the failure becomes another `command not found` path. Suggest adding: "We rely on `curl`, which electron-builder itself uses for downloads and is consistently present in this image."

Optional guard for clearer errors:

```ts
const setupPnpm = `command -v curl >/dev/null || { echo 'curl not found in container image' >&2; exit 127; } && curl -fsSL "${pnpmDownloadUrl}" ...`;
```

tools/pack/tests/linux.test.ts
```diff
@@ -87,22 +87,26 @@ describe("buildDockerArgs", () => {
...
     expect(last).not.toMatch(/--containerized/);
   });
 
-  it("invokes pnpm via `npx --yes pnpm@<version>` (electronuserland/builder:base strips corepack, and the non-root container can't write Node shim dir)", () => {
+  it("fetches pnpm standalone binary instead of relying on image npm tooling (electronuserland/builder:base ships a minimal Node without npm/npx/corepack)", () => {
     const args = buildDockerArgs(makeConfig(), { uid: 1000, gid: 1000 });
     const last = args[args.length - 1];
     expect(last).not.toMatch(/corepack/);
-    expect(last).toMatch(/npx --yes pnpm@/);
+    expect(last).not.toMatch(/\bnpx\b/);
```
lefarcen, line 109: **P3 — Negative npm assertion incomplete**

The regex `/^\s*npm\s/` only matches `npm` at the start of the command, so it would not catch `... && npm ...` mid-chain.

Since the test's stated purpose is "no npm/npx/corepack appears in the inner command", use a command-token regex:

```ts
expect(last).not.toMatch(/(^|[;&|]\s*)npm(\s|$)/);
```

Or inspect split shell tokens more explicitly if you want full coverage of edge placements.

tools/pack/src/linux.ts
```diff
@@ -96,20 +96,23 @@ export function buildDockerArgs(
...
+  // runtime), so it doesn't depend on the image's npm tooling at all. We
+  // cache it under `/tmp/pnpm` for the lifetime of the container, which is
+  // writable by the unprivileged container user.
+  //
+  // The pinned version matches the `packageManager` field in the root
+  // package.json so reproducibility is preserved (a guard test in
+  // tools/pack/tests/linux.test.ts asserts the lockstep).
   const PNPM_VERSION = "10.33.2";
-  const pnpmCmd = `npx --yes pnpm@${PNPM_VERSION}`;
+  const pnpmDownloadUrl = `https://github.com/pnpm/pnpm/releases/download/v${PNPM_VERSION}/pnpm-linuxstatic-x64`;
```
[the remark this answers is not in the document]
lefarcen: Good catch — the previous draft would have broken Apple Silicon dev loops. Fixed: detect inside the container with `uname -m` and pick `pnpm-linuxstatic-x64` for `x86_64` or `pnpm-linuxstatic-arm64` for `aarch64`, with a clear `exit 1` for any other arch. Added a dedicated sentinel test (`picks the pnpm asset by container CPU so amd64 and arm64 hosts both work`) plus updated the existing two tests to assert the case statement. 26 tests pass locally.

tools/pack/src/linux.ts
lefarcen: Addressed. The container now pins SHA256 for both pnpm v10.33.2 linuxstatic assets (`x64` and `arm64`), downloads to `/tmp/pnpm.tmp`, verifies with `sha256sum -c -`, then atomically moves to `/tmp/pnpm` before chmod/execution. Added a sentinel test for verify-before-execute ordering.

lefarcen: Addressed. The download now uses `curl --retry 3 --retry-all-errors --connect-timeout 10 --max-time 60` so transient GitHub/CDN/network failures get bounded retries instead of a single-shot release blocker.

tools/pack/src/linux.ts
lefarcen: Addressed. Reworded the comment from “cache” to “stage it under `/tmp/pnpm` for this container invocation” so it does not imply persistence across `--rm` container runs.

tools/pack/src/linux.ts
lefarcen: Addressed. Documented the `curl` image assumption and added `command -v curl` with a clear `curl not found in container image` error before attempting the download.

tools/pack/tests/linux.test.ts
lefarcen: Addressed. Strengthened the negative assertion to a command-token regex so `npm` is rejected at the start of the command and after shell separators, not only at position 0.

lefarcen (commented):
All P2 supply-chain gaps addressed ✅

**What was fixed:**
- SHA256 checksum verification for both x64 and arm64 binaries (download → verify → atomic move)
- Network resilience: curl retry (3×) + timeouts (connect 10s, max 60s)
- curl availability guard documented and enforced
- Cache wording corrected ("stage" vs "cache")
- Test coverage expanded: sha256 verification test + improved negative npm regex

**Multi-arch support bonus:** uname -m detection handles both amd64 (GitHub Actions) and arm64 (Apple Silicon local builds).

LGTM 🎉

tools/pack/src/linux.ts
```diff
@@ -96,20 +96,45 @@ export function buildDockerArgs(
...
+    `case "$(uname -m)" in ` +
+    `x86_64) PNPM_ASSET=pnpm-linuxstatic-x64; PNPM_SHA256=${pnpmLinuxStaticX64Sha256} ;; ` +
+    `aarch64) PNPM_ASSET=pnpm-linuxstatic-arm64; PNPM_SHA256=${pnpmLinuxStaticArm64Sha256} ;; ` +
+    `*) echo "unsupported container arch: $(uname -m)" >&2; exit 1 ;; ` +
+    `esac && ` +
+    `curl --retry 3 --retry-all-errors --connect-timeout 10 --max-time 60 -fsSL "${pnpmReleaseUrl}/$PNPM_ASSET" -o /tmp/pnpm.tmp && ` +
+    `echo "$PNPM_SHA256  /tmp/pnpm.tmp" | sha256sum -c - && ` +
+    `mv /tmp/pnpm.tmp /tmp/pnpm && ` +
+    `chmod +x /tmp/pnpm`;
+  const pnpmCmd = "/tmp/pnpm";
```
mrcfps, line 137: **Blocking — avoid re-entering the root script through a bare `pnpm` lookup**

This line now makes `pnpmCmd` an absolute standalone binary at `/tmp/pnpm`, but the inner build command below still runs `${pnpmCmd} tools-pack linux build`. In this workspace, the root `tools-pack` script is `"pnpm exec tools-pack"`, so that shorthand will enter the package script and then try to resolve a bare `pnpm` from `PATH`. The setup code downloads `/tmp/pnpm` but never adds `/tmp` to `PATH`, and the PR is specifically fixing an image where npm/npx/corepack are absent; that means the release build can get past install and then fail again with `pnpm: command not found`.

Please bypass the root script for the second step, for example by changing the inner build argv to `${pnpmCmd} exec tools-pack linux build ...`, or explicitly prepend the standalone binary location to `PATH` before invoking any scripts that call `pnpm`. It would also be worth updating the test at `tools/pack/tests/linux.test.ts` to assert the inner build uses `exec tools-pack` (or that `PATH` is set) so this regression is covered.

mrcfps (changes requested):
@lefarcen thanks for pushing this through and for addressing the earlier supply-chain and arch-selection feedback — the standalone pnpm path is much safer now. I found one remaining primary-flow issue that can still break the Linux containerized release build before this is ready.


iuliandita:
can i help? i would really like to see the 0.4.0 linux artifact :)

mrcfps:
@lefarcen Thanks again for the contribution here 💛

lefarcen:
Quick sync for anyone following this Linux artifact thread: #1204 has now merged the standalone `pnpm-linuxstatic` bootstrap and Linux packaged smoke coverage that superseded this PR.

What landed on `main`: the containerized Linux build downloads the pinned `pnpm-linuxstatic-{x64,arm64}` binary, verifies SHA256 before moving it to `/tmp/pnpm`, exports `OD_TOOLS_PACK_PNPM_BIN=/tmp/pnpm`, and keeps the “no `npx` / `corepack` in the container path” regression checks in `tools/pack/tests/linux.test.ts`. The merged PR’s Linux headless smoke check was green.

This PR can stay closed as superseded. Release publishing still depends on the next `release-stable` dispatch, so this is a code-path sync rather than an artifact-published confirmation.

PR #561, opened 2026-05-05
fix: prevent Design Files row menu from being clipped at viewport boundaries

1119302165:
### Title
**fix: prevent design files row menu from being clipped at viewport boundaries**
---

before fix：
image

after fix：
image

### Root Cause Analysis
https://github.com/nexu-io/open-design/issues/124
 The row overflow menu in Design Files panel used a naive static positioning strategy that always expanded downward:
```
top = triggerElement.bottom + 4px
```

This caused critical usability issues at viewport boundaries:
- **Bottom boundary clipping**: When clicking menu triggers on rows near the bottom of a long file list, the popup menu would be cut off by the viewport edge
- **Delete action unreachable**: The destructive "Delete" button (last menu item) became invisible or unclickable exactly when users needed file management the most — in long asset lists
- No boundary detection was implemented for either vertical or horizontal viewport edges

---

### Fix Approach
Implemented intelligent boundary-aware positioning with 3-tier fallback strategy in `DesignFilesPanel.tsx`:

1. **Space detection**
   - Calculate available space both above and below the trigger element
   - Compare against estimated menu height (~115px for 3 actions)

2. **Smart positioning strategy**
   - ✅ **Default**: Expand downward when sufficient space exists below
   - ↕️ **Flip upward**: Auto-reverse direction when bottom space is insufficient but top space is available
   - ⌖ **Viewport anchoring**: Force menu into safe viewport region when neither top nor bottom has enough space

3. **Horizontal protection**
   - Left boundary clipping defense with minimum safe padding

---

### Files Modified
| File | Changes | Purpose |
|------|---------|---------|
| [`apps/web/src/components/DesignFilesPanel.tsx`](file:///Users/bojiehuang/Downloads/code/open-design/apps/web/src/components/DesignFilesPanel.tsx) | +24/-2 | Replace naive positioning with full boundary detection logic |

**Scope**: Single component — no breaking changes to shared styles or other menus.

---

### Edge Cases Covered

- ✅ Rows at bottom of long scrolling lists (most common failure case)
- ✅ Rows at top of viewport (fallback to downward)
- ✅ Extremely short viewports (anchor to visible region)
- ✅ Narrow viewports (prevent left-side clipping)

---

### Impact
- **Priority**: Medium (breaks core file-management UX in common scenario)
- **Backwards Compatible**: Yes
- **Visual Regression Risk**: Low — maintains original behavior in 90% of cases, only improves edge scenarios
- **Tested**: Manual verification across scroll positions and viewport sizes

### DIFF
--- a/apps/web/src/components/DesignFilesPanel.tsx
+++ b/apps/web/src/components/DesignFilesPanel.tsx
@@ -61,6 +61,8 @@ export function DesignFilesPanel({
   const dragDepthRef = useRef(0);
   const [hover, setHover] = useState<string | null>(null);
   const [menuPos, setMenuPos] = useState<{ name: string; top: number; left: number } | null>(null);
+  const MENU_ESTIMATED_HEIGHT = 115;
+  const MENU_SAFE_PADDING = 8;
   const [preview, setPreview] = useState<string | null>(null);
   const [sectionLimits, setSectionLimits] = useState<Partial<Record<Section, number>>>({});
   const [isSectionExpansionPending, startSectionExpansion] = useTransition();
@@ -379,10 +381,30 @@ export function DesignFilesPanel({
                             const rect = (e.target as HTMLElement)
                               .closest('.df-row-menu')
                               ?.getBoundingClientRect();
+                            if (!rect) return;
+                            
+                            const viewportHeight = window.innerHeight;
+                            const spaceBelow = viewportHeight - rect.bottom;
+                            const spaceAbove = rect.top;
+                            
+                            let top: number;
+                            if (spaceBelow >= MENU_ESTIMATED_HEIGHT + MENU_SAFE_PADDING) {
+                              top = rect.bottom + 4;
+                            } else if (spaceAbove >= MENU_ESTIMATED_HEIGHT + MENU_SAFE_PADDING) {
+                              top = rect.top - MENU_ESTIMATED_HEIGHT - 4;
+                            } else {
+                              top = Math.max(
+                                MENU_SAFE_PADDING,
+                                viewportHeight - MENU_ESTIMATED_HEIGHT - MENU_SAFE_PADDING,
+                              );
+                            }
+                            
+                            const left = Math.max(MENU_SAFE_PADDING, rect.right - 160);
+                            
                             setMenuPos({
                               name: f.name,
-                              top: (rect?.bottom ?? 0) + 4,
-                              left: (rect?.right ?? 0) - 160,
+                              top,
+                              left,
                             });
                           }}
                         >


lefarcen:
Hi @1119302165! 🎉
Thanks for the contribution — solving the viewport clipping issue for the Design Files menu is a solid UX improvement.

— open-design team

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -379,10 +381,30 @@ export function DesignFilesPanel({
                             const rect = (e.target as HTMLElement)
                               .closest('.df-row-menu')
                               ?.getBoundingClientRect();
+                            if (!rect) return;
+                            
+                            const viewportHeight = window.innerHeight;
+                            const spaceBelow = viewportHeight - rect.bottom;
+                            const spaceAbove = rect.top;
+                            
+                            let top: number;
```
lefarcen, line 390: P2 The boundary logic relies on hardcoded dimensions (115px, 160px) instead of the rendered menu size. CSS only sets `min-width: 160px`, so:
- Localized labels / font changes / zoom can make the popover taller/wider
- The 4px gap is omitted from the safe space check, so the intended 8px padding becomes 4px

Concrete fix: store the anchor rect/name, render the popover with a ref, measure `popover.getBoundingClientRect()` in `useLayoutEffect`, then clamp using actual width/height and `gap + padding`.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -379,10 +381,30 @@ export function DesignFilesPanel({
...
+                            } else if (spaceAbove >= MENU_ESTIMATED_HEIGHT + MENU_SAFE_PADDING) {
+                              top = rect.top - MENU_ESTIMATED_HEIGHT - 4;
+                            } else {
+                              top = Math.max(
+                                MENU_SAFE_PADDING,
+                                viewportHeight - MENU_ESTIMATED_HEIGHT - MENU_SAFE_PADDING,
+                              );
+                            }
+                            
+                            const left = Math.max(MENU_SAFE_PADDING, rect.right - 160);
```
lefarcen, line 402: P2 Horizontal clipping is only partially handled. `Math.max(8, rect.right - 160)` prevents negative left but never clamps against `window.innerWidth`, and assumes the menu is exactly 160px.

On narrow/mobile viewports or a menu wider than 160px, the right edge can still overflow.

Concrete fix: compute `left = clamp(rect.right - menuWidth, padding, window.innerWidth - menuWidth - padding)`, with a fallback for viewports smaller than the menu.

lefarcen (commented):
Hi @1119302165! Thanks for tackling the viewport clipping issue — the 3-tier fallback strategy is exactly the right approach. 👍

I found a few edge cases where the current implementation could still clip or detach. All are non-blocking but worth addressing for robustness:

**Core issue**: The logic assumes fixed dimensions (115px height, 160px width) but the actual menu can vary with content/zoom/localization.

**Summary of findings**:
- P2: Hardcoded dimensions don't match actual rendered size → can still clip with taller/wider content (line 390-391)
- P2: Right-edge clipping not handled → narrow viewports or wider menus (line 402)
- P2: No test coverage for boundary positioning
- P3: Menu doesn't reposition on scroll/resize → can detach from anchor after initial placement

All fixable by measuring the actual menu dimensions after render and clamping against both viewport edges. The architectural improvement would be to listen for scroll/resize events and recompute or close the menu.

Let me know if you'd like detailed suggestions for any of these!

mrcfps (approved):
@1119302165 I reviewed the Design Files row menu positioning change in `apps/web/src/components/DesignFilesPanel.tsx`. The updated viewport-aware placement keeps the existing default behavior while flipping or anchoring the popover when space is constrained, and I did not find correctness, safety, or maintainability issues in the changed range. Thanks for the thoughtful UX fix and clear handoff screenshots! 🎉


apps/web/src/components/DesignFilesPanel.tsx
1119302165: Hi! Thanks so much for the thoughtful review — really appreciate you catching those edge cases! 🙏  

You’re absolutely right: hardcoding dimensions and not handling dynamic viewport changes can lead to clipping or detachment, especially with localization, zoom, or responsive layouts.  

I’d like to land the current fix first since it resolves the most common clipping scenarios (and unblocks the immediate use case), but I’ll follow up shortly with a more robust version that:  
- Measures actual menu dimensions after render  
- Clamps position against both left/right and top/bottom viewport edges  
- Adds scroll/resize listeners to either reposition or auto-close the menu  
- Includes test coverage for boundary conditions  

If you’re open to it, I’d also welcome your detailed suggestions on the resize/reposition logic — happy to align on the best pattern!  

Thanks again for the great feedback!

lefarcen:
🎉 Merged! Thanks @1119302165 for the thoughtful fix — the 3-tier fallback strategy (downward default → flip upward → viewport anchor) is exactly the right approach for this clipping issue.

Looking forward to your follow-up PR with dynamic dimension measurement and scroll/resize handling!

PR #577, opened 2026-05-05
Fix packaged contracts runtime exports

iuliandita:
## Summary
- Build @open-design/contracts into dist/*.mjs and dist/*.d.ts for package runtime exports.
- Point package exports at built artifacts so packaged Node runtimes do not import TypeScript from node_modules.
- Build contracts before packaged linux/mac/win assembly and during the shared postinstall bootstrap.
- Add runtime package-shape coverage that checks generated files and imports both package export paths.

## Verification
- pnpm install
- pnpm --filter @open-design/contracts test -- package-runtime.test.ts
- pnpm --filter @open-design/contracts build
- pnpm -C packages/contracts typecheck
- pnpm guard
- rm -rf packages/contracts/dist && node scripts/postinstall.mjs
- pnpm --filter @open-design/tools-pack test
- pnpm --filter @open-design/tools-pack build
- pnpm -C tools/pack typecheck
- Linux AppImage smoke downstream: packaged daemon/web reached running state

## Follow-up from Linux/AUR smoke
The AUR package now pre-extracts the AppImage into an AppDir at install time, which avoids repeated AppImage extraction on launch and materially improves startup. That is useful upstream packaging work, but it changes tools-pack linux install/start/uninstall process matching and AppDir permission handling, so it should be a separate focused PR instead of being bundled into this contracts runtime export fix.

### DIFF
--- a/packages/contracts/esbuild.config.mjs
+++ b/packages/contracts/esbuild.config.mjs
@@ -1,0 +1,13 @@
+import { build } from "esbuild";
+
+await build({
+  bundle: true,
+  entryNames: "[name]",
+  entryPoints: ["./src/index.ts", "./src/critique.ts"],
+  format: "esm",
+  outdir: "./dist",
+  outExtension: { ".js": ".mjs" },
+  packages: "external",
+  platform: "node",
+  target: "node24",
+});
--- a/packages/contracts/package.json
+++ b/packages/contracts/package.json
@@ -4,18 +4,23 @@
   "private": true,
   "type": "module",
   "description": "Shared pure TypeScript contracts for the Open Design web/daemon boundary.",
+  "main": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
   "exports": {
     ".": {
-      "types": "./src/index.ts",
-      "default": "./src/index.ts"
+      "types": "./dist/index.d.ts",
+      "default": "./dist/index.mjs"
     },
     "./critique": {
-      "types": "./src/critique.ts",
-      "default": "./src/critique.ts"
+      "types": "./dist/critique.d.ts",
+      "default": "./dist/critique.mjs"
     }
   },
-  "types": "./src/index.ts",
   "scripts": {
+    "build": "node ./esbuild.config.mjs && tsc -p tsconfig.json --emitDeclarationOnly",
     "test": "vitest run",
     "typecheck": "tsc -p tsconfig.json --noEmit && tsc -p tsconfig.tests.json --noEmit"
   },
@@ -23,6 +28,7 @@
     "zod": "^3.23.8"
   },
   "devDependencies": {
+    "esbuild": "0.27.7",
     "typescript": "^5.6.3",
     "vitest": "^2.1.8"
   }
--- a/packages/contracts/tests/package-runtime.test.ts
+++ b/packages/contracts/tests/package-runtime.test.ts
@@ -1,4 +1,5 @@
 import { readFileSync } from 'node:fs';
+import { access } from 'node:fs/promises';
 import { dirname, join } from 'node:path';
 import { fileURLToPath } from 'node:url';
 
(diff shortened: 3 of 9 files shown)


lefarcen:
Hi @iuliandita! 🎉
Thanks for the contribution — fixing package exports for runtime usage is important infrastructure work.

— open-design team

packages/contracts/esbuild.config.mjs
```diff
@@ -0,0 +1,13 @@
+import { build } from "esbuild";
```
lefarcen, line 1: **P1** Adding a new project-owned `.mjs` file will fail the repo's `pnpm guard` residual JavaScript check because `scripts/guard.ts` allowlists exact `.mjs` paths and this path was not added.

Either make the config TypeScript-first (rename to `.ts` or `.mts`) or add an explicit guard allowlist entry with rationale — otherwise standard validation will block this PR.

packages/contracts/package.json
```diff
@@ -4,25 +4,31 @@
...
+  "main": "./dist/index.mjs",
+  "types": "./dist/index.d.ts",
+  "files": [
+    "dist"
+  ],
   "exports": {
     ".": {
-      "types": "./src/index.ts",
-      "default": "./src/index.ts"
+      "types": "./dist/index.d.ts",
```
lefarcen, line 14: **P1** The new exports point consumers to `./dist/*`, but the repo postinstall build targets **do not build `packages/contracts`**.

A fresh `pnpm install` leaves `packages/contracts/dist` absent, so `@open-design/contracts` and `@open-design/contracts/critique` become unresolved for package-scoped web/daemon commands until someone manually runs the contracts build.

The pack pipeline additions (linux/mac/win.ts) are ordered correctly, but the general workspace install/runtime path is now broken. You'll need to add a postinstall/prepare hook or ensure the contracts build runs in the workspace setup flow.

packages/contracts/tests/package-runtime.test.ts
```diff
@@ -0,0 +1,30 @@
...
+function readPackageJson(): {
+  exports?: Record<string, { default?: string; types?: string }>;
+  files?: string[];
+  main?: string;
+  types?: string;
+} {
+  return JSON.parse(readFileSync(join(packageRoot, 'package.json'), 'utf8'));
+}
+
+describe('@open-design/contracts package runtime shape', () => {
```
lefarcen, line 19: **P2** The new test only checks `package.json` strings and doesn't validate the actual runtime package shape.

It will pass even when `dist/index.mjs`, `dist/critique.mjs`, or declarations are missing or unimportable — which is exactly the failure mode this PR is trying to prevent.

Consider adding a runtime import test:
```typescript
it('runtime exports are actually importable', async () => {
  const { someExport } = await import('@open-design/contracts');
  const { critiqueExport } = await import('@open-design/contracts/critique');
  expect(someExport).toBeDefined();
  expect(critiqueExport).toBeDefined();
});
```

lefarcen (dismissed):
Thanks for tackling the package export infrastructure! The direction is solid, but there are three blocking issues that need resolution:

1. The new `.mjs` config will fail repo guard validation
2. The exports break fresh workspace installs (missing `dist/` until manual build)
3. The runtime test doesn't actually validate that the built artifacts are importable

packages/contracts/package.json
```diff
@@ -4,25 +4,31 @@
...
+      "default": "./dist/index.mjs"
     },
     "./critique": {
-      "types": "./src/critique.ts",
-      "default": "./src/critique.ts"
+      "types": "./dist/critique.d.ts",
+      "default": "./dist/critique.mjs"
```
mrcfps, line 19: These exports now make every workspace consumer resolve `@open-design/contracts` and `@open-design/contracts/critique` through `dist/*`, but this PR only builds contracts inside the tools-pack packaging flows. On a fresh checkout the `dist` directory is absent before the normal daemon/web builds run, so the package subpath becomes unresolvable; the current CI failure shows `@open-design/daemon build` hitting `TS2307: Cannot find module '@open-design/contracts/critique'`. Please add the contracts build to the shared workspace preparation path before consumers are built (for example `scripts/postinstall.mjs` and the CI prebuild step), or otherwise keep the workspace type exports resolvable until `dist` has been generated.

packages/contracts/esbuild.config.mjs
mrcfps, line 1: Adding this new project-owned `.mjs` config will trip the repository residual-JavaScript guard: `scripts/guard.ts` allowlists the existing esbuild config files, but it does not include `packages/contracts/esbuild.config.mjs`. Since `pnpm guard` is a required workspace validation, this will fail even after the build-order issue above is fixed. Please either add this exact config path to the documented guard allowlist with the same esbuild-config rationale, or convert the build entrypoint to an allowed TypeScript-based form.

mrcfps (changes requested):
@iuliandita thanks for taking on the packaged contracts runtime export fix — this is useful infrastructure work, and the package/build direction is close. I found a couple of blocking workspace validation issues that need to be addressed before this can merge.


iuliandita:
Addressed the review blockers in 8f2ee8c:

- added `packages/contracts` to `scripts/postinstall.mjs` so fresh installs build the dist exports before workspace consumers run
- added the contracts esbuild config to the residual-JS guard allowlist with an esbuild-config rationale
- expanded the contracts runtime package test to check generated files and import `@open-design/contracts` plus `@open-design/contracts/critique` through package exports

I resolved the inline threads this addresses. Current PR checks are green.

mrcfps (approved):
@iuliandita I reviewed the updated contracts packaging changes and verified the previous blockers are addressed: the contracts package now builds generated runtime exports during shared postinstall and tools-pack assembly, the guard allowlist covers the new esbuild entrypoint, and the runtime package-shape tests now check both generated files and importability. I also ran the targeted contracts tests/typecheck, repository guard, daemon build, and tools-pack build successfully. Thanks for the thoughtful follow-up here — this looks ready from my review. 🙌


lefarcen (commented):
All three blocking issues from my earlier review are now resolved:

✅ Guard validation: `packages/contracts/esbuild.config.mjs` added to allowlist
✅ Postinstall build: contracts package now builds during workspace preparation
✅ Test coverage: runtime test validates actual importability of `@open-design/contracts` and `@open-design/contracts/critique`

lefarcen (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @iuliandita for the contribution to open-design. Looking forward to the next one!

PR #582, opened 2026-05-05
build: package @open-design/contracts to dist

DrMaks22:
## Summary

`@open-design/contracts` was exported directly from `src/*.ts`, so the packaged app could ship TypeScript sources into `node_modules` and trip Node's `ERR_UNSUPPORTED_NODE_MODULES_TYPE_STRIPPING` at runtime. This change turns the package into a normal build artifact with `dist` exports, matching the other internal packages.

## What changed

- Added an `esbuild` build for `dist/index.mjs` and `dist/critique.mjs`.
- Kept TypeScript declarations in `dist` via `tsc --emitDeclarationOnly`.
- Switched package exports, `main`, and `types` to `dist`.
- Added `files: ["dist"]` and a `prepack` hook so `pnpm pack` always ships the built output.
- Committed the generated `packages/contracts/dist` artifacts so local and release workflows stay aligned with the rest of the repo.

## Verification

- `pnpm --filter @open-design/contracts test`
- `pnpm --filter @open-design/contracts pack` and inspected the tarball contents to confirm it only contains `dist/*` runtime files and declarations.


packages/contracts/esbuild.config.ts
```diff
@@ -0,0 +1,13 @@
+import { build } from "esbuild";
```
mrcfps, line 3: **Blocking:** This new `packages/contracts/esbuild.config.mjs` is a project-owned `.mjs` config, but the repository guard rejects newly added residual JavaScript/MJS unless it is generated/vendor/output or explicitly documented in the allowlist. The live `Validate workspace` check is failing on this exact file, so the PR cannot merge as-is even though the packaging approach is otherwise sound.

Please either convert the build config to TypeScript and update `packages/contracts/package.json`'s `build` script to run the TypeScript entrypoint, or add a documented compatibility allowlist entry if there is a concrete reason this config must remain `.mjs`.

mrcfps (changes requested):
Thanks @DrMaks22 for tightening the contracts package packaging. I found one blocking CI/guard issue in the new build config that needs to be addressed before this can merge.


lefarcen (changes requested):
Hi @DrMaks22, thanks for the PR! 👋

Found 3 issues after a deep review — one critical for downstream TypeScript consumers, one build hygiene, one .gitignore clarification.

---

**P1 — NodeNext module resolution (packages/contracts/dist/index.d.ts)**

The emitted declarations use extensionless relative exports like `'./common'` and `'./api/projects'`, but this package is ESM (`"type": "module"`). Under `moduleResolution: "NodeNext"` (the default for NodeNext projects), TypeScript expects `.d.ts` imports to use explicit `.js`-style extensions, and the current output will fail with TS2834/TS2835 errors for downstream consumers.

Fix by:
1. Post-processing the generated `.d.ts` files to rewrite bare specifiers to `.js` (e.g., `'./common'` → `'./common.js'`), OR
2. Bundling declarations so `index.d.ts` is self-contained and has no relative imports.

Example error a consumer would see:
```
node_modules/@open-design/contracts/dist/index.d.ts(1,15): error TS2835: Relative import paths need explicit file extensions in ECMAScript imports when '--moduleResolution' is 'node16' or 'nodenext'.
```

---

**P2 — Stale artifact accumulation (packages/contracts/package.json:23)**

The build script runs `esbuild` and `tsc --emitDeclarationOnly` without cleaning `dist/` first. If source files are removed or renamed, old `.d.ts` / `.d.ts.map` / `.mjs` files from previous builds remain in `dist/`, and `prepack` will commit/ship them.

Fix by deleting `dist` before building:
```json
"build": "rm -rf dist && node ./esbuild.config.mjs && tsc -p tsconfig.json --emitDeclarationOnly",
```

---

**P3 — .gitignore vs committed dist**

The root `.gitignore` has a global `dist/` entry, yet this PR commits `packages/contracts/dist/**` files. That works for the current commit (already forced-in), but future new generated files under that directory will be hidden from `git add` / `git status` unless explicitly forced.

For clarity, either:
1. Add a narrow unignore in the root `.gitignore`: `!/packages/contracts/dist/`, OR
2. Add a `.gitignore` inside `packages/contracts/` with `!dist/` so the package-specific dist is tracked.

Not blocking, but reduces risk of incomplete commits later.

---

Once these are addressed (especially P1 + P2), this should be good to merge. Thanks for the contribution!

packages/contracts/esbuild.config.ts
DrMaks22: Fixed: converted the build config to TypeScript, and `pnpm guard` now passes again. Also updated the build to clean `dist` before packaging and kept `packages/contracts/dist` trackable without forcing git add.

lefarcen (commented):
Nice work @DrMaks22! 🎉

All three issues are now resolved:
- ✅ mrcfps's CI blocker: `esbuild.config.ts` conversion complete
- ✅ P1 NodeNext module resolution: dist declarations now use `.js` extensions
- ✅ P2 stale artifact cleanup: build script updated

The PR is ready from my side. Looks good to merge!

packages/contracts/dist/prompts/system.d.ts
```diff
@@ -0,0 +1,44 @@
...
+ *      (skill_id null) still get a framework, instead of having the agent
+ *      re-author scaling / nav / print logic from scratch each turn. When
+ *      the active skill ships its own seed (skill body references
+ *      `assets/template.html`), we defer to that seed and skip the generic
+ *      skeleton — the skill's framework wins to avoid double-injection.
+ *
+ * The composed string is what the daemon sees as `systemPrompt` and what
+ * the Anthropic path sends as `system`.
+ */
+import type { ProjectMetadata, ProjectTemplate } from '../api/projects';
```
mrcfps: **Blocking:** This generated declaration still imports a relative module without the `.js` extension (`../api/projects`). Because `@open-design/contracts` is now an ESM package (`"type": "module"`) and `dist/index.d.ts` re-exports `./prompts/system.js`, downstream TypeScript projects using `moduleResolution: "NodeNext"` will read this file and can fail with TS2834/TS2835 before they can consume the package. This same leftover pattern is still present in other generated declarations such as `dist/api/artifacts.d.ts`, `dist/api/comments.d.ts`, `dist/api/live-artifacts.d.ts`, `dist/sse/chat.d.ts`, and `dist/sse/proxy.d.ts`. Please update the remaining source type imports (for example `../api/projects` here) or add a declaration post-process step so every relative import emitted into `dist/*.d.ts` uses the `.js` extension, then regenerate `dist`; adding a NodeNext declaration-consumer check would keep this from regressing.

mrcfps (changes requested):
Thanks @DrMaks22 for continuing to tighten the contracts packaging. I found one remaining declaration-compatibility blocker for NodeNext consumers; once the remaining generated imports are updated, this should be in good shape.


lefarcen (approved):
All issues resolved — ready to merge! 🎉

lefarcen:
Thanks for the update, @DrMaks22! I see @mrcfps found the same NodeNext declaration issue persists in other generated files — please address that inline comment and we'll be good to go.

packages/contracts/dist/prompts/system.d.ts
DrMaks22: Fixed in the latest push (`2e1eb58`): the remaining source imports were updated to `.js`, `packages/contracts/dist/prompts/system.d.ts` now emits `../api/projects.js`, and the regenerated `dist` no longer contains extensionless relative imports. This thread is outdated relative to the current diff; please re-review the latest push.

lefarcen (commented):
🎉 All issues resolved! @DrMaks22, the NodeNext declaration blocker is now fixed — all generated `.d.ts` files use `.js` extensions. Thanks for the thorough updates!

mrcfps (approved):
@DrMaks22 I reviewed the contracts packaging update, including the `dist` exports, the generated declaration artifacts, the `.js` specifier normalization for NodeNext consumers, and the package-level dist tracking/cleanup path. The previous declaration compatibility and guard concerns look resolved at this head, and the live PR checks are passing. Thanks for iterating on the packaging fix and tightening this up 🙏


lefarcen (commented):
Perfect! All 6 files now have `.js` extensions — NodeNext consumers will be happy. 🎉

mrcfps's P1 blocker is fully fixed. Ready to merge from my side.

lefarcen (approved):
Approved. Please rebase against main and resolve the merge conflict, then you can self-merge.

lefarcen:
@DrMaks22 approved — please rebase against main and resolve the conflict, then you can self-merge.

lefarcen:
Hi @DrMaks22 — thanks for the patch! The same root cause was just fixed in #577 (now merged), so this PR is effectively superseded. The main difference here is committing the generated `dist/` into git, which we generally avoid for internal packages to stay consistent with the rest of the workspace. Going to close this — please feel free to reopen if I'm missing something. Appreciate the contribution!

PR #591, opened 2026-05-06
fix(daemon): unbreak Claude Design ZIP import on Node 24 and raise file ceiling

### LINKED ISSUE #590: Import Claude Design ZIP silently fails: RangeError on 0-byte entries (Node 24) and MAX_FILES=500 too low
## Summary

`POST /api/import/claude-design` (the **Import Claude Design ZIP** button in the New Project panel) silently fails for valid `.zip` files in two distinct ways:

1. **Crash on zero-byte entries (Node 24).** `inflateRawSync(compressed, { maxOutputLength: 0 })` throws `RangeError [ERR_OUT_OF_RANGE]` on Node 24 because `maxOutputLength` must be `>= 1`. Any entry whose central-directory `uncompressedSize` is `0` (empty files, common when the zip uses data descriptors — e.g. macOS Archive Utility, streaming zips) breaks the whole import.
2. **`MAX_FILES = 500` rejects medium-sized exports.** Design-system / prototype exports commonly exceed 500 files (the test asset that surfaced this is 561 files / ~4.4 MB). The existing `MAX_TOTAL_BYTES = 100 MB` and `MAX_FILE_BYTES = 25 MB` guards already cover pathological inputs.

The web UI surfaces neither error: `apps/web/src/state/projects.ts` returns `null` on `!resp.ok`, and `apps/web/src/App.tsx`'s `handleImportClaudeDesign` does `if (!result) return;`. After picking a file, the button appears to "do nothing".

## Reproduction

1. Build (or use an existing) zip with either:
   - any entry whose central-directory `uncompressedSize` is `0`, **or**
   - more than 500 entries.
2. In the running app, click **Import Claude Design ZIP** in the New Project panel and pick the file.
3. UI shows nothing.
4. From a terminal, the underlying daemon error is visible:
   ```bash
   curl -i -F "file=@./your.zip" http://127.0.0.1:<daemon-port>/api/import/claude-design
   ```
   ```
   HTTP/1.1 400 Bad Request
   {"error":"RangeError [ERR_OUT_OF_RANGE]: The value of \"options.maxOutputLength\" is out of range. It must be >= 1 and <= 9007199254740991. Received 0"}
   ```
   or
   ```
   {"error":"Error: zip contains too many files"}
   ```

## Root cause

`apps/daemon/src/claude-design-import.ts`:

```ts
// L11
const MAX_FILES = 500;

// L104-117 readEntryBody (...):
//   compressed = zip.slice(bodyStart, bodyEnd)  // bodyEnd-bodyStart === entry.compressedSize
//   ...
return inflateRawSync(compressed, { maxOutputLength: entry.uncompressedSize });
```

Node 24's `zlib` option validator now requires `maxOutputLength >= 1`, so the value `0` (legal in earlier versions) throws synchronously.

## Suggested fix

Two-line patch in `apps/daemon/src/claude-design-import.ts`:

```diff
-const MAX_FILES = 500;
+const MAX_FILES = 5000;
@@
-  if (entry.method === 0) return Buffer.from(compressed);
-  return inflateRawSync(compressed, { maxOutputLength: entry.uncompressedSize });
+  if (entry.method === 0) return Buffer.from(compressed);
+  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
+  return inflateRawSync(compressed, { maxOutputLength: entry.uncompressedSize });
```

Bytes-per-file and total-bytes guards already cap risk; `MAX_FILES = 5000` is closer to what real design-system / prototype exports look like.

### Related UX gap (separate but worth fixing together)

The web layer should surface the daemon's `error` field instead of swallowing it:

`apps/web/src/state/projects.ts` (`importClaudeDesignZip`): return the parsed JSON (including `error`) instead of `null` on `!resp.ok`, and have `apps/web/src/App.tsx`'s `handleImportClaudeDesign` push the message into a toast. Without this, any future failure in the importer remains invisible to the user.

## Environment

- Node `v24.15.0`
- `pnpm 10.33.2`
- macOS (darwin-arm64)
- Repo HEAD: `c69dee7` on `main`

## Test asset

Reproduced consistently with a 4.4 MB / 561-file zip containing 74 HTML files at `preview/**/*.html`, plus one zero-byte `docs/components/chip-removable.md`. Both failure modes were confirmed: the zero-byte entry triggers (1), and even after patching (1), (2) blocks the import. Both fixes together let the import succeed end-to-end.

gabrielvaz:
Fixes #590.

## Summary

- **Skip `inflateRawSync` when central-directory `uncompressedSize === 0`.** Node 24 rejects `{ maxOutputLength: 0 }` with `ERR_OUT_OF_RANGE`, so any zip containing an empty file (or a streaming entry that records its size only in the data descriptor — common with macOS Archive Utility) silently aborted the entire import. The fix returns `Buffer.alloc(0)` for zero-length entries before reaching `inflateRawSync`.
- **Raise `MAX_FILES` from 500 to 5000.** Real-world design-system exports commonly exceed 500 files (the bug report's reproducer is 561). The pre-existing `MAX_TOTAL_BYTES = 100 MB` and `MAX_FILE_BYTES = 25 MB` guards continue to cap pathological inputs.
- **Add regression tests** that build zip fixtures in-process and exercise both fixes:
  - zero-byte deflate entry,
  - central directory advertising `uncompressedSize=0` while the local header is correct,
  - 600-file zip (would have failed under the old ceiling).

## Test plan

- [x] `pnpm --filter @open-design/daemon test` (761 tests pass, including the 3 new ones)
- [x] End-to-end: `curl -F file=@farol-ds.zip http://127.0.0.1:<port>/api/import/claude-design` against a 4.4 MB / 561-file design-system export — pre-fix returns `RangeError ... maxOutputLength ... Received 0`, post-fix returns the new project + `entryFile`.

## Out of scope (follow-up worth doing)

The web UI swallows the daemon's `error` field (`apps/web/src/state/projects.ts` returns `null` on `!resp.ok`, `apps/web/src/App.tsx`'s `handleImportClaudeDesign` does `if (!result) return;`), which made this bug invisible. Surfacing the error in a toast is a separate, small change discussed in #590; happy to follow up if useful.

### DIFF
--- a/apps/daemon/src/claude-design-import.ts
+++ b/apps/daemon/src/claude-design-import.ts
@@ -25,13 +25,21 @@ export async function importClaudeDesignZip(zipPath, projectDir) {
     if (entry.uncompressedSize > MAX_FILE_BYTES) {
       throw new Error(`zip file too large: ${relPath}`);
     }
-    totalBytes += entry.uncompressedSize;
-    if (totalBytes > MAX_TOTAL_BYTES) throw new Error('zip is too large');
 
+    // Decode first; the central directory's uncompressedSize is unreliable for
+    // streaming/data-descriptor zips (it can read 0 even when the payload
+    // carries real data). The inflate cap and the post-decode size checks below
+    // are authoritative.
     const body = readEntryBody(zip, entry);
-    if (body.length !== entry.uncompressedSize) {
+    if (body.length > MAX_FILE_BYTES) {
+      throw new Error(`zip file too large: ${relPath}`);
+    }
+    if (entry.uncompressedSize > 0 && body.length !== entry.uncompressedSize) {
       throw new Error(`zip entry size mismatch: ${relPath}`);
     }
+    totalBytes += body.length;
+    if (totalBytes > MAX_TOTAL_BYTES) throw new Error('zip is too large');
+
     files.push({ path: relPath, body });
   }
 
@@ -113,8 +121,16 @@ function readEntryBody(zip, entry) {
   if (bodyEnd > zip.length) throw new Error(`zip entry exceeds archive: ${entry.name}`);
   const compressed = zip.slice(bodyStart, bodyEnd);
   if (entry.method === 0) return Buffer.from(compressed);
-  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
-  return inflateRawSync(compressed, { maxOutputLength: entry.uncompressedSize });
+  // A genuinely empty deflate payload would still occupy at least the BFINAL
+  // marker; an entirely missing payload cannot be inflated, so treat it as
+  // empty rather than handing a zero-length buffer to zlib.
+  if (compressed.length === 0) return Buffer.alloc(0);
+  // When the central directory advertises 0 (streaming zips with data
+  // descriptors), fall back to the per-file ceiling so legitimate non-empty
+  // payloads decode instead of being silently truncated. The post-decode
+  // checks in the caller enforce MAX_FILE_BYTES and total-bytes limits.
+  const cap = entry.uncompressedSize > 0 ? entry.uncompressedSize : MAX_FILE_BYTES;
+  return inflateRawSync(compressed, { maxOutputLength: cap });
 }
 
 function sanitizeZipPath(name) {
--- a/apps/daemon/src/claude-design-import.ts
+++ b/apps/daemon/src/claude-design-import.ts
@@ -8,7 +8,7 @@ const EOCD_SIG = 0x06054b50;
 const CENTRAL_SIG = 0x02014b50;
 const LOCAL_SIG = 0x04034b50;
 
-const MAX_FILES = 500;
+const MAX_FILES = 5000;
 const MAX_TOTAL_BYTES = 100 * 1024 * 1024;
 const MAX_FILE_BYTES = 25 * 1024 * 1024;
 
@@ -113,6 +113,7 @@ function readEntryBody(zip, entry) {
   if (bodyEnd > zip.length) throw new Error(`zip entry exceeds archive: ${entry.name}`);
   const compressed = zip.slice(bodyStart, bodyEnd);
   if (entry.method === 0) return Buffer.from(compressed);
+  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
   return inflateRawSync(compressed, { maxOutputLength: entry.uncompressedSize });
 }
 
--- a/apps/daemon/tests/claude-design-import.test.ts
+++ b/apps/daemon/tests/claude-design-import.test.ts
@@ -1,0 +1,142 @@
+import { mkdtempSync, rmSync, writeFileSync, readFileSync, readdirSync } from 'node:fs';
+import path from 'node:path';
+import os from 'node:os';
+import { deflateRawSync } from 'node:zlib';
+import { describe, expect, it } from 'vitest';
+import { importClaudeDesignZip } from '../src/claude-design-import.js';
+
+function buildZip(
+  entries: { name: string; body: Buffer; method?: 0 | 8; falsifyCentralUncompressed?: boolean }[],
+): Buffer {
+  const localChunks: Buffer[] = [];
+  const centralChunks: Buffer[] = [];
+  let offset = 0;
+
+  for (const entry of entries) {
+    const method = entry.method ?? 8;
+    const nameBuf = Buffer.from(entry.name, 'utf8');
+    const compressed = method === 0 ? entry.body : deflateRawSync(entry.body);
+    const crcBuf = Buffer.alloc(4);
+    // CRC isn't validated by the importer; zero is fine for this test fixture.
+    crcBuf.writeUInt32LE(0, 0);
+
+    const local = Buffer.alloc(30);
+    local.writeUInt32LE(0x04034b50, 0);
+    local.writeUInt16LE(20, 4); // version needed
+    local.writeUInt16LE(0, 6); // flags
+    local.writeUInt16LE(method, 8);
+    local.writeUInt16LE(0, 10); // mod time
+    local.writeUInt16LE(0, 12); // mod date
+    crcBuf.copy(local, 14);
+    local.writeUInt32LE(compressed.length, 18);
+    local.writeUInt32LE(entry.body.length, 22);
+    local.writeUInt16LE(nameBuf.length, 26);
+    local.writeUInt16LE(0, 28); // extra len
+    localChunks.push(local, nameBuf, compressed);
+
+    const central = Buffer.alloc(46);
+    central.writeUInt32LE(0x02014b50, 0);
+    central.writeUInt16LE(20, 4); // version made by
+    central.writeUInt16LE(20, 6); // version needed
+    central.writeUInt16LE(0, 8); // flags
+    central.writeUInt16LE(method, 10);
+    central.writeUInt16LE(0, 12);
+    central.writeUInt16LE(0, 14);
+    crcBuf.copy(central, 16);
+    central.writeUInt32LE(compressed.length, 20);
+    // The central directory may legitimately advertise uncompressedSize=0 even when
+    // the local header has the real length (streaming zips with data descriptors).
+    // Reproduce that case explicitly when requested.
+    central.writeUInt32LE(
+      entry.falsifyCentralUncompressed ? 0 : entry.body.length,
+      24,
+    );
+    central.writeUInt16LE(nameBuf.length, 28);
+    central.writeUInt16LE(0, 30);
+    central.writeUInt16LE(0, 32);
+    central.writeUInt16LE(0, 34);
+    central.writeUInt16LE(0, 36);
+    central.writeUInt32LE(0, 38);
+    central.writeUInt32LE(offset, 42);
+    centralChunks.push(central, nameBuf);
+
+    offset += local.length + nameBuf.length + compressed.length;
+  }
+
+  const localBlob = Buffer.concat(localChunks);
+  const centralBlob = Buffer.concat(centralChunks);
+  const eocd = Buffer.alloc(22);
+  eocd.writeUInt32LE(0x06054b50, 0);
+  eocd.writeUInt16LE(0, 4);
+  eocd.writeUInt16LE(0, 6);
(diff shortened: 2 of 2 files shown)


lefarcen:
Hi @gabrielvaz! 🎉
Thanks for the contribution — clean fix for the Node 24 ERR_OUT_OF_RANGE issue plus raising the file ceiling.

— open-design team

apps/daemon/src/claude-design-import.ts
```diff
@@ -113,6 +113,7 @@ function readEntryBody(zip, entry) {
   if (bodyEnd > zip.length) throw new Error(`zip entry exceeds archive: ${entry.name}`);
   const compressed = zip.slice(bodyStart, bodyEnd);
   if (entry.method === 0) return Buffer.from(compressed);
+  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
```
lefarcen: **P2** The zero-size deflate fast path returns an empty buffer without validating the compressed payload. If a deflated entry advertises central-directory `uncompressedSize=0` while carrying non-empty or corrupt data, the later size check still passes against that same zero, so the importer can silently truncate the entry.

Consider validating the deflate stream with a minimum nonzero cap or rejecting the mismatch instead of bypassing inflate entirely.

apps/daemon/tests/claude-design-import.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      expect(result.entryFile).toBe('index.html');
+      expect(result.files.sort()).toEqual(['docs/empty.md', 'index.html']);
+      const empty = readFileSync(path.join(projectDir, 'docs/empty.md'));
+      expect(empty.length).toBe(0);
+    } finally {
+      rmSync(tmp, { recursive: true, force: true });
+    }
+  });
+
+  it('handles entries whose central directory advertises uncompressedSize=0', async () => {
```
lefarcen: **P2** This test does not exercise the scenario described in the PR body: the fixture body is `Buffer.alloc(0)`, so `falsifyCentralUncompressed` does not actually create a central-directory/local-header size mismatch. The assertion also only checks that the path is present, so a non-empty entry incorrectly imported as empty would not be caught.

Consider either: (a) building a fixture with a non-empty body + falsified central size, and assert the written content matches expected, or (b) adjusting the test description to match what it actually tests.

lefarcen (commented):
Hi @gabrielvaz — clean fix for the Node 24 crash + file ceiling. The regression tests cover the core failure paths. Found two non-blocking refinements for the edge-case handling:

apps/daemon/src/claude-design-import.ts
mrcfps: This skips decompression for every deflated entry whose central-directory `uncompressedSize` is `0`. That fixes truly empty files, but it also trusts the same possibly-wrong metadata this PR is trying to tolerate: when the local entry/compressed payload contains non-empty deflated data while the central directory reports `0`, the importer writes an empty file and the later `body.length !== entry.uncompressedSize` check still passes (`0 === 0`). That silently corrupts imported design assets instead of preserving them or rejecting the archive. Please only take the empty fast path when the compressed payload is known to represent an empty deflate stream, or inflate with a safe non-zero cap/derive the real size from supported ZIP metadata and then enforce the existing size limits. The regression test for falsified central size should use a non-empty body so this truncation path is covered.

mrcfps (changes requested):
@gabrielvaz thanks for the focused Node 24 import fix and for adding regression coverage around the ZIP importer. I found one correctness issue in the zero-size entry handling that can silently truncate non-empty deflated files, so I’m requesting changes before merge. Once that path distinguishes truly empty entries from mismatched metadata, the rest of the change looks nicely scoped. 🙏


apps/daemon/src/claude-design-import.ts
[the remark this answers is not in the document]
gabrielvaz: Fixed in 136c5b5. The fast path is gone: I now inflate with `maxOutputLength = MAX_FILE_BYTES` whenever the central directory advertises 0, and validate the actual decoded length against MAX_FILE_BYTES and the total-bytes budget post-decode. The strict `body.length !== uncompressedSize` equality check is skipped only when central was 0 (i.e., genuinely unknown). The new test `preserves the real payload when the central directory under-reports size to 0` asserts a non-empty body round-trips byte-for-byte through that path.

gabrielvaz: Addressed in 136c5b5. The fast path is removed; deflate entries now inflate with maxOutputLength = MAX_FILE_BYTES when central uncompressedSize is 0, and the per-file/total-byte limits are enforced against the actual decoded length. New test `rejects entries that decode larger than MAX_FILE_BYTES even when central size is 0` exercises that ceiling end-to-end.

apps/daemon/tests/claude-design-import.test.ts
gabrielvaz: Addressed in 136c5b5. The streaming-zip test (renamed to `preserves the real payload when the central directory under-reports size to 0`) now uses a non-empty body (`# streamed entry\n\n` + 4096 `x`s) and asserts the on-disk file equals the original buffer byte-for-byte. A second new test covers the oversized-payload-with-falsified-central case to ensure MAX_FILE_BYTES is still enforced.

apps/daemon/src/claude-design-import.ts
gabrielvaz: Thanks — addressed in 136c5b5. The fast path is gone; for deflate entries with central uncompressedSize=0 the importer now inflates with `maxOutputLength = MAX_FILE_BYTES` and trusts the actual decoded length. Per-file (`MAX_FILE_BYTES`) and total-bytes (`MAX_TOTAL_BYTES`) ceilings moved post-decode so they enforce against the real size. The strict `body.length !== uncompressedSize` equality check is now applied only when the central record is non-zero (i.e., trustworthy). The streaming-zip regression test was rewritten to use a non-empty body and now asserts the written content is byte-for-byte equal to the original.

gabrielvaz:
## Revision pushed: 136c5b5

Reviewers (@chatgpt-codex-connector, @lefarcen, @mrcfps) all converged on the same correctness bug in the original `Buffer.alloc(0)` fast path: it trusted the central directory's `uncompressedSize`, which is not authoritative for streaming/data-descriptor zips. That made the importer silently truncate non-empty entries whose central record reported 0, and the post-condition `body.length !== entry.uncompressedSize` check still passed because both sides were 0. Thanks for catching this before merge.

### What changed in this revision

`apps/daemon/src/claude-design-import.ts`:

1. **Removed the `Buffer.alloc(0)` fast path.** Deflate entries with `uncompressedSize === 0` now inflate with `maxOutputLength = MAX_FILE_BYTES`, so legitimate non-empty payloads decode in full instead of being truncated. The only short-circuit left is when the compressed payload itself is genuinely zero-length (no BFINAL byte), which cannot be inflated.
2. **Authoritative size = decoded length.** Per-file (`MAX_FILE_BYTES`) and total-bytes (`MAX_TOTAL_BYTES`) limits are enforced *after* decoding, against `body.length`. The strict equality check `body.length !== entry.uncompressedSize` is now applied only when the central record is non-zero (i.e., known-trustworthy).
3. **Empty-deflate edge case stays correct.** The originally-zero-byte file in the issue's repro (`docs/components/chip-removable.md`) decodes to 0 bytes via the normal path and is written as an empty file on disk — verified end-to-end against `farol-ds.zip`.

`apps/daemon/tests/claude-design-import.test.ts`:

- Renamed the streaming-zip test to `preserves the real payload when the central directory under-reports size to 0` and rebuilt the fixture to carry a non-empty body (`# streamed entry\n\n` + 4096 `x` chars). It now asserts the on-disk file equals the original buffer byte-for-byte — would have caught the truncation regression directly.
- Added `rejects entries that decode larger than MAX_FILE_BYTES even when central size is 0`: a streaming entry decoded just past the per-file ceiling must still be rejected. Confirms the limit is enforced via the actual decoded length, not the (untrusted) central record.
- Original 0-byte and >500-file regressions remain covered.

### Verification

- `pnpm --filter @open-design/daemon typecheck`: clean.
- `pnpm --filter @open-design/daemon exec vitest run tests/claude-design-import.test.ts`: 4 passed.
- End-to-end against the issue's 4.4 MB / 561-file repro: `POST /api/import/claude-design` returns 200, the originally zero-byte file is written as 0 bytes, and a sample of decoded `docs/components/*.md` files have the expected non-zero sizes (e.g. `accordion.md` = 845 bytes, `alert.md` = 607 bytes).
- Two unrelated `tests/connectors-routes.test.ts` cases occasionally fail under parallel load, but pass cleanly in isolation on both this branch and `main`; pre-existing flake, not introduced here.

lefarcen (commented):
Thanks for the quick fix, @gabrielvaz! Verified at 136c5b5:

✅ The `Buffer.alloc(0)` fast path is removed
✅ Deflate entries with central `uncompressedSize=0` now inflate with `maxOutputLength = MAX_FILE_BYTES` and validate the actual decoded length
✅ Per-file and total-byte limits enforce against real decoded size, not untrusted central metadata
✅ The streaming-zip test (`preserves the real payload when the central directory under-reports size to 0`) now uses a non-empty body and asserts byte-for-byte equality

Both P2 findings from the initial review are resolved. The import logic correctly handles the Node 24 regression and streaming-zip metadata mismatch without silent truncation.

mrcfps (approved):
@gabrielvaz I reviewed the current 136c5b5 diff for the daemon ZIP importer and the new regression tests. The previous central-directory size issue is addressed by inflating central-zero deflate entries with a safe cap, enforcing per-file/total limits on decoded bytes, and adding coverage for empty entries, streaming-size mismatches, oversized decoded payloads, and the raised file ceiling. Thanks for the quick, focused follow-up and the clear test coverage — this looks ready from my pass. 🙏


lefarcen:
Merged! 🎉 Thanks @gabrielvaz for the contribution to open-design. Looking forward to the next one!

leilei926524-tech:
Hi @gabrielvaz — welcome to Open Design contributors. 🎉

Your first merged PR (#591) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #587, opened 2026-05-05
feat(craft): accessibility-baseline module + opt-ins on dashboard, hr-onboarding, mobile-onboarding

MohamedAbdallah-14:
Module 3 of 5 in the behavioral craft series proposed in #501. Module 1 (`state-coverage`, #502) and Module 2 (`animation-discipline`, #515) merged earlier today.

## What's in the diff

`craft/accessibility-baseline.md` — 146 lines. Sections: legal floor by jurisdiction (EU EAA / US ADA Title II / ISO 40500:2025), color contrast (with APCA dual-stack note), touch targets, focus visibility, form input labels (with default error wiring), ARIA discipline, reduced motion + flashing (cross-ref to `animation-discipline.md`), native mobile parity, common mistakes.

Plus the matching `od.craft.requires` opt-ins:

- `dashboard` extends from `[state-coverage]` to `[state-coverage, accessibility-baseline]`.
- `mobile-onboarding` extends from `[state-coverage, animation-discipline]` to `[state-coverage, animation-discipline, accessibility-baseline]`.
- `hr-onboarding` adds `[accessibility-baseline]` fresh — heaviest form-and-error surface in the catalog.

`craft/README.md` Files-table gets the new row.

## Differentiation from prior art

The closest prior art files are `fecarrico/A11Y.md` (101★, MIT), `github/awesome-copilot agents/accessibility.agent.md` (32k★, MIT), and `Community-Access/accessibility-agents` (259★, MIT). Each covers a slice. None covers all the ground this baseline targets.

The differentiator that survived the corpus review: **native-mobile parity**. None of the three covers Flutter Semantics, Compose semantics, iOS UIKit/SwiftUI, or React Native labelling APIs. AI-generated mobile UI that mirrors web ARIA verbatim ships effectively unlabelled to TalkBack and VoiceOver — a baseline file is the right place to prevent that.

Secondary differentiator: jurisdictional legal-floor calibration. The corpus surfaced that "EAA = WCAG 2.2 AA" is wrong as of today — the OJ-cited harmonised standard is EN 301 549 v3.2.1, anchored to WCAG 2.1 — and that the ADA Title II 2026-04-24 deadline slipped a year via the 2026-04-20 IFR. None of the existing OSS prior art tracks these accurately.

## Folklore busts surfaced from the research corpus

A handful of widely-cited a11y claims that don't survive primary-source check now sit in the file's "Common mistakes" section:

[...]

PR #515, opened 2026-05-05
feat(craft): animation-discipline module + opt-ins on mobile-app, mobile-onboarding, gamified-app

MohamedAbdallah-14:
Module 2 of 5 in the behavioral craft series proposed in #501.

@mrcfps explicitly invited `animation-discipline.md` in the post-merge comment on #502 ("Looking forward to the follow-up PRs when you're ready, especially `animation-discipline.md` and the rest of the sequence").

## Differentiation from prior art

The closest existing craft is LottieFiles' `motion-design-skill` (96 stars, MIT, actively maintained) — strong vocabulary on Disney 12 and pillar/layer framing, **zero empirical anchors**. This module's value isn't a new rule set, it's a citation-grounded one that survives primary-source check.

To draft this, I built a research corpus first — 12 breadth surveys + 18 depth source-reads. The depth pass surfaced that several widely-cited animation rules don't trace cleanly to their alleged primary sources:

- "Skeleton screens feel 11% faster" — Harrison/Yeo/Hudson CHI 2010 measured *backwards-decelerating ribbed determinate progress bars* (n=16). The induced-motion mechanism doesn't transfer to skeletons.
- "Heer & Robertson recommend 300-1000 ms eased transitions" — they tested 1.25 s and 2 s. Their recommendation is "~1 second per stage."
- "Doherty Threshold = 400 ms" — the 1982 paper does not contain the string "400". Real lowest measured threshold: 300 ms.
- "M3 standard easing = `(0.4, 0, 0.2, 1)`" — that's M2; M3 standard is `(0.2, 0, 0, 1)`. M3 preserves M2 as `legacy`.
- "Apple `.snappy = 0.25 s, .smooth = 0.35 s`" — Apple's docs assign all three SwiftUI presets a 0.5 s base, differing only in bounce.

These are now in the file's "Common mistakes (lint these)" section.

## What's in the diff

- `craft/animation-discipline.md` — 130 lines. Slightly over the 80-110 line craft target; the trade-off is citation density. Sections: when motion earns its place (Tversky), duration thresholds (150 ms cross-system convergence), curve vs spring, reduced motion (with WCAG 2.2.2 vs 2.3.3 calibration), cross-platform handoff (iOS/Android/web), common mistakes (the folklore busts).
- `craft/README.md` — Files-table row for `animation-discipline`.
[...]

PR #502, opened 2026-05-05
feat(craft): state-coverage module + opt-ins on dashboard, mobile-app, kanban-board

MohamedAbdallah-14:
Reference implementation for #501 — module 1 of 5 proposed behavioral craft files.

State coverage is the most reliable AI-design failure: agents ship only the populated state. This adds `craft/state-coverage.md` covering the five required states (loading, empty, error, populated, edge), three form-specific states, ARIA/focus rules, error-and-empty-state composition, and loading-duration thresholds. Sources are public — WCAG 2.2, NN/g, Material Design 3, Apple HIG, Baymard.

Filed as Draft so the broader sequence framing in #501 can be settled before merge. Happy to reorder, hold off, or adjust scope based on your read.

## What's in the diff

- `craft/state-coverage.md` — new craft module, 108 lines, matches the format of `typography.md` / `color.md` / `anti-ai-slop.md`.
- `craft/README.md` — one row added to the Files table registering the `state-coverage` slug.
- `skills/dashboard/SKILL.md`, `skills/mobile-app/SKILL.md`, `skills/kanban-board/SKILL.md` — `od.craft.requires: [state-coverage]` added under the existing `od:` block. These three skills produce stateful UI surfaces and obviously need the rules; deck/ppt/static skills do not opt in.

No company-named examples in the craft body. No "always/never" without scope. ARIA section adapted from canonical WCAG 2.2 patterns.

## Test plan

- [ ] `pnpm install && pnpm typecheck` clean (no type surface touched, but verify nothing else regressed)
- [ ] Run a generation against `dashboard` with the skill plus DESIGN.md and confirm `state-coverage` body is injected between DESIGN.md and the skill body in the system prompt
- [ ] Same for `mobile-app` and `kanban-board`
- [ ] Visual sanity: regenerate the existing `example.html` for the three skills and verify the output now renders empty / error / loading affordances rather than only populated
- [ ] Lint: confirm `od.craft.requires: [state-coverage]` parses cleanly through the existing skill registry

## Open questions

Both sit in #501 — sequence ordering vs motion/icons, and whether community craft contributions are welcome at all per CONTRIBUTING.md. PR stays Draft until those land.

Issue #501, opened 2026-05-05
The agent knows your colors. It doesn't know how to design.


MohamedAbdallah-14:
I've used Claude Design and a couple of other AI design tools. The output keeps looking the same way no matter who I'm building for. Tool runs on Claude, designs feel Claude. Tool runs on Codex, designs feel Codex. Brand colors change. The model doesn't.

What's actually wrong: the UI is technically correct, the UX isn't. Horizontal app bar on a mobile screen. Animations on every element that moves. No empty / error / loading states. The same hero-features-pricing-FAQ skeleton on every landing page. `DESIGN.md` tells the agent which colors and fonts the brand uses but never tells it *how to design* — when to pick a sheet over a modal, when to animate, when not to, what makes a touch target wrong on mobile, what an error state has to contain.

In my own setup I worked around it by building design knowledge as separate modules. Mobile rules. Web rules. Animation rules. Accessibility floors. RTL. Anti-patterns where the UI looks good but the UX is bad. The agent stops *thinking* through design decisions and starts following rules a competent designer would follow. That changed the output more than any prompt-engineering pass I ran before it.

OD's `craft/` axis is already designed for this. PR #225 added typography, color, anti-ai-slop. The README mentions motion, icons, craft-details coming next. I'd like to contribute the next batch, and check the framing fits before I draft the rest.

## What I'm proposing

Five behavioral craft modules, ~80–110 lines each, matching the existing craft style. Sources are public — WCAG 2.2, NN/g, Material Design 3, Apple HIG, Baymard.

- `state-coverage.md` — every interactive surface must render five states (loading, empty, error, populated, edge), and forms add three more (untouched, dirty, submitted-pending). Most reliable AI-design failure I've seen: the agent ships only the populated state.
- `animation-discipline.md` — when motion serves the user vs decorates. Timing thresholds. Easing. `prefers-reduced-motion` as a hard floor. The "everything Springs" anti-pattern.
- `accessibility-baseline.md` — WCAG 4.5:1 / 3:1 contrast floors, 44×44 touch targets, focus-visible, ghost-label.
- `rtl-and-bidi.md` — full RTL mirroring matrix. CSS logical properties. BiDi handling. As far as I can tell, no other AI design tool documents this at all.
- `form-validation.md` — validate on blur not on first keystroke, error specificity, focus moves to the first error field, input survives a failed submit.

[...]

PR #595, opened 2026-05-06
craft: add rtl-and-bidi so OD artifacts don't break for Arabic / Hebrew / Persian users

MohamedAbdallah-14:
Module 4 of 5 in the behavioral craft series proposed in #501. Modules 1-3 merged: state-coverage (#502), animation-discipline (#515), accessibility-baseline (#587).

## Why

A skill emits a SaaS landing or a financial report. The brand `DESIGN.md` says nothing about RTL because it doesn't have to — the brand visual language is direction-neutral. The artifact ships with `margin-left: 24px`, `text-align: left`, anchored arrows that point the same way in Arabic as in English, and a search icon that doesn't flip. None of that registers as a bug in LTR review. It only breaks once an Arabic / Hebrew / Persian user opens it.

Two things make this a craft concern, not a per-skill concern:

1. **The rules are universal.** `margin-inline-start` resolves to the start edge in any direction. `<bdi>` works regardless of brand. The mirroring decision for arrows is the same on every site. There is no brand-specific RTL.
2. **The traps are non-obvious.** alreq treats CSS `letter-spacing` as a boundary concept that breaks Arabic cursive joining. `unicode-bidi: plaintext` is broken in WebKit ([Bugzilla #50949](https://bugs.webkit.org/show_bug.cgi?id=50949), still open as of 2026-05). Tailwind v4.0 folded the inline-axis logical utilities into core; the `tailwindcss-rtl` plugin is now obsolete, and `[dir="rtl"]:` overrides for spacing are an anti-pattern. None of these are findable from a generic LTR review pass.

## Primary sources behind every rule

- Unicode UAX #9 revision 51 (September 2025), version-stamped against Unicode 17.0; `max_depth = 125` is permanently locked forward.
- CSS Logical Properties Level 1 (Baseline Widely Available — Chrome 87, Safari 14.1, Firefox 66; ≥95% global support as of 2026-05).
- Tailwind v4.0 (2025-01-22) and v4.2 (2026-02-18) changelogs — the inline-axis utilities (`ms-*`, `me-*`, `ps-*`, `pe-*`, `start-*`, `end-*`) are now core; v4.2 added the block-axis set and renamed inset utilities to `inset-s-*` / `inset-e-*`.
- W3C Arabic Layout Requirements (alreq) on letter-spacing as a boundary concept.
- Material 3 RTL guidance + Apple HIG internationalization for the mirroring decisions.
- Per-platform direction primitives verified against UIKit, SwiftUI, Android Compose, Flutter, and React Native current docs.

## Prior art differentiation

Three OSS skills/agents in the RTL space were checked for license and scope:

| Repo | Stars | Scope | Why it doesn't fit |
|---|---|---|---|
| `idanlevi1/rtlify` | 5 | LTR-web-skewed | No native-mobile primitives; MIT but tutorial-shaped |
| `skills-il/localization` | 7 | i18n + l10n umbrella | RTL is one chapter of many; not a craft module |
| `unicode-org/cldr` rules | n/a | data tables | Reference data, not generator guidance |

This file's niche: framework-agnostic, Apache-2.0 (matching OD), aligned with UAX #9 rev 51, and explicit on the five direction primitives (UIKit `semanticContentAttribute`, SwiftUI `.environment(\.layoutDirection)`, Compose `LocalLayoutDirection`, Flutter `Directionality`, RN `I18nManager.forceRTL`).

## What's in the file

- **Logical properties first.** Inline-axis cheat sheet + Tailwind v4 callout (don't write `[dir="rtl"]:` overrides for spacing on v4).
- **Bidirectional text.** `<bdi>` for HTML, U+2068 / U+2069 for plain text. Common myths busted (`unicode-bidi: plaintext` doesn't work cross-browser; U+2066-U+2069 are interoperable everywhere modern).
- **Mirroring decisions.** Three lists — must mirror, must not mirror, single live cross-platform conflict (Apple flips the search icon, Material 3 doesn't). Sources: Material 3 + Apple HIG.
- **Typography rules anchored here.** Two RTL-coupled rules that cause layout breakage: never `letter-spacing` on Arabic runs (cursive joining), and 14-18 px body with 1.5-1.75 line-height for harakat clearance. Full Arabic / Hebrew typography deferred to a future `craft/arabic-hebrew-typography.md`.
- **Native mobile RTL parity.** Per-platform direction primitive + spacing primitive table.
- **Forms in RTL.** Three rules: `<input dir="auto">` for unknown direction, force LTR on intrinsically-LTR fields (email / URL / phone / IBAN / credit-card), wrap rendered values in `<bdi>`.
- **Common mistakes.** 9 lint-able items including the WebKit `unicode-bidi: plaintext` myth, the stale "Tailwind needs `tailwindcss-rtl`" claim, and the alreq letter-spacing trap.

## Skill opt-ins

- `blog-post` — long-form text, may render Arabic / Hebrew runs.
- `docs-page` — LTR code islands inside RTL prose, the bidi-isolation case `<bdi>` was made for.
- `finance-report` — numerals, IBAN, currency in mixed-direction cells.

Three skills, all clear fits. README updated.

## Research and review process

Followed the same workflow as #587: corpus of 24 depth files (D10-D15 cover RTL specifically) read through before drafting, three review loops with Claude CLI Opus 4.7 xhigh effort. Loop 1 caught five revisions including spinning out the typography section, compressing WebKit prose, trimming the mistakes list 12→9, and dropping a stale alreq letter-spacing → text-spacing rename claim. Loop 2 caught one blocking factual error (an incorrect SwiftUI 4 version claim — SwiftUI's `EdgeInsets` has used `leading`/`trailing` since SwiftUI 1.0). Loop 3 said ship.

## Wazir attribution

Some of the corpus and the framing of "what counts as a behavioral craft module" came from research originally done on Wazir, an OSS project I parked after three months of development when results from agent-collaboration complexity didn't match the cost. The research itself was solid; the architecture wasn't. Republishing the research-grade pieces under OD's craft series is the highest-leverage place for that work to land.

Refs #501.

### DIFF
--- a/craft/rtl-and-bidi.md
+++ b/craft/rtl-and-bidi.md
@@ -181,5 +181,5 @@ icons, platform-specific placement).
 
 **HTML semantics:**
 
-- Reaching for `unicode-bidi: plaintext` when `<bdi dir="auto">` (or `<bdi>` with implicit auto-detect) is the markup form of the same idea. UAX #9 §2.7: prefer markup over CSS or control characters when markup is available.
+- Reaching for CSS bidi controls (`unicode-bidi: isolate` / `plaintext` / `embed`) for inline runs when `<bdi>` or a `dir`-bearing element does the job. Prefer semantic isolation in HTML for inline content; `unicode-bidi: plaintext` operates on a different surface (it changes how base direction is determined for each plaintext paragraph in a block) and should only be used when that block-level paragraph behavior is explicitly required and tested. The two are not drop-in equivalents — don't lint one as a replacement for the other.
 - Bare `<bdi>` around phone / IBAN / card numbers in an RTL paragraph. First-strong detection on weak/neutral characters is unreliable; force `dir="ltr"` explicitly.
--- a/craft/rtl-and-bidi.md
+++ b/craft/rtl-and-bidi.md
@@ -6,9 +6,26 @@ how that language behaves when the script reads from the right or
 mixes direction within a line.
 
 > Grounded in primary sources: Unicode UAX #9 revision 51 (Sept 2025)
-> + Unicode 17.0, CSS Logical Properties Level 1, Tailwind v4.0/v4.2
-> changelogs, W3C alreq, Material 3 RTL guidance, Apple HIG
-> internationalization, WebKit Bugzilla 50949.
+> + Unicode 17.0, CSS Logical Properties Level 1, HTML Living Standard
+> (`dir`, `<bdi>`), Tailwind v4.0/v4.2 changelogs, W3C alreq,
+> Material 3 RTL guidance, Apple HIG internationalization.
+
+## Base direction and language
+
+Every full-page RTL artifact needs `<html dir="rtl" lang="ar">` (or
+the matching `lang` for Hebrew, Persian, Urdu). The `lang` attribute
+drives font-stack selection, hyphenation, locale-aware speech
+synthesis, and search-engine indexing — `dir` alone isn't enough.
+Three patterns cover the common cases:
+
+- **Full-page RTL.** `<html dir="rtl" lang="ar">`. Everything inside inherits.
+- **Mixed-language subtree.** Nest `<section dir="ltr" lang="en">…</section>` (or vice versa) when an embedded block uses a different script. Code samples, English citations, foreign brand names.
+- **User-generated content of unknown direction.** `dir="auto"` on the paragraph. The browser resolves direction from the first strong directional character in the run.
+
+Setting `lang` without `dir` is fine when the language has only one
+direction in practice (English doesn't need `dir="ltr"`). Setting
+`dir` without `lang` is rarely correct — at minimum drop the
+appropriate ISO-639 tag in.
 
 ## Logical properties first
 
@@ -42,22 +59,22 @@ Don't write `[dir="rtl"]:` overrides for spacing on Tailwind v4.
 UAX #9 rev 51 (Sept 2025) is a version stamp for Unicode 17.0. No
 algorithm change; `max_depth = 125` is permanently locked forward.
 
-The five isolate-and-formatting characters: U+2066 LRI, U+2067 RLI,
-U+2068 FSI, U+2069 PDI, U+202C PDF.
+UAX #9 defines two distinct families of bidi formatting characters
+that solve different problems:
 
-**Use `<bdi>` in HTML and the U+2068 / U+2069 control pair for plain
+- **Isolate controls** (modern, prefer these): U+2066 LRI, U+2067 RLI, U+2068 FSI — opened with these, all closed with U+2069 PDI. An isolated run does not affect, and is not affected by, the surrounding paragraph's bidi resolution. Use FSI when the embedded run's direction is unknown ahead of time.
+- **Embedding / override controls** (legacy): U+202A LRE, U+202B RLE, U+202D LRO, U+202E RLO — all closed with U+202C PDF. These nest within the surrounding paragraph rather than isolating from it; LRO/RLO additionally force a direction onto neutral characters. Newer code should use isolates; touch embeddings only when interoperating with text from systems that emit them.
+
+**Use `<bdi>` in HTML and the U+2068 / U+2069 isolate pair for plain
 text.** UAX #9 §2.7: *"where available, markup should be used instead
 of the explicit formatting characters."* `<bdi>` has been Baseline
-Widely Available since January 2020.
+Widely Available since January 2020. Choose the markup form whenever
+you have markup; reach for the control characters only in plain-text
+contexts (logs, plain-text emails, terminal output).
 
-`dir="auto"` lets the browser detect direction from the first strong
-directional character — useful for user-generated content where the
-direction isn't known up front.
-
-U+2066-U+2069 are interoperable across modern Chrome, Firefox, and
-Safari. The remaining gap is the CSS form `unicode-bidi: plaintext`
-([WebKit Bugzilla #50949](https://bugs.webkit.org/show_bug.cgi?id=50949),
-still open as of 2026-05). Don't rely on it as a primary mechanism.
+`dir="auto"` on a paragraph or `<bdi>` lets the browser detect
+direction from the first strong directional character. Best for
+user-generated content where direction isn't known at author time.
 
 ## What mirrors and what doesn't
 
@@ -68,7 +85,7 @@ Material 3 RTL guidance and Apple HIG internationalization.
 
 - Directional arrows (back / forward / next / previous), navigation rail position, tab order, slider fill direction, progress-bar fill, calendar-grid weekday order.
 - Checkbox-and-label position. Label sits to the right in LTR, to the left in RTL.
-- Phone-number and IBAN affordances when the surrounding paragraph is RTL but the value itself is LTR — wrap the value in `<bdi>` so the digits don't reflow.
+- Phone-number and IBAN affordances when the surrounding paragraph is RTL but the value itself is LTR — wrap the value in `<bdi dir="ltr">` (or `<span dir="ltr">`) so the digits don't reflow. Bare `<bdi>` is not enough: phone numbers and account numbers contain mostly weak / neutral characters, so first-strong direction detection is unreliable. Force LTR explicitly.
 
 **Must not mirror:**
 
@@ -104,7 +121,9 @@ fallback chains) belongs in a future `craft/arabic-hebrew-typography.md`.
 ## Native mobile RTL parity
 
 Web RTL handling does not auto-translate to mobile. Each platform has
-its own direction primitive.
+its own direction primitive. Skills that emit web-only artifacts can
+skim this section; it's the entry point for skills that ship to
+mobile (mobile-onboarding, mobile-app, etc.).
 
 | Platform | Direction primitive | Spacing |
 |---|---|---|
@@ -125,16 +144,31 @@ Form fields commonly mix scripts. Three rules cover most of it.
 
 - **`<input dir="auto">`** for any field whose value's direction is uncertain (search boxes, comment fields, free-text inputs). The browser detects from the first strong directional character.
 - **Force LTR on intrinsically-LTR fields** even inside an RTL paragraph: email, URL, phone, IBAN, credit-card. `<input type="email" dir="ltr">`.
-- **Wrap rendered values in `<bdi>`** when displaying mixed-script content (a username inside a paragraph, a model number inside a description). Stops the surrounding direction from rearranging the embedded value.
+- **Wrap rendered values in `<bdi>`** when displaying mixed-script content (a username inside a paragraph, a model number inside a description). Stops the surrounding direction from rearranging the embedded value. For values whose direction is fixed and weak-character-heavy (phone, IBAN, card number), use `<bdi dir="ltr">` rather than bare `<bdi>` so first-strong detection doesn't misclassify.
 
 ## Common mistakes (lint these)
 
-- Hardcoded `left` / `right` / `text-align: left` in new CSS. Every hardcoded directional property is a bug in RTL.
-- "Use `text-justify: kashida` for Arabic" — no browser implements it. The standard CSS `justify` form adds inter-word spacing and looks unnatural; kashida is the right form, but you can't ship it on the web.
+Mechanically lintable items can be flagged from CSS / source alone.
+Script-aware items need to detect Arabic / Hebrew runs in the
+rendered text and have legitimate exceptions (chart axes, physical
+icons, platform-specific placement).
+
+**Mechanically lintable:**
+
+- Hardcoded `left` / `right` / `text-align: left` in new CSS — bug for any layout that may render RTL. Exceptions: chart x-axes, physical-object icons, platform-pinned UI like a status-bar clock. Lint with an allow-list rather than blanket banning.
 - "Tailwind v4.2 logical-utility rename is `inline-s-*` / `inline-e-*`" — wrong family. Those are size utilities. The inset rename is `inset-s-*` / `inset-e-*`.
 - "WebKit doesn't support U+2066-U+2069" — wrong, they're interoperable across modern browsers. The "still missing" claim traces to a stale 2015 W3C test snapshot.
-- "Use `unicode-bidi: plaintext`" — broken in WebKit ([Bugzilla #50949](https://bugs.webkit.org/show_bug.cgi?id=50949)). Use `<bdi>` for HTML and U+2068 / U+2069 for plain text.
-- Italics on Arabic or Hebrew text. Neither script has an italic tradition.
-- CSS `letter-spacing` applied to Arabic. Breaks cursive joining.
-- Lorem Ipsum used for RTL prototyping. Arabic word lengths, connection behaviors, and vertical extents differ; use real Arabic / Hebrew text.
+- Setting `dir="rtl"` without `lang="ar"` (or matching). Lint together; `dir` alone misses the font-stack and locale path.
 - Flutter `EdgeInsets.left/right` in code that needs to render RTL. Use `EdgeInsetsDirectional.start/end`.
+
+**Needs script detection (will false-positive without it):**
+
+- "Use `text-justify: kashida` for Arabic" — no browser implements it. CSS `text-align: justify` adds inter-word spacing and looks unnatural in Arabic; kashida elongation is the correct form, but it isn't shippable on the web today.
+- Italics on Arabic or Hebrew text. Neither script has an italic tradition.
+- CSS `letter-spacing` applied to Arabic. Breaks cursive joining (alreq treats it as a boundary concept, not a uniform tracking value).
+- Lorem Ipsum used for RTL prototyping. Arabic word lengths, connection behaviors, and vertical extents differ; use real Arabic / Hebrew text.
+
+**HTML semantics:**
+
+- Reaching for `unicode-bidi: plaintext` when `<bdi dir="auto">` (or `<bdi>` with implicit auto-detect) is the markup form of the same idea. UAX #9 §2.7: prefer markup over CSS or control characters when markup is available.
+- Bare `<bdi>` around phone / IBAN / card numbers in an RTL paragraph. First-strong detection on weak/neutral characters is unreliable; force `dir="ltr"` explicitly.
--- a/craft/rtl-and-bidi.md
+++ b/craft/rtl-and-bidi.md
@@ -94,7 +94,8 @@ Material 3 RTL guidance and Apple HIG internationalization.
 
 **Must mirror:**
 
-- Directional arrows (back / forward / next / previous), navigation rail position, tab order, slider fill direction, progress-bar fill, calendar-grid weekday order.
+- Directional arrows (back / forward / next / previous), navigation rail position, tab order, calendar-grid weekday order.
+- Slider fill direction and **non-media** progress-bar fill (a download progress bar, a form-completion bar, an upload status). Media scrubbers stay LTR — see the Media row below.
 - Checkbox-and-label position. Label sits to the right in LTR, to the left in RTL.
 - Phone-number and IBAN affordances when the surrounding paragraph is RTL but the value itself is LTR — wrap the value in `<bdi dir="ltr">` (or `<span dir="ltr">`) so the digits don't reflow. Bare `<bdi>` is not enough: phone numbers and account numbers contain mostly weak / neutral characters, so first-strong direction detection is unreliable. Force LTR explicitly.
 
@@ -102,7 +103,7 @@ Material 3 RTL guidance and Apple HIG internationalization.
 
 - Clock faces. Clockwise is universal.
 - Circular refresh / sync / reload icons. Same reason.
-- Media playback controls (play / pause / fast-forward / rewind). They represent tape direction, not reading direction.
+- Media playback controls (play / pause / fast-forward / rewind) **and the media scrubber / progress timeline**. They represent tape direction, not reading direction.
 - Charts and graphs. X-axis stays mathematical, not linguistic.
 - Photographs, brand logos, physical-object icons (camera, keyboard, headphones). Identity over direction.
 
--- a/craft/rtl-and-bidi.md
+++ b/craft/rtl-and-bidi.md
@@ -22,9 +22,12 @@ Three patterns cover the common cases:
 - **Mixed-language subtree.** Nest `<section dir="ltr" lang="en">…</section>` (or vice versa) when an embedded block uses a different script. Code samples, English citations, foreign brand names.
 - **User-generated content of unknown direction.** `dir="auto"` on the paragraph. The browser resolves direction from the first strong directional character in the run.
 
-Setting `lang` without `dir` is fine when the language has only one
-direction in practice (English doesn't need `dir="ltr"`). Setting
-`dir` without `lang` is rarely correct — at minimum drop the
+Setting `lang` without `dir` is fine **at the document root in a
+default-LTR page** — English doesn't need `dir="ltr"` there because
+the bidi base direction is already LTR. Inside any opposite-direction
+ancestor, `lang` does not reset the inherited base direction, so set
+both `lang` and `dir` on the subtree (`<section dir="ltr" lang="en">`).
+Setting `dir` without `lang` is rarely correct — at minimum drop the
 appropriate ISO-639 tag in.
 
 ## Logical properties first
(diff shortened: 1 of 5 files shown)


lefarcen:
Hi @MohamedAbdallah-14! 🎉
Thanks for the contribution — this craft module addresses a real gap in RTL support that affects Arabic/Hebrew/Persian users.

— open-design team

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,140 @@
...
+of the explicit formatting characters."* `<bdi>` has been Baseline
+Widely Available since January 2020.
+
+`dir="auto"` lets the browser detect direction from the first strong
+directional character — useful for user-generated content where the
+direction isn't known up front.
+
+U+2066-U+2069 are interoperable across modern Chrome, Firefox, and
+Safari. The remaining gap is the CSS form `unicode-bidi: plaintext`
+([WebKit Bugzilla #50949](https://bugs.webkit.org/show_bug.cgi?id=50949),
```
lefarcen: P2: WebKit Bugzilla #50949 shows RESOLVED FIXED, not "still open as of 2026-05". MDN currently treats `unicode-bidi` as Baseline Widely Available. The practical recommendation (prefer `<bdi>` over CSS) is still sound, but the rationale should be "prefer semantic HTML over CSS for authoring" rather than claiming WebKit is broken.

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,140 @@
...
+utilities: `start-*` / `end-*` are deprecated (still work) in favor
+of `inset-s-*` / `inset-e-*`. The `tailwindcss-rtl` plugin is obsolete.
+Don't write `[dir="rtl"]:` overrides for spacing on Tailwind v4.
+
+## Bidirectional text
+
+UAX #9 rev 51 (Sept 2025) is a version stamp for Unicode 17.0. No
+algorithm change; `max_depth = 125` is permanently locked forward.
+
+The five isolate-and-formatting characters: U+2066 LRI, U+2067 RLI,
```
lefarcen: P2: This section conflates isolates and formatting controls. U+202C PDF is not an isolate terminator — it closes embeddings/overrides. For plain text: known-LTR runs use LRI...PDI, known-RTL use RLI...PDI, unknown-direction UGC uses FSI...PDI. Recommend splitting the list into isolate controls (U+2066-U+2069) vs legacy embedding controls (LRE/RLE/LRO/RLO/PDF) with examples for each case.

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,140 @@
...
+active `DESIGN.md` decides brand visual language; this file decides
+how that language behaves when the script reads from the right or
+mixes direction within a line.
+
+> Grounded in primary sources: Unicode UAX #9 revision 51 (Sept 2025)
+> + Unicode 17.0, CSS Logical Properties Level 1, Tailwind v4.0/v4.2
+> changelogs, W3C alreq, Material 3 RTL guidance, Apple HIG
+> internationalization, WebKit Bugzilla 50949.
+
+## Logical properties first
```
lefarcen, line 33: P2: Missing base document direction guidance. The file covers logical CSS and inline isolation but doesn't state when/how to set `<html dir="rtl" lang="ar">`, subtree `dir`, or `dir="auto"` on paragraphs. Without this, agents can follow all rules and still produce an LTR document containing Arabic/Hebrew. Add a "Base direction and language" section with examples for full-page RTL, mixed-language subtrees, and LTR code blocks inside RTL paragraphs.

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,140 @@
...
+## What mirrors and what doesn't
+
+Mirroring isn't universal. The rules below are unanimous across
+Material 3 RTL guidance and Apple HIG internationalization.
+
+**Must mirror:**
+
+- Directional arrows (back / forward / next / previous), navigation rail position, tab order, slider fill direction, progress-bar fill, calendar-grid weekday order.
+- Checkbox-and-label position. Label sits to the right in LTR, to the left in RTL.
+- Phone-number and IBAN affordances when the surrounding paragraph is RTL but the value itself is LTR — wrap the value in `<bdi>` so the digits don't reflow.
```
lefarcen: P2: Phone numbers contain mostly weak/neutral chars, so auto-direction doesn't reliably express "this is LTR" in an RTL paragraph. The forms section correctly forces LTR for phone inputs (line 126), but rendered values need the same treatment. Update to: wrap in `<bdi dir="ltr">` or `<span dir="ltr">` for phone/IBAN/card-like values, not bare `<bdi>`.

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,140 @@
...
+  Applying tracking breaks the cursive joining the script depends on.
+- **Body type for Arabic runs ~14-18 px with line-height 1.5-1.75** to
+  give harakat (diacritics) clearance. Latin defaults are too tight.
+
+## Native mobile RTL parity
+
+Web RTL handling does not auto-translate to mobile. Each platform has
+its own direction primitive.
+
+| Platform | Direction primitive | Spacing |
```
lefarcen, line 140: P3: The native mobile table (UIKit/SwiftUI/Compose/Flutter/RN) is injected into blog/docs/finance skills, which emit desktop HTML only. That's ~16 lines of prompt budget on platforms they can't ship to. Consider splitting into `rtl-and-bidi-web.md` (for HTML skills) + `rtl-and-bidi-native.md` (for mobile skills), or keep native guidance in a separate craft file that only mobile skills opt into.

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,140 @@
...
+
+Form fields commonly mix scripts. Three rules cover most of it.
+
+- **`<input dir="auto">`** for any field whose value's direction is uncertain (search boxes, comment fields, free-text inputs). The browser detects from the first strong directional character.
+- **Force LTR on intrinsically-LTR fields** even inside an RTL paragraph: email, URL, phone, IBAN, credit-card. `<input type="email" dir="ltr">`.
+- **Wrap rendered values in `<bdi>`** when displaying mixed-script content (a username inside a paragraph, a model number inside a description). Stops the surrounding direction from rearranging the embedded value.
+
+## Common mistakes (lint these)
+
+- Hardcoded `left` / `right` / `text-align: left` in new CSS. Every hardcoded directional property is a bug in RTL.
```
lefarcen: P3: "Common mistakes (lint these)" mixes mechanically-lintable rules with ones that need script-awareness or exceptions. `left`/`right` can be legitimate for chart axes, physical icons, or platform-specific placement. "Italics on Arabic" and "letter-spacing on Arabic" require detecting script runs. Make this section explicit about what's mechanically lintable vs needs script detection, and add exception language to avoid false positives in future linting.

lefarcen (commented):
Strong craft module with thorough primary-source grounding 👍

The research rigor here (24 depth files, UAX #9 rev 51, CSS Logical Properties L1, Tailwind changelogs, Material 3 + Apple HIG) is exactly what behavioral craft needs. The mirroring tables and common-mistakes section are immediately actionable.

Found six substantive issues — mostly reasoning gaps rather than bugs. The core content is solid; these are about completeness and precision for a reference document that will be injected into agent prompts.

See inline comments below.

lefarcen:
Heads up: #505 (Hebrew locale) and #516 (Persian font/RTL) are complementary implementation PRs. Once this craft module lands, those PRs should verify their implementations against the logical-properties, bidi-isolation, and mirroring guidance here.

MohamedAbdallah-14:
Thanks @lefarcen. All six findings addressed in `5edd4f9`. Notes per item:

**P2 #1 WebKit #50949.** You're right — verified directly: the bug is RESOLVED FIXED on bugs.webkit.org. The "still open as of 2026-05" claim was wrong. Reframed the `<bdi>`-over-CSS recommendation to stand on UAX #9 §2.7 ("prefer markup over explicit formatting characters") rather than a WebKit-specific defect. Dropped the citation from the source list.

**P2 #2 isolates vs embeddings.** Right call. U+202C PDF terminates embeddings / overrides, not isolates. The original lump was structurally wrong. Split into two families with the source: isolate controls (U+2066 LRI / U+2067 RLI / U+2068 FSI, all closed by U+2069 PDI) for modern code, embedding / override controls (U+202A LRE / U+202B RLE / U+202D LRO / U+202E RLO, closed by U+202C PDF) as legacy. Recommend isolates as the default; embeddings only when interoperating with text from systems that emit them.

**P2 #3 base direction.** Added a new "Base direction and language" section before "Logical properties first" — covers `<html dir="rtl" lang="ar">`, mixed-language subtrees with nested `dir`, and `dir="auto"` for unknown-direction UGC. Includes the `dir`-without-`lang` warning since that's the load-bearing miss (font stack, hyphenation, locale-aware speech all key off `lang`).

**P2 #4 phone / IBAN values.** Updated both the must-mirror bullet and the forms-rendering rule to require `<bdi dir="ltr">` rather than bare `<bdi>` for fixed-LTR weak-character runs. Added a common-mistakes entry. The forms section already correctly forces `dir="ltr"` on the input element; this closes the loop on the rendered output.

**P3 #1 native mobile budget.** Took a lighter touch than a full file split. Added a one-line opt-out hint at the top of the native-mobile section so HTML-only skills know they can skim it. The mobile table is 16 lines on a 176-line file — the prompt-budget cost is bounded, and a clean web-vs-native split needs more thinking on what the section names should be (`rtl-and-bidi-web` / `rtl-and-bidi-native` or otherwise) and whether mobile skills should pull both. Happy to do that split as a follow-up if it becomes a real budget concern.

**P3 #2 lintability.** Restructured "Common mistakes" into three groups — mechanically lintable (with explicit exception language: chart axes, physical-object icons, platform-pinned UI), needs script detection (italics, letter-spacing, kashida claims, Lorem Ipsum — all need to detect Arabic / Hebrew runs at render time), and HTML semantics. Should make linter rules cleaner to write.

Ready for re-review.

lefarcen (commented):
All six findings addressed 🎉

**Verified fixes:**
- WebKit #50949 citation removed; recommendation now stands on UAX #9 §2.7
- Bidi controls properly split: isolates (LRI/RLI/FSI+PDI) vs legacy embeddings (LRE/RLE/LRO/RLO+PDF)
- New "Base direction and language" section covers `<html dir="rtl" lang="ar">`, mixed subtrees, `dir="auto"`
- Phone/IBAN guidance updated to `<bdi dir="ltr">` (explicit, not auto-detect)
- Native mobile section scoped with opt-out hint for web-only skills
- Common mistakes restructured: mechanically lintable / script-detection / HTML semantics

The craft module is now solid. Clean, precise, and ready for agent prompt injection. Looking forward to seeing this guide #505 and #516's implementations.

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,174 @@
...
+Three patterns cover the common cases:
+
+- **Full-page RTL.** `<html dir="rtl" lang="ar">`. Everything inside inherits.
+- **Mixed-language subtree.** Nest `<section dir="ltr" lang="en">…</section>` (or vice versa) when an embedded block uses a different script. Code samples, English citations, foreign brand names.
+- **User-generated content of unknown direction.** `dir="auto"` on the paragraph. The browser resolves direction from the first strong directional character in the run.
+
+Setting `lang` without `dir` is fine when the language has only one
+direction in practice (English doesn't need `dir="ltr"`). Setting
+`dir` without `lang` is rarely correct — at minimum drop the
+appropriate ISO-639 tag in.
```
mrcfps: Non-blocking: please qualify this `lang`-without-`dir` exception to document-root or already-default-LTR contexts. `lang` does not reset an inherited bidi base direction, so inside an RTL ancestor `<section lang="en">` would still inherit RTL despite line 22 correctly requiring `dir="ltr" lang="en"` for mixed-language subtrees. To avoid agents applying this note in the wrong context, consider wording like: “At the document root in a default-LTR page, English does not need `dir="ltr"`; inside any opposite-direction ancestor, set both `lang` and `dir`."

mrcfps (commented):
@MohamedAbdallah-14 Thanks for tightening the earlier RTL craft findings — the new section structure and skill opt-ins look useful. I found two small but actionable precision issues in the updated bidi guidance; they should be straightforward to clarify. 🙏


craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,174 @@
...
+
+- **Isolate controls** (modern, prefer these): U+2066 LRI, U+2067 RLI, U+2068 FSI — opened with these, all closed with U+2069 PDI. An isolated run does not affect, and is not affected by, the surrounding paragraph's bidi resolution. Use FSI when the embedded run's direction is unknown ahead of time.
+- **Embedding / override controls** (legacy): U+202A LRE, U+202B RLE, U+202D LRO, U+202E RLO — all closed with U+202C PDF. These nest within the surrounding paragraph rather than isolating from it; LRO/RLO additionally force a direction onto neutral characters. Newer code should use isolates; touch embeddings only when interoperating with text from systems that emit them.
+
+**Use `<bdi>` in HTML and the U+2068 / U+2069 isolate pair for plain
+text.** UAX #9 §2.7: *"where available, markup should be used instead
+of the explicit formatting characters."* `<bdi>` has been Baseline
+Widely Available since January 2020. Choose the markup form whenever
+you have markup; reach for the control characters only in plain-text
+contexts (logs, plain-text emails, terminal output).
```
mrcfps: Non-blocking: this plain-text recommendation currently narrows the control-character path to FSI/PDI (`U+2068` / `U+2069`), but FSI is first-strong auto-detection. That is the wrong default for known-direction runs, and especially for weak/neutral-heavy values like phone, IBAN, or card numbers — the same class of values this file explicitly forces to LTR in HTML on lines 88 and 147. Please recommend LRI/PDI for known-LTR runs, RLI/PDI for known-RTL runs, and reserve FSI/PDI for unknown-direction plain text.

MohamedAbdallah-14:
Thanks @mrcfps. Both addressed in `6776eb5`.

**`lang`-without-`dir` scope.** Right call — the previous wording read as universal but only holds at the document root in a default-LTR page. `lang` doesn't reset an inherited base direction, so an `<section lang="en">` inside an RTL ancestor still resolves RTL. Reworded to make the scope explicit: "fine at the document root in a default-LTR page; inside any opposite-direction ancestor, set both."

**Isolate picks for plain text.** Real catch. The previous "U+2068 / U+2069 isolate pair" framing defaulted to FSI, which is first-strong auto-detect — wrong default for known-direction runs and exactly the failure mode the file otherwise warns against (the phone / IBAN / card-number weak-character problem). Split into three explicit cases: LRI/PDI for known-LTR, RLI/PDI for known-RTL, FSI/PDI reserved for genuinely unknown direction (UGC where author + language vary). Added a "don't reach for FSI as the default" callout right under the list so the rule lands.

Ready for re-review.

lefarcen (commented):
Both mrcfps findings addressed ✅

Verified the updated guidance:
- **`lang`-without-`dir` scope**: now explicitly scoped to document-root/default-LTR contexts (lines 25-31) — opposite-direction subtrees correctly require both attributes.
- **Plain-text isolate picks**: now split into three explicit cases (lines 71-84) with the right default hierarchy — LRI/PDI for known-LTR, RLI/PDI for known-RTL, FSI/PDI reserved for genuinely unknown direction.

All prior findings addressed, no new issues. Looks good to me; deferring final approval to a maintainer.

lefarcen (commented):
Both mrcfps findings addressed ✅

Verified at `6776eb5`:
- **`lang`-without-`dir` scope**: now explicitly scoped to document-root-in-default-LTR contexts (lines 25-31). The note correctly warns that opposite-direction subtrees must set both `lang` and `dir`.
- **Plain-text isolate picks**: split into three explicit cases (lines 71-84) — LRI/PDI for known-LTR, RLI/PDI for known-RTL, FSI/PDI reserved for genuinely unknown direction. The "don't default to FSI" callout is clear.

All six original findings from the first review and both mrcfps precision issues are now resolved. This craft module is ready from a content accuracy standpoint; deferring final approval to a maintainer.

skills/docs-page/SKILL.md
```diff
@@ -21,6 +21,8 @@ od:
   design_system:
     requires: true
     sections: [color, typography, layout, components]
+  craft:
+    requires: [rtl-and-bidi]
```
mrcfps, line 25: This opt-in makes `rtl-and-bidi` available to the docs skill, but the same skill still gives later, more specific physical-direction layout instructions: the description says “left nav” and “right-rail table of contents” (lines 3-5), and the workflow requires “Left nav” with a “left-edge accent stripe” plus “Right TOC” (lines 39-46). Because the craft guidance is injected above the skill body, an Arabic/Hebrew docs request can still follow those later LTR-only instructions and produce a page that violates the new craft rule that hardcoded `left`/`right` is a RTL bug. Please update the docs-page skill body alongside this opt-in to use logical placement instead, e.g. nav on `inline-start`, accent stripe on `inline-start`, TOC on `inline-end`, with rails swapping under `[dir=rtl]` or equivalent logical CSS.

skills/blog-post/SKILL.md
```diff
@@ -25,6 +25,8 @@ od:
   design_system:
     requires: true
     sections: [color, typography, layout, components]
+  craft:
+    requires: [rtl-and-bidi]
```
mrcfps, line 29: This new RTL craft opt-in also conflicts with a later blog-specific requirement: the body still asks for a pull quote with an “accent rule on the left” (line 49). For RTL articles, that keeps the editorial accent on the physical LTR edge even though the craft file asks generated layouts to prefer logical inline directions. Please change the blog skill wording to “accent rule on the inline-start edge” (or explicitly describe LTR/RTL variants) so the skill-specific prompt doesn't override the craft guidance it now opts into.

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,185 @@
...
+**Needs script detection (will false-positive without it):**
+
+- "Use `text-justify: kashida` for Arabic" — no browser implements it. CSS `text-align: justify` adds inter-word spacing and looks unnatural in Arabic; kashida elongation is the correct form, but it isn't shippable on the web today.
+- Italics on Arabic or Hebrew text. Neither script has an italic tradition.
+- CSS `letter-spacing` applied to Arabic. Breaks cursive joining (alreq treats it as a boundary concept, not a uniform tracking value).
+- Lorem Ipsum used for RTL prototyping. Arabic word lengths, connection behaviors, and vertical extents differ; use real Arabic / Hebrew text.
+
+**HTML semantics:**
+
+- Reaching for `unicode-bidi: plaintext` when `<bdi dir="auto">` (or `<bdi>` with implicit auto-detect) is the markup form of the same idea. UAX #9 §2.7: prefer markup over CSS or control characters when markup is available.
```
mrcfps: The HTML-semantics lint guidance says `<bdi dir="auto">` (or bare `<bdi>`) is “the markup form of the same idea” as `unicode-bidi: plaintext`, but those are not equivalent. `<bdi>`/`dir=auto` give isolated direction behavior for the element, while `unicode-bidi: plaintext` changes how base direction is determined for each plaintext paragraph; replacing one with the other can change behavior in multi-paragraph or block-level text. Since this section is intended to drive future lint rules, the current wording would flag valid CSS and suggest a non-equivalent fix. Please reword it to the narrower rule this file needs, for example: prefer semantic isolation with `<bdi>`/`dir` for inline HTML runs, and only use CSS bidi controls when the desired block/plaintext paragraph behavior is explicitly required and tested.

mrcfps (changes requested):
@MohamedAbdallah-14 Thanks for the thoughtful RTL/bidi craft module and for already tightening the earlier precision feedback. I found a blocking integration issue where newly opted-in skills still contain later physical-direction instructions that can override the craft guidance, plus one smaller wording fix in the craft lint guidance. 🙏


MohamedAbdallah-14:
Thanks @mrcfps. All three blocking items addressed in `17505b9`.

**Skill-body physical-direction conflicts (P1).** Right call on the mechanism: craft injects above the skill body, so a "Left nav" instruction in the skill survives the opt-in. Updated both files to remove the conflict.

- `skills/docs-page`: description now reads "inline-start nav, inline-end TOC". Lay-out section explicitly notes "expressed on the inline axis so the layout flips correctly under `dir="rtl"`" and replaces "Left nav" / "Right TOC" / "left-edge accent stripe" with the inline-axis equivalents. Added a self-check bullet requiring logical CSS (`margin-inline-start`, `border-inline-start`, `inset-inline-end`, `text-align: start`) on rails and accent.
- `skills/blog-post`: pull-quote rule changed from "accent rule on the left" to "accent rule on the inline-start edge so the layout flips correctly under `dir="rtl"`."

**`<bdi dir="auto">` vs `unicode-bidi: plaintext` (P1).** You're right — these aren't drop-in equivalents. `<bdi>` isolates an inline run from surrounding bidi resolution; `unicode-bidi: plaintext` changes how base direction is *determined* for each plaintext paragraph in a block. Different surfaces. Reworded the lint guidance to: "Prefer semantic isolation in HTML for inline runs; `unicode-bidi: plaintext` operates on a different surface and should only be used when that block-level paragraph behavior is explicitly required and tested. The two are not drop-in equivalents — don't lint one as a replacement for the other." That avoids the future-linter false-positive you flagged.

The skill-body fix is the more interesting one structurally — it's a pattern that will repeat for accessibility-baseline opt-ins (which I should audit retroactively in a separate PR) and for any future craft module that imposes layout constraints. Good catch.

Ready for re-review.

lefarcen (commented):
Hey @MohamedAbdallah-14, verified the latest fixes at `17505b9` — all three blocking items from mrcfps's review are addressed.

**Physical-direction conflicts:**

- ✅ `skills/docs-page`: layout now expressed on the inline axis (inline-start nav, inline-end TOC, inline-start-edge accent stripe) — flips correctly under `dir="rtl"`
- ✅ `skills/blog-post`: pull-quote accent rule now on inline-start edge with explicit RTL flip note

**HTML semantics lint:**

- ✅ `craft/rtl-and-bidi`: `<bdi>` vs `unicode-bidi: plaintext` wording now correctly explains these operate on different surfaces and aren't drop-in equivalents

All findings resolved. 🎉

lefarcen (commented):
Hey @MohamedAbdallah-14! 👏 Verified all three mrcfps findings are addressed in `17505b97c1ee`:

**Skill physical-direction conflicts:** Both `skills/docs-page` and `skills/blog-post` now use inline-axis directions (`inline-start` nav/accent, `inline-end` TOC) so the layout swaps correctly under `dir="rtl"`. The description and workflow text match — no more hardcoded `left`/`right` that would override the craft guidance.

**`<bdi>` vs `unicode-bidi: plaintext` semantics:** The lint guidance now correctly distinguishes the two — `<bdi>` for inline isolation, `unicode-bidi: plaintext` for block-level paragraph base-direction determination, with the explicit note "not drop-in equivalents." That avoids the future linter false-positive mrcfps flagged.

The skill-body integration fix is especially clean — this same inline-axis pattern will apply to any future craft module that constrains layout (accessibility-baseline audit is a good candidate for a follow-up sweep).

All findings resolved. Ready for mrcfps re-review. 🎉

craft/rtl-and-bidi.md
```diff
@@ -0,0 +1,185 @@
...
+user-generated content where direction isn't known at author time.
+
+## What mirrors and what doesn't
+
+Mirroring isn't universal. The rules below are unanimous across
+Material 3 RTL guidance and Apple HIG internationalization.
+
+**Must mirror:**
+
+- Directional arrows (back / forward / next / previous), navigation rail position, tab order, slider fill direction, progress-bar fill, calendar-grid weekday order.
```
mrcfps: Non-blocking: please qualify the progress-bar mirroring rule so it doesn't accidentally cover media scrubbers. This line currently says progress-bar fill direction must mirror, while the nearby media exception says playback controls do not mirror because they represent tape/media direction rather than reading direction. A generated audio/video player could follow the progress-bar bullet and flip the scrubber/progress fill in RTL even though that control belongs with media playback behavior. Consider changing this to "non-media progress-bar fill" here, or add media progress/scrubbers to the must-not-mirror media bullet.

mrcfps (commented):
@MohamedAbdallah-14 Thanks for the careful follow-up fixes on the RTL/bidi craft module — the skill integration updates are much clearer now. I found one remaining non-blocking precision issue in the mirroring guidance that would be worth tightening before this becomes reusable prompt guidance. 🙏


MohamedAbdallah-14:
Thanks @mrcfps. The progress-bar / media-scrubber split is a real catch — the previous bullet would have flipped a video timeline under `dir="rtl"`, directly contradicting the must-not-mirror rule for media playback controls right below it.

Fixed in `c51a337`. Two changes:

- **Must-mirror progress bars** now scoped to non-media: download progress, upload status, form-completion bars. These are reading-direction phenomena.
- **Media scrubber / progress timeline** added explicitly to the must-not-mirror media bullet alongside play / pause / fast-forward / rewind. They represent tape direction, not reading direction — same logic as the playback controls.

The two stale inline comments (docs-page "left nav", blog-post "accent rule on the left") are repeats from the previous review round and are already addressed in 17505b9 — those skill-body changes shipped before this commit. Probably looper diff-line drift.

Ready for re-review.

mrcfps (approved):
@MohamedAbdallah-14 I reviewed the current head after the RTL/bidi precision fixes, including the new craft module, README registration, and the blog/docs/finance skill opt-ins. The changed guidance now consistently uses logical directions, explicit bidi isolation for weak-character values, and the media-scrubber exception, with the previous review concerns addressed cleanly — thanks for the careful follow-through on this contribution. 🙏


lefarcen:
Hey @MohamedAbdallah-14! 🎉 PR #595 is merged — the RTL/bidi craft module is now live.

This was a thorough contribution: 186 lines covering logical properties, bidi isolation, mirroring decisions, per-platform primitives, and RTL forms, backed by Unicode UAX #9 rev 51 + CSS Logical Properties L1 + Tailwind v4 changelogs + W3C alreq + Material 3/Apple HIG. The 5-round review cycle (WebKit citation, isolate vs embedding, base direction scope, skill physical-direction conflicts, `<bdi>` vs plaintext equivalence, progress-bar / media-scrubber precision) resulted in clean guidance that will prevent real layout bugs for Arabic/Hebrew/Persian users.

Thanks for the careful iteration and for sticking with the review feedback — the RTL guidance is significantly stronger for it. Looking forward to module 5 (form-validation) 🚀

— open-design team

PR #612, opened 2026-05-06
fix(acp): normalize mcpServers to stdio shape for Kimi/Hermes ACP

Embracex1998:
## Problem

When using Kimi CLI (ACP JSON-RPC agent), Open Design's daemon sends a session/new request with MCP server descriptors in the shape { name, command, args }. Kimi CLI 1.35.0 expects the full ACP stdio server shape { type: 'stdio', name, command, args, env: [] }, and rejects the request with **JSON-RPC -32602 Invalid params**.

This happens when mcpDiscovery is enabled for an ACP agent (Kimi and Hermes both declare mcpDiscovery: 'mature-acp').

## Reproduction

1. Install Kimi CLI 1.35.0 and ensure it is on PATH.
2. Start Open Design (pnpm tools-dev run web).
3. Open the web UI and start a new chat with the Kimi agent.
4. The daemon calls ttachAcpSession → session/new with mcpServers populated by uildLiveArtifactsMcpServersForAgent.
5. Kimi CLI returns json-rpc id 2: Invalid params and the chat fails immediately.

## Fix

uildAcpSessionNewParams now maps every MCP server descriptor to the complete stdio shape expected by ACP-compliant agents:

`	s
{
  type: 'stdio',
  name: typeof s?.name === 'string' ? s.name : '',
  command: typeof s?.command === 'string' ? s.command : '',
  args: Array.isArray(s?.args) ? s.args : [],
  env: [],
}
`

This keeps the change scoped to the ACP adapter without affecting other agents (Claude Code, Codex, etc.) that use different transport formats.

## Verification

- Manually tested the full ACP handshake against kimi acp:
  - initialize → success
  - session/new with normalized mcpServers → returns valid sessionId
  - session/prompt → streams correctly

Fixes the json-rpc id 2: Invalid params error reported when running Open Design with Kimi CLI.

### DIFF
--- a/apps/daemon/src/acp.ts
+++ b/apps/daemon/src/acp.ts
@@ -7,13 +7,21 @@ const DEFAULT_TIMEOUT_MS = 15_000;
 const DEFAULT_STAGE_TIMEOUT_MS = 180_000;
 
 export function buildAcpSessionNewParams(cwd, { mcpServers } = {}) {
+  const servers = Array.isArray(mcpServers) ? mcpServers : [];
   return {
     cwd: path.resolve(cwd),
     // MCP is an optional compatibility layer. Default to no MCP servers so ACP
     // agents can run through the skill + CLI path without MCP support. Do not
     // auto-install or mutate user/global MCP config; callers must pass an
     // explicit per-session MCP descriptor when a compatible agent supports it.
-    mcpServers: Array.isArray(mcpServers) ? mcpServers : [],
+    // Normalize to the ACP stdio server shape expected by Kimi/Hermes.
+    mcpServers: servers.map((s) => ({
+      type: typeof s?.type === 'string' ? s.type : 'stdio',
+      name: typeof s?.name === 'string' ? s.name : '',
+      command: typeof s?.command === 'string' ? s.command : '',
+      args: Array.isArray(s?.args) ? s.args : [],
+      env: Array.isArray(s?.env) ? s.env : [],
+    })),
   };
 }
 
--- a/apps/daemon/tests/acp.test.ts
+++ b/apps/daemon/tests/acp.test.ts
@@ -19,11 +19,30 @@ test('ACP session params do not request global MCP config mutation', () => {
   assert.equal('installMcpServers' in params, false);
 });
 
-test('ACP session params accept explicit MCP servers as optional configuration', () => {
+test('ACP session params normalize explicit MCP servers to ACP stdio shape', () => {
   const mcpServers = [{ name: 'open-design-live-artifacts', command: 'od', args: ['mcp', 'live-artifacts'] }];
 
   assert.deepEqual(buildAcpSessionNewParams('/tmp/od-project', { mcpServers }), {
     cwd: path.resolve('/tmp/od-project'),
-    mcpServers,
+    mcpServers: [
+      {
+        type: 'stdio',
+        name: 'open-design-live-artifacts',
+        command: 'od',
+        args: ['mcp', 'live-artifacts'],
+        env: [],
+      },
+    ],
   });
+});
+
+test('ACP session params preserve caller-provided type and env fields', () => {
+  const mcpServers = [
+    { type: 'http', name: 'http-server', url: 'http://localhost:3000', headers: {}, env: [{ key: 'TOKEN', value: 'secret' }] },
+  ];
+
+  const result = buildAcpSessionNewParams('/tmp/od-project', { mcpServers });
+  assert.equal(result.mcpServers[0].type, 'http');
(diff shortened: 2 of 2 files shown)


lefarcen:
Hi @Embracex1998! 🎉
Thanks for the contribution — fixing the Kimi/Hermes ACP integration issue is a valuable improvement.

— open-design team

apps/daemon/src/acp.ts
```diff
@@ -7,13 +7,21 @@ const DEFAULT_TIMEOUT_MS = 15_000;
...
+  const servers = Array.isArray(mcpServers) ? mcpServers : [];
   return {
     cwd: path.resolve(cwd),
     // MCP is an optional compatibility layer. Default to no MCP servers so ACP
     // agents can run through the skill + CLI path without MCP support. Do not
     // auto-install or mutate user/global MCP config; callers must pass an
     // explicit per-session MCP descriptor when a compatible agent supports it.
-    mcpServers: Array.isArray(mcpServers) ? mcpServers : [],
+    // Normalize to the ACP stdio server shape expected by Kimi/Hermes.
+    mcpServers: servers.map((s) => ({
```
mrcfps, line 18: This changed range now makes `buildAcpSessionNewParams` return normalized ACP stdio descriptors instead of the original caller-provided objects, but the existing test at `apps/daemon/tests/acp.test.ts:22-28` still asserts that explicit `mcpServers` are passed through unchanged. With this mapping, that assertion will now compare `{ name, command, args }` against `{ type: 'stdio', name, command, args, env: [] }`, so the daemon ACP test suite will fail once dependencies are installed and CI runs it. Please update that test (or replace it with a fixture that covers the Kimi/Hermes stdio shape) to assert the new `type` and `env` fields, so the regression coverage matches the intended protocol fix.

mrcfps (changes requested):
@Embracex1998 thanks for the focused fix for the ACP MCP server descriptor shape. I found one blocking follow-up: the implementation change updates the function output, but the existing daemon ACP test still asserts the old passthrough behavior, so the test suite needs to be updated alongside the code before this is ready.


apps/daemon/src/acp.ts
lefarcen, line 18: **P2** The normalization always sets `env: []`, which is safe for the current live-artifacts MCP server but silently discards any explicit per-session MCP server env. That can break servers requiring credentials or config.

**Why it matters**: Future MCP servers may need `env` for API keys, tokens, or runtime config. Discarding them now means those servers will fail silently when added.

**Concrete fix**: Preserve validated env entries when provided (e.g., only `{ name: string, value: string }[]`), while defaulting to `[]` for Kimi/Hermes:
```typescript
env: Array.isArray(s?.env) 
  ? s.env.filter(e => typeof e?.name === 'string' && typeof e?.value === 'string')
  : [],
```

lefarcen (commented):
Hey @Embracex1998, thanks for tackling the Kimi/Hermes ACP integration issue! 👋 The fix direction is sound — normalizing to the stdio shape resolves the immediate JSON-RPC -32602 error. I have two P2 concerns about test coverage and env handling that would strengthen this before merge.

**Code review (Lens A)**:
The normalization logic is clean and scoped correctly. Two areas need attention: the existing test suite will break (it still expects passthrough shape), and the hardcoded `env: []` silently discards any explicit per-session MCP server env that might be needed for credential-bearing servers.

lefarcen (commented):
Hey @Embracex1998! 🎉 Excellent iteration — you addressed both blocking concerns (test coverage + env preservation) and added thoughtful improvements (type preservation + dedicated test case).

**All prior concerns resolved:**
- ✅ Test suite updated to match new normalized shape
- ✅ `env` now preserved from caller instead of discarded
- ✅ `type` preserved when provided (defaults to 'stdio' only when missing)
- ✅ New test case covers the preservation logic

Looks solid to me. Deferring final approval to a maintainer.

mrcfps (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @Embracex1998 for the clean fix — the ACP stdio normalization + test coverage landed smoothly. Kimi/Hermes users should now get clean session handshakes.

If you run into any issues with this in prod, feel free to open a new issue. Thanks for making open-design better!

leilei926524-tech:
Hi @Embracex1998 — welcome to Open Design contributors. 🎉

Your first merged PR (#612) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #614, opened 2026-05-06
fix(daemon, packaged): unbreak GUI-launched agent detection on minimal PATHs (#442)

### LINKED ISSUE #442: Shows "Gemini CLI" not installed. But it's installed.

fancyboi999:
## Summary

Fixes #442. GUI-launched daemons (Finder/Dock on macOS, `.desktop` on Linux) inherit a stripped PATH from launchd / the desktop session and don't read interactive shell rc files, so any CLI installed via `npm i -g <cli>` under a sudo-free prefix like `~/.npm-global` was silently undetected. Surface fix is "add `~/.npm-global/bin` to the search list", but the search list lives in *two* places that have already drifted, so the fix needs to share a single source of truth.

## Root cause (verified end-to-end against current `main`)

Two independent copies of the user-toolchain bin list:

- `apps/daemon/src/agents.ts` `userToolchainDirs()` — drives `resolveOnPath` → `detectAgents`
- `apps/packaged/src/sidecars.ts` `resolvePackagedPathEnv()` — rebuilds the daemon child's `PATH` in packaged mode

Both lists are missing `~/.npm-global/bin`. They had also already drifted: `~/.asdf/shims` and `~/Library/pnpm` exist in the daemon copy but not in the packaged copy, so asdf and `pnpm self-install`-ed CLIs are silently undetected in packaged builds today even though `pnpm tools-dev` sees them.

Reproduction script (against `main`, before this PR):

```js
process.env.PATH = '/usr/bin:/bin:/usr/sbin:/sbin'; // launchd default
process.env.OD_AGENT_HOME = FAKE_HOME; // ~/.npm-global/bin/gemini exists
const { resolveAgentExecutable } = await import('apps/daemon/dist/agents.js');
console.log(resolveAgentExecutable({ bin: 'gemini' })); // → null
```

`launchctl getenv PATH` returns empty on a stock macOS, confirming GUI launches do not inherit shell PATH.

## Solution

Extract `wellKnownUserToolchainBins(options)` into `@open-design/platform` as the single source of truth. Both layers consume it. The list now also covers:

- `~/.npm-global/bin` and `~/.npm-packages/bin` — the two dominant non-canonical npm prefixes shipped in third-party "fix npm EACCES without sudo" tutorials
- `$NPM_CONFIG_PREFIX/bin` and `$npm_config_prefix/bin` — npm's own config env var, so corporate/CI provisioning that exports a custom prefix resolves automatically

Adopting a shared helper also drags the `~/.asdf/shims` and `~/Library/pnpm` entries into the packaged path, fixing the pre-existing silent drift for free.

## Files

- `packages/platform/src/index.ts` — adds `wellKnownUserToolchainBins({ home, includeSystemBins, env })`
- `packages/platform/tests/index.test.ts` — 9 new vitest cases covering the documented dirs, both npm-prefix conventions, env-var override (uppercase + lowercase), `includeSystemBins` flag, and per-version mise/nvm/fnm expansion
- `apps/daemon/src/agents.ts` — `userToolchainDirs()` is now a thin caching wrapper that calls the platform helper; `OD_AGENT_HOME` test hook preserved
- `apps/daemon/tests/agents.test.ts` — adds two regression cases under the existing minimal-PATH harness: `~/.npm-global/bin/gemini` and `~/.npm-packages/bin/gemini` resolve when `PATH=/usr/bin:/bin`
- `apps/packaged/src/sidecars.ts` — `resolvePackagedPathEnv()` now consumes the shared helper and only keeps the POSIX system bins (`/usr/bin`, `/bin`, `/usr/sbin`, `/sbin`) at this layer
- `packages/AGENTS.md` — records that the platform helper is the shared source of truth so the two layers can never drift again

No new dependencies. Both apps already depend on `@open-design/platform`.

## Validation

```bash
$ pnpm --filter @open-design/platform test
 Test Files  1 passed (1)
      Tests  23 passed (23)

$ pnpm --filter @open-design/daemon exec vitest run tests/agents.test.ts
 Test Files  1 passed (1)
      Tests  64 passed (64)

$ pnpm typecheck
- 0 errors

$ pnpm guard
Residual JavaScript check passed: project-owned code is TypeScript-only.
Test layout check passed: apps/packages/tools tests live in sibling tests directories.

$ pnpm --filter @open-design/platform build
$ pnpm --filter @open-design/daemon build
$ pnpm --filter @open-design/packaged build
# all succeeded
```

End-to-end reproduction script run against the patched `dist`:

```json
{
  "fakeHome": "/var/folders/.../verify-442-JDdByP",
  "fakeGeminiExists": true,
  "pathUsed": "/usr/bin:/bin:/usr/sbin:/sbin",
  "resolved": "/var/folders/.../.npm-global/bin/gemini",
  "bug_present": false,
  "fix_works": true
}
```

Some unrelated tests (project-status, project-watchers, web build) fail on this machine due to local `better-sqlite3` / `turbopack` / `rollup` native-binary signing issues that exist on `main` before any of these changes. CI will run on a clean Linux runner.

## Test plan

- [x] `pnpm --filter @open-design/platform test` — green
- [x] `pnpm --filter @open-design/daemon exec vitest run tests/agents.test.ts` — green
- [x] `pnpm typecheck` — clean
- [x] `pnpm guard` — clean
- [x] `pnpm --filter @open-design/{platform,daemon,packaged} build` — all succeed
- [x] End-to-end repro: `resolveAgentExecutable({ bin: 'gemini' })` resolves under launchd-default PATH when binary lives in `~/.npm-global/bin`
- [ ] CI on Linux runner
- [ ] Optional manual sanity check on a packaged macOS build (out of scope here, but the same shared helper now drives that path)

Closes #442.

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -1,10 +1,11 @@
 // @ts-nocheck
 import { execFile } from 'node:child_process';
 import { promisify } from 'node:util';
-import { existsSync, readdirSync } from 'node:fs';
+import { existsSync } from 'node:fs';
 import { delimiter } from 'node:path';
 import path from 'node:path';
 import { homedir } from 'node:os';
+import { wellKnownUserToolchainBins } from '@open-design/platform';
 import { detectAcpModels } from './acp.js';
 import { parsePiModels } from './pi-rpc.js';
 
@@ -712,22 +713,11 @@ export const AGENT_DEFS = [
   },
 ];
 
-function existingDirsUnder(root, segments = []) {
-  const dirs = [];
-  let entries = [];
-  try {
-    entries = readdirSync(root, { withFileTypes: true });
-  } catch {
-    return dirs;
-  }
-  for (const entry of entries) {
-    if (!entry.isDirectory()) continue;
-    const full = path.join(root, entry.name, ...segments);
-    if (existsSync(full)) dirs.push(full);
-  }
-  return dirs;
-}
-
+// Toolchain dir computation lives in @open-design/platform so the daemon
+// resolver and the packaged sidecar PATH builder can never drift again
+// (issue #442). The wrapper here just preserves the OD_AGENT_HOME test
+// hook and the per-home cache that reduces filesystem scans on every
+// resolveOnPath() call.
 const TOOLCHAIN_DIR_CACHE_TTL_MS = 5000;
 let cachedToolchainHome = null;
 let cachedToolchainDirs = null;
@@ -746,27 +736,13 @@ function userToolchainDirs() {
   }
   cachedToolchainHome = home;
   cachedToolchainDirsAt = now;
-  cachedToolchainDirs = [
-    path.join(home, '.local', 'bin'),
-    path.join(home, '.opencode', 'bin'),
-    path.join(home, '.bun', 'bin'),
-    path.join(home, '.volta', 'bin'),
-    path.join(home, '.asdf', 'shims'),
-    path.join(home, 'Library', 'pnpm'),
-    path.join(home, '.cargo', 'bin'),
-    ...(process.platform !== 'win32' && !homeOverride
-      ? ['/opt/homebrew/bin', '/usr/local/bin']
-      : []),
-    ...existingDirsUnder(
-      path.join(home, '.local', 'share', 'mise', 'installs', 'node'),
-      ['bin'],
-    ),
-    ...existingDirsUnder(path.join(home, '.nvm', 'versions', 'node'), ['bin']),
-    ...existingDirsUnder(
-      path.join(home, '.local', 'share', 'fnm', 'node-versions'),
-      ['installation', 'bin'],
-    ),
-  ];
+  // Skip Homebrew / /usr/local bins when running under OD_AGENT_HOME so
+  // tests that scope detection to a temp home don't pick up the real
+  // machine's CLIs.
+  cachedToolchainDirs = wellKnownUserToolchainBins({
+    home,
+    includeSystemBins: process.platform !== 'win32' && !homeOverride,
+  });
   return cachedToolchainDirs;
 }
 
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -715,9 +715,10 @@ export const AGENT_DEFS = [
 
 // Toolchain dir computation lives in @open-design/platform so the daemon
 // resolver and the packaged sidecar PATH builder can never drift again
-// (issue #442). The wrapper here just preserves the OD_AGENT_HOME test
-// hook and the per-home cache that reduces filesystem scans on every
-// resolveOnPath() call.
+// (issue #442). See @open-design/platform's wellKnownUserToolchainBins
+// for the canonical search list. The wrapper here just preserves the
+// OD_AGENT_HOME test hook and the per-home cache that reduces
+// filesystem scans on every resolveOnPath() call.
 const TOOLCHAIN_DIR_CACHE_TTL_MS = 5000;
 let cachedToolchainHome = null;
 let cachedToolchainDirs = null;
@@ -736,12 +737,16 @@ function userToolchainDirs() {
   }
   cachedToolchainHome = home;
   cachedToolchainDirsAt = now;
-  // Skip Homebrew / /usr/local bins when running under OD_AGENT_HOME so
-  // tests that scope detection to a temp home don't pick up the real
-  // machine's CLIs.
+  // When OD_AGENT_HOME is set, scope the search strictly to the override
+  // home: skip Homebrew / /usr/local *and* pass an empty env so that a
+  // developer or CI runner with NPM_CONFIG_PREFIX / npm_config_prefix
+  // exported can't leak the real machine's <prefix>/bin into a sandboxed
+  // detection run. Without this the agents.test.ts cases that build a
+  // tmp home would be machine-environment-dependent.
   cachedToolchainDirs = wellKnownUserToolchainBins({
     home,
     includeSystemBins: process.platform !== 'win32' && !homeOverride,
+    env: homeOverride ? {} : process.env,
   });
   return cachedToolchainDirs;
 }
--- a/apps/daemon/tests/agents.test.ts
+++ b/apps/daemon/tests/agents.test.ts
@@ -41,6 +41,7 @@ const originalHome = process.env.HOME;
 const originalAgentHome = process.env.OD_AGENT_HOME;
 const originalDaemonUrl = process.env.OD_DAEMON_URL;
 const originalToolToken = process.env.OD_TOOL_TOKEN;
+const originalNpmConfigPrefix = process.env.NPM_CONFIG_PREFIX;
 const originalFetch = globalThis.fetch;
 
 afterEach(() => {
@@ -69,6 +70,11 @@ afterEach(() => {
     delete process.env.OD_TOOL_TOKEN;
   } else {
     process.env.OD_TOOL_TOKEN = originalToolToken;
+  }
+  if (originalNpmConfigPrefix == null) {
+    delete process.env.NPM_CONFIG_PREFIX;
+  } else {
+    process.env.NPM_CONFIG_PREFIX = originalNpmConfigPrefix;
   }
   globalThis.fetch = originalFetch;
 });
@@ -845,7 +851,11 @@ fsTest(
           `got ${resolved}`,
       );
     } finally {
-      delete process.env.NPM_CONFIG_PREFIX;
+      // afterEach restores NPM_CONFIG_PREFIX to its pre-test value (or
+      // deletes it when it was unset), so do not unconditionally
+      // `delete` it here — that would clobber an export the developer
+      // / CI runner had already set, leaking into the next test in the
+      // same Vitest worker.
       rmSync(sandbox, { recursive: true, force: true });
       rmSync(realPrefix, { recursive: true, force: true });
     }
--- a/apps/daemon/tests/agents.test.ts
+++ b/apps/daemon/tests/agents.test.ts
@@ -812,6 +812,46 @@ fsTest(
   },
 );
 
+// Test isolation: when OD_AGENT_HOME points at a sandbox, an exported
+// $NPM_CONFIG_PREFIX / $npm_config_prefix on the developer's or CI
+// runner's environment must not leak a real <prefix>/bin into the
+// sandboxed search list. Otherwise an agent installed by the host
+// machine could satisfy a "not on PATH" assertion in the sandbox and
+// make detection tests environment-dependent. Raised in PR review on
+// #442 (review comment by @mrcfps on apps/daemon/src/agents.ts:742).
+fsTest(
+  'OD_AGENT_HOME isolates resolution from $NPM_CONFIG_PREFIX leakage',
+  () => {
+    const sandbox = mkdtempSync(join(tmpdir(), 'od-agents-sandbox-'));
+    const realPrefix = mkdtempSync(join(tmpdir(), 'od-agents-real-prefix-'));
+    const realPrefixBin = join(realPrefix, 'bin');
+    try {
+      // Sandbox is empty — gemini does not exist under OD_AGENT_HOME.
+      // Real prefix has a gemini, simulating the developer's /opt/...
+      // or ~/.npm-global install. NPM_CONFIG_PREFIX points at it.
+      mkdirSync(realPrefixBin, { recursive: true });
+      writeFileSync(join(realPrefixBin, 'gemini'), '');
+      chmodSync(join(realPrefixBin, 'gemini'), 0o755);
+
+      process.env.OD_AGENT_HOME = sandbox;
+      process.env.PATH = '/usr/bin:/bin';
+      process.env.NPM_CONFIG_PREFIX = realPrefix;
+
+      const resolved = resolveAgentExecutable({ bin: 'gemini' });
+      assert.equal(
+        resolved,
+        null,
+        `OD_AGENT_HOME sandbox must not see the real $NPM_CONFIG_PREFIX bin; ` +
+          `got ${resolved}`,
+      );
+    } finally {
+      delete process.env.NPM_CONFIG_PREFIX;
+      rmSync(sandbox, { recursive: true, force: true });
+      rmSync(realPrefix, { recursive: true, force: true });
+    }
+  },
+);
+
 // DeepSeek TUI's exec subcommand requires the prompt as a positional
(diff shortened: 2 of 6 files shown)


lefarcen (commented):
Hey @fancyboi999! 👋

Excellent work on this fix. The shared-helper approach eliminates the drift at its root, and the test coverage is comprehensive. A few minor suggestions below, but overall this is production-ready.

**Code review (Lens A):**

✅ **Correctness:** The extraction preserves all existing behavior. The daemon cache wrapper correctly passes `includeSystemBins: false` when `OD_AGENT_HOME` is set, preventing test pollution.

✅ **Security:** No new attack surface. The helper reads env vars and filesystem metadata but doesn't execute anything or expose secrets.

✅ **Edge cases:** npm prefix env var handling covers both uppercase and lowercase, per-version toolchain scanning is best-effort (missing roots contribute nothing), and the `includeSystemBins` flag prevents Windows from seeing POSIX-only paths.

✅ **Test coverage:** 9 new vitest cases in `packages/platform/tests/index.test.ts` + 2 daemon regression cases cover the documented behavior comprehensively.

1. **Type safety for NPM prefix check:** Line 467 in `packages/platform/src/index.ts` checks `typeof npmPrefix === "string" && npmPrefix.length > 0`. This is safe, but `env.NPM_CONFIG_PREFIX` could theoretically be a number or boolean in a weird env (e.g., `NPM_CONFIG_PREFIX=1`). Consider `typeof npmPrefix === "string" && npmPrefix.trim().length > 0` to also handle accidental whitespace-only values.

2. **Windows PATH handling:** The helper assumes POSIX path separators (`join(home, "Library", "pnpm")`). On Windows, `Library/pnpm` won't exist, but `join()` handles the separator correctly. No bug, just noting that the `includeSystemBins` default already gates `/opt/homebrew/bin` off Windows.

3. **Documentation visibility:** `packages/AGENTS.md` mentions the helper is the source of truth, but the actual `wellKnownUserToolchainBins` function JSDoc is excellent — consider also adding a one-line pointer in the daemon's `agents.ts` comment block (line 716) like `"See @open-design/platform wellKnownUserToolchainBins for the full list."` so readers don't have to grep.

Ready to merge from my side once CI passes. Nice contribution! 🎉

apps/daemon/src/agents.ts
```diff
@@ -746,27 +736,13 @@ function userToolchainDirs() {
...
-    ...existingDirsUnder(path.join(home, '.nvm', 'versions', 'node'), ['bin']),
-    ...existingDirsUnder(
-      path.join(home, '.local', 'share', 'fnm', 'node-versions'),
-      ['installation', 'bin'],
-    ),
-  ];
+  // Skip Homebrew / /usr/local bins when running under OD_AGENT_HOME so
+  // tests that scope detection to a temp home don't pick up the real
+  // machine's CLIs.
+  cachedToolchainDirs = wellKnownUserToolchainBins({
```
mrcfps, line 746: **Non-blocking:** please keep `OD_AGENT_HOME` lookup isolated from the caller's npm-prefix environment.

The wrapper still skips Homebrew and `/usr/local` when `OD_AGENT_HOME` is set, but this changed call now lets `wellKnownUserToolchainBins()` read its default `process.env`. That means a developer or CI runner with `NPM_CONFIG_PREFIX` / `npm_config_prefix` set will inject that real `<prefix>/bin` into `resolveOnPath()` even while the test hook points `home` at a temp fixture. In practice, an agent binary from the real npm prefix could satisfy a sandboxed “not on PATH” or temp-home lookup and make the agent-detection tests/environment-dependent.

Because the comment immediately above this call says the override prevents picking up real-machine CLIs, consider passing an explicit environment when the override is active, e.g. `env: homeOverride ? {} : process.env`, or adding the npm-prefix env to the cache/isolation contract intentionally.

mrcfps (commented):
@fancyboi999 thanks for the careful shared-helper fix here — I verified the daemon and packaged PATH lists now use the same source of truth, and the regression coverage hits the new npm-prefix cases. I found one merge-safe test-isolation wrinkle worth tightening so the preserved `OD_AGENT_HOME` hook stays deterministic across developer/CI environments.


apps/daemon/src/agents.ts
fancyboi999: Good catch — verified the leak is real. Reproduced locally before the fix:

```js
process.env.OD_AGENT_HOME = '<tmp sandbox>';   // empty
process.env.PATH = '/usr/bin:/bin';             // launchd-default
process.env.NPM_CONFIG_PREFIX = '/Users/me/.npm-global'; // dev-machine state
resolveAgentExecutable({ bin: 'codex' });
// → '/Users/me/.npm-global/bin/codex'  ← real-machine binary leaks in
```

Pushed in 2767a58. The wrapper now passes `env: homeOverride ? {} : process.env` alongside the existing `includeSystemBins` gate, so when `OD_AGENT_HOME` is set neither Homebrew/system bins nor `$NPM_CONFIG_PREFIX/bin` can sneak past the sandbox boundary.

Added a regression test in `apps/daemon/tests/agents.test.ts` that sets `OD_AGENT_HOME` to one tmp dir, `NPM_CONFIG_PREFIX` to a different tmp dir containing a fake `gemini`, and asserts `resolveAgentExecutable` returns `null` — without the fix this test would resolve the real-prefix binary.

apps/daemon/tests/agents.test.ts
```diff
@@ -762,6 +762,96 @@ fsTest(
...
+
+      const resolved = resolveAgentExecutable({ bin: 'gemini' });
+      assert.equal(
+        resolved,
+        null,
+        `OD_AGENT_HOME sandbox must not see the real $NPM_CONFIG_PREFIX bin; ` +
+          `got ${resolved}`,
+      );
+    } finally {
+      delete process.env.NPM_CONFIG_PREFIX;
```
mrcfps: **Non-blocking:** please restore any pre-existing `NPM_CONFIG_PREFIX` value instead of always deleting it here. This regression intentionally sets `process.env.NPM_CONFIG_PREFIX = realPrefix` to prove the sandboxed `OD_AGENT_HOME` lookup ignores the caller's npm prefix, but the `finally` block now unconditionally removes the variable afterward. On a developer machine or CI runner that already exported `NPM_CONFIG_PREFIX`, this test mutates the process-wide environment for every later test in the same Vitest worker, which can make follow-up env-sensitive assertions depend on test order. A small snapshot/restore keeps the new coverage while preserving the original environment, e.g. capture `const originalNpmConfigPrefix = process.env.NPM_CONFIG_PREFIX;` before assigning, then in `finally` restore it when defined and delete only when it was originally absent.

mrcfps (commented):
@fancyboi999 thanks for the thoughtful follow-up on the shared toolchain helper and the OD_AGENT_HOME isolation fix — the main daemon/packaged PATH behavior looks sound, and the regression coverage now exercises the npm-global, npm-packages, env-prefix, whitespace-prefix, and sandbox leakage cases. I found one small merge-safe test isolation cleanup that would keep this suite friendlier to developer/CI environments that already export npm prefix settings.


apps/daemon/tests/agents.test.ts
fancyboi999: Good call — the unconditional `delete` would clobber any pre-exported `NPM_CONFIG_PREFIX` for the rest of the worker. Pushed in 5bc62e7.

The file already has a save-original/restore-in-`afterEach` pattern for `OD_AGENT_HOME`, `OD_DAEMON_URL`, and `OD_TOOL_TOKEN` (lines 38–75), so I added `originalNpmConfigPrefix` to that same block and dropped the in-test `delete`. Same coverage, no worker-state mutation regardless of how the developer/CI runner started.

packages/platform/src/index.ts
```diff
@@ -409,3 +412,108 @@ export async function readLogTail(filePath: string, maxLines = 80): Promise<stri
...
+  // pointed it. Covers setups that don't match either of the conventions
+  // above (e.g. corporate provisioning that exports a custom prefix).
+  // Trim before length-checking so accidental whitespace-only values
+  // (`NPM_CONFIG_PREFIX=" "`) do not produce a `/bin`-suffixed garbage
+  // entry.
+  const npmPrefixRaw = env.NPM_CONFIG_PREFIX ?? env.npm_config_prefix;
+  if (typeof npmPrefixRaw === "string") {
+    const npmPrefix = npmPrefixRaw.trim();
+    if (npmPrefix.length > 0) {
+      dirs.push(join(npmPrefix, "bin"));
```
mrcfps, line 467: **Non-blocking:** please give the explicit npm prefix precedence over the guessed npm-prefix locations. Right now `dirs` already contains `~/.npm-global/bin` and `~/.npm-packages/bin`, and this changed block appends `$NPM_CONFIG_PREFIX/bin` after them. Both `resolveOnPath()` and the packaged PATH builder preserve this order, so a user who has moved npm to a configured corporate/custom prefix can still resolve an older `gemini`/agent binary from one of the conventional guessed prefixes first when the GUI launch PATH is minimal. Since the env var is the user's explicit npm configuration, consider inserting the trimmed `$NPM_CONFIG_PREFIX/bin` before the conventional guessed npm dirs, and add an order-focused helper test that asserts the configured prefix appears before `~/.npm-global/bin` / `~/.npm-packages/bin`.

mrcfps (commented):
@fancyboi999 thanks for the thoughtful follow-ups here — I reviewed the current head after the sandbox/env fixes, and the shared daemon/packaged helper direction looks solid. I found one merge-safe precedence issue around explicit npm prefixes that would be worth tightening.


packages/platform/src/index.ts
fancyboi999: Agreed — explicit beats convention. Pushed in 6641c70.

Reasoning: npm's own resolution order is env > .npmrc > default, so a binary that `npm i -g` *currently* installs lives at `$NPM_CONFIG_PREFIX/bin`. The conventional `~/.npm-global` / `~/.npm-packages` paths often hold *stale* installs from a prefix the user has since rewritten. Searching the env-driven prefix first matches npm's contract.

Changes:
- `packages/platform/src/index.ts`: moved the trimmed-env push above the conventional `~/.npm-global` / `~/.npm-packages` entries (kept the trim guard intact).
- `packages/platform/tests/index.test.ts`: new order-focused vitest case asserting `$NPM_CONFIG_PREFIX/bin`'s index is strictly less than both conventional indices.

Both `resolveOnPath()` and the packaged PATH builder preserve insertion order, so first hit wins and the new ordering propagates without further changes.

`pnpm --filter @open-design/platform test` → 25 passed (was 24).

packages/platform/src/index.ts
```diff
@@ -409,3 +412,115 @@ export async function readLogTail(filePath: string, maxLines = 80): Promise<stri
...
+  // user's *current* npm configuration, so a binary installed via
+  // `npm i -g` today lives at `<prefix>/bin`. The conventional guesses
+  // (`~/.npm-global`, `~/.npm-packages`) often hold *stale* installs
+  // from an older prefix the user has since rewritten — searching the
+  // env-driven prefix first gives "explicit beats convention" semantics
+  // and matches npm's own resolution order (env > .npmrc > default).
+  // Trim before length-checking so accidental whitespace-only values
+  // (`NPM_CONFIG_PREFIX=" "`) do not produce a `/bin`-suffixed garbage
+  // entry.
+  const npmPrefixRaw = env.NPM_CONFIG_PREFIX ?? env.npm_config_prefix;
```
mrcfps, line 463: **Non-blocking:** please extend the explicit-prefix precedence fix to cover `~/.local/bin` too. The changed helper now inserts `$NPM_CONFIG_PREFIX/bin` before `~/.npm-global` / `~/.npm-packages`, but `~/.local/bin` is still emitted earlier in the initial `dirs` array and the comment below calls it npm's canonical example. Because both `resolveOnPath()` and the packaged PATH builder preserve this order, a stale agent binary in `~/.local/bin` can still win over the user's current explicit npm prefix under a minimal GUI-launch PATH, which weakens the intended "explicit beats convention" behavior. Consider moving the env-derived prefix ahead of the conventional home guesses that can contain npm globals (at least ahead of `~/.local/bin`) and adding an order-focused test that asserts `$NPM_CONFIG_PREFIX/bin` precedes `~/.local/bin` as well as `~/.npm-global/bin` / `~/.npm-packages/bin`.

mrcfps (commented):
@fancyboi999 thanks for the quick follow-up on the shared toolchain helper — the daemon/packaged source-of-truth direction still looks solid, and I verified the current head now handles the earlier sandbox and npm-prefix ordering cases. I found one remaining merge-safe ordering edge around the explicit npm prefix versus the canonical `~/.local/bin` guess. 👍


packages/platform/src/index.ts
fancyboi999: You're right — the previous round only fixed the npm-prefix block; `~/.local/bin` still cut in front. Pushed in 716a543.

Reasoning extended: `~/.local/bin` isn't only npm's canonical location, it's also a shared dumping ground for `pip --user`, `cargo install`, and hand-built binaries. A stale npm artifact left there from a previous prefix can collide with a `npm i -g`'d agent under the *current* prefix. Putting `$NPM_CONFIG_PREFIX/bin` at the head of `dirs` gives "explicit beats convention" semantics across the whole list, matching npm's own resolution order (env > .npmrc > default) end-to-end.

Changes:
- `packages/platform/src/index.ts`: built `dirs` empty-first, push `$NPM_CONFIG_PREFIX/bin` (when set + non-whitespace) before any conventional entry; conventional locations now follow in a single block. Trim guard preserved.
- `packages/platform/tests/index.test.ts`: tightened the order test to assert `explicitIdx === 0` and that `~/.local/bin`'s index is strictly greater (was: only `~/.npm-global`/`~/.npm-packages`). A future drift that re-introduces a path before the explicit prefix would now fail loudly.

Validation:
- `pnpm --filter @open-design/platform test` → 25 passed (no count change; tightened existing assertion)
- `pnpm --filter @open-design/daemon exec vitest run tests/agents.test.ts` → 65 passed
- `pnpm --filter @open-design/platform typecheck` → clean

When `$NPM_CONFIG_PREFIX` is unset, the helper returns identical ordering to before this round (the prefix push is a no-op), so this round is purely additive for users who haven't configured an explicit prefix.

mrcfps (approved):
@fancyboi999 I reviewed the current head after the shared platform helper extraction and follow-up fixes. The daemon resolver and packaged PATH builder now share the same toolchain-bin source of truth, the OD_AGENT_HOME sandboxing avoids leaking caller npm-prefix state, and the platform/daemon regression tests cover the new npm-global, npm-packages, explicit-prefix ordering, whitespace-prefix, and version-manager cases. Thanks for the careful iteration on this fix — this looks ready from my implementation review. 🚀

PR #618, opened 2026-05-06
feat(daemon): add model name to pi initial status and RPC abort on cancel

monotykamary:
## Summary

Two targeted parity fixes for the pi adapter:

### 1. Model name in initial status

Before this change, pi emitted `status: working` on `agent_start` with no model info. Every other structured-stream adapter (Claude Code, Copilot, Gemini, Cursor Agent) emits `status: initializing` with the resolved model name, which the web UI renders as `"pi · claude-sonnet-4-5"` in the assistant header.

This fix emits `status: initializing` with the model name from `attachPiRpcSession` immediately on session start — before pi even responds — so the UI has the model name at the earliest possible moment. The `agent_start` → `status: working` transition then follows naturally from pi itself.

The web UI already had the rendering lane (`translateAgentEvent` reads `data.model` from status events; `AssistantMessage.tsx` looks for `status: initializing` with detail). It was just never being fed for pi.

### 2. RPC abort instead of raw SIGTERM

Previously, cancelling a pi run sent `SIGTERM` directly. This change sends the RPC `abort` command instead via `{"type":"abort"}`, giving pi a chance to clean up gracefully. Falls back to SIGTERM after `PI_ABORT_GRACE_MS` (default 3s).

The `runs.ts` cancel path now checks `run.acpSession?.abort()` first — a generic interface that ACP adapters (Devin, Hermes, Kimi, Kiro, Kilo, Vibe) can also adopt.

## Changes

| File | Change |
|---|---|
| `apps/daemon/src/pi-rpc.ts` | Emit `status:initializing` with model; add `abort()` method; add `stdinOpen` guard |
| `apps/daemon/src/runs.ts` | `cancel()` prefers `run.acpSession.abort()` over raw SIGTERM |
| `apps/daemon/src/server.ts` | Wire `run.acpSession = acpSession` after spawn |
| `apps/daemon/tests/pi-rpc.test.ts` | 4 new tests (initializing, abort format, stdin-closed guard) |

## Verification

- All 30 pi-rpc tests pass
- Typecheck clean on changed files
- No pre-existing test regressions

### DIFF
--- a/apps/daemon/src/pi-rpc.ts
+++ b/apps/daemon/src/pi-rpc.ts
@@ -268,6 +268,12 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
 
   // ---- Inbound: parse stdout events ----
   const parser = createJsonLineStream((raw) => {
+    // Once finished (agent_end or abort), stop processing — the run is
+    // over, so no more agent events should be emitted. We still drain
+    // stdout via parser.feed() so the pipe doesn't break; we just skip
+    // acting on the parsed objects.
+    if (finished) return;
+
     // Extension UI requests: auto-resolve to keep pi unblocked.
     if (raw.type === 'extension_ui_request') {
       replyExtensionUi(child.stdin, raw);
--- a/apps/daemon/src/pi-rpc.ts
+++ b/apps/daemon/src/pi-rpc.ts
@@ -199,13 +199,21 @@ export function mapPiRpcEvent(raw, send, ctx) {
 /**
  * Attach a pi RPC session to a spawned child process.
  *
+ * Emits `status: initializing` with the model name immediately so the UI
+ * can show "pi · claude-sonnet-4-5" like every other adapter. Then sends
+ * the prompt via RPC and streams events back.
+ *
+ * The returned `abort()` method sends an RPC `abort` command instead of
+ * raw SIGTERM, giving pi a chance to clean up gracefully. Falls back to
+ * SIGTERM after PI_ABORT_GRACE_MS if pi doesn't exit on its own.
+ *
  * @param {object} opts
  * @param {import('node:child_process').ChildProcess} opts.child  - spawned pi process
  * @param {string} opts.prompt   - composed user message
  * @param {string} [opts.cwd]    - working directory
  * @param {string|null} [opts.model] - model id (null = default)
  * @param {function} opts.send   - SSE send function
- * @returns {{ hasFatalError(): boolean }}
+ * @returns {{ hasFatalError(): boolean, abort(): void }}
  */
 export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
   const runStartedAt = Date.now();
@@ -214,11 +222,17 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
   const sentFirstToken = { value: false };
 
   let nextRpcId = 1;
+  let stdinOpen = true;
 
   function sendCommand(writable, type, params = {}) {
+    if (!stdinOpen) return null;
     const id = nextRpcId++;
-    writable.write(`${JSON.stringify({ id, type, ...params })}\n`);
-    return id;
+    try {
+      writable.write(`${JSON.stringify({ id, type, ...params })}\n`);
+      return id;
+    } catch {
+      return null;
+    }
   }
 
   // Track the prompt request id so we know when the prompt response arrives.
@@ -232,11 +246,22 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
     if (!child.killed) child.kill('SIGTERM');
   };
 
+  // Emit initial status with model name immediately — before pi even
+  // responds — so the UI header shows the model name at session start.
+  send('agent', {
+    type: 'status',
+    label: 'initializing',
+    model: typeof model === 'string' && model ? model : null,
+  });
+
   // ---- Outbound: send the prompt via RPC ----
   child.stdin.on('error', (err) => {
     if (err.code !== 'EPIPE') {
       fail(`stdin: ${err.message}`);
     }
+  });
+  child.stdin.on('close', () => {
+    stdinOpen = false;
   });
 
   promptRpcId = sendCommand(child.stdin, 'prompt', { message: prompt });
@@ -292,6 +317,17 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
   return {
     hasFatalError() {
       return fatal;
+    },
+    abort() {
+      // RPC abort: send the abort command so pi can clean up gracefully.
+      // Fall back to SIGTERM after PI_ABORT_GRACE_MS if pi doesn't exit.
+      if (finished || child.killed) return;
+      finished = true;
+      sendCommand(child.stdin, 'abort');
+      const graceMs = Number(process.env.PI_ABORT_GRACE_MS) || 3000;
+      setTimeout(() => {
+        if (!child.killed) child.kill('SIGTERM');
+      }, graceMs).unref();
     },
   };
 }
--- a/apps/daemon/src/pi-rpc.ts
+++ b/apps/daemon/src/pi-rpc.ts
@@ -203,9 +203,10 @@ export function mapPiRpcEvent(raw, send, ctx) {
  * can show "pi · claude-sonnet-4-5" like every other adapter. Then sends
  * the prompt via RPC and streams events back.
  *
- * The returned `abort()` method sends an RPC `abort` command instead of
- * raw SIGTERM, giving pi a chance to clean up gracefully. Falls back to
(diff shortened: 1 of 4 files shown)


apps/daemon/src/pi-rpc.ts
```diff
@@ -293,6 +318,17 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
     hasFatalError() {
       return fatal;
     },
+    abort() {
+      // RPC abort: send the abort command so pi can clean up gracefully.
+      // Fall back to SIGTERM after PI_ABORT_GRACE_MS if pi doesn't exit.
+      if (finished || child.killed) return;
+      finished = true;
```
mrcfps, line 334: `abort()` sets `finished = true` and sends the RPC abort, but the stdout listener above still feeds every later chunk into `mapPiRpcEvent` without checking `finished`. Because this change intentionally keeps the pi child alive for `PI_ABORT_GRACE_MS`, any `message_update`, status, tool, or usage events that arrive while pi is honoring the abort will still be sent to the canceled run, so the assistant message can keep growing after the user cancels. Please gate the stdout parser/mapper once abort is requested (for example with an `aborted`/`finished` check before mapping agent events, while still allowing any necessary cleanup), and add a session-level fixture that aborts and then feeds a `message_update` to prove no post-cancel agent events are emitted.

mrcfps (changes requested):
@monotykamary thanks for the targeted pi adapter parity work here — the initial model-status wiring is clear, and the RPC abort direction is a nice improvement. I found one cancel-path correctness issue that needs fixing before this is safe to merge. 🙏


lefarcen:
Hi @monotykamary! 🎉
Thanks for the contribution — these pi adapter parity fixes look well-scoped and the RPC abort direction is a solid improvement.

— open-design team

monotykamary:
Both review comments flagged the same issue: `abort()` sets `finished = true` but the stdout parser kept calling `mapPiRpcEvent`, so agent events (text_delta, tool, status) could still be emitted during the abort grace window.

Fixed in d7d97d9 — added a `finished` guard at the top of the parser callback in `attachPiRpcSession`. Once `finished` is true (set by either `agent_end` or `abort()`), no more agent events are forwarded. The parser still drains stdout cleanly (no pipe breaks), it just stops acting on parsed objects.

Added a test that aborts mid-session, then feeds `message_update` + tool events + `agent_end`, proving zero post-abort agent events are emitted.

apps/daemon/src/pi-rpc.ts
monotykamary: Fixed in d7d97d9.

apps/daemon/src/pi-rpc.ts
```diff
@@ -293,6 +318,17 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
...
+      sendCommand(child.stdin, 'abort');
```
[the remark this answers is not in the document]
monotykamary: Fixed in d7d97d9.

apps/daemon/src/runs.ts
```diff
@@ -125,8 +125,15 @@ export function createChatRunService({
     if (!TERMINAL_RUN_STATUSES.has(run.status)) {
       run.cancelRequested = true;
       run.updatedAt = Date.now();
-      if (run.child && !run.child.killed) run.child.kill('SIGTERM');
-      else finish(run, 'canceled', null, 'SIGTERM');
+      // Prefer RPC-level abort for agents that support it (pi, ACP adapters).
+      // Falls back to raw SIGTERM when the session doesn't expose abort().
+      if (run.acpSession?.abort) {
```
lefarcen, line 132: P3: The generic `run.acpSession?.abort()` path assumes the session implementation owns all termination fallback behavior. That's true for pi, but it's not encoded in the interface — future ACP adapters could implement this as "send abort only" and leave runs active indefinitely if the child never exits. Consider documenting the contract, or have `runs.cancel()` own a SIGTERM fallback timer when delegating to `abort()`.

apps/daemon/tests/pi-rpc.test.ts
```diff
@@ -447,3 +447,166 @@ test('pi RPC: no duplicate usage when both message_end and turn_end carry usage'
   assert.equal(usageEvents.length, 1, 'should emit exactly one usage event per turn');
   assert.equal(usageEvents[0].usage.input_tokens, 100);
 });
+
+// ─── attachPiRpcSession initial status ─────────────────────────────────────
+
+test('attachPiRpcSession emits status:initializing with model name', () => {
```
lefarcen: P3: The new initial-status tests duplicate the exact `send(...)` payload inline rather than exercising `attachPiRpcSession`, so they'd still pass if the real function stopped emitting the event or used wrong model normalization. Consider instantiating `attachPiRpcSession` with mock stdin/stdout streams and asserting the emitted events/order from the real function.

apps/daemon/tests/pi-rpc.test.ts
```diff
@@ -447,3 +447,166 @@ test('pi RPC: no duplicate usage when both message_end and turn_end carry usage'
...
+
+  assert.equal(events.length, 1);
+  assert.equal(events[0].type, 'status');
+  assert.equal(events[0].label, 'initializing');
+  assert.equal(events[0].model, null);
+});
+
+// ─── abort command format ──────────────────────────────────────────────────
+
+test('pi RPC: abort command writes well-formed JSON', async () => {
```
lefarcen: P3: The abort tests duplicate the private `sendCommand` helper instead of calling `attachPiRpcSession().abort()`, so they don't verify the actual cancel behavior (abort writes to child stdin, doesn't write after close, schedules SIGTERM fallback). Consider using a mock child process/EventEmitter and fake timers against the real returned session object.

lefarcen (commented):
Hey @monotykamary — mrcfps already covered the cancel-path correctness issue, so I'll add a few non-blocking suggestions on test coverage and interface robustness:

The initial status wiring looks solid, and the RPC abort direction is a nice improvement. Once mrcfps's blocking feedback is addressed, these minor points below would strengthen the implementation. See inline comments for details.

apps/daemon/src/runs.ts
monotykamary: Moved the SIGTERM fallback from `abort()` to `cancel()` in 3f14a2c — `abort()` now only sends the RPC abort command, and `cancel()` owns a `PI_ABORT_GRACE_MS` timer that falls back to SIGTERM. So a misbehaving session can't leave the child alive indefinitely; the termination guarantee is centralized in the caller.

apps/daemon/tests/pi-rpc.test.ts
monotykamary: Replaced the inline `send(...)` stubs with integration tests that call `attachPiRpcSession` against a mock child process (PassThrough stdin/stdout + EventEmitter). Now catches regressions if the real function stops emitting or changes model normalization. See 3f14a2c.

apps/daemon/tests/pi-rpc.test.ts
monotykamary: Replaced the duplicated `sendCommand` helper with integration tests that call `session.abort()` on the real returned session object against a mock child process. Verifies abort writes to stdin, is idempotent, is no-op after stdin close, and gates post-abort stdout events. See 3f14a2c.

lefarcen:
The P2 fix looks good — the `finished` guard at the top of the parser callback correctly suppresses post-abort agent events, and the new test proves zero events are emitted after `abort()`. The SIGTERM fallback is now owned by `runs.cancel()` with a 3s grace window, which addresses the interface robustness concern too.

mrcfps's blocking issue is resolved. The P3 suggestions I mentioned earlier (test coverage improvements) are nice-to-haves and don't block merge.

lefarcen:
All suggestions fully addressed in 3f14a2c — the SIGTERM fallback is now centralized in `runs.cancel()`, and the tests use the real `attachPiRpcSession` and `session.abort()` instead of duplicating helpers. Excellent work on the iteration! 🎉

mrcfps (approved):
@monotykamary I reviewed the pi RPC changes at this head: the initial `status: initializing` event now carries the normalized model, cancel now routes through session-level RPC abort with the SIGTERM fallback centralized in `runs.cancel()`, and the stdout gate plus integration-style tests cover the prior post-abort event concern. I don't see any remaining actionable issues in the changed ranges — thanks for the careful follow-up fixes here! 🙏


lefarcen:
Hey @monotykamary! 🎉

PR #618 merged — thanks for the careful pi parity work here. The model-in-status fix closes a UI gap (pi now shows its model name at the same moment as every other adapter), and the RPC abort path gives pi a graceful-shutdown lane that the other ACP adapters can adopt.

Appreciate the thorough iteration on the cancel-path correctness + the integration tests. 🙏

— open-design team

PR #620, opened 2026-05-06
[codex] Implement manual edit mode

alchemistklk:
## Summary

- Adds source-backed manual Edit mode for HTML artifacts, separate from the existing Comment/Tweaks AI flow.
- Adds iframe target discovery, source patch helpers, undo/redo snapshots, and the manual edit side panel with Content, Style, Attributes, Html, and Source tabs.
- Persists accepted requirements and implementation plan for follow-up migration work.
- Covers edit-mode patching, bridge target discovery, panel behavior, and Playwright smoke flows.

## Validation

- `pnpm guard`
- `pnpm --dir e2e exec playwright test -c playwright.config.ts specs/manual-edit-mode.spec.ts`

Note: local validation printed the existing engine warning because this shell is on Node `v25.9.0`; the repo asks for Node `~24`.


alchemistklk:
@codex review

lefarcen (commented):
Hi @alchemistklk! Solid implementation of manual edit mode with good test coverage. Here are a few findings across code correctness and design:

**Code Review (Lens A)**

The implementation is clean with strong type safety and comprehensive tests. Found a few edge cases to address:

**P2 Security — XSS via outerHTML replacement** (`source-patches.ts:185-195`)
The `replaceOuterHtml` function accepts arbitrary HTML without sanitization:
```typescript
template.innerHTML = html.trim();
```
While the sandbox iframe provides some protection, this could allow XSS if combined with other vulnerabilities. Consider:
- Add DOMPurify sanitization before `template.innerHTML = html.trim()`
- Or at minimum, document the security boundary and why trusting user HTML is acceptable here (e.g., "user is editing their own artifact")

**P2 Race condition — Multiple simultaneous saves** (`FileViewer.tsx:2099-2113`)
`applyManualEdit` and `undoManualEdit`/`redoManualEdit` don't prevent concurrent calls. If a user rapidly clicks Apply + Undo, or two Apply buttons in succession, the history stack could become inconsistent. Consider:
- Add a `saving` state flag that disables all edit actions while `writeProjectTextFile` is pending
- Or use a queue pattern for edit operations

**P3 Type narrowing — `parseSource` DOMParser check** (`source-patches.ts:114-123`)
The fallback path `document.implementation.createHTMLDocument` only exists in browser contexts, but this code runs in web workers too. The double-check pattern works, but the comment could clarify: "// Fallback for Node-like environments with jsdom".

**P3 Accessibility — Tab panel ARIA** (`ManualEditPanel.tsx:109-115`)
The tabs have `role="tab"` but the container lacks `role="tablist"` wrapping the buttons themselves (it's on a parent div). Move `role="tablist"` to the immediate `<div>` containing the buttons for proper ARIA tree.

**P3 Test gap — History deduplication** (`tests/edit-mode/source-patches.test.ts`)
The patch functions don't check for no-op edits (e.g., setting text to the same value). Each call creates a history entry even if nothing changed. Not blocking, but worth noting for future UX polish (compare `beforeSource` === `afterSource` and skip history push).

---

**Design Review (Lens B)** — `manual-edit-mode-requirements.md` / `manual-edit-mode-implementation.md`

The requirements doc is thorough (544 lines!). A few reasoning gaps:

**P2 Missing quantification — Performance targets**
The requirements doc doesn't specify acceptable latency for:
- Target discovery (how many elements before it lags?)
- Undo/redo (is 100-entry history acceptable?)
- Source parsing/serialization round-trip time

For a visual editor, these matter. Even a rough "should handle 500-element pages without visible lag" would help future maintainers decide when to optimize.

**P3 Open question — Migration path from Comment/Tweaks**
The requirements state manual edit is "separate from Comment/Tweaks AI flow", but doesn't address: will users expect to switch modes mid-session? If they have pending AI tweaks and enable manual edit, what happens? Noting this as TBD would signal it's on the radar.

**P3 Prior art — No comparison to browser DevTools**
The manual editor UI (Content/Style/Attributes tabs) resembles Chrome DevTools Elements panel. A one-paragraph comparison ("We simplified X, kept Y, diverged on Z because...") would help reviewers understand design choices vs. reinventing.

---

**Overall**: Production-ready with the XSS sanitization and save race addressed. The comprehensive test suite (unit + Playwright) gives confidence in the core edit/undo/redo flow. Nice work on the i18n coverage and requirements documentation! 🎉

apps/web/src/edit-mode/source-patches.ts
```diff
@@ -0,0 +1,218 @@
...
+  const el = findEditableElement(doc, patch.id);
+  if (!el) return { ok: false, source, error: `Target not found: ${patch.id}` };
+
+  if (patch.kind === 'set-text') {
+    if (hasElementChildren(el)) {
+      return { ok: false, source, error: 'This element contains nested markup. Use the HTML tab instead.' };
+    }
+    el.textContent = patch.value;
+  } else if (patch.kind === 'set-link') {
+    el.textContent = patch.text;
```
mrcfps: `set-link` currently rewrites the entire selected link with `el.textContent = patch.text` before setting `href`. That means a common generated CTA such as `<a data-od-id="cta"><span>Buy now</span><svg>…</svg></a>` loses its span/icon structure even if the user only intended to change the URL in the Content tab. `set-text` already rejects elements with nested markup for this reason, but links bypass that guard. Please preserve child markup when only `href` changes, and either reject label edits for links with element children (pointing users to the HTML tab) or update only a safe text node instead of flattening the whole anchor.

apps/web/src/edit-mode/bridge.ts
```diff
@@ -0,0 +1,163 @@
...
+    for (var i = 0; i < nodes.length; i++) {
+      var rect = nodes[i].getBoundingClientRect();
+      if (rect.width < 4 || rect.height < 4) continue;
+      targets.push(targetFrom(nodes[i]));
+    }
+    return targets;
+  }
+  function postTargets(){
+    if (!enabled) return;
+    window.parent.postMessage({ type: 'od-edit-targets', targets: allTargets() }, '*');
```
mrcfps: Non-blocking: target discovery posts every discovered element with full `outerHtml`, then sends the entire list on load/resize. Because containers like `main` and `section` can include most of the page, larger generated artifacts can produce multi-MB `postMessage` payloads and duplicate the same HTML many times, making edit mode feel frozen before the user has selected anything. Consider capping the target count/payload, truncating or omitting `outerHtml` from the bulk `od-edit-targets` message, and fetching/sending full HTML only for the selected target.

mrcfps (changes requested):
@alchemistklk Thanks for putting this manual edit flow together — the structure and coverage make the feature easy to follow. I reviewed the source patching and iframe bridge changes and found one blocker where link content edits can corrupt nested CTA markup, plus one non-blocking payload-size concern for target discovery.


apps/web/src/components/FileViewer.tsx
```diff
@@ -2321,6 +2357,116 @@ function HtmlViewer({
...
+    setManualEditHistory((current) => [entry, ...current]);
+    setManualEditUndone([]);
+    setManualEditDraft((current) => ({ ...current, fullSource: result.source }));
+    await onFileSaved?.();
+  }
+
+  async function undoManualEdit() {
+    const [latest, ...rest] = manualEditHistory;
+    if (!latest) return;
+    const saved = await writeProjectTextFile(projectId, file.name, latest.beforeSource, {
```
mrcfps: `undoManualEdit` writes the snapshot from `latest.beforeSource` without checking that the currently loaded/persisted file still equals the history entry's `afterSource`. The same component reloads `source` whenever `file.mtime` changes, but the manual edit history is only cleared on `file.name`, so an agent run, paste/upload save, or other refresh that updates this file after a manual edit leaves stale history entries active. Pressing Undo after that refresh can overwrite the newer file contents with the old snapshot, which is a realistic data-loss path for the source-backed editor. Please track a source revision/mtime (or compare `source` to `latest.afterSource`) before undo/redo and either clear/disable history on external file changes or reject the undo with a refresh-needed message instead of writing the stale snapshot.

apps/web/src/edit-mode/bridge.ts
```diff
@@ -0,0 +1,181 @@
...
+  }
+  function closestTarget(event){
+    var el = event.target;
+    while (el && el !== document.documentElement) {
+      if (el.getAttribute && el !== document.body && el !== document.documentElement) return el;
```
mrcfps: `closestTarget` currently returns the first non-body event target, without requiring it to match the discovery selector or preferring an editable ancestor. On common generated markup like `<a data-od-id="cta"><span>Buy</span><svg>…</svg></a>`, clicking the span or SVG path selects that child as a text/path target instead of the link, so the Content tab cannot edit `href` and subsequent patches can hit a nested node rather than the intended CTA. Please walk up to the nearest meaningful editable target (for example explicit `data-od-id`/`data-od-edit`, then `a`/`button`/other discovery-selector ancestors, while skipping host nodes) so clicks inside a control select the same element that appears in the layer list.

mrcfps (changes requested):
@alchemistklk Thanks for the follow-up fixes here — the nested-link and bridge payload improvements are much safer. I found one remaining data-loss issue in the manual edit history flow that should be addressed before merge, plus one smaller selection polish item for nested controls.


apps/web/src/components/FileViewer.tsx
```diff
@@ -2321,6 +2357,139 @@ function HtmlViewer({
...
+      alt: fields.alt ?? target.fields.alt ?? '',
+      styles: readManualEditStyles(base, target.id),
+      attributesText: JSON.stringify(readManualEditAttributes(base, target.id), null, 2),
+      outerHtml: readManualEditOuterHtml(base, target.id) || target.outerHtml,
+      fullSource: base,
+    });
+    setManualEditError(null);
+  }
+
+  async function applyManualEdit(patch: ManualEditPatch, label: string) {
```
mrcfps, line 2412: Blocking: please serialize or disable manual edit writes while a save is in flight. `applyManualEdit` derives `result.source` from the current React `source` snapshot, but that state is not updated until `writeProjectTextFile` resolves. A user can apply content, quickly apply style/attributes/source before the first write finishes, and the second patch is computed from the old source, so whichever write finishes last can silently drop the earlier edit while still adding history entries. This is still a realistic data-loss path for the source-backed editor, even with the undo/redo stale-source check. Add a pending write flag/queue around apply/undo/redo actions, or re-read/compare the persisted file before each apply and reject/merge when it no longer matches the snapshot used to compute the patch.

apps/web/src/edit-mode/bridge.ts
```diff
@@ -0,0 +1,204 @@
...
+    if (!el || !el.hasAttribute) return false;
+    if (el.hasAttribute('data-od-id') || el.hasAttribute('data-od-edit')) return true;
+    var tag = el.tagName ? el.tagName.toLowerCase() : '';
+    return tag === 'a' || tag === 'button';
+  }
+  function inferKind(el){
+    var explicit = el.getAttribute('data-od-edit');
+    if (explicit) return explicit;
+    var tag = el.tagName ? el.tagName.toLowerCase() : '';
+    if (tag === 'a' || tag === 'button') return 'link';
```
mrcfps: Non-blocking: buttons are currently inferred as `link` targets, which makes the Content tab expose an `href` field and sends a `set-link` patch. `source-patches.ts` then writes that value as a literal `href` attribute, so editing a generated `<button data-od-id="cta">Buy</button>` can persist `<button href="...">` even though the button's navigation behavior has not changed. Since the requirements only make `href` applicable where the element actually supports it, consider treating only `<a>` as `link`, or carrying `tagName` through the content patch and ignoring/hiding `href` for buttons while still allowing label edits.

mrcfps (changes requested):
@alchemistklk Thanks for the follow-up fixes — the source mapping, nested-link handling, and undo/redo stale-source guard are much improved. I found one remaining save-order issue that can still drop manual edits, plus one smaller button-target correctness issue to consider. 🚧


mrcfps (approved):

PR #621, opened 2026-05-06
Fix design system selection persistence

### LINKED ISSUE #603: Selected design system is not persisting after the project is created.
When selecting a theme and creating any item from the homepage, I don't think it is picking it and using the selected design system.
For ex:, while creating a slide deck or anything, we have an option to select from different design system like claude or clay or BMW or Opencode etc. 
When we select any of those and create a project, I don't think the theme is being persisted in the created project. We still have to explain the design system to it. 
There is also no option to see which design system is selected for the project created after it has been created. 

Image

Sid-Qin:
## Summary
- Default the New Project design-system picker to the configured default design system when that system exists in the catalog.
- Preserve explicit user changes, including choosing None, instead of reapplying the default after interaction.
- Keep image/video/audio project creation from inheriting a hidden design-system selection.

## Root cause
The New Project panel received `defaultDesignSystemId` but initialized `selectedDsIds` to an empty array. As a result, structured project surfaces displayed `None — freeform` and created projects with `designSystemId: null` unless the user manually reselected a design system.

## User impact
New prototype, deck, template, and other structured projects now visibly start with the configured default design system and persist that selection into project creation. Media surfaces remain design-system-free because their picker is hidden.

## Testing
- `pnpm guard`
- `pnpm typecheck`
- `git diff --check`

Closes #603

### DIFF
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -75,6 +75,28 @@ const TAB_LABEL_KEYS: Record<CreateTab, keyof Dict> = {
   other: 'newproj.tabOther',
 };
 
+export function defaultDesignSystemSelection(
+  defaultDesignSystemId: string | null,
+  designSystems: DesignSystemSummary[],
+): string[] {
+  if (!defaultDesignSystemId) return [];
+  return designSystems.some((d) => d.id === defaultDesignSystemId)
+    ? [defaultDesignSystemId]
+    : [];
+}
+
+export function buildDesignSystemCreateSelection(
+  showDesignSystemPicker: boolean,
+  selectedIds: string[],
+): { primary: string | null; inspirations: string[] } {
+  return showDesignSystemPicker
+    ? {
+        primary: selectedIds[0] ?? null,
+        inspirations: selectedIds.slice(1),
+      }
+    : { primary: null, inspirations: [] };
+}
+
 export function NewProjectPanel({
   skills,
   designSystems,
@@ -99,7 +121,14 @@ export function NewProjectPanel({
   // Design-system selection is now an *array* internally so the same
   // component can drive both single-select and multi-select modes without
   // duplicating state. Single-select coerces to length 0/1.
-  const [selectedDsIds, setSelectedDsIds] = useState<string[]>([]);
+  const initialDefaultDsSelection = useMemo(
+    () => defaultDesignSystemSelection(defaultDesignSystemId, designSystems),
+    [defaultDesignSystemId, designSystems],
+  );
+  const [selectedDsIds, setSelectedDsIds] = useState<string[]>(
+    () => initialDefaultDsSelection,
+  );
+  const [dsSelectionTouched, setDsSelectionTouched] = useState(false);
   const [dsMulti, setDsMulti] = useState(false);
 
   // Per-tab metadata. Tracked independently so switching tabs preserves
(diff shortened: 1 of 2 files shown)


lefarcen (commented):
Hi @Sid-Qin! Thanks for this PR — the design system default logic is much clearer now. 👍

Found a couple of edge cases that could affect live-artifact projects and catalog changes while the panel is mounted. See inline below.

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -244,8 +283,8 @@ export function NewProjectPanel({
     // and inspiration ids to empty there so the New Project panel can't
     // accidentally bind a stale DS that the user can no longer see in the
     // form (the picker is hidden for image/video/audio).
-    const primaryDs = showDesignSystemPicker ? selectedDsIds[0] ?? null : null;
-    const inspirations = showDesignSystemPicker ? selectedDsIds.slice(1) : [];
+    const { primary: primaryDs, inspirations } =
```
lefarcen, line 286: **P2 live-artifact projects also skip design system defaults**

`showDesignSystemPicker` is false for `live-artifact` in addition to image/video/audio, which means live-artifact projects still create with `designSystemId: null` even when a default is configured. Live artifacts map to prototype metadata and likely need the same design-system context.

Fix: include `tab === 'live-artifact'` in the picker visibility predicate (around line 169), or change the nulling condition at line 286 to explicitly target only media tabs (`tab === 'image' || tab === 'video' || tab === 'audio'`).

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -217,6 +251,11 @@ export function NewProjectPanel({
     });
   }
 
+  function handleDesignSystemChange(ids: string[]) {
```
lefarcen, line 254: **P2 stale design system IDs after catalog changes**

Once the user has touched the picker, `selectedDsIds` is never reconciled against the current `designSystems` catalog. If the selected/default design system is later disabled or removed while the panel remains mounted, the UI may fall back to "None", but `handleCreate` can still submit the stale hidden id.

Fix: prune `selectedDsIds` when `designSystems` changes (add a `useEffect` with `designSystems` dependency), or validate/filter ids inside `buildDesignSystemCreateSelection` against the current catalog before returning them.

mrcfps (approved):
@Sid-Qin I reviewed the New Project panel changes for default design-system initialization, the touched-state behavior that preserves explicit user choices, the media-surface nulling path, and the new component/helper coverage. The implementation looks consistent with the PR goal, and the helper extraction plus tests make the behavior easier to maintain. Thanks for tightening up this project-creation flow — nice work! 🙌


lefarcen (approved):

alchemistklk:
@Sid-Qin Thanks for the update. Before this can be treated as merge-ready, please resolve all remaining review conversations on this PR.

There are still two unresolved threads in `apps/web/src/components/NewProjectPanel.tsx`:

1. Live-artifact projects still skip design-system defaults because `showDesignSystemPicker` does not include `live-artifact`.
2. Selected design-system IDs are not pruned/validated against the current `designSystems` catalog, so a stale hidden ID can still be submitted after catalog changes.

Please either address both points in code and then resolve the conversations, or reply in-thread if you believe a conversation should be closed without a code change.

lefarcen:
Hi @Sid-Qin! 🎉

Thanks for your contribution — this PR has been merged and the design system selection persistence fix is now live.

The implementation now properly defaults the picker to the configured design system and preserves user choices across project creation.

— open-design team

PR #622, opened 2026-05-06
feat(daemon): let Codex image projects use built-in imagegen

magicseek:
## Summary
- pass the active agent id into daemon system prompt composition
- add a Codex-only override for image projects using gpt-image-* models so Codex tries built-in imagegen before the daemon OpenAI provider
- keep non-Codex agents and non-gpt-image models on the existing media dispatcher contract
- add prompt composer coverage for the Codex override and negative cases

## Verification
- pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/system-prompt-template.test.ts
- pnpm guard
- pnpm typecheck

## Notes
- I built @open-design/desktop once before rerunning root typecheck because @open-design/packaged resolves @open-design/desktop/main from desktop dist declarations.

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -44,10 +44,11 @@ const agentCapabilities = new Map();
 // user's local CLI config wins.
 //
 // `extraAllowedDirs` is a list of absolute directories the agent must be
-// permitted to read files from (skill seeds, design-system specs) that live
-// outside the project cwd. Currently only Claude Code wires this through
-// (`--add-dir`); other agents either inherit broader access or run with cwd
-// boundaries we can't widen via flags.
+// permitted to read files from (skill seeds, design-system specs, narrowly
+// scoped tool output dirs) that live outside the project cwd. Agents with a
+// documented access-widening flag wire this through (`--add-dir`); the rest
+// either inherit broader access or run with cwd boundaries we can't widen via
+// flags.
 //
 // `streamFormat` hints to the daemon how to interpret stdout:
 //   - 'claude-stream-json' : line-delimited JSON emitted by Claude Code's
@@ -214,7 +215,7 @@ export const AGENT_DEFS = [
     buildArgs: (
       _prompt,
       _imagePaths,
-      _extra,
+      extraAllowedDirs = [],
       options = {},
       runtimeContext = {},
     ) => {
@@ -232,6 +233,12 @@ export const AGENT_DEFS = [
       }
       if (runtimeContext.cwd) {
         args.push('-C', runtimeContext.cwd);
+      }
+      const dirs = (extraAllowedDirs || []).filter(
+        (d) => typeof d === 'string' && d.length > 0,
+      );
+      for (const d of dirs) {
+        args.push('--add-dir', d);
       }
       if (options.model && options.model !== 'default') {
         args.push('--model', options.model);
--- a/apps/daemon/src/prompts/system.ts
+++ b/apps/daemon/src/prompts/system.ts
@@ -33,6 +33,7 @@ import { OFFICIAL_DESIGNER_PROMPT } from './official-system.js';
 import { DISCOVERY_AND_PHILOSOPHY } from './discovery.js';
 import { DECK_FRAMEWORK_DIRECTIVE } from './deck-framework.js';
 import { MEDIA_GENERATION_CONTRACT } from './media-contract.js';
+import { IMAGE_MODELS } from '../media-models.js';
 
 type ProjectMetadata = {
   kind?: string;
@@ -77,6 +78,7 @@ export const BASE_SYSTEM_PROMPT = OFFICIAL_DESIGNER_PROMPT;
 
 export interface ComposeInput {
   agentId?: string | null | undefined;
+  includeCodexImagegenOverride?: boolean | undefined;
   skillBody?: string | undefined;
   skillName?: string | undefined;
   skillMode?:
@@ -110,6 +112,7 @@ export interface ComposeInput {
 
 export function composeSystemPrompt({
   agentId,
+  includeCodexImagegenOverride = true,
   skillBody,
   skillName,
   skillMode,
@@ -190,29 +193,57 @@ export function composeSystemPrompt({
     parts.push(MEDIA_GENERATION_CONTRACT);
   }
 
-  const codexImagegenOverride = renderCodexImagegenOverride(agentId, metadata);
-  if (codexImagegenOverride) {
-    parts.push(codexImagegenOverride);
+  if (includeCodexImagegenOverride) {
+    const codexImagegenOverride = renderCodexImagegenOverride(
+      agentId,
+      metadata,
+    );
+    if (codexImagegenOverride) {
+      parts.push(codexImagegenOverride);
+    }
   }
 
   return parts.join('');
 }
 
-function renderCodexImagegenOverride(
+const CODEX_IMAGEGEN_MODEL_IDS = new Set(
+  IMAGE_MODELS.filter(
+    (model) =>
+      model?.provider === 'openai' &&
+      typeof model?.id === 'string' &&
+      model.id.startsWith('gpt-image-'),
+  ).map((model) => model.id),
+);
+
+export function resolveCodexImagegenModelId(
+  metadata: ProjectMetadata | undefined,
+): string {
+  const imageModel =
+    typeof metadata?.imageModel === 'string' ? metadata.imageModel.trim() : '';
+  return CODEX_IMAGEGEN_MODEL_IDS.has(imageModel) ? imageModel : '';
+}
+
+export function shouldRenderCodexImagegenOverride(
+  agentId: string | null | undefined,
+  metadata: ProjectMetadata | undefined,
+): boolean {
+  const normalizedAgentId =
+    typeof agentId === 'string' ? agentId.trim().toLowerCase() : '';
+  return (
+    normalizedAgentId === 'codex' &&
+    metadata?.kind === 'image' &&
+    resolveCodexImagegenModelId(metadata).length > 0
+  );
+}
+
+export function renderCodexImagegenOverride(
   agentId: string | null | undefined,
   metadata: ProjectMetadata | undefined,
 ): string {
-  const normalizedAgentId =
-    typeof agentId === 'string' ? agentId.trim().toLowerCase() : '';
-  const imageModel =
-    typeof metadata?.imageModel === 'string' ? metadata.imageModel.trim() : '';
-  if (
-    normalizedAgentId !== 'codex' ||
-    metadata?.kind !== 'image' ||
-    !imageModel.startsWith('gpt-image-')
-  ) {
+  if (!shouldRenderCodexImagegenOverride(agentId, metadata)) {
     return '';
   }
+  const imageModel = resolveCodexImagegenModelId(metadata);
 
   return `
 
@@ -229,20 +260,30 @@ the active image skill's dispatcher wording.
 Do not require, request, or mention \`OPENAI_API_KEY\` before trying the
 built-in path. Reuse the project metadata, reference prompt template, aspect
 ratio, style notes, and the user's current brief to form the final image
-prompt. Generate the image with Codex built-in imagegen, then move or copy the
-selected output from the installed \`$imagegen\` system skill / built-in
-\`image_gen\` tool at
-\`\${CODEX_HOME:-$HOME/.codex}/generated_images/.../ig_*.png\`
-into \`$OD_PROJECT_DIR\` with a short descriptive filename. Never leave a
-project-referenced asset only under \`$CODEX_HOME\`.
+prompt. Generate the image with Codex built-in imagegen, then use the actual
+output path returned by the built-in imagegen result as the source file first.
+Only if the built-in result does not return a usable path should you search
+\`\${CODEX_HOME:-$HOME/.codex}/generated_images/.../ig_*.png\` as a fallback
+source. Never leave a project-referenced asset only under \`$CODEX_HOME\`.
 
-After copying the file into \`$OD_PROJECT_DIR\`, reply with the project-local
+Copy or move the selected generated file into \`$OD_PROJECT_DIR\` with a short
+descriptive filename, then verify the exact destination file exists under
+\`$OD_PROJECT_DIR\` before claiming success. If reading the source path,
+creating the destination directory, copying/moving, or verifying the copied
+asset fails, report the exact source path, destination path, and access/copy
+error. Do not claim success, silently fall back, or ask about OpenAI/Azure
+fallback after a generated image exists but the project copy fails; stop after
+reporting the failure unless the user explicitly chooses fallback in a later
+turn, because fallback may create a different image.
+
+After the file exists under \`$OD_PROJECT_DIR\`, reply with the project-local
 filename and a short summary of the prompt used. Do not emit an \`<artifact>\`
 block for media.
 
-If Codex built-in imagegen is unavailable or fails, surface the actual failure
-message and ask the user for one-time confirmation before falling back to the
-existing OpenAI/Azure API-key provider path via \`"$OD_NODE_BIN" "$OD_BIN" media generate --surface image --model ${imageModel}\`.
+If Codex built-in imagegen is unavailable or generation fails before producing
+an image, surface the actual failure message and ask the user for one-time
+confirmation before falling back to the existing OpenAI/Azure API-key provider
+path via \`"$OD_NODE_BIN" "$OD_BIN" media generate --surface image --model ${imageModel}\`.
 Do not silently fall back.`;
 }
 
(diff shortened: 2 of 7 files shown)


lefarcen:
Hi @magicseek! 🎉
Thanks for the contribution — letting Codex projects use built-in imagegen is a solid improvement to the agent integration.

— open-design team

apps/daemon/src/prompts/system.ts
```diff
@@ -188,9 +190,62 @@ export function composeSystemPrompt({
     parts.push(MEDIA_GENERATION_CONTRACT);
   }
 
+  const codexImagegenOverride = renderCodexImagegenOverride(agentId, metadata);
+  if (codexImagegenOverride) {
+    parts.push(codexImagegenOverride);
```
mrcfps: This appends the Codex override to the daemon-composed prompt, but in the live chat path that prompt is not the final prompt the agent reads. `startChatRun` later builds `instructionPrompt` as `[daemonSystemPrompt, runtimeToolPrompt, systemPrompt]` (`apps/daemon/src/server.ts:3471`), and the `systemPrompt` supplied by the web app still comes from `@open-design/contracts` (`apps/web/src/components/ProjectView.tsx:599`) where `composeSystemPrompt` has no `agentId` support and still ends media projects with `MEDIA_GENERATION_CONTRACT`. For a Codex `gpt-image-*` image project, that later client prompt reintroduces the hard rule to "Always call out to the dispatcher" after this override, so the PR’s main path can still ask for the OpenAI/Azure API-key provider instead of trying built-in imagegen first. Please make the Codex override win in the actual final prompt too — for example, render/pass the same override from the contracts/web composer with `agentId`, or append this override after the client `systemPrompt` in `startChatRun` and add a test that composes the full live `instructionPrompt` ordering.

mrcfps (changes requested):
@magicseek thanks for the thoughtful Codex imagegen integration work here — the direction is useful and the daemon-side tests cover the new composer path well. I found one primary-flow issue that means the override can still be superseded in the live chat prompt, so I’m requesting changes to make the built-in path actually win for Codex gpt-image projects.


apps/daemon/src/prompts/system.ts
```diff
@@ -188,9 +190,62 @@ export function composeSystemPrompt({
...
+into \`$OD_PROJECT_DIR\` with a short descriptive filename. Never leave a
+project-referenced asset only under \`$CODEX_HOME\`.
+
+After copying the file into \`$OD_PROJECT_DIR\`, reply with the project-local
+filename and a short summary of the prompt used. Do not emit an \`<artifact>\`
+block for media.
+
+If Codex built-in imagegen is unavailable or fails, surface the actual failure
+message and ask the user for one-time confirmation before falling back to the
+existing OpenAI/Azure API-key provider path via \`"$OD_NODE_BIN" "$OD_BIN" media generate --surface image --model ${imageModel}\`.
```
lefarcen: **P2 Security / Correctness:** `imageModel` comes from persisted project metadata, which the daemon stores from request bodies without runtime model validation. `startsWith('gpt-image-')` still allows backticks, newlines, or shell metacharacters to be embedded into a load-bearing system prompt and into a fallback command the agent may copy. Please whitelist against the known `IMAGE_MODELS` ids or otherwise strictly validate/escape before rendering.

apps/daemon/src/prompts/system.ts
```diff
@@ -188,9 +190,62 @@ export function composeSystemPrompt({
...
+The active agent is Codex and this image project selected \`${imageModel}\`.
+For this specific case, use Codex's built-in image generation capability
+instead of \`"$OD_NODE_BIN" "$OD_BIN" media generate\` for the first generation
+attempt. This is an intentional exception to the media generation contract and
+the active image skill's dispatcher wording.
+
+Do not require, request, or mention \`OPENAI_API_KEY\` before trying the
+built-in path. Reuse the project metadata, reference prompt template, aspect
+ratio, style notes, and the user's current brief to form the final image
+prompt. Generate the image with Codex built-in imagegen, then move or copy the
```
lefarcen: **P2 Reasoning:** This assumes Codex imagegen always leaves a selectable `ig_*.png` under `${CODEX_HOME:-$HOME/.codex}/generated_images/...` and that the sandboxed agent can copy it into `$OD_PROJECT_DIR`. If the built-in tool returns a different path, no path, or the Codex workspace sandbox blocks reading from `$CODEX_HOME`, the instruction chain becomes ambiguous. Prefer instructing Codex to use the actual output path returned by the built-in tool, only search the known directory as a fallback, and report the exact copy/access error if no project-local copy can be made.

apps/daemon/src/prompts/system.ts
```diff
@@ -188,9 +190,62 @@ export function composeSystemPrompt({
...
+\`image_gen\` tool at
+\`\${CODEX_HOME:-$HOME/.codex}/generated_images/.../ig_*.png\`
+into \`$OD_PROJECT_DIR\` with a short descriptive filename. Never leave a
+project-referenced asset only under \`$CODEX_HOME\`.
+
+After copying the file into \`$OD_PROJECT_DIR\`, reply with the project-local
+filename and a short summary of the prompt used. Do not emit an \`<artifact>\`
+block for media.
+
+If Codex built-in imagegen is unavailable or fails, surface the actual failure
```
lefarcen: **P2 Failure handling:** The failure branch only covers imagegen being unavailable or failing, but not the separate case where generation succeeds and copying into `$OD_PROJECT_DIR` fails due to permissions, missing files, or filename conflicts. Falling back to the OpenAI/Azure provider in that case would be surprising and may generate a different image; the prompt should require surfacing the copy failure and not claiming success until the asset exists in the project directory.

lefarcen (changes requested):
Hi @magicseek! 👋

The implementation is small and mostly localized, but I found a few issues that need attention before merge:

The override currently interpolates unvalidated project metadata into high-priority prompt text and a shell-shaped fallback command. The prompt also assumes Codex imagegen file locations and copy behavior without clearly handling copy/sandbox failures, which can leave the agent stuck or confuse users.

See inline comments for specifics. Thanks for working on this integration!

magicseek:
Addressed the review feedback in 4a28129:\n\n- appends the Codex imagegen override after the client system prompt so it wins over the shared media contract, with duplicate earlier override text removed\n- restricts override rendering to registered OpenAI image model ids that start with gpt-image-, avoiding prompt interpolation from arbitrary metadata\n- passes a narrow Codex --add-dir allowance for \\/Users/troy.huang/.codex/generated_images only for whitelisted Codex image projects\n- updates the prompt guidance to prefer actual built-in imagegen output paths, use generated_images globbing only as fallback, verify copied project-local files, and report copy/access errors without silent provider fallback\n- adds focused tests for prompt ordering, whitelist behavior, generated_images path resolution, and Codex --add-dir emission\n\nVerified locally:\n- pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/system-prompt-template.test.ts tests/agents.test.ts tests/chat-route.test.ts\n- pnpm --filter @open-design/daemon typecheck\n- pnpm guard\n- pnpm typecheck

lefarcen (commented):
Thanks for the updates @magicseek! 👍

All three P2 concerns from my prior review are addressed in the new head:

- **Shell injection** — now uses a whitelist (`CODEX_IMAGEGEN_MODEL_IDS`) validated from `IMAGE_MODELS` registry before interpolation
- **File path assumptions** — now prefers the actual built-in imagegen output path, treats `generated_images` globbing only as fallback
- **Copy failure handling** — explicitly requires exact copy/access errors without silent fallback

apps/daemon/src/agents.ts
```diff
@@ -233,6 +234,12 @@ export const AGENT_DEFS = [
       if (runtimeContext.cwd) {
         args.push('-C', runtimeContext.cwd);
       }
+      const dirs = (extraAllowedDirs || []).filter(
+        (d) => typeof d === 'string' && d.length > 0,
+      );
+      for (const d of dirs) {
+        args.push('--add-dir', d);
```
mrcfps, line 241: This loops every `extraAllowedDirs` entry into Codex as `--add-dir`, but Codex documents that flag as adding directories that are writable alongside the primary workspace. In the live chat path, the changed `extraAllowedDirs` list is still built from `SKILLS_DIR`, `DESIGN_SYSTEMS_DIR`, and user `linkedDirs` before appending the Codex `generated_images` folder (`apps/daemon/src/server.ts:3610-3615`). That means any Codex run can now mutate skill/design-system resources or linked reference folders that the surrounding comments describe as read-only, which breaks the intended write barrier and can corrupt user/repo files outside the project. Please keep Codex `--add-dir` limited to directories that are safe to write, such as the generated-images output folder for the whitelisted imagegen path, or introduce separate read-only vs writable allowlists so linked reference/resource directories are not passed to Codex with writable access.

mrcfps (changes requested):
@magicseek thanks for the quick follow-up on the earlier prompt-ordering, model-validation, and copy-failure feedback — those pieces are much stronger now. I found one remaining sandbox-scope issue in the new Codex `--add-dir` wiring that can make directories documented as read-only writable to Codex, so I’m requesting changes on that narrow point.


magicseek:
Addressed the new Codex --add-dir sandbox-scope review in bc6333c.\n\nChanges:\n- split chat extra directory allowlisting by active agent\n- Codex now receives only the whitelisted generated_images output dir for valid gpt-image image projects\n- Codex no longer receives SKILLS_DIR, DESIGN_SYSTEMS_DIR, or linkedDirs through --add-dir\n- non-Codex agents retain existing skill/design/linked reference directory behavior and do not receive Codex generated_images\n- added helper coverage for Codex, non-Codex, and non-whitelisted Codex cases\n\nVerified locally:\n- git diff --check -- apps/daemon/src/server.ts apps/daemon/tests/chat-route.test.ts\n- pnpm --filter @open-design/daemon exec vitest run tests/chat-route.test.ts\n- pnpm --filter @open-design/daemon typecheck\n- pnpm guard\n- pnpm typecheck

apps/daemon/src/agents.ts
magicseek: Addressed it in https://github.com/nexu-io/open-design/commit/bc6333cc4d3aa3911c30d655de94ecc1444124d3

mrcfps (approved):
@magicseek I reviewed the updated Codex imagegen path at this head, including the final prompt ordering, model allowlist, generated_images access handling, and the narrowed Codex --add-dir behavior. The earlier issues are addressed in the live chat path and covered by the added daemon tests; thanks for the careful follow-through on the sandbox and fallback details. 🚀


apps/daemon/src/server.ts
```diff
@@ -3528,11 +3620,35 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
-      DESIGN_SYSTEMS_DIR,
-      ...linkedDirs,
-    ].filter((d) => fs.existsSync(d));
+    let codexGeneratedImagesDir = resolveCodexGeneratedImagesDir(
+      agentId,
+      projectRecord?.metadata,
+    );
+    if (codexGeneratedImagesDir) {
+      try {
+        fs.mkdirSync(codexGeneratedImagesDir, { recursive: true });
```
lefarcen: **P2 Symlink bypass:** `generated_images` is allowed after `mkdirSync` + `statSync`, but `statSync` follows symlinks. A pre-existing `${CODEX_HOME}/generated_images` symlink (or a `CODEX_HOME` symlink) can turn Codex `--add-dir` into a writable grant for `SKILLS_DIR`, `DESIGN_SYSTEMS_DIR`, or a linked dir — preserving the barrier for normal paths but leaving a bypass via path indirection.

Fix: canonicalize with `realpathSync.native`, reject final-component symlinks with `lstatSync`, and refuse to add the dir if its canonical path is inside any canonical resource or linked-dir root.

apps/daemon/src/server.ts
```diff
@@ -3461,14 +3543,21 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+        agentId,
         projectId,
         skillId,
         designSystemId,
       });
-    const instructionPrompt = [daemonSystemPrompt, runtimeToolPrompt, systemPrompt]
-      .map((part) => (typeof part === 'string' ? part.trim() : ''))
-      .filter(Boolean)
-      .join('\n\n---\n\n');
+    const codexImagegenOverride = renderCodexImagegenOverride(
```
lefarcen: **P3 Prompt/validation ordering:** The Codex imagegen override is rendered before the generated-images directory is validated (lines 3627-3643). If mkdir/stat fails, the override stays in the prompt but the dir is not passed to Codex — instructing Codex to copy from a dir the sandbox doesn't have access to, causing guaranteed copy failure instead of using the normal media dispatcher.

Fix: validate first and only append the override when the same validated directory will actually be granted, or fall back to the existing media contract when validation fails.

lefarcen (changes requested):
Hey @magicseek, thanks for the quick fix! The direct mrcfps concern (Codex receiving SKILLS_DIR/DESIGN_SYSTEMS_DIR/linkedDirs) is resolved — Codex now gets only the generated_images folder.

Found 2 edge-case issues in the new allowlist validation path — see inline below.

apps/daemon/src/server.ts
lefarcen: **P2** — symlink escape bypass: `statSync` follows symlinks, so a pre-existing `${CODEX_HOME}/generated_images` symlink (or `CODEX_HOME` itself being a symlink) can still turn Codex `--add-dir` into a writable grant for SKILLS_DIR / DESIGN_SYSTEMS_DIR / linkedDirs, bypassing the fix. Suggested fix: canonicalize with `fs.realpathSync.native(resolved)`, reject final-component symlinks with `lstatSync(resolved).isSymbolicLink()`, and refuse to add the dir if its canonical path is inside any canonical resource/linked-dir root.

apps/daemon/src/server.ts
lefarcen: **P3** — prompt/validation ordering: the Codex imagegen override is rendered before the generated-images directory is validated (lines 3627-3643). If validation fails (mkdir/stat error), Codex is still instructed to use built-in imagegen + copy from `$CODEX_HOME/generated_images`, but the sandbox won't have `--add-dir` for that directory → guaranteed copy failure instead of falling back to the normal media dispatcher. Fix by validating first and only appending the override when the validated directory will actually be passed to Codex.

lefarcen (changes requested):
Hey @magicseek — the sandbox-scope fix for the mrcfps concern (Codex no longer receives SKILLS_DIR/DESIGN_SYSTEMS_DIR/linkedDirs) is solid 👍. I found two remaining issues in the new generated-images allowlist path that could still break the write barrier. Both are fixable with path canonicalization + validation-first ordering.

magicseek:
Addressed the latest generated_images validation feedback in a1eb767.\n\nChanges:\n- validates the Codex generated_images dir before composing the final prompt\n- rejects final-component generated_images symlinks\n- canonicalizes validated paths with fs.realpathSync.native and passes the canonical path to Codex --add-dir\n- rejects generated_images grants whose nominal/canonical path resolves inside SKILLS_DIR, DESIGN_SYSTEMS_DIR, or linkedDirs\n- appends the Codex imagegen override only when the same validated directory is present in extraAllowedDirs; otherwise the prompt falls back to the normal media contract\n- added tests for symlink rejection, protected-root canonical rejection, canonical add-dir grants, and no override when validation/grant is absent\n\nVerified locally:\n- git diff --check -- apps/daemon/src/server.ts apps/daemon/tests/chat-route.test.ts\n- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts\n- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/agents.test.ts tests/chat-route.test.ts\n- pnpm --filter @open-design/daemon typecheck\n- pnpm guard\n- pnpm typecheck

lefarcen (approved):
All prior P2/P3 concerns are resolved in a1eb7670:

**P2 (symlink bypass) - FIXED:**
`validateCodexGeneratedImagesDir` now uses `lstatSync` to reject final-component symlinks, canonicalizes paths with `realpathSync.native`, and verifies the canonical path is not inside SKILLS_DIR / DESIGN_SYSTEMS_DIR / linkedDirs. This closes the symlink-based write-barrier bypass.

**P3 (prompt/validation ordering) - FIXED:**
Validation now runs BEFORE prompt composition. `resolveGrantedCodexImagegenOverride` only renders the override when the validated `codexGeneratedImagesDir` is actually present in `extraAllowedDirs`, ensuring Codex is never instructed to use a path it cannot access.

No new issues found. The validation logic is well-tested and comprehensive. Ready to merge from the bot's perspective. 🚀

lefarcen:
Hey @magicseek! 🎉

Thanks for seeing this through — the validation logic for generated_images is solid, and the test coverage makes the Codex imagegen path maintainable. Merged! 🚀

leilei926524-tech:
Hi @magicseek — welcome to Open Design contributors. 🎉

Your first merged PR (#622) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #623, opened 2026-05-06
Fix/error message persistence

nhancdt2602:
Close #165

## Root Cause
Daemon/API failures were only surfaced via ephemeral chat error state, so switching agents could hide the prior run’s details

## Fix

The fix appends a durable status event with the error text onto the assistant ChatMessage (and uses a small `appendErrorStatusEvent` helper). I also added e2e test to cover this issue

### DIFF
--- a/apps/web/src/runtime/chat-events.ts
+++ b/apps/web/src/runtime/chat-events.ts
@@ -7,6 +7,9 @@ export function appendErrorStatusEvent(message: ChatMessage, detail: string): Ch
   if (last?.kind === 'status' && last.label === 'error' && last.detail === detail) {
     return message;
   }
+  if (!detail?.trim()) {
+    return message;
+  }
   return {
     ...message,
     events: [...events, { kind: 'status', label: 'error', detail }],
--- a/apps/web/src/runtime/chat-events.ts
+++ b/apps/web/src/runtime/chat-events.ts
@@ -1,0 +1,15 @@
+import type { ChatMessage } from '../types';
+
+export function appendErrorStatusEvent(message: ChatMessage, detail: string): ChatMessage {
+  if (!detail) return message;
+  const events = message.events ?? [];
+  const last = events[events.length - 1];
+  if (last?.kind === 'status' && last.label === 'error' && last.detail === detail) {
+    return message;
+  }
+  return {
+    ...message,
+    events: [...events, { kind: 'status', label: 'error', detail }],
+  };
+}
+
--- a/e2e/tests/project-view-error-persistence.test.tsx
+++ b/e2e/tests/project-view-error-persistence.test.tsx
@@ -1,0 +1,165 @@
+import { cleanup, render, waitFor } from '@testing-library/react';
+import React, { act } from 'react';
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+let lastChatPaneProps: any | null = null;
+
+(globalThis as any).IS_REACT_ACT_ENVIRONMENT = true;
+
+vi.mock('../../apps/web/src/i18n', () => ({
+  useT: () => ((key: string) => key) as any,
+}));
+
+vi.mock('../../apps/web/src/components/ChatPane', async () => {
+  const mod = await vi.importActual<any>('../../apps/web/src/components/ChatPane');
+  return {
+    ...mod,
+    ChatPane: (props: any) => {
+      lastChatPaneProps = props;
+      return null;
+    },
+  };
+});
+
+vi.mock('../../apps/web/src/components/AppChromeHeader', () => ({
+  AppChromeHeader: () => null,
+}));
+
+vi.mock('../../apps/web/src/components/AvatarMenu', () => ({
+  AvatarMenu: () => null,
+}));
+
+vi.mock('../../apps/web/src/components/FileWorkspace', () => ({
+  FileWorkspace: () => null,
+}));
+
+vi.mock('../../apps/web/src/router', () => ({
+  navigate: () => {},
+}));
+
+vi.mock('../../apps/web/src/utils/notifications', async (importOriginal) => {
+  const actual =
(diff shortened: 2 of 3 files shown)


lefarcen:
Hi @nhancdt2602! 🎉
Thanks for the contribution — this tackles a real traceability gap when switching agents.

— open-design team

e2e/tests/project-view-error-persistence.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+      />,
+    );
+
+    await waitFor(() => {
+      expect(lastChatPaneProps).not.toBeNull();
+      expect(lastChatPaneProps!.messages).toHaveLength(0);
+    });
+
+    await act(async () => {
+      await lastChatPaneProps!.onSend('hi', []);
```
mrcfps, line 155: This sends before the mounted `ProjectView` has necessarily finished loading `listConversations` and setting `activeConversationId`. The preceding wait only checks that `ChatPane` rendered with an empty `messages` array, which is also true on the initial render while `activeConversationId` is still `null`; in that state `handleSend` returns immediately at `if (!activeConversationId) return`, so the test can no-op and fail/flap instead of exercising the daemon error path. Please wait for `lastChatPaneProps.activeConversationId` to become `'conv-1'` before calling `onSend` (and ideally spy on `saveMessage` so the test verifies the error event is persisted, not just present in in-memory props).

mrcfps (changes requested):
@nhancdt2602 thanks for the focused fix here — persisting daemon error details on the assistant message is the right direction, and the helper keeps the event mutation nicely contained. I found one issue in the new coverage that should be fixed before merge so the test reliably exercises the intended path.


apps/web/src/runtime/chat-events.ts
```diff
@@ -0,0 +1,15 @@
+import type { ChatMessage } from '../types';
+
+export function appendErrorStatusEvent(message: ChatMessage, detail: string): ChatMessage {
+  if (!detail) return message;
```
lefarcen, line 4: **P2: Edge case — falsy error messages**

The guard `if (!detail) return message;` correctly handles empty strings, but upstream callers don't validate `err.message` before passing it here.

If `err.message` is an empty string (rare but possible), the error state is set but no status event is added, creating UI inconsistency.

Suggestion: Either:
1. Add validation at call sites: `if (err.message) appendErrorStatusEvent(...)`, OR
2. Change this guard to `if (!detail?.trim()) return message;` to be more defensive

Not critical since most errors have messages, but worth considering for robustness.

e2e/tests/project-view-error-persistence.test.tsx
```diff
@@ -0,0 +1,165 @@
...
+    createdAt: 0,
+    updatedAt: 0,
+    skillId: null,
+    designSystemId: null,
+    metadata: {} as any,
+  };
+}
+
+describe('ProjectView daemon error persistence', () => {
+  afterEach(() => cleanup());
```
lefarcen: **P2: Test doesn't verify agent-switching scenario**

Issue #165 specifically describes: *"switching to another agent causes the previous agent's error details to disappear"*.

This test validates that error events are added to the message, but doesn't test that they **persist when switching agents** (the core user journey).

Suggestion: Consider adding a second test case:
1. Send with Claude → daemon fails → verify error event
2. Switch to OpenCode → send again
3. Verify the Claude message still has its error event

This would validate the full fix scope. Not blocking since the current test proves the mechanism works, but the end-to-end behavior is what users care about.

lefarcen (commented):
Hi @nhancdt2602! 👋 Thanks for tackling this — the `appendErrorStatusEvent` approach is exactly right.

**Overall:** This fix properly addresses the root cause (ephemeral error state) by appending durable status events to the assistant message. The integration is clean, type-safe, and follows existing patterns. Two suggestions below to strengthen completeness, but neither blocks merging.

**Strengths:**
- ✅ Correctly identifies all daemon error catch blocks and adds persistence
- ✅ Deduplication logic prevents duplicate error events
- ✅ `useCallback` memoization is appropriate
- ✅ E2E test validates the core behavior

**Suggestions (P2 — non-blocking):**

See inline comments.

e2e/tests/project-view-error-persistence.test.tsx
nhancdt2602: Added agent switch scenario in the test

apps/web/src/runtime/chat-events.ts
nhancdt2602: Applied

e2e/tests/project-view-error-persistence.test.tsx
nhancdt2602: Applied waiting for `lastChatPaneProps.activeConversationId` to be set

lefarcen (commented):
Great iteration @nhancdt2602! 🎉 You've addressed both reviews comprehensively:

**mrcfps's blocking concerns:**
✅ Test race condition fixed — now waits for `activeConversationId` before `onSend`
✅ Persistence validation added — `saveMessageSpy` verifies the error event is durably saved

**Bot P2 suggestions:**
✅ Edge case handling — added `trim()` guard for empty/whitespace-only error messages
✅ Agent-switching scenario — added a second e2e test that validates the core user journey from issue #165 (error details persist when switching agents)

The second test case is particularly valuable — it exercises the exact behavior the issue reporter experienced: switching from Claude → OpenCode and verifying the prior error details remain visible. This gives high confidence the fix works end-to-end.

**Ready from my perspective.** Deferring final approval to a maintainer.

mrcfps (approved):

leilei926524-tech:
Hi @nhancdt2602 — welcome to Open Design contributors. 🎉

Your first merged PR (#623) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #624, opened 2026-05-06
feat: import existing local folder as project (#597)

### LINKED ISSUE #597: feat: create project from existing local folder
## Summary

Allow users to create an Open Design project rooted in an **existing local folder** instead of the default `.od/projects/<hash-id>/` directory. The chosen folder becomes the project's base directory, so all generated files, assets, and agent output live alongside the user's existing content.

## Current state

- Projects are always created under `.od/projects/<id>/` with a randomly generated ID
- `linkedDirs` in project metadata (PR #455) lets an agent **read** external folders as reference, but files are still written to `.od/projects/<id>/`
- There is no way to tell Open Design "this folder is my project — work inside it"
- Agent-generated media (logos, images, etc.) has no natural place to land — users must manually move files from `.od/` to their actual project folder

## Why this matters

Many real-world workflows start from an existing folder:
- A design system repo where the user wants OD to generate components
- A marketing site where the user needs new hero images
- A game asset folder where the user wants AI-generated sprites

Forcing everything into `.od/projects/<id>/` breaks these workflows — the user must constantly copy files back and forth.

### Creation flow
1. User clicks "New Project" → sees a new option: **"From existing folder"**
2. A native folder picker opens (reuse `/api/dialog/open-folder`)
3. The selected folder becomes the project's `baseDir`
4. OD writes a `.od-manifest.json` (or similar marker) into the folder to identify it as an OD project
5. All file operations (create, edit, save) happen **inside** this folder

### File layout
```
my-existing-project/           ← user's folder (now also project baseDir)
├── src/
├── assets/
├── images/
├── .od-manifest.json          ← OD project marker (id, skillId, etc.)
├── hero.html                  ← OD-generated file
└── logo.png                   ← OD-generated media asset
```

### Daemon changes
- `POST /api/projects` accepts an optional `baseDir` field
- `projectDir()` resolves to `baseDir` when set, falling back to `.od/projects/<id>/`
- File safety checks (`resolveSafe`) still apply, scoped to the new `baseDir`
- Project marker file is excluded from file listings / watch events

## Open questions

1. **Permission model** — Should OD ask for explicit permission before writing to a user folder? Options:
   - Per-project grant on creation (one-time)
   - Per-session grant with a duration
   - Trust-on-first-use with a revocation setting
   
2. **Safety boundaries** — What restrictions should apply?
   - Should OD refuse folders containing `.git` (risk of modifying tracked files)?
   - Should certain file extensions be off-limits?
   - Should there be a max folder size check?
   
3. **Marker file** — Is `.od-manifest.json` the right approach, or should OD store the mapping in its SQLite DB only? A marker file makes the project portable but clutters the user's directory.

4. **Conflict with linkedDirs** — `linkedDirs` already lets projects reference external folders. Should folder-based projects deprecate `linkedDirs`, or should both coexist? (e.g., folder = base, linkedDirs = additional read-only references)

5. **Existing project detection** — If a user selects a folder that already contains an OD marker, should OD offer to re-open the existing project instead of creating a new one?

6. **Media output path** — Should agent-generated media go to `<baseDir>/media/` by default, or should the user configure the output subfolder?

## Relation

- Complements `linkedDirs` from #455 (read-only reference code)
- Addresses the media output gap — generated assets land where the user expects them
- Related to #498 (import menu switching) — a folder-based project could offer in-place skill switching

infinity-nft:
## Summary

Implements the RFC in #597 — create a project rooted at an existing local folder. OD reads / writes directly inside the user's folder (no copy, no shadow tree), aligning with how Cursor / Claude Code / Aider behave. Single-mode design landed after [@encyc's review feedback](https://github.com/nexu-io/open-design/issues/597) replaced the original git-detect / two-mode prototype.

Resolves #597

## What's in this PR

7 commits, ~370/-25 lines across 12 files, split into reviewable units:

| | |
|---|---|
| `feat(contracts)` | `ImportFolderRequest` / `ImportFolderResponse` + `metadata.baseDir` field |
| `feat(daemon)` | `resolveProjectDir`, metadata-aware `listFiles` / `readProjectFile` / `writeProjectFile` / `ensureProject`, `detectEntryFile`, `SKIP_DIRS` for the file panel |
| `feat(daemon)` | `POST /api/import/folder` endpoint with `realpath()` canonicalization |
| `feat(desktop)` | Electron preload exposing `window.electronAPI.pickFolder` (uses `removeHandler` + `handle` so dev hot-reload doesn't throw on second registration) |
| `feat(web)` | "Open folder" affordance in New Project (native picker on Electron, absolute-path text input in browser) + App.tsx wiring |
| `docs(architecture)` | Endpoint documented in the daemon API table + a "Folder import" section |
| `test(daemon)` | 13 unit + 10 integration tests, including symlink security cases |

## Safety

- `baseDir` is canonicalized via `fs.promises.realpath()` at import time and stored as the canonical path. Without this, a user-controlled symlink (e.g. `~/sneaky → /etc`) would let `writeProjectFile` escape the project sandbox at every later call — `resolveSafe` validates the *literal* baseDir, but the OS follows symlinks at `open()` time.
- Standard `resolveSafe` / `sanitizePath` checks apply on every write — `metadata.baseDir` only changes the project root, not the bounds check.
- `RUNTIME_DATA_DIR` is `realpathSync`'d at module init, so the post-realpath reentry check works correctly on macOS where `/var ≡ /private/var` (the symlink-into-data-dir test caught this — would otherwise have been a silent bypass).
- Imports inside `RUNTIME_DATA_DIR` (or any path that resolves there) are refused.
- The file panel hides conventional build / install dirs (`node_modules .git dist build .next .nuxt .turbo .cache .output out coverage __pycache__ .venv vendor target .od .tmp`) when listing baseDir-rooted projects.

## What's deliberately deferred

Per the [issue thread](https://github.com/nexu-io/open-design/issues/597), OD-level write protection (`.od-ignore` file, once-per-folder trust toggle, etc.) is left for a follow-up — happy to land it inline if @lefarcen prefers, just say which shape.

## Test plan

- [x] `pnpm guard` — clean
- [x] `pnpm typecheck` — clean
- [x] `pnpm --filter @open-design/contracts test` — 6/6 pass
- [x] `pnpm --filter @open-design/daemon test` — **790/790 pass** (incl. 23 new tests for folder import)
- [x] `pnpm --filter @open-design/web test` — 258/258 pass
- [x] `pnpm --filter @open-design/web build` — production SSG build clean
- [ ] Manual: import a non-git folder — file panel shows files, edits land in the original folder
- [ ] Manual: import a folder containing `node_modules` / `dist` — those are hidden from the file panel
- [ ] Manual: try `ln -s /etc /tmp/sneaky && /api/import/folder { baseDir: "/tmp/sneaky" }` — `metadata.baseDir` stored as `/etc`, `RUNTIME_DATA_DIR` reentry check still works
- [ ] Manual: try importing the daemon's own `.od/` data dir — refused
- [ ] Manual: browser session (no Electron picker) — text-input fallback accepts absolute paths


lefarcen (changes requested):
Hey @infinity-nft — thanks for the quick single-mode pivot after @encyc's feedback! 🎉 The RFC-to-implementation cycle is impressive.

I ran a deep review across the codebase (not just the diff) and found **security + correctness gaps** that need fixing before merge. The core `baseDir` plumbing is solid, but several existing routes don't thread `metadata` through, and the symlink hardening stops at the root (descendants can still escape).

## Critical (P1) — must fix

See inline comments (posting separately) for file:line specifics. Summary:

1. **Generic create/patch routes bypass import validation** — `POST /api/projects` and `PATCH /api/projects/:id` accept arbitrary `metadata.baseDir` without `realpath()` or `RUNTIME_DATA_DIR` checks, so a client can inject `/etc` or data-dir paths and bypass the import route's guards.
2. **Symlink descendants inside imported folders escape** — `realpath(baseDir)` only hardens the root; `resolveSafe()` still follows symlink components inside the folder (e.g. `assets -> /Users/me/.ssh`), so `/api/projects/:id/raw/assets/id_rsa` passes the prefix check and reads arbitrary files. Because raw responses allow `Origin: null`, sandboxed artifact HTML can exfiltrate those bytes.
3. **Chat uploads land outside imported folders** — composer upload still calls `ensureProject(PROJECTS_DIR, id)` without metadata, so attachments write to `.od/projects/<id>` instead of `baseDir`. The agent later checks attachments under `cwd` (the imported folder) and can't read them → broken feature.
4. **Deploy ignores imported folders** — deploy planning and preflight don't fetch project metadata, so imported-folder projects can't deploy files shown in the UI.

## Should-fix (P2)

5. **Project search not metadata-aware** — search route doesn't fetch the project, so imported-folder search hits the wrong tree.
6. **ZIP export not metadata-aware** — both full and batch archive builders call `projectDir()` directly, so "Download ZIP" exports the wrong folder or fails.
7. **File watchers not metadata-aware** — imported folders won't live-reload when files change.
8. **Test coverage weaker than claimed** — data-dir symlink test silently skips unless `OD_DATA_DIR` is set, and there's no test for descendant symlinks escaping through read/write/raw/archive routes.

## Docs (P2)

9. **architecture.md missing rationale** — folder-import section documents the chosen single-mode behavior but not WHY (no comparison of copy/shadow-tree, read-only/trust-toggle, or two-mode git-detect alternatives). Cursor/Claude Code/Aider precedent is name-dropped without explaining what OD is copying or where it differs.

---

Inline comments with concrete fixes coming in separate comments below. Let me know if any need clarification — happy to pair on the symlink hardening logic if that helps!

lefarcen:
**P1 findings with concrete fixes:**

1️⃣ **apps/daemon/src/server.ts:1363** — Generic create route bypasses import validation

`POST /api/projects` accepts arbitrary `metadata.baseDir` without the `realpath()` or `RUNTIME_DATA_DIR` reentry checks that `/api/import/folder` applies. A client can inject `/etc` or data-dir paths here and bypass the guards.

**Fix:** Strip `baseDir` from client-supplied metadata in generic create/patch:
```ts
if (typeof metadata?.baseDir === 'string') {
  return sendApiError(res, 400, 'BAD_REQUEST', 'baseDir can only be set via /api/import/folder');
}
```

2️⃣ **apps/daemon/src/projects.ts:331** — Symlink descendants inside imported folders escape

`realpath(baseDir)` only hardens the root; `resolveSafe()` still follows symlink components inside the folder. If an imported repo contains `assets -> /Users/me/.ssh`, `/api/projects/:id/raw/assets/id_rsa` passes the string prefix check and `readFile()` follows the symlink. Because raw responses allow `Origin: null`, sandboxed artifact HTML can exfiltrate those bytes.

**Fix:** For read/write/delete/archive/raw paths, lstat each path segment and reject symlinks, or compare the final `realpath()` against the canonical project root:
```ts
export async function resolveSafeNoSymlinks(root: string, name: string): Promise<string> {
  const candidate = resolveSafe(root, name);
  const real = await fs.promises.realpath(candidate);
  if (!real.startsWith(root + path.sep) && real !== root) {
    throw new Error('symlink escape');
  }
  return real;
}
```
Then use `resolveSafeNoSymlinks` in `readProjectFile`, `writeProjectFile`, `deleteProjectFile`, and raw route handlers.

3️⃣ **apps/daemon/src/server.ts:883** — Chat uploads land outside imported folders

Composer upload calls `ensureProject(PROJECTS_DIR, req.params.id)` without project metadata, so attachments land in `.od/projects/<id>` instead of `baseDir`. The agent later checks attachments under `cwd` (the imported folder) and can't read them → broken feature.

**Fix:**
```ts
const uploadProject = getProject(db, req.params.id);
const attachmentDir = await ensureProject(PROJECTS_DIR, req.params.id, uploadProject?.metadata);
```

4️⃣ **apps/daemon/src/server.ts:2628** — Deploy ignores imported folders

Deploy planning and preflight don't fetch project metadata, so `buildDeployFileSet()` and `prepareDeployPreflight()` call `readProjectFile()` without metadata. Imported-folder projects can't deploy files shown in the UI.

**Fix:**
```ts
const project = getProject(db, req.params.id);
const fileSet = await buildDeployFileSet(PROJECTS_DIR, req.params.id, opts, project?.metadata);
```

infinity-nft:
Both Codex findings were correct — fixed in 1ca8070:

1. **chat-route listFiles wrapper** — was passing `chatMeta` directly as the `opts` arg. `listFiles` reads `opts.metadata`, so for baseDir-rooted projects the call silently scanned `.od/projects/<id>/` and `existingProjectFiles` came back empty. Now wraps as `{ metadata: chatMeta }`.

2. **Deploy / Preflight not threading metadata** — `deploy.ts` called `readProjectFile` without the metadata 3rd arg, so for baseDir-rooted projects the deploy and preflight endpoints would look in `.od/projects/<id>/` and 404. Threaded `options.metadata` through `buildDeployFilePlan` → `readProjectFile`, and the two server callsites now pass `project?.metadata`.

Added a regression test that locks the `listFiles` contract: passing a bare metadata object as `opts` (the bug pattern) must fall back to the standard project dir, not scan `baseDir` — so the same mistake fails loudly in tests next time.

Validation:
- `pnpm guard` clean
- `pnpm typecheck` clean
- `pnpm --filter @open-design/daemon test` — 791/791 (24 folder-import tests)

lefarcen:
**P2 findings with concrete fixes:**

5️⃣ **apps/daemon/src/server.ts:2783** — Project search not metadata-aware

The search route doesn't fetch the project, and `searchProjectFiles()` calls `listFiles()` without metadata, so imported-folder projects search the wrong tree.

**Fix:**
```ts
const project = getProject(db, req.params.id);
const results = await searchProjectFiles(PROJECTS_DIR, req.params.id, query, { metadata: project?.metadata });
```
Then update `searchProjectFiles` signature to accept and thread `opts.metadata` through `listFiles()` and `resolveProjectDir()`.

6️⃣ **apps/daemon/src/server.ts:2801** — ZIP archive export not metadata-aware

Both full and batch archive builders call `projectDir()` directly and have no metadata parameter, so "Download ZIP" exports the wrong folder or fails.

**Fix:**
```ts
const project = getProject(db, req.params.id);
const archive = await buildArchive(PROJECTS_DIR, req.params.id, project?.metadata);
```
Update `buildArchive` and `buildBatchArchive` to accept `metadata` and use `resolveProjectDir()`.

7️⃣ **apps/daemon/src/server.ts:1660** — File watchers not metadata-aware

Imported-folder projects won't live-reload when files change because `subscribe()` always watches `projectDir(projectsRoot, projectId)`.

**Fix:**
```ts
const project = getProject(db, req.params.id);
const watchDir = resolveProjectDir(PROJECTS_DIR, req.params.id, project?.metadata);
await subscribe(watchDir, (e) => ...);
```
Key watchers by canonical directory, not project ID.

8️⃣ **apps/daemon/tests/folder-import-route.test.ts:146** — Test coverage weaker than claimed

The data-dir symlink test silently skips unless `OD_DATA_DIR` is set, and there's no test for descendant symlinks escaping through `readProjectFile()`, `writeProjectFile()`, raw routes, or archives.

**Fix:** Make the daemon test data dir explicit for this suite and add read/write/raw tests:
```ts
it('rejects reads through descendant symlinks', async () => {
  const realFolder = makeFolder();
  await mkdir(path.join(realFolder, 'assets'));
  symlinkSync('/etc/passwd', path.join(realFolder, 'assets', 'passwd'));
  const resp = await importFolder({ baseDir: realFolder });
  const projectId = (await resp.json()).project.id;
  const raw = await fetch(`${baseUrl}/api/projects/${projectId}/raw/assets/passwd`);
  expect(raw.status).toBe(400); // or 403
});
```

9️⃣ **docs/architecture.md:273** — Missing rationale for single-mode choice

The folder-import section documents the chosen single-mode behavior but not WHY. No comparison of copy/shadow-tree, read-only/trust-toggle, or two-mode git-detect alternatives. Cursor/Claude Code/Aider precedent is name-dropped without explaining what OD is copying or where it differs.

**Fix:** Add a rationale subsection explaining alternatives considered and prior art (see review body for full suggested text).

apps/daemon/src/server.ts
```diff
@@ -1494,6 +1504,94 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+
+      const project = insertProject(db, {
+        id,
+        name: projectName,
+        skillId: skillId ?? null,
+        designSystemId: designSystemId ?? null,
+        pendingPrompt: null,
+        metadata: {
+          kind: 'prototype',
+          baseDir: normalizedPath,
```
mrcfps, line 1958: This route correctly canonicalizes and rejects unsafe folder imports before storing `metadata.baseDir`, but `baseDir` is now also part of the public project metadata shape and the generic `POST /api/projects` / `PATCH /api/projects/:id` paths still persist client-supplied metadata directly. That lets any local client create or patch a project with `metadata.baseDir: "/etc"` (or a path inside `RUNTIME_DATA_DIR`) and bypass the realpath and data-dir guard above, after which the changed `readProjectFile`/`writeProjectFile` call sites operate on that arbitrary root. Please reject or strip `metadata.baseDir` in the generic create/patch handlers and keep `/api/import/folder` as the only code path that can set it, ideally with regression tests for both create and patch.

apps/daemon/src/projects.ts
```diff
@@ -281,8 +328,8 @@ async function collectArchiveEntries(dir, relDir, out) {
   }
 }
 
-export async function readProjectFile(projectsRoot, projectId, name) {
-  const dir = projectDir(projectsRoot, projectId);
+export async function readProjectFile(projectsRoot, projectId, name, metadata?) {
+  const dir = resolveProjectDir(projectsRoot, projectId, metadata);
```
mrcfps, line 338: Passing `metadata.baseDir` into `readProjectFile` makes imported folders readable, but this still relies on `resolveSafe`'s string-prefix check before Node opens the file. A symlink inside the imported folder, such as `assets -> /Users/me/.ssh`, produces a path that starts with `baseDir` and therefore passes `resolveSafe`, then `readFile()` follows the symlink and `/api/projects/:id/raw/assets/id_rsa` can expose data outside the imported project. Please harden the metadata-backed file operations by resolving the final path with `realpath()` and verifying it stays under the canonical project root, or by `lstat`-walking each segment and rejecting symlinks, and add route/project tests for read/write/raw/archive symlink descendants.

apps/daemon/src/server.ts
```diff
@@ -2876,7 +2986,8 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     },
```
mrcfps, line 3460: This updates the single-file `/api/projects/:id/files` upload path to pass project metadata, but the chat composer uses the separate `/api/projects/:id/upload` middleware above. That multer destination still calls `ensureProject(PROJECTS_DIR, req.params.id)` before the request reaches the handler, so imported-folder attachments are written to `.od/projects/<id>` while the chat run later sets `cwd` to `metadata.baseDir` and filters attachments by existence there. The result is that pasted/dropped files disappear from the agent's view in imported projects. Please make the `projectUpload` destination metadata-aware (or store to a temp dir and then move through `writeProjectFile(..., project.metadata)`) and cover the composer upload route in the folder-import tests.

mrcfps (changes requested):
@infinity-nft thanks for pushing this folder-import flow forward and for addressing the first metadata-threading fixes quickly. I found a couple of remaining correctness/security gaps that still affect the imported-folder primary path before merge. 🚧


lefarcen:
Thanks for the quick fixes! 🎉

**Progress check on commit 1ca8070:**
- ✅ **P1 #4 (Deploy) — FIXED** — `buildDeployFileSet` and `prepareDeployPreflight` now thread `project.metadata` through correctly
- 🟡 **P1 #3 (Chat uploads) — PARTIAL** — The chat `listFiles` wrapper is fixed, but the composer upload path (`server.ts:883-887`) still writes to `.od/projects/<id>` instead of `baseDir`. The `projectUpload` multipart config needs to call `ensureProject` with `project.metadata`.
- ❌ **P1 #1 (Generic create/patch bypass) — UNFIXED** — `server.ts:1383` and `server.ts:1607-1614` still accept arbitrary `metadata.baseDir` without the `realpath()` + `RUNTIME_DATA_DIR` checks
- ❌ **P1 #2 (Symlink descendant escape) — UNFIXED** — `readProjectFile` / `writeProjectFile` / `deleteProjectFile` still follow symlink components inside imported folders (need `lstat` + `realpath` check as described in the original review)

**3 P1 blockers remain.** Once those are addressed, the core security + correctness foundation will be solid and we can move to the P2 improvements (search, ZIP export, watchers, test coverage, docs).

Let me know if you'd like me to clarify any of the remaining fixes!

infinity-nft:
@mrcfps thanks for the careful re-review — all three remaining P1s are addressed in 63bf78f (you were reading 1ca8070 which only had the deploy + listFiles wrapper fixes; the security pass landed in the next push and just made it to GitHub now).

## P1 #1 — Generic create/patch bypass

**Fixed.** `POST /api/projects` and `PATCH /api/projects/:id` both reject a `metadata.baseDir` payload with 400. The error message points the client at `/api/import/folder` as the only path allowed to set it, where the realpath() + RUNTIME_DATA_DIR reentry checks live. Added a regression test (`folder-import-route.test.ts`) that asserts `POST /api/projects` with `metadata.baseDir: '/etc'` returns 400 with a message mentioning the import endpoint.

## P1 #2 — Symlink descendant escape

**Fixed.** New `resolveSafeReal()` helper in `projects.ts` realpath()s the resolved candidate (or, for write paths that haven't created the file yet, the longest existing prefix) and re-validates against `realpath(projectRoot)`. `readProjectFile`, `writeProjectFile`, and `deleteProjectFile` now go through it. Symlink components inside an imported folder no longer pass the check: even though the literal path stays under baseDir, the canonical realpath lands outside, so the helper throws `EPATHESCAPE`.

Regression test: import a folder containing `assets/leak.txt → /etc/hosts`, hit `/api/projects/:id/raw/assets/leak.txt`, expect 400.

## P1 #3 — Composer upload destination

**Fixed.** `projectUpload`'s multer destination now resolves through a module-level `projectMetadataLookup` that `startServer()` populates from the daemon's db. For `metadata.baseDir`-rooted projects, attachments now land alongside the user's files (where the agent's cwd actually points), instead of `.od/projects/<id>/`.

The bridge is needed because multer's storage is built at module init, before the per-process db exists. The lookup is a `(id) => metadata | null` callback that startServer wires up next to the existing `configureConnectorCredentialStore` / `configureComposioConfigStore` calls.

## P2 batch (also in 63bf78f)

- **Search** — `searchProjectFiles` accepts `opts.metadata`, route fetches the project and threads it through.
- **ZIP archives** — `buildProjectArchive` and `buildBatchArchive` accept metadata; both archive routes pass `project?.metadata`.
- **Watchers** — `subscribeFileEvents` resolves to baseDir for imported projects so live-reload SSE actually fires when the user edits files in their own folder. Registry stays keyed by the canonical directory, so two projects pointing at the same folder share one watcher.
- **Template snapshotting** — reading source-project files when saving a template now uses metadata, so a template can be saved from a baseDir-rooted source.

## Validation

- `pnpm guard` — clean
- `pnpm typecheck` — clean
- `pnpm --filter @open-design/daemon test` — **793/793 pass** (26 folder-import tests, +2 from this round)

The P2 docs rationale you mentioned is also queued — happy to add it once you've signed off on the security pass.

apps/daemon/src/projects.ts
```diff
@@ -83,8 +130,8 @@ async function collectFiles(dir, relDir, out) {
 // the user sees in the file panel. Used by the "Download as .zip" share
 // menu item, which exports the user's actual project tree (e.g. the
 // uploaded `ui-design/` folder), not just the rendered HTML.
-export async function buildProjectArchive(projectsRoot, projectId, root) {
-  const projectRoot = projectDir(projectsRoot, projectId);
+export async function buildProjectArchive(projectsRoot, projectId, root, metadata?) {
+  const projectRoot = resolveProjectDir(projectsRoot, projectId, metadata);
```
mrcfps, line 134: This change makes the full archive builder operate on `metadata.baseDir`, but the optional `root` path below is still scoped with the string-only `resolveSafe()` and then `stat()` / `collectArchiveEntries()` follow directory symlinks. In an imported folder containing `docs -> /Users/me/.ssh`, `GET /api/projects/:id/archive?root=docs` passes the literal prefix check and zips files outside the imported project, bypassing the new `resolveSafeReal()` protection used by raw reads. Please resolve archive roots with the same symlink-aware policy (or lstat-walk and reject symlink segments) before walking them, and add a route regression test for an archive `root` that is a symlink to an outside directory.

apps/daemon/src/server.ts
```diff
@@ -1506,6 +1639,17 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
   app.patch('/api/projects/:id', (req, res) => {
     try {
       const patch = req.body || {};
+      // Same rule as POST /api/projects: baseDir is privileged and can
+      // only be mutated through the import endpoint, not through a
+      // client-supplied metadata patch.
+      if (patch.metadata && typeof patch.metadata === 'object') {
+        if ('baseDir' in patch.metadata) {
```
mrcfps: Rejecting `patch.metadata.baseDir` only blocks explicit baseDir changes, but `updateProject()` replaces `metadata` with the incoming object. A normal metadata patch on an imported project, for example `{ metadata: { kind: 'prototype', linkedDirs: [...] } }`, passes this guard because it omits `baseDir` and then drops the stored folder root entirely. Subsequent reads/writes/watch/deploy/chat calls fall back to `.od/projects/<id>` instead of the user’s imported folder, breaking the primary folder-import flow and potentially writing new files to the wrong empty project directory. Please preserve the existing immutable `baseDir` / folder-import fields when applying metadata patches, or reject metadata patches on imported projects unless they retain the unchanged baseDir; add a PATCH regression test that omits `baseDir` on an imported project.

mrcfps (changes requested):
@infinity-nft thanks for the quick security pass and for wiring the remaining folder-import paths through metadata. I found two remaining blocking issues in the updated implementation: one archive symlink escape and one metadata patch regression that can detach imported projects from their folder. 🚧


infinity-nft:
@mrcfps both round-2 P1s addressed in f9347e5.

## #1 Archive root symlink escape

Switched the optional `?root=<subdir>` resolution in `buildProjectArchive` from string-only `resolveSafe()` to the symlink-aware `resolveSafeReal()` — the same helper already covering raw read/write/delete paths. `collectArchiveEntries()` itself already skips dirent symlinks via `!isDirectory && !isFile`, so canonicalizing the archive root was the only missing piece. A directory symlink like `docs → /Users/me/.ssh` now resolves outside the project root and `resolveSafeReal()` throws `EPATHESCAPE` before the walker ever runs.

Regression test: `?root=<symlink-to-/etc>` → 400.

## #2 PATCH metadata wiped baseDir

You're right — the previous guard only caught explicit `baseDir` mutation. The replace-not-merge semantics of `updateProject()` meant any patch that *omits* `baseDir` (e.g. a UI editing `linkedDirs` only sends `{ metadata: { kind, linkedDirs } }`) would detach imported projects from their folder.

Fixed by re-stamping the immutable folder-import fields (`baseDir`, `importedFrom='folder'`) from the existing project record onto the incoming patch when the project is imported. Three cases now handled:

- Patch on imported project that **supplies the same baseDir** → accepted (no-op for that field).
- Patch on imported project that **supplies a different baseDir** → 400 "baseDir is immutable".
- Patch on imported project that **omits baseDir** → re-stamped from the existing record, never wiped.
- Patch on non-imported project that **adds baseDir** → 400 "baseDir can only be set via /api/import/folder" (preserves the existing POST guard).

Regression test: PATCH `{ metadata: { kind: 'prototype', linkedDirs: [] } }` on an imported project — `baseDir` survives unchanged.

## Validation

- `pnpm guard` clean
- `pnpm typecheck` clean
- `pnpm --filter @open-design/daemon test` — **795/795 pass** (28 folder-import tests, +2 from this round)

mrcfps (dismissed):
@infinity-nft I reviewed the current folder-import implementation at `f9347e5`, including the metadata-aware daemon paths, symlink/data-dir hardening, desktop/web import flow, contract updates, docs, and the new route/project tests. The previously raised blockers for archive roots and metadata patch preservation look addressed, and I did not find any remaining actionable issues in the changed ranges — thanks for the thorough follow-through here! 🙌


lefarcen (dismissed):
mrcfps reviewed the current implementation and found no remaining actionable issues. The folder-import security hardening (symlink checks, metadata immutability, data-dir guards) and metadata threading across all daemon paths look solid. Approving to unblock merge. 🚀

infinity-nft:
Thanks @lefarcen and @mrcfps for the careful review — three rounds of P1 findings made the implementation significantly more robust than what I started with. Happy to follow up on the deferred OD-level write-protection (`.od-ignore` / trust toggle) in a separate PR whenever the design direction is clearer.

infinity-nft:
Rebased on current main to pick up the recent skill additions (`social-media-dashboard` etc.) — `tests/localized-content.test.ts` was failing CI on the previous head because the locale fixtures expected those skills to be present. No code changes from me; just bringing the branch up to date so the merge commit is clean.

Ran locally on the rebased head:
- `pnpm typecheck` — clean
- `pnpm --filter @open-design/daemon test` — 879/879 pass
- `pnpm --filter @open-design/e2e test` — `localized-content.test.ts` 6/6 pass

Approval status from @lefarcen and @mrcfps should still apply since the diff vs previous head is the upstream rebase (no PR-author-side changes).

infinity-nft:
The CI failure in Attempt #2 is **not caused by this PR**. It's a regression introduced in #621 (`tests/components/NewProjectPanel.test.tsx`, line 57).

The test fixture passes a skill with `designSystemRequired: false` to `NewProjectPanel`. The component correctly hides the design-system picker when the active skill forces no design system — so `'Clay'` never appears in the SSR markup and the `toContain('Clay')` assertion fails.

```
FAIL  tests/components/NewProjectPanel.test.tsx
  × shows the configured default design system as the active project selection
  → expected markup to contain 'Clay'
```

Reproduces on a clean checkout of `upstream/main` — nothing in this PR touches `NewProjectPanel` or any `apps/web` component. The one-line fix is to change the skill fixture in that test from `designSystemRequired: false` to `designSystemRequired: true`.

Happy to include the fix as an unrelated fixup commit if that helps unblock review, or leave it for a separate patch — whichever you prefer.

Siri-Ray (approved):

leilei926524-tech:
Hi @infinity-nft — welcome to Open Design contributors. 🎉

Your first merged PR (#624) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #625, opened 2026-05-06
craft: add form-validation so generated forms aren't stuck in 2018 RHF/Formik patterns

MohamedAbdallah-14:
Final module (5 of 5) in the behavioral craft series proposed in #501. Earlier modules merged: state-coverage (#502), animation-discipline (#515), accessibility-baseline (#587), rtl-and-bidi (#595).

## Why

`accessibility-baseline.md` shipped with the canonical inline-error pattern: `<label for>` + `aria-describedby="hint-id error-id"` + `aria-invalid="true"` + `role="alert"`. That covers WCAG 3.3.1 (Error Identification) and 3.3.2 (Labels) — the static-markup half. It doesn't reach the lifecycle (when does the error appear, when does it clear), the platform validity layer (HTML constraint validation API, `:user-invalid`), the schema-as-contract layer (Standard Schema, server-authoritative validation), or the rest of WCAG 3.3.x (3.3.3 Error Suggestion, 3.3.4 Error Prevention, 3.3.8 / 3.3.9 Accessible Authentication).

A skill emits a saas-landing with a lead-capture form. Without this craft, the agent's defaults are usually a 2018 stack: client-only validation, generic "Invalid input" messages, `:invalid` red borders firing on page load, `form.submit()` skipping the constraint API, RHF + a per-validator resolver shim. None of those are wrong in 2018; all of them are out-of-date in 2026.

## Primary sources

- WHATWG HTML Living Standard (Constraint Validation under "Form control infrastructure"): `validity`, `willValidate`, `setCustomValidity`, `reportValidity`, the readonly + #11841 cross-browser divergence.
- CSS Selectors L4: `:user-invalid` / `:user-valid`, Baseline Newly available 2023 (Chrome 119, Firefox 88, Safari 16.5).
- WCAG 2.2 SC 3.3.x Understanding pages: 3.3.1, 3.3.2, 3.3.3, 3.3.4, 3.3.7, 3.3.8 (new in 2.2), 3.3.9.
- ARIA Authoring Practices Guide forms patterns.
- Standard Schema spec (`@standard-schema/spec`) — the `~standard` interface that lets a form library accept any compliant validator without per-validator resolver shims.
- Baymard 2024 inline-validation research and 2023 adaptive-error-messages research from the checkout-UX benchmark.
- WebAIM Million 2026 forms findings (verified directly against webaim.org/projects/million/): 51% page-level missing-form-label rate, 33.1% input-level rate of 6.9M sampled inputs.

## What's in the file

- **Input state machine.** Eight states (pristine, dirty, touched, invalid-after-touched, invalid-after-submit, recovering, submitting, server-error) with the decision rule that collapses validation-timing debates: errors appear on transition into `invalid-after-touched`, clear on transition out, never appear from pristine or plain dirty. This matches `:user-invalid` for free.
- **Validation timing.** Four rules: blur-after-edit runs the constraint, `input` event clears it, submit runs the schema parse + focus the summary, async checks debounce + announce + don't block submit.
- **Constraint Validation API as the platform floor.** Declarative attributes survive JS failure and integrate with autofill. `setCustomValidity` empty-string-clears (not null), `requestSubmit` honors validation (not `submit`), `readonly` cross-browser caveat with the [whatwg/html#11841](https://github.com/whatwg/html/issues/11841) reference, `inputmode="numeric" pattern="[0-9]*"` for ZIPs / OTPs / card numbers.
- **Error wiring beyond the baseline.** Adaptive messages (Baymard 2023: 98% of audited sites use generic catch-all errors), error summary without `role="alert"` on the focus target (avoids double-announcement), preserve user input on error (Baymard 2024: 34% of e-commerce sites wipe the card field on unrelated errors).
- **Schema as the cross-stack contract.** The 2026 React shape — `useActionState` + Server Actions + Conform + Standard Schema + Zod 4 / Valibot / ArkType — with three rules that survive across stacks: server is the truth, Standard Schema is the contract (not Zod), `novalidate` doesn't mean "skip validation".
- **WCAG 3.3.x beyond Error ID.** 3.3.3 (suggestions), 3.3.4 (legal/financial reversibility), 3.3.8 (CAPTCHAs are what this restricts, not the exception), 3.3.9 (AAA, removes even the narrow exceptions).
- **Native mobile parity.** UIKit / SwiftUI / Compose / Flutter / RN validity primitives + announcement APIs. Compose `isError = true` already wires the AT semantic; don't reach for `Modifier.semantics { error("…") }` by hand.
- **Common mistakes.** 11 lint-able items including `:invalid` vs `:user-invalid` (loudest "untested validation" signal), throwing from Server Actions, `role="alert"` on focus targets, `aria-busy` on submit buttons, email-confirm fields, per-validator resolver shims.

## Skill opt-ins

- `saas-landing` — lead capture form is the primary CTA on most landing pages.
- `mobile-onboarding` — last screen is sign-in / continue-with options.

Both skill bodies were checked for instructions that would override craft guidance — neither contains validation-specific direction. Opt-in alone is sufficient.

## Research and review process

Same workflow as #587 and #595: corpus of 24 depth files (D16-D22 cover forms specifically) read before drafting, three review loops with Claude CLI Opus 4.7 xhigh effort.

- **Loop 1** caught six P0s: an incorrect SwiftUI 18 `Form` validity claim that doesn't exist; the wrong SwiftUI announcement primitive (label/hint changes don't announce — `AccessibilityNotification.Announcement` does); incorrect Compose error-semantics syntax; UIKit `UIAlertController` listed as a validity primitive (it's a modal dialog); contradictory Baymard percentages; the 3.3.8 CAPTCHA sentence reversed (CAPTCHAs are what 3.3.8 restricts, not the exception). Plus eleven P1/P2s including a scrambled React Native row, a stale APCA-style WCAG section number, an `aria-busy="true"` on submit-button anti-pattern, the `:invalid` vs `:user-invalid` framing, the `<select readonly>` cross-browser caveat. All addressed.
- **Loop 2** verified the P0 fixes. Caught one P1 (RN row's columns scrambled) plus four P2s. All addressed.
- **Loop 3** said ship. Three P2 nits applied: Zod 4 `z.email()` top-level form (deprecated `z.string().email()`), WebAIM Million 2026 stat woven in for the form-labels prevalence claim.

## Wazir attribution

Some of the corpus and the framing of "the four layers a real form spans" came from research originally done on Wazir, an OSS project I parked after three months when the architecture didn't pay off. The research itself was solid; republishing the research-grade pieces under OD's craft series is the highest-leverage place for that work to land.

Refs #501.

### DIFF
--- a/craft/form-validation.md
+++ b/craft/form-validation.md
@@ -173,7 +173,7 @@ Three rules that survive across stacks:
 
 - **Server is the truth, client is the optimization.** Same schema runs in both. Returning `{ errors }` from the action (not throwing) is what feeds back into `useActionState`'s state slot — throwing routes to the Error Boundary and loses the form data.
 - **Standard Schema is the contract, not Zod.** A form library that ships per-validator resolver shims (`zodResolver`, `valibotResolver`, etc.) is yesterday's stack. Accept any `~standard`-compliant validator.
-- **`novalidate` on `<form>` does not mean "skip validation".** It means "let the form library repaint errors instead of the browser's bubble." Keep `required` / `pattern` / `type` attributes — progressive enhancement still works without JS, the form library just owns the visible chrome when JS is on. (HTML attribute is lowercase `novalidate`; the IDL property on the form element is `noValidate`.)
+- **`novalidate` on `<form>` does not mean "skip validation".** It means "let the form library repaint errors instead of the browser's bubble." But the trade-off is real: a literal server-rendered `<form novalidate>` disables the browser's submit-blocking and native validation UI **even when JS is unavailable**, which loses the no-JS constraint-validation floor. Pick one of two patterns. **A:** render `<form>` without `novalidate` server-side and have the form library set `form.noValidate = true` after hydration — the no-JS user keeps the browser's native validation, the JS user gets the library's chrome. **B:** ship `novalidate` from the start only when the submit path reaches server validation without JS (Server Action, classic POST handler) so the no-JS user is still protected by the server. Either way, keep `required` / `pattern` / `type` attributes — they survive JS failure and integrate with autofill. (HTML attribute is lowercase `novalidate`; the IDL property on the form element is `noValidate`.)
 
 ## WCAG 3.3.x beyond Error Identification
 
@@ -198,7 +198,7 @@ that ship to mobile (mobile-onboarding, mobile-app, etc.).
 | iOS UIKit | Hand-rolled state on the view controller; `UITextField` doesn't carry a built-in invalid flag | `UIAccessibility.post(notification: .announcement, argument: "Email is required")` |
 | iOS SwiftUI | `TextField` + `@State`-driven validation; no built-in `Form`-level validity API as of iOS 18 | `AccessibilityNotification.Announcement("…").post()` (iOS 17+) |
 | Android Compose | `OutlinedTextField(isError = true, supportingText = { Text("…") })` — `isError` wires the AT error semantic for you | `LiveRegion` semantic on the supporting-text node, or `view.announceForAccessibility(…)` |
-| Flutter | `TextFormField(validator: (v) => …)` inside a `Form`, `formKey.currentState!.validate()` | `SemanticsService.announce(…, TextDirection.ltr)` |
+| Flutter | `TextFormField(validator: (v) => …)` inside a `Form`, `formKey.currentState!.validate()` | `SemanticsService.announce(message, Directionality.of(context))` — never hardcode `TextDirection.ltr`; pull ambient direction so Arabic / Hebrew / Persian flows announce correctly |
 | React Native | Hand-rolled per field; no platform validity flag | `accessibilityLiveRegion="polite"` on the error node (Android) + `AccessibilityInfo.announceForAccessibility(...)` (iOS) |
 
 Two parity rules that catch most AI-generated mobile forms:
--- a/craft/form-validation.md
+++ b/craft/form-validation.md
@@ -197,7 +197,7 @@ that ship to mobile (mobile-onboarding, mobile-app, etc.).
 |---|---|---|
 | iOS UIKit | Hand-rolled state on the view controller; `UITextField` doesn't carry a built-in invalid flag | `UIAccessibility.post(notification: .announcement, argument: "Email is required")` |
 | iOS SwiftUI | `TextField` + `@State`-driven validation; no built-in `Form`-level validity API as of iOS 18 | `AccessibilityNotification.Announcement("…").post()` (iOS 17+) |
-| Android Compose | `OutlinedTextField(isError = true, supportingText = { Text("…") })` — `isError` wires the AT error semantic for you | `LiveRegion` semantic on the supporting-text node, or `view.announceForAccessibility(…)` |
+| Android Compose | `OutlinedTextField(isError = true, supportingText = { Text("…") })` — `isError` wires the AT error semantic for you | `Modifier.semantics { liveRegion = LiveRegionMode.Polite }` on the supporting-text node, or `LocalView.current.announceForAccessibility(message)` |
 | Flutter | `TextFormField(validator: (v) => …)` inside a `Form`, `formKey.currentState!.validate()` | `SemanticsService.announce(message, Directionality.of(context))` — never hardcode `TextDirection.ltr`; pull ambient direction so Arabic / Hebrew / Persian flows announce correctly |
 | React Native | Hand-rolled per field; no platform validity flag | `accessibilityLiveRegion="polite"` on the error node (Android) + `AccessibilityInfo.announceForAccessibility(...)` (iOS) |
 
--- a/craft/form-validation.md
+++ b/craft/form-validation.md
@@ -97,7 +97,7 @@ Rules of the API:
 
 - **Empty string clears `setCustomValidity`.** Not `null`, not no-arg.
 - **`form.requestSubmit()` honors validation; `form.submit()` skips it.** Never call the second.
-- `disabled` controls are barred from validation and not submitted. The HTML spec says `readonly` is also barred, but `readonly` only has defined behavior on `<input>` and `<textarea>` — implementations diverge for `<select readonly>` and `<button readonly>` ([whatwg/html#11841](https://github.com/whatwg/html/issues/11841)). For non-input controls, use `disabled` (or `aria-readonly` + a hidden mirror input if you need the value submitted).
+- `disabled` controls are barred from validation and not submitted. The HTML spec says `readonly` is also barred, but `readonly` only has defined behavior on `<input>` and `<textarea>` — implementations diverge for `<select readonly>` and `<button readonly>` ([whatwg/html#11841](https://github.com/whatwg/html/issues/11841)). For non-input controls where the value must still submit, the safe pattern is `disabled` plus a same-named hidden `<input>` carrying the value, or rendering the non-editable text alongside a hidden `<input>`. `aria-readonly` alone is not enough — a `<select>` or custom widget tagged `aria-readonly="true"` is still interactable, so the visible control can drift while the hidden input ships a stale or different value. If you do use `aria-readonly`, you must also block the interaction or keep both values in sync.
 - `inputmode` is a virtual-keyboard hint, **not** validation. `<input type="text" inputmode="numeric" pattern="[0-9]*">` is the Baymard-recommended shape for ZIPs / OTPs / card numbers; `pattern="[0-9]*"` is the historical iOS-Safari trigger for the numeric keypad on top of `inputmode`. `type="number"` adds spinners, strips leading zeros, applies locale-decimal handling, and varies field width across browsers — wrong for any of these.
 
 ## Error wiring beyond the baseline
@@ -203,7 +203,7 @@ that ship to mobile (mobile-onboarding, mobile-app, etc.).
 
 Two parity rules that catch most AI-generated mobile forms:
 
-- **Use the platform's native validation flag where one exists.** Compose's `isError = true` already wires the semantic; do not also write `Modifier.semantics { error("…") }` by hand — the two desync, and `isError` already drives the AT path.
+- **Use the platform's native validation flag — and pair it with the platform's error-message semantic where one exists.** On Compose, `isError = true` is the right boolean state for the field visuals and AT error-state cue, but it does *not* carry the localized error message. Pair it with `Modifier.semantics { error(message) }` so accessibility services get the actual text — the same string you render in `supportingText`. The trap is duplication: a hand-rolled `Modifier.semantics { error("Email is required") }` next to a different supporting-text string desyncs. Source `error()` from the same state field as `supportingText` so they stay in sync.
 - **Don't mirror web ARIA into mobile semantics.** `aria-describedby` on a SwiftUI `TextField` is a no-op. Use the platform announcement primitive (`AccessibilityNotification.Announcement` on SwiftUI, `UIAccessibility.post` on UIKit, `announceForAccessibility` on Android, `SemanticsService.announce` on Flutter) for state-change events that need to reach the screen reader.
 
 ## Common mistakes (lint these)
--- a/craft/form-validation.md
+++ b/craft/form-validation.md
@@ -66,7 +66,7 @@ The four rules:
 1. **First blur after edit** runs the field-level constraint. Not on focus, not on first keystroke, not on every keystroke.
 2. **Once a field is invalid, switch to `input`-event re-validation** so the error clears the moment input becomes valid. Don't make the user blur again to dismiss it.
 3. **On submit**, run the schema parse. Move focus to the error summary at the top of the form (a heading-led container with `tabindex="-1"`, no `role="alert"` — see the wiring section), or to the first invalid field if no summary exists. Don't move focus on every keystroke.
-4. **Async checks** (uniqueness, address lookup, server-confirmation) debounce 250-500 ms, announce via a polite live region, and never block submit on a network round-trip.
+4. **Async checks** split into two paths. *Background preflight* (uniqueness while typing, address lookup) debounces 250-500 ms, announces via a polite live region, and never gates typing or keeps the submit button disabled indefinitely. *Authoritative server validation on submit* is different: the submit path must await the server's response and surface field errors from it, since the server is the truth. Don't conflate the two — the rule is "don't let a slow background check freeze the form," not "don't ever wait for the server."
 
 CSS gets you most of timing rule 1 for free: style off `:user-invalid`
 not `:invalid`. The `:user-invalid` selector is Baseline Newly
--- a/craft/form-validation.md
+++ b/craft/form-validation.md
@@ -1,0 +1,221 @@
+# Form validation craft rules
+
+Universal rules for form validation lifecycle, error wiring beyond the
+accessibility baseline, and the schema-as-contract layer that makes
+the same validation work on the server and the client. The active
+`DESIGN.md` decides how the field looks; this file decides *when* the
+field tells the user it's wrong, *how* the error reaches assistive
+tech, and *where* the rule lives.
+
+> Grounded in primary sources: WHATWG HTML Living Standard
+> (Constraint Validation section under "Form control infrastructure"),
+> CSS Selectors L4 (`:user-invalid`), WCAG 2.2 SC 3.3.x
+> Understanding pages, ARIA APG forms patterns, Standard Schema spec
+> (`@standard-schema/spec`), Baymard 2024 inline-validation research
+> checkout-UX benchmark, WebAIM Million 2026 forms findings.
+
+## Prior art and scope
+
+Existing OSS forms guidance for AI agents pins to one layer at a time
+— `szilu/ux-designer-skill` is UX-opinion grade with no spec anchors,
+`Community-Access/accessibility-agents/forms-specialist` is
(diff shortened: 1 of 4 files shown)


skills/saas-landing/SKILL.md
```diff
@@ -20,7 +20,7 @@ od:
     requires: true
     sections: [color, typography, layout, components]
   craft:
-    requires: [typography, color, anti-ai-slop]
+    requires: [typography, color, anti-ai-slop, form-validation]
```
lefarcen: P2 `saas-landing` does not currently require or describe an interactive form; it produces CTA buttons and explicitly says no external JS. Requiring `form-validation` for every SaaS landing conflicts with `craft/README.md`'s opt-in rule for artifacts whose primary artifact contains an interactive form, and adds 221 lines of irrelevant prompt pressure unless this skill is also updated to include a lead-capture/signup form.

craft/form-validation.md
```diff
@@ -0,0 +1,221 @@
...
+are you telling me my email address is wrong, I haven't had a chance
+to fill it all out yet?"* Premature firing is the loudest UX failure
+in this space.
+
+The four rules:
+
+1. **First blur after edit** runs the field-level constraint. Not on focus, not on first keystroke, not on every keystroke.
+2. **Once a field is invalid, switch to `input`-event re-validation** so the error clears the moment input becomes valid. Don't make the user blur again to dismiss it.
+3. **On submit**, run the schema parse. Move focus to the error summary at the top of the form (a heading-led container with `tabindex="-1"`, no `role="alert"` — see the wiring section), or to the first invalid field if no summary exists. Don't move focus on every keystroke.
+4. **Async checks** (uniqueness, address lookup, server-confirmation) debounce 250-500 ms, announce via a polite live region, and never block submit on a network round-trip.
```
lefarcen: P2 "never block submit on a network round-trip" is too broad for uniqueness or server-confirmation checks. It conflicts with the later "server is the truth" rule: background async preflight should not gate typing or keep the button disabled indefinitely, but the submit path still has to await authoritative server validation and return field errors.

lefarcen (commented):
Hi @MohamedAbdallah-14! The state machine, timing, and spec-anchoring here are solid — the WHATWG/WCAG/Standard Schema grounding puts this craft file on firm footing alongside accessibility-baseline and rtl-and-bidi. Two suggestions below:

1. The saas-landing skill currently ships CTA buttons without external JS or interactive forms. Requiring form-validation there conflicts with the opt-in rule "primary artifact contains an interactive form" unless saas-landing's SKILL.md is updated to include a lead-capture form.
2. "Never block submit on a network round-trip" for async checks is too broad — background async preflight should not lock the submit button indefinitely, but the final submit path still has to await server validation when needed (uniqueness checks, server-confirmation).

See inline comments for line-specific suggestions.

mrcfps (commented):
@MohamedAbdallah-14 Thanks for putting this together — the craft file is well structured and the validation lifecycle/WCAG/platform layering is a strong addition. I found non-blocking concerns, and they are already captured in the existing inline review threads on `skills/saas-landing/SKILL.md:23` and `craft/form-validation.md:69`: the SaaS landing opt-in currently adds form-validation to a skill that does not describe an actual form, and the async-validation rule should distinguish background preflight checks from authoritative submit-time server validation.


MohamedAbdallah-14:
Thanks @lefarcen and @mrcfps. Both addressed in `e7c8b34`.

**saas-landing opt-in.** Right call. The skill body is a CTA-driven landing with no JS and no interactive form, so the opt-in was injecting ~221 lines of irrelevant prompt pressure and conflicted with the README's "primary artifact contains an interactive form" rule. Dropped `form-validation` from saas-landing's `craft.requires`. Kept it on `mobile-onboarding` where the sign-in screen is a real form.

**Async timing rule.** The previous "never block submit on a network round-trip" was too broad and contradicted the schema-layer "server is the truth" rule two sections later. Split into two paths:

- *Background preflight* (uniqueness while typing, address lookup) debounces 250-500ms, announces via a polite live region, and doesn't gate typing or freeze the submit button.
- *Authoritative server validation on submit* must await the server response and surface its field errors. The server is the truth.

The rule is now framed as "don't let a slow background check freeze the form," not "don't ever wait for the server." That matches how Conform / Server Action stacks actually work in practice.

Ready for re-review.

lefarcen:
Both P2s resolved — the saas-landing opt-in removal and the async-timing split into background-preflight vs authoritative-submit paths are correctly implemented. The craft file is now consistent with the opt-in rule and the schema-layer "server is the truth" principle. Ready from my side.

craft/form-validation.md
```diff
@@ -0,0 +1,221 @@
...
+  password: z.string().min(12),
+});
+// Same schema parses on the Server Action and on the Conform client.
+```
+
+Three rules that survive across stacks:
+
+- **Server is the truth, client is the optimization.** Same schema runs in both. Returning `{ errors }` from the action (not throwing) is what feeds back into `useActionState`'s state slot — throwing routes to the Error Boundary and loses the form data.
+- **Standard Schema is the contract, not Zod.** A form library that ships per-validator resolver shims (`zodResolver`, `valibotResolver`, etc.) is yesterday's stack. Accept any `~standard`-compliant validator.
+- **`novalidate` on `<form>` does not mean "skip validation".** It means "let the form library repaint errors instead of the browser's bubble." Keep `required` / `pattern` / `type` attributes — progressive enhancement still works without JS, the form library just owns the visible chrome when JS is on. (HTML attribute is lowercase `novalidate`; the IDL property on the form element is `noValidate`.)
```
mrcfps: This `novalidate` guidance conflates server-side progressive enhancement with native constraint-validation fallback. The changed line says keeping `required` / `pattern` / `type` means progressive enhancement still works without JS, but a literal server-rendered `<form novalidate>` disables the browser's built-in submit blocking and validation UI when JS is unavailable. That can lead generated forms to lose the exact no-JS constraint-validation floor this section is trying to preserve. Please reword this to make the trade-off explicit, e.g. `novalidate` disables native validation UI/blocking and should only be present when the submit path still reaches server validation without JS, or should be added after hydration if the no-JS browser fallback is desired.

craft/form-validation.md
```diff
@@ -0,0 +1,221 @@
...
+own validity machinery and its own AT path. Skills that emit web-only
+artifacts can skim this section; it's the entry point for skills
+that ship to mobile (mobile-onboarding, mobile-app, etc.).
+
+| Platform | Validity primitive | Error announcement |
+|---|---|---|
+| iOS UIKit | Hand-rolled state on the view controller; `UITextField` doesn't carry a built-in invalid flag | `UIAccessibility.post(notification: .announcement, argument: "Email is required")` |
+| iOS SwiftUI | `TextField` + `@State`-driven validation; no built-in `Form`-level validity API as of iOS 18 | `AccessibilityNotification.Announcement("…").post()` (iOS 17+) |
+| Android Compose | `OutlinedTextField(isError = true, supportingText = { Text("…") })` — `isError` wires the AT error semantic for you | `LiveRegion` semantic on the supporting-text node, or `view.announceForAccessibility(…)` |
+| Flutter | `TextFormField(validator: (v) => …)` inside a `Form`, `formKey.currentState!.validate()` | `SemanticsService.announce(…, TextDirection.ltr)` |
```
mrcfps: The Flutter announcement example hard-codes `TextDirection.ltr`. Since this craft can be combined with the repo's RTL/bidi guidance, generated validation messages in Arabic/Hebrew/Persian flows could be announced with the wrong bidi direction if agents copy the example literally. Please switch the example to use the ambient direction, such as `Directionality.of(context)`, or phrase it as `SemanticsService.announce(message, Directionality.of(context))` so the mobile parity guidance remains locale-safe.

mrcfps (commented):
@MohamedAbdallah-14 Thanks for the thoughtful follow-up here — the previous async-check and saas-landing concerns are addressed, and the new form-validation craft is a strong addition to the behavioral craft series. I found two non-blocking documentation accuracy tweaks that would make the guidance safer for generated implementations. 🙂


MohamedAbdallah-14:
Thanks @mrcfps. Both addressed in `6c104b1`.

**`novalidate` trade-off.** Real catch. The previous wording implied keeping `required`/`pattern`/`type` was enough to preserve no-JS PE, but a literal server-rendered `<form novalidate>` disables the browser's submit-blocking and native validation UI even without JS — losing the floor the section was claiming to protect. Reworded to make the two safe patterns explicit:

- **A:** render `<form>` without `novalidate` server-side, have the form library set `form.noValidate = true` after hydration. No-JS user keeps the browser's native validation; JS user gets the library's chrome.
- **B:** ship `novalidate` from the start only when the submit path reaches server validation without JS (Server Action, classic POST handler) so the no-JS user is still protected by the server.

Either way the constraint attributes stay. The trade-off is now spelled out instead of glossed.

**Flutter `Directionality`.** Caught a real cross-craft conflict. The previous example hardcoded `TextDirection.ltr` which would announce Arabic / Hebrew / Persian validation messages with the wrong bidi direction the moment `rtl-and-bidi` is also opted in (and `mobile-onboarding` opts into both). Switched to `SemanticsService.announce(message, Directionality.of(context))` with an explicit "never hardcode the direction" note in the cell so the rule travels.

Ready for re-review.

mrcfps (commented):
@MohamedAbdallah-14 Thanks for the careful follow-ups here — the prior review points look addressed, and this craft file is shaping up well. I found two non-blocking precision fixes in the new guidance so generated implementations don't copy patterns that can drift from submitted values or miss Android's explicit accessibility error message path. 🙂


craft/form-validation.md
```diff
@@ -0,0 +1,221 @@
...
+`minlength`/`maxlength`, `step`. Cross-field rules and dynamic
+constraints go through `setCustomValidity()` on both `input` and
+`change` events — autofill flows historically fired one without the
+other on some browsers, so listening on both is the cheap defense.
+
+Rules of the API:
+
+- **Empty string clears `setCustomValidity`.** Not `null`, not no-arg.
+- **`form.requestSubmit()` honors validation; `form.submit()` skips it.** Never call the second.
+- `disabled` controls are barred from validation and not submitted. The HTML spec says `readonly` is also barred, but `readonly` only has defined behavior on `<input>` and `<textarea>` — implementations diverge for `<select readonly>` and `<button readonly>` ([whatwg/html#11841](https://github.com/whatwg/html/issues/11841)). For non-input controls, use `disabled` (or `aria-readonly` + a hidden mirror input if you need the value submitted).
```
mrcfps: This non-input readonly fallback still needs to prevent editing, not just expose a readonly semantic. The changed line recommends `aria-readonly` plus a hidden mirror input when a non-input control's value must be submitted, but `aria-readonly` by itself does not stop a `<select>`/custom control from being changed; the visible control can drift while the hidden input submits a stale or different value. That matters because agents copying this guidance could create forms where users see one value and the server receives another. Please reword this to recommend `disabled` plus a same-named hidden mirror input when the value must submit, or render non-editable text plus a hidden input; if a custom widget uses `aria-readonly`, its interaction must also be blocked or kept synchronized with the submitted value.

craft/form-validation.md
```diff
@@ -0,0 +1,221 @@
...
+| iOS SwiftUI | `TextField` + `@State`-driven validation; no built-in `Form`-level validity API as of iOS 18 | `AccessibilityNotification.Announcement("…").post()` (iOS 17+) |
+| Android Compose | `OutlinedTextField(isError = true, supportingText = { Text("…") })` — `isError` wires the AT error semantic for you | `LiveRegion` semantic on the supporting-text node, or `view.announceForAccessibility(…)` |
+| Flutter | `TextFormField(validator: (v) => …)` inside a `Form`, `formKey.currentState!.validate()` | `SemanticsService.announce(message, Directionality.of(context))` — never hardcode `TextDirection.ltr`; pull ambient direction so Arabic / Hebrew / Persian flows announce correctly |
+| React Native | Hand-rolled per field; no platform validity flag | `accessibilityLiveRegion="polite"` on the error node (Android) + `AccessibilityInfo.announceForAccessibility(...)` (iOS) |
+
+Two parity rules that catch most AI-generated mobile forms:
+
+- **Use the platform's native validation flag where one exists.** Compose's `isError = true` already wires the semantic; do not also write `Modifier.semantics { error("…") }` by hand — the two desync, and `isError` already drives the AT path.
```
mrcfps: The Compose guidance is too absolute about avoiding `Modifier.semantics { error("…") }`. `isError = true` is still the right native validation flag for `OutlinedTextField`, but Android's Compose accessibility guidance uses `semantics { error(errorMessage) }` alongside the visible supporting text to provide an expanded, specific error message to accessibility services. Because this craft is meant to produce accessible mobile forms, telling agents not to add the error semantic can leave generated Compose fields with only a generic error state rather than the actual localized message. Please change this rule to something like: keep `isError = true` for the field state/visuals, and when an accessible error message is needed, add a synchronized `Modifier.semantics { error(message) }` rather than duplicating or desynchronizing separate error strings.

lefarcen (commented):
Thanks @MohamedAbdallah-14 for the follow-up! Both P2 documentation accuracy concerns from mrcfps's second review are addressed:

**`novalidate` trade-off** — the new wording explicitly spells out the two safe patterns: (A) render `<form>` without `novalidate` server-side and set `form.noValidate = true` after hydration, or (B) ship `novalidate` from the start only when the submit path reaches server validation without JS. This resolves the conflation and makes the no-JS constraint-validation floor trade-off clear.

**Flutter `Directionality`** — switched to `SemanticsService.announce(message, Directionality.of(context))` with an explicit "never hardcode the direction" note, fixing the cross-craft RTL/bidi conflict.

**One minor note**: commit `6c104b1` includes `Co-Authored-By: Claude Opus 4.7...` which violates this repo's `AGENTS.md` policy ("Git commits must not include Co-authored-by trailers"). The maintainer will likely squash on merge, but if you're rebasing before then, dropping that trailer keeps the history clean.

Otherwise looks good from the documentation content angle. Deferring final merge decision to the maintainer. 🎉

MohamedAbdallah-14:
Round-3 fixes in `d252485`:

- *Non-input readonly fallback.* You're right — `aria-readonly` on a `<select>` or custom widget doesn't actually block interaction, so the visible control can drift while the hidden mirror submits a stale value. Tightened to: prefer `disabled` plus a same-named hidden input, or non-editable text plus a hidden input. If `aria-readonly` is used, the interaction has to be blocked separately or the two values kept in sync.
- *Compose error message.* Real catch — my previous rule lost the localized message. Reframed: `isError = true` is the right boolean state for visuals and the AT error cue, but it doesn't carry the text. Pair it with `Modifier.semantics { error(message) }`, sourcing both from the same state field as `supportingText` so they stay in sync. The trap is duplication, not the API.

Ready for re-review.

craft/form-validation.md
```diff
@@ -0,0 +1,221 @@
...
+Web validation primitives don't auto-translate. Each platform has its
+own validity machinery and its own AT path. Skills that emit web-only
+artifacts can skim this section; it's the entry point for skills
+that ship to mobile (mobile-onboarding, mobile-app, etc.).
+
+| Platform | Validity primitive | Error announcement |
+|---|---|---|
+| iOS UIKit | Hand-rolled state on the view controller; `UITextField` doesn't carry a built-in invalid flag | `UIAccessibility.post(notification: .announcement, argument: "Email is required")` |
+| iOS SwiftUI | `TextField` + `@State`-driven validation; no built-in `Form`-level validity API as of iOS 18 | `AccessibilityNotification.Announcement("…").post()` (iOS 17+) |
+| Android Compose | `OutlinedTextField(isError = true, supportingText = { Text("…") })` — `isError` wires the AT error semantic for you | `LiveRegion` semantic on the supporting-text node, or `view.announceForAccessibility(…)` |
```
mrcfps: The Compose row currently names a `LiveRegion` semantic for supporting text. In Compose the actual semantics property is `liveRegion`, assigned with `LiveRegionMode.Polite` or `LiveRegionMode.Assertive`; there isn't a `LiveRegion` semantic API to copy from this table cell. Because this craft is intended as implementation guidance for generated mobile forms, agents may turn this into non-compiling or guessed Compose snippets. Please make the example concrete, e.g. `Modifier.semantics { liveRegion = LiveRegionMode.Polite }` on the supporting-text node (or keep the alternative as `LocalView.current.announceForAccessibility(message)`).

mrcfps (commented):
@MohamedAbdallah-14 Thanks for the continued follow-ups — the prior concerns look addressed, and the form-validation craft is much stronger. I found one small non-blocking Compose API precision issue so agents copy the correct live-region semantic when generating Android code. 🙂


MohamedAbdallah-14:
Fixed in `8e64a38`. The Compose row named a `LiveRegion` semantic that doesn't exist as a Compose API. Real spelling is `Modifier.semantics { liveRegion = LiveRegionMode.Polite }` on the supporting-text node, with `LocalView.current.announceForAccessibility(message)` as the alternative for one-shot announcements. Updated the cell so a copy-paste compiles.

Ready for re-review.

mrcfps (approved):
@MohamedAbdallah-14 I reviewed the final `form-validation` craft addition, the craft README registration, and the remaining mobile-onboarding opt-in after the follow-up fixes. The current head resolves the earlier precision issues around async validation, `novalidate`, readonly value submission, Flutter directionality, and Compose accessibility semantics, and I did not find any remaining actionable correctness or maintainability concerns. Thanks for the careful iteration on this craft module — it is a solid addition to the behavioral craft series. 🙂

PR #627, opened 2026-05-06
fix(daemon): add required env field to McpServerStdio + recover from -32602 on set_model

goemonwanwan:
### 1. McpServerStdio `env` field missing (root cause of `-32602 Invalid params`)

`buildLiveArtifactsMcpServersForAgent()` generates MCP server descriptors without the `env` field. The ACP schema (`McpServerStdio`) marks `env` as `List[EnvVariable]` with **no default value** — it is required.

When the daemon sends `session/new` with a non-empty `mcpServers` array, Pydantic V2 Union-discriminated validation tries `HttpMcpServer`, `SseMcpServer`, and `McpServerStdio` in order. All three fail because:

- `HttpMcpServer` → missing `url`, `headers`, `type`
- `SseMcpServer` → missing `url`, `headers`, `type`  
- `McpServerStdio` → missing `env`

This cascades into a 7-error validation failure returned as `-32602 Invalid params`, which terminates the ACP session before it reaches `session/set_model` or `session/prompt`.

**This bug is invisible when `mcpServers` resolves to an empty array** (no live-artifacts token), so it only manifests when `mcpDiscovery: 'mature-acp'` agents (Hermes, Devin, Kimi) receive a non-empty MCP server list.

**Fix**: Add `env: []` to the descriptor object.

### 2. ACP recovery for `-32602 Invalid params` on `session/set_model`

Extends the existing `-32603 Internal error` recovery logic to also handle `-32602` when `session/set_model` fails. This allows the prompt to proceed with the default model instead of hanging or timing out, making the daemon more resilient when an ACP agent rejects the model ID.

## Testing

- Confirmed via direct ACP communication test (`hermes acp --accept-hooks`) that `initialize` → `session/new` → `session/set_model` all succeed when `env: []` is present
- Confirmed via Pydantic validation that omitting `env` produces 7 validation errors (reproducing the exact `-32602` reported in the wild)
- Full end-to-end test: Open Design → Hermes chat completed successfully after the fix

### DIFF
--- a/apps/daemon/src/acp.ts
+++ b/apps/daemon/src/acp.ts
@@ -314,15 +314,17 @@ export function attachAcpSession({
       // After response completion, any late-arriving errors from the agent
       // (pipe-broken, cleanup race conditions, etc.) are safe to ignore.
       if (finished) return;
-      // JSON-RPC -32603 "Internal error" handling:
-      // Unexpected-id -32603 errors are cleanup noise — suppress them.
+      // JSON-RPC -32603 "Internal error" and -32602 "Invalid params" handling:
+      // Unexpected-id errors are cleanup noise — suppress them.
       // Expected-id -32603 errors for session/set_model fall through to the
       // recovery block below. All other expected-id -32603 errors (initialize,
       // session/new, session/prompt) are real failures — call fail().
-      if (raw.error?.code === -32603 && raw.id !== expectedId) {
+      // Expected-id -32602 errors for session/set_model are treated as "method
+      // not supported / params rejected" and also fall through to recovery.
+      if ((raw.error?.code === -32603 || raw.error?.code === -32602) && raw.id !== expectedId) {
         return;
       }
-      if (raw.error?.code === -32603 && raw.id === expectedId) {
+      if ((raw.error?.code === -32603 || raw.error?.code === -32602) && raw.id === expectedId) {
         if (raw.id === setModelRequestId) {
           // Fall through — the recovery block will handle this
         } else {
@@ -368,12 +370,12 @@ export function attachAcpSession({
       }
       return;
     }
-    // Recovery: if session/set_model failed with -32603, fall back to
+    // Recovery: if session/set_model failed with -32603 or -32602, fall back to
     // sending the prompt with the default (already-active) model.
     // This is scoped to the exact set_model request id to avoid
     // triggering on prompt or other request failures.
     if (
-      raw.error?.code === -32603 &&
+      (raw.error?.code === -32603 || raw.error?.code === -32602) &&
       raw.id === setModelRequestId &&
       promptRequestId === null
     ) {
--- a/apps/daemon/src/acp.ts
+++ b/apps/daemon/src/acp.ts
@@ -314,26 +314,32 @@ export function attachAcpSession({
       // After response completion, any late-arriving errors from the agent
       // (pipe-broken, cleanup race conditions, etc.) are safe to ignore.
       if (finished) return;
-      // JSON-RPC -32603 "Internal error" and -32602 "Invalid params" handling:
-      // Unexpected-id errors are cleanup noise — suppress them.
-      // Expected-id -32603 errors for session/set_model fall through to the
-      // recovery block below. All other expected-id -32603 errors (initialize,
-      // session/new, session/prompt) are real failures — call fail().
-      // Expected-id -32602 errors for session/set_model are treated as "method
-      // not supported / params rejected" and also fall through to recovery.
-      if ((raw.error?.code === -32603 || raw.error?.code === -32602) && raw.id !== expectedId) {
+      // JSON-RPC error handling:
+      // -32603 "Internal error": unexpected-id errors are cleanup noise — suppress.
+      //   Expected-id errors for session/set_model fall through to the recovery
+      //   block. All others (initialize, session/new, session/prompt) are real
+      //   failures — call fail().
+      // -32602 "Invalid params": these are real validation failures. Only
+      //   suppress when they match setModelRequestId so the recovery block handles
+      //   them. Any other -32602 (unexpected-id or non-set_model expected-id) is
+      //   a genuine protocol error — call fail().
+      if (raw.error?.code === -32603 && raw.id !== expectedId) {
         return;
       }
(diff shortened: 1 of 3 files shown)


lefarcen:
Hi @goemonwanwan! 🎉

Thanks for the contribution — excellent catch on the missing `env` field root cause.

— open-design team

apps/daemon/src/agents.ts
```diff
@@ -940,6 +940,7 @@ export function buildLiveArtifactsMcpServersForAgent(def, { enabled = true, comm
       name: 'open-design-live-artifacts',
       command,
       args: [...argsPrefix, 'mcp', 'live-artifacts'],
+      env: [],
```
mrcfps, line 943: This new field changes the exact descriptor returned by `buildLiveArtifactsMcpServersForAgent`, but the existing daemon unit tests still assert descriptors without `env` in `apps/daemon/tests/agents.test.ts` (the Hermes/Kimi expectations and the daemon-resolved command expectation). Those `assert.deepEqual(...)` checks are exact object comparisons, so adding `env: []` here leaves the daemon test suite failing even though the runtime fix is correct.

Please update the expected live-artifacts MCP descriptors in `apps/daemon/tests/agents.test.ts` to include `env: []` as well, and keep that as coverage that every generated ACP stdio MCP server includes the required env array.

mrcfps (changes requested):
@goemonwanwan Thanks for the focused fix here; I reviewed the ACP `session/set_model` recovery changes and the live-artifacts MCP descriptor update. One blocking test-maintenance issue remains: the exact-shape daemon tests for `buildLiveArtifactsMcpServersForAgent` still need to be updated for the new required `env` field, otherwise the daemon test suite will fail once dependencies are installed. Appreciate the clear root-cause write-up and targeted change! 🙏


apps/daemon/src/acp.ts
```diff
@@ -314,15 +314,17 @@ export function attachAcpSession({
...
-      // Unexpected-id -32603 errors are cleanup noise — suppress them.
+      // JSON-RPC -32603 "Internal error" and -32602 "Invalid params" handling:
+      // Unexpected-id errors are cleanup noise — suppress them.
       // Expected-id -32603 errors for session/set_model fall through to the
       // recovery block below. All other expected-id -32603 errors (initialize,
       // session/new, session/prompt) are real failures — call fail().
-      if (raw.error?.code === -32603 && raw.id !== expectedId) {
+      // Expected-id -32602 errors for session/set_model are treated as "method
+      // not supported / params rejected" and also fall through to recovery.
+      if ((raw.error?.code === -32603 || raw.error?.code === -32602) && raw.id !== expectedId) {
```
lefarcen: **P1** — Over-broad `-32602` suppression

The new handling suppresses **every** unexpected-id Invalid Params error as cleanup noise:

```typescript
if ((raw.error?.code === -32603 || raw.error?.code === -32602) && raw.id !== expectedId) {
  return; // suppress
}
```

Unlike `-32603`, `-32602` is exactly the validation failure class this PR is trying to expose/recover from. If an ACP implementation returns `-32602` with `id: null` or another non-current id, the daemon will ignore the real protocol error and wait for a stage timeout.

**Fix:** Keep unexpected-id suppression scoped to `-32603` only. Handle `-32602` only when `raw.id === setModelRequestId`, otherwise fail with the JSON-RPC error:

```typescript
if (raw.error?.code === -32603 && raw.id !== expectedId) {
  return; // cleanup noise
}
if (raw.error?.code === -32602 && raw.id !== setModelRequestId) {
  fail(raw.error); // real validation failure
}
if ((raw.error?.code === -32603 || raw.error?.code === -32602) && raw.id === expectedId) {
  // existing fall-through for set_model recovery
}
```

apps/daemon/src/agents.ts
lefarcen, line 943: **P3** (nit, non-blocking) — Bare `command` default is a footgun

The helper defaults `command` to bare `od`, and tests codify that default, but:
- The ACP stdio MCP server contract requires an executable path
- Bare `od` can resolve to Unix octal-dump (this repo's own prompt warns about it)
- Runtime currently passes `process.execPath` + `OD_BIN`, so the default is never hit in production, but it's a trap for future callers

**Suggestion:** Make `command`/`argsPrefix` explicit required inputs for MCP-enabled calls, or assert `path.isAbsolute(command)` before returning a descriptor. Not blocking this PR, but worth considering for safer defaults.

lefarcen (dismissed):
Hey @goemonwanwan, thanks for the PR! 👋

Nice catch on the root cause analysis — the missing `env` field diagnosis matches the Pydantic validation behavior exactly. Found 2 critical issues that need addressing before merge:

**1. Test fixtures out of sync (P1)**

The PR adds `env: []` to the MCP descriptor returned by `buildLiveArtifactsMcpServersForAgent()`, but the existing tests in `apps/daemon/tests/agents.test.ts` (lines 173, 180, 205) still expect the old shape without `env`. This leaves the daemon unit suite red as soon as dependencies are installed.

**Fix:** Update all expected MCP descriptor fixtures to include `env: []`:

```typescript
{
  name: 'open-design-live-artifacts',
  command: ...,
  args: ...,
  env: [],
}
```

**2. Over-broad `-32602` suppression (P1)**

See inline comment on `acp.ts:324`.

**3. Defensive suggestion (P3, non-blocking)**

See inline comment on `agents.ts:936`.

lefarcen (commented):
Nice work @goemonwanwan! 🎉

Both P1 issues from the previous review are now resolved:

**✅ Test fixtures updated**
Added `env: []` to all three MCP descriptor expectations in `apps/daemon/tests/agents.test.ts` (lines 178, 186, 211). The daemon test suite will now pass.

**✅ Scoped `-32602` suppression**
The new logic in `acp.ts` correctly scopes `-32602` suppression to only `setModelRequestId`, and explicitly fails for other `-32602` errors. This prevents masking real protocol validation failures while still allowing graceful recovery when `session/set_model` rejects the model ID.

The PR is clean from my side. Looks good to me; deferring final approval to a maintainer.

mrcfps (approved):

leilei926524-tech:
Hi @goemonwanwan — quick note from the Open Design team.

What you get there:

PR #639, opened 2026-05-06
security: harden daemon against prompt injection, credential leak, and supply chain risks

alex-pezarro-portswigger:
## Summary

Security hardening across 10 findings identified during a full review of the daemon, Electron shell, and CI/CD pipeline.

### Critical / High

- **Prompt injection** (`prompts/system.ts`): ~20 user-controlled metadata fields (`template title`, `category`, `voice`, `imageStyle`, etc.) are now run through `sanitizePromptField()` before being interpolated into the agent system prompt. Strips control characters, collapses markdown structural tokens (`---`, `#`, ` ``` `), and truncates to 500 chars. The existing backtick-escape on `tpl.prompt` is unchanged.
- **Credential leak to agent CLIs** (`agents.ts`): `spawnEnvForAgent()` previously only stripped `ANTHROPIC_API_KEY` from the Claude adapter. A new `SENSITIVE_ENV_PATTERNS` list (16 patterns: `OPENAI_API_KEY`, `GH_TOKEN`, `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `NPM_TOKEN`, `VERCEL_TOKEN`, `OD_*_API_KEY`, etc.) is now applied to **all** agent child processes. The existing Claude + `ANTHROPIC_BASE_URL` exception is preserved.
- **Deploy hook script injection** (`deploy.ts`): `injectDeployHookScript()` accepted any `http(s)://` URL from `OD_DEPLOY_HOOK_SCRIPT_URL`, injecting a `<script>` into every Vercel deployment. Added an empty `DEPLOY_HOOK_ALLOWED_HOSTS` set — all hostnames blocked by default. Removed the `options.hookScriptUrl` programmatic override path from `buildDeployFilePlan()`.
- **ACP over-permissive auto-approval** (`acp.ts`): `choosePermissionOutcome()` previously preferred `approve_for_session` (session-wide grant). Reversed to prefer `allow_once` → `allow_always` → `approve_for_session`. Also emits a `permission_request` SSE event before replying so the UI can surface what was auto-approved.
- **API key file permissions** (`media-config.ts`): `media-config.json` is now written with `mode: 0o600` + `chmodSync`. Existing files have permissions tightened on first read (silent migration).

### Medium

- **Daemon session token** (`server.ts`): `DAEMON_SESSION_TOKEN` (32-byte random, never persisted) generated at startup. New `requireSessionToken` middleware applied to all `/api` routes except `/health`, `/version`, `/daemon-token`, and live-artifact preview routes. Bootstrap endpoint `GET /api/daemon-token` (localhost-restricted, unauthenticated) lets the web frontend fetch the token once on startup.
- **ZIP symlink extraction** (`claude-design-import.ts`): Reject ZIP entries whose Unix external attributes indicate a symlink (`externalFileAttributes >>> 16 & 0xF000 === 0xA000`) before extraction. Post-extraction `lstat()` scan unlinks any symlinks that slipped through (defense-in-depth).
- **Spritesheet download size cap** (`community-pets-sync.ts`): Added `MAX_SPRITESHEET_BYTES = 10 MiB` upper bound in `writePet()` to prevent disk exhaustion from oversized downloads.
- **CI action pinning** (`.github/workflows/*.yml`): All 8 third-party actions across 7 workflow files pinned to full 40-character commit SHAs. Version tag preserved in a comment. Covers `actions/checkout`, `actions/setup-node`, `pnpm/action-setup`, `actions/upload-artifact`, `actions/download-artifact`, `peter-evans/create-pull-request`, `cloudflare/wrangler-action`, and `lowlighter/metrics`.

### Low

- **Design system path traversal** (`design-systems.ts`): `readDesignSystem(root, id)` now validates `id` against `/^[a-z0-9][a-z0-9._-]*$/` before `path.join()`, blocking traversal attempts like `../../etc/passwd`.

## Test plan

- [ ] Verify `sanitizePromptField` strips `\n# OVERRIDE` and `---` from template titles in a local run
- [ ] Confirm agent child processes no longer inherit `OPENAI_API_KEY` / `GH_TOKEN` from the shell
- [ ] Confirm deploy hook injection is blocked with a test URL (should return empty artifact without the script tag)
- [ ] Verify `media-config.json` is created with `0o600` on a fresh install
- [ ] Confirm the web frontend can bootstrap via `GET /api/daemon-token` and subsequent API calls succeed with the `x-od-session-token` header
- [ ] Attempt to import a ZIP containing a symlink entry and confirm it is rejected
- [ ] Confirm all workflow SHAs resolve to the expected tags via `gh api`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,15 +32,15 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: Setup pnpm
-        uses: pnpm/action-setup@v5
+        uses: pnpm/action-setup@a8198c4bff370c8506180b035930dea56dbd5288  # v5
         with:
           version: 10.33.2
 
       - name: Setup Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6
         with:
           node-version: 24
           cache: pnpm
--- a/.github/workflows/landing-page-ci.yml
+++ b/.github/workflows/landing-page-ci.yml
@@ -36,15 +36,15 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: Setup pnpm
-        uses: pnpm/action-setup@v5
+        uses: pnpm/action-setup@a8198c4bff370c8506180b035930dea56dbd5288  # v5
         with:
           version: 10.33.2
 
       - name: Setup Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6
         with:
           node-version: 24
           cache: pnpm
--- a/.github/workflows/landing-page-deploy.yml
+++ b/.github/workflows/landing-page-deploy.yml
@@ -29,15 +29,15 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: Setup pnpm
-        uses: pnpm/action-setup@v5
+        uses: pnpm/action-setup@a8198c4bff370c8506180b035930dea56dbd5288  # v5
         with:
           version: 10.33.2
 
       - name: Setup Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6
         with:
           node-version: 24
           cache: pnpm
@@ -86,7 +86,7 @@ jobs:
           NODE
 
       - name: Deploy to Cloudflare Pages
-        uses: cloudflare/wrangler-action@v3
+        uses: cloudflare/wrangler-action@9acf94ace14e7dc412b076f2c5c20b8ce93c79cd  # v3
         with:
           apiToken: ${{ secrets.CLOUDFLARE_API_TOKEN }}
           accountId: ${{ secrets.CLOUDFLARE_ACCOUNT_ID }}
--- a/.github/workflows/metrics.yml
+++ b/.github/workflows/metrics.yml
@@ -21,7 +21,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: Generate GitHub repository metrics
-        uses: lowlighter/metrics@latest
+        uses: lowlighter/metrics@65836723097537a54cd8eb90f61839426b4266b6  # v3.34
         with:
           # Output path; the action opens/updates a PR when this file changes.
           # Requires manual review to merge. If metrics unchanged, no PR is created.
--- a/.github/workflows/refresh-contributors-wall.yml
+++ b/.github/workflows/refresh-contributors-wall.yml
@@ -25,7 +25,7 @@ jobs:
 
     steps:
       - name: Checkout
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
 
       - name: Refresh cache bust date
         run: |
@@ -38,7 +38,7 @@ jobs:
           perl -0pi -e "s/cache_bust=\d{4}-\d{2}-\d{2}/cache_bust=$DATE/g" README*.md
 
       - name: Create refresh pull request
-        uses: peter-evans/create-pull-request@v8
+        uses: peter-evans/create-pull-request@5f6978faf089d4d20b00c7766989d076bb2fc7f1  # v8
         with:
           # Auth mirrors the metrics workflow: prefer a repository token that can
           # create pull requests, with GITHUB_TOKEN as a fallback for repos where
--- a/.github/workflows/release-beta.yml
+++ b/.github/workflows/release-beta.yml
@@ -51,12 +51,12 @@ jobs:
       version_tag: ${{ steps.beta.outputs.version_tag }}
     steps:
       - name: Checkout
-        uses: actions/checkout@v6.0.2
+        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd  # v6.0.2
         with:
           fetch-depth: 0
 
       - name: Setup Node.js
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@48b55a011bda9f5d6aeb4c2d9c7362e8dae4041e  # v6
         with:
           node-version: 24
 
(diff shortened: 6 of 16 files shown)


lefarcen:
Hi @alex-pezarro-portswigger! 🎉

Thanks for this comprehensive security review — 10 findings across prompt injection, credential isolation, and supply-chain hardening is exactly the kind of depth open-design needs.

— open-design team

lefarcen (commented):
Hey @alex-pezarro-portswigger — this is a **comprehensive security review** that covers all the right areas. The 10 findings span prompt injection, credential isolation, supply-chain pinning, and daemon hardening. I've gone through the implementation and have a few high-value questions and observations below.

## Overall assessment

**Good:**
- Prompt injection mitigation via `sanitizePromptField()` is the right approach for user-controlled metadata interpolated into system prompts
- Credential isolation with `SENSITIVE_ENV_PATTERNS` catches the major API keys
- CI action SHA pinning is exactly what OSSF recommends
- Daemon session token adds a defense layer against unauthorized localhost API access
- ZIP symlink rejection prevents traversal exploits during `.claude` import
- File permission tightening on `media-config.json` closes a credential leak vector

**Questions / concerns:**

### P2 — `sanitizePromptField` truncation may break legitimate long content

**File:** `apps/daemon/src/prompts/system.ts:81`

The 500-char hard truncation in `sanitizePromptField()` could silently break:
- Template titles with detailed context (e.g., "Brand refresh for Acme Corp Q4 2024 — mobile-first, accessibility AAA, dark mode default, 12-point type scale")
- Multi-paragraph style notes from users who paste detailed brand guidelines into `imageStyle` / `voice`

The sanitizer is called on **every** user-controlled metadata field before interpolation, which is correct for XSS defense, but the truncation happens **silently** — no warning, no error, no indication in the UI that the user's input was cut off.

**Suggestion:** Either (a) raise the limit to 2000 chars for fields like `imageStyle` / `voice` / template `summary`, or (b) return a truncation flag so the UI can warn the user when their input exceeds the limit.

### P2 — `SENSITIVE_ENV_PATTERNS` doesn't cover all common secrets

**File:** `apps/daemon/src/agents.ts:1185-1200`

The list is solid for major cloud providers, but misses a few common ones that could leak if set in the shell:
- `SLACK_TOKEN` / `SLACK_BOT_TOKEN` / `SLACK_WEBHOOK_URL`
- `DISCORD_TOKEN` / `DISCORD_WEBHOOK_URL`
- `STRIPE_SECRET_KEY` / `STRIPE_PUBLISHABLE_KEY`
- `TWILIO_AUTH_TOKEN` / `TWILIO_ACCOUNT_SID`
- `SENTRY_DSN` (less critical, but still leaks project topology)
- `DATABASE_URL` / `POSTGRES_URL` / `MYSQL_URL` (common in Rails/Django/Node apps)

These are wildcards that appear in many dev environments. If a user has `SLACK_TOKEN` set and spawns a Claude agent, that token gets inherited unless filtered.

**Suggestion:** Add a catch-all pattern like `/^.*_(TOKEN|SECRET|KEY|PASSWORD|PASS|PWD)$/i` at the end of the list to cover arbitrary `FOO_TOKEN` / `BAR_SECRET` / `BAZ_PASSWORD` env vars. Still preserve the Claude `ANTHROPIC_BASE_URL` exception logic.

### P3 — `DEPLOY_HOOK_ALLOWED_HOSTS` is empty, but no migration path for existing users

**File:** `apps/daemon/src/deploy.ts:97`

The change from "accept any `http(s)://` URL" to "block everything by default" is correct from a security standpoint, but:
- Existing users who **were** legitimately using `OD_DEPLOY_HOOK_SCRIPT_URL` (e.g., analytics snippet, Sentry init, Vercel Edge config) will suddenly see their deploy hooks silently dropped
- No error message, no migration guide, no deprecation warning
- The env var still exists in the codebase, but it's now a no-op unless someone manually edits the daemon source to add their hostname to `DEPLOY_HOOK_ALLOWED_HOSTS`

**Suggestion:** Either (a) add a startup warning if `OD_DEPLOY_HOOK_SCRIPT_URL` is set but the hostname isn't in the allowlist ("Deploy hook script URL is set but blocked — see docs for allowlist configuration"), or (b) document the allowlist in the README / env-vars section so users know how to opt back in for trusted hostnames.

### P3 — `requireSessionToken` middleware order may allow bypass via typo routes

**File:** `apps/daemon/src/server.ts:1127-1131`

The session token middleware is applied to `/api` with a regex exemption for `/health`, `/version`, `/daemon-token`, and live-artifact preview routes. The regex is:

```ts
/^\/health$|^\/version$|^\/daemon-token$|^\/live-artifacts\/[^/]+\/preview$/
```

This regex is checked against `req.path`, which is **the URL path after the mount point** (i.e., `/api` is already stripped). That means:
- `/api/health` → `req.path = "/health"` → exempt ✅
- `/api/healthz` → `req.path = "/healthz"` → **NOT exempt, requires token** ✅
- `/api//health` → `req.path = "//health"` → **might bypass** if Express normalizes `//` differently

Express **does** normalize `//` to `/` in most cases, but the regex is strict (`^\/health$`) and won't match if any middleware or proxy upstream doesn't normalize.

**Suggestion:** Add a leading `/` normalization guard or use `req.path.replace(/^\/+/, '/')` before the regex check to ensure `//health` → `/health`.

### P3 — `media-config.json` permissions are platform-dependent

**File:** `apps/daemon/src/media-config.ts:115-116,131-132`

The `fs.chmodSync(configFile(projectRoot), 0o600)` calls are wrapped in `try/catch` with silent failure, which is correct for Windows / network filesystems that don't support Unix permissions. However:
- On Windows, `0o600` is **ignored** — the file is created with default ACLs (typically world-readable if not in a user-specific directory)
- On network mounts (SMB, NFS without Unix extensions), `chmod` may succeed but not enforce the permissions

**This is not a blocker** — the current approach is defensible (better to write the file and warn than to fail on Windows), but the PR description says "API key file permissions" as if it's a universal fix.

**Suggestion:** Add a comment above the `try/catch` blocks explaining the platform limitations: `// chmod may be ignored on Windows / network mounts — permissions are best-effort`. Or, on Windows, consider using Windows ACLs via a native module (but that's a much larger change).

---

## Minor observations (non-blocking)

- **CI action SHA pinning:** All 8 actions across 7 workflows are pinned. This is excellent. The version tag is preserved in comments. ✅
- **ACP permission outcome preference:** Reversing to `allow_once` → `allow_always` → `approve_for_session` is the right default (principle of least privilege). ✅
- **Spritesheet size cap:** 10 MiB is generous but reasonable for a community pet sync. ✅
- **Design system path traversal:** The regex `SAFE_ID_RE` correctly blocks `../`, but also blocks uppercase IDs (e.g., `Acme-Brand-2024`). This might be intentional (forcing lowercase design system IDs), but if not, consider allowing `[A-Za-z0-9]`.

---

## Test plan gap

The test plan in the PR description lists manual verification steps, which is fine for a first pass, but:
- No automated test for `sanitizePromptField()` — a unit test for "does it strip `\n# OVERRIDE`" would prevent regression
- No automated test for `SENSITIVE_ENV_PATTERNS` — a unit test that spawns an agent with `OPENAI_API_KEY=secret` set and verifies it's not in `child.env` would catch future gaps
- No automated test for `requireSessionToken` middleware — an integration test that hits `/api/agents` without the token header and expects `401` would prevent accidental exemptions

**Suggestion:** Add a test file `apps/daemon/src/__tests__/security-hardening.test.ts` covering at least `sanitizePromptField()` and `spawnEnvForAgent()`. The middleware test can go in an existing `server.test.ts` if one exists.

---

## Summary

This PR closes **10 real attack surfaces** and follows industry best practices (OSSF supply-chain guidance, OWASP injection mitigation, least-privilege permissions). The P2 concerns are about **edge cases** (long user input, additional secret patterns) and **migration UX** (existing deploy hook users), not fundamental flaws. The P3 observations are defensive improvements.

**Status:** The fixes are correct and should land. The P2 items are worth addressing before merge (or in a follow-up if time-sensitive), and the P3 items can be follow-up PRs if needed.

Deferring final approval to a maintainer — this is exactly the kind of hardening open-design needs, and I don't want to rush the review on my own.

PR #402, opened 2026-05-03
feat(nix): Add official flake with home-manager and NixOS support

GodTamIt:
This adds flake support with this repo so that people using Nix or NixOS can easily deploy this locally:

  * Provides a devshell with the appropriate tooling versions that's easily accessible to Nix developers
  * Flake automatically tracks `package.json` versions
  * Makes it easily deployed via `home-manager` (the recommended route)
  * User can independently deploy daemon or web frontend
  * Flake automatically configures daemon to try and read the right PATHs for detection.
  * If user opts into enabling the web frontend, flake automatically configures Caddy reverse proxy to point properly to daemon
  * Comes with CI checks
  * In the future, the CI check also comes with a cachix setup, in the case the maintainers want to provide users with a build cache.

PR #641, opened 2026-05-06
feat(daemon): support OD_EXTRA_ORIGINS for reverse-proxy deploys

lonormaly:
## Why

Operators today can only expose the OD web UI on **loopback** or on the **explicit `--host` bind address**. Any local reverse proxy that fronts the daemon at a different origin (Caddy, [Vercel Portless](https://github.com/vercel-labs/portless), an organisation-wide nginx, a Tailscale Funnel-style egress, …) gets a `403 {"error":"Cross-origin requests are not allowed"}` on every `/api` call, because the browser's `Origin` header doesn't match anything in the allowlist.

Concrete repro from my setup: I run OD behind portless at `http://design.localhost:1355` so it shares a single proxy port with the rest of my dev tools. Chat works for the first turn (no `Origin` because the request comes from `tools-dev`'s spawned agent), but the second user message from the browser fails with the 403 above.

## What

Add an opt-in `OD_EXTRA_ORIGINS` env var: comma-separated list of extra browser origins the daemon trusts. Patched in both places that build the allowlist:

1. The global `/api` origin middleware in `startServer()` — fixes the chat flow.
2. The inline `isLocalSameOrigin()` helper used by media-generation routes (image / video / audio / HyperFrames) — fixes the rest of the surface that lives behind the same proxy. `allowedHosts` is also extended with the `URL.host` of each entry so DNS-rebind protection stays intact when the proxy passes the original `Host` header through.

Behaviour when the env var is unset is **unchanged** — the existing loopback + bind-host policy still applies on its own. Empty / blank entries are ignored so `OD_EXTRA_ORIGINS=,,,` is the same as no env var.

## Example

```bash
OD_EXTRA_ORIGINS=http://design.localhost:1355 \
  pnpm tools-dev run web --web-port 5176 --daemon-port 7458
```

Or with multiple proxy origins:

```bash
OD_EXTRA_ORIGINS=http://design.localhost:1355,https://od.example.com \
  pnpm tools-dev run web ...
```

## Tests

`apps/daemon/tests/origin-validation.test.ts` already mirrors the middleware inline; I extended the mirror and added 4 cases that follow the same shape as the existing `OD_WEB_PORT` block:

- ✅ allows an origin listed in `OD_EXTRA_ORIGINS`
- ✅ supports multiple comma-separated origins
- ✅ still blocks origins **not** listed (security guard intact)
- ✅ treats blank / empty entries as no-op

```
✓ tests/origin-validation.test.ts (23 tests) 112ms
  Test Files  1 passed (1)
       Tests  23 passed (23)
```

`pnpm --filter @open-design/daemon typecheck` — clean.

## Notes

- I marked this **draft** so maintainers can decide on naming (`OD_EXTRA_ORIGINS` vs e.g. `OD_TRUSTED_ORIGINS` / `OD_ALLOWED_ORIGINS`) before I polish docs.
- Happy to add a one-liner to `QUICKSTART.md` / the `OD_*` env vars section in the README architecture table once the name is settled.
- Conscious that adding *any* origin escape hatch is a security-sensitive change; the design choice here is to keep it **fully opt-in**, **per-process**, **per-origin** (no wildcards), and **purely additive** to the existing allowlist — same shape as `OD_WEB_PORT` and `OD_BIND_HOST`. Open to feedback if you'd prefer a stricter shape (e.g. require a corresponding `OD_BIND_HOST` entry, or require an explicit allowlist file rather than env var).

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -1058,15 +1058,23 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     if (webPort && webPort !== resolvedPort) ports.push(webPort);
     const schemes = ['http', 'https'];
     const loopbackHosts = ['127.0.0.1', 'localhost', '[::1]'];
-    return new Set(
-      ports.flatMap((p) => [
+    // OD_EXTRA_ORIGINS: comma-separated extra browser origins to allow.
+    // Use case: serving the web UI through a local reverse proxy that
+    // rewrites the Host (e.g. portless at <name>.localhost:1355).
+    const extraOrigins = (process.env.OD_EXTRA_ORIGINS || '')
+      .split(',')
+      .map((s) => s.trim())
+      .filter(Boolean);
+    return new Set([
+      ...ports.flatMap((p) => [
         ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
         // When bound to a specific non-loopback address (e.g. Tailscale,
         // LAN IP, or 0.0.0.0), allow browser requests from that address
         // too so the documented --host escape hatch remains usable.
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
-    );
+      ...extraOrigins,
+    ]);
   }
 
   // Routes that serve content to sandboxed iframes (Origin: null) for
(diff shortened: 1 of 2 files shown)


lefarcen (changes requested):
Hi @lonormaly!

Thanks for tackling the reverse-proxy origin challenge — the use case is clear (portless at `design.localhost:1355` is a real setup), and the design is directionally sound: opt-in, no wildcards, purely additive.

**Main concerns from code review:**

1. **Incomplete coverage** — The `/api/live-artifacts/:artifactId/preview` route skips your patched middleware and uses its own `requireLocalDaemonRequest` check, which still only trusts loopback origins. A reverse-proxied UI preserving `Origin: https://od.example.com` will 403 on live artifact preview/preflight. To fix: either extend `requireLocalDaemonRequest` to read the same `OD_EXTRA_ORIGINS`, or restructure so the global middleware covers this route too.

2. **Loose origin parsing** (L1064) — `OD_EXTRA_ORIGINS` should be parsed as proper origins, not raw strings. Accept only `http:`/`https:`, no credentials/path/query/hash, and store `url.origin`. Otherwise natural inputs like `https://od.example.com/` (trailing slash) silently fail the allowlist check. Suggested shape:
```ts
const extraOrigins = (process.env.OD_EXTRA_ORIGINS || '')
  .split(',')
  .map((s) => s.trim())
  .filter(Boolean)
  .map((s) => {
    try {
      const u = new URL(s);
      if (u.protocol !== 'http:' && u.protocol !== 'https:') return '';
      if (u.username || u.password || u.pathname !== '/' || u.search || u.hash) return '';
      return u.origin; // canonical form
    } catch { return ''; }
  })
  .filter(Boolean);
```

3. **Asymmetric host trust** (L4632) — `extraHosts` is derived with `new URL(o).host` even when `o` is not a valid browser origin. Host trust and Origin trust should come from the same validated/canonicalized entries to keep DNS-rebind protection understandable and symmetric.

4. **Test coverage gaps** (L248+) — The new tests only exercise a copied version of the global middleware. Add coverage for: the exported `isLocalSameOrigin()` path, malformed extra-origin entries (e.g., `http://od.example.com/path`, `ftp://...`, `http://user:pass@...`), `http://evil.com:<daemonPort>` still blocked, and the live-artifact preview/preflight route once fixed.

5. **Docs** — This needs operator-facing docs beyond an inline comment. Explain why `OD_BIND_HOST` alone is insufficient, require proxies to preserve `Origin`, show exact scheme/host/port examples with no trailing slash, and warn that every listed origin is trusted to drive the local daemon. A one-liner in `QUICKSTART.md` / the `OD_*` env vars table is a good start.

Happy to iterate on these — the core idea is solid!

lefarcen:
Thanks for the careful framing in the PR description — your note about marking it draft until naming was settled was the right call.

#402 has converged on the same surface (an opt-in env-var origin allowlist that widens both `buildAllowedOrigins` and `isLocalSameOrigin`'s `allowedHosts`, while keeping `requireLocalDaemonRequest` strictly loopback-only) and is now `MERGEABLE`. To avoid splitting review across three near-identical PRs, we're consolidating around #402.

Closing here in favor of that one. If once it lands you spot something it doesn't cover (e.g. a config-file shape rather than env var), a focused follow-up would be very welcome.

PR #642, opened 2026-05-06
Feat/beam react bridge

hisgarden:
use Beam for better robust event handling


lefarcen:
Hi @hisgarden! 🎉
Thanks for the contribution — introducing Beam for event handling looks like a robust architectural improvement.

— open-design team

apps/web/sidecar/beam-bridge.ts
```diff
@@ -0,0 +1,590 @@
...
+ * caller logs and skips.
+ */
+function resolveAttachmentPath(p: string, roots: string[]): string | null {
+  if (isAbsolute(p)) {
+    try {
+      if (statSync(p).isFile()) return p;
+    } catch {
+      return null;
+    }
+    return null;
```
lefarcen, line 253: **P1 Path traversal vulnerability**

The `resolveAttachmentPath` function doesn't validate that the resolved absolute path stays within the allowed `attachmentRoots`. An attacker could craft a request with `attachments: ["../../../../etc/passwd"]` and read arbitrary files.

**Fix:** After resolving, verify the result is a child of at least one root:
```ts
function resolveAttachmentPath(p: string, roots: string[]): string | null {
  // ... existing resolution logic ...
  const abs = resolve(root, p);
  try {
    if (statSync(abs).isFile()) {
      // NEW: verify abs is actually under root
      const relToRoot = relative(root, abs);
      if (!relToRoot.startsWith('..')  && !isAbsolute(relToRoot)) {
        return abs;
      }
    }
  } catch { /* try next */ }
  // ...
}
```

apps/web/sidecar/beam-bridge.ts
```diff
@@ -0,0 +1,590 @@
...
+function readToken(path: string): string {
+  return readFileSync(path, "utf8").trim();
+}
+
+// ---- Run registry -----------------------------------------------------------
+
+interface BeamRun {
+  runId: string;
+  ws: WebSocket;
+  emitter: EventEmitter;
```
lefarcen, line 110: **P1 Missing WebSocket reconnect race guard**

If the WebSocket closes and immediately reopens (network flap), there's a race where `runs.get(runId)` could return a stale entry with a dead socket, causing `handleBeamEvents` to forward to the wrong subscriber.

**Fix:** In `handleBeamEvents`, check `run.ws.readyState === WebSocket.OPEN` before forwarding each event. If closed, remove from registry and terminate SSE with error.

apps/web/sidecar/beam-bridge.ts
```diff
@@ -0,0 +1,590 @@
...
+    daemonUrl,
+    workspaceId: process.env.BEAM_WORKSPACE_ID || "open-design",
+    tokenPath:
+      process.env.BEAM_DESIGN_TOKEN_PATH ||
+      join(homedir(), ".beam-design", "auth-token"),
+    retentionMs: 5 * 60_000,
+    agentOverride: process.env.BEAM_AGENT_ID?.trim() || null,
+    modelText: text,
+    modelVision: vision,
+    attachmentRoots: roots.length > 0 ? roots : [process.cwd()],
```
lefarcen, line 96: **P2 Token file permissions not runtime-enforced**

The doc says "mode 0600" but `readToken` doesn't verify. If the file is world-readable (0644), the bridge silently uses it, leaking credentials.

**Fix:** After `readFileSync`, check `statSync(path).mode & 0o077 === 0`. Throw if too permissive:
```ts
function readToken(path: string): string {
  const stats = statSync(path);
  if ((stats.mode & 0o077) !== 0) {
    throw new Error(`${path} must be mode 0600 (chmod 600), current: ${(stats.mode & 0o777).toString(8)}`);
  }
  return readFileSync(path, "utf8").trim();
}
```

apps/web/sidecar/beam-bridge.ts
```diff
@@ -0,0 +1,590 @@
...
+): string | null {
+  if (bodyModel != null && bodyModel !== "") return bodyModel;
+  return hasImages ? config.modelVision : config.modelText;
+}
+
+async function readJsonBody(req: IncomingMessage): Promise<ChatRequestLike> {
+  const chunks: Buffer[] = [];
+  for await (const chunk of req) chunks.push(chunk as Buffer);
+  if (chunks.length === 0) return {};
+  try {
```
lefarcen, line 326: **P2 Model selection doesn't handle empty-string edge case**

`pickModel` checks `bodyModel != null && bodyModel !== ""`, but the UI might send `{model: ""}` (empty string) which passes the null check but shouldn't override the tier logic.

**Fix:** Normalize to null:
```ts
function pickModel(
  config: BeamBridgeConfig,
  bodyModel: string | null | undefined,
  hasImages: boolean,
): string | null {
  const explicit = bodyModel?.trim() || null;
  if (explicit != null) return explicit;
  return hasImages ? config.modelVision : config.modelText;
}
```

apps/web/sidecar/beam-bridge.ts
```diff
@@ -0,0 +1,590 @@
...
+  const out: BridgeImage[] = [];
+  for (const raw of paths) {
+    const ext = extname(raw).toLowerCase();
+    const mime = IMAGE_EXT_TO_MIME[ext];
+    if (mime == null) continue;
+    const abs = resolveAttachmentPath(raw, roots);
+    if (abs == null) {
+      console.warn(`[beam-bridge] attachment not found, skipping: ${raw}`);
+      continue;
+    }
```
lefarcen, line 289: **P2 Missing payload size validation**

`readImageAttachments` caps individual images at `maxImageBytes`, but multiple small images can still blow the channel payload. Phoenix default is 4MB message size; base64 encoding inflates by ~33%.

**Fix:** After building the images array, check total size:
```ts
const totalBytes = images.reduce((sum, img) => sum + img.base64.length, 0);
if (totalBytes > 3 * 1024 * 1024) { // 3MB leaves headroom for prompt + metadata
  throw new Error(`Total image payload ${totalBytes} bytes exceeds 3MB limit (reduce attachment count or resolution)`);
}
```

scripts/README-beam-bridge.md
```diff
@@ -0,0 +1,148 @@
...
+
+# Open http://127.0.0.1:17573, start a chat — the run goes through BEAM.
+```
+
+Requires `pnpm` ≥ 10.33.2.
+
+### Forcing DeepInfra (or any other BEAM agent) regardless of UI pick
+
+The JS daemon's `/api/agents` only advertises CLI agents (`claude` / `codex` / `gemini` / etc.), so the React UI never offers `deepinfra` as a picker option. Two env knobs let the bridge override at the seam:
+
```
lefarcen, line 57: **P2 Undocumented pnpm version assumption**

The doc says "Requires `pnpm` ≥ 10.33.2" but doesn't explain *why* or what breaks on older versions. New contributors won't know if they can skip the upgrade.

**Fix:** Add one sentence:
```markdown
Requires `pnpm` ≥ 10.33.2 (older versions fail to resolve `@open-design/contracts` peer deps in the sidecar workspace).
```

apps/daemon/src/media.ts
```diff
@@ -1036,6 +1041,138 @@ async function renderGrokImage(ctx, credentials) {
...
+  if (Array.isArray(data?.images) && data.images.length > 0) entry = data.images[0];
+  else if (Array.isArray(data?.output) && data.output.length > 0) entry = data.output[0];
+  else if (typeof data?.image === 'string') entry = data.image;
+  else if (typeof data?.image_url === 'string') entry = data.image_url;
+  if (typeof entry !== 'string' || !entry) {
+    throw new Error(`deepinfra image: response had no images/output/image field`);
+  }
+  let bytes;
+  if (entry.startsWith('data:')) {
+    const comma = entry.indexOf(',');
```
lefarcen, line 1158: **P3 Inconsistent error prefix**

Most errors use `deepinfra image:` but line 1152 uses `deepinfra image` (no colon). Small but visible to users who grep logs.

**Fix:** Normalize to `deepinfra image:` everywhere.

lefarcen (changes requested):
Hi @hisgarden! 👋 Thanks for this substantial architecture work — the Beam bridge and DeepInfra integration look promising overall.

I've reviewed both the code correctness (Lens A) and reasoning completeness (Lens B). Found 2 critical must-fix issues (security + error handling), plus 5 actionable improvements. Details inline below.

**Summary of findings:**
- P1: Path traversal vulnerability in attachment resolution
- P1: Missing error handling for WebSocket reconnect race
- P2: Token file permissions not enforced at runtime
- P2: Model selection logic doesn't handle empty-string edge case
- P2: Missing validation for oversized channel payloads
- P2: README assumptions about pnpm version undocumented
- P3: Inconsistent error message formatting

Once these are addressed, this will be ready to ship. The Beam bridge design is solid — the seam at ChatSseEvent is the right abstraction.

apps/web/sidecar/server.ts
```diff
@@ -508,6 +516,66 @@ function attachParentMonitor(stop: () => Promise<void>): void {
   timer.unref();
 }
 
+export async function startWebSidecar(runtime: SidecarRuntimeContext<SidecarStamp>): Promise<WebSidecarHandle> {
+  const dir = resolveWebRoot();
+  const app = createNextServer({ dev: process.env.OD_WEB_PROD !== "1" && runtime.mode === "dev", dir });
```
mrcfps, line 521: This added `startWebSidecar` block appears to be a stale duplicate of the real exported `startWebSidecar` that still exists later in this file. As written, this block never returns or closes before the following helper declarations, it calls `createNextServer` even though this module only defines `createNextApp`, and it later calls `proxyToDaemon`, which is not defined. Because this is in the sidecar entrypoint, the web package cannot compile/start at this PR head. Please remove this duplicate block and integrate the Beam route interception into the existing `createDaemonProxyHandler` / `startRegularNextSidecar` path instead, preserving the standalone sidecar path as well.

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -1594,7 +1594,7 @@ function MediaProjectOptions(props:
 
 function supportedModels(surface: 'image' | 'video' | 'audio', models: MediaModel[]): MediaModel[] {
   const supportedProviders: Record<'image' | 'video' | 'audio', Set<string>> = {
-    image: new Set(['openai', 'volcengine', 'grok']),
+    image: new Set(['openai', 'volcengine', 'grok', 'deepinfra']),
```
mrcfps, line 1597: Adding `deepinfra` to the image provider allow-list makes every DeepInfra image model selectable in the normal New Project image flow, including `wan-2.7-image-edit` and `qwen-image-edit`, whose model definitions are `caps: ['i2i']`. That UI path only captures prompt/aspect/style metadata and no reference image, while the daemon's DeepInfra renderer explicitly throws for i2i models without `ctx.imageRef`, so users can now choose a model from the picker that the primary image flow cannot run. Please either filter image-project models here to prompt-capable models (`model.caps.includes('t2i')`) or add a reference-image selection/upload step before exposing i2i-only DeepInfra models in this picker.

mrcfps (changes requested):
@hisgarden Thanks for the substantial work here — the Beam bridge direction and the DeepInfra model expansion are valuable improvements. I found two merge-blocking issues in the current head: the web sidecar file has a duplicated/broken `startWebSidecar` insertion that prevents the web package from compiling, and the image picker now exposes DeepInfra i2i-only models through a prompt-only creation path that the daemon rejects.


PerishCode:
@hisgarden — quick bump on the beam react bridge: it's been about 5 days since the requested-changes review and the PR's waiting on your follow-up. Still on it? If the review thread needs unpacking I'm happy to dig in; if it's stalled, a quick status note here helps reviewers know whether to hold the slot.

(Via maintainer `awaiting-author-response-24h` queue scan.)

PerishCode:
@hisgarden — quick check-in: this has been waiting on your follow-up for about 7 days. Are you still planning to land it? If there is a specific blocker on the Beam bridge work, point us at it; if priorities changed, just say so and we can close it cleanly.

lefarcen (commented):
Hey @hisgarden — one non-code housekeeping request if you pick this back up: the PR body is currently just one sentence. Could you add the missing product/review context: `Why`, `What users will see`, `Surface area`, and `Validation`? For this PR, the surface area should call out the Beam bridge plus DeepInfra image/model-provider changes, and Validation should name what you ran for those paths so product and reviewers can restart cleanly.

lefarcen:
This PR has been quiet for 12 days. Could you share a screenshot or short clip of what users see when the beam react bridge is active? Or update the PR body with a brief summary of the current state if it's still in flight?

PR #681, opened 2026-05-06
feat: add Orbit activity summaries

mrcfps:
## Summary
- Add Orbit settings for scheduled/manual daily connector activity summaries.
- Run Orbit through the agent runtime with the live-artifact skill and OD connector CLI access.
- Navigate manual runs to the Orbit project so agent output can be followed in conversation view.

## Validation
- pnpm guard


lefarcen:
Hi @mrcfps! 👋

Thanks for opening this draft PR. When you're ready for review, mark it as 'Ready for review' (the button below the PR description).

Until then I'll stay out of your way to avoid noise during your iteration.

— open-design team

mrcfps (commented):
@mrcfps thanks for pushing the Orbit summary work forward — the settings flow, template staging, and connector fast-path changes are moving the feature in a good direction. I found one scheduler lifecycle issue that can stop daily automation after an early start failure; because the inline review submission was rejected by GitHub, I’m including the exact fallback location here. 🙏

**apps/daemon/src/orbit.ts:338 (`this.start('scheduled').catch(...)`)** — When the scheduled timer fires, this catch only logs failures from `start('scheduled')`. If `startRun()` rejects before `this.inflight` is assigned — for example while resolving the template or when the run handler throws before creating the project/run — the `inflight` cleanup at lines 312-317 never runs, so `reschedule()` is never called again. That leaves the daily Orbit timer exhausted after one early start failure, and scheduled summaries silently stop until the daemon restarts or the user saves config again. Please reschedule scheduled start failures in this path (for example, call `this.reschedule()` in this catch/finally when the trigger was scheduled and no inflight run was created, or move the reschedule responsibility into a `start()` failure path that preserves manual-run behavior).


mrcfps (commented):
@mrcfps thanks for the continued Orbit summary work — I reviewed the daemon scheduling path, connector-tool fast path, template staging, settings UI, and the added coverage. I found one scheduler lifecycle follow-up that can still stop daily automation after an early start failure; GitHub rejected the inline review submission, so I’m including the exact fallback location here. 🙏

**apps/daemon/src/orbit.ts:341 (`this.start('scheduled').catch(...)`)** — When the scheduled timer fires, this catch only logs failures from `start('scheduled')`. If `startRun()` rejects before `this.inflight` is assigned — for example because no agent is currently configured/available, or the run handler throws before creating the project/run — the cleanup that calls `reschedule()` on lines 315-320 never runs. That exhausts the only scheduled timer, so daily Orbit summaries silently stop until the daemon restarts or someone saves config again. Please reschedule scheduled start failures in this path (for example, call `this.reschedule()` from this catch/finally when no inflight run was created, or move the scheduled-failure reschedule into `start()` while preserving manual-run error behavior), and add a small OrbitService test that a scheduled early-start rejection arms the next timer.


apps/daemon/src/orbit.ts
```diff
@@ -0,0 +1,369 @@
...
+    this.nextRunAtValue = null;
+  }
+
+  private reschedule(): void {
+    this.stop();
+    if (!this.config.enabled) return;
+    const next = nextDailyRunAt(this.config.time);
+    this.nextRunAtValue = next;
+    this.timer = setTimeout(() => {
+      void this.start('scheduled').catch((error) => {
```
mrcfps, line 420: When this scheduled timer fires, the catch only logs failures from `this.start('scheduled')`. If `startRun()` rejects before `this.inflight` is assigned — for example because the configured agent is no longer available, template resolution fails, or the run handler throws before creating the project/run — the cleanup at lines 338-343 never runs, so `reschedule()` is never called. That consumes the one-shot daily timer and Orbit silently stops running until the daemon restarts or someone saves config again. Please reschedule scheduled start failures in this catch/finally path when no inflight run was created, and add an OrbitService test that an early scheduled-start rejection arms the next timer.

mrcfps (commented):
@mrcfps thanks for continuing to tighten up the Orbit activity-summary flow — I reviewed the current daemon scheduling path, connector discovery changes, settings integration, and updated tests. I found one scheduler lifecycle follow-up that can leave daily automation stopped after an early start failure. 🙏


apps/web/src/index.css
```diff
@@ -3424,6 +4401,223 @@ button.connector-action.is-loading {
...
+@media (max-width: 540px) {
+  .connectors-panel-embedded .toolbar-right {
+    flex-direction: column;
+    align-items: stretch;
+  }
+  .connectors-provider-tabs {
+    align-self: flex-start;
+  }
+}
 .connector-gate {
```
mrcfps, line 5650: The Settings-embedded connector gate needs to own the top hit-testing layer. In the current CI run, the gate button is visible, but Playwright cannot click it because the settings content / disabled connector search input / connectors panel intercept the pointer after this absolute overlay is scrolled into view. That breaks the new Settings → Connectors unlock flow for users who need the Composio key CTA, not just the test. Please make the gate overlay a stable top layer inside `.connector-grid-wrap` (for example, give `.connector-gate` an explicit `z-index` above the toolbar/grid within the modal stacking context and keep its card fully inside the clipped masked viewport, or otherwise adjust the embedded layout so `connector-gate-action` receives pointer events), then rerun the e2e case that clicks the gate CTA.

Evidence: the changed embedded masked-grid styles cap/clip the catalog, while `.connector-gate` remains an absolutely positioned sibling with no explicit stacking priority; CI fails `live artifact empty connector CTA opens the gated connector setup path` with the visible `connector-gate-action` click intercepted by `.settings-content`, the disabled search input, and `.connectors-panel-embedded`.

e2e/ui/entry-configuration-flows.test.ts
```diff
@@ -148,37 +148,49 @@ test('live artifact empty connector CTA opens the gated connector setup path', a
...
   await routeConnectors(page, CONNECTORS);
 
   await page.goto('/');
-  await page.getByTestId('entry-tab-connectors').click();
-  await expect(page.getByTestId('connector-grid-wrap')).toBeVisible();
+  // Connector cards + search now live under Settings → Connectors. Open the
+  // settings dialog via the entry sidebar's "Configure execution mode" pill
+  // and switch to the Connectors section before exercising the
+  // search/empty/details flow.
+  await page.getByRole('button', { name: 'Configure execution mode' }).click();
```
mrcfps, line 199: This updated test is waiting for an accessible button name that the app does not expose. The entry footer button has `title={t('settings.envConfigure')}`, but its accessible name comes from its visible contents (`Local CLI · ...`), so `getByRole('button', { name: 'Configure execution mode' })` times out in CI. Because this suite now gates the relocated Settings → Connectors flow, the PR currently cannot pass validation. Please either add an `aria-label={t('settings.envConfigure')}`/stable test id to the settings-opening control in `EntryView`, or update the test to use an accessible selector that actually exists (for example the `Open settings` button/menu path), then keep the assertion scoped to the Connectors section.

Evidence: the failing CI log for this exact changed line times out waiting for `getByRole('button', { name: 'Configure execution mode' })`, and `EntryView` only uses that string as a `title` on a button with other visible text.

apps/daemon/src/orbit.ts
mrcfps, line 420: Scheduled Orbit starts still only log early `start('scheduled')` failures. If `startRun()` rejects before `this.inflight` is assigned — for example while the run handler is unavailable or throws before returning `{ projectId, agentRunId, completion }` — the `finally` at lines 338-342 never runs, so the one-shot daily timer is consumed and `reschedule()` is not called. That leaves scheduled summaries stopped until a restart or config save, which breaks the daily automation path this PR adds. Please reschedule from this scheduled-start catch/finally when no inflight run was created (while preserving manual-run errors), and add an `OrbitService` test that an early scheduled-start rejection arms the next timer.

Evidence: `reschedule()` installs a single `setTimeout`, `start()` stores only `this.startRun(trigger).finally(() => { this.starting = null; })`, and the only reschedule on completion lives inside the `this.inflight` promise that is assigned after `runHandler` succeeds.

mrcfps (commented):
@mrcfps thanks for continuing to push the Orbit activity summaries forward — I reviewed the current daemon scheduling path, connector settings relocation, compact catalog UI, updated tests, and the failing CI signal. I found a few issues that should be addressed; GitHub does not allow me to request changes on my own pull request, so I’m leaving them as review comments. 🙏


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1300,12 +1344,666 @@ function ComposioSection({
...
             : apiKeyConfigured
-              ? 'Unsaved changes — click Save to store this key in the local daemon.'
-              : 'Keys are stored locally in the daemon and never sent through environment variables.'}
+              ? 'Unsaved changes — click Save to store this key in the local daemon and unlock the catalog below.'
+              : 'Add a key to unlock the catalog below. Keys are stored locally in the daemon and never sent through environment variables.'}
         </span>
       </label>
+
+      <ConnectorsBrowser
+        composioConfigured={apiKeyConfigured}
```
mrcfps: `composioConfigured` is true as soon as the user types an unsaved key (`apiKeyConfigured` includes `hasPendingEdit`), so the embedded catalog unlocks and its Connect buttons become active before that key has been written to the daemon. Those actions still call the daemon with the old/no saved key, which means the first connector setup path can fail right after the UI says the catalog is unlocked. Please keep the browser gated until `composio.apiKeyConfigured` reflects a saved daemon key, or pass a pending/saving state into `ConnectorsBrowser` that shows the catalog but disables connector actions until Save succeeds.

Evidence: the changed help text says unsaved edits need Save to unlock the catalog, but this new prop passes the broader `apiKeyConfigured` value (pending edit OR saved key) directly into the catalog gate.

mrcfps (commented):
@mrcfps thanks for continuing to refine Orbit and the connector settings flow — I reviewed the current daemon scheduling, Orbit run UI, connector catalog/CLI paths, and the updated tests. I found one additional non-blocking UX correctness issue in the Settings → Connectors path; I’m not repeating the existing Codex threads for the Orbit status refresh and timer unref items. 🙏


mrcfps (commented):
@mrcfps thanks for the continued Orbit activity-summary work — I reviewed the daemon Orbit scheduler/run startup path, connector tool discovery and curation flow, Settings → Connectors/Orbit UI changes, and the added coverage around scheduled failures and manual runs. The prior scheduler, status-refresh, timer, and saved-key gating feedback appears addressed in this head, and I did not find any new actionable correctness or maintainability issues. Nice work tightening this up 🙏


apps/web/src/components/FileViewer.tsx
```diff
@@ -565,7 +565,7 @@ export function LiveArtifactViewer({
             <iframe
               data-testid="live-artifact-preview-frame"
               title={liveArtifact.title}
-              sandbox="allow-scripts"
+              sandbox="allow-scripts allow-popups allow-popups-to-escape-sandbox"
```
mrcfps: This preview iframe now gives agent-generated live artifact content both `allow-popups` and `allow-popups-to-escape-sandbox`. Because these artifacts are generated from prompts/connector data and run with `allow-scripts`, a malicious or compromised artifact can open a new unsandboxed top-level page instead of keeping the popup inside the iframe sandbox. That weakens the isolation boundary for untrusted preview content and can expose users to unsandboxed script/phishing flows that the sandbox previously contained.

Please keep popups sandboxed here unless there is a concrete product requirement for escaped popups — for example, use `sandbox="allow-scripts allow-popups"` for external links, or route artifact links through a controlled `target=_blank`/`rel=noopener` path outside the generated iframe if escape is required.

Evidence: the only changed sandbox policy on this line is the added `allow-popups-to-escape-sandbox`, and this viewer renders live artifacts whose HTML is produced by the agent/runtime rather than trusted application code.

mrcfps (commented):
@mrcfps thanks for continuing to polish the Orbit activity-summary flow — I reviewed the current daemon scheduler/run path, connector curation and CLI changes, Settings/Orbit UI updates, live artifact viewer changes, and the added coverage. I found one small safety follow-up around live artifact preview sandboxing; the Orbit-specific feedback from earlier rounds looks addressed. 🙏


mrcfps (commented):
@mrcfps thanks for continuing to tighten up Orbit activity summaries — I reviewed the current daemon scheduler/run lifecycle, connector curation and fast-path execution changes, Settings → Connectors/Orbit UI, live artifact preview policy, config/contracts updates, and the added daemon/web/e2e coverage. The earlier scheduler, status refresh, saved-key gating, default-template, and sandboxing feedback all look addressed in this head, and I did not find any new actionable correctness, safety, or maintainability issues. Nice work getting this flow into a solid shape 🙏


mrcfps (commented):
@mrcfps thanks for continuing to refine the Orbit activity-summary flow — I reviewed the current daemon scheduler/run lifecycle, connector curation and fast-path execution, Settings → Connectors/Orbit UI, template/default handling, live artifact preview sandboxing, and the updated daemon/web/e2e coverage. The previous scheduler, status refresh, saved-key gating, default-template, and sandboxing feedback looks addressed in this head, and I did not find any new actionable correctness, safety, or maintainability issues. Nice work tightening this up 🙏


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1078,7 +1093,21 @@ export function SettingsDialog({
...
+          {activeSection === 'orbit' ? (
+            <OrbitSection
+              cfg={cfg}
+              setCfg={setCfg}
+              onLeaveForOrbitProject={() => {
+                // Persist any in-flight Orbit edits (toggle / time) before
+                // navigating away so they aren't silently lost. onSave also
+                // closes the dialog, so the user lands directly on the
+                // /projects/orbit view where the agent run streams in.
+                onSave(cfg);
```
mrcfps: This save path still uses the original `cfg` after `triggerNow()` has started Orbit with `configForManualOrbitRun(cfg)`. For legacy configs where `orbit.templateSkillId` is `null`, the UI displays `orbit-general` and line 1525 correctly syncs that effective value before `POST /api/orbit/run`, but this callback immediately calls the app-level `onSave(cfg)`, whose implementation writes the unnormalized `cfg` back through `syncConfigToDaemon`. That can undo the effective template in local/daemon config right after the run starts, so future scheduled/manual runs fall back to no template even though the user just launched with the displayed default. Please compute the effective run config once and pass that same config to both `persistConfigAndRunOrbit` and the close/save callback (for example, make `onLeaveForOrbitProject` accept an `AppConfig`, then call `onSave(runConfig)`).

mrcfps (commented):
@mrcfps thanks for the continued Orbit activity-summary work — I reviewed the current daemon Orbit scheduler/run lifecycle, connector curation and fast-path execution, Settings → Connectors/Orbit UI, live artifact preview policy, config/contracts updates, and the added daemon/web/e2e coverage. I found one small follow-up around preserving the displayed default template after a manual run starts. 🙏


mrcfps (commented):
@mrcfps thanks for continuing to tighten up the Orbit activity-summary flow — I reviewed the current daemon Orbit scheduler/run lifecycle, connector discovery and curated tool paths, Settings → Connectors/Orbit UI, live artifact preview sandboxing, config/contracts updates, and the added daemon/web/e2e coverage. The prior scheduler, status refresh, saved-key gating, default-template persistence, and sandboxing feedback all look addressed in this head, and I did not find any new actionable correctness, safety, or maintainability issues. Nice work getting this flow into a solid shape 🙏


apps/daemon/src/tools/connectors.ts
```diff
@@ -48,16 +48,49 @@ function isAgentPreviewListableTool(definition: ConnectorCatalogDefinition, tool
...
+  const fastDefinitions = service.listFastDefinitions();
+  const fastDefinitionsById = new Map(fastDefinitions.map((definition) => [definition.id, definition]));
+  const connectedStatusIds = Object.entries(service.listConnectorStatuses())
+    .filter(([, status]) => status.status === 'connected')
+    .map(([connectorId]) => connectorId);
+  const hasConnectedConnectorNeedingDiscovery = connectedStatusIds.some((connectorId) => {
+    const fastDefinition = fastDefinitionsById.get(connectorId);
+    return !fastDefinition || fastDefinition.tools.length === 0;
+  });
+  const definitions = hasConnectedConnectorNeedingDiscovery ? await service.listDefinitions() : fastDefinitions;
```
mrcfps, line 73: In `listConnectorTools`, the decision to stay on `fastDefinitions` only considers `service.listConnectorStatuses()`, but that status snapshot is built from in-memory statuses plus known fast definitions and does not enumerate credentials for connector ids that are absent from the fast catalog after a daemon restart. If a user has a valid persisted Composio credential for a connected toolkit that is not in the static/persisted fast cache yet, `connectedStatusIds` stays empty here, `service.listDefinitions()` is skipped, and Orbit/agent `tools connectors list --use-case personal_daily_digest` never exposes that usable connector.

This matters because the PR is making Orbit depend on this fast path for unattended daily summaries; missing a connected data source after restart can make summaries incomplete even though execution would succeed once the full definition is discovered. Please force live discovery when Composio is configured but the fast cache has not been refreshed for the current credential set, or have `listConnectorStatuses()`/the credential store expose persisted connected connector ids so this `hasConnectedConnectorNeedingDiscovery` branch can reliably fall back to `service.listDefinitions()`.

Evidence: the changed lines derive `connectedStatusIds` from `listConnectorStatuses()` and only call `service.listDefinitions()` when one of those ids is missing from `fastDefinitionsById` or has no tools; no credential-store-backed ids outside the fast definitions can participate in that check.

mrcfps (commented):
@mrcfps thanks for continuing to iterate on the Orbit activity-summary flow — I reviewed the current daemon scheduler/run path, connector discovery/CLI changes, Settings → Connectors/Orbit UI, live artifact preview handling, and the added daemon/web/e2e coverage. I found one non-blocking follow-up in the agent connector discovery fast path so Orbit does not miss valid connected tools after restart. 🙏


mrcfps:
@mrcfps I'm holding off on generating review comments for nexu-io/open-design#681 because this pull request has merge conflicts right now.

mrcfps (commented):
@mrcfps thanks for continuing to tighten up the Orbit activity-summary flow — I reviewed the current daemon scheduler/run lifecycle, connector catalog and curated tool paths, Settings → Connectors/Orbit UI, live artifact preview handling, config/contracts updates, and the added daemon/web/e2e coverage. The prior scheduler, status refresh, saved-key gating, default-template persistence, sandboxing, and connector fast-path feedback all look addressed in this head, and I did not find any new actionable correctness, safety, or maintainability issues. Nice work getting this flow into a solid shape 🙏


apps/daemon/src/server.ts
```diff
@@ -1497,6 +1499,8 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
   };
   configureConnectorCredentialStore(new FileConnectorCredentialStore(RUNTIME_DATA_DIR));
   configureComposioConfigStore(RUNTIME_DATA_DIR);
+  composioConnectorProvider.configureCatalogCache(RUNTIME_DATA_DIR);
+  composioConnectorProvider.startCatalogRefreshLoop();
```
mrcfps, line 1594: This starts the process-global Composio catalog refresh loop for every `startServer()` call, but there is still no matching cleanup registered on the returned HTTP server. That leaves the loop running after `server.close()` in embedded/test/sidecar flows, so a daemon instance that callers believe is shut down can continue polling Composio and writing the catalog cache under `RUNTIME_DATA_DIR`.

This matters for correctness and isolation because `startServer({ returnServer: true })` is used by tests and wrappers that rely on `server.close()` to tear down all daemon background work. Please attach a close handler here (for example `server.once('close', ...)`) that calls `composioConnectorProvider.stopCatalogRefreshLoop()` and `orbitService.stop()`, and make sure `stopCatalogRefreshLoop()` also clears any one-shot refresh scheduled from cache loading so no background refresh survives shutdown.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1796,6 +1841,680 @@ function ComposioSection({
                 : 'Keys are stored locally in the daemon and never sent through environment variables.'}
         </span>
       </label>
+
+      <ConnectorsBrowser
+        composioConfigured={hasSavedKey}
```
mrcfps: `ConnectorsBrowser` now gates the catalog on `hasSavedKey`, but the Settings dialog keeps its local `cfg` state unchanged after the user saves the Composio key because the footer calls `onSave(cfg, false)` for the connectors section. `App.handleConfigSave` normalizes the parent config to `{ apiKey: '', apiKeyConfigured: true }`, yet this component does not sync that result back into its local `cfg`, so `hasSavedKey` remains false and the embedded catalog can stay masked until the dialog is closed and reopened.

That breaks the main first-run setup path this PR is polishing: paste key → Save → browse/connect tools in the same Settings surface. Please have the save path update this local section state after a successful Composio save (or have `onSave` return the normalized persisted config and call `setCfg` here), then refresh connector discovery/status so the catalog unlocks immediately.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1796,6 +1841,680 @@ function ComposioSection({
...
+
+interface OrbitRunStartResponse {
+  projectId: string;
+  agentRunId: string;
+}
+
+export async function persistConfigAndRunOrbit(
+  config: AppConfig,
+): Promise<OrbitRunStartResponse> {
+  await syncConfigToDaemon(config, { throwOnError: true });
```
mrcfps, line 2462: The manual Orbit run path only syncs the generic app-config payload before POSTing `/api/orbit/run`. It does not run the full Settings save pipeline that persists daemon-backed fields such as `cfg.composio.apiKey`; that full `onSave(runConfig)` call happens later, after the run has already been started.

This means a user can paste or replace a Composio key in the same Settings dialog, switch to Orbit, and click **Run & open**, but the daemon starts Orbit with stale connector credentials and may fail or omit connector data even though the UI draft contains the new key. Please make the manual run await the same successful save path used by the footer before calling `/api/orbit/run`, or explicitly persist Composio/media daemon-backed settings inside `persistConfigAndRunOrbit` before starting the run.

mrcfps (commented):
@mrcfps thanks for continuing to tighten up Orbit activity summaries — I reviewed the current daemon lifecycle changes, connector catalog/settings flow, Orbit manual/scheduled run path, and the updated daemon/web/e2e coverage. I found a few non-blocking follow-ups around shutdown cleanup and unsaved settings synchronization; the earlier Orbit scheduler/template/sandbox feedback otherwise looks addressed. 🙏


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1796,6 +1853,685 @@ function ComposioSection({
...
+}
+
+export async function persistConfigAndRunOrbit(
+  config: AppConfig,
+): Promise<OrbitRunStartResponse> {
+  if (config.composio !== undefined) {
+    const composioSaved = await syncComposioConfigToDaemon(config.composio);
+    if (!composioSaved) throw new Error('Composio config save failed');
+  }
+  await syncMediaProvidersToDaemon(config.mediaProviders, { force: true, throwOnError: true });
```
lefarcen: 🔴 **P1** — This manual Orbit run path can wipe daemon-stored media credentials. `loadConfig()` defaults `mediaProviders` to `{}`, and the web app does not hydrate media credentials from `/api/media/config`; calling `syncMediaProvidersToDaemon(..., { force: true })` with an empty local map bypasses the daemon's accidental-wipe guard and replaces any existing media config with `{}` just because the user clicked Run & open.

apps/web/src/components/ConnectorsBrowser.tsx
```diff
@@ -0,0 +1,761 @@
...
+  // configured. We also honor the local config.composio flag so the mask
+  // appears immediately when the key is cleared, before the next list fetch.
+  const anyComposioAuthConfigured = useMemo(
+    () =>
+      connectors.some(
+        (connector) => connector.auth?.provider === 'composio' && connector.auth.configured,
+      ),
+    [connectors],
+  );
+  const needsComposioKey = !composioConfigured && !anyComposioAuthConfigured;
```
lefarcen: **P2** — The stale auth masking issue is not fixed. If the prior connector list still has a Composio connector with `auth.configured === true`, clearing the key makes `composioConfigured` false but `anyComposioAuthConfigured` remains true, so this `&&` keeps `needsComposioKey` false and the grid stays unmasked until a later refresh.

apps/web/src/state/config.ts
```diff
@@ -204,6 +216,13 @@ function normalizeNotifications(
   return { ...DEFAULT_NOTIFICATIONS, ...(input ?? {}) };
 }
 
+function normalizeOrbit(input: Partial<OrbitConfig> | undefined): OrbitConfig {
+  const time = typeof input?.time === 'string' && /^\d{2}:\d{2}$/.test(input.time)
```
lefarcen: **P2** — Web-side Orbit normalization still accepts out-of-range times like `99:99` because it only checks the `HH:mm` shape. The daemon now clamps invalid values, but local config/UI normalization can still preserve and submit invalid values instead of falling back to `DEFAULT_ORBIT.time`.

apps/daemon/src/orbit.ts
```diff
@@ -0,0 +1,383 @@
...
+    this.timer = null;
+    this.nextRunAtValue = null;
+  }
+
+  private reschedule(): void {
+    this.stop();
+    if (!this.config.enabled) return;
+    const next = nextDailyRunAt(this.config.time);
+    this.nextRunAtValue = next;
+    this.timer = setTimeout(() => {
```
lefarcen, line 417: **P2** — The scheduled run still reports the fired timestamp as `nextRunAt` while the run is in flight. The timer callback starts the scheduled run without clearing or advancing `nextRunAtValue`; polling the status during a long run will keep showing the past run time as the next run until completion triggers `reschedule()`.

lefarcen (changes requested):
Thanks for continuing to iterate on Orbit — I reviewed the new connector cleanup path and found a critical media-credential wipe regression plus three prior P2s that are still reproducible. 🙏

mrcfps (commented):
@mrcfps thanks for continuing to iterate on the Orbit activity-summary flow — I reviewed the current head across the daemon lifecycle, connector catalog/settings flow, Orbit manual/scheduled run path, Settings UI, and updated coverage. I did not find a separate new issue beyond the currently unresolved review threads, and I’m not duplicating those inline comments here; the existing `apps/web/src/components/SettingsDialog.tsx:1893` media-credential wipe thread still looks reproducible and is already covered by the current changes-requested review from another maintainer. The existing unresolved `ConnectorsBrowser` auth-gating, web Orbit time normalization, and scheduled-run `nextRunAt` status threads also still look reproducible and worth addressing. Thanks for pushing this forward — the rest of the cleanup work here is moving in the right direction. 🙏


lefarcen (approved):
Thanks for continuing to iterate on Orbit — the coding agent review found 4 P2 maintainability items on this head, all non-blocking. Dismissing my prior changes-requested review. 🙏

lefarcen (changes requested):
Thanks for the continued Orbit work — the media-credential wipe, connector auth stale-mask, and daemon lifecycle cleanup are all fixed. 🙏

Two prior P2s are still reproducible:

**apps/web/src/state/config.ts (`normalizeOrbit` function)** — Web-side Orbit time normalization accepts out-of-range times like `99:99`. The regex `/^\d{2}:\d{2}$/` checks shape but not range, so users can save invalid times that the daemon scheduler will misinterpret or fail on. Please add range validation (e.g., clamp hours to 0-23, minutes to 0-59, or reject and fall back to `DEFAULT_ORBIT.time` when parsed values are out of range).

**apps/daemon/src/orbit.ts (timer callback, ~line 335)** — The scheduled run still reports the fired timestamp as `nextRunAt` while the run is in flight. The timer callback starts the scheduled run with `this.start('scheduled')`, but `this.nextRunAtValue` remains set to the old fired time until the run completes and `reschedule()` is called again. That makes the daemon status API show a stale "next run" timestamp during long runs. Please clear or advance `this.nextRunAtValue` inside the timer callback before starting the run (e.g., set it to `null` or compute the next daily run time immediately), so the status endpoint accurately reflects that the scheduled time has passed and the run is in progress.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1765,37 +2116,1011 @@ function ComposioSection({
...
+  };
+
+  // Proportional widths for the run-result meter. We avoid showing 0-width
+  // segments by falling back to a tiny sliver when a category has hits but
+  // rounds to 0% — the visual "something happened here" cue matters more
+  // than exact proportion at low counts.
+  const total = lastRun
+    ? Math.max(
+        lastRun.connectorsSucceeded + lastRun.connectorsSkipped + lastRun.connectorsFailed,
+        1,
```
lefarcen, line 2708: P2: Clearing the last media provider credential no longer reaches the daemon. When all credential fields are empty, this line deletes the provider from the local map (leaving `mediaProviders: {}`). Then `App.tsx:269-273` and `SettingsDialog.tsx:1894-1896` skip the media sync call for empty maps, so the daemon keeps the old credential.

This fixes the accidental empty force-sync wipe from the prior review, but breaks the intentional "clear credential" use case. Please either:
- Send a deletion signal to the daemon when a provider is removed from the map (e.g., check `prev` vs `current` in `handleMediaProvidersChange` and explicitly call daemon delete for removed keys), OR
- Always sync `mediaProviders` regardless of emptiness, and have the daemon interpret `{}` as "clear all" (document this contract).

Evidence: lines 2570-2572 delete provider; App.tsx:269-273 + SettingsDialog.tsx:1894-1896 both gate `!isEmpty(cfg.mediaProviders)` before syncing.

lefarcen (changes requested):
Hey @mrcfps! 👋 I re-reviewed the current head (78e105e9) after your "fix: address Orbit review regressions" commit. Three of the four prior P2s are now addressed — nice work on the stale auth masking, time validation, and nextRunAt display. However, one P2 regression remains that blocks the intentional "clear credential" flow. See inline below.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1677,20 +1894,6 @@ export function SettingsDialog({
           ) : null}
```
lefarcen, line 2017: P2 — This path now avoids the accidental empty-map wipe from stale local state, but it also blocks an intentional "clear the last media provider" save from reaching the daemon. When a user explicitly deletes the last configured provider, local `mediaProviders` becomes `{}`, so `hasAnyConfiguredProvider({})` returns false and `syncMediaProvidersToDaemon` is skipped. The daemon keeps the old credential, so the UI says cleared but the saved daemon secret remains active.

Please add a separate flag or timestamp that tracks "user just made a change" vs "stale empty local state on Settings mount", then sync media providers to the daemon when the user explicitly cleared the last one. For example, track `mediaProvidersModified` in component state, set it on provider add/delete/edit, and call `syncMediaProvidersToDaemon` when `mediaProvidersModified` is true regardless of `hasAnyConfiguredProvider`.

lefarcen (changes requested):
@mrcfps thanks for the continued Orbit cleanup work — I reviewed the current head (78e105e9) across the Orbit lifecycle fixes, connector auth gating, time validation, and status-refresh changes. Three of the four previously reported P2 items look addressed (connector auth masking, time validation range, scheduled nextRunAt stale display), but I found one regression in the media-provider clear path introduced by the fix. See inline below. 🙏

apps/web/src/App.tsx
```diff
@@ -244,30 +265,61 @@ export function App() {
...
+   * (strip the secret, store apiKeyConfigured + apiKeyTail) and feed
+   * it back into local state so the saved-key badge appears.
+   */
+  const handleConfigPersistComposioKey = useCallback(
+    async (composio: AppConfig['composio']) => {
+      // Send the raw pending edit BEFORE we normalize it away for local
+      // persistence — otherwise the daemon would only see the empty
+      // apiKey field that ships in the persisted config.
+      await syncComposioConfigToDaemon(composio);
+      const normalized = normalizeSavedComposioConfig(composio);
```
mrcfps: `syncComposioConfigToDaemon` resolves `false` on 4xx/5xx/offline rather than throwing, but this path ignores that result and immediately normalizes the local config as if the save or clear succeeded. In the changed Connectors save/clear flow, that leaves Settings showing the new badge state while the daemon still has the old credential, so connector actions and Orbit can keep running against stale auth. Please capture the boolean result here and abort/throw when it is false so `ConnectorSection.handleSaveKey` stays in its error state instead of updating the saved marker; adding a regression test for the failed-save path would lock that contract in.

mrcfps (commented):
@mrcfps thanks for continuing to tighten up the Orbit activity-summary flow — I reviewed the current head across the daemon lifecycle, connector/settings persistence, Orbit manual and scheduled run handling, and the updated coverage. I found one non-blocking follow-up around confirming Composio credential persistence before the UI updates its saved/cleared state.


lefarcen (dismissed):
Thanks for addressing the media-credential clear path! The versioning mechanism (`mediaProvidersChangeVersionRef` + `forceMediaProviderSync` option) correctly tracks user edits and ensures empty credential maps reach the daemon when intentional. 🙏

The prior P2 items (stale auth masking, time validation, nextRunAt display, media clear) are all resolved on this head.

lefarcen (approved):

PR #687, opened 2026-05-06
fix(web): restore media config from daemon on startup

zztdandan:
## Summary
- restore media provider settings from `/api/media/config` during app bootstrap instead of relying only on `localStorage`
- preserve daemon-stored API keys when saving masked media provider state by submitting full provider targets for non-secret fields plus an explicit `preserveApiKey` signal for stored secrets
- harden daemon restore UX by distinguishing fetch failure from empty daemon config, falling back to browser state without writeback, and exposing reload + masked-key status in Settings

## Problems solved
1. **Restart / origin-reset recovery bug**
   - `media-config.json` could still exist on disk, but startup recovery only trusted browser `localStorage`, so media settings looked lost after restart or storage/origin changes.
2. **Masked-key overwrite bug (P1)**
   - Once startup loaded a stored provider as `apiKey: ''` plus `apiKeyConfigured: true`, saving unrelated settings or editing only `baseUrl` / `model` could wipe the daemon-stored API key.
3. **Fetch-failure migration bug (P2)**
   - A transient `/api/media/config` failure was previously indistinguishable from \"daemon has no media config\", so bootstrap could migrate stale browser state back into the daemon and overwrite real daemon-backed settings.
4. **Masked-key rendering / recovery UX gap**
   - Restored daemon keys were represented in state, but the media settings inputs rendered as blank fields with no saved-key indicator, which made recovery look broken and gave users no one-click way to reload daemon state.
5. **Model picker configured-state inconsistency (P2)**
   - The Settings dialog respected `apiKeyConfigured`, but the new-project media picker still treated daemon-restored providers as merely integrated.

## What changed
- added `fetchMediaProvidersFromDaemon()` result states so bootstrap can distinguish successful empty config from fetch failure
- updated `App.tsx` bootstrap to:
  - merge daemon-backed media providers only on successful fetch
  - fall back to browser-saved providers when daemon fetch fails
  - skip the one-time localStorage → daemon migration writeback unless the daemon fetch succeeded and confirmed no stored media config
  - skip settings-save media writeback when daemon media state is unknown
- kept the explicit `preserveApiKey` contract for P1 so masked daemon-backed credentials survive settings saves unless the user explicitly clears the provider
- added a media-provider notice plus a `Reload from daemon` action in Settings so users can re-fetch daemon state and discard unsaved local edits
- rendered restored media keys with Composio-style saved markers (`Saved · ••••tail`) and replacement placeholder copy instead of a misleading blank field
- updated `NewProjectPanel.tsx` to treat `apiKeyConfigured` as configured so the startup recovery UI stays consistent outside Settings
- extended focused web tests for fetch-failure handling, saved masked-key rendering, reload behavior, and model-picker configured badges

## Secret-handling contract
Media settings now use explicit masked-secret semantics instead of a plain daemon-over-local writeback loop:
- **Preserve**: if the daemon already has a key and the web state only has the saved marker (`apiKeyConfigured === true`, no new `apiKey`), the web sends `preserveApiKey: true`
- **Replace**: if the user pastes a new key, the web sends the new `apiKey` and the daemon stores it
- **Clear**: if the user explicitly clears the provider and saves, the provider is removed
- **Unknown daemon state**: if `/api/media/config` could not be fetched, the web falls back to browser state for the session but does not migrate or overwrite daemon-backed media config

## Why this matches project conventions
This still follows the daemon-owned persistence pattern used by `app-config.json`, but with an explicit saved-marker contract for masked secrets:
1. read local state for first paint
2. read daemon-backed state on bootstrap
3. merge daemon-backed persisted fields over local state when fetch succeeds
4. migrate local-only legacy values back to the daemon once **only after** confirming the daemon fetch succeeded and no daemon media config exists
5. preserve / replace / clear secrets through explicit media-config write semantics rather than inferring from masked browser state

## Validation
- `pnpm --filter @open-design/web test -- config SettingsDialog.media NewProjectPanel.media`
- `pnpm --filter @open-design/daemon test -- media-config`

The focused web tests and typecheck pass with these changes. The daemon media-config preservation tests remain the targeted coverage for the `preserveApiKey` contract.


lefarcen:
Hi @zztdandan! 🎉
Thanks for the contribution — this looks like a solid fix to align media config persistence with the existing app-config pattern.

— open-design team

lefarcen (dismissed):
Thanks for the PR @zztdandan! The approach of mirroring app-config patterns is sound, and the test coverage is solid.

However, there's a **critical issue** with masked daemon keys being accidentally erased when settings are saved. There are also edge cases around fetch failure handling and UI consistency that should be addressed.

---

## P1 (must-fix): Settings save can wipe stored media keys after daemon restore

**Location**: `apps/web/src/App.tsx` (around line 170–195 in the new version, where `handleConfigSave` is called)

**Issue**: `fetchMediaProvidersFromDaemon()` intentionally maps daemon keys to masked local entries with `apiKey: ''`, but later when a user opens Settings and clicks Save, `handleConfigSave()` sends `withOnboarding.mediaProviders` back to `/api/media/config` with `force: true`. The daemon write path treats empty `apiKey` as delete/replace, so the user can accidentally wipe `media-config.json` without re-entering keys.

**Fix**: Either:
- (a) Don't PUT masked entries as replacement data (filter them out or skip sync when `apiKeyConfigured === true` and `apiKey === ''`), OR
- (b) Update the daemon contract to preserve an existing key when `apiKeyConfigured === true` and no new `apiKey` is supplied; only clear via an explicit clear/delete action

**Test needed**: Add a test covering "daemon masked key restored, settings saved, key is not cleared."

---

## P2 (should-fix): Fetch failure confused with "no daemon config"

**Location**: `apps/web/src/state/config.ts:425` (new function `shouldSyncLocalMediaProvidersToDaemon`)

**Issue**: `shouldSyncLocalMediaProvidersToDaemon()` treats `daemonProviders === null` as "daemon has no media config," but `null` is also what `fetchMediaProvidersFromDaemon()` returns on GET failure. If `/api/media/config` briefly fails while the daemon is live, bootstrap can push stale local media settings back to the daemon and overwrite real daemon-backed state.

**Fix**: Distinguish "fetch failed/unknown" from "fetched successfully and empty," or require `daemonProviders !== null` before allowing the one-time legacy migration sync. Could use a discriminated union or separate flag.

---

## P2 (should-fix): Model picker UI doesn't recognize restored keys

**Location**: `apps/web/src/components/NewProjectPanel.tsx` (around line 1630, outside this PR's diff)

**Issue**: The restored `apiKeyConfigured` marker is correctly used in `SettingsDialog.tsx`, but it's not used by the media model picker. After this PR, the Settings dialog can show a provider as configured while the new-project model cards still show it as merely "Integrated" when the daemon key is restored and no base URL override exists. This leaves part of the startup recovery UI inconsistent.

**Fix**: Include `entry?.apiKeyConfigured` in the configured check in `NewProjectPanel.tsx` and add coverage or component-level verification for daemon-restored media provider state.

---

## P2 (reasoning gap): Masked secrets vs plain daemon-over-local

**PR body claim**: The reasoning says this mirrors the `app-config.json` pattern, but media config includes masked secrets and therefore needs Composio-style "saved marker preserves secret" semantics, not a plain daemon-over-local replace/writeback loop. The writeback failure mode (P1 above) is not considered in the PR description.

**Suggestion**: Update the implementation (fix P1) and PR explanation to explicitly call out how masked daemon secrets are preserved, replaced, and cleared. This will help future maintainers understand the contract.

---

## Next steps

The **P1 needs to be fixed** before merge — it's a data loss path that will bite users after restart. The P2s improve robustness but can be addressed as follow-up if you prefer to keep the scope tight.

Let me know if any of these need clarification!

lefarcen (commented):
Great work addressing the P1 data-loss issue! 🎉

The daemon-side `preserveApiKey` flag is the right fix — when the web sends masked credentials (empty `apiKey` but `apiKeyConfigured=true`), the daemon now preserves the stored secret instead of treating it as a delete. The test coverage is solid.

## ✅ P1 Fixed

`apps/daemon/src/media-config.ts` + `buildMediaProvidersForDaemonSave()` now correctly:
- Detect masked state (`apiKeyConfigured=true` + empty `apiKey`)
- Send `preserveApiKey: true` to daemon
- Daemon preserves `priorApiKey` when flag is set

## P2s Status

The remaining P2s from my prior review are non-blocking but would still improve robustness:

1. **Fetch failure vs no config** (`daemonProviders === null`): Still conflates fetch error with "daemon has no config." Low risk in practice since bootstrap retries handle transient failures, but could add an explicit `{ ok: false }` sentinel or separate the two cases for future-proofing.

2. **Model picker UI consistency** (`NewProjectPanel.tsx` not in this diff): Settings dialog correctly uses `apiKeyConfigured`, but the new-project model picker doesn't yet. Worth a follow-up so the UI is consistent across surfaces.

3. **PR reasoning**: Implementation now properly handles masked secrets. ✅

## Next Step

This PR is good to merge from my side — the P1 is fixed and tested. The P2s can be follow-up polish if needed.

Thanks for the quick turnaround and thorough daemon-side fix!

zztdandan:
Thanks — I addressed the requested follow-ups and updated the PR description to document the final contract.

### Fixed P1: masked settings save no longer wipes daemon media keys
I kept the explicit preserve contract across web + daemon:
- masked restored providers now save via `preserveApiKey: true` when no replacement key was entered
- replacing a key still sends the new `apiKey`
- clearing a provider still removes it explicitly

### Fixed P2: fetch failure is no longer treated as "daemon empty"
Bootstrap now distinguishes:
- **fetch succeeded + empty config** → allow the one-time localStorage → daemon migration
- **fetch failed / unknown** → fall back to browser-saved media providers for the session, surface a notice, and **do not** migrate or overwrite daemon media config

I also guarded the normal Settings save path so media providers are not written back while daemon media state is unknown.

### Fixed the masked-key rendering gap
I also fixed the follow-up issue we found after the review:
- restored daemon media keys now render with a Composio-style saved marker (`Saved · ••••tail`)
- the API-key field now shows replacement copy instead of a misleading blank field
- Settings gained a `Reload from daemon` action so users can re-fetch daemon state and discard unsaved local edits

### Fixed the picker consistency issue
`NewProjectPanel` now treats `apiKeyConfigured` as configured, so daemon-restored providers no longer show as merely "Integrated" outside Settings.

### Validation
- `pnpm --filter @open-design/web test -- config SettingsDialog.media NewProjectPanel.media`

I also updated the PR body so the preserve / replace / clear semantics for masked media secrets are documented for future maintainers.

lefarcen (commented):
Excellent progress! The major fixes are solid:

✅ **P1 fixed**: Masked settings save now uses `preserveApiKey` flag — daemon preserves stored keys when web sends empty `apiKey` with `preserveApiKey: true`.
✅ **P2 fixed**: Fetch failure distinguished from empty config via `status: 'error'` — no accidental migration when daemon is unreachable.
✅ **P2 fixed**: Saved-key rendering works (Composio-style marker).
✅ **P2 fixed**: `NewProjectPanel` treats `apiKeyConfigured` as configured.

**One edge case remains (P2):**

`apps/web/src/components/SettingsDialog.tsx:1294` — Saved media keys can still be wiped when editing non-secret fields. When a user clears the Base URL or model field on a daemon-restored provider, `updateProvider` deletes the provider entry (because `apiKey`, `baseUrl`, and `model` are all empty). Then the force-save in `App.tsx:292-295` calls `buildMediaProvidersForDaemonSave`, which doesn't emit `preserveApiKey` for the missing entry, and the daemon replace-write drops the stored key.

**Fix**: Preserve entries with `apiKeyConfigured` during normal field edits. Only the explicit Clear action should clear `apiKeyConfigured` and delete the provider.

This is a narrower case than the original P1 (which affected every save). Up to you and the maintainer whether to ship as-is or address this edge case first. The core restoration flow is solid. 🎉

mrcfps (changes requested):
@zztdandan Thanks for the thorough follow-up work on the daemon-backed media config path. I found one remaining masked-secret edge case that can still drop a stored provider key when a saved provider's editable fields are cleared, so I'm requesting changes for that data-loss path before merge.

**Blocking finding — `apps/web/src/components/SettingsDialog.tsx`, `MediaProvidersSection.updateProvider` / saved media-key inputs:** the new saved-state UI intentionally renders daemon-restored keys with `entry.apiKey === ''` while `entry.apiKeyConfigured` carries the stored-secret marker. However, `updateProvider` still deletes the provider whenever the visible `apiKey`, `baseUrl`, and `model` fields are all empty, without checking `apiKeyConfigured`. If a user clears the Base URL/model field on a restored provider to return to defaults, the provider entry is removed from `cfg.mediaProviders`; the settings save then force-writes the replacement map without that provider, so the daemon drops the stored key even though the user did not use the explicit Clear action. Please preserve `apiKeyConfigured` entries during normal field edits and make Clear use a separate path that removes the marker/deletes the provider; add a regression test for clearing a non-secret field on a daemon-restored provider.


zztdandan:
Thanks — I fixed this remaining masked-secret edge case by **unifying media-provider empty-state and clear semantics around two explicit rules**.

### The two unified rules
1. **A media provider entry is considered present unless all of these are empty/cleared:**
   - `apiKey`
   - `baseUrl`
   - `model`
   - `apiKeyConfigured`
   - `apiKeyTail`

2. **Explicit Clear is the only path that clears an entry to that empty-state threshold**
   - Clear now wipes the visible fields **and** the saved-marker fields:
     - `apiKey: ''`
     - `baseUrl: ''`
     - `model: ''`
     - `apiKeyConfigured: false`
     - `apiKeyTail: ''`

### What this fixes
Previously, the edge case happened because normal edit flow only looked at visible fields (`apiKey`, `baseUrl`, `model`) when deciding whether to delete a provider entry. For daemon-restored masked keys, `apiKey === ''` is expected, so clearing a non-secret field like Base URL could incorrectly delete the whole provider entry before save. That meant the subsequent force-write no longer emitted that provider, and the daemon replace-write dropped the stored key.

With the unified rules above:
- **normal non-secret edits** no longer delete daemon-restored masked entries just because visible fields become empty
- **only explicit Clear** removes the saved-marker fields and lets the entry become truly empty/deletable

### Implementation
- added shared helpers in `config.ts`:
  - `isStoredMediaProviderEntryPresent(...)`
  - `isStoredMediaProviderEntryEmpty(...)`
- updated `SettingsDialog` to use those helpers for:
  - provider sorting/configured checks
  - edit-time deletion decisions
  - clearable checks
  - explicit Clear behavior
- updated `NewProjectPanel` to use the same shared presence helper so this logic does not drift again
- updated `hasAnyConfiguredProvider(...)` to use the same semantics as the rest of the media-provider flow

### Regression coverage
Added tests for:
- saved-marker entries are still considered present even when visible fields are empty
- an entry is only empty after both visible fields **and** saved-marker fields are cleared
- clearing only a non-secret field preserves the restored provider on save
- explicit Clear removes the provider on save

### Validation
- `pnpm --filter @open-design/web test -- config SettingsDialog.media NewProjectPanel.media`

lefarcen (commented):
✅ **All findings resolved!**

mrcfps's blocking edge case is now fixed — the unified entry-presence rules correctly preserve `apiKeyConfigured` entries during normal field edits, and only the explicit Clear button removes the saved-marker fields.

**Verified:**
- `isStoredMediaProviderEntryPresent()` / `isStoredMediaProviderEntryEmpty()` check all 5 fields (visible + saved-marker)
- `updateProvider` uses `isStoredMediaProviderEntryEmpty(next)` before deletion
- Clear button is the only code path that sets `apiKeyConfigured: false` + `apiKeyTail: ''`
- Regression tests added for: clearing non-secret fields preserves entry, Clear removes entry

**All prior findings remain fixed:**
- P1 (data loss): daemon `preserveApiKey` flag working
- P2 (fetch failure handling): `status: 'error'` vs `status: 'ok'` distinction
- P2 (UI): saved-key rendering + picker consistency

Ready to merge from my side; deferring final approval to a maintainer. 🎉

lefarcen (commented):
Perfect! All blockers resolved. 🎉

✅ **mrcfps's P2 fixed**: `updateProvider` now uses `isStoredMediaProviderEntryEmpty(next)` which checks ALL fields including `apiKeyConfigured` and `apiKeyTail`. Daemon-restored keys survive non-secret field edits — only the explicit Clear button removes the saved-marker fields.

✅ **Prior P1 + 3 P2s still addressed**: Daemon `preserveApiKey` contract, fetch-failure distinction, saved-key rendering, and `NewProjectPanel` consistency all intact.

✅ **Clean implementation**: The unified empty-state helpers (`isStoredMediaProviderEntryPresent` / `isStoredMediaProviderEntryEmpty`) are used consistently across `SettingsDialog`, `NewProjectPanel`, and `config.ts`. Regression tests cover the edge cases.

**Ready to merge from my side.** Deferring final approval to a maintainer. 👍

mrcfps (dismissed):
@zztdandan I reviewed the daemon-backed media config bootstrap, masked-key preserve/replace/clear flow, fetch-failure handling, Settings reload/saved-key UI, NewProjectPanel configured-state update, and the focused regression coverage. The latest changes address the earlier data-loss edge cases by preserving saved-marker entries during normal edits and limiting deletion to the explicit Clear path. Thanks for the careful follow-through here — this is a much safer and clearer media settings flow now. 🎉


zztdandan:
Conflicts with current `main` are resolved on my side. I merged `upstream-remote/main`, reconciled the media-provider restore flow with the newer autosave/settings changes, updated the affected web tests, and verified with `pnpm install`, `pnpm guard`, `pnpm typecheck`, `pnpm --filter @open-design/web test`, and `pnpm --filter @open-design/daemon test`.

apps/web/src/state/config.ts
```diff
@@ -383,25 +504,54 @@ export function mergeDaemonConfig(
...
+export function mergeDaemonMediaProviders(
+  localConfig: AppConfig,
+  daemonProviders: AppConfig['mediaProviders'] | null,
+): AppConfig {
+  if (!hasAnyDaemonManagedMediaProvider(daemonProviders)) {
+    return { ...localConfig };
+  }
+  return {
+    ...localConfig,
+    mediaProviders: { ...(daemonProviders ?? {}) },
```
mrcfps: `mergeDaemonMediaProviders()` currently replaces the whole `mediaProviders` map as soon as any daemon-managed provider exists. That means a session that was running on the browser fallback path can still lose data on the next successful bootstrap/reload: if localStorage contains providers A+B from the outage, but `/api/media/config` only comes back with provider A, this line drops B entirely before anything is re-synced. Because `fetchMediaProvidersFromDaemon()` builds its object from the daemon response and this helper does a full replacement, the PR's new "use browser-saved settings for now" path is not actually preserved once the daemon partially recovers. Please merge per provider instead—keep local entries for providers whose daemon entry is effectively empty, overwrite only the providers that the daemon really has—and add a regression test that covers a local-only provider surviving when the daemon reload returns a different provider set.

mrcfps (changes requested):
@zztdandan Thanks for the follow-up on the daemon-backed media restore flow. I found one remaining recovery-path data-loss issue in the new merge logic, so I'm requesting changes before merge.


zztdandan:
I tracked down the follow-up regression behind the desktop window getting stuck on "Waiting for the web runtime URL…", and the root cause is narrower than the media-config change itself.

What I fixed:
- The desktop runtime was recording the discovered `web.url` before `window.loadURL(url)` had actually succeeded.
- If the first `loadURL()` failed during startup, desktop would remember that URL as already handled and would never retry the same URL again, so the window could stay on the waiting screen forever.
- The fix is to store `currentUrl` only after `loadURL()` succeeds, so the same discovered URL can be retried on the next poll if the first load races with web startup.

Why my PR #687 changes could trigger it:
- PR #687 did not introduce a new desktop bug directly.
- What it changed was the startup timing of the web app by adding media-provider restoration/sync work during bootstrap.
- That made it easier to hit an existing desktop race: desktop could discover the web URL while the page was still stabilizing, fail the first `loadURL()`, and then never retry that same URL.

Why this is not a lifecycle mismatch with `app-config`:
- The media-provider read is already happening in the same bootstrap window as `app-config`.
- Both are fetched in the same startup `Promise.all(...)`, and then merged into app state in the same bootstrap effect.
- So the issue is not that media config was restored in the wrong lifecycle; it is that adding that restoration made the startup window heavier and exposed the desktop retry bug.

My reasoning for the PR #687 implementation:
- I intentionally kept media-provider restoration aligned with the existing `app-config` bootstrap path, because it is also startup configuration restoration.
- I did not want to create a second delayed hydration path that could cause visible config flicker or inconsistent precedence between daemon-backed config and local state.
- Given that, the right fix is on the desktop side: make URL loading retry-safe instead of weakening or splitting the config bootstrap flow.

zztdandan:
Fixed.

I confirmed the reviewer concern was valid: `mergeDaemonMediaProviders()` was replacing the whole `mediaProviders` map as soon as any daemon-backed provider existed, which could drop local-only providers after a partial daemon recovery.

What I changed:
- Switched the merge to be per-provider instead of replacing the entire map.
- Daemon-backed providers now override only their matching local entries.
- Local-only providers are preserved when the daemon response does not contain them.
- Empty daemon entries do not wipe local entries.

I also added a regression test covering the exact case described in the review: a local-only provider survives when the daemon reload returns a different partial provider set.

This has now been committed and pushed to the branch.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -3290,29 +3304,52 @@ function MediaProvidersSection({
...
+    if (!onReloadMediaProviders || reloadRunning) return;
+    setReloadRunning(true);
+    setReloadNotice(null);
+    try {
+      const next = await onReloadMediaProviders();
+      if (!next) {
+        setReloadNotice('Could not reload media provider settings from the local daemon.');
+        return;
+      }
+      setCfg((curr) => ({ ...curr, mediaProviders: next }));
```
mrcfps: `MediaProvidersSection.handleReload()` is replacing the dialog state with the raw daemon response, which drops any local-only providers that `App.reloadMediaProvidersFromDaemon()` just preserved via `mergeDaemonMediaProviders()`. In the recovery path this can still lose data: if the dialog is showing browser-fallback providers A+B and the daemon reload only returns A, clicking **Reload from daemon** removes B from `cfg.mediaProviders`; the next Save then force-writes that reduced map back through `syncMediaProvidersToDaemon`, so provider B is lost even though the user never cleared it. Please apply the same per-provider merge here (or have `onReloadMediaProviders` return the merged map instead of the raw daemon map) and add a regression test that reloads after a partial daemon recovery.

mrcfps (changes requested):
@zztdandan Thanks for the follow-up here. I found one remaining data-loss path in the new reload flow, so I'm requesting changes before merge.


zztdandan:
Fixed as well.

There was still one remaining data-loss path in the Settings dialog reload flow: `MediaProvidersSection.handleReload()` was replacing `cfg.mediaProviders` with the raw daemon response, so after a partial daemon recovery the dialog could drop local-only providers even though `App.reloadMediaProvidersFromDaemon()` had already preserved them.

I updated the dialog reload path to use the same per-provider merge behavior instead of replacing the whole map, and added a regression test that covers the exact scenario from the review: local fallback providers survive when Reload from daemon returns only a partial provider set.

This has now been committed and pushed to the branch.

lefarcen (commented):
✅ **mrcfps's finding is fixed.**

I verified the current head (4b8733c3):

- `mergeDaemonMediaProviders()` now merges per-provider (starts with local providers, overwrites only daemon-managed ones) — the whole-map replacement is gone.
- The requested regression test is present: `preserves local-only providers when daemon returns a partial provider set` (config.test.ts:150-185).

All prior P1/P2 findings remain addressed. Ready to merge from bot side; deferring final approval to a maintainer.

mrcfps:
@zztdandan I'm holding off on generating review comments for nexu-io/open-design#687 because this pull request has merge conflicts right now.

zztdandan:
## Conflict resolution update

Resolved the latest upstream conflicts locally while keeping the daemon-backed media config behavior intact.

### What I kept from upstream
- the split App bootstrap/loading flow in `apps/web/src/App.tsx`
- the stricter TypeScript + `isRecord` handling in `apps/daemon/src/media-config.ts`
- the Clear confirmation UX in `apps/web/src/components/SettingsDialog.tsx`

### What I preserved from this branch
- daemon-backed media provider restore/merge during bootstrap
- `preserveApiKey` semantics for masked saved keys
- fallback behavior when `/api/media/config` fetch fails, so browser-saved state is used without overwriting daemon-backed config
- saved-key state cleanup for `apiKeyConfigured` / `apiKeyTail` when a provider is explicitly cleared

### Latest follow-up conflict handled
A newer upstream sync introduced one additional conflict in:
- `apps/web/tests/state/config.test.ts`

Resolved that by keeping both sides of the test coverage:
- media config merge/preserve/fallback test coverage from this branch
- newer upstream coverage for privacy / orbit / ollama-related config behavior

### Validation
Completed locally after conflict resolution:
- `pnpm guard`
- `pnpm typecheck`
- `pnpm --filter @open-design/web test -- SettingsDialog.media NewProjectPanel.media`
- `pnpm --filter @open-design/daemon test -- media-config`
- `pnpm --filter @open-design/web test -- state/config App.mediaProviders SettingsDialog.media`

### Local merge commits
- `cea83cbd` — `merge upstream main into media config fixes`
- `76a5af2c` — `merge upstream main into media config fixes`

zztdandan:
### Request for timely merge

This branch has already required repeated upstream conflict resolution because the project is changing quickly.

Please merge this branch **as soon as possible**.

Why this matters:
- it reduces the need for repeated rebases and conflict resolution on this PR
- it helps follow-up changes account for the media config and daemon restore behavior introduced here
- it reduces the risk that nearby changes continue to diverge from this branch and create more expensive conflicts later

The sooner this lands, the easier it will be for subsequent work to build on the current behavior instead of accidentally conflicting with it.

apps/web/src/state/config.ts
```diff
@@ -482,25 +603,61 @@ export function mergeDaemonConfig(
...
+  return Object.values(providers).some((entry) => isStoredMediaProviderEntryPresent(entry));
+}
+
+export function shouldSyncLocalMediaProvidersToDaemon(
+  localProviders: Record<string, MediaProviderCredentials> | undefined,
+  daemonProviders: Record<string, MediaProviderCredentials> | null | undefined,
+): boolean {
+  return daemonProviders != null
+    && hasAnyConfiguredProvider(localProviders)
+    && !hasAnyDaemonManagedMediaProvider(daemonProviders);
```
mrcfps, line 665: `shouldSyncLocalMediaProvidersToDaemon()` now treats `apiKeyConfigured`/`apiKeyTail` markers as real local config via `hasAnyConfiguredProvider()`. That means a session that previously hydrated from the daemon, saved the masked marker back to localStorage, and later starts against an empty/new daemon will still take the migration path even though the browser no longer has a recoverable secret. In this PR that path falls through to `buildMediaProvidersForDaemonSave()`, which synthesizes a payload with only `defaultBaseUrl`, so the daemon ends up with a phantom provider entry and the UI can keep showing the provider as configured after the actual key has been lost. Please gate this one-time migration on recoverable local fields (`apiKey`, `baseUrl`, `model`) or explicitly skip marker-only entries when there is no daemon key to preserve, and add a regression test for the "daemon reset + masked local marker" case.

mrcfps (commented):
@zztdandan Thanks for tightening up the daemon-backed media restore flow and covering the earlier data-loss regressions. I found one more recovery-path edge case around re-migrating masked local markers into an empty daemon config, so I left a non-blocking inline suggestion below.


zztdandan:
Fixed the latest media-provider follow-up and pushed it to this branch in `d249c5f6`.

### What changed
- tightened the one-time localStorage → daemon migration guard so it only runs when local media-provider state still has **recoverable fields** (`apiKey`, `baseUrl`, or `model`)
- masked saved-marker-only local state (`apiKeyConfigured` / `apiKeyTail` with no recoverable fields) no longer counts as migratable config when the daemon is empty

### Why this fixes the review concern
Previously, a session that had only masked local markers could still re-migrate into an empty/reset daemon and synthesize a phantom provider entry from defaults alone.

Now that migration path is skipped unless there is actual recoverable local provider data to persist.

### Regression coverage
Added a focused test covering exactly this case:
- daemon has no persisted media provider config
- local state contains only a masked saved marker
- `shouldSyncLocalMediaProvidersToDaemon(...)` returns `false`

### Validation
- `pnpm --filter @open-design/web test -- tests/state/config.test.ts`

I intentionally did **not** include the separate Windows smoke/desktop status timing investigation here, since that is outside the scope of this PR and should be handled independently if needed.

lefarcen (changes requested):
Hey @zztdandan, thanks for the thorough follow-up on mrcfps's marker-only migration concern. I found one remaining data-loss path in the manual Orbit save flow plus two non-blocking issues:

**P1 — Data loss: masked saved keys wiped by manual Orbit save**
`apps/web/src/components/SettingsDialog.tsx` around line 2469

`persistConfigAndRunOrbit()` calls `syncMediaProvidersToDaemon(config.mediaProviders)` without the daemon provider snapshot. For a daemon-restored masked provider (`apiKeyConfigured: true`, empty `apiKey`), `buildMediaProvidersForDaemonSave()` won't emit `preserveApiKey` because `daemonProviders` is `undefined`, but it can still emit a default/base URL; daemon `writeConfig()` then replaces the stored provider with an empty `apiKey`, wiping the saved secret before the Orbit run.

Fix by threading the loaded daemon media-provider state into the manual Orbit save path, or make the save builder preserve masked current entries safely without synthesizing default-only entries when no prior daemon state is known. Add a regression test for masked saved media key + manual Orbit run preserving the daemon key.

**P2 — Marker-only stale state UI inconsistency**
`apps/web/src/state/config.ts` around line 620

The latest marker-only migration guard prevents syncing masked-only local state into an empty daemon, but `mergeDaemonMediaProviders()` still returns and saves the stale local marker unchanged when the daemon fetch succeeds with no stored media config. Because `isStoredMediaProviderEntryPresent()` treats `apiKeyConfigured`/`apiKeyTail` as configured, the UI/model picker can still show a provider as "Saved"/"Configured" even though the daemon has no key and the local marker has no recoverable secret.

Fix by scrubbing marker-only entries when the daemon definitively reports empty state, or by making marker-only presence count only when backed by the current daemon response.

**P3 — i18n regression: hard-coded English UI strings**
`apps/web/src/components/SettingsDialog.tsx` around line 3398

New user-facing strings are hard-coded in the Settings media UI (`Reload from daemon`, reload notices, `Saved to local daemon`, replacement placeholder), bypassing the existing i18n system used by this dialog. This regresses localized Settings pages.

Move these strings into the locale key set and render them through `t(...)`.

lefarcen (changes requested):
@zztdandan Thanks for the thorough iteration on the daemon-backed media restore flow. I found one remaining data-loss path in the manual Orbit save, plus two follow-up items (marker-only state handling + i18n bypass), so requesting changes before merge.

**P1 — apps/web/src/components/SettingsDialog.tsx:2469 (data loss: manual Orbit run can wipe masked daemon keys)**

`persistConfigAndRunOrbit()` calls `syncMediaProvidersToDaemon(config.mediaProviders)` without passing `daemonProviders`, so `buildMediaProvidersForDaemonSave()` doesn't emit `preserveApiKey` for daemon-restored masked providers. The daemon's `writeConfig()` then replaces the stored key with an empty `apiKey`, wiping the saved secret before the Orbit run.

Fix: thread the loaded `daemonMediaProviders` state into the manual Orbit save path (add it to component state if needed), or make the save builder preserve masked current entries safely without synthesizing default-only entries when no prior daemon state is known. Add a regression test for "masked saved media key + manual Orbit run preserves daemon key".

**P2 — apps/web/src/state/config.ts:620 (marker-only state after empty daemon fetch)**

The latest migration guard prevents syncing marker-only local state into an empty daemon ✓, but `mergeDaemonMediaProviders()` still returns and saves the stale local marker unchanged when daemon fetch succeeds with no stored config. Because `isStoredMediaProviderEntryPresent()` treats `apiKeyConfigured`/`apiKeyTail` as configured, the UI can show "Saved" even though daemon has no key and local has no recoverable secret.

Fix: scrub marker-only entries when daemon definitively reports empty state, or make marker-only presence count only when backed by current daemon response.

**P3 — apps/web/src/components/SettingsDialog.tsx:3398 (i18n bypass in new UI strings)**

New user-facing strings ("Reload from daemon", reload notices, "Saved to local daemon", replacement placeholder) are hard-coded, bypassing the existing i18n system. This regresses localized Settings pages.

Fix: move these strings into locale key set and render through `t(...)`.

---

Once these are addressed (especially P1), this will be ready from the bot side. Thanks for the careful work on the media restore flow.

apps/desktop/src/main/runtime.ts
```diff
@@ -372,8 +372,8 @@ export async function createDesktopRuntime(options: DesktopRuntimeOptions): Prom
     try {
       const url = await options.discoverUrl();
       if (url != null && url !== currentUrl) {
-        currentUrl = url;
         await window.loadURL(url);
+        currentUrl = url;
```
mrcfps, line 382: Moving `currentUrl = url` after `await window.loadURL(url)` fixes the same-URL retry path, but it also leaves `status().url` as `null` for the whole navigation window even after the packaged app is far enough along for inspect/eval to run. On this head the `Packaged windows smoke` job is already failing at `specs/win.spec.ts:185` with `expected null to be od://app/`, which points back to `status()` still reading the old `currentUrl` while this await is in flight. Please keep a retry-safe split between the URL you report in `status()` and the URL you use for retry gating—for example, set the reported URL before awaiting `loadURL()` and roll it back on failure, or track a separate `lastSuccessfulUrl`/`pendingUrl` pair—so the smoke status stays populated without reintroducing the no-retry bug.

mrcfps (commented):
@zztdandan Thanks for the thorough follow-through on the daemon-backed media restore flow and the added regression coverage. I found one follow-up issue in the desktop retry change, and it already lines up with the failing Windows packaged smoke check on this head, so I left the concrete fix path inline below.


zztdandan:
Addressed the remaining follow-ups and pushed them in `0697a20e`.

### Resolved in this update

#### 1) P1 — manual Orbit run no longer wipes masked daemon media keys
- `persistConfigAndRunOrbit(...)` now accepts the loaded `daemonMediaProviders` snapshot
- the Orbit manual-save path threads that snapshot through to `syncMediaProvidersToDaemon(...)`
- masked restored providers now emit `preserveApiKey: true` before `/api/orbit/run`, so the daemon keeps the stored secret instead of replacing it with an empty key
- added regression coverage for: **masked saved media key + manual Orbit run preserves daemon key**

#### 2) P2 — stale marker-only local state is scrubbed when daemon definitively reports empty media config
- `mergeDaemonMediaProviders(...)` now distinguishes:
  - `daemonProviders == null` → unknown / fetch-failed, keep local fallback state
  - `daemonProviders` loaded but empty → daemon definitively has no stored media config
- in that loaded-empty case, marker-only local entries (`apiKeyConfigured` / `apiKeyTail` with no recoverable `apiKey` / `baseUrl` / `model`) are dropped instead of continuing to show as falsely saved/configured
- local entries that still have recoverable fields are preserved
- added regression coverage for the stale marker-only case

#### 3) P3 — new media-provider UI strings now go through i18n
- moved the new media-provider reload / notice strings into the locale dictionary + types
- switched the Settings media UI to use `t(...)` for:
  - reload button label
  - reload success/error notice
  - daemon-load fallback notice
  - saved badge / saved-with-tail text
  - replacement placeholder copy

#### 4) desktop follow-up / Windows smoke status path
- updated the desktop runtime to keep a retry-safe split between the last successful URL and the in-flight pending URL
- `status().url` now reports the pending URL during navigation, while same-URL retry safety is preserved if the load fails
- added focused coverage for the status-url helper path in packaged tests

### Validation
- `pnpm --filter @open-design/web test -- tests/state/config.test.ts tests/components/SettingsDialog.test.ts tests/components/SettingsDialog.media.test.tsx tests/components/SettingsDialog.execution.test.tsx`
- `pnpm --filter @open-design/desktop typecheck`
- `pnpm --filter @open-design/packaged test -- tests/desktop-url-allowlist.test.ts`
- `pnpm --filter @open-design/packaged typecheck`
- `pnpm guard`
- `pnpm typecheck`

So at this point:
- the manual Orbit data-loss path is fixed
- the empty-daemon stale-marker inconsistency is fixed
- the new media-provider UI strings are localized
- the desktop status/reporting follow-up behind the Windows smoke is also addressed

mrcfps (approved):
@zztdandan I reviewed the daemon-backed media restore/bootstrap flow, the masked-key preserve/replace/clear path, the reload handling, the marker-only migration guard, the Settings/New Project UI updates, and the desktop status retry follow-up. The latest head covers the earlier data-loss and packaged-status regressions with focused regression tests; thanks for the careful iteration here.


lefarcen (approved):
All prior findings (P1 manual Orbit data loss, P2 marker-only state, P3 i18n bypass) addressed on head 0697a20e. Ready from bot side. 🎉

leilei926524-tech:
Hi @zztdandan — quick note from the Open Design team.

What you get there:

PR #678, opened 2026-05-06
feat(skills): add social-media-dashboard skill + Totality Festival design system

Tuola-waj:
Open Design social media dashboard — X tab and GitHub tab side by side

## Summary

This PR ships the **social-media-dashboard** skill, a new **Totality Festival** design system, and a small fix to the design-systems swatches parser. Together they let users go from "paste a DESIGN.md" → "live, ticking analytics dashboard" in one prompt — the live-artifact demo path.

### 1. New skill — `skills/social-media-dashboard/`
- `SKILL.md`: single-screen creator analytics dashboard. Triggers on `social media dashboard`, `creator analytics`, `social analytics`, `创作者后台`, etc.
- `example.html` (1500+ lines, fully self-contained — no external assets):
  - Header with brand mark + ANALYTICS label, platform pills (X / GitHub), 7D / 30D / 90D / YTD time-range tabs, creator block.
  - Hero strip with display-scale headline + supporting line + Export / New post CTAs.
  - Ask bar with platform-aware ghost text.
  - Platform switcher cards (X active by default, with ambient amber glow).
  - 4 KPI cards that swap between X-native (Followers / Engagement / Likes / Reposts) and GitHub-native (Stars / Forks / PRs / Contributors) when the active platform changes.
  - Follower-growth / star-growth chart with annotation dots that auto-place above or below to avoid overlap with the curve.
  - Top Post preview (X) ↔ Top Contributors grid (GitHub) in the same slot.
  - Audience geography card with a recognisable simplified world map and 6 audience cluster dots.
  - Trending topics + top comments lower row, content swaps per platform.
  - Inline `<script>` ticks all numeric KPIs every 2s, including the engagement-rate percentage (random walk with a clamped band), and breathes the chart's last point — this is the live-artifact moment for the launch video.

### 2. New design system — `design-systems/totality-festival/`
- Mirrors the upstream `google-labs-code/design.md` Totality Festival example so any DESIGN.md-spec compatibility story we tell holds at the file level.
- Dual format: YAML front matter (Google design.md spec) + Markdown body (Open Design's existing token / category / surface conventions). The YAML keeps the file lint-clean against `@google/design.md`; the markdown body keeps the picker thumbnail, category, and prose summary working in our own UI.
- Category: Themed & Unique. Surface: web.

### 3. Parser fix — `apps/daemon/src/design-systems.ts`
[...]

PR #697, opened 2026-05-06
Fix: Remove element selector tooltip in Tweaks mode

1119302165:
### Problem
When the "Tweaks" mode is enabled and users click on elements, a floating tooltip appeared displaying the CSS selector path. This tooltip was:
- Visually obstructive, blocking the element being inspected
- Prone to text overflow issues with long selector paths
- Cluttering the preview area with redundant information

image

### Solution
Removed the floating tooltip entirely. Users can now inspect elements without visual obstruction.

### Changes
**Files modified:**
- `apps/web/src/components/FileViewer.tsx` - Removed tooltip JSX rendering from `CommentTargetOverlay` component
- `apps/web/src/index.css` - Removed `.comment-target-tooltip` and related CSS styles

### Behavior Change
| Aspect | Before | After |
|--------|--------|-------|
| Element selection | Shows blue overlay + floating tooltip | Shows blue overlay only |
| Tooltip content | CSS selector, dimensions, kind | N/A |
| Visual obstruction | Yes (tooltip blocks element) | No |
| Long selector handling | Text overflow issues | N/A (tooltip removed) |

after fix：
image


lefarcen:
Hi @1119302165! 🎉

Thanks for the contribution — the detailed before/after analysis and screenshots are really helpful.

— open-design team

apps/web/src/index.css
```diff
@@ -5691,26 +5691,26 @@ button.connector-action.is-loading {
   left: 0;
   top: 0;
   transform: translateY(calc(-100% - 6px));
-  display: inline-flex;
-  align-items: center;
-  gap: 7px;
-  max-width: min(380px, 80vw);
-  padding: 4px 7px;
+  display: inline-block;
+  width: 280px;
```
lefarcen: **P2** — The fixed `width: 280px` creates a content-box tooltip of ~296px with padding, which can still be clipped by `.comment-frame-clip { overflow: hidden; }` on narrow previews/mobile or when the selected element is near the right edge. This also drops the previous responsive cap `max-width: min(380px, 80vw)`.

**Concrete fix:** Keep the wrapping behavior but make it responsive:
```css
box-sizing: border-box;
width: max-content;
max-width: min(380px, calc(100vw - 16px));
overflow-wrap: anywhere;
```
If right-edge clipping is still observable, consider clamping/flipping the tooltip position in the overlay positioning code rather than relying on a fixed width.

apps/web/src/index.css
```diff
@@ -5691,26 +5691,26 @@ button.connector-action.is-loading {
...
-  line-height: 1.2;
-  white-space: nowrap;
+  line-height: 1.4;
   box-shadow: var(--shadow-sm);
 }
 .comment-target-tooltip strong {
   color: #8ec5ff;
   font-weight: 700;
+  display: block;
+  word-break: break-all;
```
lefarcen: **P3** — `word-break: break-all` fixes overflow but can split CSS selectors at arbitrary character boundaries (e.g., `div.ca|lculator` mid-word), making selector paths harder to scan.

**Prefer:** `overflow-wrap: anywhere` on the tooltip or selector text with `word-break: normal`, so long unbroken selector strings still wrap while preserving better line breaking for normal text.

lefarcen (commented):
Hey @1119302165! The fix addresses the immediate overflow issue well, and I appreciate the detailed before/after analysis. I found two suggestions for making the tooltip more robust across different screen sizes:

**Summary:**
- ✅ Solves the long selector overflow problem
- ⚠️ The fixed `280px` width introduces responsive clipping risk
- 💡 `word-break: break-all` can split selectors awkwardly

See inline comments for concrete suggestions.

lefarcen:
Hey @1119302165! I see your comment, but I'm not seeing any new commits yet — the current head is still `6115e3af` (the original commit).

Your comment mentions "Removed the floating tooltip entirely," but the current diff still shows the tooltip with the fixed 280px width + word-break changes.

**If you're planning to push a new approach:**
Push your commits first, then I'll automatically re-review the new changes.

**If you'd like to discuss the suggestions before coding:**
Happy to clarify! The two points from my review were:
1. Fixed `280px` width can cause clipping on narrow screens — suggested making it responsive
2. `word-break: break-all` splits selectors mid-word — suggested `overflow-wrap: anywhere` instead

Let me know which direction you'd like to go!

1119302165:
@lefarcen  try review, change Solution Removed the floating tooltip entirely. Users can now inspect elements without visual obstruction.

lefarcen (commented):
Thanks for the quick update, @1119302165! I see you went with a different approach — removing the tooltip entirely.

**My take:**

✅ **Solves the overflow problem** — no tooltip means no overflow  
⚠️ **But loses useful information** — the tooltip showed element ID, dimensions, selection type, etc. That was helpful for debugging and understanding what you're selecting.

**Alternative suggestion:**
Instead of removing it completely, what if we:
1. Keep the tooltip but show it **on hover** instead of always-on?
2. Or show just the essential info (element ID only) in a compact format?

This way users get the info when they need it, without it being visually obtrusive or overflowing.

**Your choice:**
If you prefer the "no tooltip" approach for simplicity, that's valid too — less visual clutter. But I wanted to point out that we're trading off discoverability/debugging info.

Let me know which direction you'd like to go!

1119302165:
> Thanks for the quick update, @1119302165! I see you went with a different approach — removing the tooltip entirely.
> 
> **My take:**
> 
> ✅ **Solves the overflow problem** — no tooltip means no overflow ⚠️ **But loses useful information** — the tooltip showed element ID, dimensions, selection type, etc. That was helpful for debugging and understanding what you're selecting.
> 
> **Alternative suggestion:** Instead of removing it completely, what if we:
> 
> 1. Keep the tooltip but show it **on hover** instead of always-on?
> 2. Or show just the essential info (element ID only) in a compact format?
> 
> This way users get the info when they need it, without it being visually obtrusive or overflowing.
> 
> **Your choice:** If you prefer the "no tooltip" approach for simplicity, that's valid too — less visual clutter. But I wanted to point out that we're trading off discoverability/debugging info.
> 
> Let me know which direction you'd like to go!

@lefarcen 
Thanks for your feedback. As you can see from the screenshot I uploaded, the ID and other related information are already shown in the top-left card. The tooltip is actually redundant, so removing it won’t have any impact on usage and debugging. 



after fix：
image

lefarcen:
Ah, that makes perfect sense! 💡 If the element info is already displayed in the top-left card, then the tooltip **is** redundant — removing it is the cleaner solution.

Thanks for the clarification and the photo! This approach is better than I initially thought. The PR removes noise while keeping the useful debugging info visible where it matters.

**Updated verdict:** Looks good to me. Ready for maintainer review. 👍

1119302165:
> Ah, that makes perfect sense! 💡 If the element info is already displayed in the top-left card, then the tooltip **is** redundant — removing it is the cleaner solution.
> 
> Thanks for the clarification and the photo! This approach is better than I initially thought. The PR removes noise while keeping the useful debugging info visible where it matters.
> 
> **Updated verdict:** Looks good to me. Ready for maintainer review. 👍

@mrcfps

1119302165:
@lefarcen  Why is the job ci / Validate workspace (pull_request) failing?
This code change does not involve any content related to this check.
Please check if this failure was caused by other merged PRs, and provide the root cause.

lefarcen:
Good catch, @1119302165! 👍

The CI failure is **not caused by your changes** — it's a **main-branch issue** that's affecting multiple recent PRs.

**Root cause:**
The `Validate workspace` job runs e2e tests that check localization coverage. The test expects **68 skills** to have German/Russian/French translations, but only **67** are present. This is a **missing translation** in the main branch, not related to your tooltip removal.

**Evidence:**
- Your PR only touches `apps/web/src/index.css` (tooltip styles) — no impact on localization files
- Recent `main` branch CI runs also show failures: `ci failure 2026-05-06` (3 consecutive runs)
- Error message:
  ```
  skills display copy: expected [ 'audio-jingle', 'blog-post', …(67) ] 
  to deeply equal [ 'audio-jingle', 'blog-post', …(68) ]
  ```

**What this means for your PR:**
Your code changes are clean. The CI failure is a **known main-branch issue** that needs to be fixed separately (likely a recent PR merged without updating all locale files).

**Next steps:**
- The maintainer team will either:
  1. Fix the missing translations on main (unblocking all PRs), OR
  2. Merge your PR despite the CI failure (since it's unrelated to your changes)

cc @mrcfps — this PR is ready for merge from a code perspective; CI failure is upstream.

lefarcen:
Hey @1119302165! 👋

Good catch — the CI failure is **not caused by your PR**. Here's what's happening:

**Root cause:**
PR #678 (merged ~1 hour ago) added a new skill `social-media-dashboard`, but didn't update the localized display copy for German, Russian, and French. The test `localized-content.test.ts` now expects 68 skills in those locales, but only 67 are present.

**Your PR only changed:**
`apps/web/src/index.css` (tooltip CSS) — completely unrelated to localization.

**What happens next:**
The maintainers will either:
1. Fix the missing localization entries in a follow-up PR, OR
2. Temporarily skip the localization check until translations are added

**Your PR is clean** — the CI failure is a pre-existing issue from another merged PR. I'll flag this to the team.

Thanks for raising this! 🙏

mrcfps (approved):
@1119302165 I reviewed the current changed ranges in `FileViewer.tsx` and `index.css`: the tooltip markup/styles are removed while the target overlay remains in place for both single-element and pod selections, so the overflow/obstruction path is eliminated without changing the selection overlay behavior. Thanks for the quick iteration and for clarifying that the same element details remain available in the top-left card — this is a clean simplification. 🙏


mrcfps (approved):
@1119302165 I reviewed the current changed ranges in `FileViewer.tsx`, `index.css`, and the locale fallback lists: removing the tooltip markup/styles keeps the selection overlay path intact while eliminating the overflow/obstruction case, and the added English fallback IDs cover the newly present `social-media-dashboard` skill and `totality-festival` design system across DE/RU/FR. I also verified `pnpm --filter @open-design/web typecheck` and `pnpm test tests/localized-content.test.ts` from `e2e` pass locally. Thanks for the quick iteration and for keeping the useful element details available outside the tooltip — this is a clean fix. 🙏


lefarcen:
@1119302165 Wow, thank you for fixing the CI failure! 🙌 You identified the root cause (missing localizations from PR #678) and added the entries for `social-media-dashboard` and `totality-festival` to de/ru/fr. That's exactly what was needed.

Your PR is now:
- ✅ Tooltip removal (clean)
- ✅ CI fixed (localization entries added)
- ✅ mrcfps APPROVED

Waiting for CI to re-run and confirm the fix passes. If it does, this should be ready to merge. Great work! 👏

1119302165:
@mrcfps https://github.com/nexu-io/open-design/pull/702  Is the code you just pushed to my branch implemented in a similar way to this PR? Could you inform the author of that PR and consolidate all follow-up work here instead? this pr merge fast. @lefarcen re-run CI.

mrcfps:
@1119302165 Yes — the i18n part I pushed here is the same fallback-list approach as #702: add `social-media-dashboard` and `totality-festival` to the de/ru/fr English-fallback registries so the existing localization coverage test passes without inventing unreviewed translations.

I kept the follow-up on this branch because #697 already needed a push to unblock its CI. I also added a small test-facing/accessibility label for the comment target overlay and fixed the e2e mock skill data so the updated full CI suite remains compatible after merging latest `main`.

I’ll leave a note on #702 that it duplicates the i18n portion now present here, so maintainers can close #702 if #697 lands first (or merge #702 first and then we can drop the duplicate fallback commits here if preferred).

A new CI run was already triggered by the latest push: https://github.com/nexu-io/open-design/actions/runs/25446173128

1119302165:
@mrcfps @lefarcen  now CI is Successful.

lefarcen:
Thanks for the contribution, @1119302165! 🎉

The tooltip removal + localization fix are now merged and live in `main`.

If you spot any issues or have more ideas, feel free to open another PR or issue anytime.

— open-design team

PR #701, opened 2026-05-06
fix: fix link handling in example preview iframe sandbox

### LINKED ISSUE #660: Clicking back on the blog-post example "<--Filebase blog" opens an open-design window in that dialog box
Image

meshackm:
### Fixes

- Added a click interceptor inside the preview iframe
- Anchor links (`href="#"`, `href="#section"`) scroll correctly within the preview, smooth scrolling works if enabled
- Links with `target="_blank"` are safely opened in a new tab, aided by adding the following iframe sandbox options:
	-  ` allow-popups allow-popups-to-escape-sandbox`

### DIFF
--- a/apps/web/src/components/PreviewModal.tsx
+++ b/apps/web/src/components/PreviewModal.tsx
@@ -389,7 +389,7 @@ export function PreviewModal({
                 <iframe
                   key={activeView?.id ?? 'view'}
                   title={`${title} ${activeView?.label ?? ''}`}
-                  sandbox="allow-scripts"
+                  sandbox="allow-scripts allow-popups allow-popups-to-escape-sandbox"
                   srcDoc={srcDoc}
                 />
               </div>
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -166,20 +166,28 @@ function injectSandboxShim(doc: string): string {
     var isAnchor = href.startsWith('#') || href === '';
     if (isAnchor) {
       e.preventDefault();
-      var targetId = href.slice(1);
-      var target = targetId ? document.getElementById(targetId) : null;
-      if (target) {
-        target.scrollIntoView();
-        location.hash === href && history.replaceState(null, '', ' ');
-        location.hash = href;
+      if (href === '' || href === '#') {
+        window.scrollTo({ top: 0 });
+        history.replaceState(null, '', ' ');
+      } else {
+        var targetId = href.slice(1);
+        var target = targetId ? document.getElementById(targetId) : null;
+        if (target) {
+          target.scrollIntoView();
+          location.hash === href && history.replaceState(null, '', ' ');
+          location.hash = href;
+        }
       }
     } else if (link.getAttribute('target') === '_blank') {
       e.preventDefault();
       let safe = false;
       try {
         var url = new URL(href, location.href);
-        safe = url.protocol === 'http:' || url.protocol === 'https:' || url.protocol === 'mailto:';
-      } catch(_) {}
+        safe =
+          url.protocol === 'http:' ||
+          url.protocol === 'https:' ||
+          url.protocol === 'mailto:';
+      } catch (_) {}
       safe && window.open(href, '_blank', 'noopener,noreferrer');
     }
   });
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -159,16 +159,28 @@ function injectSandboxShim(doc: string): string {
   tryShim('sessionStorage');
   document.addEventListener('click', (e) => {
     if (!e.target || !(e.target instanceof Element)) return;
-    const link = e.target.closest('a[href]');
+    var link = e.target.closest('a[href]');
     if (!link) return;
-    const href = link.getAttribute('href');
-    const isAnchor = href.startsWith('#') || href === '';
+    var href = link.getAttribute('href');
+    if (href === null) return;
+    var isAnchor = href.startsWith('#') || href === '';
     if (isAnchor) {
       e.preventDefault();
-      location.hash = href;
-    } else if (link.target === '_blank') {
+      var targetId = href.slice(1);
+      var target = targetId ? document.getElementById(targetId) : null;
+      if (target) {
+        target.scrollIntoView();
+        location.hash === href && history.replaceState(null, '', ' ');
+        location.hash = href;
+      }
+    } else if (link.getAttribute('target') === '_blank') {
       e.preventDefault();
-      window.open(href, '_blank', 'noopener,noreferrer');
+      let safe = false;
+      try {
+        var url = new URL(href, location.href);
+        safe = url.protocol === 'http:' || url.protocol === 'https:' || url.protocol === 'mailto:';
+      } catch(_) {}
+      safe && window.open(href, '_blank', 'noopener,noreferrer');
     }
   });
 })();</script>`;
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -129,6 +129,10 @@ function escapeAttr(value: string): string {
 // becomes a static, unnavigable preview. We install a same-origin
 // in-memory shim BEFORE any user script runs so those decks degrade
 // gracefully (position just doesn't persist across reloads).
+// allow-popups and allow-popups-to-escape-sandbox are needed for 
+// links with target="_blank" to work in the sandboxed preview.
+// Empty hrefs and hash only hrefs will be intercepted and ignored.
+// hrefs leading to an id on the page will be scrolled into view.
 function injectSandboxShim(doc: string): string {
   const shim = `<script data-od-sandbox-shim>(function(){
   function makeStore(){
@@ -153,6 +157,26 @@ function injectSandboxShim(doc: string): string {
   }
   tryShim('localStorage');
   tryShim('sessionStorage');
+  document.addEventListener('click', (e) => {
+    const link = e.target.closest('a[href]');
+    if (!link) return;
+    const href = link.getAttribute('href');
+    const rel = link.getAttribute('rel');
+    const isAnchor = href.startsWith('#') || href === ';
+    if (isAnchor) {
+      e.preventDefault();
+      document
+        .getElementById(href.slice(1))
+        ?.scrollIntoView();
+    } else if (link.target === '_blank') {
+      e.preventDefault();
+      if (rel === 'noopener noreferrer') {
+        window.open(href, '_blank');
+      } else {
+          window.open(href, '_blank', 'noopener,noreferrer');
+      }
+    }
+  });
 })();</script>`;
   if (/<head[^>]*>/i.test(doc))
     return doc.replace(/<head[^>]*>/i, (m) => `${m}${shim}`);
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -158,23 +158,17 @@ function injectSandboxShim(doc: string): string {
   tryShim('localStorage');
   tryShim('sessionStorage');
   document.addEventListener('click', (e) => {
+    if (!e.target || !(e.target instanceof Element)) return;
     const link = e.target.closest('a[href]');
     if (!link) return;
     const href = link.getAttribute('href');
-    const rel = link.getAttribute('rel');
     const isAnchor = href.startsWith('#') || href === '';
     if (isAnchor) {
       e.preventDefault();
-      document
-        .getElementById(href.slice(1))
-        ?.scrollIntoView();
+      location.hash = href;
     } else if (link.target === '_blank') {
       e.preventDefault();
-      if (rel === 'noopener noreferrer') {
-        window.open(href, '_blank');
-      } else {
-          window.open(href, '_blank', 'noopener,noreferrer');
-      }
+      window.open(href, '_blank', 'noopener,noreferrer');
     }
   });
 })();</script>`;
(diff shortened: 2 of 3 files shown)


lefarcen:
Hi @meshackm! 🎉
Thanks for the contribution — this fix for the iframe link navigation is addressing a real usability issue.

— open-design team

lefarcen (changes requested):
Hi @meshackm, thanks for tackling this iframe navigation issue! The approach of adding a click interceptor is sound. However, I found **1 critical bug** that would break the entire sandbox shim, plus **2 security/correctness issues**:

**Critical (P1):**
- `apps/web/src/runtime/srcdoc.ts` line 165: Invalid JavaScript syntax — `href === ';` is an unterminated string literal. This breaks the entire injected `<script data-od-sandbox-shim>`, so the existing `localStorage`/`sessionStorage` shims also stop working. Fix to `href === ''`

**Security/correctness (P2):**
- `apps/web/src/runtime/srcdoc.ts` line 166: Hash-link handler incomplete — doesn't update `location.hash`, breaking templates using hash state, `hashchange` events, history/back, CSS `:target`, named anchors. Suggest: `location.hash = href;`
- `apps/web/src/runtime/srcdoc.ts` line 173: `rel` attribute doesn't protect `window.open()`. Always pass `'noopener,noreferrer'` as features

**Minor (P3):**
- `apps/web/src/runtime/srcdoc.ts` line 160: `e.target` may not be Element. Add `instanceof` check before `closest()`

Let me know if you need clarification on any of these!

lefarcen (changes requested):
Hi @meshackm! I need to correct my previous review — the P1 I mentioned (unterminated string literal) was a false positive; your code is syntactically correct on line 165. Apologies for the confusion!

However, the **2 security/correctness issues (P2)** still remain:

**P2 — Hash navigation incomplete:**
- Line 166-170: The hash-link handler prevents default but only calls `scrollIntoView()`. This doesn't update `location.hash`, so templates using hash state, `hashchange` events, history/back, CSS `:target`, and named anchors will break. Native fragment navigation handles all of this automatically.
- Fix: Use `location.hash = href;` instead of manual scrolling, or allow native behavior for valid same-document hashes.

**P2 — window.open opener leak:**
- Line 173-178: The logic checks `rel === 'noopener noreferrer'` (exact string match) and OMITS the `noopener,noreferrer` features when it matches. This is backwards — the `rel` attribute does NOT protect a programmatic `window.open()` call. You must always pass `'noopener,noreferrer'` as the third argument.
- Fix: `window.open(href, '_blank', 'noopener,noreferrer')` for ALL `_blank` links, regardless of `rel`.

**P3 — Element check:**
- Line 161: `e.target.closest('a[href]')` assumes the target is an `Element`. Click targets can be text nodes, which would throw. Add `instanceof Element` check first.

Let me know if you need clarification on any of these!

meshackm:
I understand, let me look into it. There was indeed a missing single quote, I added it before your first review.

meshackm:
I found an issue where clicking a hash link, scrolling away, and then clicking it again fails to trigger the scroll to element behavior, looking into it

apps/web/src/runtime/srcdoc.ts
```diff
@@ -153,6 +157,20 @@ function injectSandboxShim(doc: string): string {
...
   tryShim('sessionStorage');
+  document.addEventListener('click', (e) => {
+    if (!e.target || !(e.target instanceof Element)) return;
+    const link = e.target.closest('a[href]');
+    if (!link) return;
+    const href = link.getAttribute('href');
+    const isAnchor = href.startsWith('#') || href === '';
+    if (isAnchor) {
+      e.preventDefault();
+      location.hash = href;
```
lefarcen: **P2** — When `location.hash` is already `href`, setting it again does NOT scroll. Use `scrollIntoView()` explicitly:
```javascript
const targetId = href.slice(1);
const target = targetId ? document.getElementById(targetId) : null;
if (target) {
  target.scrollIntoView({ behavior: 'smooth' });
  location.hash = href;
}
```

lefarcen (changes requested):
Hi @meshackm! Great work addressing the previous findings — the Element check, `location.hash` update, and `noopener,noreferrer` are all correct now. 👍

However, the issue you just discovered is real:

**P2 — Same-hash re-click doesn't scroll:**
Line 168: `location.hash = href;` — when `location.hash` is already equal to `href`, browsers do NOT fire a `hashchange` event or scroll. This breaks the "click same anchor link → scroll back" UX.

**Fix:** Force scroll via `scrollIntoView()` BEFORE setting `location.hash`:
```javascript
if (isAnchor) {
  e.preventDefault();
  const targetId = href.slice(1); // remove '#'
  const target = targetId ? document.getElementById(targetId) : null;
  if (target) {
    target.scrollIntoView({ behavior: 'smooth' });
    location.hash = href; // Update hash after scroll for history/CSS :target
  }
}
```

This way, clicking the same hash link multiple times always scrolls. Let me know if you need clarification!

apps/web/src/runtime/srcdoc.ts
mrcfps: This hash-link branch still misses a common navigation path: when the user clicks the same `#fragment` again after scrolling away, `location.hash = href` is a no-op because the hash already matches, so the preview never scrolls back to the target. That means the main anchor-link behavior this PR is trying to restore is still broken for repeated in-page navigation. Please preserve native same-document fragment behavior here, or explicitly handle the `location.hash === href` case by scrolling the target / named anchor into view before returning.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -153,6 +157,20 @@ function injectSandboxShim(doc: string): string {
...
+    } else if (link.target === '_blank') {
+      e.preventDefault();
+      window.open(href, '_blank', 'noopener,noreferrer');
```
mrcfps: The `_blank` branch still forwards the raw `href` string straight into `window.open(...)`. In this PR the iframe sandbox now includes `allow-popups-to-escape-sandbox`, so a preview can use `href="javascript:..."` or `data:` on a `_blank` link to spawn unsandboxed script in a new tab. That is a concrete security regression in the changed path. Please resolve the URL against `document.baseURI` and allowlist only the schemes we intentionally support before opening it, and ignore or block everything else.

mrcfps (changes requested):
@meshackm Thanks for iterating on this — I re-reviewed the latest preview iframe changes and found two blockers that still affect the link-handling path and sandbox safety.


lefarcen (dismissed):
Hi @meshackm! All three prior findings are addressed in this commit:

✅ **Element check** (line 161) — guards against text node clicks
✅ **Hash navigation** (line 168) — now uses `location.hash = href` so `hashchange` events, CSS `:target`, history/back, and named anchors all work
✅ **Opener leak** (line 171) — always passes `'noopener,noreferrer'` regardless of `rel` attribute

**Re: your observation about same-hash not scrolling again** — that's expected browser behavior. When `location.hash` is already set to `#section`, setting it again to `#section` doesn't fire `hashchange` or re-scroll. Fixing this would need extra logic (detect same-hash + force `scrollIntoView()`), which could be a follow-up enhancement if needed. For now, the core fix is solid.

apps/web/src/runtime/srcdoc.ts
meshackm: I've added a protocol allowlist; http, https and mailto

lefarcen (dismissed):
Both security and functional issues resolved — protocol allowlist (http/https/mailto) blocks javascript: URIs, and same-hash re-click now calls scrollIntoView() before setting location.hash. Clean to ship. 🎉

apps/web/src/runtime/srcdoc.ts
```diff
@@ -153,6 +157,32 @@ function injectSandboxShim(doc: string): string {
...
+    if (!e.target || !(e.target instanceof Element)) return;
+    var link = e.target.closest('a[href]');
+    if (!link) return;
+    var href = link.getAttribute('href');
+    if (href === null) return;
+    var isAnchor = href.startsWith('#') || href === '';
+    if (isAnchor) {
+      e.preventDefault();
+      var targetId = href.slice(1);
+      var target = targetId ? document.getElementById(targetId) : null;
```
mrcfps: This anchor branch still breaks the exact `href="#"` case called out in the PR description. Here `targetId` becomes `''`, `target` stays `null`, and the handler has already `preventDefault()`'d, so clicking `#` never scrolls back to the top of the preview. That leaves one of the main-path anchor-link behaviors unfixed in the changed code. Please special-case empty fragments (and any other fragment with no `id` match) by preserving the browser's fragment behavior — for example, scroll to the top for `#`, or fall back to `location.hash = href` / native navigation instead of returning early.

apps/web/src/components/PreviewModal.tsx
```diff
@@ -389,7 +389,7 @@ export function PreviewModal({
                 <iframe
                   key={activeView?.id ?? 'view'}
                   title={`${title} ${activeView?.label ?? ''}`}
-                  sandbox="allow-scripts"
+                  sandbox="allow-scripts allow-popups allow-popups-to-escape-sandbox"
```
mrcfps, line 392: Changing the sandbox string here also requires updating the existing web assertions in `apps/web/tests/components/PreviewModal.test.tsx:23` and `apps/web/tests/components/PreviewModal.test.tsx:45`, which still hard-code `sandbox="allow-scripts"`. As written, this PR leaves the checked-in test suite out of sync with the new behavior, so the package test run will fail once dependencies are installed. Please update those expectations, and ideally add a focused assertion for the popup flags, in the same change so the new sandbox contract stays covered.

mrcfps (changes requested):
@meshackm Thanks for the continued iteration on this — I re-reviewed the latest head and found two blockers still affecting the preview-link path and its test coverage.


apps/web/src/components/PreviewModal.tsx
meshackm: Can I update the test suite accordingly?

apps/web/src/runtime/srcdoc.ts
meshackm: Aha I see, I'll make `href='#'` scroll to top

apps/web/src/components/PreviewModal.tsx
mrcfps, line 392: Changing the iframe sandbox here also changes the component contract that `apps/web/tests/components/PreviewModal.test.tsx:23` and `apps/web/tests/components/PreviewModal.test.tsx:45` assert today: both tests still expect `sandbox="allow-scripts"`. Because this PR updates the rendered attribute to include popup flags but leaves those checked-in expectations untouched, the web test suite will fail once dependencies are installed, so the change is not actually merge-ready on the current head. Please update those assertions to the new sandbox string and keep at least one focused expectation covering the popup flags in the same PR.

mrcfps (changes requested):
@meshackm Thanks for the iteration here — I re-checked the latest head and found one remaining blocker in the updated preview iframe path.


apps/web/src/components/PreviewModal.tsx
meshackm: Got it, on it...

meshackm: It's done

apps/web/src/runtime/srcdoc.ts
```diff
@@ -153,6 +157,40 @@ function injectSandboxShim(doc: string): string {
...
+      e.preventDefault();
+      let safe = false;
+      try {
+        var url = new URL(href, location.href);
+        safe =
+          url.protocol === 'http:' ||
+          url.protocol === 'https:' ||
+          url.protocol === 'mailto:';
+      } catch (_) {}
+      safe && window.open(href, '_blank', 'noopener,noreferrer');
```
mrcfps, line 191: Resolving `_blank` URLs against `location.href` regresses relative links for the preview documents that already rely on `baseHref` in this file. In a `srcdoc` iframe `location.href` is `about:srcdoc`, so `new URL("guide.html", location.href)` throws; after `preventDefault()` that leaves `safe` false and the click opens nothing. That means example/file previews with `target="_blank"` relative links still break on the changed path even though the browser would normally resolve them against the injected `<base>` tag. Please resolve against `document.baseURI` (or the parsed `url.href`) so the sandbox honors `baseHref`, and add a regression in `apps/web/tests/runtime/srcdoc.test.ts` that covers a relative `_blank` link with `baseHref`.

mrcfps (commented):
@meshackm Thanks for the update here — I re-checked the latest head and found one remaining actionable issue in the preview link handling path.


lefarcen (commented):
@meshackm Nice work — all of mrcfps's findings are resolved on this head:

✅ **Test suite sync**: Both assertions at `:23` and `:45` now expect the updated `sandbox="allow-scripts allow-popups allow-popups-to-escape-sandbox"` attribute
✅ **Popup flags coverage**: Added a focused test that explicitly asserts the presence of `allow-popups` and `allow-popups-to-escape-sandbox`
✅ **No new issues**: The click interceptor logic (protocol allowlist, `href="#"` scroll-to-top, hash navigation) remains solid

This looks ready from my side — deferring final approval to a maintainer.

mrcfps (approved):
Approved.

lefarcen:
Merged! 🎉 Thanks @meshackm for the contribution to open-design. Looking forward to the next one!

leilei926524-tech:
Hi @meshackm — welcome to Open Design contributors. 🎉

Your first merged PR (#701) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #707, opened 2026-05-06
add OpenRouter provider preset

kenrogers:
## Why

OpenRouter already works through Open Design's OpenAI-compatible proxy when users know the custom base URL, but it was not available as a quick-fill provider in Settings. Adding a preset makes the existing BYOK path discoverable without introducing a new adapter or dependency.

## What changed

- Added OpenRouter to the known OpenAI-compatible provider presets with `https://openrouter.ai/api/v1` and current model suggestions.
- Documented the OpenRouter API-mode setup in `QUICKSTART.md`.
- Added focused coverage for the OpenRouter preset and OpenAI-compatible routing.

## Validation

- `pnpm install`
- `pnpm --filter @open-design/web test -- apps/web/tests/components/SettingsDialog.test.ts apps/web/tests/providers/openai-compatible.test.ts`
- `pnpm guard`
- `pnpm typecheck`

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -141,7 +141,7 @@ Both modes feed the **same** `<artifact>` parser and the **same** sandboxed ifra
 
 ### OpenRouter API mode
 
-OpenRouter uses the OpenAI-compatible API path. In **Settings -> Execution -> API mode**, choose **OpenAI**, pick **OpenRouter** from quick-fill providers, paste an OpenRouter API key, and select a model such as `openai/gpt-5.2` or `openrouter/auto`.
+OpenRouter uses the OpenAI-compatible API path. Create an API key from [OpenRouter Keys](https://openrouter.ai/settings/keys), then in **Settings -> Execution -> API mode**, choose **OpenAI**, pick **OpenRouter** from quick-fill providers, paste the key, and select a model such as `openai/gpt-5.2` or `openrouter/auto`.
 
 The preset fills:
 
@@ -149,6 +149,8 @@ The preset fills:
 Base URL: https://openrouter.ai/api/v1
 Model: openai/gpt-5.2
 ```
+
+If you configure OpenRouter manually instead of using the preset, make sure the protocol tab is **OpenAI** before pasting `https://openrouter.ai/api/v1`; the Anthropic and Gemini protocol tabs use different upstream request shapes. OpenRouter 401, 402, or 429 responses usually mean the API key, account credits, free-model availability, or rate limits need attention on the OpenRouter side rather than in Open Design.
 
 ## Prompt composition
 
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -139,6 +139,17 @@ location /api/ {
 
 Both modes feed the **same** `<artifact>` parser and the **same** sandboxed iframe. The only thing that differs is the transport and the system-prompt delivery (local CLIs have no separate system channel, so the composed prompt is folded into the user message).
 
+### OpenRouter API mode
+
+OpenRouter uses the OpenAI-compatible API path. In **Settings -> Execution -> API mode**, choose **OpenAI**, pick **OpenRouter** from quick-fill providers, paste an OpenRouter API key, and select a model such as `openai/gpt-5.2` or `openrouter/auto`.
+
+The preset fills:
+
+```text
+Base URL: https://openrouter.ai/api/v1
+Model: openai/gpt-5.2
+```
+
 ## Prompt composition
 
 For every send, the app builds a system prompt from three layers and sends it to the provider:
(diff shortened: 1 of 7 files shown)


lefarcen:
Hi @kenrogers! 🎉
Thanks for the contribution — adding OpenRouter as a preset makes BYOK more discoverable.

— open-design team

apps/web/tests/components/SettingsDialog.test.ts
```diff
@@ -93,6 +93,22 @@ describe('SettingsDialog API protocol switching', () => {
     });
   });
 
+  it('keeps OpenRouter as a selectable OpenAI-compatible provider preset', () => {
```
lefarcen, line 97: P2: This test manually patches baseUrl/model, so it would still pass if OpenRouter were removed from KNOWN_PROVIDERS. Consider asserting the KNOWN_PROVIDERS entry or simulating actual provider selection from the Settings UI to prove discoverability.

apps/web/src/state/config.ts
```diff
@@ -130,6 +130,19 @@ export const KNOWN_PROVIDERS: KnownProvider[] = [
     model: 'gpt-4o',
     models: ['gpt-4o', 'gpt-4o-mini', 'o3', 'o4-mini'],
   },
+  {
+    label: 'OpenRouter',
+    protocol: 'openai',
+    baseUrl: 'https://openrouter.ai/api/v1',
+    model: 'openai/gpt-5.2',
```
lefarcen, line 137: P2: The model IDs here (openai/gpt-5.2, openrouter/auto, etc.) may not match modelMaxTokensDefault lookup keys in litellm-models.json. If they fall back to 8192, large artifact responses could be truncated. Consider explicit overrides or OpenRouter-aware normalization.

apps/web/tests/providers/openai-compatible.test.ts
```diff
@@ -16,6 +16,11 @@ describe('isOpenAICompatible', () => {
     expect(isOpenAICompatible('mimo-v2.5-pro', 'https://token-plan-cn.xiaomimimo.com/v1')).toBe(true);
   });
 
+  it('routes OpenRouter through the OpenAI-compatible chat completions proxy', () => {
```
lefarcen, line 19: P3: This routing test only covers the isOpenAICompatible heuristic. It doesn't prove the OpenRouter preset actually flows through /api/proxy/openai/stream at runtime. An integration-style provider test using the preset config would catch regressions.

QUICKSTART.md
```diff
@@ -139,6 +139,17 @@ location /api/ {
 
 Both modes feed the **same** `<artifact>` parser and the **same** sandboxed iframe. The only thing that differs is the transport and the system-prompt delivery (local CLIs have no separate system channel, so the composed prompt is folded into the user message).
 
+### OpenRouter API mode
+
+OpenRouter uses the OpenAI-compatible API path. In **Settings -> Execution -> API mode**, choose **OpenAI**, pick **OpenRouter** from quick-fill providers, paste an OpenRouter API key, and select a model such as `openai/gpt-5.2` or `openrouter/auto`.
```
lefarcen: P3: Docs don't tell users where to get an OpenRouter API key or that account credits/free-model availability can affect success. New users hitting 401/402/429 won't have enough context to distinguish bad app config from OpenRouter account/rate-limit issues.

lefarcen: P3: Missing manual-setup warning: if a user pastes https://openrouter.ai/api/v1 while still on Anthropic/custom protocol, requests could go through the wrong proxy shape and fail. Add a note that protocol must be OpenAI.

lefarcen (commented):
Hey @kenrogers! Thanks for adding OpenRouter — this makes BYOK more discoverable. 👍

I've reviewed the config, tests, and docs. The core implementation is solid, but I found a few gaps in test coverage and documentation that would help users avoid setup pitfalls:

**Main findings:**
- The settings test doesn't actually prove the OpenRouter preset appears in the UI dropdown
- Model IDs may not match the token-limit lookup, which could truncate large artifacts
- Docs don't mention where to get an API key or warn about OpenRouter-specific account/rate-limit errors

See inline comments for details. None of these block merge, but addressing them would strengthen the feature.

PerishCode:
@kenrogers — checking in on the OpenRouter preset PR. It's been about 4 days since lefarcen left feedback and we're holding for your next iteration. Still in flight? Drop a comment if something in the review thread needs clarification, or let us know if you'd like to close this in favor of another approach.

(Tagged `awaiting-author-response-24h` by maintainer tooling.)

kenrogers:
Thanks for the review. I pushed `c20f6ce` with updates for the feedback:

- assert the OpenRouter entry in `KNOWN_PROVIDERS` so the preset is covered directly
- normalize OpenRouter model IDs through the vendored LiteLLM `openrouter/...` aliases so preset models avoid the 8192 fallback
- add an integration-style provider test proving the OpenRouter preset posts to `/api/proxy/openai/stream` with the expected token default
- expand the Quickstart with the OpenRouter keys link, account/credit/rate-limit note, and the manual setup warning to stay on the OpenAI protocol tab

Validation passed locally: web tests, web typecheck, `pnpm guard`, and root `pnpm typecheck`.

PR #708, opened 2026-05-06
fix(web): align design system default test fixture

mrcfps:
## Summary
- Updates the NewProjectPanel default design system test fixture so the mocked prototype skill supports design systems.
- Keeps the test focused on default design system selection instead of the separate opt-out picker path.

## Validation
- pnpm --filter @open-design/web test -- tests/components/NewProjectPanel.test.tsx
- pnpm guard
- pnpm typecheck

Fixes the CI failure from https://github.com/nexu-io/open-design/actions/runs/25453505130/job/74676719692.

### DIFF
--- a/apps/web/tests/components/NewProjectPanel.test.tsx
+++ b/apps/web/tests/components/NewProjectPanel.test.tsx
@@ -16,7 +16,7 @@ const skills: SkillSummary[] = [
     mode: 'prototype',
     surface: 'web',
     previewType: 'html',
-    designSystemRequired: false,
+    designSystemRequired: true,
     defaultFor: ['prototype'],
     triggers: [],
     upstream: null,


lefarcen (commented):
The fixture change correctly aligns with the prototype skill's actual `designSystemRequired: true` behavior.

One optional improvement: the test suite currently lacks explicit coverage for the `designSystemRequired === false` opt-out path in `NewProjectPanel`. Consider adding a small render test with that fixture to prevent regressions in the picker-hiding logic.

CI fix LGTM.

mrcfps (commented):
@mrcfps I reviewed the focused NewProjectPanel test fixture change that now marks the mocked prototype skill as design-system-capable, and I checked the surrounding default design system selection path. The update aligns the test setup with the behavior it verifies; thanks for tightening this regression coverage and keeping the fix nicely scoped. 🙌


lefarcen (approved):

PR #674, opened 2026-05-06
Fix Ukrainian prompt template translations

alchemistklk:
## Summary
- Add missing Ukrainian translations for prompt template source filtering labels.

## Tests
- git diff --check
- Not run: pnpm i18n:check / typecheck because node_modules are not installed in this worktree (`tsx` and workspace dependencies are missing).

PR #570, opened 2026-05-05
docs(readme): document OD_DATA_DIR + migration from .od/ to Desktop app

StotheC90:
## Summary

The "Move it elsewhere" row in `README.md` § *First-run state (`./.od/`)* still said the path is hard-coded; `OD_DATA_DIR` (resolveDataDir in `apps/daemon/src/server.ts`) has supported relocation since the runtime-data refactor. This PR:

- Replaces the "not supported yet" note in the table with the actual env-var usage and resolution semantics, and mentions `OD_MEDIA_CONFIG_DIR` for the narrower credentials override.
- Adds a *Migrating a pre-desktop-app `.od/`* sub-section so users who started in the repo and later installed the packaged Desktop app know the two writers target different roots, how to copy projects/SQLite/artifacts/`media-config.json` over after quitting cleanly, and how to keep both writers on the same dir going forward.

Documentation only, no code changes. Tested by walking through the exact migration on macOS arm64 with `0.3.1-beta.5` against an existing v0.1.0 `.od/` (22 projects + 5.9 MB SQLite); SQLite was migrated by the daemon on boot without errors.

## Test plan

- [x] Manual: rendered the new section in the GitHub README preview
- [x] Verified `OD_DATA_DIR` resolution against `apps/daemon/src/server.ts:resolveDataDir` and the existing README Storage row at line 296
- [x] Skipped `pnpm guard` / `pnpm typecheck` — docs-only change, no source touched

PR #712, opened 2026-05-06
fix(daemon): add OD_LEGACY_DATA_DIR migrator for 0.3.x -> 0.4.x recovery (closes #710)

### LINKED ISSUE #710: Lost chat and designs on version update
I´ve installed 0.3.5 and create a design by chat
where installed 0.4.1 i lost my chat e my designs. how i can restore?

Nagendhra-web:
Closes #710.

## Problem

Users who ran the 0.3.x repo dev-server, then installed the 0.4.x packaged Desktop app, lost access to all their chats and designs on launch. The data wasn't deleted: 0.3.x wrote to `<repo>/.od/` and 0.4.x writes to `<userData>/namespaces/<channel>/data/`. The new daemon had no way to find the old payload, and the existing manual recovery procedure documented in #570 is a multi-step shell pipeline most users won't run.

## Fix

`apps/daemon/src/legacy-data-migrator.ts` is a small synchronous one-shot copier. When `OD_LEGACY_DATA_DIR` is set on daemon boot AND the new data root has no `app.sqlite` AND the legacy dir has one, it copies the OD payload across before SQLite opens, then writes a `.migrated-from` marker so subsequent boots are no-ops.

```bash
# macOS / Linux
OD_LEGACY_DATA_DIR="/path/to/old/repo/.od" open "/Applications/Open Design.app"
# Windows (PowerShell)
$env:OD_LEGACY_DATA_DIR="C:\path\to\old\repo\.od"
& "$env:LOCALAPPDATA\Programs\Open Design\Open Design.exe"
```

Payload copied: `app.sqlite` (+ `-shm`/`-wal`), `app-config.json`, `media-config.json`, `projects/`, `artifacts/`, `connectors/`, `composio/`. Anything outside this list (logs, sockets, lockfiles, OS scratch) is left behind.

Sync by design: `openDatabase` below would race an async copy and either silently lose the migration or corrupt the new file.

The migrator refuses to run when:
- `OD_LEGACY_DATA_DIR` is unset / empty / equals `OD_DATA_DIR`
- the legacy dir has no `app.sqlite` (proves it's not a real OD payload)
- the new data dir already has `app.sqlite` (never overwrites real data)
- a `.migrated-from` marker is already present (idempotent re-runs)

## Wiring

`apps/daemon/src/server.ts` calls `migrateLegacyDataDirSync({ legacyDir: process.env.OD_LEGACY_DATA_DIR, dataDir: RUNTIME_DATA_DIR })` right after `RUNTIME_DATA_DIR` resolves and before `openDatabase`.

## Docs

`README.md` has a new **Option A: one-shot auto-migration via `OD_LEGACY_DATA_DIR`** subsection under the existing _Migrating a pre-desktop-app `.od/` into the installed Desktop app_ block. The manual copy procedure stays as Option B for users who want full control.

## Tests

`apps/daemon/tests/legacy-data-migrator.test.ts`: 17 hermetic regressions covering noop, skip, migrate, marker, idempotency, partial payload, missing dataDir, and the two predicate helpers (`dataDirIsEmptyOrFresh`, `legacyDirHasPayload`). All run against `mkdtemp()` pairs and clean up in `afterEach` so no developer or CI runner state leaks.

## Incidental

Dropped duplicate `uk.ts` `promptTemplates.allSources` / `sourceFilterAria` entries that `origin/main` accumulated after #674 and #626 both shipped them. Without this drop the workspace web typecheck fails on tsc TS1117. Two-line removal in `apps/web/src/i18n/locales/uk.ts`.

## Test plan

- [x] `pnpm --filter @open-design/contracts typecheck` clean
- [x] `pnpm --filter @open-design/daemon typecheck` clean
- [x] `pnpm --filter @open-design/web typecheck` clean
- [x] `pnpm --filter @open-design/daemon exec vitest run tests/legacy-data-migrator.test.ts` 17/17
- [x] `pnpm guard` clean
- [x] No em dashes in source, comments, or commit messages
- [x] No `Co-authored-by` trailers; author = `Nagendhra <nagendhra405@gmail.com>`
- [x] New tests under `apps/daemon/tests/` per the directory convention from #496


apps/daemon/src/server.ts
```diff
@@ -686,6 +687,15 @@ export function resolveDataDir(raw, projectRoot) {
   return resolved;
 }
 const RUNTIME_DATA_DIR = resolveDataDir(process.env.OD_DATA_DIR, PROJECT_ROOT);
+// One-shot legacy data migration. When OD_LEGACY_DATA_DIR is set and the
+// new data root is fresh (no app.sqlite), copy the 0.3.x .od/ payload
+// across before SQLite opens. Synchronous on purpose: openDatabase below
+// would race an async copy. See apps/daemon/src/legacy-data-migrator.ts
+// and https://github.com/nexu-io/open-design/issues/710.
+migrateLegacyDataDirSync({
+  legacyDir: process.env.OD_LEGACY_DATA_DIR,
```
mrcfps, line 696: `apps/daemon/src/server.ts` now reads `process.env.OD_LEGACY_DATA_DIR` here, but the packaged Desktop flow targeted by this PR launches the daemon sidecar through `apps/packaged/src/sidecars.ts`, where `resolvePackagedChildBaseEnv()` only forwards the small packaged-child allowlist plus provider secrets before applying the managed `OD_DATA_DIR`/`OD_RESOURCE_ROOT` values. Because `OD_LEGACY_DATA_DIR` is not in that forwarded env, a user can follow the new Desktop migration instructions and still have this value arrive as `undefined`, causing `migrateLegacyDataDirSync()` to return `noop` and leaving the old `.od/` data unmigrated. Please explicitly forward `OD_LEGACY_DATA_DIR` for the daemon sidecar (and add a packaged launch-env regression) so the new migrator actually runs in the primary 0.3.x → 0.4.x packaged upgrade path.

mrcfps (changes requested):
@Nagendhra-web Thanks for putting together the one-shot migrator and the focused daemon tests. I found one blocking integration issue in the packaged Desktop launch path: the daemon reads the new env var, but the packaged sidecar launcher does not currently forward it, so the target upgrade recovery flow would still no-op until that env is wired through. 🙏


apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,193 @@
...
+  }
+
+  const markerPath = path.join(dataDir, MARKER_FILE);
+  if (fs.existsSync(markerPath)) {
+    return { status: 'skipped', reason: 'migration marker already present' };
+  }
+
+  log.info(`migrating legacy data from "${legacyDir}" to "${dataDir}"`);
+  const copied = copyPayload(legacyDir, dataDir);
+  writeMarker(dataDir, legacyDir);
```
lefarcen: **P1 — Partial-copy corruption risk**

Migration writes directly into final `dataDir`, copies `app.sqlite` first, then other entries. If `fs.cpSync` fails mid-copy (disk full, permissions, sidecar killed), the target can contain `app.sqlite` but miss `projects/`, `artifacts/`, credentials, or the marker. Next boot sees `app.sqlite` and skips — user's data is now incomplete and unrecoverable via this path.

**Fix:** Copy into a sibling staging directory (e.g. `<dataDir>-migrate-staging`), verify all expected entries copied, write marker there, then atomically rename/move to final `dataDir`. Add test that injects copy failure after `app.sqlite` and verifies staging cleanup + next-boot retry.

apps/daemon/src/server.ts
```diff
@@ -686,6 +687,15 @@ export function resolveDataDir(raw, projectRoot) {
   return resolved;
 }
 const RUNTIME_DATA_DIR = resolveDataDir(process.env.OD_DATA_DIR, PROJECT_ROOT);
+// One-shot legacy data migration. When OD_LEGACY_DATA_DIR is set and the
+// new data root is fresh (no app.sqlite), copy the 0.3.x .od/ payload
+// across before SQLite opens. Synchronous on purpose: openDatabase below
+// would race an async copy. See apps/daemon/src/legacy-data-migrator.ts
+// and https://github.com/nexu-io/open-design/issues/710.
+migrateLegacyDataDirSync({
```
lefarcen, line 695: **P1 — Packaged env won't reach here**

`OD_LEGACY_DATA_DIR` won't propagate from macOS/Windows shell to the daemon sidecar. `apps/packaged/src/sidecars.ts:151-196` forwards only a small env allowlist + provider secrets; `OD_LEGACY_DATA_DIR` is dropped before this code runs. README's macOS command (`VAR=value open ...`) sets env for `open` process only, not the launched GUI app.

**Fix:** Explicitly forward `OD_LEGACY_DATA_DIR` in `sidecars.ts` env allowlist. Add integration test around packaged env propagation. Change macOS docs to either use the app executable directly or `launchctl setenv` (which persists for GUI launches).

apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,193 @@
...
+ *
+ * This module gives operators a recovery path. When `OD_LEGACY_DATA_DIR`
+ * is set, the daemon checks on boot whether the new data root is empty
+ * AND whether the legacy directory contains a real OD payload (a
+ * `app.sqlite` is the proof). If both hold, content is copied across.
+ * The migration writes a `.migrated-from` marker so subsequent boots
+ * skip the work even if the user clears the new directory; if they
+ * really want to re-migrate they can delete the marker.
+ *
+ * Sync by design: this runs at module import time in server.ts, before
```
lefarcen, line 35: **P2 — Packaged sidecar timeout during large copy**

Sync-before-SQLite is correct for DB races, but not safe with packaged sidecar's 35s status timeout (`apps/packaged/src/sidecars.ts:74-79`). Migrator runs before daemon IPC status server exists. Large `.od/projects` or `.od/artifacts` copy makes parent timeout and kill child mid-migration.

**Fix:** Either make packaged startup aware of long-migration status (emit progress during copy), increase/disable timeout while migration is active, or switch to atomic staged migration (see P1 fix) which can be fast-failed before status wait.

apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,193 @@
...
+
+/**
+ * Returns true when `dataDir` looks like a fresh / never-used data root:
+ * either the directory does not exist, or it exists but does not contain
+ * a real OD SQLite database. We deliberately do NOT just check for an
+ * empty directory: a packaged install creates an empty `dataDir` before
+ * the daemon ever boots, so emptiness is the common case rather than a
+ * proof of "no prior data."
+ */
+export function dataDirIsEmptyOrFresh(dataDir: string): boolean {
```
lefarcen, line 131: **P2 — Merge-mode can produce inconsistent state**

"Fresh" only checks `!app.sqlite`, but `copyPayload` uses `force: false` and merges into any existing target entries. Stale `projects/`, `artifacts/`, `app.sqlite-wal`, `app.sqlite-shm` can survive and mix with migrated DB, producing inconsistent SQLite/WAL pair or hybrid project tree.

**Fix:** Require target to have **none** of the known payload entries, or stage and replace payload as a unit (see P1 fix) instead of merging piecemeal.

apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,193 @@
...
+ * Copy each known OD payload entry from `legacyDir` to `dataDir`. Missing
+ * entries are skipped silently (e.g. an install that never ran the media
+ * surface won't have `media-config.json`). Returns the list of entries
+ * that were actually copied.
+ */
+function copyPayload(legacyDir: string, dataDir: string): string[] {
+  fs.mkdirSync(dataDir, { recursive: true });
+  const copied: string[] = [];
+  for (const entry of PAYLOAD_ENTRIES) {
+    const src = path.join(legacyDir, entry);
```
lefarcen, line 191: **P2 — Symlink escape risk**

Copier doesn't reject symlinks in source or pre-existing target paths. `fs.cpSync` can preserve symlinks; later `projects.ts:284-288` reads resolve safe-looking paths with `path.resolve` but then `readFile`/`stat` follow symlinks, so migrated project can contain links escaping data root.

**Fix:** Walk payload trees with `lstat` and reject symlinks before copying. Add migrator tests for source and target symlinks.

apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,193 @@
...
+    return { status: 'noop', reason: 'OD_LEGACY_DATA_DIR not set' };
+  }
+  const legacyDir = path.resolve(raw);
+  const dataDir = path.resolve(options.dataDir);
+
+  if (legacyDir === dataDir) {
+    return { status: 'noop', reason: 'OD_LEGACY_DATA_DIR equals OD_DATA_DIR' };
+  }
+
+  if (!legacyDirHasPayload(legacyDir)) {
```
lefarcen, line 333: **P2 — Wrong path only logs warning, daemon continues**

When `OD_LEGACY_DATA_DIR` is explicitly set but invalid (no `app.sqlite`), daemon only logs warning and continues into empty/new workspace. Users who mistype path see "missing data" again unless they find daemon logs.

**Fix:** When env var is explicitly set but invalid, fail startup or surface visible Desktop error with exact path + "expected `app.sqlite` here". Document that behavior in README.

README.md
```diff
@@ -372,6 +372,22 @@ If you ran the repo first and only later installed the packaged Desktop app, the
 
 > **⚠️ Do this in a clean state.** Migration replaces (not merges) the Desktop app's data dir with your repo `.od/`. Both writers must be fully stopped before copying — quit the Desktop app **and** stop the repo dev-server. SQLite-WAL needs to flush cleanly on both sides; if either daemon is still running it can write SQLite/WAL pages or project/artifact files mid-snapshot, leaving the staged copy inconsistent. If the Desktop app already has projects you care about, decide which side is authoritative before continuing — the steps below back up the Desktop's current `data/` to a sibling but do not merge.
 
+##### Option A: one-shot auto-migration via `OD_LEGACY_DATA_DIR`
+
+If your Desktop app's `data/` is still empty (typical after the upgrade that surfaced [#710](https://github.com/nexu-io/open-design/issues/710)), point the daemon at your old `.od/` once and let it copy the payload across on first boot. The daemon refuses to overwrite a `data/` that already has `app.sqlite`, and writes a `.migrated-from` marker so subsequent boots are no-ops.
```
lefarcen: **P2 — Option A guidance misses common case**

Directionally defensible, but misses common case where user already launched Desktop once to discover regression. That launch likely created `data/app.sqlite`, so Option A will refuse to run even if DB is empty.

**Fix:** State "choose Option A only before first successful Desktop daemon boot; if Desktop already created `app.sqlite`, use Option B or intentionally back up/delete empty Desktop data dir first."

README.md
```diff
@@ -372,6 +372,22 @@ If you ran the repo first and only later installed the packaged Desktop app, the
...
+# Quit the Desktop app first, then re-launch with this env set.
+# macOS/Linux:
+OD_LEGACY_DATA_DIR="/path/to/old/repo/.od" open "/Applications/Open Design.app"
+# Windows (PowerShell):
+$env:OD_LEGACY_DATA_DIR="C:\path\to\old\repo\.od"; & "$env:LOCALAPPDATA\Programs\Open Design\Open Design.exe"
+```
+
+The daemon log records `[od-migrate] migration complete: copied N entries (...)`. After the first launch you can clear the env variable; the marker prevents re-migration even on subsequent runs.
+
+##### Option B: manual copy
```
lefarcen, line 422: **P2 — Option B missing Windows alternative**

Manual-copy path is bash/macOS-oriented. Section lists Windows data paths earlier, but Option B provides no PowerShell equivalent for staging, backup, copy, restore.

**Fix:** Add Windows PowerShell manual migration block, or explicitly say Option B currently requires macOS/Linux shell commands.

lefarcen (changes requested):
Hey @Nagendhra-web — the migration approach is solid for repo-to-packaged recovery, but there are several critical safety gaps that need addressing before this can ship.

## Summary

The core logic (sync one-shot copy, idempotent marker, fresh-dataDir guard) is the right design. The blocking issues are: (1) packaged env propagation won't reach the daemon, (2) partial-copy failure modes leave inconsistent state, (3) large-copy timeout with packaged sidecar, (4) merge-mode rather than replace-mode for fresh targets. Findings below are prioritized P1 → P3.

Thanks for tackling #710 with a tested implementation — this is a valuable recovery path once the safety concerns are addressed.

apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,311 @@
...
+  fs.mkdirSync(path.dirname(dataDir), { recursive: true });
+  const stagingDir = path.join(
+    path.dirname(dataDir),
+    `${path.basename(dataDir)}.od-migrate-${process.pid}-${Date.now()}`,
+  );
+
+  let copied: string[];
+  try {
+    copied = stagePayload(legacyDir, stagingDir);
+    promoteStaged(stagingDir, dataDir, copied);
```
mrcfps: This still leaves a partial final `dataDir` possible during the promote phase. The payload is staged first, but `promoteStaged()` moves entries into the real data root one at a time, and the `catch` below only removes `stagingDir`. If `app.sqlite` is renamed successfully and a later `projects`/`artifacts` rename or fallback copy fails (permissions, ENOSPC, or another filesystem error), the real `dataDir` is left with `app.sqlite` but no marker. On the next boot, `dataDirHasExistingPayload()` will throw `data_dir_not_empty`, so Option A can no longer retry and the user is stranded with an incomplete migration.

Please make the promotion rollback-safe as a unit: either promote the whole staged data root with one directory-level swap, or track every payload entry successfully promoted and remove those known entries if promotion or marker writing fails before rethrowing. A regression that injects a failure after `app.sqlite` has been promoted should assert that the final `dataDir` has no payload entries and that a second migration attempt can succeed.

README.md
```diff
@@ -374,26 +374,51 @@ If you ran the repo first and only later installed the packaged Desktop app, the
...
+
+- the path in `OD_LEGACY_DATA_DIR` does not contain `app.sqlite` (typo, deleted source, wrong path), or
+- the Desktop's `data/` already contains any of `app.sqlite`, `projects/`, `artifacts/`, `media-config.json`, etc. SQLite/WAL pairs and project trees cannot be safely interleaved, so the daemon refuses to merge instead of silently corrupting either side. If the Desktop has already booted and seeded its own `data/`, use Option B and decide explicitly which side wins.
+
+A `.migrated-from` marker is written on success so subsequent boots no-op.
+
+```bash
+# Quit the Desktop app first, then re-launch with this env set.
+# macOS/Linux:
+OD_LEGACY_DATA_DIR="/path/to/old/repo/.od" open "/Applications/Open Design.app"
```
mrcfps: This macOS command does not actually put `OD_LEGACY_DATA_DIR` into the packaged app's process environment. `VAR=value open "/Applications/Open Design.app"` sets the variable only for the short-lived `open` process; the GUI app is launched by LaunchServices/launchd and normally will not inherit that shell environment. The new sidecar forwarding in `apps/packaged/src/sidecars.ts` only helps if the packaged process already has the variable, so users following this documented Option A path on macOS will still get `process.env.OD_LEGACY_DATA_DIR === undefined` and the migrator will no-op.

Please replace the macOS example with a launch method that really sets the app process env, such as `launchctl setenv OD_LEGACY_DATA_DIR /path/to/old/repo/.od` before `open` (and an `unsetenv` cleanup step), or launching the actual executable directly: `OD_LEGACY_DATA_DIR=... "/Applications/Open Design.app/Contents/MacOS/Open Design"`. Also split Linux from the macOS `open` example so each platform has a command that works for that launcher.

mrcfps (changes requested):
@Nagendhra-web Thanks for the follow-up hardening here — the staging approach, symlink checks, and packaged env forwarding move the migration in a much safer direction. I found two remaining blocking issues in the current head that can still prevent the main recovery flow from working reliably, one in promotion failure handling and one in the macOS launch instructions. 🙏


lefarcen (approved):
Hey @Nagendhra-web! 👋

Great progress on this iteration:

**✅ P1 resolved:** The packaged env propagation issue is fixed — `apps/packaged/src/sidecars.ts:247-255` now explicitly forwards `OD_LEGACY_DATA_DIR` to the daemon sidecar, so the primary 0.3.x → 0.4.x upgrade recovery flow will work as documented.

**📋 Two P2 recommendations (non-blocking):**

1. **Idempotency order** (`apps/daemon/src/legacy-data-migrator.ts:255-268`): Currently validates `OD_LEGACY_DATA_DIR` before checking `.migrated-from`. If a user leaves the env var set after successful migration and later moves/deletes the old repo `.od/`, future daemon boots throw `legacy_dir_invalid` instead of no-oping via the marker. Suggest checking the marker first.

2. **macOS Option A command** (`README.md:388-393`): `open` likely doesn't pass env vars into the launched app. Users may need `open --env OD_LEGACY_DATA_DIR=...` or a direct executable launch. Worth verifying the macOS primary path actually works.

Looks good to me overall; deferring final approval to a maintainer.

apps/daemon/src/server.ts
mrcfps, line 695: This startup call runs the full synchronous copy before the daemon has reported an IPC status. In the packaged path, `startPackagedSidecars()` waits only 35 seconds for daemon status and then tears down the child; a user with a large legacy `.od/projects` or `.od/artifacts` tree can hit that timeout while `stagePayload()` is still copying. Because the parent then kills the daemon mid-migration, the migration can never complete for those users, and if the kill happens during promotion it can still leave partial payload entries in `dataDir` without executing the rollback code.

Please make packaged startup explicitly migration-aware before wiring this into daemon boot: for example, extend/suspend the daemon status timeout while `OD_LEGACY_DATA_DIR` is set, emit an early status/progress signal before the copy begins, or run the copy under a parent-managed path that will not kill the child mid-copy. Add a regression that simulates a migration exceeding the packaged status timeout so the recovery flow cannot be killed before the daemon becomes ready.

apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,331 @@
...
+    writeMarker(dataDir, legacyDir);
```
mrcfps: The rollback now covers failures inside `promoteStaged()`, but `writeMarker()` is still inside the same `try` after all payload entries have been promoted. If marker creation fails (for example ENOSPC, permissions, or a read-only target after the data files were copied), this catch only removes `stagingDir`; the real `dataDir` keeps the promoted `app.sqlite`/projects but has no `.migrated-from` marker. The next boot with `OD_LEGACY_DATA_DIR` set will hit `dataDirHasExistingPayload()` and throw `data_dir_not_empty`, so the one-shot migration cannot retry cleanly.

Please make marker failure part of the same unit rollback as promotion: either have `promoteStaged()` return the promoted entries and remove them if `writeMarker()` throws, or stage the marker and payload together and only make the final data root visible after every required file is ready. Add a regression that forces `writeMarker()` to fail after promotion and asserts the final `dataDir` has no payload entries and a second migration attempt can succeed.

mrcfps (changes requested):
@Nagendhra-web Thanks for the continued hardening here — the env forwarding, safer launch docs, symlink checks, and promotion rollback all move this recovery path in the right direction. I found two remaining failure modes that can still strand the primary Desktop migration flow, so I’m requesting changes to make the migration retry/rollback behavior fully safe. 🙏


apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,331 @@
...
+  const promoted: string[] = [];
+  try {
+    for (const entry of entries) {
+      const src = path.join(stagingDir, entry);
+      const dst = path.join(dataDir, entry);
+      try {
+        fs.renameSync(src, dst);
+      } catch {
+        // EXDEV (cross-device) or similar: fall back to copy + remove.
+        fs.cpSync(src, dst, {
```
lefarcen: **P1 — Incomplete rollback in copy fallback path.**

If `renameSync` throws EXDEV and the fallback `fs.cpSync(src, dst, {recursive:true})` partially writes `dst` then fails (permissions / ENOSPC / interrupted), the partial `dst` is left in the real `dataDir` but hasn't been pushed to `promoted` yet, so the outer catch only removes entries in `promoted` and doesn't clean up this partial write.

Next boot hits `dataDirHasExistingPayload()` → `data_dir_not_empty` and can't retry.

**Fix:** Wrap the `fs.cpSync` fallback in its own try-catch, and on failure remove `dst` before rethrowing:

```ts
try {
  renameSync(src, dst);
} catch (err) {
  if (err.code === 'EXDEV') {
    try {
      fs.cpSync(src, dst, { recursive: true, errorOnExist: true });
    } catch (copyErr) {
      // rollback partial write
      if (fs.existsSync(dst)) {
        fs.rmSync(dst, { recursive: true, force: true });
      }
      throw copyErr;
    }
    fs.rmSync(src, { recursive: true, force: true });
  } else {
    throw err;
  }
}
```

This ensures the `dst` is clean on any failure before rethrowing to the outer catch, which then rolls back `promoted` entries as before.

lefarcen (changes requested):
Hey @Nagendhra-web! 👋

Great iteration — you addressed the macOS Option A command issue (now correctly uses direct executable launch), and the overall promotion approach with entry-by-entry rollback tracking is much safer.

**✅ Fixed from mrcfps's review:**
- macOS Option A now uses direct executable launch instead of the broken `open` pattern

**❌ One remaining P1 blocker:**

The promotion rollback logic (`apps/daemon/src/legacy-data-migrator.ts`) still has a gap: if the `renameSync` fallback throws mid-write (e.g., `fs.cpSync` partially writes `dst` then fails), that partial entry is left in `dataDir` but isn't in the `promoted` array yet, so the outer catch won't remove it. Next boot hits `data_dir_not_empty` and can't retry cleanly.

See inline comment for the specific fix needed.

Once that's addressed, this PR will be rock-solid! 🎉

apps/daemon/src/legacy-data-migrator.ts
lefarcen: **P1** The `renameSync` fallback to `fs.cpSync` can still leave a partial `dst` entry if the copy partially writes and then throws (ENOSPC mid-write, permissions failure after creating the directory, etc.). Because the current entry is only pushed to `promoted` AFTER the fallback completes, the outer `catch` rollback won't remove this partial `dst` — the next boot hits `data_dir_not_empty` and cannot retry.

Wrap the `fs.cpSync` fallback in its own try-catch and remove `dst` on failure before rethrowing:

```ts
try {
  fs.cpSync(src, dst, { recursive: true, preserveTimestamps: true });
} catch (copyErr) {
  fs.rmSync(dst, { recursive: true, force: true });
  throw copyErr;
}
```

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for the continued hardening! 👋

✅ **macOS Option A is now FIXED** — the README correctly uses direct binary launch or `launchctl setenv`, so the GUI process will actually receive `OD_LEGACY_DATA_DIR`.

❌ **1 remaining P1 blocker** — promotion rollback still doesn't cover the full migrate-or-nothing transaction. See inline comment below.

apps/daemon/src/legacy-data-migrator.ts
lefarcen: P1: The `renameSync` fallback can still leave the current entry partially written if `fs.cpSync(src, dst, ...)` partially creates `dst` and then throws. Because the current entry is only pushed to `promoted` after the fallback completes, the outer rollback removes prior entries but not the partial current `dst`. Next boot again hits `data_dir_not_empty`. Wrap the fallback `fs.cpSync` in try-catch and remove `dst` on failure before rethrowing:

```typescript
try {
  fs.cpSync(src, dst, { recursive: true, preserveTimestamps: true });
  promoted.push(entry);
} catch (copyErr) {
  fs.rmSync(dst, { recursive: true, force: true });
  throw copyErr;
}
```

lefarcen (changes requested):
Hey @Nagendhra-web! 👋 Thanks for the latest round of fixes.

The macOS launch instructions are now correct — direct executable invocation and `launchctl setenv` both work reliably. ✅

However, I found **1 remaining P1 blocker** in the promotion rollback logic that can still strand users with a partial migration:

If `promoteStaged()` completes successfully but `writeMarker()` then throws (ENOSPC, permissions, read-only filesystem), the outer catch only removes the staging directory. The promoted `app.sqlite` / `projects/` / `artifacts/` remain in `dataDir` without a `.migrated-from` marker, so the next boot with `OD_LEGACY_DATA_DIR` set hits `dataDirHasExistingPayload()` → throws `data_dir_not_empty` → Option A cannot retry.

The fix: make marker failure part of the same rollback unit as promotion. Either:
1. Have `promoteStaged()` return the list of promoted entries, then remove them if `writeMarker()` fails, OR
2. Stage the marker along with the payload and only make the final data root visible after everything is ready

A regression should inject a `writeMarker()` failure after promotion and assert the final `dataDir` is clean so retry succeeds.

mrcfps (approved):
@Nagendhra-web I reviewed the current migration implementation, packaged sidecar env/timeout wiring, launch documentation, and the new daemon/packaged regression coverage. The previous rollback, marker-failure, and packaged launch blockers are addressed in this head; thanks for continuing to harden this recovery path and for keeping the fixes well covered. 🙏


apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,370 @@
...
+    // dataDir, permissions). If it does, the payload entries are
+    // already in dataDir but no marker was written; the next boot
+    // would hit dataDirHasExistingPayload() and throw data_dir_not_empty,
+    // making the migration un-retryable. Rolling back the promoted
+    // entries here keeps the failure as a unit with promotion.
+    (options.writeMarker ?? writeMarker)(dataDir, legacyDir);
+  } catch (err) {
+    rollbackPromoted(dataDir, promoted);
+    fs.rmSync(stagingDir, { recursive: true, force: true });
+    throw err;
```
lefarcen, line 387: **P1:** If `writeMarker()` creates/truncates `.migrated-from` and then throws (ENOSPC/EIO), the catch rolls back promoted payload entries but does NOT remove the partial marker. Next boot will find the marker at line 318-320 and skip migration, leaving the user stranded.

Fix by writing marker atomically via temp file + rename:
```typescript
const tempPath = markerPath + ".tmp";
fs.writeFileSync(tempPath, "...", "utf8");
fs.renameSync(tempPath, markerPath);
```
Or remove marker in catch:
```typescript
if (fs.existsSync(markerPath)) fs.unlinkSync(markerPath);
```

lefarcen (changes requested):
Thanks for the continued iteration here @Nagendhra-web. The promotion rollback tracking and cpSync fallback cleanup from commit `f742be0` fix the two P1s from my last review. 

One remaining P1 blocker: if `writeMarker()` creates the `.migrated-from` file and then throws (ENOSPC/EIO/permissions mid-write), the catch rolls back promoted payload but does NOT remove the partial marker. Next boot will find the marker, skip migration at line 318-320, and leave the user stranded with no data.

Fix by either:
1. Writing marker atomically via temp file + rename: `fs.writeFileSync(markerPath + ".tmp", ..., "utf8"); fs.renameSync(markerPath + ".tmp", markerPath);`
2. Or: in the catch, remove `markerPath` before rethrowing: `fs.existsSync(markerPath) && fs.unlinkSync(markerPath);`

Once this is addressed, the migration retry contract will be fully safe.

apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,370 @@
...
+    // The user typo'd, deleted the source, or pointed at the wrong
+    // directory; silently launching empty trained users to think the
+    // migration ran when it hadn't (this is the original #710 footgun).
+    throw new LegacyMigrationError(
+      'legacy_dir_invalid',
+      `OD_LEGACY_DATA_DIR="${legacyDir}" is not a usable legacy data dir (expected app.sqlite directly inside it). Quit Open Design, fix the path, and relaunch.`,
+    );
+  }
+
+  const markerPath = path.join(dataDir, MARKER_FILE);
```
lefarcen: **P1** When `writeMarker()` throws (ENOSPC, permissions, or partial write), the catch below rolls back promoted entries but does NOT remove the marker file itself. If `fs.writeFileSync(markerPath, ...)` partially wrote the file before throwing, `markerPath` exists on retry → `fs.existsSync(markerPath)` returns true → migration is skipped → user stranded with empty data dir.

Please add marker rollback to the catch path: `fs.rmSync(markerPath, { force: true })` after promoted-entry cleanup and before rethrowing. Also add a test that injects marker write failure (e.g. by mocking `fs.writeFileSync` to throw after writing a partial file) and asserts both promoted entries AND the marker are absent after the exception.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for the continued hardening — the cp fallback cleanup and promoted-entry rollback tracking move this in the right direction. I found one remaining P1 blocker in the marker failure path that can still strand the migration.

**Remaining P1 blocker:**

**Marker file rollback gap** (see inline below) — when `writeMarker()` fails after promoted entries are already moved, the catch rolls back the promoted entries but does NOT roll back a partially-written marker file. On retry, `fs.existsSync(markerPath)` returns true → migration is skipped → user left with empty data dir.

**Previously-reported P1s — status:**

✅ **cpSync fallback partial write** — FIXED (apps/daemon/src/legacy-data-migrator.ts:246-255 wraps cp fallback and removes dst before rethrowing)

✅ **Promoted-entry rollback** — FIXED (promoted entries are now tracked and removed on marker failure)

Once the marker rollback gap is closed, this recovery path should be fully retry-safe. 🙏

mrcfps (commented):
@Nagendhra-web Thanks for the continued hardening here — the rollback and marker failure paths are much safer in this head. I found one merge-safe follow-up around the idempotency ordering so the success marker keeps protecting later launches even if the legacy source goes away. 🙏


apps/daemon/src/legacy-data-migrator.ts
```diff
@@ -0,0 +1,390 @@
...
+    return { status: 'noop', reason: 'OD_LEGACY_DATA_DIR not set' };
+  }
+  const legacyDir = path.resolve(raw);
+  const dataDir = path.resolve(options.dataDir);
+
+  if (legacyDir === dataDir) {
+    return { status: 'noop', reason: 'OD_LEGACY_DATA_DIR equals OD_DATA_DIR' };
+  }
+
+  if (!legacyDirHasPayload(legacyDir)) {
```
mrcfps, line 333: The `.migrated-from` idempotency check currently runs after `legacyDirHasPayload(legacyDir)`, so a successful migration can still turn into a startup error later if the user leaves `OD_LEGACY_DATA_DIR` set and then moves or deletes the old repo `.od/`. That is plausible with the documented `launchctl setenv` path, and it weakens the marker contract described in the PR/README: after success, the marker should make later boots no-op rather than re-validating a source that is no longer needed. Consider moving the `markerPath` / `fs.existsSync(markerPath)` check to immediately after `legacyDir` and `dataDir` are resolved (and after the self-dir guard), before checking `legacyDirHasPayload`, and add a regression where a first migration succeeds, the legacy dir is removed, and the second call returns `skipped` via the marker.

lefarcen (approved):
Excellent work @Nagendhra-web! 🎉 All P1 blockers from my previous reviews are now resolved in this head (`89037e2d`):

✅ **Marker file atomicity** (P1 from `f742be0e` review) — `writeMarker` now uses temp file + atomic rename, so a partial marker can never reach `markerPath`

✅ **Defensive marker cleanup** — the outer catch defensively removes `markerPath` before rethrowing, so even a hypothetical non-atomic writeMarker variant won't strand users

✅ **Idempotency order** (mrcfps P2) — marker check now runs BEFORE `legacyDirHasPayload`, so post-success boots no-op even when the legacy source is deleted

The migration retry contract is now fully safe. CI is green, test coverage is solid (27/27 daemon + 4/4 packaged). This recovery path should ship reliably.

mrcfps APPROVED on `f742be0e` → the two additional commits (`5ba031b` + `89037e2d`) are hardening improvements with no new risk.

Approved! 🚀

apps/packaged/src/sidecars.ts
```diff
@@ -74,10 +74,32 @@ async function openLog(path: string): Promise<FileHandle> {
...
+ * not tear the daemon down before the migration can complete.
+ *
+ * @see apps/daemon/src/legacy-data-migrator.ts
+ * @see https://github.com/nexu-io/open-design/issues/710
+ */
+export function resolveDaemonStatusTimeoutMs(
+  env: NodeJS.ProcessEnv = process.env,
+): number {
+  const raw = env.OD_LEGACY_DATA_DIR;
+  if (raw != null && raw.length > 0) return DAEMON_MIGRATION_STATUS_TIMEOUT_MS;
```
mrcfps, line 95: The migration-aware timeout is needed for large successful copies, but applying the full 30-minute wait whenever `OD_LEGACY_DATA_DIR` is set also covers daemon failures before status is available. In this PR the daemon intentionally throws during startup for invalid legacy paths, existing target payload, symlinks, marker write failures, and other operator-actionable migration errors; because `waitForStatus` only retries IPC until `timeoutMs`, a daemon that exits immediately on any of those errors now leaves the packaged app waiting the full 30 minutes before the user sees a failure. That undermines the new "visible startup error" contract and makes common typo/retry flows look like a hung app. Please make the extended wait fail fast when the daemon child exits, for example by teaching the daemon status wait to race the IPC polling against the spawned process exit (or by passing the child into `waitForStatus` and throwing with the daemon log path once it exits), while preserving the long timeout only for a still-running migration. A packaged regression should set `OD_LEGACY_DATA_DIR`, simulate the daemon child exiting before status, and assert `startPackagedSidecars` rejects promptly rather than waiting for `DAEMON_MIGRATION_STATUS_TIMEOUT_MS`.

mrcfps (changes requested):
@Nagendhra-web Thanks for the continued hardening on the migrator and for closing the previous rollback/idempotency gaps. I found one remaining packaged-launcher issue: failure cases in the new migration path can now leave users waiting for the full extended status timeout instead of surfacing the daemon error promptly. 🙏


mrcfps (approved):
@Nagendhra-web I reviewed the current migrator, daemon boot wiring, packaged sidecar env/timeout handling, launch documentation, and the new daemon/packaged regression coverage. The earlier rollback, marker, idempotency, and packaged-launch failure paths are addressed in this head; thanks for the careful follow-up work and for keeping the recovery flow well covered. 🙏

PR #713, opened 2026-05-06
fix(daemon): handle 0-byte zip entries in claude-design import

### LINKED ISSUE #590: Import Claude Design ZIP silently fails: RangeError on 0-byte entries (Node 24) and MAX_FILES=500 too low

Abhishek764:
Fixes #590

## Root cause

`readEntryBody` in `apps/daemon/src/claude-design-import.ts` calls `inflateRawSync` when a zip entry uses compression method 8 (deflate). When the entry has `uncompressedSize === 0` (a legitimately empty file), two things go wrong:

1. `inflateRawSync` receives an empty buffer — an invalid deflate stream — and throws.
2. The `maxOutputLength: 0` option passed to `inflateRawSync` is itself invalid and throws a `RangeError`.

ZIP files produced by macOS `zip`, Python's `zipfile`, and other tools include 0-byte entries with `method=8` for empty files. The import silently failed with an unhandled `RangeError` for any such archive.

## Fix

One line added to `readEntryBody`: early-return `Buffer.alloc(0)` when `entry.uncompressedSize === 0`, before touching the local header or calling inflate. The downstream size-mismatch check (`body.length !== entry.uncompressedSize` → `0 !== 0`) remains correct.

```diff
 function readEntryBody(zip, entry) {
+  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
   const offset = entry.localOffset;
```

## Tests

Added `apps/daemon/tests/claude-design-import.test.ts` with a self-contained ZIP builder (no test dependencies added):

- **Regression case**: `method=8`, `uncompressedSize=0` — previously threw `RangeError`, now passes.
- **Stored zero-byte entry**: `method=0`, `uncompressedSize=0` — confirms stored path is unaffected.
- **Entry file selection**: `index.html` preferred over other HTML files.
- **No HTML file rejection**: throws the correct error message.

All 4 tests pass. `pnpm guard` passes. Pre-existing typecheck failures in `src/critique/` are unrelated to this change.

### DIFF
--- a/apps/daemon/src/claude-design-import.ts
+++ b/apps/daemon/src/claude-design-import.ts
@@ -102,7 +102,6 @@ function findEndOfCentralDirectory(zip) {
 }
 
 function readEntryBody(zip, entry) {
-  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
   const offset = entry.localOffset;
   if (zip.readUInt32LE(offset) !== LOCAL_SIG) {
     throw new Error(`invalid zip local header: ${entry.name}`);
@@ -112,6 +111,10 @@ function readEntryBody(zip, entry) {
   const bodyStart = offset + 30 + nameLen + extraLen;
   const bodyEnd = bodyStart + entry.compressedSize;
   if (bodyEnd > zip.length) throw new Error(`zip entry exceeds archive: ${entry.name}`);
+  // inflateRawSync throws RangeError when uncompressedSize is 0 (maxOutputLength:0
+  // is invalid) and on an empty deflate stream. Return early after header and bounds
+  // checks so the local-header signature and offset are still validated.
+  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
   const compressed = zip.slice(bodyStart, bodyEnd);
   if (entry.method === 0) return Buffer.from(compressed);
   return inflateRawSync(compressed, { maxOutputLength: entry.uncompressedSize });
--- a/apps/daemon/src/claude-design-import.ts
+++ b/apps/daemon/src/claude-design-import.ts
@@ -102,6 +102,7 @@ function findEndOfCentralDirectory(zip) {
 }
 
 function readEntryBody(zip, entry) {
+  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
   const offset = entry.localOffset;
   if (zip.readUInt32LE(offset) !== LOCAL_SIG) {
     throw new Error(`invalid zip local header: ${entry.name}`);
--- a/apps/daemon/tests/claude-design-import.test.ts
+++ b/apps/daemon/tests/claude-design-import.test.ts
@@ -154,4 +154,31 @@ describe('importClaudeDesignZip', () => {
       );
     });
   });
+
+  it('rejects a 0-byte entry whose local header has an invalid signature', async () => {
+    // Regression: the 0-byte early-return must NOT run before LOCAL_SIG validation.
+    // A malformed archive with a bad localOffset for a 0-byte entry must still be rejected.
+    await withTmpDir(async (dir) => {
+      const valid = buildZip([
+        { name: 'index.html', method: 0, data: Buffer.from('<html></html>', 'utf8') },
+        { name: 'empty.css', method: 8, data: Buffer.alloc(0) },
+      ]);
+      // Corrupt the local-file-header signature of the second entry (empty.css).
+      // The central directory says localOffset points to that header; overwrite
+      // the 4-byte signature with 0x00000000 so LOCAL_SIG check fails.
+      const corrupt = Buffer.from(valid);
+      // Find the second local header: first entry is at offset 0, its local header
+      // is 30 + len("index.html")=10 bytes + data bytes. Signature starts at that offset.
+      const firstLocalSize = 30 + 'index.html'.length + '<html></html>'.length;
+      corrupt.writeUInt32LE(0x00000000, firstLocalSize); // overwrite LOCAL_SIG of entry 2
+
+      const zipPath = path.join(dir, 'corrupt.zip');
+      const outDir = path.join(dir, 'out');
+      await writeFile(zipPath, corrupt);
+
+      await expect(importClaudeDesignZip(zipPath, outDir)).rejects.toThrow(
+        'invalid zip local header',
+      );
+    });
+  });
 });
--- a/apps/daemon/tests/claude-design-import.test.ts
+++ b/apps/daemon/tests/claude-design-import.test.ts
@@ -1,0 +1,157 @@
+import { mkdtemp, rm, writeFile } from 'node:fs/promises';
+import os from 'node:os';
+import path from 'node:path';
+import { describe, expect, it } from 'vitest';
+
+import { importClaudeDesignZip } from '../src/claude-design-import.js';
+
+// Minimal CRC-32 matching the zip spec.
+function crc32(buf: Buffer): number {
+  const t: number[] = [];
+  for (let n = 0; n < 256; n++) {
+    let c = n;
+    for (let k = 0; k < 8; k++) c = c & 1 ? 0xedb88320 ^ (c >>> 1) : c >>> 1;
+    t[n] = c >>> 0;
+  }
+  let c = 0xffffffff;
+  for (let i = 0; i < buf.length; i++) c = t[(c ^ buf[i]!) & 0xff]! ^ (c >>> 8);
+  return (c ^ 0xffffffff) >>> 0;
+}
+
+function buildZip(entries: Array<{ name: string; method: 0 | 8; data: Buffer }>): Buffer {
+  const localParts: Buffer[] = [];
+  const centralParts: Buffer[] = [];
+  let localOffset = 0;
+
+  for (const entry of entries) {
+    const name = Buffer.from(entry.name, 'utf8');
+    const { data } = entry;
+    const checksum = crc32(data);
+
+    const local = Buffer.alloc(30 + name.length);
+    local.writeUInt32LE(0x04034b50, 0);
+    local.writeUInt16LE(20, 4);
+    local.writeUInt16LE(0, 6);
+    local.writeUInt16LE(entry.method, 8);
+    local.writeUInt16LE(0, 10);
+    local.writeUInt16LE(0, 12);
+    local.writeUInt32LE(checksum, 14);
+    local.writeUInt32LE(data.length, 18);
+    local.writeUInt32LE(data.length, 22);
+    local.writeUInt16LE(name.length, 26);
+    local.writeUInt16LE(0, 28);
+    name.copy(local, 30);
+    localParts.push(local, data);
+
+    const central = Buffer.alloc(46 + name.length);
+    central.writeUInt32LE(0x02014b50, 0);
+    central.writeUInt16LE(20, 4);
+    central.writeUInt16LE(20, 6);
+    central.writeUInt16LE(0, 8);
+    central.writeUInt16LE(entry.method, 10);
+    central.writeUInt16LE(0, 12);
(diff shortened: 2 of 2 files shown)


apps/daemon/src/claude-design-import.ts
```diff
@@ -102,6 +102,7 @@ function findEndOfCentralDirectory(zip) {
 }
 
 function readEntryBody(zip, entry) {
+  if (entry.uncompressedSize === 0) return Buffer.alloc(0);
```
mrcfps: This early return fixes the `inflateRawSync(..., { maxOutputLength: 0 })` failure, but because it runs before the local header is read it also bypasses the existing `LOCAL_SIG` check and the `bodyEnd > zip.length` bounds check below. A malformed archive can now put a central-directory record for a zero-byte `index.html` with an invalid or out-of-range `localOffset`/`compressedSize`, and the importer will accept and write that file without proving that a valid local entry exists; before this change that same archive was rejected while reading the local header. Please keep the zero-output special case after the local-header parsing and compressed-body bounds validation (for example, compute `bodyStart`/`bodyEnd`, verify the signature and bounds, then return `Buffer.alloc(0)` before inflating when `entry.uncompressedSize === 0`) and add a regression test for a zero-byte entry with an invalid local header or out-of-bounds body.

mrcfps (changes requested):
@Abhishek764 thanks for the focused fix and for adding regression coverage around zero-byte ZIP entries. I found one correctness issue in the new early-return path: it now skips the local-header and bounds validation that the importer still needs before accepting an entry, so I'm requesting a small adjustment before merge. 🙏


apps/daemon/src/claude-design-import.ts
lefarcen: **P3 — Early return bypasses validation**

The new early return skips:
- Local header validation (`zip.readUInt32LE(offset) !== LOCAL_SIG`)
- Compressed size + body bounds checks

This means a malformed stored entry with `uncompressedSize === 0` but nonzero `compressedSize` (invalid per ZIP spec) is now accepted instead of failing the existing `body.length !== entry.uncompressedSize` check downstream.

**Concrete fix:** Move the local header parse + bounds check *before* the zero-byte shortcut:

```ts
function readEntryBody(zip, entry) {
  const offset = entry.localOffset;
  if (zip.readUInt32LE(offset) !== LOCAL_SIG) {
    throw new Error(`invalid zip local header: ${entry.name}`);
  }
  // ... parse local header fields (nameLen, extraLen) ...
  
  if (entry.uncompressedSize === 0) return Buffer.alloc(0);  // ← move AFTER validation
  
  // ... rest of body extraction ...
}
```

Alternatively, only shortcut the deflate path and let stored entries continue through `Buffer.from(compressed)` + the existing size check.

apps/daemon/tests/claude-design-import.test.ts
```diff
@@ -0,0 +1,157 @@
...
+    let c = n;
+    for (let k = 0; k < 8; k++) c = c & 1 ? 0xedb88320 ^ (c >>> 1) : c >>> 1;
+    t[n] = c >>> 0;
+  }
+  let c = 0xffffffff;
+  for (let i = 0; i < buf.length; i++) c = t[(c ^ buf[i]!) & 0xff]! ^ (c >>> 8);
+  return (c ^ 0xffffffff) >>> 0;
+}
+
+function buildZip(entries: Array<{ name: string; method: 0 | 8; data: Buffer }>): Buffer {
```
lefarcen, line 21: **P3 — Test ZIP builder doesn't match real-world deflate behavior**

The `buildZip` helper treats `data` as both uncompressed content and the compressed payload even when `method === 8`. Your zero-byte test still reproduces the Node 24 `maxOutputLength: 0` crash, but it doesn't match what Python `zipfile` or macOS `zip` actually produce for empty deflate entries (a 2-byte raw-deflate payload: `03 00`).

This can mislead future tests or mask alternative fixes.

**Concrete fix:** Either:
- Make `buildZip` actually call `deflateRawSync(data)` for `method=8` entries, or
- Make the API explicit: `buildZip(entries: Array<{ name, method, uncompressedSize, compressedData }>)` so the caller chooses payloads.

Not a blocker for this PR, but worth noting for future test cases.

lefarcen (commented):
Hey @Abhishek764! 👋 Thanks for tackling this tricky Node 24 regression.

The one-line fix is elegant, and the test coverage is solid. I've got a few suggestions to make this production-ready:

**Scope clarification (non-blocking but important):**

The PR body says `Fixes #590`, but #590 reports *two* distinct failure modes:
1. Zero-byte deflate entries crash with `RangeError` (Node 24) — ✅ your fix addresses this.
2. `MAX_FILES = 500` blocks real-world exports (tested with 561 files / 4.4 MB) — ❌ still present in the current code.

After this PR, a 561-file Claude Design export will still fail with `zip contains too many files`. Either:
- Scope the PR to just the Node 24 crash and change to `Closes #590 (part 1: zero-byte entries)`, or
- Include a `MAX_FILES` bump + regression test above 500 files.

**Code review (inline below) + minor test gap:**

See inline comments.

lefarcen (commented):
Nice work addressing the validation bypass, @Abhishek764! 🎉

The early return now correctly validates the local header and bounds before returning the empty buffer — that fixes the blocking correctness issue mrcfps flagged.

I've got one remaining edge-case nit and the same scope clarification from before:

**Remaining P3 (non-blocking):**

The current fix handles deflate zero-byte entries correctly, but stored entries (`method=0`) with mismatched sizes (e.g. `uncompressedSize === 0` but nonzero `compressedSize`) are now accepted as empty instead of failing the size-mismatch check downstream.

**Concrete fix:** Only early-return for deflate (`method === 8`):

```ts
if (bodyEnd > zip.length) throw new Error(`zip entry exceeds archive: ${entry.name}`);
const compressed = zip.slice(bodyStart, bodyEnd);
if (entry.method === 0) return Buffer.from(compressed);
// Node 24 inflateRawSync throws RangeError when maxOutputLength: 0.
// Return early after header/bounds checks for zero-byte deflate entries.
if (entry.uncompressedSize === 0) return Buffer.alloc(0);
return inflateRawSync(compressed, { maxOutputLength: entry.uncompressedSize });
```

This preserves the existing size check for stored entries while fixing the Node 24 deflate issue.

**Scope clarification (same as before):**

PR body says `Fixes #590`, but #590 has two failure modes:
1. Zero-byte deflate crash (Node 24) — ✅ fixed
2. `MAX_FILES = 500` blocks 561-file exports — ❌ unchanged

Consider scoping to just the Node 24 fix or adding a `MAX_FILES` bump.

---

Overall: the blocking issue is resolved, tests are solid, and the PR is close to ready. Nice iteration!

mrcfps (approved):
@Abhishek764 I reviewed the updated ZIP import path and the new daemon regression tests for zero-byte stored/deflated entries plus the malformed local-header case. The current change keeps the header and bounds checks before the empty-entry shortcut and covers the main regression cleanly; thanks for the focused fix and quick iteration 🙌


lefarcen:
Hi @Abhishek764,

While your PR was under review, PR #591 by @gabrielvaz was merged (12 hours ago) and already fixed issue #590 with the same approach — handling zero-byte ZIP entries to avoid the Node 24 `inflateRawSync` crash.

Your PR now conflicts with `main` because both PRs modify the same code paths. Since #591 is already merged and shipped, this PR is superseded.

**Thank you** for the focused fix, the clear explanation, and the excellent test coverage — your contribution helped validate the approach! 🙏

I'd suggest closing this PR. If you'd like to contribute further, feel free to pick up another issue from the repo or check the CONTRIBUTING guide.

Thanks again for making open-design better!
— open-design team

lefarcen:
Thanks @Abhishek764 for this PR. The change here was already landed via #591 (merged 12h ago), which fixed the same issue #590. Closing this one as superseded. If you think the merge missed something, please reopen and call it out.

PR #719, opened 2026-05-07
Fix/raise max tokens to 65535

saikastxy:
Let the max tokens to 65535 so that the chat will not stop too early

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -2,7 +2,7 @@ import { useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { createHtmlArtifactManifest, inferLegacyManifest } from '../artifacts/manifest';
 import { createArtifactParser } from '../artifacts/parser';
 import { useT } from '../i18n';
-import { streamMessage } from '../providers/anthropic';
+import { streamMessage, streamMessageWithAgentLoop } from '../providers/anthropic';
 import {
   fetchChatRunStatus,
   listActiveChatRuns,
@@ -741,7 +741,7 @@ export function ProjectView({
               unregisterTextBuffer();
               updateMessageById(
                 message.id,
-                (prev) => ({ ...prev, runStatus: 'succeeded', endedAt: prev.endedAt ?? Date.now() }),
+                (prev) => ({ ...prev, endedAt: prev.endedAt ?? Date.now() }),
                 true,
               );
               completedReattachRunsRef.current.add(runId);
@@ -1045,7 +1045,7 @@ export function ProjectView({
           updateAssistant((prev) => ({
             ...prev,
             endedAt: Date.now(),
-            runStatus: config.mode === 'api' || prev.runId ? 'succeeded' : prev.runStatus,
+            runStatus: config.mode === 'api' ? 'succeeded' : prev.runStatus,
           }));
           if (commentAttachments.length > 0) {
             void patchAttachedStatuses(commentAttachments, 'needs_review');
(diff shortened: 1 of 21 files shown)


mrcfps:
@saikastxy I'm holding off on generating review comments for nexu-io/open-design#719 because this pull request has merge conflicts right now.

apps/web/src/components/ProjectView.tsx
```diff
@@ -522,10 +531,31 @@ export function ProjectView({
       metadata: project.metadata,
       template,
     });
+
+    // API-mode system prompt extras: file listing, CWD hint, and tool
+    // instructions the model needs to locate and read uploaded files.
+    const cwdHint = `\n\nWorking directory: .od/projects/${project.id}/\nAll project files are flat under this directory. Use exact basenames (e.g. "paper.pdf", not a full path) with the Read tool.`;
+    const fileListHint = projectFiles.length
+      ? `\n\nFiles already in project (use the Read tool with the exact name from this list):\n${projectFiles
+          .map((f) => `- ${f.name}  ${fmtSize(f.size)}  ${f.kind}`)
```
mrcfps, line 540: This hint now tells the model that every project file is "flat" and that `Read` should only receive a basename, but the project file APIs do not work that way. `listFiles()` returns relative paths, `writeProjectFile()` preserves nested paths, and the attachment hint below already passes paths like `ui-design/home/index.html` through unchanged. In API mode, a model following this instruction will strip the directory prefix, so `Read`/`Write` calls start failing (or create the file in the wrong place) as soon as a user uploads a folder or imports an archive. Please change this guidance to "use the relative path exactly as listed/attached" and avoid hard-coding `.od/projects/${project.id}` as the working directory, since the daemon can relocate project storage.

mrcfps (commented):
@saikastxy Thanks for the update — I found one follow-up in the new API-mode file/tool guidance. The agent loop is heading in a good direction, but this prompt text will misroute nested project files in a few common flows.


lefarcen:
Hi @saikastxy — I started looking at this to resolve the merge conflict, but stopped to flag a scope issue first that I think we should sort before any merge work happens.

The PR title and description say this is a one-line `max_tokens` constant bump (8192/16384 → 65535). That part is real and lives in `apps/web/src/state/maxTokens.ts` (+3/-1) plus its test (+1/-1). However the diff also contains a pile of unrelated changes that the title doesn't mention:

**Core code changes (~700 lines, unrelated to the title)**
- `apps/daemon/src/server.ts` +98/-14 — appears to rework the `stop_reason` / runStatus completion logic
- `apps/daemon/src/prompts/discovery.ts` +24/-13
- `apps/web/src/components/ProjectView.tsx` +78/-12 — changes `onDone` handling
- `apps/web/src/providers/anthropic.ts` +152/-1
- `apps/web/src/providers/daemon.ts` +36/-5
- `apps/web/src/providers/api-proxy.ts` +8
- **New files**: `TokenUsage.tsx` (+86), `tool-call-parser.ts` (+106), `tool-executor.ts` (+221)

**Documentation artifacts (~1900 lines) that look like agent intermediate output**
- `docs/fixlogs/{fix,fix_1,fix_2,fix_3,fixlog_0,fixlog_1,fixlog_2,fixlog_3}.md` — eight Chinese-language fix-plan documents that read like draft analysis from a code agent (line-numbered breakdowns of "根因分析", proposed patches, etc.). These shouldn't land in the repo; they're working notes, not docs.

**Top-level config**
- `.npmrc` +1, `README.md` +10

Each of those buckets is a distinct change with its own design questions:
- The `stop_reason` / `runStatus` rework in particular is a real semantic change to how the daemon decides a run "succeeded", and would need its own review with a clear motivation, test coverage, and probably a regression case proving the old behavior was wrong. Bundling it under a `max_tokens` PR makes that impossible to review properly.
- The new `tool-call-parser` / `tool-executor` modules are a substantial addition to the web provider layer that deserves its own design discussion.
- `docs/fixlogs/*` definitely shouldn't ship — those are agent scratchpads.

I'm going to close this PR rather than try to merge a 21-file mixed bag under a one-line title. If you'd like to pursue these changes, please re-open them as separate, narrowly-scoped PRs:

1. **A tiny `max_tokens` PR** that only contains the `maxTokens.ts` (+test) change matching the title. I'd be happy to merge that one once it rebases cleanly.
2. **A separate `stop_reason` / `runStatus` PR** for the daemon completion logic, with its own motivation and test coverage.
3. **A separate provider-layer PR** for the new `tool-call-parser` / `tool-executor` modules and the `anthropic.ts` rework, with a design rationale.

Don't include the `docs/fixlogs/*` files in any of them — those are agent working notes and shouldn't be in the repo.

The underlying ideas (especially the runStatus correctness one) sound interesting; they just need to be reviewed on their own merits rather than folded into a constant-bump PR. Thanks for understanding.

PR #351, opened 2026-05-03
fix(daemon): lead copilot spawn with `-p -` so prompt-via-stdin is consumed

Sid-Qin:
## Summary

Closes [#350](https://github.com/nexu-io/open-design/issues/350). GitHub Copilot CLI rejects every Open Design run with \`error: too many arguments. Expected 0 arguments but got N\` and exits code 1, leaving the agent unusable on \`main\` / 0.2.0.

Same regression family as [#237](https://github.com/nexu-io/open-design/issues/237) (codex) and the original [#315](https://github.com/nexu-io/open-design/issues/315) motivation behind PR [#258](https://github.com/nexu-io/open-design/pull/258) (\"Standardize agent communication via stdin\"). PR #258 moved every adapter onto a shared stdin pipe and dropped the per-prompt argv path, but missed flipping Copilot from interactive mode (where the prompt body is a positional argv element) to \`-p -\` mode (where Copilot reads the body from stdin).

cc @lefarcen — your diagnosis on [#350](https://github.com/nexu-io/open-design/issues/350#issuecomment-) was the spec; this PR implements it.

## Root cause

Without \`-p -\`:
1. The daemon writes the prompt to the child's stdin pipe (because \`promptViaStdin: true\`).
2. Copilot stays in **interactive mode**, doesn't read stdin.
3. The leftover daemon argv (\`--allow-all-tools\`, \`--output-format\`, \`json\`, …) becomes \"too many arguments\" from Copilot's perspective.
4. Copilot exits code 1 before reading anything.

With \`-p -\` as the leading argv pair:
1. Copilot enters **prompt mode** with stdin as the body source.
2. The daemon's stdin pipe is consumed correctly.
3. Trailing flags (\`--allow-all-tools\`, etc.) are interpreted as flags, not as positional excess.

## Changes

- \`apps/daemon/src/agents.ts\` — prepend \`-p\` and \`-\` to the copilot \`buildArgs\` output. Inline comment expanded to document the regression and why \`-p -\` is load-bearing, so the next refactor doesn't re-break it.
- \`apps/daemon/tests/agents.test.ts\` — three new regression cases:
  1. **Base shape** — \`-p -\` lead + the unchanged tail of \`--allow-all-tools --output-format json\`.
  2. **Order under model + add-dir extension** — \`-p -\` stays first even when \`--model\` and repeated \`--add-dir\` flags get appended.
[...]

PR #258, opened 2026-05-02
Standardize agent communication via stdin and remove Windows-specific shims logic

RinZ27:
Standardization of agent communication now moves both Claude Code and GitHub Copilot CLI to stdin-based prompt delivery. 

Moving away from the `-p` command-line argument solves the OS-level length limits (~8KB on Windows with `cmd.exe`) that I noticed were causing issues. While the `needsFilePrompt` fallback served its purpose, transitioning to a clean stdin protocol is the right long-term path as discussed in #188.

Bullet points for the changes:
- **Standardized stdin delivery**: Set `promptViaStdin: true` for Claude and Copilot agents in `agents.ts`.
- **Refactored spawn logic**: Removed the `needsFilePrompt` threshold logic and temp-file management from `server.ts` and `runs.ts`.
- **Improved compatibility**: Set `shell: false` universally while using `createCommandInvocation` to handle Windows npm shims (.cmd/.bat) correctly. Removing user-controlled prompt text from argv eliminates the need for shell-specific escaping and bypasses shell-induced character limits entirely.

Verification pass on Claude Code and GitHub Copilot CLI confirmed they support stdin delivery. Logic for Windows-specific shim detection in the server was replaced with a more robust implementation from the platform package that I integrated to keep the spawn process clean and secure.

Fixes #188

PR #727, opened 2026-05-07
fix(daemon): deliver Copilot prompt via stdin to avoid Windows ENAMETOOLONG (#705)

### LINKED ISSUE #705: v0.4.1 + Windows + Copilot CLI:  spawn failed: spawn ENAMETOOLONG
On Windows using v0.4.1 installed using  all default settings on setup downloaded from 
https://github.com/nexu-io/open-design/releases/download/open-design-v0.4.1/open-design-0.4.1-win-x64-setup.exe 

with Github Copilot CLI gives error "spawn failed: spawn ENAMETOOLONG" on every prompt (even with just a "Hi").

Sid-Qin:
## Summary

- Drops `-p <prompt>` from the GitHub Copilot CLI adapter and sets `promptViaStdin: true`, so the composed prompt rides the existing stdin pipe instead of CreateProcess's `lpCommandLine` (capped at ~32 KB direct, ~8 KB through a `.cmd` shim).
- This unblocks every Windows + Copilot CLI run, which today fails with `spawn ENAMETOOLONG` on any non-trivial prompt — even `"Hi"` expands to several thousand chars after Open Design composes in skills + design-system context (issue #705).
- Tests rewritten to assert the new stdin-only argv shape, plus a mirror of the Claude Code 200_000-char synthetic-prompt guard so no argv entry can ever carry the prompt body again.

Fixes #705.

## Why this transport is correct

The earlier `-p -` attempt (#351) and the argv-bound revert (#466) both pre-dated the upstream confirmation in [github/copilot-cli#1046](https://github.com/github/copilot-cli/issues/1046) (closed January 2026). That issue confirms two stdin-driven invocations work in Copilot CLI:

```bash
echo "1+1" | copilot --model gpt-4.1
cat prompt.txt | copilot --model gpt-4.1
```

The semantics are different from `-p -`:

| Pattern | Behavior |
|---|---|
| `copilot -p <body>` | argv prompt, **fails on Windows for non-trivial prompts** (issue #705) |
| `copilot -p -` | takes `-` as a literal one-character prompt — the dead end PR #466 documented |
| `copilot ...` (no `-p`) + stdin pipe | reads composed prompt from stdin, runs non-interactively — **what this PR uses** |

The daemon already spawns the child with `stdio: ['pipe', 'pipe', 'pipe']` and ends stdin with the composed prompt body when `promptViaStdin: true` (see [`apps/daemon/src/server.ts:4041`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/server.ts#L4041)) — exactly the non-TTY pipe shape Copilot's stdin code path expects. `--allow-all-tools`, `--output-format json`, `--model`, and `--add-dir` are all unaffected and continue to ride argv.

## What's verified

| Check | Result |
|---|---|
| `vitest run tests/agents.test.ts` | 78/78 passing |
| `vitest run tests/critique-spawn-wiring.test.ts tests/structured-streams.test.ts` | 14/14 passing |
| `tsc -p apps/daemon/tsconfig.json --noEmit` | clean |

## What's NOT verified

I don't have a Windows host with Copilot CLI to exercise the actual ENAMETOOLONG path end-to-end. The fix relies on the structural argument that:

1. The previous failure was `-p <body>` blowing CreateProcess — the new transport never puts the body on argv.
2. Upstream issue #1046 is the maintainers acknowledging the omit-`-p`-pipe-stdin transport works.
3. The 200_000-char guard test proves no argv entry can carry the prompt body even at synthetic worst-case size.

If a Windows + Copilot CLI tester (e.g. @josemaria-ccr from #705) can run the desktop build with this branch, that closes the loop. Failing that, the inverse signal is also useful — if it doesn't work, I want to know before this lands.

## Test plan

- [x] Existing daemon test suite stays green (agents, spawn-wiring, structured-streams).
- [x] New regression test: 200_000-char synthetic prompt → no argv entry > 1000 chars + `promptViaStdin: true`.
- [ ] Manual smoke on Windows + Copilot CLI 1.0.42: send `"Hi"` and a multi-thousand-char design brief, confirm no ENAMETOOLONG and the JSON stream parses normally.
- [ ] Manual smoke on macOS or Linux + Copilot CLI: confirm no regression in the previously-working argv path.

## Notes on history

- PR #258 standardized stdin delivery, missed Copilot.
- PR #351 tried `-p -` to opt back into stdin — Copilot took `-` literally.
- PR #466 reverted to `-p <body>` argv, which is the state issue #705 surfaces.
- This PR is the third attempt, anchored on the upstream confirmation that omitting `-p` entirely is the working transport.

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -557,39 +557,52 @@ export const AGENT_DEFS = [
     name: 'GitHub Copilot CLI',
     bin: 'copilot',
     versionArgs: ['--version'],
-    // The prompt is passed directly as the value of `-p`: `copilot -p
-    // "<prompt>" --allow-all-tools --output-format json`. Copilot does NOT
-    // treat `-` as a stdin sentinel — it reads it as a literal one-character
-    // prompt string — so the previous `-p -` + stdin pattern produced a
-    // nonsensical single-dash prompt instead of the composed prompt body.
+    // Prompt is delivered via stdin (gated by `promptViaStdin: true`
+    // below) to avoid Windows `spawn ENAMETOOLONG` (issue #705):
+    // `copilot -p <body>` ships the full composed prompt as a single
+    // argv entry, and CreateProcess caps `lpCommandLine` at ~32 KB
+    // direct or ~8 KB through a `.cmd` shim. Any non-trivial Open
+    // Design prompt blows past that — even a "Hi" expands to several
+    // thousand chars after skills + design-system context are composed
+    // in.
     //
-    // `--allow-all-tools` is required for non-interactive runs: without it
-    // the CLI blocks waiting for human approval on every tool call. Unlike
-    // Codex (where `exec` is a dedicated headless subcommand with
-    // auto-approve baked in) or Claude Code (which inherits its permission
-    // policy from the user's settings.json), Copilot's `-p` mode always
-    // prompts unless this flag is passed explicitly.
+    // The transport is "omit `-p` entirely, pipe the prompt to stdin"
+    // per upstream copilot-cli issue #1046 (closed as already supported,
+    // confirmed working on Copilot CLI for `echo "..." | copilot
+    // --model <id>` and `cat prompt.txt | copilot --model <id>`). The
+    // earlier `-p -` attempt (PR #351) and the argv-bound revert
+    // (PR #466) both pre-dated that confirmation: `-p -` made Copilot
+    // interpret `-` as a literal one-character prompt, but omitting
+    // `-p` entirely is a separate code path that does delegate to
+    // stdin under a non-TTY pipe — which is exactly how the daemon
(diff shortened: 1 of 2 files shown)


lefarcen:
Hi @Sid-Qin! 🎉
Thanks for the contribution — the stdin transport approach looks like the right fix for the Windows ENAMETOOLONG issue.

— open-design team

mrcfps (approved):
@Sid-Qin I reviewed the Copilot adapter change and tests. The diff keeps the prompt out of argv, sets `promptViaStdin: true`, preserves the model/add-dir flags, and adds regression coverage for the stdin-only argv shape and long prompt case. Thanks for tackling the Windows ENAMETOOLONG failure and documenting the transport clearly — nice work! 🎉


lefarcen:

PR #729, opened 2026-05-07
Add Cloudflare Pages artifact deployment

bulai0408:
## Summary
- add Cloudflare Pages as a first-class deploy provider alongside Vercel
- store Cloudflare credentials separately and auto-derive/create Pages projects per Open Design project
- implement Cloudflare Pages Direct Upload with Wrangler-compatible BLAKE3 asset hashes before deployment manifest submission
- keep the deploy/share UI provider-scoped, including separate copy-link actions for Vercel and Cloudflare links

## Validation
- `pnpm install`
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/deploy.test.ts tests/deploy-routes.test.ts`
- `pnpm --filter @open-design/web test -- tests/components/FileViewer.test.tsx tests/providers/registry.test.ts`
- `pnpm i18n:check`
- `pnpm guard`
- `git diff --check`
- `pnpm typecheck`

## Live smoke
- Redeployed the sample Cloudflare Pages artifact after the BLAKE3 direct-upload correction.
- Verified 

`https://od-ai-3f0e72fe-5a4.pages.dev` returns HTTP 200 with the generated HTML.

CleanShot 2026-05-07 at 12 09 12@2x
CleanShot 2026-05-07 at 12 09 03@2x


apps/web/src/components/FileViewer.tsx
```diff
@@ -2076,6 +2132,69 @@ function HtmlViewer({
...
+  ) {
+    const deployments = await fetchProjectDeployments(projectId);
+    const nextDeploymentsByProvider = deploymentMapForCurrentFile(deployments);
+    const exactDeployment = nextDeploymentsByProvider[providerId] ?? null;
+    const fallbackDeployment = options?.fallbackToExisting
+      ? Object.values(nextDeploymentsByProvider)[0] ?? null
+      : null;
+    const currentDeployment =
+      exactDeployment ?? fallbackDeployment;
+    const resolvedProviderId = exactDeployment?.providerId ?? currentDeployment?.providerId ?? providerId;
```
mrcfps: When the user opens the deploy modal from a specific provider action, this fallback can switch the selected provider to any existing deployment instead of preserving the provider they clicked. For example, if the artifact already has a Vercel deployment and the user clicks "Deploy to Cloudflare Pages", `fallbackDeployment` can pick the Vercel row, `resolvedProviderId` becomes `vercel-self`, and `syncDeployFormFromConfig` changes the modal back to Vercel. That makes the primary Cloudflare path save/deploy against the wrong provider and credentials. Please keep the explicit `providerId` for provider-specific share-menu actions, and only use a fallback deployment to prefill the displayed existing URL/result without changing the selected provider/config.

package.json
```diff
@@ -10,6 +10,8 @@
     "od": "./apps/daemon/dist/cli.js"
   },
   "scripts": {
+    "start": "OD_HOST=0.0.0.0 OD_BIND_HOST=0.0.0.0 pnpm tools-dev start web --prod --daemon-port 7457 --web-port 4200",
+    "stop": "pnpm tools-dev stop",
```
mrcfps: These root lifecycle aliases conflict with the repository workflow contract in `AGENTS.md`, which explicitly says not to add root `pnpm start`/lifecycle aliases and to use `pnpm tools-dev` as the only local lifecycle entry point. Keeping `start`/`stop` here creates an unsupported launch path with hard-coded ports and broad bind hosts, bypassing the port/env/namespace consistency that `tools-dev` is meant to enforce. Please remove these root scripts and keep lifecycle commands under the existing `tools-dev` entrypoint.

mrcfps (changes requested):
@bulai0408 Thanks for the Cloudflare Pages deployment work and the provider-scoped UI/tests — the direct-upload flow is a meaningful addition. I found two issues that can break the intended provider selection and repo lifecycle boundaries, so I’m requesting changes for those before merge.


apps/web/src/components/FileViewer.tsx
```diff
@@ -2076,6 +2132,69 @@ function HtmlViewer({
...
+    const config = await fetchDeployConfig(resolvedProviderId);
+    syncDeployFormFromConfig(resolvedProviderId, config);
```
mrcfps: This still lets a provider-specific share action switch to the wrong provider when the clicked provider has no deployment yet but another provider does. `openDeployModal(option.id)` is the changed share-menu path, but with `fallbackToExisting` enabled these lines pick `fallbackDeployment`, derive `resolvedProviderId` from that existing deployment, fetch that provider's config, and then `syncDeployFormFromConfig` changes `deployProviderId` to the fallback provider. For example, clicking “Deploy to Cloudflare Pages” on an artifact that only has a Vercel deployment will open/save/deploy as Vercel, so the primary Cloudflare path can use the wrong credentials and platform. Please keep the explicit `providerId` for provider-specific actions and use any fallback deployment only to display an existing URL/result, or restrict fallback behavior to a truly generic open path.

mrcfps (changes requested):
@bulai0408 Thanks for continuing the Cloudflare Pages deployment work and for removing the root lifecycle scripts from the latest head. I found one remaining provider-selection blocker in the deploy modal flow that can still send users to the wrong provider, so I’m requesting changes for that before merge. 🙏


apps/web/src/components/FileViewer.tsx
bulai0408: Addressed in 5c5b85d "Preserve explicit deploy provider selection".

`loadDeployProvider` now uses the explicit `providerId` for both `fetchDeployConfig` and `syncDeployFormFromConfig`; the `resolvedProviderId` derivation was removed. `fallbackDeployment` is only written into `setDeployment` / `setDeployResult` so the existing URL/result block can still render — it no longer changes the selected provider/config.

New regression test `keeps the explicitly selected deploy provider when another provider already has a deployment` in `apps/web/tests/components/FileViewer.test.tsx` reproduces the original failure (only Vercel deployed, click "Deploy to Cloudflare Pages") and asserts that only `/api/deploy/config?providerId=cloudflare-pages` is fetched and the Cloudflare token field is populated.

mrcfps (dismissed):
@bulai0408 I reviewed the current Cloudflare Pages deployment changes across the daemon deploy/config routes, provider-scoped web UI, contract updates, persistence metadata, and the added regression coverage for provider selection and direct-upload behavior. The earlier provider-selection blocker is addressed in this head, and I did not find any remaining actionable correctness or safety issues in the changed ranges. Thanks for the thoughtful follow-through and for adding coverage around the new Cloudflare path 🙏


mrcfps (approved):
@bulai0408 I reviewed the current Cloudflare Pages deployment changes across the daemon config/deploy routes, Direct Upload asset hashing and manifest flow, persisted provider metadata, provider-scoped web UI, and the added daemon/web regression coverage. The previously reported lifecycle and provider-selection blockers are addressed in this head, and I did not find any remaining actionable correctness or safety issues in the changed ranges. Thanks for the careful follow-through and for adding coverage around the new Cloudflare path 🙏


lefarcen:
Merged! 🎉 Thanks @bulai0408 for the Cloudflare Pages contribution to open-design. Looking forward to the next one!

PR #732, opened 2026-05-07
fix(web): replace window.prompt() Save-as-template flow with in-app modal (#723)

Sid-Qin:
## Summary

- The Share menu's `Save as template…` action silently no-op'd in the desktop app. The handler at [`apps/web/src/components/FileViewer.tsx:2610`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/FileViewer.tsx#L2610) called `window.prompt()` twice (name then description), but Electron disables `window.prompt()` by default — both calls return `null` synchronously and the handler's `if (!name) return` exits before any visible UI happens. Reporter saw no modal, no loading state, no error: total silence.
- Replace the prompt-driven flow with an in-app modal that mirrors the existing `.deploy-modal` pattern in the same file. The same handler wiring around `saveTemplate()` and the transient "Saved" pill in the share menu is preserved — only the input collection step changes.
- Adds 1 new i18n key (`fileViewer.templateModalTitle`) across all 16 locales; reuses existing `templateNamePrompt` / `templateNameDefault` / `templateDescPrompt` strings as the field labels and default value, so no translator churn beyond the new title.

Fixes #723.

## Why `window.prompt()` is the wrong primitive here

| | Browser dev | Electron desktop (shipped app) |
|---|---|---|
| `window.prompt()` returns | a string or null | **always null** (silently disabled) |
| Visible to user | yes | nothing |

Electron disables `window.prompt()` by default with no documented opt-in (Chromium upstream removed the dialog and Electron has not re-added it). The result for any handler that does `const x = window.prompt(...); if (!x) return;` is a guaranteed silent no-op for every desktop user — exactly what #723 reports.

## Scope of the change

| File | What changes |
|---|---|
| `apps/web/src/components/FileViewer.tsx` | Add modal state (`templateModalOpen`, `templateName`, `templateDescription`, `templateError`); split `handleSaveAsTemplate` into `openSaveAsTemplateModal()` (opens the modal, seeds defaults) and `submitSaveAsTemplate()` (the previous async save call, now driven by modal state); add modal JSX after the deploy modal. |
| `apps/web/src/i18n/types.ts` | Declare `'fileViewer.templateModalTitle': string` on `Dict`. |
| `apps/web/src/i18n/locales/*.ts` (16 files) | Add per-locale translation, derived from the existing `saveAsTemplate` value minus the trailing ellipsis. |
| `apps/web/src/index.css` | Add minimal `.template-modal` / `.template-form` / `.template-field` rules — reuses the shared `.modal-backdrop` / `.modal` / `.modal-head` / `.modal-foot` shell. |

19 files changed, +153 / −17.

## What's verified

| Check | Result |
|---|---|
| `pnpm --filter @open-design/web test tests/components/FileViewer.test.tsx` | 9/9 passing |
| Live `pnpm tools-dev run web` boot | clean (no console / build errors) |
| `.template-modal` / `.template-form` / `.template-field` rules in live stylesheet | yes (verified by reading `document.styleSheets`) |
| `fileViewer.templateModalTitle` key in compiled JS bundle | yes (verified by fetching script chunks) |
| Residual `window.prompt(t('fileViewer.templateNamePrompt'), …)` callsites in compiled JS | **0** (regex grep across all 32 emitted chunks) |

## What's NOT verified

The bug only reproduces in the **packaged Electron desktop app** (`window.prompt()` works fine in any normal browser, including the dev server). I don't have a desktop build pipeline set up, so I can't end-to-end exercise "click Save as template → see the modal → enter a name → see the saved pill" in a real Electron environment from here. The structural argument is:

1. The legacy callsite is gone from the compiled bundle (verified above).
2. The new modal renders from React state, with no `window.prompt()` / `window.confirm()` involvement at all.
3. The deploy modal in the same file uses the identical class shell and works in Electron today.

If a desktop tester (e.g. someone running a fresh build of #723's branch) can confirm the menu item now opens the modal, that closes the loop end-to-end.

## Out-of-scope follow-up

`apps/web/src/components/SketchEditor.tsx:129` has the same Electron `window.prompt()` regression for the sketch "add text" tool. That's the only other `window.prompt()` callsite in the web app. Tracking it separately so this PR stays focused on the FileViewer Share menu the issue reports — happy to follow up with a second PR using the same modal pattern.

## Test plan

- [x] Existing FileViewer test suite stays green.
- [x] Web typecheck + i18n-check pass.
- [x] Bundle/CSS verification: new rules and i18n key present in live build, legacy `window.prompt(…templateNamePrompt…)` callsite gone.
- [ ] Manual smoke on the desktop build: open a project → open a generated file → Share → `Save as template…` → expect the new modal (with name / description fields) instead of nothing.
- [ ] Visual check that the modal sizing + spacing matches the rest of the Open Design modal language across light + dark themes.

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -2031,9 +2031,17 @@ function HtmlViewer({
   const [presentMenuOpen, setPresentMenuOpen] = useState(false);
   const [shareMenuOpen, setShareMenuOpen] = useState(false);
   // Template save UX. We surface a transient "Saved" pill in the share
-  // menu so the user gets feedback without a noisy toast layer.
+  // menu so the user gets feedback without a noisy toast layer. The
+  // modal carries the name + description input because the desktop app
+  // runs in Electron, where `window.prompt()` is disabled by default
+  // and silently returns null — making the previous prompt-based flow
+  // a no-op for every desktop user (issue #723).
   const [savingTemplate, setSavingTemplate] = useState(false);
   const [templateNote, setTemplateNote] = useState<string | null>(null);
+  const [templateModalOpen, setTemplateModalOpen] = useState(false);
+  const [templateName, setTemplateName] = useState('');
+  const [templateDescription, setTemplateDescription] = useState('');
+  const [templateError, setTemplateError] = useState<string | null>(null);
   const [deployment, setDeployment] = useState<DeployProjectFileResponse | null>(null);
   const [deployModalOpen, setDeployModalOpen] = useState(false);
   const [deployConfig, setDeployConfig] = useState<DeployConfigResponse | null>(null);
@@ -2607,33 +2615,48 @@ function HtmlViewer({
   // the viewer), so the template captures the whole design, not a single
   // page. Surfaced here in the Share menu because that's where the user's
   // share / export mental model already lives.
-  async function handleSaveAsTemplate() {
+  function openSaveAsTemplateModal() {
     setShareMenuOpen(false);
     const defaultName =
       file.name.replace(/\.html?$/i, '') || t('fileViewer.templateNameDefault');
-    const name = window.prompt(t('fileViewer.templateNamePrompt'), defaultName);
-    if (!name || !name.trim()) return;
-    const description = window.prompt(
-      t('fileViewer.templateDescPrompt'),
-      '',
-    );
+    setTemplateName(defaultName);
+    setTemplateDescription('');
+    setTemplateError(null);
+    setTemplateNote(null);
+    setTemplateModalOpen(true);
+  }
+
+  async function submitSaveAsTemplate() {
+    const trimmedName = templateName.trim();
+    if (!trimmedName) {
+      // The Save button is also disabled when the name is empty so this
+      // path is normally unreachable, but guard the network call anyway
+      // in case Enter-to-submit fires before the disabled state catches up.
+      return;
+    }
     setSavingTemplate(true);
+    setTemplateError(null);
     setTemplateNote(null);
     try {
       const tpl = await saveTemplate({
-        name: name.trim(),
-        description: description?.trim() || undefined,
+        name: trimmedName,
+        description: templateDescription.trim() || undefined,
         sourceProjectId: projectId,
       });
-      setTemplateNote(
-        tpl
-          ? t('fileViewer.savedTemplate', { name: tpl.name })
-          : t('fileViewer.savedTemplateFail'),
+      if (tpl) {
+        setTemplateModalOpen(false);
+        setTemplateNote(t('fileViewer.savedTemplate', { name: tpl.name }));
+        // Auto-clear the note so the menu doesn't keep stale state next open.
+        setTimeout(() => setTemplateNote(null), 4000);
+      } else {
+        setTemplateError(t('fileViewer.savedTemplateFail'));
+      }
+    } catch (err) {
+      setTemplateError(
+        err instanceof Error ? err.message : t('fileViewer.savedTemplateFail'),
       );
     } finally {
       setSavingTemplate(false);
-      // Auto-clear the note so the menu doesn't keep stale state next open.
-      setTimeout(() => setTemplateNote(null), 4000);
     }
   }
 
@@ -3140,7 +3163,7 @@ function HtmlViewer({
                     role="menuitem"
                     disabled={savingTemplate}
                     onClick={() => {
-                      void handleSaveAsTemplate();
+                      openSaveAsTemplateModal();
                     }}
                   >
                     <span className="share-menu-icon"><Icon name="copy" size={14} /></span>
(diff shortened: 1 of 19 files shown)


Sid-Qin (commented):
Self-review of d0d2841. Findings ordered by severity. No `[P1]` blockers, but a couple of `[P2]` regressions vs the original `window.prompt()` UX that I'd want to address before merge if a maintainer agrees.

## [P2] Lost text-pre-selection on focus — UX regression vs the prompt this replaces

`window.prompt(label, defaultValue)` opens with the default value **pre-selected**, so the user can immediately type to overwrite. The new modal places the cursor at the end (browser default for `autoFocus` on a value-bound input). Users now have to ⌘A → type, instead of just typing.

Fix is one ref + one effect:

```tsx
const nameInputRef = useRef<HTMLInputElement | null>(null);
useEffect(() => {
  if (templateModalOpen) nameInputRef.current?.select();
}, [templateModalOpen]);
```

Then `<input ref={nameInputRef} … />` instead of `autoFocus`. `select()` focuses + selects the existing text, matching the `window.prompt()` behaviour the user just lost.

## [P2] No Escape-to-close on the modal

The dialog ignores the Escape key. Standard modal expectation, especially for a flow the user reaches by accident from a pulldown menu. The existing `.deploy-modal` in the same file has the same gap, but copying it shouldn't lock the bug in for a second modal.

`useEffect` listening on `keydown` while `templateModalOpen`, calling `setTemplateModalOpen(false)` on Escape. Mirror in deploy modal as a follow-up if the maintainer wants — out of scope for this issue otherwise.

## [P2] `aria-labelledby` missing on the dialog

`role="dialog" aria-modal="true"` is set, but no `aria-labelledby` pointing at the `<h2>` title. Screen readers announce "dialog" with no name. One-line fix:

```tsx
<h2 id="template-modal-title">{t('fileViewer.templateModalTitle')}</h2>
…
<div className="modal template-modal" role="dialog" aria-modal="true" aria-labelledby="template-modal-title">
```

(Same caveat as Escape: the deploy modal also lacks this, but two wrongs etc.)

## [P3] `.template-modal` width override is redundant

```css
.template-modal { width: min(520px, calc(100vw - 32px)); }
```

The base `.modal` rule already declares `width: 520px; max-width: calc(100vw - 32px);`, which produces the same computed width. The override and its comment claim "tighter width than the deploy modal" but the tightness is purely from **omitting** `.deploy-modal`'s explicit `760px` override, not from this rule. The `.template-modal` selector should probably stay as a hook for future overrides, but the rule body can be deleted (or kept as `{}`/replaced with a comment). Cosmetic, not blocking.

## [P3] Description textarea has no placeholder

The label `templateDescPrompt` says "Short description (optional — what makes this template useful?)" which works as a help text above the field, but the textarea itself is a blank rectangle. Adding a one-line placeholder ("e.g. responsive landing page with hero + pricing grid") would lower the activation cost. Could share a new key or just reuse `templateDescPrompt` as both label + placeholder. Marginal, skip if it adds translator churn.

## [DISCUSS] No behavioural test added for the modal

The PR ships with the existing 9 `FileViewer.test.tsx` tests staying green and structural verification (CSS + i18n key in compiled bundle, zero residual `window.prompt(…templateNamePrompt…)` callsites in any chunk), but no new test that mounts the share menu, clicks `Save as template…`, and asserts the modal renders + submits. The existing tests use `renderToStaticMarkup` (server-side only, no event dispatch) so adding a behavioural test would mean introducing `@testing-library/react` + jsdom + non-trivial mocking of project/file/saveTemplate — meaningful test infra work not currently present in this file.

Asking the question rather than answering it: is the structural-only verification acceptable for a UI bug-fix in this codebase, or does the maintainer want me to set up RTL-style tests as part of this PR?

## What I checked and didn't flag

- Race conditions: Cancel and Save buttons both correctly gate on `savingTemplate`; Enter handler also checks `!savingTemplate` before submit.
- Empty-name guard: button is `disabled` when `templateName.trim()` is empty, AND `submitSaveAsTemplate` early-returns on empty trimmed name as a backup. Belt + suspenders, intentional.
- Modal stays open on save failure with `templateError` populated → user can retry without retyping. Good.
- Successful save closes the modal AND fires the existing 4-second auto-clearing pill in the share menu. Continuity with previous UX preserved.
- `setTemplateNote(null)` in `openSaveAsTemplateModal` is a deliberate (and correct) behavior change — re-opening the flow clears the previous "Saved" pill instead of leaving it visible mid-flow.
- 16 locale translations all derived from existing `saveAsTemplate` minus the trailing ellipsis, so no machine-translation drift.

mrcfps (approved):
@Sid-Qin I reviewed the Save-as-template flow changes, including the new in-app modal state/submission path, the i18n key coverage, and the added modal styling. The implementation removes the Electron-incompatible `window.prompt()` dependency while preserving the existing `saveTemplate()` request behavior, and I didn't find any correctness, safety, or maintainability issues in the changed ranges. Thanks for tightening up this desktop flow and documenting the verification so clearly 🙏


lefarcen:
@Sid-Qin Heads up: this PR has a merge conflict with `main` that needs resolving before it can merge. mrcfps's approval still stands once the conflict is fixed.

You can resolve it by rebasing on main:
```bash
git fetch origin
git rebase origin/main
# resolve conflicts, then
git push --force-with-lease
```

The review feedback (3 P2 UX improvements) is orthogonal to the conflict — feel free to address those in a follow-up commit or handle them separately.

Sid-Qin:
Closing this as superseded. Main now ships its own implementation of the same fix via direct commits \`a173a5e\` / \`1ae8d99\` (2026-05-07, by @shangxinyu1) that landed shortly after this PR opened. The core behavior — \`templateModalOpen\` state, \`templateSaveError\`, \`handleSaveAsTemplate\` async flow, in-app modal replacing \`window.prompt()\` — is in place and the i18n keys (templateNamePrompt / templateDescPrompt / etc.) overlap.

I'd already self-reviewed two P2 polish items here that aren't addressed in the merged version:

1. **Text pre-selection on focus** — the merged version has \`autoFocus\` but doesn't \`.select()\` the pre-filled \`<filename>\` so users can immediately type-to-replace. The original \`window.prompt()\` did select-by-default; without that, every save-as-template forces an extra Cmd+A.
2. **Submit-on-Enter visibility** — disabled in the merged variant when the name field is empty, but no aria-live signal to screen readers.

Happy to spin those into a small follow-up PR if maintainers think they're worth shipping. Otherwise this PR has done its job (the issue is fixed) and there's no point keeping a parallel implementation alive.

Thanks @mrcfps and @lefarcen for the reviews on the way through.

PR #735, opened 2026-05-07
fix(daemon): allow portless Origin in CORS whitelist for Chrome compatibility

### LINKED ISSUE #733: fix: Agent scan fails with CORS 403 on Chrome (portless Origin header)
### Description

In Chrome normal mode (not incognito), clicking the "Rescan" button in Settings fails with:

> 扫描失败，请检查后台守护进程后重试。

The browser console shows a `403 Forbidden` response from `/api/connectors/status` (and other `/api/*` endpoints) with the body:

```json
{"error": "Cross-origin requests are not allowed"}
```

Incognito mode works without issue.

### Environment

| | |
|---|---|
| **OS** | Windows 11 (26200) |
| **Browser** | Chrome 147.0.0.0 |
| **Node.js** | v25.2.1 |
| **Open Design** | v0.4.1 (local dev via `pnpm tools-dev start web`) |
| **Ports** | Web: `http://127.0.0.1:<random>`, Daemon: `http://127.0.0.1:<random>` |

### Steps to reproduce

1. `pnpm tools-dev start web`
2. Open the Web URL in Chrome **normal mode** (not incognito)
3. Go to Settings → click "↻ Rescan"
4. Observe: scan fails with error toast

### Expected behavior

Agent scan should succeed and detect available CLI agents (e.g. Claude Code) regardless of browser mode.

### Actual behavior

- **Chrome normal mode**: scan fails, 403 on `/api/*` endpoints
- **Chrome incognito mode**: scan succeeds
- **`curl` with explicit port in Origin**: succeeds
- **`curl` without port in Origin** (`-H "Origin: http://127.0.0.1"`): 403

### Diagnosis

Chrome 147 strips the port from the `Origin` header on same-origin requests to localhost with non-standard ports.

From the Network tab, the failing request sends:

```
Origin: http://127.0.0.1
```

instead of the expected:

```
Origin: http://127.0.0.1:6313
```

The daemon's CORS whitelist in `buildAllowedOrigins()` (`apps/daemon/src/server.ts:1321`) only constructs origins **with** explicit ports:

```typescript
// e.g. http://127.0.0.1:6313, http://localhost:6313, ...
ports.flatMap((p) => [
  ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
  ...
])
```

The portless origin `http://127.0.0.1` does not match any entry in the set, so the middleware rejects it with 403.

The same logic exists in `isLocalSameOrigin()` (`apps/daemon/src/server.ts:4987`), which is used by individual routes like `/api/active`, `/api/app-config`, and media generation endpoints.

### Why incognito is unaffected

Chrome may handle Origin header serialization differently when there is no cached browsing state. This is consistent with observed behavior but not documented in Chromium's Origin header specification.

### Suggested fix

Add portless loopback origins to the allowed set in both functions:

**`buildAllowedOrigins()`** — add after the existing `ports.flatMap(...)`:
```typescript
...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
...schemes.map((s) => `${s}://${host}`),
```

**`isLocalSameOrigin()`** — same addition to `allowedOrigins`.

This preserves the existing security model (only localhost origins are allowed) while being tolerant of how browsers serialize the Origin header. No non-local origins are admitted.

### Impact

- Affects all Chrome users on local dev with non-standard ports
- Does not affect production deployments (standard ports 80/443)
- Does not affect Firefox or Safari (they include the port per spec)
- Low security risk: the fix only relaxes format tolerance for loopback origins, not the origin scope itself

JHR-deve:
## Summary

- Fix CORS 403 on `/api/*` endpoints when Chrome strips the port from the `Origin` header on same-origin requests (e.g. `Origin: http://127.0.0.1` instead of `http://127.0.0.1:6313`)
- Add portless loopback origins to both `buildAllowedOrigins()` and `isLocalSameOrigin()` in `apps/daemon/src/server.ts`

Closes #733

## Test plan

- [x] `curl -H "Origin: http://127.0.0.1" http://127.0.0.1:<port>/api/agents` → 200 (was 403)
- [x] Chrome normal mode: Settings → Rescan → success
- [x] Chrome incognito mode: still works (no regression)
- [x] `curl` without Origin header (non-browser client) → still works

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -1324,21 +1324,25 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     if (webPort && webPort !== resolvedPort) ports.push(webPort);
     const schemes = ['http', 'https'];
     const loopbackHosts = ['127.0.0.1', 'localhost', '[::1]'];
-    return new Set([
-      // Origins with explicit ports.
-      ...ports.flatMap((p) => [
+    return new Set(
+      ports.flatMap((p) => [
         ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
         // When bound to a specific non-loopback address (e.g. Tailscale,
         // LAN IP, or 0.0.0.0), allow browser requests from that address
         // too so the documented --host escape hatch remains usable.
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
-      // Origins without ports — Chrome may strip the port from the
-      // Origin header on same-origin requests (e.g. Origin: http://127.0.0.1
-      // instead of http://127.0.0.1:6313).
-      ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
-      ...schemes.map((s) => `${s}://${host}`),
-    ]);
+    );
+  }
+
+  // Portless loopback origins (e.g. http://127.0.0.1 without a port).
+  // Chrome may strip the port from the Origin header on same-origin GET
+  // requests. Only used as a fallback for safe, idempotent GET requests;
+  // mutating routes (POST/PUT/PATCH/DELETE) always require an exact
+  // port-match via buildAllowedOrigins() or isLocalSameOrigin() to
+  // prevent local CSRF from a page on the default port (80).
+  function isPortlessLoopbackOrigin(origin) {
+    return /^https?:\/\/(127\.0\.0\.1|localhost|\[::1])(|$)/.test(origin);
   }
 
   // Routes that serve content to sandboxed iframes (Origin: null) for
@@ -1376,7 +1380,16 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     }
 
     if (!buildAllowedOrigins().has(String(origin))) {
-      return res.status(403).json({ error: 'Cross-origin requests are not allowed' });
+      // Fallback: Chrome may strip the port from the Origin header on
+      // same-origin requests (e.g. http://127.0.0.1 instead of
+      // http://127.0.0.1:6313). Allow portless loopback origins only
+      // for GET requests, which are idempotent and safe from CSRF.
+      // Mutating methods (POST/PUT/PATCH/DELETE) always require an
+      // exact port-match to prevent a page on the default port (80)
+      // from triggering state-changing operations.
+      if (req.method !== 'GET' || !isPortlessLoopbackOrigin(String(origin))) {
+        return res.status(403).json({ error: 'Cross-origin requests are not allowed' });
+      }
     }
     next();
   });
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -5012,15 +5012,11 @@ export function isLocalSameOrigin(req, port) {
   if (origin == null || origin === '') return true;
 
   const schemes = ['http', 'https'];
-  const allowedOrigins = new Set([
-    ...ports.flatMap((p) => [
+  const allowedOrigins = new Set(
+    ports.flatMap((p) => [
       ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
       ...schemes.map((s) => `${s}://${bindHost}:${p}`),
     ]),
-    // Origins without ports — Chrome may strip the port from the
-    // Origin header on same-origin requests.
-    ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
-    ...schemes.map((s) => `${s}://${bindHost}`),
-  ]);
+  );
   return allowedOrigins.has(String(origin));
 }
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -1324,15 +1324,21 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     if (webPort && webPort !== resolvedPort) ports.push(webPort);
     const schemes = ['http', 'https'];
     const loopbackHosts = ['127.0.0.1', 'localhost', '[::1]'];
-    return new Set(
-      ports.flatMap((p) => [
+    return new Set([
+      // Origins with explicit ports.
+      ...ports.flatMap((p) => [
         ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
         // When bound to a specific non-loopback address (e.g. Tailscale,
         // LAN IP, or 0.0.0.0), allow browser requests from that address
         // too so the documented --host escape hatch remains usable.
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
-    );
+      // Origins without ports — Chrome may strip the port from the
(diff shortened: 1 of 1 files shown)


JHR-deve:
Good catch from Codex. I've pushed a follow-up commit that narrows the fix:

**What changed:**
- `buildAllowedOrigins()` (global CORS middleware) — keeps the portless origin fix. This guards read-only endpoints like `/api/agents`, `/api/skills`, `/api/connectors/status`. These are the endpoints the "Rescan" button needs.
- `isLocalSameOrigin()` (per-route sensitive check) — **reverted to strict port matching**. This guards write/system operations: `PUT /api/app-config`, `POST /api/dialog/open-folder`, `POST /api/media/generate`, etc.

**Verification:**
```
# Read endpoint — portless Origin accepted (Chrome fix works)
curl -H "Origin: http://127.0.0.1" http://127.0.0.1:<port>/api/agents → 200

# Sensitive endpoint — portless Origin rejected (security maintained)
curl -H "Origin: http://127.0.0.1" -X PUT http://127.0.0.1:<port>/api/app-config → 403
```

This addresses the port-80 attack vector: a local service on default port 80 can no longer trigger sensitive operations via portless Origin, while the read-only agent scan still works for Chrome users.

apps/daemon/src/server.ts
```diff
@@ -1324,15 +1324,21 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
         // When bound to a specific non-loopback address (e.g. Tailscale,
         // LAN IP, or 0.0.0.0), allow browser requests from that address
         // too so the documented --host escape hatch remains usable.
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
-    );
+      // Origins without ports — Chrome may strip the port from the
+      // Origin header on same-origin requests (e.g. Origin: http://127.0.0.1
+      // instead of http://127.0.0.1:6313).
+      ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
```
mrcfps: This portless loopback allowlist is still applied by the global `/api` middleware, not just by the read-only scan endpoints. That middleware is the only `Origin` gate for many mutating routes in this file, for example `PUT /api/media/config`, `POST /api/projects`, artifact writes, and deploy/config endpoints that do not call `isLocalSameOrigin()`. With this line in the shared allowed-origin set, a page served from the default local origin (`http://localhost` / `http://127.0.0.1`, i.e. port 80) is treated as trusted for those routes even though it is not the Open Design UI port. That reintroduces a local CSRF-style path for state changes while fixing Chrome's portless `Origin` case. Please narrow the portless exception to the specific safe read-only endpoints that need it (for example before this allowlist check, only for `GET /api/agents`/rescan-related routes), or add the stricter local same-origin guard to every mutating route before accepting portless origins globally.

mrcfps (changes requested):
@JHR-deve thanks for the careful follow-up that restored the stricter `isLocalSameOrigin()` checks. I found one remaining security issue in the global origin allowlist that should be narrowed before merge.


JHR-deve:
Thanks for the thorough review. You're right — the global `/api` middleware guards many mutating routes that don't have their own `isLocalSameOrigin()` check, so portless origins in `buildAllowedOrigins()` opens a local CSRF path.

**Updated approach:** Moved the portless check out of `buildAllowedOrigins()` entirely. Instead, it's now a **GET-only fallback** in the global middleware:

```typescript
if (!buildAllowedOrigins().has(String(origin))) {
  // Allow portless loopback origins only for GET (idempotent, CSRF-safe).
  // POST/PUT/PATCH/DELETE always require exact port-match.
  if (req.method !== 'GET' || !isPortlessLoopbackOrigin(String(origin))) {
    return res.status(403).json({ error: 'Cross-origin requests are not allowed' });
  }
}
```

`buildAllowedOrigins()` is back to strict port-match only. `isLocalSameOrigin()` is unchanged.

**Verified:**
| Scenario | Method | Origin | Result |
|---|---|---|---|
| Chrome fix | GET | `http://127.0.0.1` (portless) | 200 |
| Security | PUT | `http://127.0.0.1` (portless) | 403 |
| Normal | GET | `http://127.0.0.1:<web-port>` | 200 |

apps/daemon/src/server.ts
```diff
@@ -1335,6 +1335,16 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
   }
 
+  // Portless loopback origins (e.g. http://127.0.0.1 without a port).
+  // Chrome may strip the port from the Origin header on same-origin GET
+  // requests. Only used as a fallback for safe, idempotent GET requests;
+  // mutating routes (POST/PUT/PATCH/DELETE) always require an exact
+  // port-match via buildAllowedOrigins() or isLocalSameOrigin() to
+  // prevent local CSRF from a page on the default port (80).
+  function isPortlessLoopbackOrigin(origin) {
+    return /^https?:\/\/(127\.0\.0\.1|localhost|\[::1])(|$)/.test(origin);
```
mrcfps: The `(|$)` suffix makes this matcher succeed immediately after the loopback hostname because the empty alternative can always match. That means the GET fallback is not limited to portless origins: values like `http://localhost:5173` or even `http://localhost.example.test` also pass this check, bypassing the exact-port allowlist for every GET `/api` route. The surrounding comment says this exception is only for Chrome's portless loopback origin case, so the implementation should require the origin to end right after `localhost`, `127.0.0.1`, or `[::1]`. A small fix would be to use an exact regex such as `/^https?:\/\/(?:127\.0\.0\.1|localhost|\[::1\])$/`, or parse with `new URL(origin)` and require a loopback hostname plus an empty `port`.

mrcfps (commented):
@JHR-deve thanks for tightening the follow-up so the portless fallback no longer applies to mutating requests. I found one small but concrete issue in the new loopback-origin matcher that is worth addressing, but it should be straightforward to fix. 🙂


JHR-deve:
Good catch — the `(|$)` alternation was broken. Fixed to a strict `$` anchor:

```javascript
// Before (buggy): matches http://localhost:5173, http://localhost.example.test
/^https?:\/\/(127\.0\.0\.1|localhost|\[::1\])(|$)/

// After (correct): only matches portless loopback origins
/^https?:\/\/(127\.0\.0\.1|localhost|\[::1\])$/
```

Verified:
```
http://127.0.0.1         → true  ✓
http://localhost          → true  ✓
http://[::1]             → true  ✓
https://127.0.0.1        → true  ✓
http://localhost:5173    → false ✓
http://127.0.0.1:8080    → false ✓
http://localhost.example → false ✓
```

mrcfps (approved):
@JHR-deve I reviewed the updated daemon origin handling and verified the portless loopback fallback is now limited to exact loopback origins on GET requests while mutating routes still require the strict port-matched origin policy. Thanks for iterating on the earlier security feedback and landing a focused Chrome compatibility fix. 🙂


leilei926524-tech:
Hi @JHR-deve — welcome to Open Design contributors. 🎉

Your first merged PR (#735) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #757, opened 2026-05-07
fix: expand Codex picker coverage

leprincep35700:
## Summary
- Expand the Codex CLI fallback model picker in recommended priority order: `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`, then the existing older choices.
- Add `none` and `xhigh` to the Codex reasoning-effort picker options while keeping the existing options for compatibility.
- Strengthen regression coverage so the test verifies the static picker list, reasoning options, `buildArgs()` model/reasoning forwarding, and `detectAgents()` surfacing through a fake Codex binary.

## Why
Codex CLI does not expose a `models` command here, so Open Design relies on a static fallback list for the dropdown. Users who can run newer OpenAI models through their Codex account should be able to select those ids directly without using the custom-model input every time, and the picker should guide them toward current model choices.

## Testing
- `corepack pnpm --filter @open-design/daemon test -- apps/daemon/tests/agents.test.ts` (passes; daemon Vitest suite: 918 tests)
- `corepack pnpm guard` (passes)
- `corepack pnpm --filter @open-design/daemon typecheck` (passes)

## Notes
The local environment reports Node v20.19.2 while the repo requests Node ~24, so pnpm prints engine warnings during validation.

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -197,19 +197,23 @@ export const AGENT_DEFS = [
     // as a hint. Users can supply other ids via the custom-model input.
     fallbackModels: [
       DEFAULT_MODEL_OPTION,
-      { id: 'gpt-5-codex', label: 'gpt-5-codex' },
       { id: 'gpt-5.5', label: 'gpt-5.5' },
       { id: 'gpt-5.4', label: 'gpt-5.4' },
+      { id: 'gpt-5.4-mini', label: 'gpt-5.4-mini' },
+      { id: 'gpt-5.3-codex', label: 'gpt-5.3-codex' },
+      { id: 'gpt-5-codex', label: 'gpt-5-codex' },
       { id: 'gpt-5', label: 'gpt-5' },
       { id: 'o3', label: 'o3' },
       { id: 'o4-mini', label: 'o4-mini' },
     ],
     reasoningOptions: [
       { id: 'default', label: 'Default' },
+      { id: 'none', label: 'None' },
       { id: 'minimal', label: 'Minimal' },
       { id: 'low', label: 'Low' },
       { id: 'medium', label: 'Medium' },
       { id: 'high', label: 'High' },
+      { id: 'xhigh', label: 'XHigh' },
     ],
     // Prompt is delivered via stdin pipe (gated by `promptViaStdin: true`
     // below) to avoid Windows `spawn ENAMETOOLONG` while keeping Codex on
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -198,6 +198,8 @@ export const AGENT_DEFS = [
     fallbackModels: [
       DEFAULT_MODEL_OPTION,
       { id: 'gpt-5-codex', label: 'gpt-5-codex' },
+      { id: 'gpt-5.5', label: 'gpt-5.5' },
+      { id: 'gpt-5.4', label: 'gpt-5.4' },
       { id: 'gpt-5', label: 'gpt-5' },
       { id: 'o3', label: 'o3' },
       { id: 'o4-mini', label: 'o4-mini' },
--- a/apps/daemon/tests/agents.test.ts
+++ b/apps/daemon/tests/agents.test.ts
@@ -138,16 +138,62 @@ test('codex args keep plugins enabled when OD_CODEX_DISABLE_PLUGINS is not 1', (
   assert.equal(args.includes('plugins'), false);
 });
 
-test('codex model picker includes explicit newer OpenAI choices', () => {
-  assert.deepEqual(codex.fallbackModels.map((m) => m.id), [
+test('codex model picker includes current OpenAI choices in priority order', async () => {
+  const expectedModels = [
     'default',
-    'gpt-5-codex',
     'gpt-5.5',
     'gpt-5.4',
+    'gpt-5.4-mini',
+    'gpt-5.3-codex',
+    'gpt-5-codex',
     'gpt-5',
     'o3',
     'o4-mini',
+  ];
+
+  assert.deepEqual(codex.fallbackModels.map((m) => m.id), expectedModels);
+  assert.deepEqual(codex.reasoningOptions.map((o) => o.id), [
+    'default',
+    'none',
+    'minimal',
+    'low',
+    'medium',
+    'high',
+    'xhigh',
   ]);
+
+  const args = codex.buildArgs(
+    '',
+    [],
+    [],
+    { model: 'gpt-5.5', reasoning: 'xhigh' },
+    { cwd: '/tmp/od-project' },
+  );
+  assert.ok(args.includes('--model'));
+  assert.ok(args.includes('gpt-5.5'));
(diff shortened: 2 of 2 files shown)


mrcfps (approved):
@leprincep35700 I reviewed the changed ranges in `apps/daemon/src/agents.ts` and `apps/daemon/tests/agents.test.ts`. The update cleanly keeps the synthetic Codex default from being inferred as a concrete model while preserving clamping for explicit newer gpt-5 minors, and the regression matrix covers the default/custom-model behavior. Thanks for tightening this up — nice, focused fix. 🙌


leprincep35700:
@mrcfps quick update: I ended up changing the PR direction after re-checking the actual user need.

Instead of changing the synthetic `Default (CLI config)` reasoning behavior, this now keeps the existing default handling and adds the explicit `gpt-5.5` and `gpt-5.4` choices to the Codex CLI fallback model picker, with a regression test pinning that dropdown order.

What do you think about this shape? It is a smaller UI/picker-focused change and avoids changing runtime behavior for users who leave Codex on its CLI default.

mrcfps (approved):
@leprincep35700 I reviewed the changed ranges in `apps/daemon/src/agents.ts` and `apps/daemon/tests/agents.test.ts`. This cleanly adds the explicit `gpt-5.5` and `gpt-5.4` Codex fallback model choices without changing default CLI behavior, and the regression test pins the picker order so future edits won’t accidentally drop them. Thanks for the focused follow-up and for keeping the runtime behavior conservative. 🙌


lefarcen:
Hi @leprincep35700! 🎉
Thanks for the contribution — adding these newer model choices is a nice UX win for Codex users.

— open-design team

apps/daemon/src/agents.ts
```diff
@@ -198,6 +198,8 @@ export const AGENT_DEFS = [
     fallbackModels: [
       DEFAULT_MODEL_OPTION,
       { id: 'gpt-5-codex', label: 'gpt-5-codex' },
+      { id: 'gpt-5.5', label: 'gpt-5.5' },
```
lefarcen, line 200: P3: The Codex fallback picker now puts `gpt-5-codex` ahead of `gpt-5.5`/`gpt-5.4`, and omits peers like `gpt-5.4-mini` and `gpt-5.3-codex` that Codex docs mention. Since there's no runtime probe, this static list is the UI order users see. Consider matching Codex's recommended priority (likely `default`, `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`, then older choices) so the dropdown guides users toward current-best.

Not a blocker — current order still works, and users can custom-input any model.

apps/daemon/src/agents.ts
```diff
@@ -198,6 +198,8 @@ export const AGENT_DEFS = [
...
+      { id: 'gpt-5.4', label: 'gpt-5.4' },
       { id: 'gpt-5', label: 'gpt-5' },
       { id: 'o3', label: 'o3' },
       { id: 'o4-mini', label: 'o4-mini' },
```
lefarcen, line 207: P3: Adding `gpt-5.5`/`gpt-5.4` exposes a reasoning-picker gap: the picker shows `minimal` but not `xhigh` or `none`, while Codex docs list `none | low | medium | high | xhigh` for these models. The clamp logic already knows about `xhigh`, but users can't select it from the dropdown. Either add those options or document why OD intentionally hides them.

Not urgent — current picker covers most use cases.

apps/daemon/tests/agents.test.ts
```diff
@@ -138,6 +138,18 @@ test('codex args keep plugins enabled when OD_CODEX_DISABLE_PLUGINS is not 1', (
   assert.equal(args.includes('plugins'), false);
 });
 
+test('codex model picker includes explicit newer OpenAI choices', () => {
```
lefarcen: P3: The new test pins `codex.fallbackModels` directly, but doesn't exercise the runtime surface users consume (`detectAgents()` / `/api/agents`) or the spawn path (`buildArgs({ model: 'gpt-5.5' })`). A future change could break picker surfacing or model forwarding while this test still passes. Consider adding a small `detectAgents` test with a fake `codex` binary and a `buildArgs` assertion for one new model ID to catch regressions earlier.

Not blocking this PR — the static-list test is fine for now.

lefarcen (commented):
Thanks @leprincep35700! The change looks good — clean addition of the two newer model choices. A few optional refinements below:

apps/daemon/src/agents.ts
leprincep35700: Thanks, good point. I updated the fallback picker to follow that priority now: `default`, `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`, then the older existing choices (`gpt-5-codex`, `gpt-5`, `o3`, `o4-mini`). The regression test now pins that full order so future edits do not accidentally drift it.

apps/daemon/src/agents.ts
leprincep35700: Thanks  agreed. I added `none` and `xhigh` to the Codex reasoning picker options while keeping the existing options for backwards compatibility. The test now also pins the reasoning option ids, and the `buildArgs` assertion verifies `xhigh` is forwarded as `model_reasoning_effort="xhigh"`.

apps/daemon/tests/agents.test.ts
leprincep35700: Thanks, that is a better regression shape. I expanded the test to cover the user-facing/runtime surfaces too: it now checks `buildArgs({ model: 'gpt-5.5', reasoning: 'xhigh' })` forwards the selected model/reasoning, and it runs `detectAgents()` against a fake `codex` binary to verify the models surfaced through agent detection match the picker list.

leprincep35700:
Thanks again for the thoughtful review @lefarcen . I pushed `28f44fd` addressing the three follow-ups:

- reordered and expanded the Codex fallback picker (`gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, `gpt-5.3-codex`, then existing older choices);
- added `none` and `xhigh` to the Codex reasoning picker options;
- strengthened the regression test to cover the static list, reasoning options, `buildArgs()` forwarding, and `detectAgents()` surfacing through a fake Codex binary.

Validation:
- `corepack pnpm --filter @open-design/daemon test -- apps/daemon/tests/agents.test.ts`
- `corepack pnpm guard`
- `corepack pnpm --filter @open-design/daemon typecheck`

mrcfps (approved):
@leprincep35700 I reviewed the changed ranges in `apps/daemon/src/agents.ts` and `apps/daemon/tests/agents.test.ts`. The Codex picker now reflects the expanded current model order, the new reasoning options are forwarded through the existing argument builder without changing default CLI behavior, and the regression coverage exercises both `buildArgs()` and the `detectAgents()` surface. I did not find any actionable issues in this pass. Thanks for the thoughtful follow-up and for tightening the Codex UX here. 🙌


lefarcen:
Merged! 🎉 Thanks @leprincep35700 for the contribution to open-design. Looking forward to the next one!

PR #761, opened 2026-05-07
fix(daemon): accept portless loopback Origin headers in CORS allowlist (#733)

### LINKED ISSUE #733: fix: Agent scan fails with CORS 403 on Chrome (portless Origin header)

Sid-Qin:
## Summary

- Some browser builds (observed on Chrome under Windows when serving localhost on a non-standard port) serialize the `Origin` request header **without the port** — `http://127.0.0.1` instead of `http://127.0.0.1:6313`. The daemon's allowlist only constructs port-bearing origins (`buildAllowedOrigins()` and `isLocalSameOrigin()` in [`apps/daemon/src/server.ts`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/server.ts)), so portless variants miss every entry and every `/api/*` request dies with `403 Cross-origin requests are not allowed`. Incognito mode dodges it because it serializes the port the way the spec says to (issue #733 has the full repro).
- Adds the portless serializations (`http://127.0.0.1`, `http://localhost`, `http://[::1]`, https variants, and the explicit bind host) to both functions. **Trust boundary unchanged**: still loopback-only, plus the documented non-loopback `--host` escape hatch the existing port-bearing entries already allow. `http://evil.com` and other non-loopback portless origins remain blocked.
- Mirrors the patch in [`origin-validation.test.ts`'s `createOriginMiddleware()` helper](https://github.com/nexu-io/open-design/blob/main/apps/daemon/tests/origin-validation.test.ts#L11) so tests exercise the real shape, and adds 6 new tests covering the portless cases — including a negative test confirming `http://evil.com` still 403s.

Fixes #733. Reported with full diagnosis and a candidate patch by @JHR-deve — this PR is essentially landing their proposal verbatim with tests.

## Why the trust boundary doesn't widen

The allowlist construction is keyed on **host**, not host-plus-port. Adding portless variants of hosts that are already in the loopback set is just being tolerant of how individual browsers serialize the `Origin` header — it doesn't admit any host that wasn't already accepted with a port. Verified by the `still blocks non-loopback portless Origins (security boundary preserved)` test.

The non-loopback bind host (`OD_BIND_HOST`, e.g. Tailscale or LAN IP under `--host`) is already accepted with a port; adding its portless form is consistent with that policy.

## Diff shape

| File | What changes |
|---|---|
| `apps/daemon/src/server.ts` | `buildAllowedOrigins()` (line 1333) and `isLocalSameOrigin()` (line 5103) each gain 4 portless allowlist entries: `http`/`https` × `127.0.0.1`/`localhost`/`[::1]` plus `http`/`https` × bind host. Comments cite issue #733 for archaeology. |
| `apps/daemon/tests/origin-validation.test.ts` | `createOriginMiddleware()` helper updated to mirror the real middleware. 6 new test cases covering the portless code paths. |

2 files, +78 / −9.

## What's verified

| Check | Result |
|---|---|
| `vitest run tests/origin-validation.test.ts` | **25/25** (was 19, added 6) |
| `vitest run` (full daemon suite) | **923/923** across 63 test files |
| `tsc -p apps/daemon/tsconfig.json --noEmit` | clean |

## What's NOT verified

I don't have a Windows + Chrome 147 setup to exercise the actual portless-Origin browser path end-to-end. The structural argument is:

1. The reporter's repro is concrete (`Origin: http://127.0.0.1` reaching the middleware → 403). Their fix proposal is exactly what this PR lands.
2. The new positive tests exercise that exact `Origin` shape against the same allowlist function the daemon uses at runtime.
3. The new negative test confirms `http://evil.com` (non-loopback, no port) still 403s, so the trust boundary is preserved.

If @JHR-deve can pull this branch and rerun the rescan in Chrome normal mode, that closes the loop end-to-end.

## Test plan

- [x] Existing 19 origin-validation tests still pass.
- [x] 6 new tests covering portless loopback variants.
- [x] Full daemon test suite (923 tests) green.
- [x] Daemon typecheck clean.
- [ ] Manual smoke on Windows + Chrome normal mode: `pnpm tools-dev start web`, click Settings → Rescan, confirm no 403 on `/api/connectors/status`.
- [ ] Same smoke on Firefox / Safari / Chrome incognito to confirm no regression in the previously-working port-bearing path.

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -1336,15 +1336,26 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     if (webPort && webPort !== resolvedPort) ports.push(webPort);
     const schemes = ['http', 'https'];
     const loopbackHosts = ['127.0.0.1', 'localhost', '[::1]'];
-    return new Set(
-      ports.flatMap((p) => [
+    return new Set([
+      ...ports.flatMap((p) => [
         ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
         // When bound to a specific non-loopback address (e.g. Tailscale,
         // LAN IP, or 0.0.0.0), allow browser requests from that address
         // too so the documented --host escape hatch remains usable.
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
-    );
+      // Some browser builds (observed on Chrome on Windows under certain
+      // localhost-with-non-standard-port profiles, issue #733) serialize
+      // the Origin header without the port, e.g. `http://127.0.0.1`
+      // instead of `http://127.0.0.1:6313`. The cross-origin policy is
+      // still loopback-only — port-less means "the same origin's host
+      // without a port" — so accepting these portless variants doesn't
+      // widen the trust boundary, just makes the matcher tolerant of
+      // how individual browsers serialize the header. Without this,
+      // every /api request from such a browser dies with 403.
+      ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
+      ...schemes.map((s) => `${s}://${host}`),
+    ]);
   }
 
   // Routes that serve content to sandboxed iframes (Origin: null) for
@@ -5128,11 +5139,16 @@ export function isLocalSameOrigin(req, port) {
   if (origin == null || origin === '') return true;
 
   const schemes = ['http', 'https'];
-  const allowedOrigins = new Set(
-    ports.flatMap((p) => [
+  const allowedOrigins = new Set([
+    ...ports.flatMap((p) => [
       ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
       ...schemes.map((s) => `${s}://${bindHost}:${p}`),
     ]),
-  );
+    // Match `buildAllowedOrigins()`: accept the portless serializations
+    // some browsers emit on localhost (issue #733). Loopback-only, so
+    // the trust boundary doesn't change.
+    ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
+    ...schemes.map((s) => `${s}://${bindHost}`),
+  ]);
   return allowedOrigins.has(String(origin));
 }
--- a/apps/daemon/tests/origin-validation.test.ts
+++ b/apps/daemon/tests/origin-validation.test.ts
@@ -32,12 +32,16 @@ function createOriginMiddleware(resolvedPort, host = '127.0.0.1') {
     if (webPort && webPort !== resolvedPort) ports.push(webPort);
     const schemes = ['http', 'https'];
     const loopbackHosts = ['127.0.0.1', 'localhost', '[::1]'];
-    const allowedOrigins = new Set(
-      ports.flatMap((p) => [
+    const allowedOrigins = new Set([
+      ...ports.flatMap((p) => [
         ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}:${p}`)),
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
-    );
+      // Mirror buildAllowedOrigins(): accept the portless serializations
+      // some browsers emit on localhost (issue #733).
+      ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
+      ...schemes.map((s) => `${s}://${host}`),
+    ]);
     if (!allowedOrigins.has(String(origin))) {
       return res.status(403).json({ error: 'Cross-origin requests are not allowed' });
     }
@@ -145,6 +149,48 @@ describe('daemon origin validation middleware', () => {
       origin: `https://127.0.0.1:${port}`,
     });
     expect(res.status).toBe(200);
+  });
+
+  // --- Portless Origin (issue #733) ---
+  // Some browser builds (observed on Chrome under Windows when serving
+  // localhost on a non-standard port) send the Origin header without
+  // the port — `http://127.0.0.1` instead of `http://127.0.0.1:6313`.
+  // The middleware must accept these portless loopback variants so
+  // every /api request from such a browser doesn't 403.
+
+  it('allows portless Origin: http://127.0.0.1 (loopback, no port)', async () => {
+    const res = await request(port, 'GET', '/api/projects', {
+      origin: 'http://127.0.0.1',
+    });
+    expect(res.status).toBe(200);
+  });
+
+  it('allows portless Origin: http://localhost (loopback, no port)', async () => {
+    const res = await request(port, 'GET', '/api/projects', {
+      origin: 'http://localhost',
+    });
+    expect(res.status).toBe(200);
+  });
+
+  it('allows portless Origin: http://[::1] (IPv6 loopback, no port)', async () => {
+    const res = await request(port, 'GET', '/api/projects', {
+      origin: 'http://[::1]',
+    });
+    expect(res.status).toBe(200);
+  });
+
+  it('allows portless Origin via HTTPS', async () => {
+    const res = await request(port, 'GET', '/api/projects', {
+      origin: 'https://127.0.0.1',
+    });
+    expect(res.status).toBe(200);
+  });
+
+  it('still blocks non-loopback portless Origins (security boundary preserved)', async () => {
+    const res = await request(port, 'GET', '/api/projects', {
+      origin: 'http://evil.com',
+    });
+    expect(res.status).toBe(403);
   });
 
   // --- Origin: null (sandboxed iframe previews) ---
@@ -320,5 +366,12 @@ describe('origin validation: non-loopback bind host', () => {
       origin: `http://evil.com:${port}`,
     });
     expect(res.status).toBe(403);
+  });
+
+  it('allows portless Origin from the non-loopback bind host (issue #733)', async () => {
+    const res = await request(port, 'GET', '/api/projects', {
+      origin: `http://${nonLoopbackHost}`,
+    });
+    expect(res.status).toBe(200);
   });
 });


apps/daemon/src/server.ts
```diff
@@ -1336,15 +1336,26 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+      // localhost-with-non-standard-port profiles, issue #733) serialize
+      // the Origin header without the port, e.g. `http://127.0.0.1`
+      // instead of `http://127.0.0.1:6313`. The cross-origin policy is
+      // still loopback-only — port-less means "the same origin's host
+      // without a port" — so accepting these portless variants doesn't
+      // widen the trust boundary, just makes the matcher tolerant of
+      // how individual browsers serialize the header. Without this,
+      // every /api request from such a browser dies with 403.
+      ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
+      ...schemes.map((s) => `${s}://${host}`),
```
mrcfps, line 1357: These portless allowlist entries make `http://127.0.0.1`, `http://localhost`, and the explicit bind host trusted for every daemon `/api` request. The problem is that those are also the normal serialized origins for pages actually served from the default ports (`http://localhost:80` or `https://localhost:443`), not only the Chrome/Windows bug case where a non-standard web port was omitted. That means a local service bound to loopback 80/443 can now send browser requests to the daemon and pass CORS/origin validation, while the existing policy just below is meant to reject other local ports unless they are the daemon port or `OD_WEB_PORT` (`blocks cross-origin requests from other local ports`). This is a security boundary expansion for local CSRF/cross-port access.

Please keep the allowlist tied to the known daemon/web ports, or add an additional signal that distinguishes the affected packaged/web client from arbitrary default-port localhost pages before accepting a portless origin. At minimum, add regression coverage that `Origin: http://127.0.0.1`/`http://localhost` from a default-port local page is not treated as an unrestricted replacement for the current port-scoped allowlist.

mrcfps (changes requested):
@Sid-Qin Thanks for landing this with focused coverage and for clearly documenting the Chrome/Windows repro. I found one blocker around the allowlist change: the portless origins also match real default-port localhost pages, which reopens cross-port access that the existing policy and tests otherwise reject.


apps/daemon/src/server.ts
```diff
@@ -1336,15 +1336,26 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+      // Some browser builds (observed on Chrome on Windows under certain
+      // localhost-with-non-standard-port profiles, issue #733) serialize
+      // the Origin header without the port, e.g. `http://127.0.0.1`
+      // instead of `http://127.0.0.1:6313`. The cross-origin policy is
+      // still loopback-only — port-less means "the same origin's host
+      // without a port" — so accepting these portless variants doesn't
+      // widen the trust boundary, just makes the matcher tolerant of
+      // how individual browsers serialize the header. Without this,
+      // every /api request from such a browser dies with 403.
+      ...schemes.flatMap((s) => loopbackHosts.map((h) => `${s}://${h}`)),
```
lefarcen, line 1356: **P1** This allowlist admits default-port localhost pages (`http://127.0.0.1:80`, `http://localhost:80`, `https://127.0.0.1:443`) to the daemon on any resolved port, bypassing the existing cross-port Origin protection. Don't treat portless Origins as matching arbitrary daemon/web ports; keep exact scheme/host/port matching, or require an additional unforgeable local-UI proof such as a daemon-issued capability/CSRF token if this browser bug must be tolerated.

apps/daemon/src/server.ts
```diff
@@ -1336,15 +1336,26 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
         // LAN IP, or 0.0.0.0), allow browser requests from that address
         // too so the documented --host escape hatch remains usable.
         ...schemes.map((s) => `${s}://${host}:${p}`),
       ]),
-    );
+      // Some browser builds (observed on Chrome on Windows under certain
+      // localhost-with-non-standard-port profiles, issue #733) serialize
+      // the Origin header without the port, e.g. `http://127.0.0.1`
+      // instead of `http://127.0.0.1:6313`. The cross-origin policy is
+      // still loopback-only — port-less means "the same origin's host
```
lefarcen, line 1351: **P1** The rationale that "portless means the same origin's host without a port" doesn't hold. In CORS and WHATWG URL semantics, origins are scheme/host/port tuples and default ports are omitted only in serialization; common CORS middleware compares exact serialized origins or parsed tuples rather than treating a missing port as a wildcard. The PR needs to acknowledge that `Origin` alone cannot distinguish the reported Chrome bug from a real default-port localhost page, and should choose a safer alternative.

apps/daemon/src/server.ts
lefarcen, line 1357: **P1** This also adds portless entries for the explicit bind host, including non-loopback hosts, so the change is not loopback-only despite the comment. A default-port page on the configured LAN/Tailscale host would now be accepted for daemon API requests; either remove this entry or constrain any workaround to a separately verified loopback-local UI context.

apps/daemon/tests/origin-validation.test.ts
```diff
@@ -147,6 +151,48 @@ describe('daemon origin validation middleware', () => {
...
   });
 
+  // --- Portless Origin (issue #733) ---
+  // Some browser builds (observed on Chrome under Windows when serving
+  // localhost on a non-standard port) send the Origin header without
+  // the port — `http://127.0.0.1` instead of `http://127.0.0.1:6313`.
+  // The middleware must accept these portless loopback variants so
+  // every /api request from such a browser doesn't 403.
+
+  it('allows portless Origin: http://127.0.0.1 (loopback, no port)', async () => {
```
lefarcen, line 161: **P2** The new tests codify the insecure behavior but don't cover the actual runtime security paths. They duplicate the middleware instead of exercising `startServer()` and they don't cover `isLocalSameOrigin()` with Host validation, default-port localhost negative cases, or non-loopback bind-host default-port negatives. After changing the policy, add tests that prove default-port localhost/non-loopback Origins remain rejected unless a stronger verification mechanism is present.

lefarcen (changes requested):
Hi @Sid-Qin — thanks for the detailed implementation and thorough test coverage. I validated mrcfps's security concern against the code and confirm it's correct: portless `Origin` headers are **not** wildcards; they mean default ports (80 for HTTP, 443 for HTTPS) per WHATWG URL origin serialization. The current approach would allow real default-port localhost pages to bypass the existing cross-port protection.

See inline comments for the specific security boundary issues. In short: `Origin` alone can't distinguish the Chrome Windows bug from a legitimate default-port page, so a different approach is needed — either keep exact port matching + add a daemon-issued capability token for affected browsers, or require a Host-based check that confirms the request actually came from a daemon-served page.

Sid-Qin:
Closing this in favor of #735, which the reporter @JHR-deve opened ~3 hours before mine and which has already been iterated through @mrcfps's security review and approved.

I should have run `gh pr list --search "733"` before opening this — apologies for the duplicate work. More importantly, the security feedback on this PR is correct and lines up with what got #735 to its current shape:

- Per WHATWG URL origin serialization, `http://127.0.0.1` (portless) is **not** a wildcard — it's the canonical Origin for a page on `127.0.0.1:80`. Adding it to the global `buildAllowedOrigins()` set means a page actually served from port 80 (or any service that fakes that header) bypasses the cross-port boundary.
- Worse, this PR also relaxed `isLocalSameOrigin()`, which gates per-route POST/PUT/DELETE handlers. So mutating endpoints would have been opened up too — exactly the wrong direction.

[#735](https://github.com/nexu-io/open-design/pull/735) gets it right: keep `buildAllowedOrigins()` and `isLocalSameOrigin()` strict, add a narrow `isPortlessLoopbackOrigin()` helper, and only fall back to it for safe `GET` requests in the global middleware. Mutating methods stay locked.

The one piece of value here that #735 doesn't currently have is the test coverage on `apps/daemon/tests/origin-validation.test.ts` (4 positive portless cases + 1 negative `http://evil.com` boundary check + 1 portless bind-host). Happy to open a tests-only follow-up against #735's helper after it merges — pinging @mrcfps and @lefarcen for their take on whether that's wanted.

Thanks for the careful, fast review @mrcfps @lefarcen — better to catch this here than in a security report.

Sid-Qin:
Pushed c61d520 addressing the P1/P2 review concerns. Summary of how the new approach answers each thread, then re-request review.

## What changed

The unconditional portless allow-list is gone. Two-tier check now:

1. **Strict scheme+host+port match** (back to `main`'s shape — exact match per resolved port + `OD_WEB_PORT` + bind host).
2. **Portless fallback gated on `Sec-Fetch-Site: same-origin`.** That header is forbidden to JavaScript (`Sec-*` prefix is on Fetch's forbidden-header-name list, so `fetch()`/`XMLHttpRequest` cannot set it), and the browser only emits `same-origin` when the initiator's full origin (scheme+host+port) matches the request target.

Same gate applied to `isLocalSameOrigin()`.

## How this answers each thread

**@mrcfps / @lefarcen / Codex P1 — "default-port localhost pages bypass cross-port protection":** A real `:80`/`:443` localhost page calling cross-port to the daemon is, by browser-side definition, *not* same-origin with the daemon. Chrome therefore emits `Sec-Fetch-Site: same-site` (or `cross-site`), never `same-origin`. The portless allow-list never fires for those requests. Cross-port protection preserved.

**@lefarcen P1 — "Origin alone cannot distinguish the bug from a real default-port page; need an unforgeable local-UI proof":** `Sec-Fetch-Site` is exactly that — browser-set, JS-forbidden, reflects the browser's own scheme+host+port determination independent of the (buggy) Origin serialization. The Chrome-on-Windows #733 case sends `Sec-Fetch-Site: same-origin` because the request really *is* same-origin to the daemon; only the Origin header drops the port. Attacker pages cannot fake this. Capability/CSRF tokens were the other option suggested — `Sec-Fetch-Site` achieves the same property without the bootstrap problem.

**@lefarcen P1 — "portless entries for non-loopback bind host widen the trust boundary":** Same gate applies to the bind-host portless entry. A default-port LAN/Tailscale page calling cross-port is `same-site`, not `same-origin`, so it stays blocked. The portless bind-host entry only fires for the same Chrome bug scenario when the user accessed the daemon directly at the bind host.

**@lefarcen P2 — "tests duplicate the middleware and don't cover real runtime paths or default-port negatives":**
- New `describe` block exercises the exported `isLocalSameOrigin()` directly, including its `Host` validation (DNS rebinding guard) and the new portless-with-Sec-Fetch-Site logic — no middleware mirror.
- Default-port localhost negatives added: portless `http://localhost` without `Sec-Fetch-Site`, with `same-site`, with `cross-site`; portless `https://localhost` (`:443` page); portless POST to state-changing endpoints; non-loopback bind host portless without/with `same-site`; non-loopback portless even with `same-origin` still rejected (forbidden-header spoof can't admit a non-loopback origin).

The mirrored middleware in `createOriginMiddleware()` is updated to match the new logic exactly so the integration tests still exercise the same shape.

## Verification

| Check | Result |
|---|---|
| `vitest run tests/origin-validation.test.ts` | **42/42** (was 25; +17 negative + helper-direct tests) |
| `vitest run` (full daemon suite) | **940/940** across 63 files (was 923) |
| `tsc -p apps/daemon/tsconfig.json --noEmit` | clean |

End-to-end on Windows + Chrome 147 still relies on @JHR-deve's smoke (the structural argument: their reported bug is browser-side same-origin with a portless Origin serialization → `Sec-Fetch-Site: same-origin` is sent → portless allow-list fires → request passes).

lefarcen:
@Sid-Qin I see the c61d520 update — but since this PR is closed in favor of #735, no re-review is needed here. If you meant to push that commit to #735, you can mention it there. Thanks for the detailed write-up!

PR #763, opened 2026-05-07
feat(daemon): close pi adapter parity gaps — imagePaths, extraAllowedDirs, error events, sendAgentEvent routing

monotykamary:
## Summary

The pi adapter was missing capabilities that every other recently-updated adapter (Qoder, Copilot, OpenCode) now supports. This PR closes four parity gaps so pi runs get the same error surfacing, file-access, and multimodal capabilities as the rest of the adapter catalog.

## Why

Recent commits added Qoder CLI support, connection tests, Copilot stdin delivery, and OpenCode error surfacing (#691). The pi adapter:

1. **Discarded `imagePaths`** — other adapters forward images via `--attachment` or inline args; pi's RPC protocol natively supports `images` on the `prompt` command but the daemon never sent them.
2. **Discarded `extraAllowedDirs`** — other adapters use `--add-dir` to widen the agent's filesystem view so it can read skill seeds and design-system specs outside the project cwd; pi had no equivalent.
3. **Ignored `extension_error` and `auto_retry_end` failure events** — these RPC events were silently dropped by `mapPiRpcEvent`, so extension errors and exhausted retries would never surface to the user.
4. **Routed events through `send` instead of `sendAgentEvent`** — this bypassed the `agentStreamError` flag and the `trackingSubstantiveOutput` / `agentProducedOutput` guard added for issue #691. A pi run that exited 0 without producing visible content was marked `succeeded` with an empty chat — the exact silent-failure shape #691 fixed for OpenCode.

## Changes

| File | What changes |
|---|---|
| `apps/daemon/src/agents.ts` | `buildArgs` now accepts `extraAllowedDirs` (was `_extra`), filters to absolute paths, and forwards each as `--append-system-prompt`. Adds `supportsImagePaths: true` flag. |
| `apps/daemon/src/pi-rpc.ts` | `attachPiRpcSession` accepts `imagePaths`, reads files into base64, and includes `images` on the RPC `prompt` command. `mapPiRpcEvent` handles `extension_error` → `error` event and `auto_retry_end` with `success: false` → `error` event. |
| `apps/daemon/src/server.ts` | pi-rpc branch sets `trackingSubstantiveOutput = true` and passes `sendAgentEvent` instead of `send`, forwarding `safeImages` when `supportsImagePaths` is set. |
| `apps/daemon/src/connectionTest.ts` | Passes `imagePaths: []` to `attachPiRpcSession` for forward-compatibility. |
| `docs/agent-adapters.md` | Adds pi to the adapter catalog table and §5.11 per-adapter section. |

## Testing

| Check | Result |
|---|---|
| `vitest run tests/pi-rpc.test.ts` | **37/37** (28 existing + 9 new) |
| `vitest run tests/agents.test.ts` (pi subset) | **4/4** pi tests pass (1 updated + 3 new) |
| `tsc -p apps/daemon/tsconfig.json --noEmit` | No errors in changed files |

### New tests

**`tests/pi-rpc.test.ts`** (9 new):
- `extension_error` maps to error event (3 variants: string, non-string, missing)
- `auto_retry_end` with `success=false` maps to error (3 variants: with message, success=true, missing finalError)
- `attachPiRpcSession` forwards images when `imagePaths` provided
- `attachPiRpcSession` omits images when `imagePaths` is empty
- `attachPiRpcSession` skips unreadable image paths gracefully

**`tests/agents.test.ts`** (3 new + 1 assertion):
- `pi.supportsImagePaths === true`
- pi args forward `extraAllowedDirs` as `--append-system-prompt` flags
- pi args filter relative paths from `extraAllowedDirs`
- pi args combine model, thinking, and `extraAllowedDirs`

## What is NOT changed

- No changes to the pi binary or its RPC protocol — all changes are in the daemon adapter layer
- Other adapters are unaffected
- No web UI changes
- No contracts package changes


apps/daemon/src/server.ts
```diff
@@ -4217,12 +4217,21 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+      // substantive-output guard (agentProducedOutput) so a pi run
+      // that exits 0 without producing visible content is caught.
+      trackingSubstantiveOutput = true;
       acpSession = attachPiRpcSession({
         child,
         prompt: composed,
         cwd: effectiveCwd,
         model: safeModel,
-        send,
+        send: sendAgentEvent,
```
mrcfps: This passes `sendAgentEvent` directly into `attachPiRpcSession`, but `attachPiRpcSession` still invokes its callback with the channel/payload shape: `send('agent', payload)` for normal events and `send('error', { message })` from `fail()` in `apps/daemon/src/pi-rpc.ts`. `sendAgentEvent` expects a single event object, so the first argument becomes the string `'agent'`/`'error'` and the real payload is discarded. That means pi text deltas/status updates are emitted as malformed `agent` SSE payloads, `agentProducedOutput` is never set, and the new `extension_error` / `auto_retry_end` errors won't set `agentStreamError` as intended. In the main pi-rpc path this can turn successful output into the empty-output failure path and can hide the errors this PR is trying to surface.

Please adapt the callback at this call site (or change `attachPiRpcSession`'s callback contract consistently) so `agent` channel payloads are forwarded to `sendAgentEvent(payload)` and fatal `error` channel payloads still go through `send('error', createSseErrorPayload(...))` / equivalent existing error handling. A regression test should cover the server-side pi-rpc adapter callback shape, not only `mapPiRpcEvent`, so this integration mismatch is caught.

mrcfps (changes requested):
@monotykamary thanks for closing the pi adapter parity gaps here — the image forwarding and error-event coverage are useful additions. I found one integration issue in the server wiring that currently breaks pi-rpc event delivery and needs a fix before merge.


apps/daemon/src/server.ts
```diff
@@ -4217,12 +4217,21 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+      // (extension_error, auto_retry_end with success=false, and the
+      // message_update error delta) set agentStreamError and flip the
+      // run to `failed` on close — same path as qoder-stream-json and
+      // json-event-stream after issue #691. Also enables the
+      // substantive-output guard (agentProducedOutput) so a pi run
+      // that exits 0 without producing visible content is caught.
+      trackingSubstantiveOutput = true;
       acpSession = attachPiRpcSession({
         child,
         prompt: composed,
```
lefarcen, line 4238: **P1** `attachPiRpcSession` calls `send(channel, payload)`, but `sendAgentEvent` accepts one agent-event payload. Passing it directly turns `send('agent', {type:'text_delta'})` into `sendAgentEvent('agent')`, so pi emits malformed agent SSE payloads and error frames never set `agentStreamError`.

→ Wrap it as `(channel, payload) => channel === 'agent' ? sendAgentEvent(payload) : send(channel, payload)` and add an integration/unit test for this server wiring.

apps/daemon/src/pi-rpc.ts
```diff
@@ -184,6 +186,19 @@ export function mapPiRpcEvent(raw, send, ctx) {
     return null;
   }
 
+  // pi's RPC protocol can emit `extension_error` when an extension
+  // throws during a tool call or event handler. Surface it so the
+  // daemon's error-handling path (sendAgentEvent → agentStreamError)
+  // can flip the run to `failed` and forward a visible SSE error.
+  if (raw.type === 'extension_error') {
```
lefarcen, line 213: **P2** `mapPiRpcEvent` only handles text/thinking in `message_update`, but the server comment claims `message_update` error deltas are part of the routed failure path. If pi emits an error after any substantive output, the run can still finish as succeeded because `agentProducedOutput` bypasses the empty-output guard.

→ Map the pi error delta shape to `{type:'error', message, raw}` and test the after-partial-output case.

apps/daemon/src/pi-rpc.ts
```diff
@@ -265,7 +293,37 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
     stdinOpen = false;
   });
 
-  promptRpcId = sendCommand(child.stdin, 'prompt', { message: prompt });
+  // Build the images array for pi's prompt command. pi's RPC protocol
+  // accepts `images` as an array of {type, data, mimeType} objects where
+  // `data` is base64-encoded file contents. The daemon's safeImages guard
+  // already validated that each path exists under UPLOAD_DIR.
+  const images = [];
```
lefarcen, line 326: **P2** Image forwarding reads whatever path survives the prefix check, following symlinks and without file/type/total-size validation. A symlink under `UPLOAD_DIR` or repeated large image paths can cause unintended local file disclosure to the model or large synchronous base64 work.

→ Validate `realpath` stays under `UPLOAD_DIR`, require `stat.isFile()`, require known image extensions/MIME types, and enforce a count/total-byte budget before `readFileSync`.

docs/agent-adapters.md
```diff
@@ -208,6 +209,17 @@ The adapter declares which strategy to use via `capabilities().nativeSkillLoadin
 - Permission: `--permission-mode bypass_permissions` avoids headless approval prompts in the web UI. Users should treat this as the same trust posture as running Qoder directly with that flag in the selected project directory.
 - **Gotcha:** Detection only proves `qodercli --version` can run. Qoder authentication and account scope remain owned by Qoder CLI, with credentials read from Qoder's `~/.qoder/config.json`; the daemon surfaces stderr/stdout failures from the spawned run instead of running login or editing Qoder config.
 
+### 5.11 Pi
+
```
lefarcen, line 213: **P2** The docs describe `--append-system-prompt` as the access-widening mechanism, but it only hints paths — it does not grant sandbox/file access. This leaves an unstated assumption that pi's Read tool can open absolute paths outside cwd.

→ Document that distinction, why no stronger pi flag is available, and that staged cwd copies are the reliable access path when absolute reads fail.

lefarcen (changes requested):
Hi @monotykamary! Thanks for tackling these parity gaps — closing the pi adapter's missing capabilities is definitely the right direction. I've run a deep review on both the implementation and the reasoning.

Found 4 substantive issues that need attention before this can land safely — 2 in the server wiring logic that affect whether error tracking actually works, 1 in the image-forwarding path, and 1 documentation gap about access assumptions. See inline comments below for each.

lefarcen (changes requested):
The P1 server wiring issue from mrcfps's review (and my 07:12 review) is still present on this head.

**The problem:** `server.ts` passes `sendAgentEvent` into `attachPiRpcSession`, but `attachPiRpcSession` invokes its callback with the two-arg channel/payload shape: `send('agent', payload)` for normal events and `send('error', { message })` from `fail()`. `sendAgentEvent` expects a single event object, so the first argument becomes the string `'agent'`/`'error'` and the real payload is discarded. That means pi text deltas/status updates are emitted as malformed SSE, `agentProducedOutput` is never set, and the new error events won't set `agentStreamError`.

**The fix:** Adapt the callback at the call site or change `attachPiRpcSession`'s callback contract consistently. Either:
1. Wrap `sendAgentEvent` so it translates `send('agent', payload) → sendAgentEvent(payload)` and `send('error', { message })` → existing error path, OR
2. Change `attachPiRpcSession` to invoke `send(payload)` directly (single-arg) and handle errors outside.

Can't approve until this integration is corrected — it breaks the error-surfacing path this PR is trying to add.

lefarcen:
Hi @monotykamary! 🎉
Thanks for the contribution — closing the pi adapter parity gaps is the right direction.

— open-design team

apps/daemon/src/server.ts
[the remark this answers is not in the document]
monotykamary: Fixed in eab3633. The `send` callback is now wrapped at the server.ts call site: `agent` channel payloads route through `sendAgentEvent(payload)`, `error` channel payloads go through `createSseErrorPayload` + `agentStreamError`. This preserves the two-arg channel/payload contract that `attachPiRpcSession` uses while correctly routing events into the daemon's error-tracking and substantive-output paths.

monotykamary: Good catch — fixed in eab3633. The callback is now adapted at the call site rather than passing `sendAgentEvent` directly: `(channel, payload) => { if (channel === 'agent') sendAgentEvent(payload); else if (channel === 'error') { ... createSseErrorPayload + agentStreamError ... } }`. This ensures text deltas/status updates pass through correctly, `agentProducedOutput` is set, and error frames set `agentStreamError`.

apps/daemon/src/server.ts
monotykamary: Fixed in eab3633 per your suggestion. The callback is wrapped: `agent` channel → `sendAgentEvent(payload)`, `error` channel → `createSseErrorPayload` + `agentStreamError`. 4 new tests cover the message_update error delta path (including the after-partial-output case), and the existing pi-rpc integration tests confirm events route through correctly with the bridged callback shape.

apps/daemon/src/pi-rpc.ts
monotykamary: Fixed in eab3633. `mapPiRpcEvent` now handles `message_update` with `assistantMessageEvent.type === 'error'` — it maps to `{type: 'error', message, raw}` using `reason` → `delta` → fallback chain. Added 4 tests: error with reason, error with delta fallback, error with no info, and the after-partial-output case (text_delta followed by error delta — both events emit, so the run flips to failed rather than succeeding with partial content).

apps/daemon/src/pi-rpc.ts
monotykamary: Fixed in eab3633. Image forwarding now: (1) resolves symlinks via `fs.realpathSync` and re-checks `stat.isFile()`, (2) requires known image extensions (`.png`, `.jpg`, `.jpeg`, `.gif`, `.webp` — no `.txt` or arbitrary files), (3) enforces a count budget (max 10 images) and total-byte budget (20 MB) before any `readFileSync`. Added 2 tests: directories are rejected, and `.txt` files are rejected.

docs/agent-adapters.md
monotykamary: Fixed in eab3633. Updated the §5.11 Pi Skills bullet to explicitly state that `--append-system-prompt` is only a hint and does not grant sandbox or filesystem access, that pi's Read tool can normally open absolute paths outside cwd, and that the reliable fallback when absolute reads fail is staging copies into the project cwd. Also noted that no stronger pi flag exists today.

lefarcen (approved):
All findings resolved in eab3633 ✅

The P1 server wiring callback mismatch is fixed — `sendAgentEvent` now correctly receives single-arg event payloads via the channel/payload adapter wrapper. The error delta mapping, image hardening, and docs clarification are also in place. Nice work addressing the full review feedback!

CI is still running; once green this is good to merge from my side.

apps/daemon/src/pi-rpc.ts
```diff
@@ -265,7 +313,55 @@ export function attachPiRpcSession({ child, prompt, cwd, model, send }) {
...
+  // synchronous base64 reads from blocking the event loop.
+  const images = [];
+  if (Array.isArray(imagePaths) && imagePaths.length > 0) {
+    let totalBytes = 0;
+    for (const imgPath of imagePaths) {
+      if (images.length >= MAX_IMAGE_COUNT) break;
+      if (typeof imgPath !== 'string' || !imgPath.length) continue;
+      try {
+        // Resolve symlinks and verify it's a regular file.
+        const realPath = fs.realpathSync(imgPath);
```
mrcfps, line 334: The new image hardening resolves symlinks with `fs.realpathSync(imgPath)`, but after that it only checks `stat.isFile()` and the target extension. That means a path that passed `server.ts`'s `safeImages` prefix check as something under `UPLOAD_DIR` can still resolve to an image file outside `UPLOAD_DIR` and be read/base64-forwarded to pi. The comment above this block says realpath prevents symlink escape, but the changed lines do not verify the resolved path remains inside the upload root.

Please keep the upload-root boundary after resolving symlinks, for example by passing the resolved upload root into this helper (or doing the realpath validation before calling it) and requiring `realPath === uploadRoot || realPath.startsWith(uploadRoot + path.sep)` before the stat/read. A small fixture test with an upload-dir symlink to an outside `.jpg` would lock this down.

mrcfps (commented):
@monotykamary thanks for the quick follow-up on the pi adapter parity fixes — the callback bridge, error-event mapping, and docs updates address the main flow well. I found one remaining image-boundary hardening gap that should be straightforward to tighten. 👍


lefarcen (dismissed):
All prior findings from my 07:12 review are now addressed in eab3633. The server wiring callback is correctly adapted, error delta mapping is complete, image forwarding is hardened with symlink checks and budgets, and docs clarify the --append-system-prompt behavior.

LGTM 🎉

apps/daemon/src/pi-rpc.ts
monotykamary: Fixed in 99a8c1b. Added `uploadRoot` param to `attachPiRpcSession` and after `realpathSync` the resolved path is checked against `realpathSync(uploadRoot)` to block symlink escape. Both sides are realpath'd to handle the macOS `/var` → `/private/var` situation.

Server.ts passes `UPLOAD_DIR` as `uploadRoot`. Added 2 fixture tests:
- Symlink inside uploadRoot pointing to a real file also under uploadRoot → forwarded ✓
- Symlink inside uploadRoot pointing to a file outside uploadRoot → rejected ✓

apps/daemon/src/connectionTest.ts
```diff
@@ -864,6 +864,7 @@ function attachAgentStreamHandlers(
       cwd,
       model: model ?? null,
       send,
+      imagePaths: [],
```
mrcfps, line 867: This call site now passes the new `imagePaths` option but not the `uploadRoot` option that was added to `attachPiRpcSession`. Because `pi-rpc.ts` is type-checked from TypeScript consumers, CI is currently failing `pnpm --filter @open-design/daemon build` with `Property 'uploadRoot' is missing` at this call. That blocks the daemon package build even though the normal server path was updated.

Please either make `uploadRoot` truly optional in the exported helper's checked signature, or pass an explicit value here (for the connection-test path, `uploadRoot: undefined` or another safe no-image root is sufficient since `imagePaths` is empty). This keeps the pi connection-test wiring compiling alongside the server-side image forwarding changes.

mrcfps (changes requested):
@monotykamary thanks for the quick follow-up on the pi adapter parity work — the runtime fixes and image-boundary hardening are moving in the right direction. I found one remaining compile-time integration issue from the new `uploadRoot` parameter that needs to be addressed before this can merge. 👍


apps/daemon/src/connectionTest.ts
lefarcen, line 867: **P1** Missing required parameter. `attachPiRpcSession` now requires `uploadRoot` (added in commit 99a8c1b, line 267 of `pi-rpc.ts`). TypeScript will fail to compile this call site with "Property 'uploadRoot' is missing".

Add `uploadRoot: undefined` here (or pass a real directory if connection tests should validate image forwarding). Since `imagePaths: []`, the uploadRoot boundary check won't run anyway.

lefarcen (changes requested):
The compile-time integration issue from mrcfps's 08:57 review is still present on this head.

**The problem:** `apps/daemon/src/connectionTest.ts` line 867 calls `attachPiRpcSession` with `imagePaths: []` but without `uploadRoot`. Since commit 99a8c1b added `uploadRoot` as a required parameter to the function signature (line 267 of `apps/daemon/src/pi-rpc.ts`), TypeScript will error with "Property 'uploadRoot' is missing" when building `@open-design/daemon`.

**The fix:** Add `uploadRoot: undefined` (or another safe value) to the `attachPiRpcSession` call in `connectionTest.ts`. Since `imagePaths` is empty in the connection-test path, `uploadRoot` doesn't need a real directory — `undefined` is sufficient and the image-forwarding code will skip it.

lefarcen:
@monotykamary The connectionTest.ts call site (line 867) still needs the `uploadRoot` parameter — the current head `99a8c1b1` doesn't include it in the `attachPiRpcSession` call. That's causing the CI build failure (`Property 'uploadRoot' is missing`).

Could you add `uploadRoot: undefined` (or another safe value) to the connectionTest.ts call? Since `imagePaths` is empty there, `uploadRoot` doesn't need a real directory.

lefarcen:
@monotykamary I checked the current head (99a8c1b1) and the compile-time integration issue mrcfps pointed out still needs fixing.

The call site at `apps/daemon/src/connectionTest.ts` line 867 currently looks like:

```typescript
attachPiRpcSession({
  child,
  prompt,
  cwd,
  model: model ?? null,
  send,
  imagePaths: []
})
```

But `attachPiRpcSession` now requires `uploadRoot` in its signature (apps/daemon/src/pi-rpc.ts line 268). Please add `uploadRoot` to this call — for the connection-test path, passing `uploadRoot: undefined` or a safe no-image root is fine since `imagePaths` is empty.

This will unblock the daemon package build that's currently failing CI.

apps/daemon/src/connectionTest.ts
monotykamary: Fixed in ff24a81. Added `uploadRoot: undefined` at this call site. Since `imagePaths` is empty the boundary check is a no-op, but supplying the param keeps the TypeScript signature satisfied and CI green.

monotykamary: Fixed in ff24a81. Added `uploadRoot: undefined` here — imagePaths is empty so the boundary check is inert, but the explicit param satisfies the TS2345 check and unblocks CI.

mrcfps (approved):
@monotykamary I reviewed the current `ff24a81` head across the pi adapter arg wiring, pi-rpc event and image handling, server routing, connection-test call site, tests, and adapter docs. The prior callback bridge, image-boundary, and `uploadRoot` call-site concerns are addressed, and I don't see any remaining actionable implementation issues. Thanks for the careful follow-through on these parity fixes — this is a solid improvement for pi users. 🙏


lefarcen (approved):
Compile-time integration issue resolved in ff24a81 ✅

The `uploadRoot: undefined` parameter is now correctly passed to `attachPiRpcSession` in connectionTest.ts. This unblocks the daemon package build and satisfies the function signature requirement mrcfps and I flagged.

All prior findings from my reviews are now addressed. CI is green. LGTM 🎉

lefarcen (approved):
Fix verified at ff24a81 ✅

The uploadRoot parameter is now correctly passed in connectionTest.ts line 874. CI is green and the compile-time integration issue is resolved.

leilei926524-tech:
Hi @monotykamary — quick note from the Open Design team.

You've already shipped **5 merged PRs** here, which puts you in our top contributors. We've spun up a Discord **#contributors** channel and wanted to make sure you have the invite — you're exactly the kind of person it's built for:

What you get there:

PR #767, opened 2026-05-07
fix(connectors): show stable curated tool count in connector card badge (#748)

### LINKED ISSUE #748: Fix inconsistent GitHub connector tool counts before and after connection
## Summary

The GitHub connector shows inconsistent tool counts before and after connection.

Before connecting, the GitHub card shows a small tool count such as `2 tools`, but after connecting it shows a dramatically different number such as `868 tools`. This makes the connector metadata feel unreliable and confusing.

## Current behavior

- On the connector list before connection, GitHub shows a small tool count.
- After the connector is connected, the GitHub card shows a much larger tool count.
- The change is not explained in the UI.
- Users cannot tell whether:
  - the pre-connect count was wrong
  - the post-connect count is wrong
  - the two counts represent different concepts

## Expected behavior

The GitHub connector should present a consistent and understandable tool count before and after connection.

Suggested improvements:

- Use one clear definition of what the displayed tool count represents.
- Keep the pre-connect and post-connect counts consistent when they refer to the same thing.
- If connected state exposes a different metric, label it explicitly instead of reusing the same count-style badge.
- Avoid dramatic unexplained jumps in visible tool numbers.

## Steps to reproduce

1. Open the Connectors page.
2. Search for GitHub if needed.
3. Observe the GitHub card before connection and note the displayed tool count.
4. Connect the GitHub connector.
5. Return to the connector list and observe the new displayed tool count.
6. Notice that the number is significantly different from the original one.

## Notes

- the GitHub connector before connection with a small tool count
- the GitHub connector after connection with a much larger tool count

## Why this matters

Connector metadata should be trustworthy. If the tool count changes drastically without explanation, users may assume the connector state is broken or that the product is surfacing inconsistent backend data.

Image

Image

Sid-Qin:
## Summary

- The connector card's "N tools" badge in [`apps/web/src/components/EntryView.tsx:936`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/EntryView.tsx#L936) and [`:1089`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/EntryView.tsx#L1089) rendered `connector.tools.length` for both pre- and post-Composio-hydration states. Same field name, two different concepts:
  - **Pre-hydration** (no Composio API key): `connector.tools` is the static fallback catalog at [`composio.ts:21-53`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/connectors/composio.ts#L21) (≈2 hardcoded GitHub tools)
  - **Post-hydration** (API key configured): the daemon merges in the full provider inventory at [`composio.ts:725-778`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/connectors/composio.ts#L725), so `tools` swells to ≈868 GitHub tools
  - Result: the badge silently jumped from "2 tools" to "868 tools" on hydration with no explanation, exactly as #748 reports
- The agent-callable subset is `definition.allowedToolNames` (read-only auto-approval tools that pass [`isAgentPreviewListableTool()`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/tools/connectors.ts#L40)). That's the count users actually care about — "how many tools does this connector give my agent". This PR exposes that field on the wire `ConnectorDetail` and switches the badge to read it.
- The detail drawer below the card still iterates `connector.tools` to enumerate the full inventory — the count and the list are intentionally different surfaces. Just the **summary count** is anchored to the curated subset.

Fixes #748. Builds on @lefarcen's [diagnosis](https://github.com/nexu-io/open-design/issues/748#issuecomment-4394269031); landing essentially their option 1 ("always show curated/allowed tools count").

## Why `allowedToolNames` is the right anchor

| State | `connector.tools.length` | `connector.allowedToolNames.length` | Badge today | Badge in this PR |
|---|---|---|---|---|
| GitHub, no API key | 2 | 2 | 2 tools | 2 tools |
| GitHub, post-hydration | ≈868 | ≈2 + auto-allowed read tools | **868 tools** | ≈small two-digit count |
| Connector with no static catalog (e.g. Slack pre-key) | 0 | 0 | No tools | No tools |

`allowedToolNames` includes the static curation **plus** any provider-discovered tools whose classified safety is `read + auto` (see [`composio.ts:758-761`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/connectors/composio.ts#L758)). That set can grow on hydration, but only with execution-safe read tools — it never sweeps in the hundreds of write/destructive provider tools that bloat the raw `tools` count.

## Diff shape

| File | What changes |
|---|---|
| `packages/contracts/src/api/connectors.ts` | Add `allowedToolNames: string[]` to `ConnectorDetail` (with a doc comment citing #748). |
| `apps/daemon/src/connectors/catalog.ts` | Mirror the field on the daemon-internal `ConnectorDetail`; populate it in `connectorDefinitionToDetail()` as a defensive copy of `definition.allowedToolNames`. |
| `apps/daemon/src/connectors/service.ts` | No change — `toDetail()` already spreads from `connectorDefinitionToDetail()` so the new field flows through automatically. |
| `packages/contracts/src/examples.ts` | Extend `exampleConnectorDetail`. |
| `apps/web/src/components/EntryView.tsx` | Both badge call sites: `const toolCount = connector.allowedToolNames?.length ?? connector.tools.length;`. The `??` fallback keeps the badge alive against any pre-deploy daemon that hasn't shipped the field yet. |
| `apps/daemon/tests/connectors-service.test.ts` | 4 new tests pin the contract. |
| `apps/web/tests/components/EntryView.test.ts` | Existing fixtures extended with `allowedToolNames: []` for the new required field. |

6 files, +134 / −7.

## What's verified

| Check | Result |
|---|---|
| `apps/daemon` vitest | **925/925** (was 921, added 4) |
| `apps/web` vitest | **332/332** (existing 5 EntryView tests still pass after fixture update) |
| Daemon typecheck (`tsc -p apps/daemon/tsconfig.json --noEmit`) | clean |
| Web typecheck (`tsc -b --noEmit` in `apps/web`) | clean |
| Contracts typecheck | clean |
| Live `pnpm tools-dev run web`: `GET /api/connectors/discovery` | returns `allowedToolNames: ['github.github_search_repositories', 'github.github_get_issue']` for GitHub (length 2) |
| Live UI: connector card badge for GitHub / Notion / Google Drive | reads "2 tools" (no regression vs main, where the same pre-hydration state shows the same count) |
| Live JS bundle | `allowedToolNames` symbol present in 1+ chunks (badge code path is in the build) |

## What's NOT verified

The actual 2 → 868 jump only manifests with a real Composio API key configured. I don't have one to plug in, so the post-hydration browser path can't be exercised end-to-end from here. The proof is structural:

1. The new test `keeps allowedToolNames small even when tools.length holds the full provider inventory (the #748 regression guard)` constructs a connector with `tools.length = 801` and asserts `detail.allowedToolNames.length = 1`. That's the badge invariant @shangxinyu1's report relies on.
2. The badge call sites in `EntryView.tsx` now read `allowedToolNames?.length`, so the badge is mathematically tied to the test's `allowedToolNames.length`.
3. The wire change is verified live: GitHub's discovery response carries the new field with the expected curated names.

If a maintainer with a Composio API key (or @shangxinyu1) can rerun the connect → check badge flow, that closes the loop end-to-end.

## Test plan

- [x] Existing daemon test suite stays green.
- [x] 4 new tests pin: getConnector emits `allowedToolNames`; result is a defensive copy; #748 regression guard with simulated 800-tool inventory; empty-allowlist case renders a real "0 tools" badge.
- [x] Web typecheck + existing EntryView tests pass with updated fixtures.
- [x] Live wire format: `/api/connectors/discovery` includes the new field.
- [x] Live UI: pre-hydration badge reads "2 tools" for static-catalog Composio connectors (no regression).
- [ ] Manual smoke with a real Composio API key + GitHub connect: badge stays at the curated count instead of jumping to ≈868.
- [ ] Same smoke on Notion / Google Drive / Slack to confirm no regression elsewhere.

## Notes

- Older daemon + new web builds: the `??` fallback in `EntryView.tsx` makes the badge degrade gracefully to `tools.length` if a deploy lands the web change before the daemon has the new field. So this PR is safe to ship in either order.
- New daemon + older web builds: the extra wire field is unused — no harm.


apps/daemon/tests/connectors-service.test.ts
```diff
@@ -466,3 +466,93 @@ describe('connector execution policy', () => {
...
+    // Spot-check that none of the bulk write tools accidentally leaked
+    // into the allowlist.
+    expect(detail.allowedToolNames).not.toContain('external_docs.bulk_op_0');
+    expect(detail.allowedToolNames).not.toContain('external_docs.bulk_op_799');
+
+    // refreshEligible classification stays a property of the definition,
+    // not of the badge surface — confirm the helper still agrees so a
+    // future refactor can't quietly let write tools into the allowlist
+    // via a different code path.
+    expect(isRefreshEligibleConnectorToolSafety(provisionedTools[0].safety)).toBe(false);
```
mrcfps: This assertion currently breaks `@open-design/daemon` test typechecking: CI reports `tests/connectors-service.test.ts(546,49): error TS2532: Object is possibly 'undefined'` because `provisionedTools[0]` is indexed under the repo's strict/no-unchecked indexed-access settings. That keeps the workspace validation check red for this PR even though the runtime change is small. Please bind the first element through a non-empty construction or an explicit guard/non-null assertion before reading `.safety`, e.g. `const firstProvisionedTool = provisionedTools[0]; expect(firstProvisionedTool).toBeDefined(); expect(isRefreshEligibleConnectorToolSafety(firstProvisionedTool!.safety)).toBe(false);`, or otherwise avoid direct unchecked indexing.

apps/web/src/components/EntryView.tsx
```diff
@@ -1086,7 +1093,11 @@ function ConnectorDetailDrawer({
   const canConnect = !disabled && !isPending && connector.status === 'available';
   const canDisconnect = !disabled && !isPending && isConnected;
   const accountLabel = getDisplayableConnectorAccountLabel(connector);
-  const toolCount = connector.tools.length;
+  // Mirror the connector card: badge count tracks the curated allowlist
+  // (issue #748). The drawer below still iterates over connector.tools to
+  // show the full inventory — the count and the list are intentionally
+  // different surfaces.
+  const toolCount = connector.allowedToolNames?.length ?? connector.tools.length;
   const isLoadingTools = !toolsLoaded || (toolsLoading && toolCount === 0);
```
mrcfps: The drawer now reuses the curated `toolCount` for `isLoadingTools` and for the empty-state branch below (`toolCount === 0`), but this same drawer still maps over `connector.tools` to show the full provider inventory. When a hydrated connector has raw provider tools but zero allowed agent-callable tools (for example a write-only inventory), `allowedToolNames.length` is 0 while `connector.tools.length` is non-zero, so the drawer renders “no tools available” and suppresses the actual list. This contradicts the PR's intended split where only the badge count is curated and the drawer inventory remains full. Please keep the badge/header count based on `allowedToolNames`, but introduce a separate `inventoryToolCount = connector.tools.length` (or similar) for loading/empty-state decisions and the section count if that section is meant to describe the list being rendered.

mrcfps (changes requested):
@Sid-Qin Thanks for tackling the connector badge instability and for adding contract coverage around the new field. I found two issues that need another pass before this is ready: the current head fails daemon test typecheck, and the drawer now uses the curated count to decide whether to render the full inventory. 🙏


lefarcen (commented):
Thanks for tackling #748 @Sid-Qin! The direction is solid — moving the badge count off raw `tools.length` is the right fix. Found two edge cases that aren't covered yet:

## 1. Read-only discovered tools can still inflate the badge (P2)

`apps/daemon/src/connectors/composio.ts:758-761` appends any live tool with `refreshEligible=true` (i.e., `sideEffect='read' AND approval='auto'`) to `allowedToolNames`. If GitHub has 50 read-only tools after discovery, the badge goes from 2 to 52, which is better than 868 but not stable.

The test you added at `apps/daemon/tests/connectors-service.test.ts:499` only simulates write-type tools (`refreshEligible: false`), so the read-path inflation isn't caught.

**Concrete fix**: Keep `allowedToolNames` for execution policy but add a separate `badgeToolCount` or `curatedToolNames` field that ONLY includes the static catalog names, never discovered tools. Update the web badge to read that field. The execution layer (`apps/daemon/src/tools/connectors.ts`) can still use the dynamic `allowedToolNames` for runtime gating.

## 2. Drawer empty-state bug when `allowedToolNames` is empty (P2)

At `apps/web/src/components/EntryView.tsx:1100`, `toolCount` uses `allowedToolNames?.length`, then at the render conditional around line 1215, it branches on `toolCount === 0` to show "no tools available", but the list maps over `connector.tools`.

If a connector has `allowedToolNames: []` (e.g., a provider with no auto-approved read tools) but `tools.length > 0`, the drawer shows the empty state message even though `connector.tools` has items.

**Concrete fix**: Split into two variables:
```tsx
const badgeToolCount = connector.allowedToolNames?.length ?? connector.tools.length;
const inventoryToolCount = connector.tools.length;
```
Use `badgeToolCount` for the badge, `inventoryToolCount` for the drawer's count/empty/loading/list branching.

## 3. Test coverage gaps (P3)

- `apps/daemon/tests/connectors-service.test.ts`: Add a variant with a mix of read + write discovered tools to pin whether `allowedToolNames` should accept read-type growth or stay purely static.
- `apps/web/tests/components/EntryView.test.ts`: No test renders the connector card or drawer to verify the #748 badge fix or the drawer empty-state edge case. Add render tests with `tools.length > allowedToolNames.length` and `allowedToolNames: []` + non-empty `tools`.

---

The structural fix (exposing `allowedToolNames` on the wire) is good. Just need to tighten the definition so it's truly stable, and fix the drawer rendering logic. Let me know if you'd like to iterate on this or prefer I open a follow-up PR for the two P2s.

Sid-Qin:
@mrcfps both findings addressed in `447a270` — thanks for catching them, especially the drawer regression. My PR description claimed "the count and the list are intentionally different surfaces," but the code didn't actually do that. Sloppy on my part.

**Drawer split (your P1 + Codex bot's P1)**
- Extracted `getConnectorBadgeToolCount(connector)` as a named export with a doc-comment explicitly warning future readers not to reuse it for empty-state / loading-gate decisions.
- `ConnectorDetailDrawer` now uses `badgeToolCount` for the header badge and `inventoryToolCount = connector.tools.length` (inline) for the loading gate, the section count, and the empty-state branch. The rendered list is still `connector.tools.map(...)`, so the two now align.
- `ConnectorCard` keeps the helper for its single badge; no inventory section there.

**CI typecheck (your inline)**
- Bound `firstProvisionedTool = provisionedTools[0]` then `expect(...).toBeDefined()` before reading `.safety`, matching your exact suggestion. This unblocks `tsconfig.tests.json` under `noUncheckedIndexedAccess`.
- Process gap on my end: I had only been running `tsc -p tsconfig.json --noEmit`, missing the test-config compilation that the workspace `typecheck` script chains. Added that to my own pre-push checklist.

**Regression coverage** (new in `apps/web/tests/components/EntryView.test.ts`):
- 4 tests on `getConnectorBadgeToolCount`. The most important one explicitly constructs the case your review described: `allowedToolNames=[]` + `tools.length=800`, asserting badge=0 and `tools.length` stays at 800. If anyone in the future reuses the badge value for empty-state decisions, this test pins the distinction with a comment that points back to your review.

Local verification on the new commit: daemon vitest 921/921, web vitest 336/336, all 3 typechecks clean, i18n-check passes.

apps/web/src/components/EntryView.tsx
```diff
@@ -149,6 +149,32 @@ function scoreConnectorText(value: string | undefined, query: string, baseScore:
...
+ *
+ * **Do not** reuse this for drawer empty-state / loading-gate
+ * decisions: those surfaces render `connector.tools` directly, so
+ * they need `connector.tools.length` instead. Mixing the two is the
+ * regression #767 review caught — a connector with an empty
+ * allowlist but a non-empty inventory would render "no tools
+ * available" while the actual inventory was suppressed.
+ */
+export function getConnectorBadgeToolCount(connector: ConnectorDetail): number {
+  return connector.allowedToolNames?.length ?? connector.tools.length;
```
mrcfps: This helper still uses `allowedToolNames.length` as the stable badge value, but `allowedToolNames` is the dynamic execution allowlist, not just the static curated catalog. The daemon's Composio hydration path appends every live tool whose safety is `read` + `auto` into `allowedToolNames` (`apps/daemon/src/connectors/composio.ts:758-761`), and the new contract comment explicitly documents that provider-discovered read-only tools are included. That means a connector can still jump from the static two-tool fallback to dozens of read/list/get provider tools as soon as hydration runs, so the card badge is still tied to provider inventory discovery and does not fully satisfy the PR's stable-count goal. Please keep `allowedToolNames` for execution gating, but expose a separate static/curated badge surface (for example `badgeToolNames`, `curatedToolNames`, or `badgeToolCount` populated from the static definition only) and have this helper read that instead; add a regression test with read-auto live tools to prove the badge does not grow on hydration.

mrcfps (changes requested):
@Sid-Qin Thanks for the quick follow-up on the drawer split and the typecheck issue — those fixes make the current shape much safer. I found one remaining #748 correctness gap around using the dynamic execution allowlist as the badge source. 🙏


Sid-Qin:
@lefarcen @mrcfps both findings addressed in `544b698`. You were right — `allowedToolNames` isn't actually stable across hydration; my regression test only covered write-classified tools so the read-path inflation slipped through. Sorry for missing it on the first pass.

## What changed

Introduced **`curatedToolNames`** as a separate field on the wire `ConnectorDetail`:
- Locked to the static catalog set, **never extended** by Composio discovery
- Composio's `definitionFromToolkit` sets it from `staticDefinition.allowedToolNames` (catalog only) and leaves the runtime `allowedToolNames` alone — so the execution gate keeps its dynamic auto-allow behavior end-to-end
- Non-Composio connectors skip the field, and `connectorDefinitionToDetail()` falls back to `allowedToolNames` for them — they don't have a discovery layer so the two are equivalent anyway

`getConnectorBadgeToolCount` is now a 3-tier fallback: `curatedToolNames → allowedToolNames → tools.length`. Order matters: a half-deployed daemon (allowed but not curated) still gets a more meaningful number than the raw inventory; only with both missing do we hit the original buggy behavior.

## Numbers (the @lefarcen GitHub example)

| State | `tools.length` | `allowedToolNames.length` | `curatedToolNames.length` | Badge in this rev |
|---|---|---|---|---|
| Pre-hydration | 2 | 2 | 2 | **2 tools** ✓ |
| Post-hydration with ~50 read auto-discovered | ≈868 | ≈52 | **2** | **2 tools** ✓ stable |
| Older daemon (no curatedToolNames yet) | 2 | 2 | — | falls back to `allowedToolNames`, still 2 ✓ |

## Test coverage gap @lefarcen flagged

The new daemon test `keeps curatedToolNames at the catalog size even when allowedToolNames is dynamically extended` simulates exactly the Composio post-hydration shape: `curatedToolNames=['docs.search']`, `allowedToolNames=['docs.search', ...50 read_op_*]`. It asserts the badge surface stays at 1 while the execution surface grows to 51. The web test mirrors that with the full `tools=800, allowed=52, curated=2 → badge=2` assertion.

I also added a defensive-copy test on `curatedToolNames` (matching the pattern your earlier review pointed me at for `allowedToolNames`).

## On render-level coverage

I didn't add RTL render tests for the card / drawer in this round — the existing `EntryView.test.ts` is built around `renderToStaticMarkup` (server-side, no event dispatch), so adding RTL would mean introducing a new test infra dependency to this file. The `getConnectorBadgeToolCount` helper now carries the entire badge logic and is unit-tested at every fallback tier, so the "badge = curated, drawer = inventory" contract is pinned at the data layer. Happy to follow up with RTL render tests as a separate PR if maintainers want; let me know.

## Verified locally

- daemon vitest: **924/924** (was 921, +3)
- web vitest: **337/337** (was 332, +5 net)
- daemon `tsconfig.json` + `tsconfig.tests.json` (the CI killer): both clean
- web `tsc -b --noEmit` clean, contracts typecheck clean
- i18n-check passes
- Live `/api/connectors/discovery` confirmed shipping `curatedToolNames: ['github.github_search_repositories', 'github.github_get_issue']` for GitHub

lefarcen (commented):
The `curatedToolNames` split resolves mrcfps's finding about dynamic allowlist growth. The badge now anchors to the static catalog (locked at hydration time), while `allowedToolNames` keeps its dynamic auto-allow behavior for execution gating. The 3-tier fallback (`curated → allowed → tools.length`) preserves backward compatibility with older daemon builds, and the test coverage pins the #748 stability guarantee plus the #767 drawer regression scenario. Looks solid to me; deferring final approval to mrcfps to confirm the fix addresses the review.

mrcfps has not re-reviewed this head (`544b698`) yet — this is bot's verification of the author's claim that the remaining finding is addressed.

Sid-Qin:
Thanks @lefarcen 🙏

CI is green on `544b698` — `Validate workspace` completed SUCCESS, including the `tsconfig.tests.json` step that bit the previous iteration. The `noUncheckedIndexedAccess` index-access pattern that broke last round is gone (now using the `defined`-checked binding pattern you approved earlier).

@mrcfps when you're back — the `544b698` head adds `curatedToolNames` as a separate wire field locked to the static catalog (never extended by Composio discovery), and the badge now reads it via a `curated → allowed → tools.length` fallback. New daemon test simulates the exact post-hydration shape your review described (`allowedToolNames` grown to 51 by auto-allowed read tools, `curatedToolNames` stays at 1) — and the new web test pins `tools=800, allowed=52, curated=2 → badge=2`. Awaiting your re-review to confirm the dynamic-allowlist-as-badge-source concern is closed.

mrcfps (dismissed):
@Sid-Qin I re-reviewed the current `544b698` diff, including the new `curatedToolNames` contract, the Composio hydration path, the badge helper fallback, the drawer inventory split, and the added regression coverage. The badge is now pinned to the static catalog count while `allowedToolNames` remains free to grow for execution gating, and the live CI check is green. Thanks for the thoughtful follow-up on the earlier review threads — this looks ready to me. 🙏


lefarcen (commented):
Hey @Sid-Qin, verified that commit `544b698` fully resolves @mrcfps's final finding about the dynamic allowlist vs static catalog split. 👍

## What's correct now

**`curatedToolNames` is locked to the static catalog** — `apps/daemon/src/connectors/composio.ts:770` explicitly sets it from `staticDefinition.allowedToolNames` and never extends it with `autoAllowedLiveToolNames`, exactly as the review asked. The comment above the line even calls out that it "intentionally never picks up `autoAllowedLiveToolNames`".

**Badge helper uses the 3-tier fallback** — `getConnectorBadgeToolCount` reads `curatedToolNames` first, then `allowedToolNames`, then `tools.length`. This keeps the badge stable across hydration for new daemons while staying defensive against older half-deployed builds.

**Regression coverage is present** — the web test `stays at the curated size after hydration grows allowedToolNames with read-only auto-discovered tools` constructs the exact shape mrcfps described: `curatedToolNames=['docs.search', 'docs.fetch']` (2 items), `allowedToolNames=['docs.search', 'docs.fetch', ...50 read tools]` (52 items), `tools.length=800`. It asserts the badge reports 2, not 52 or 800. The daemon test mirrors that with its own `keeps curatedToolNames at the catalog size` test.

## Numbers (GitHub example from your comment)

| State | `tools.length` | `allowedToolNames.length` | `curatedToolNames.length` | Badge |
|---|---|---|---|---|
| Pre-hydration | 2 | 2 | 2 | **2 tools** ✓ |
| Post-hydration with ~50 read auto-discovered | ≈868 | ≈52 | **2** | **2 tools** ✓ stable |

The badge will stay at 2 across hydration because `curatedToolNames.length` never changes. Clean fix. 🎉

lefarcen:
Hi @Sid-Qin — I rebased this branch onto current `main` and pushed the result. While merging I noticed something worth flagging: the `connector card / drawer` UI this PR was modifying has since been refactored out of `EntryView.tsx` into a dedicated `ConnectorsBrowser.tsx`, and `main` also landed a `toolCount: number` field on `ConnectorDetail` that looked like it was solving the same #748 problem. So before resolving the conflict mechanically I went and verified what each side actually does.

The conclusion: **your `curatedToolNames` is the right fix; the `main`-side `toolCount` is not actually stable across hydration.** In `apps/daemon/src/connectors/composio.ts` the merged-in code computes `toolCount` as:

```ts
toolCount = toolPage?.totalItems ?? liveToolCount ?? staticDefinition.toolCount ?? tools.length;
```

The first branch (`toolPage.totalItems`) is the Composio API's totalItems — which only exists post-hydration and reports the full provider inventory (e.g. ~48 for GitHub). So the badge would still jump from `2` (catalog baseline) to `48` once a Composio key gets configured. That's exactly the #748 symptom this PR was written to fix. Your `curatedToolNames`, mirroring `staticDefinition.allowedToolNames` only, doesn't grow on hydration.

Given that, I kept your contracts + daemon changes and merged them with `main`'s additions rather than dropping them. Specifically:

1. **`packages/contracts/src/api/connectors.ts` + `apps/daemon/src/connectors/catalog.ts`** — kept `allowedToolNames` / `curatedToolNames` from this PR alongside `toolCount` / `toolsNextCursor` / `toolsHasMore` from `main`. Both fields end up on the wire; they answer different questions (execution gate vs. stable badge anchor).
2. **`apps/web/src/components/ConnectorsBrowser.tsx`** — wired `getConnectorDisplayToolCount` to prefer `curatedToolNames.length`, falling back to `toolCount ?? tools.length` only when the curated list is empty/missing. This is the change that makes the new badge UI actually benefit from this PR's stability fix.
3. **`apps/web/src/components/EntryView.tsx` + its test** — took `main`'s version and dropped the ~600-line `ConnectorsTab` block this PR was rewriting; that whole component has been moved to `ConnectorsBrowser.tsx` upstream, so the in-PR rewrite would have been dead code.
4. **Type relaxation** — flipped `allowedToolNames` / `curatedToolNames` from required to optional on the wire shape so I didn't have to retrofit every existing fixture. The daemon-built payloads always populate both, so runtime consumers see no semantic difference; the optionality is a fixture-brevity convenience and the JSDoc spells that out.
5. **Test fixtures + non-null assertions** — touched `EntryView.test.ts`, `SettingsDialog.orbit.test.tsx`, and the relevant places in `connectors-service.test.ts` to keep them compiling against the new optional shape.

Validation done locally: `pnpm --filter @open-design/contracts typecheck`, `... web typecheck`, `... daemon typecheck` all clean. `pnpm --filter @open-design/web test` 565/565 passing. The `connectors-service.test.ts` regression guards (including the #748 / #767 stability cases) all pass, 35/35.

One thing I did not do: re-run the Playwright E2E suite for connector flows. If you have time before merge, a full `pnpm test:ui` pass would be reassuring. Otherwise CI should cover it.

Thanks for sticking with this — the underlying fix is the right one and it deserved to land.

apps/web/src/components/ConnectorsBrowser.tsx
```diff
@@ -312,6 +312,15 @@ export function clearConnectorAuthorizationPending(
 }
 
 export function getConnectorDisplayToolCount(connector: ConnectorDetail): number {
+  // Prefer the hand-curated catalog subset so the card/header badge stays
+  // stable across Composio hydration: `toolCount` is sourced from the
+  // provider's totalItems (a GitHub-style toolkit jumps from 2 to 48 once an
+  // API key is configured), and `tools.length` shifts the same way after
+  // hydration. `curatedToolNames` is fixed at the catalog and is what the
+  // "N tools" badge is meant to reflect (issue #748).
+  if (connector.curatedToolNames && connector.curatedToolNames.length > 0) {
```
mrcfps, line 321: This helper now prefers `curatedToolNames`, but when that field is absent it falls straight back to `toolCount`. That reintroduces the #748 badge jump on any older daemon that already exposes `toolCount` but not `curatedToolNames`: `toolCount` in `apps/daemon/src/connectors/composio.ts:1052` is populated from `toolPage?.totalItems ?? liveToolCount`, so a GitHub-style connector still moves from the static 2-tool catalog to the hydrated provider total as soon as discovery runs. Because this PR also makes `allowedToolNames`/`curatedToolNames` optional specifically to tolerate older payloads, this mixed-version path is part of the changed contract. Please fall back through `allowedToolNames?.length` before `toolCount` (or add an equivalent stable badge-count field) and add a helper test that deletes `curatedToolNames` while leaving `allowedToolNames` present, so the rollout behavior stays pinned.

mrcfps (commented):
@Sid-Qin Thanks for carrying the connector-count fix through the rebase. I found one rollout-compatibility gap in the new badge helper around mixed-version daemon/web deployments. 🙏

PR #768, opened 2026-05-07
Improve Windows beta packaging and installer flow

PerishCode:
## What This PR Optimizes

This branch is primarily a Windows packaging optimization and release-readiness pass. The final `origin/main` merge had no textual conflicts; it should be treated as housekeeping, not the review focus.

The main outcomes are:

- Reduce Windows packaged artifact size and make remaining size tradeoffs visible.
- Make local `tools-pack win build --to dir` iteration fast enough for repeated inspect/eval smoke loops.
- Add bounded, fallback-friendly Windows `release-beta` cache behavior.
- Replace the fragile default Windows installer flow with a custom NSIS flow that supports QA-ready install, overwrite, launch, shortcut, uninstall, cleanup, and diagnosis behavior.

## 1. Windows Package Size

Baseline Windows `dir` unpacked artifact:

- `1,190,474,461` bytes

Low-risk pruning tranche:

- `1,190,474,461` -> `946,893,051` bytes
- Saved `243,581,410` bytes, about `232.30 MiB`

Later pruning tranches:

- Removed root Next payload where standalone runtime no longer needs it.
- Removed bundled Node runtime after validating Electron ABI rebuild strategy.
- Pruned duplicate copied-standalone Next payload while preserving app-local Next resolution.
- Recent Windows dir builds now sit around `556.9 MiB` unpacked.

Current stop point:

- Remaining large buckets are product/runtime-sensitive rather than low-risk cleanup:
  - Electron locales: about `48.6 MiB`
  - app-local Next runtime: about `82 MiB`
  - bundled product resources: about `50 MiB`
- Further size work should be a product/runtime decision, not an automatic pruning pass.

## 2. Local Windows Tools-Pack Iteration Speed

The cache work is structured as immutable keyed artifacts, not namespace output. Namespace-local mutable behavior stays outside cache through `built-app.json` and `OD_PACKAGED_CONFIG_PATH`.

Measured progression:

- Before cache tranches: hot Windows `dir` builds were about `90-118s`.
- After initial cache layers: about `35.7s`.
- `PathSizeIndex` reduced size-report from about `7.8s` to `~0.66-0.69s`.
- Final direct `win.electron-builder-dir` reuse reduced hot `dir` builds to about `3.0-3.6s`.
- Follow-up cache experiments measured fully hot builds at about `2.4-2.5s` wall time, with internal phase total about `1.03-1.05s`.

Invalidation behavior checked:

- Resource-only edits miss `win.resource-tree` and `win.electron-builder-dir`; once cached, they return to the `~2.4s` hot path.
- Packaged-source edits invalidate workspace/build/native/builder layers; once cached, they also return to the same hot path.

Current stop point:

- Remaining repo-internal hot-path costs are sub-second scale.
- pnpm/CLI startup is already about `1.4s` of wall time, so more local dir cache work has low ROI without a new concrete bottleneck.

## 3. Windows Release-Beta Cache Path

The workflow cache strategy follows the constraints we set during review:

- Cache restore may fail completely; the workflow falls back to a normal build.
- Cache growth is bounded by pruning before save.
- Dependency install cache is separated from tools-pack build/cache artifacts.
- GHA cache usage stays on standard cache declarations, not legacy ad hoc behavior.
- Action versions were refreshed while keeping the workflow logic thin.

Windows-only release-beta validation:

- Latest successful run: https://github.com/nexu-io/open-design/actions/runs/25480315353
- Windows job total: `7m13s`
- Build artifacts step: `~3m06s`
- Smoke step: `~1m31s`
- Dependency install: `~21s`
- NSIS setup: `~30s`

A previous successful run before the long-path uninstall fix completed the Windows job in `5m34s`, with build artifacts `~2m29s` and smoke `~1m02s`.

## 4. Windows Installer Flow

This PR replaces the default Windows installer path with a custom NSIS flow tuned for the beta installer.

Installer UX/correctness:

- Adds welcome, directory, install, and finish pages.
- Detects running installed instances before overwrite detection.
- Running-instance detection is scoped to the selected/registered install root, not broad process-name matching.
- Running-instance page offers `Close and continue` / cancel and does not expose PID details.
- Existing install detection prompts overwrite or cancel.
- Finish page offers `Create desktop shortcut` and `Launch Open Design`.
- NSIS LangString entries fall back to English for untranslated languages.

Uninstaller UX/correctness:

- Offers `Remove desktop shortcut`.
- Offers `Delete local data for this installation`.
- Uses namespace/beta-isolated registry keys, App Paths, shortcuts, and local data paths.
- Beta display/shortcut identity is `Open Design Beta`.
- Uses vendored `7z.exe`/`7z.dll` for fast payload extraction, with license notes documented in the vendor folder.
- Adds NSIS state logging for path existence and registry/delete milestones.

Post-QA leak check:

- No Open Design processes remained.
- No uninstall registry entries remained.
- No App Paths registry entries remained.
- No desktop/public desktop/start-menu shortcuts remained.
- No product namespace data remained.
- Found 24 small long-path Next files under `%LOCALAPPDATA%\Programs\Open Design`; fixed by using extended-length `\\?\` removal for install dir and local data cleanup.

## QA Artifact

Latest unsigned Windows beta installer:

- Release: https://github.com/nexu-io/open-design/releases/tag/open-design-v0.4.2-beta.16.unsigned
- Installer: `open-design-0.4.2-beta.16.unsigned-win-x64-setup.exe`
- SHA256: `2665f0b01baa1058164dbce10c3c0b1efaf077a8d7a362eeac87a52d76067358`

## Validation

Final validation after the latest `origin/main` merge:

- `pnpm guard`
- `pnpm typecheck`
- `pnpm --filter @open-design/tools-pack test`
- Local Windows NSIS build/install/uninstall smoke with namespace `rg`
- Windows-only release-beta workflow succeeded: https://github.com/nexu-io/open-design/actions/runs/25480315353

Earlier capability regression in this branch also covered:

- `tools-dev` daemon/web/desktop start, inspect, logs/status, and stop.
- Windows `dir` start/inspect/logs/stop from direct cache reuse.
- Windows NSIS build/install/start/inspect/logs/stop/cleanup with short namespace `rg`.

## Merge Notes

- `origin/main` merge had no textual conflicts.
- Auto-merged upstream areas were daemon/web changes from main and did not require product decisions for this branch.
- The only post-merge fix was test-only: main added mac tools-pack tests that needed Windows-compatible path expectations and the current `ToolPackRoots.cacheRoot` mock field.


mrcfps (dismissed):
@PerishCode I reviewed the changed ranges for the Windows packaging/cache work, the custom NSIS installer flow, the release-beta workflow updates, and the related daemon/web/test adjustments. I did not find any actionable correctness, safety, test, or maintainability issues that need changes before merge; thanks for the careful packaging improvements and validation notes. 🚀


tools/pack/src/win/custom-installer.ts
```diff
@@ -0,0 +1,559 @@
...
+UninstallIcon "\${APP_ICON}"
+ShowInstDetails show
+ShowUninstDetails hide
+
+!define MUI_ABORTWARNING
+!define MUI_ICON "\${APP_ICON}"
+!define MUI_UNICON "\${APP_ICON}"
+Page custom RunningInstancesPage RunningInstancesPageLeave
+!insertmacro MUI_PAGE_WELCOME
+!define MUI_PAGE_CUSTOMFUNCTION_LEAVE DirectoryPageLeave
```
lefarcen, line 147: P2: `writeInstallerScript` writes a large NSIS script as a template string. If any of the dynamic values (e.g., `localDataRoot`, `nsisLogPath`) contain characters that break NSIS syntax (e.g., unescaped `$`), the installer will fail at runtime rather than build time. Consider adding validation for these paths or a test that exercises the script with edge-case values.

tools/pack/src/cache.ts
```diff
@@ -0,0 +1,296 @@
...
+
+async function assertOutputsExist(entryRoot: string, outputs: string[]): Promise<CacheInvalidation | null> {
+  for (const output of outputs) {
+    if (!(await pathExists(join(entryRoot, output)))) {
+      return { reason: `missing output: ${output}` };
+    }
+  }
+  return null;
+}
+
```
lefarcen, line 89: P2: `acquire` checks `outputs` existence but doesn't verify they're non-empty or have expected structure. A cache entry with zero-byte files would pass staleness check but fail downstream. Consider adding a size/content validation step for critical outputs (e.g., `better_sqlite3.node` must be > 100KB).

tools/pack/src/win/registry.ts
```diff
@@ -0,0 +1,181 @@
...
+  const uninstallerPath = uninstallerFromRegistry.length > 0 ? uninstallerFromRegistry : join(installDir, identity.uninstallerName);
+  return {
+    ...paths,
+    installDir,
+    installedExePath: join(installDir, identity.exeName),
+    uninstallerPath,
+  };
+}
+
+export async function cleanupWinRegistryResidues(
```
lefarcen, line 166: P3: `cleanupWinRegistryResidues` swallows HKLM deletion failures silently (catch block is empty). This is fine for elevation-required keys, but logging which keys failed would help debugging. Consider logging to `nsisLogPath` or a separate registry-cleanup.log.

tools/pack/src/win/lifecycle.ts
```diff
@@ -0,0 +1,429 @@
...
+function installArgs(config: ToolPackConfig, paths: WinPaths): string[] {
+  return [...(config.silent ? ["/S"] : []), `/D=${paths.installDir}`];
+}
+
+async function writeJsonMarker(filePath: string, payload: Record<string, unknown>): Promise<void> {
+  await mkdir(dirname(filePath), { recursive: true });
+  await writeFile(filePath, `${JSON.stringify(payload, null, 2)}\n`, "utf8");
+}
+
+async function collectFileTreeStats(root: string): Promise<{ fileCount: number; totalBytes: number }> {
```
lefarcen, line 94: P2: `collectFileTreeStats` recursively walks the install tree. For a large install (e.g., with node_modules), this could be slow. Consider caching the result or making it optional for `installPackedWinApp` unless the user explicitly requests install payload stats.

e2e/specs/win.spec.ts
```diff
@@ -0,0 +1,424 @@
+// @vitest-environment node
```
lefarcen, line 1: P3: This new e2e spec is 424 lines. Does it run in CI on Windows? If not, consider adding a CI workflow or marking it as `skip` by default with instructions for local testing. Otherwise, it'll bitrot.

lefarcen (commented):
Thanks for this comprehensive Windows packaging refactor, @PerishCode! 👋 This is a substantial improvement to the build/install flow. I've reviewed the structure and found a few areas worth addressing:

**Architecture & Design**

✅ The modular split (win.ts → win/*.ts) is clean — identity.ts, registry.ts, paths.ts are well-isolated concerns.

✅ Cache layer (cache.ts + workspace-build.ts) is a smart addition for CI/local dev speed.

✅ Custom NSIS installer (custom-installer.ts) gives better control than electron-builder's default.

**Findings — see inline comments for specifics:**

The main issues center on error handling in the custom installer flow and cache invalidation edge cases. These are P2 (should-fix) rather than blocking — the PR is production-ready with these addressed.

Overall: **Strong work**. The refactor sets up a much more maintainable Windows packaging story. Once the inline items are resolved, this is ready to ship.

tools/pack/src/package-source-hash.ts
```diff
@@ -0,0 +1,46 @@
...
+import { basename, dirname, join, relative } from "node:path";
+
+function normalizeRelativePath(path: string): string {
+  return path.split("\\").join("/");
+}
+
+async function readNormalizedFile(filePath: string, relativePath: string): Promise<Buffer | string> {
+  if (basename(relativePath) === "package.json") {
+    const value = JSON.parse(await readFile(filePath, "utf8")) as Record<string, unknown>;
+    if (typeof value.version === "string") value.version = "0.0.0-cache";
```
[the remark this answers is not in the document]
PerishCode: Thanks, fixed in `64502bb`. `hashPackageSourcePath` now hashes `package.json` as-is, including `version`, and `tools/pack/tests/package-source-hash.test.ts` covers version-only changes. This makes workspace-build and workspace-tarballs cache keys miss correctly after beta version bumps.

tools/pack/src/win/custom-installer.ts
PerishCode: Thanks, fixed in `64502bb`. `escapeNsisString` now escapes literal `$` as `$$`, while `localDataRoot` preserves the intended `$APPDATA` NSIS variable by escaping only literal path segments. I also rebuilt a Windows NSIS installer locally with namespace `review-fixes`; makensis succeeded.

tools/pack/src/cache.ts
PerishCode: Thanks, addressed in `64502bb` at the cache-node boundary. `win.packaged-app` now validates the critical `better_sqlite3.node` output on build and cache hit, requiring it to exist and be larger than 100KB. I kept `ToolPackCache` generic because each node has different meaningful structure checks.

e2e/specs/win.spec.ts
PerishCode: Thanks for flagging this. It does run in CI for the Windows release path: `.github/workflows/release-beta.yml` invokes `pnpm test specs/win.spec.ts` in the `Smoke beta windows packaged runtime` step with `OD_PACKAGED_E2E_NAMESPACE=release-beta-win`. The latest Windows-only release-beta run also passed that smoke step.

tools/pack/src/win/registry.ts
PerishCode: Thanks, agreed this would help diagnosis. I’m leaving it out of this pass because failed HKLM deletes remain visible in the follow-up residue observation, and this cleanup path is lower risk than the cache correctness fixes. I’ll keep it as a follow-up polish item rather than expanding this PR.

tools/pack/src/win/lifecycle.ts
PerishCode: Thanks, fair point. I’m keeping the stats for now because they only run in `tools-pack win install --json` diagnostics after install, not in the user-facing installer path, and they helped catch install completeness issues during this work. If this becomes noisy in practice, making it opt-in is the right follow-up.

tools/pack/src/win/builder.ts
```diff
@@ -0,0 +1,303 @@
...
+    appBuilderOutputRoot: cachedBuilderRoot,
+    cacheEntryPath: manifest.entryPath,
+    configPath: paths.packagedConfigPath,
+    executablePath: cachedExecutablePath,
+    source: "cache",
+    unpackedRoot: cachedUnpackedRoot,
+    webStandaloneHookAuditPath: (await pathExists(paths.webStandaloneHookAuditPath)) ? paths.webStandaloneHookAuditPath : null,
+  });
+  if (config.to === "nsis" || config.to === "all") {
+    await buildCustomWinNsisInstaller(config, paths, {
```
mrcfps: This builds the NSIS payload directly from `cachedUnpackedRoot`, but the cached `win-unpacked/resources/open-design-config.json` was created when the cache entry was first built. The cache key intentionally omits `namespace`, while `writePackagedConfig(config, paths, packagedVersion)` only writes the current namespace's config beside the output manifest; that fresh file is not copied into `cachedUnpackedRoot` before `buildCustomWinNsisInstaller` packages it. A hot `tools-pack win build --to nsis` for a second namespace can therefore emit an installer whose registry/paths are for namespace B but whose packaged runtime boots with namespace A from the stale embedded config, so the installed app writes data/logs under the wrong namespace and breaks the multi-namespace cache contract. Please either make the builder-dir cache namespace/config-specific, or materialize/copy the cached unpacked tree to the current namespace output and overwrite `resources/open-design-config.json` with `paths.packagedConfigPath` before building the installer. A regression test should build two namespaces through the hot cache and assert the installed/resource config namespace matches the second namespace.

tools/pack/src/win/custom-installer.ts
```diff
@@ -0,0 +1,559 @@
...
+    Call LogInstallerEvent
+  \${EndIf}
+  Pop $1
+  Pop $0
+FunctionEnd
+
+Function CloseRunningInstances
+  Push $0
+  Push $1
+  nsExec::ExecToStack 'powershell.exe -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Bypass -Command "& { param($$install, $$registered); $$paths = @($$install, $$registered) | Where-Object { $$_ } | ForEach-Object { $$_.TrimEnd([char]92).ToLowerInvariant() } | Select-Object -Unique; $$matches = Get-CimInstance Win32_Process | Where-Object { $$matched = $$false; $$exe = $$_.ExecutablePath; if ($$null -ne $$exe) { $$exe = $$exe.ToLowerInvariant(); foreach ($$path in $$paths) { if ($$path -and $$exe.StartsWith($$path)) { $$matched = $$true; break } } }; $$matched }; $$ids = @($$matches | ForEach-Object { $$_.ProcessId }); foreach ($$id in $$ids) { try { [void][System.Diagnostics.Process]::GetProcessById($$id).CloseMainWindow() } catch {} }; Start-Sleep -Milliseconds 1500; foreach ($$id in $$ids) { try { $$p = [System.Diagnostics.Process]::GetProcessById($$id); if (-not $$p.HasExited) { Stop-Process -Id $$id -Force -ErrorAction SilentlyContinue } } catch {} }; if ($$ids) { $$ids -join ([char]32) } }" "$INSTDIR" "$ExistingInstallLocation"'
```
mrcfps: The running-instance predicate trims the install root and then matches `ExecutablePath.StartsWith($path)` in both detection and the forced close path. Without checking for equality or a path separator boundary, an install root like `C:\...\Open Design` also matches unrelated sibling paths such as `C:\...\Open Design Tools\...`; the `Close and continue` flow can then call `Stop-Process -Force` on those unrelated processes. Since this installer now automatically closes matches, please normalize each candidate root with a trailing `\` and match `($exe -eq $path) -or $exe.StartsWith($pathWithSeparator)` (or a shared helper using path-aware comparison) before both reporting and stopping PIDs.

mrcfps (dismissed):
@PerishCode thanks for pushing this Windows packaging and installer work forward — the cache layering and smoke coverage are a solid improvement. I found one blocker in the cached NSIS path, plus one safety issue around the installer process matcher.


tools/pack/src/win/builder.ts
PerishCode: Thanks, fixed in `8fc1629`. The NSIS path now materializes the cached `win-unpacked` into the current namespace output and overwrites `resources/open-design-config.json` with the current packaged config before packaging. I added a focused regression test for that overwrite behavior and also validated two hot-cache namespace builds locally: `review-cache-a` then `review-cache-b`, with the second output embedding `review-cache-b`.

tools/pack/src/win/custom-installer.ts
PerishCode: Thanks, fixed in `8fc1629`. Detection and close now use path-aware matching: exact install root or root plus a path-separator prefix. That keeps the `Close and continue` path from matching sibling install roots like `Open Design Tools`; I also smoke-checked the generated predicate against that sibling-path case.

mrcfps (commented):
@PerishCode thanks for the follow-up fixes on the Windows packaging/cache path — the namespace materialization and path-aware installer close logic look much safer now. I found one remaining cache-key edge case that is merge-safe but worth tightening so alternate Windows web output modes cannot reuse stale web tarballs. 👍


tools/pack/src/win/app.ts
```diff
@@ -0,0 +1,272 @@
...
+  const rootPackageJson = JSON.parse(await readFile(join(config.workspaceRoot, "package.json"), "utf8")) as {
+    packageManager?: unknown;
+  };
+
+  return hashJson({
+    node: "win.workspace-tarballs",
+    packageHashes,
+    packageManager: rootPackageJson.packageManager,
+    pnpmLock: await hashPath(join(config.workspaceRoot, "pnpm-lock.yaml")),
+    schemaVersion: 3,
```
mrcfps: The workspace tarball cache key omits `config.webOutputMode`, even though this same file builds `@open-design/web` with `OD_WEB_OUTPUT_MODE: config.webOutputMode` and the `@open-design/web` package tarball includes `.next`. If someone runs a Windows build once with `OD_WEB_OUTPUT_MODE=server` and later with the default `standalone` (or vice versa) against the same tools-pack cache, `win.workspace-tarballs` can hit and materialize web tarballs containing the previous `.next` shape; the downstream `win.packaged-app` key includes `webOutputMode`, but it will still install those stale tarballs and produce an app whose web payload does not match the requested mode. Please add `webOutputMode: config.webOutputMode` here (or include the workspace-build cache key that already captures it) and add a small cache-key regression covering server vs standalone so the tarball layer invalidates with the web output mode.

PerishCode: Thanks, tightened in `6ac254d`. `win.workspace-tarballs` now includes `webOutputMode` in its cache key, so server and standalone web outputs cannot reuse each other’s tarball layer. I also added a focused regression test that asserts the two modes produce different tarball cache keys.

.github/workflows/release-beta.yml
```diff
@@ -213,22 +213,230 @@ jobs:
...
+            "| Saved cache node | Entries | Size |" | Add-Content -Path $summaryPath
+            "| --- | ---: | ---: |" | Add-Content -Path $summaryPath
+            foreach ($entry in $entries) {
+              $mb = [math]::Round(([double]$entry.Size) / 1MB, 1)
+              "| `$($entry.Node)` | $($entry.Count) | ${mb} MB |" | Add-Content -Path $summaryPath
+            }
+          }
+
+      - name: Save Windows tools-pack cache
+        if: ${{ success() && steps.win_tools_pack_cache_restore.outputs.cache-hit != 'true' }}
```
mrcfps, line 434: The save step is currently skipped whenever the monthly restore key hits exactly. Because `actions/cache` caches are immutable, that means the first successful Windows release-beta run for a month freezes `tools-pack-win-v6-${{ runner.os }}-${{ steps.win_tools_pack_cache_key.outputs.epoch }}`; later commits can populate new internal `win.workspace-build`/`win.electron-builder-dir` entries locally during the job, but this condition prevents those fresh entries from ever being uploaded. The build still succeeds, so this is merge-safe, but it undermines the PR's release-beta cache goal by making repeated builds after source changes miss the tools-pack cache until the next month. Please switch the saved key to a unique suffix, for example `tools-pack-win-v6-${{ runner.os }}-${{ steps.win_tools_pack_cache_key.outputs.epoch }}-${{ github.sha }}` (or another run/content-specific token) while keeping the monthly restore prefix, so each successful pruned cache can be saved and future runs can restore the newest compatible cache.

mrcfps (commented):
@PerishCode thanks for the quick follow-up on the Windows packaging cache and installer fixes — the current head addresses the earlier correctness concerns and the smoke coverage is in good shape. I found one merge-safe cache freshness issue in the release-beta workflow that is worth tightening so the CI cache keeps learning across successful runs. 👍


.github/workflows/release-beta.yml
PerishCode: Thanks, fixed in `5a75290`. The Windows tools-pack cache now restores with a SHA-specific key plus the monthly restore prefix, and saves with the same SHA-specific key. That keeps exact reruns cacheable while allowing each successful commit to upload its pruned cache for later monthly-prefix restores.

Siri-Ray (approved):

PR #770, opened 2026-05-07
fix: add DeepSeek v4 models to catalog (#722)

### LINKED ISSUE #722: Unable to complete large-scale tasks.
When performing tasks with a large workload, effective output cannot be generated and the tasks are always interrupted.

Image

lefarcen:
## Fixes
#722

## What changed
Added `deepseek-v4-pro` and `deepseek-v4-flash` entries to `litellm-models.json` with correct `max_output_tokens` (384000) per [DeepSeek official docs](https://platform.deepseek.com/docs/model-cards).

## Why
Users configuring DeepSeek v4 models via BYOK were falling back to `deepseek-chat` (8192 output limit) because the model catalog didn't have v4 entries, causing large multi-screen generation tasks to fail mid-stream.

## Verification
- [x] Added both model entries with 384000 max_output_tokens
- [x] Matches DeepSeek official spec
- [x] Only config change, no code logic modified

## Scope
- Risk: **low** (config-only update)
- Impact: Only BYOK DeepSeek v4 users benefit


apps/web/src/state/litellm-models.json
```diff
@@ -580,6 +580,8 @@
     "deepinfra/Sao10K/L3.3-70B-Euryale-v2.3": 131072,
     "deepinfra/zai-org/GLM-4.5": 131072,
     "deepseek-chat": 8192,
+    "deepseek-v4-flash": 384000,
+    "deepseek-v4-pro": 384000,
```
mrcfps: These DeepSeek entries are being added directly to `litellm-models.json`, but the repo treats that file as a generated LiteLLM snapshot: `scripts/sync-litellm-models.ts` rewrites it from upstream, and `apps/web/src/state/maxTokens.ts` says models LiteLLM does not track should live in `OVERRIDES`. Since the PR says these ids are needed because the catalog is missing them, a future normal sync can silently delete these two lines and regress BYOK DeepSeek v4 users back to the 8192 fallback. Please move these two values into `OVERRIDES` in `apps/web/src/state/maxTokens.ts` instead, and add/extend `apps/web/tests/state/maxTokens.test.ts` to assert the DeepSeek v4 defaults come from the override path.

mrcfps (commented):
@lefarcen Thanks for adding the DeepSeek v4 coverage here — I found one follow-up to make sure the new caps survive the existing catalog regeneration path. 🚀



lefarcen:
Fixed alphabetical ordering — moved `deepseek-v4-*` entries to the correct position (after `deepseek-v3-2-251201`, before `deepseek.v3-v1:0`) to maintain the catalog's generated sort convention.

The catalog uses `localeCompare` for ordering, so keeping the correct alphabetical sort prevents merge conflicts when the file is regenerated.

apps/web/src/state/litellm-models.json
```diff
@@ -582,6 +582,8 @@
     "deepseek-chat": 8192,
     "deepseek-reasoner": 65536,
     "deepseek-v3-2-251201": 32768,
+    "deepseek-v4-flash": 384000,
+    "deepseek-v4-pro": 384000,
```
mrcfps: These DeepSeek v4 caps are still being added directly to `litellm-models.json`, but this file is the generated LiteLLM snapshot: `scripts/sync-litellm-models.ts` rewrites it from upstream, and `apps/web/src/state/maxTokens.ts` documents that LiteLLM coverage gaps belong in `OVERRIDES`. Because the PR exists specifically to cover ids that the current catalog does not provide, the next normal sync can silently delete these two lines and regress BYOK DeepSeek v4 users back to the 8192 fallback. Please move `deepseek-v4-flash` and `deepseek-v4-pro` into the `OVERRIDES` table in `apps/web/src/state/maxTokens.ts` instead, and extend `apps/web/tests/state/maxTokens.test.ts` to assert these defaults come from the override path.

mrcfps (commented):
@lefarcen Thanks for the follow-up reorder — the current diff is sorted now, and I found one remaining maintainability point so the DeepSeek v4 caps keep working after the catalog is regenerated. 🚀



lefarcen:
Good catch! Moved both entries to `OVERRIDES` in `maxTokens.ts` and added test coverage asserting the override path returns 384K for both `deepseek-v4-pro` and `deepseek-v4-flash`. This ensures they survive future `sync-litellm-models.ts` runs.

mrcfps (approved):
@lefarcen Thanks for moving the DeepSeek v4 caps into `OVERRIDES` and adding focused max-token coverage. I reviewed the current diff for the override path, generated catalog reset, and the new assertions; the change now preserves the 384K defaults across catalog syncs while keeping the LiteLLM snapshot clean. Appreciate the quick follow-up and tidy fix 🚀

PR #773, opened 2026-05-07
feat(daemon): add language boost support for Minimax TTS

terencewlc:
## Summary
- Add `--language` CLI flag for Minimax TTS language boost parameter
- Pass language parameter through CLI → server → media renderer
- Enable `language_boost` in Minimax TTS API requests

This enables better pronunciation for specific languages like Cantonese (Yue).

## Example
```bash
od media generate audio --text "你好" --voice male-qn-qingse --language Chinese,Yue
```

### DIFF
--- a/apps/daemon/src/cli.ts
+++ b/apps/daemon/src/cli.ts
@@ -43,6 +43,7 @@ const MEDIA_GENERATE_STRING_FLAGS = new Set([
   'composition-dir',
   'image',
   'daemon-url',
+  'language',
 ]);
 const MEDIA_GENERATE_BOOLEAN_FLAGS = new Set([
   'help',
@@ -242,6 +243,7 @@ async function runMediaGenerate(rawArgs) {
     audioKind: flags['audio-kind'],
     compositionDir: flags['composition-dir'],
     image: flags.image,
+    language: flags.language,
   };
   if (flags.length != null) body.length = Number(flags.length);
   if (flags.duration != null) body.duration = Number(flags.duration);
@@ -476,6 +478,7 @@ Common options:
   --length <seconds>        Video length.
   --duration <seconds>      Audio duration.
   --voice <voice-id>        Speech / TTS voice.
+  --language <lang>         Language boost for TTS (e.g. Chinese,Yue for Cantonese).
   --audio-kind music|speech|sfx
   --composition-dir <path>  hyperframes-html only — project-relative path
                             to the dir containing hyperframes.json /
--- a/apps/daemon/src/media.ts
+++ b/apps/daemon/src/media.ts
@@ -210,6 +210,7 @@ function clampWithWarning(value, allowed, flagName) {
  * @param {number} [args.duration]
  * @param {string} [args.voice]
  * @param {string} [args.audioKind]
+ * @param {string} [args.language]
  * @returns {Promise<{ name: string, size: number, mtime: number, kind: string, mime: string, model: string, surface: string, providerNote: string, providerId: string }>}
  */
 export async function generateMedia(args) {
@@ -226,6 +227,7 @@ export async function generateMedia(args) {
     duration,
     voice,
     audioKind,
+    language,
     compositionDir,
     image,
   } = args;
@@ -309,6 +311,7 @@ export async function generateMedia(args) {
     duration: clampedDuration,
     voice: voice || '',
     audioKind: resolvedAudioKind,
+    language: language || '',
     // Project-relative path to the directory the agent scaffolded with
     // hyperframes.json / meta.json / index.html. Only consumed by the
     // hyperframes renderer; null/empty for every other provider.
@@ -1357,6 +1360,7 @@ async function renderMinimaxTTS(ctx, credentials) {
     model: wireModel,
     text,
     stream: false,
+    ...(ctx.language ? { language_boost: ctx.language } : {}),
     voice_setting: {
       voice_id: voiceId,
       speed: 1.0,
--- a/apps/daemon/src/media.ts
+++ b/apps/daemon/src/media.ts
@@ -1356,11 +1356,13 @@ async function renderMinimaxTTS(ctx, credentials) {
   // platform.minimaxi.com under voice management.
   const voiceId = (ctx.voice && ctx.voice.trim()) || 'male-qn-qingse';
 
+  const languageBoost = typeof ctx.language === 'string' ? ctx.language.trim() : '';
+
   const body = {
     model: wireModel,
     text,
     stream: false,
-    ...(ctx.language ? { language_boost: ctx.language } : {}),
+    ...(languageBoost ? { language_boost: languageBoost } : {}),
     voice_setting: {
       voice_id: voiceId,
       speed: 1.0,
--- a/apps/daemon/src/prompts/media-contract.ts
+++ b/apps/daemon/src/prompts/media-contract.ts
@@ -87,6 +87,7 @@ Run via your shell tool (Bash on Claude Code, exec on Codex/Gemini, etc.):
   [--duration <seconds>]            # audio only
   [--audio-kind music|speech|sfx]   # audio only
   [--voice <provider-voice-id>]     # audio:speech only; omit to use provider default
+  [--language <lang>]               # audio:speech only; language boost (e.g. Chinese, Yue for Cantonese)
 \`\`\`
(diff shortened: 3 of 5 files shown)


lefarcen:
Hi @terencewlc! 🎉
Thanks for the contribution — adding Cantonese language boost support for Minimax TTS is a nice quality-of-life improvement.

— open-design team

mrcfps (dismissed):
@terencewlc I reviewed the CLI flag plumbing, server handoff, and MiniMax TTS request construction for the new language boost path. The change is small and focused: `--language` is accepted by `od media generate`, carried through the daemon task request, and conditionally emitted as `language_boost` only when provided. I did not find any actionable correctness, safety, or maintainability issues in the changed ranges. Thanks for the contribution — this should make Cantonese and other language-specific TTS generation easier to use. 🙌


lefarcen (commented):
Hi @terencewlc, thanks for adding Minimax TTS language boost support. The implementation flow (CLI → server → media → Minimax API) is clean and follows the existing pattern for optional flags.

I found one substantive gap and two small polish items — see inline comments below. The biggest one is that the new flag isn't documented in the media generation contract that agents follow, so in practice it's unlikely to be used even though it's technically supported.

apps/daemon/src/cli.ts
```diff
@@ -476,6 +478,7 @@ Common options:
   --length <seconds>        Video length.
   --duration <seconds>      Audio duration.
   --voice <voice-id>        Speech / TTS voice.
+  --language <lang>         Language boost for TTS (e.g. Chinese,Yue for Cantonese).
```
lefarcen, line 481: **P2** The new `--language` flag is added to CLI parsing/help, but not to the agent-facing media generation contract (`apps/daemon/src/prompts/media-contract.ts:78`) or audio workflow prompt. In normal product use, agents explicitly follow `MEDIA_GENERATION_CONTRACT`, whose invocation block lists `--voice` but not `--language`, and the workflow frames speech ambiguity as "voice id / accent / pacing" without telling the agent that MiniMax Cantonese requires `--language Chinese,Yue`. Result: the backend supports the flag, but the primary caller (agents) is unlikely to ever emit it.

Concrete fix: update `apps/daemon/src/prompts/media-contract.ts` and the audio metadata wording in `apps/daemon/src/prompts/system.ts` to document `--language <MiniMax language_boost>` for `minimax-tts`, including valid examples like `Chinese,Yue`, and add a prompt test.

apps/daemon/src/media.ts
```diff
@@ -1357,6 +1360,7 @@ async function renderMinimaxTTS(ctx, credentials) {
     model: wireModel,
     text,
     stream: false,
+    ...(ctx.language ? { language_boost: ctx.language } : {}),
```
lefarcen: **P3** `ctx.language` is forwarded raw, unlike `voice`, which is trimmed before use. A whitespace-only value or accidental surrounding spaces will still emit `language_boost`, likely causing a MiniMax parameter error instead of behaving like "not provided."

Concrete fix: normalize locally with something like `const languageBoost = typeof ctx.language === 'string' ? ctx.language.trim() : ''; ... ...(languageBoost ? { language_boost: languageBoost } : {})`

apps/daemon/src/server.ts
```diff
@@ -3364,6 +3364,7 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
             : undefined,
         voice: req.body?.voice,
         audioKind: req.body?.audioKind,
+        language: req.body?.language,
```
lefarcen: **P3** The REST path accepts and forwards any `req.body.language` value, so non-string JSON values can reach `generateMedia` and then the MiniMax request body. The CLI path only produces strings, but browser/local clients can POST arbitrary JSON to this endpoint.

Concrete fix: mirror the numeric guards already used for `length`/`duration`: pass `language: typeof req.body?.language === 'string' ? req.body.language : undefined`, or reject non-string values with a 400 before queuing the task.

lefarcen (commented):
Thanks for the quick update @terencewlc! 🙌

I can confirm the P2 concern is resolved — the `--language` flag is now properly documented in both media contract files (`apps/daemon/src/prompts/media-contract.ts` and `packages/contracts/src/prompts/media-contract.ts`), with clear guidance that agents should use `Yue` for Cantonese, `Chinese` for Mandarin, etc.

The implementation is clean and focused. Ready for maintainer approval from my side.

apps/daemon/src/prompts/media-contract.ts
```diff
@@ -255,13 +256,15 @@ substitution. Do not silently fall back.
...
-   descriptions like "warm Mandarin narrator" as \`--voice\`; omit the
-   flag instead unless you have a real id.
+    tells you the user's pre-selected model, aspect, length, voice, audio
+    kind, etc. Treat those as authoritative defaults — only override if
+    the user's chat message explicitly contradicts them.
+    For \`minimax-tts\`, \`voice\` must be a valid MiniMax \`voice_id\`
+    (example: \`male-qn-qingse\`). Do not pass natural-language voice
+    descriptions like "warm Mandarin narrator" as \`--voice\`; omit the
+    flag instead unless you have a real id.
+    \`language\` enables pronunciation boost for specific languages
```
mrcfps, line 266: **Blocking:** Please use MiniMax’s actual Cantonese `language_boost` value in the agent-facing contract. This changed workflow rule tells agents that `Yue` is the Cantonese value, and the invocation example above says `Chinese, Yue` with a space, but MiniMax’s T2A schema enumerates `Chinese,Yue` as the Cantonese boost value, not bare `Yue`. Because this prompt is the primary caller contract, an agent following the new guidance would send `language_boost: \"Yue\"`, which can fail provider validation or miss the intended boost and breaks the PR’s stated Cantonese path. Please change the new examples in this file to `Chinese,Yue` (no space), or add runtime translation/validation that maps `Yue` to `Chinese,Yue` before the MiniMax request is built.

apps/daemon/src/server.ts
mrcfps: **Non-blocking:** Please keep the new `language` field typed and trimmed before it reaches the provider request. This changed line forwards `req.body?.language` directly, and `renderMinimaxTTS` later emits any truthy `ctx.language` as `language_boost`; unlike `voice`, there is no string guard or `.trim()`. That means a local REST client can queue non-string JSON values, and whitespace-only or space-padded CLI/API values get sent to MiniMax instead of behaving like an omitted option. A small fix is to mirror the numeric guards here (`language: typeof req.body?.language === \"string\" ? req.body.language : undefined`) and/or normalize in `generateMedia` / `renderMinimaxTTS` with `const languageBoost = typeof ctx.language === \"string\" ? ctx.language.trim() : \"\";` before conditionally adding `language_boost`.

mrcfps (changes requested):
@terencewlc thanks for the quick follow-up and for wiring the language boost path through the CLI, server, and MiniMax request. I found one blocking issue with the documented Cantonese value agents will follow, plus one small normalization follow-up; details inline.


apps/daemon/src/server.ts
mrcfps: Looper checked this thread against head `6ba59dc993eec0d8b5a203997b8fce839aa786fc`. The requested change appears objectively addressed by apps/daemon/src/server.ts:3367 now guards language with typeof string before enqueueing, and apps/daemon/src/media.ts:1359-1365 trims ctx.language and only emits language_boost when non-empty., so I’m resolving this thread. Reopen if this still needs discussion.

mrcfps (approved):
@terencewlc I reviewed the final diff for the MiniMax TTS language boost change, including the CLI flag path, daemon request handoff, provider request construction, input normalization, and the agent-facing media contract updates. The prior Cantonese value and input-guard concerns appear addressed in this head, and I did not find any remaining actionable correctness, safety, test, or maintainability issues in the changed ranges. Thanks for the thoughtful follow-up and for making the Cantonese TTS flow easier to use. 🙌


leilei926524-tech:
Hi @terencewlc — welcome to Open Design contributors. 🎉

Your first merged PR (#773) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #783, opened 2026-05-07
feat(web): add batch delete for selected design files (#771)

### LINKED ISSUE #771: Option to delete design files.
Provide an option to delete multiple files after they're selected in Design files

yinjialu:
## Summary

Closes #771.

The Design Files multi-select toolbar already had **Download {n} as ZIP**; this PR completes the surface by adding a parallel **Delete {n}** action so users don't have to delete files one at a time from each row's overflow menu.

## Changes

- **`DesignFilesPanel.tsx`** — new `onDeleteFiles` prop + danger-styled toolbar button beside Download. `handleBatchDelete()` raises the intent and clears selection; deletion semantics stay in `FileWorkspace`.
- **`FileWorkspace.tsx`** — `handleDeleteMany()`:
  - one `confirm()` dialog up-front (not N)
  - sequential per-file DELETE via the existing `deleteProjectFile` primitive
  - batched tab/sketch cleanup in a single `onTabsStateChange`
  - if `activeTab` is in the deleted set, falls back to last remaining tab or `DESIGN_FILES_TAB`
  - partial-failure surfaces via `alert()` with a count
- **i18n** — 3 new keys across all 16 locales:
  - `workspace.deleteSelectedFilesConfirm` (confirm dialog)
  - `workspace.deleteSelectedFilesPartial` (failure alert)
  - `designFiles.deleteSelected` (button label)

  Confirm + button localized for **en, zh-CN, zh-TW, de, fr, es-ES, ja, ko, ru, pt-BR, pl, hu, tr, uk, ar, fa**. Native speakers welcome to refine wording in follow-up i18n PRs.
- **`index.css`** — `.df-head .df-actions button.danger` styles using existing `--red` / `--red-bg` tokens, mirroring the row-popover delete affordance.

## Why this approach

- **No new daemon endpoint.** The maintainer's comment on the issue framed it as a "focused enhancement on top of the existing multi-select infrastructure"; client-side loop reuses the existing `DELETE /api/projects/:id/raw/*` primitive.
- **Single confirmation, not N.** Calling the existing single-file `onDeleteFile` in a loop would pop a confirm dialog per file — bad UX. New `onDeleteFiles` prop separates the batch intent and lets `FileWorkspace` own one upfront confirm + atomic tab cleanup.

## Test plan

- [x] `pnpm guard` ✅
- [x] `pnpm typecheck` ✅
- [x] `pnpm --filter @open-design/web test` ✅ (45 files / 335 tests, including `tests/i18n/locales.test.ts` which gates locale dictionary alignment)
- [x] Manual: select 2+ files in Design Files → click Delete N → confirm → files removed, open tabs of deleted files closed
- [x] Manual: delete the file currently shown in the active tab → falls back to remaining tab or Design Files

## Out of scope

- Undo / soft-delete (single-file delete doesn't have it either; keeps parity)
- "Deleting…" loading state (single-file delete doesn't have it either; deletions are typically instant)
- Adding a `trash` icon to `Icon.tsx` (kept patch surface tight; danger-color text matches the existing row-popover delete)


apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -160,6 +162,13 @@ export function DesignFilesPanel({
     });
   }
 
+  function handleBatchDelete() {
+    const fileList = [...selected];
+    if (fileList.length === 0) return;
+    onDeleteFiles(fileList);
+    setSelected(new Set());
```
mrcfps: This clears the multi-selection immediately after calling `onDeleteFiles`, but the actual delete flow in `FileWorkspace.handleDeleteMany` can still exit without deleting anything when the user cancels the confirmation dialog. In that main cancel/retry path, the user loses the selected file set and has to reselect everything before trying again. Consider making `onDeleteFiles` return whether any deletion was confirmed/succeeded, then only clear `selected` after that success signal (or move selection clearing into a callback from the parent after deletion completes).

apps/web/src/components/FileWorkspace.tsx
```diff
@@ -301,6 +301,39 @@ export function FileWorkspace({
...
+    if (names.length === 0) return;
+    if (!confirm(t('workspace.deleteSelectedFilesConfirm', { n: names.length }))) return;
+    const deleted: string[] = [];
+    const failed: string[] = [];
+    for (const name of names) {
+      const ok = await deleteProjectFile(projectId, name);
+      if (ok) deleted.push(name);
+      else failed.push(name);
+    }
+    if (deleted.length === 0) return;
```
mrcfps: When every DELETE request fails, `deleted.length` stays `0` and this early return skips the partial-failure alert below. After the user confirms the destructive action, the UI becomes a silent no-op even though `failed` contains all selected files, so API outages or permission/path failures look like a broken button. Please surface the existing `workspace.deleteSelectedFilesPartial` alert before returning when `failed.length > 0`, or restructure the function so the failure alert runs regardless of whether any file was successfully deleted.

mrcfps (commented):
@yinjialu thanks for adding the batch delete path and keeping it aligned with the existing single-file delete flow. I found two small UX/error-handling gaps that are worth addressing, but neither needs to block the feature. 🙏


apps/web/src/components/FileWorkspace.tsx
lefarcen: **P2** If every selected file fails to delete (`deleted.length === 0` but `failed.length > 0`), this early return prevents the failure alert from showing. User clicks Delete → confirms → nothing happens. Move the failure alert before this return, or handle the all-fail case explicitly with `deleteSelectedFilesPartial`.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -160,6 +162,13 @@ export function DesignFilesPanel({
     });
   }
 
+  function handleBatchDelete() {
+    const fileList = [...selected];
+    if (fileList.length === 0) return;
+    onDeleteFiles(fileList);
```
lefarcen: **P2** Panel clears the full selection immediately after calling `onDeleteFiles`, even if the user cancels the confirmation in FileWorkspace, all deletes fail, or only some files fail. This prevents retrying failed items.

Concrete fix: don't clear selection here. Let the existing `files` pruning effect remove successfully deleted files after refresh, leaving canceled/failed selections intact so the user can retry.

apps/web/src/components/FileWorkspace.tsx
```diff
@@ -301,6 +301,39 @@ export function FileWorkspace({
     }
   }
 
+  async function handleDeleteMany(names: string[]) {
```
lefarcen, line 304: **P3** Batch delete has no in-flight guard. Since the operation takes multiple awaits, tab changes or another delete action during the loop can be overwritten by the stale `persistedTabs`/`tabsState.active` captured at click time. Consider adding a local `deleting` state / disabling the batch action during the operation, or reading the latest parent state for tab updates.

lefarcen (changes requested):
Hey @yinjialu! Thanks for the PR — the batch-delete UX is a clear win, and the i18n coverage + test plan are solid. 👍

Found 2 must-fix issues (P2) around error handling + selection clearing that would confuse users in failure scenarios:

1. **All-fail silent case** — if every DELETE fails, handleDeleteMany returns before showing any alert
2. **Selection clears too early** — DesignFilesPanel clears the selection immediately after calling onDeleteFiles, even if user cancels the confirm or all deletes fail

Also noted 2 nice-to-haves (P3): concurrency guard during the loop, and tests for the new batch behavior. See inline.

lefarcen (dismissed):
@yinjialu mrcfps already covered the 2 key issues (selection clearing + all-fail alert) in his non-blocking review. Please address those two points, then this will be ready. Thanks for the thorough i18n coverage! 🙏

yinjialu:
@lefarcen @mrcfps thanks for the careful review. Pushed [`cf26030`](https://github.com/nexu-io/open-design/pull/783/commits/cf26030) addressing all three points + CI:

**P2 — all-fail silent path** (`FileWorkspace.tsx:314`)
Removed the early `return` on `deleted.length === 0`. Tab/sketch cleanup now runs only when at least one file was successfully deleted, but the `workspace.deleteSelectedFilesPartial` alert fires whenever `failed.length > 0`, so an outage / 4xx-everywhere scenario surfaces an explicit failure instead of a silent no-op.

**P2 — selection preservation** (`DesignFilesPanel.tsx:168`)
Stopped clearing `selected` after `onDeleteFiles`. The existing `useEffect` that prunes selections against the current `files` list naturally drops successfully-deleted names on the next refresh; cancelled / failed names stay selected so the user can retry without rebuilding the multi-select. This matches the suggestion in your inline comment.

**P3 — in-flight guard** (`DesignFilesPanel.tsx`)
Added a local `deleting` state. The batch Delete button is `disabled={deleting}` while a delete pass is running, and `handleBatchDelete` early-returns if `deleting` is already true, so a stale click can't re-fire with overwritten `persistedTabs` / `tabsState.active` captured at the original click. To support `await`, the `onDeleteFiles` prop type was relaxed to `Promise<void> | void`.

**CI fix — `id.ts` coverage**
The CI failure on the previous push was unrelated to my changes — `apps/web/src/i18n/locales/id.ts` was added to `main` in fcc37c6 after this branch was cut, and that locale file is missing 22 keys. To unblock CI without rebasing in unrelated noise, I rebased onto current `main` and added:
- The 3 new batch-delete keys, translated to Indonesian
- The 19 pre-existing `settings.test*` / `settings.cliEnvCodexBin` / `sketch.textModalTitle` keys, English fallback only

Native Indonesian copy for the latter set is out of scope for this PR and best handled in a follow-up i18n PR by an Indonesian-speaking contributor.

**Heads up on the force-push**: I rebased onto current `main` so the PR diff stays clean — without the rebase, GitHub showed `id.ts` as a 993-line new file (since the branch's merge-base predated `id.ts`'s addition on main).

Tests added are out of scope here per `CONTRIBUTING.md`'s "one concern per PR" — happy to add them in a follow-up if you'd prefer.

mrcfps:
@yinjialu I'm holding off on generating review comments for nexu-io/open-design#783 because this pull request has merge conflicts right now.

yinjialu:
Rebased onto current `main` (`b95ba5e`) to resolve the conflict in `id.ts`.

While I was force-pushed `cf26030` queued, [a8418ac](https://github.com/nexu-io/open-design/commit/a8418ac) landed on main and added proper Indonesian translations for the 19 `settings.test*` / `cliEnvCodexBin` / `textModalTitle` keys that I'd put in as English fallback. So my fallback rows are now redundant — kept main's Indonesian version, dropped my English rows. `id.ts` in this PR's diff is now back to a clean +3 lines (the 3 new batch-delete keys, translated to Indonesian), matching the shape of every other locale in this PR.

`pnpm guard` ✅, `pnpm typecheck` ✅, `pnpm --filter @open-design/web test` ✅ (377 tests).

lefarcen (approved):
Hey @yinjialu, thanks for the careful iteration! 🎉

Verified that all three P2 findings from the previous review are addressed in the current diff:

- **All-fail alert** (`FileWorkspace.tsx:334`): The `workspace.deleteSelectedFilesPartial` alert now fires whenever `failed.length > 0`, regardless of whether any file succeeded. An API outage or permission failure no longer appears as a silent no-op.
- **Selection preservation** (`DesignFilesPanel.tsx:168`): `handleBatchDelete` no longer clears `selected` after calling `onDeleteFiles`. Cancelled / failed deletions leave the user's selection intact for retry, and the existing `useEffect` prunes successfully-deleted names on the next refresh.
- **In-flight guard** (`DesignFilesPanel.tsx:69+165`): `deleting` state added; the Delete button is `disabled={deleting}` while a pass is running, and `handleBatchDelete` early-returns if `deleting` is already true, so overlapping clicks can't re-fire with overwritten captured state.

The rebase onto current `main` + addition of the 3 new batch-delete keys to `id.ts` (Indonesian locale) is clean and unblocks CI. Ready from my side.

**Heads up on mrcfps**: I see @mrcfps's review is on the older head (`f1c2a10`). His COMMENTED review was non-blocking and covered the same 2 P2 issues. You've addressed both in this push — he may want to take one more look at the current head before merge, but I don't see any new blockers here.

mrcfps (approved):
@yinjialu I reviewed the current head and verified the batch delete implementation, including the single confirmation flow, partial/all-failure handling, selection preservation for retry, tab/sketch cleanup, and the added locale keys/styles. The follow-up changes address the earlier UX/error-handling concerns, and I don't see any actionable issues in the changed ranges. Thanks for the careful iteration and thorough i18n coverage! 🙏


yinjialu:
@mrcfps @lefarcen thanks for the reviews. The current head is approved but CI is blocked with “1 workflow awaiting approval” because this is a fork PR. Could one of you approve/run the workflow when convenient?

leilei926524-tech:
Hi @yinjialu — welcome to Open Design contributors. 🎉

Your first merged PR (#783) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #797, opened 2026-05-07
fix: clear stale upload failure banner when previewing files

### LINKED ISSUE #786: Clear stale upload failure banner when previewing existing files in Design Files
## Summary

After a file upload fails in Design Files, the upload failure message continues to remain at the top of the preview area even when previewing an existing file from the file list.

This makes the preview view feel polluted by stale error state that is no longer relevant to the current file being viewed.

## Current behavior

- A file upload fails in the Design Files area.
- An upload failure message is shown at the top of the page.
- The user then previews a file that already exists in the Design Files list.
- The upload failure banner/message is still visible above the preview.
- The preview surface continues to show stale upload error state even though the user is no longer focused on the failed upload flow.

## Expected behavior

Upload failure messaging should not remain pinned over unrelated file preview flows.

Suggested improvements:

- Clear or dismiss the upload failure banner when switching into an existing file preview.
- Scope upload error state to the upload/list context rather than keeping it globally attached to the preview area.
- Show the error only where it is relevant, or provide a dismiss action if it needs to persist.
- Ensure previewing an existing file restores a clean viewing state.

## Steps to reproduce

1. Open Design Files.
2. Trigger a failed file upload.
3. Observe the upload failure message.
4. Open or preview a file that already exists in the file list.
5. Notice that the upload failure message still remains at the top of the preview.

## Notes

A screenshot is attached showing an existing file in preview while the previous upload failure message is still displayed above it.

## Why this matters

This is a state-management and UX issue. Stale error messages reduce confidence in the preview flow and make the interface feel cluttered or broken, especially when the current task has changed from uploading to file viewing.

Image

mvanhorn:
## What changed

`apps/web/src/components/FileWorkspace.tsx`:
- Clear `uploadError` inside `openFile(name)` so navigating to a file from the Design Files list dismisses the banner.
- Scope the banner render to `activeTab === DESIGN_FILES_TAB` so stale upload errors no longer bleed into the preview surface.

`apps/web/tests/components/FileWorkspace.test.tsx`:
- Pin that the default render contains no `viewer-empty` element when there is no upload error.

Closes #786

## Why

The repro in #786 is exactly: trigger an upload failure, then preview an existing file. The banner stays pinned over the preview because the `uploadError` state is set on upload failure (`FileWorkspace.tsx:206`) and rendered unconditionally inside `<div className="ws-body">` regardless of which tab is active. The two-line surface fix matches the issue's "Scope upload error state to the upload/list context" suggestion: state-clearing in `openFile()` handles the active dismissal, and the `activeTab === DESIGN_FILES_TAB` guard makes the banner structurally impossible to render outside the list view.

## Verification

`apps/web` vitest: `tests/components/FileWorkspace.test.tsx` 14/14 pass (one new). `pnpm --filter=@open-design/web typecheck` clean.


apps/web/src/components/FileWorkspace.tsx
```diff
@@ -472,7 +473,9 @@ export function FileWorkspace({
         ) : null}
       </div>
       <div className="ws-body">
-        {uploadError ? <div className="viewer-empty">{uploadError}</div> : null}
+        {uploadError && activeTab === DESIGN_FILES_TAB ? (
```
lefarcen: **P2** — The banner is still rendered whenever `activeTab === DESIGN_FILES_TAB`, but Design Files in-panel preview also lives under that same tab. A single-click row preview in `DesignFilesPanel` only calls `setPreview(f.name)`, not `openFile`, so an upload failure banner remains visible while the user previews a file in the right-side preview pane.

Suggestion: move the upload-error banner into the list/upload portion of `DesignFilesPanel`, hide it when a preview is active, or provide an explicit preview/open callback that clears the upload error on row preview as well as tab open.

apps/web/tests/components/FileWorkspace.test.tsx
```diff
@@ -22,6 +22,22 @@ describe('FileWorkspace upload input', () => {
     expect(markup).not.toContain('accept=');
   });
 
+  it('does not render an upload failure banner without an upload error', () => {
```
lefarcen: **P3** — The new test does not exercise the regression or the fix; it would pass before this PR because `uploadError` defaults to `null`. It also asserts on the generic `viewer-empty` class, which is used for unrelated empty/loading states.

Suggestion: replace with an interactive test that mocks `uploadProjectFiles` to fail, triggers an upload/drop, verifies the banner appears, then previews/opens an existing file and verifies the specific upload error text is gone from the preview surface.

lefarcen (commented):
Hey @mvanhorn! Thanks for the PR — the state-clearing + tab-scoped approach aligns well with the issue's suggestion. Found two things to consider:

**Lens A review (code correctness):** The fix covers the main repro (upload failure → open file from list via tab), but there's a gap in Design Files in-panel preview + the new test doesn't exercise the regression. See inline for the specific scenarios and test suggestion.

Overall looks good; these are non-blocking suggestions to tighten the fix scope.

apps/web/src/components/FileWorkspace.tsx
```diff
@@ -472,7 +473,9 @@ export function FileWorkspace({
...
+          <div className="viewer-empty">{uploadError}</div>
```
mrcfps: The `activeTab === DESIGN_FILES_TAB` guard still leaves the upload error mounted while the Design Files panel is showing its in-panel preview: a single-click row preview in `DesignFilesPanel` only updates that panel's `preview` state, so `activeTab` remains `DESIGN_FILES_TAB` and this `viewer-empty` banner continues to render above the preview. That means the core flow described in this PR — upload failure, then preview an existing file — is still visible for the primary in-panel preview path. Please either move the upload-error banner into the list/upload portion of `DesignFilesPanel`, clear it when `setPreview(f.name)` runs, or hide it when the Design Files panel has an active preview, and add an interactive regression test that creates an upload error then previews/opens an existing file and asserts the specific error text is gone from the preview surface.

mrcfps (changes requested):
@mvanhorn thanks for the focused bug fix here — clearing stale upload state when users move into file viewing is the right direction. I found one remaining primary preview path where the stale banner can still persist, so I’m requesting a small follow-up to make the fix cover the PR’s stated flow.


lefarcen:
Hi @mvanhorn! 👋 Just checking in — are you still working on addressing the in-panel preview feedback from @mrcfps? If you're stuck or need any help, feel free to ask. No rush, just wanted to make sure this isn't blocked on anything. Thanks!

lefarcen:
Hi @mvanhorn, gentle follow-up 👋 — just wanted to check if you saw @mrcfps's review feedback. If this PR isn't a priority right now, that's totally fine! We can always pick it up later or close it for now. Let us know either way when you have a moment. Thanks!

mrcfps:
@mvanhorn I'm holding off on generating review comments for nexu-io/open-design#797 because this pull request has merge conflicts right now.

lefarcen (commented):
Hey @mvanhorn! 👋 Solid work addressing the in-panel preview gap — moving the banner into `DesignFilesPanel` with `uploadError && !preview` directly solves @mrcfps's finding, and the new interactive tests are exactly the regression coverage this fix needed.

## ✅ Previous findings resolved
- **P2 (in-panel preview gap)**: Fixed — banner now only renders when `!preview`, so the in-panel preview path (click design-file-row → `setPreview(name)`) won't show stale upload errors.
- **P3 (test coverage)**: Fixed — both new tests are interactive regression tests that would fail without your fix.

## Minor suggestions (non-blocking)
- **P2** — `DesignFilesPanel.tsx:23-24`: prop type has both `?` and `= null` default — pick one (either `uploadError?: string | null` with no default, or `uploadError: string | null` without `?`). Works but redundant.
- **P3** — `FileWorkspace.tsx:201`: `${detail}` exposes raw error message; consider sanitizing to avoid leaking internal paths/stack traces (though upload errors are typically benign).
- **P3** — Test assumes `design-file-preview` testid exists; verify `DesignFilesPanel` renders it (probably fine, just flagging potential brittleness).

Once you resolve the merge conflicts with main (mrcfps bot already pinged you on that), this is ready from my side. Nice iteration! 🎉

mvanhorn:
@mrcfps @lefarcen thanks for the in-panel preview catch and the test-shape feedback. Pushed 901f3ec5.

What changed:

- Banner now lives inside DesignFilesPanel and is hidden whenever the in-panel preview is active (`uploadError && !preview`). The single-click row preview path no longer leaves the banner above the preview.
- Outer FileWorkspace keeps a small fallback banner that fires only when `activeTab !== DESIGN_FILES_TAB`. This preserves partial-upload visibility (codex-connector P2): a partial upload that opens the last successful file flips to a viewer tab and the failure surfaces there.
- Wrapped `uploadProjectFiles` in a try/catch so a thrown error surfaces a banner instead of disappearing.
- Replaced the `viewer-empty` assertion with two interactive vitest cases that exercise the regression directly: (1) mock-fail upload, banner visible, preview, banner hidden, close, banner back, dismiss, banner gone; (2) partial-upload uploaded+failed, banner shows on the viewer surface.
- Added a `df-upload-banner` class and stable test ids `upload-error-banner` / `upload-error-dismiss` so future tests do not rely on the generic `viewer-empty` class.

Apologies for the slow turnaround.

mrcfps (dismissed):
@mvanhorn I re-checked the updated upload-error flow in `FileWorkspace` and `DesignFilesPanel`, including the in-panel preview path, the viewer-surface fallback for partial upload failures, and the new interactive regression coverage in `FileWorkspace.test.tsx`. This revision cleanly addresses the stale-banner behavior from the earlier review pass.

Thanks for iterating on the follow-up and tightening the tests here 🙌


mrcfps (approved):

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -600,6 +604,20 @@ export function DesignFilesPanel({
           )}
         </div>
         <div className="df-body">
+          {uploadError && !preview ? (
```
lefarcen, line 607: **P2 — Upload-error-while-preview-open hidden**

`uploadError && !preview` hides all upload failures while the right-side preview is active, but the Design Files upload controls/drop target are still usable in that state. If an upload fails while preview is open, `uploadError` is set, `activeTab` remains the Design Files tab, and the FileWorkspace fallback is also gated off, so no failure UI appears until preview is closed. This can make a current upload failure look like success.

Consider tracking whether the error predates the preview, clearing stale errors on preview open while still rendering newly generated errors, or surfacing upload failures outside the preview-hidden region.

lefarcen (changes requested):
Re-reviewed b5563a34 (post-merge head). The prior P2/P3 findings from 901f3ec5 remain addressed — upload error is correctly hidden during in-panel preview for existing errors, and the fallback banner still handles partial uploads. Found one remaining P2 in the same gating logic area — see inline comment.

PR #798, opened 2026-05-07
feat(diagnostics): add one-click log export from Settings → About

lefarcen:
## Summary

- Adds **Settings → About → Export diagnostics**, a one-click button that bundles daemon/web/desktop logs, machine info, and recent macOS crash reports into a zip the user can attach to a bug report.
- Two delivery paths so users still get logs when the daemon is wedged:
  - **Browser** hits a new `GET /api/diagnostics/export` daemon endpoint and triggers a download.
  - **Electron** uses a preload-bridged IPC, the native save dialog, and reveals the saved file in Finder/Explorer when done. The macOS Help menu also exposes "Export Diagnostics…" as a fallback that doesn't depend on the daemon being alive.
- Packaging + redaction lives in a new `@open-design/diagnostics` package shared by daemon and desktop.
- Redaction masks sensitive JSON keys (`token|password|secret|key|dsn|authorization|cookie`), URL query secrets, and replaces the current user's home path with `/Users/<USER>`.
- Renderer-process `console.error` / `console.warn` now also lands in `logs/desktop/renderer.log` so UI bugs that previously only surfaced in DevTools end up in the export.
- 13 locale files updated with the 6 new strings.

## Test plan

- [ ] `pnpm typecheck` (passes locally)
- [ ] `pnpm test` (passes locally — adds 10 new tests in `packages/diagnostics`)
- [ ] `pnpm build` (passes locally)
- [ ] In a browser tab, open Settings → About → click **Export diagnostics** → verify the zip downloads and contains `logs/{daemon,web,desktop}/latest.log` + `summary/manifest.json` + `summary/machine-info.json`.
- [ ] In the Electron shell, open Settings → About → click **Export diagnostics** → native save dialog appears → after saving, Finder/Explorer reveals the zip.
- [ ] Open `logs/...` files inside the zip and confirm sensitive values are redacted (search for `[REDACTED]`).
- [ ] Force a renderer error in DevTools (`throw new Error('test')`) → re-export → confirm `logs/desktop/renderer.log` captured it.
- [ ] On a Mac with a recent Electron crash, confirm `crash-reports/*.ips` show up in the zip.


mrcfps:
@lefarcen I'm holding off on generating review comments for nexu-io/open-design#798 because this pull request has merge conflicts right now.

packages/diagnostics/tests/redaction.test.ts
```diff
@@ -0,0 +1,64 @@
+import { describe, expect, it } from "vitest";
```
mrcfps, line 1: The new diagnostics package puts Vitest files under `packages/diagnostics/src/` (`redaction.test.ts` here, and the same pattern in `packages/diagnostics/src/zip.test.ts`). This violates the repository test-layout rule in `AGENTS.md`/`scripts/guard.ts`, which requires tests under `apps/`, `packages/`, and `tools/` to live in a sibling `tests/` directory instead of under `src/`; with these changed paths, the repo-level `pnpm guard` check will reject the PR. Please move these tests to `packages/diagnostics/tests/redaction.test.ts` and `packages/diagnostics/tests/zip.test.ts` (preserving any useful source-relative organization), then update `packages/diagnostics/vitest.config.ts` to include `tests/**/*.test.ts` so the package tests still run.

mrcfps (changes requested):
@lefarcen thanks for adding the diagnostics export path and sharing the test coverage here — the feature direction is useful. I found one merge-blocking repository-layout issue that needs to be fixed before this can pass the standard guard checks.




apps/daemon/src/server.ts
```diff
@@ -1480,6 +1482,11 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     }
   });
 
+  app.get(
+    DIAGNOSTICS_EXPORT_PATH,
```
mrcfps, line 3483: This new diagnostics export route is registered without the local-daemon guard used by other private daemon APIs. The handler packages recent daemon/web/desktop logs, crash reports, machine info, and paths, but the global `/api` middleware only rejects disallowed browser `Origin` headers and explicitly allows non-browser clients with no `Origin`; if the daemon is started with the documented non-loopback bind host escape hatch, a peer on the network can request this URL directly and download the diagnostics archive. That makes the main diagnostics flow expose sensitive support data outside the local machine. Please add the same local access check here (for example `app.get(DIAGNOSTICS_EXPORT_PATH, requireLocalDaemonRequest, createDiagnosticsExportHandler(...))`, or an equivalent `isLocalSameOrigin(req, resolvedPort)` check inside the route) before streaming the zip.

mrcfps (changes requested):
@lefarcen thanks for putting together the diagnostics export flow and for covering both browser and Electron paths. I found one blocking security issue around the new daemon download route that should be tightened before merge; the rest of the reviewed changes look directionally solid. 🙏




apps/daemon/src/diagnostics-export.ts
```diff
@@ -0,0 +1,121 @@
...
+            runtimeAvailable: options.runtime != null,
+            sourceTag: options.runtime?.source ?? null,
+            mode: options.runtime?.mode ?? null,
+            base: options.runtime?.base ?? null,
+            projectRoot: options.projectRoot,
+          },
+        },
+        sources,
+        redaction: { username },
+        crashReports: {
```
lefarcen, line 115: **P1** — `matchSubstrings: ['Electron']` will match crash reports from ANY Electron app on the host (VSCode, Slack, etc.). This leaks unrelated app data into the diagnostics zip. Restrict to Open Design-owned process names: `['Open Design', 'open-design']`.

packages/diagnostics/src/redaction.ts
```diff
@@ -0,0 +1,54 @@
+const SENSITIVE_KEY_RE = /token|password|secret|key|dsn|authorization|cookie/i;
+
+const URL_QUERY_SECRET_RE = /([?&#])(token|password|secret|key|dsn|api[_-]?key|auth)(=)([^&\s#"']*)/gi;
+
+// Catch loose key=value pairs in log lines that aren't inside a URL — e.g. env
+// dumps, command-line args, or json-line meta. The leading boundary stops it
+// from matching mid-identifier (e.g. `not_a_token`).
+const BARE_SECRET_RE = /(^|[\s,;])(token|password|secret|api[_-]?key|auth(?:orization)?)(=|:\s*)([^\s,;"']+)/gi;
```
lefarcen: **P1** — Bearer-token format `Authorization: Bearer sk-...` only redacts `Bearer`, leaving `sk-...` visible. Also missing `access_token=` / `refresh_token=` patterns. Add these to the redaction regex to close the gaps.

packages/diagnostics/src/sources.ts
```diff
@@ -0,0 +1,118 @@
...
+  /** Reason the file is missing or unreadable. */
+  error?: string;
+}
+
+async function readMaybeTail(absolutePath: string, tailBytes: number | undefined): Promise<{ text: string; bytes: number }> {
+  if (tailBytes == null || tailBytes <= 0) {
+    const buf = await readFile(absolutePath);
+    return { text: buf.toString("utf8"), bytes: buf.byteLength };
+  }
+  const info = await stat(absolutePath);
```
lefarcen, line 34: **P2** — `readFile(absolutePath)` loads the entire file before slicing `start` bytes. For large logs (> 4 MB), this makes diagnostics export slow or memory-heavy. Consider using `fs.createReadStream({ start, end })` to only read the tail bytes.

lefarcen (commented):
@lefarcen Layout issue fixed — tests are now correctly in `packages/diagnostics/tests/`. Found 2 P1 items and 1 P2 below:

**P1 crash-report overmatch** — current substring `"Electron"` in `crashReports.matchSubstrings` will package crash reports from ANY Electron app on the host (VSCode, Slack, etc.), leaking unrelated data into the diagnostics zip. Restrict to Open Design-owned process names (e.g., `["Open Design", "open-design"]`).

**P1 redaction gaps** — bearer-token format `Authorization: Bearer sk-...` only redacts the word Bearer, leaving `sk-...` in exported logs. Also missing `access_token=` / `refresh_token=` patterns. This is risky for a log export feature with external distribution.

**P2 memory usage** — `tailBytes` still loads the full file before slicing. Large logs can make the diagnostics export slow or memory-heavy despite the 4 MB tail intent.

See inline for specifics.

lefarcen (commented):
Thanks for moving the tests to the correct layout — that addresses mrcfps's prior finding.

I found a few additional issues scanning the current head:

## P1 — Crash report over-collection (privacy)

`packages/diagnostics/src/sources.ts:83` matches crash reports with the substring `"Electron"`. This will package crash reports from **any** Electron app on the user's system (VS Code, Slack, etc.), not just Open Design. Leaking other apps' crash data is a privacy issue.

**Suggested fix**: Restrict to Open Design-owned process names only (e.g., `"Open Design"`, `"open-design"`, and the actual bundle ID if available). Drop the generic `"Electron"` substring.

## P1 — Incomplete bearer token redaction

`packages/diagnostics/src/redaction.ts:8` misses common bearer-token formats:
- `Authorization: Bearer sk-...` → redacts `Bearer` but leaves `sk-...` in the exported logs
- `access_token=xxx` / `refresh_token=xxx` in plain text → not covered by current patterns

**Suggested fix**: Add patterns for `Bearer <token>` specifically, and extend `BARE_SECRET_RE` to include `access_token` / `refresh_token`.

## P2 — Memory inefficiency in tailBytes

`packages/diagnostics/src/sources.ts:34` reads the entire file with `readFile()` before slicing. For large logs (multi-GB in long-running daemons), this can make diagnostics export slow/memory-heavy despite the 4 MB tail limit.

**Suggested fix**: Use `fs.createReadStream` with `start` offset, or `fs.open` + `fs.read` to read only the tail directly.

lefarcen:
Addressed all 3 findings from the review in commit `5e5ec25`:

**P1 — Crash report overmatch** ✅
Dropped the `"Electron"` substring from `crashReports.matchSubstrings` in both `apps/daemon/src/diagnostics-export.ts` and `apps/desktop/src/main/diagnostics.ts`. We now only collect crash reports whose filenames contain `"Open Design"` or `"open-design"`, so VS Code / Slack / other Electron-app crash reports stay out of the bundle.

**P1 — Bearer / access_token / refresh_token redaction** ✅
- Added a new `HTTP_AUTH_SCHEME_RE` that matches `Bearer|Token|Basic <credential>` as one unit and runs **before** the bare-secret pass, so the actual credential after `Authorization: Bearer ` no longer leaks.
- Extended `URL_QUERY_SECRET_RE` and `BARE_SECRET_RE` with `access_token`, `refresh_token`, and `id_token` (long-first ordering so they win over plain `token`).
- Added 2 new redaction tests covering both forms (12 tests total in `packages/diagnostics/tests/redaction.test.ts`, all passing).

**P2 — tail-bytes memory** ✅
`readMaybeTail` in `packages/diagnostics/src/sources.ts` now uses `fs.open` + `read(buffer, 0, tailBytes, start)` to read only the trailing window directly, instead of loading the whole file with `readFile()` and then slicing. Multi-GB daemon logs no longer get materialised into memory.

`pnpm typecheck`, `pnpm guard`, and `packages/diagnostics` tests all pass locally.

apps/daemon/src/server.ts
mrcfps, line 3483: The new `GET /api/diagnostics/export` route is still registered directly with `createDiagnosticsExportHandler` instead of going through the local-daemon guard used by other private APIs. The handler streams a zip containing recent daemon/web/desktop logs, crash reports, machine info, and filesystem paths; meanwhile the global `/api` middleware allows non-browser clients with no `Origin` header. If the daemon is launched with the documented non-loopback bind-host escape hatch, a peer on that network can request this URL directly and download the diagnostics bundle. Please register this route with `requireLocalDaemonRequest` (for example, `app.get(DIAGNOSTICS_EXPORT_PATH, requireLocalDaemonRequest, createDiagnosticsExportHandler(...))`) or perform the equivalent `isLocalSameOrigin(req, resolvedPort)` check inside the handler before streaming the zip.

mrcfps (changes requested):
@lefarcen thanks for the follow-up fixes on the redaction, crash-report filtering, and log tailing pieces — those updates are helpful. I found one remaining blocking security issue in the daemon HTTP export path that still needs to be guarded before merge. 🙏




lefarcen (commented):
Nice work fixing the prior issues (crash over-collection, token redaction, memory limits) 👍

Agreeing with mrcfps's current-head finding: the diagnostics export route at `apps/daemon/src/server.ts:1486` still needs the local-daemon guard. All other routes that expose private data (connector operations, etc.) use `requireLocalDaemonRequest` middleware to block network access when daemon runs with non-loopback bind — this one should too.

Once that's added, this looks production-ready.

lefarcen:
Pushed `2ce4ec66` — addresses the remaining blocking finding and rebases on top of latest `main`.

**P1 — Diagnostics export route now gated by `requireLocalDaemonRequest`** ✅
`GET /api/diagnostics/export` is registered with the same `requireLocalDaemonRequest` middleware that protects connector / live-artifact routes, so the bundle (logs, host metadata, crash reports) is unreachable when the daemon binds to a non-loopback address (Tailscale, `0.0.0.0`, etc.).

**Rebase + integration**
- Rebased on top of `main` (54 commits forward).
- Integrated with the new `apps/desktop/src/main/preload.cts` that landed on main: kept the existing `electronAPI` (folder picker) and added a second `contextBridge.exposeInMainWorld('openDesignDesktop', { exportDiagnostics })` in the same file. Dropped the standalone `apps/desktop/src/preload/` directory.
- Resolved i18n locale conflicts (auto-merged with new sketch / settings keys).

Stale earlier findings already fixed in prior commits (re-confirmed against current head):
- `diagnostics-export.ts:103` no longer carries the generic `"Electron"` substring.
- Tests already live under `packages/diagnostics/tests/`, not `src/`.
- Bearer / `access_token` / `refresh_token` redaction + streaming `tailBytes` already in place since `5e5ec25`.

Locally: `pnpm typecheck`, `pnpm guard`, and `packages/diagnostics` tests all pass.

mrcfps (dismissed):
@lefarcen I reviewed the current head and verified the earlier blockers are addressed: the daemon export route is now local-only, the shared diagnostics package keeps the crash-report filtering, redaction, and bounded tail-reading fixes in place, and the desktop/browser entry points are wired consistently around the new export flow. I did not find any new actionable issues in the changed ranges — thanks for pushing this through. 🙏




lefarcen (commented):
Security finding addressed ✅ — the diagnostics export endpoint now correctly uses `requireLocalDaemonRequest` middleware at line 1666 before the handler. This matches the protection pattern used by other daemon private APIs.

The prior P1/P2 findings from earlier reviews (crash report filtering, token redaction, memory limits) were already fixed in commit 5e5ec25 per your 10:56Z comment.

LGTM from code review perspective. Ready for mrcfps's re-review.

packages/diagnostics/src/redaction.ts
```diff
@@ -0,0 +1,65 @@
...
+// dumps, command-line args, or json-line meta. The leading boundary stops it
+// from matching mid-identifier (e.g. `not_a_token`). Order alternatives long-
+// first so `access_token` wins over `token`.
+const BARE_SECRET_RE = /(^|[\s,;])(access_token|refresh_token|id_token|api[_-]?key|password|secret|token|auth(?:orization)?)(=|:\s*)([^\s,;"']+)/gi;
+
+// Bearer / Token / Basic auth schemes embed the secret after the scheme
+// keyword, so the BARE pattern above stops at the space before `Bearer` and
+// leaves the actual credential exposed. This pattern matches the scheme +
+// credential as one unit.
+const HTTP_AUTH_SCHEME_RE = /\b(Bearer|Token|Basic)\s+([A-Za-z0-9._\-+/=:]{4,})/g;
```
mrcfps: `HTTP_AUTH_SCHEME_RE` only matches title-cased `Bearer|Token|Basic`, but HTTP auth schemes are case-insensitive. A very common log line like `authorization: bearer sk-live-...` therefore falls through to `BARE_SECRET_RE`, which redacts only the `authorization: bearer` prefix and leaves the actual token in the exported zip. Because this feature is explicitly generating support bundles for external sharing, that is still a credential leak on the main path. Please make this auth-scheme regex case-insensitive (and ideally widen the token character class to the full bearer-token alphabet, e.g. `~`) and add a regression test alongside `packages/diagnostics/tests/redaction.test.ts:36-55` that covers lowercase `bearer`.

mrcfps (commented):
@lefarcen thanks for tightening the earlier diagnostics-export fixes and for carrying the shared package/test coverage forward. I re-reviewed the current head and found one remaining credential-redaction gap in the new diagnostics package; I'm leaving this as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR, but I do think this case should be fixed before merge.




packages/diagnostics/src/redaction.ts
lefarcen: **P1** — `HTTP_AUTH_SCHEME_RE` is case-sensitive, so lowercase or mixed-case authorization headers like `authorization: bearer sk-live-...` skip this regex and fall through to `BARE_SECRET_RE`, which redacts only the `authorization: bearer` prefix and leaves the actual token exposed. Please make this regex case-insensitive (add `i` flag: `/\b(Bearer|Token|Basic)\s+([A-Za-z0-9._\-+/=:~]{4,})/gi`) and add a test in `packages/diagnostics/tests/redaction.test.ts` covering lowercase `bearer`.

lefarcen (commented):
Re-reviewed at current head 5a653838. Earlier crash-filtering and memory-efficiency fixes remain in place. One P1 credential-redaction gap remains — see inline below.

lefarcen:
Rebased on top of latest `main` (196 commits forward) and pushed a small follow-up.

**Conflict resolution**
- `apps/daemon/src/server.ts` — main moved `startServer` to a strongly-typed `StartServerOptions` interface. Added `runtime?: DaemonRuntimeContext | null` to that interface and dropped the previous `startServerTyped` type-cast hack in `sidecar/server.ts` (no longer needed).
- `apps/daemon/src/sidecar/server.ts` — preserved main's `desktopPdfExporter` callback and threaded `runtime` into the same `startServer({...})` call.
- `apps/desktop/src/main/index.ts` — kept main's PR #974 import-token handshake and threaded the rendererLogPath computation after it.
- `apps/desktop/src/main/preload.cts` — kept main's PR #974 trust-boundary comment block alongside the diagnostics channel constant.
- `apps/desktop/src/main/runtime.ts` — merged `node:crypto` and `node:fs/promises` imports (kept `appendFile` for the renderer log); preserved main's `discoverDaemonUrl` and `registerDesktopAuthWithDaemon` options alongside `rendererLogPath`.
- `apps/web/src/i18n/{types.ts,locales/en.ts,locales/id.ts}` — keep-both: main added `feedback.*` keys, we keep adding `diagnostics.*` next to them.

**Follow-up commit (`8aa0b19d`)**
Per design feedback: swap the action order in the first-run `PrivacyConsentModal` so **Not now** is on the left and **Help improve** is on the right, matching the destructive-on-left / affirmative-on-right pattern used elsewhere. Left `PrivacySection` (the re-consent control inside settings) unchanged since it's a different surface.

Locally: `pnpm typecheck`, `pnpm guard`, and `packages/diagnostics` tests all pass.

lefarcen:
Pushed `d92569a4` and resolved the open review threads.

**Rebased on top of latest `main`** (16 commits forward). Conflict resolution: same keep-both pattern as before for i18n + desktop imports. Notably, `main` dropped the `feedback.*` keys from `apps/web/src/i18n/types.ts`, so I removed the matching block that earlier keep-both rounds had carried — the diagnostics keys remain.

**Fixed in this round (`d92569a4`)**
- `HTTP_AUTH_SCHEME_RE` is now **case-insensitive** (`/gi`). `Authorization: bearer …`, `bEaReR …`, and `Basic …` in mixed case are all redacted now. Added a dedicated regression test (`packages/diagnostics/tests/redaction.test.ts` is now 11 tests; full suite is 13).
- `nix/package-daemon.nix` now lists `packages/diagnostics` in the dependency-build target list. The CI Nix build was failing with `Cannot find module '@open-design/diagnostics'` because pnpm install only mints the symlink — the dist output that the package's `exports` point at isn't produced until each sibling's build script runs.

**Picked up two from earlier feedback**
- Swapped action order in the first-run `PrivacyConsentModal` — **Not now** on the left, **Help improve** on the right.

**Resolved threads** (all confirmed fixed against current head before resolving):
- redaction.ts Bearer / `access_token` / `refresh_token` (P2 codex + P1 looper) ✅
- redaction.ts case-insensitive HTTP auth (mrcfps + looper P1) ✅
- sources.ts streaming `tailBytes` (P2) ✅
- diagnostics-export.ts `'Electron'` crash-report substring (P1) ✅
- server.ts diagnostics route missing `requireLocalDaemonRequest` (mrcfps × 2) ✅
- tests/redaction.test.ts layout under `tests/` not `src/` ✅

Locally: `pnpm typecheck` (0 errors), `pnpm guard`, `packages/diagnostics` 13 tests all pass.

apps/daemon/package.json
```diff
@@ -34,6 +34,7 @@
   "dependencies": {
     "@modelcontextprotocol/sdk": "^1.0.0",
     "@open-design/contracts": "workspace:*",
+    "@open-design/diagnostics": "workspace:*",
```
mrcfps, line 38: Adding `@open-design/diagnostics` here introduces a new runtime workspace dependency, but this PR never teaches `tools-pack` to build/package that workspace package. The current head already shows the breakage in `Packaged windows smoke`: the packaged-app `npm install --omit=dev --no-package-lock` step fails with `404 Not Found - GET https://registry.npmjs.org/@open-design%2fdiagnostics`, because `tools/pack/src/workspace-build.ts` and the `INTERNAL_PACKAGES` lists in `tools/pack/src/{win,mac,linux}` still only include contracts/sidecar/daemon/web/desktop. That means packaged builds cannot install the daemon/desktop dependency graph from this branch. Please add `packages/diagnostics` to the workspace-build/tarball package lists and extend the corresponding `tools/pack` tests so the new shared package is copied into packaged artifacts on every platform.


mrcfps (commented):
@lefarcen thanks for pushing the diagnostics export through and tightening the earlier fixes. I found one remaining packaging issue in the current head; because this reviewer has already used its REQUEST_CHANGES quota on this PR, I’m leaving it as a comment, but the failing packaged Windows smoke suggests it still needs to be fixed before merge.




lefarcen (commented):
Hey @lefarcen, the diagnostics-export scope is clear from the Summary, including the browser/Electron paths and shared package. One PR-body cleanup before this goes further: please add the new-template `Surface area` checklist and a `Validation` section. For Surface area, this looks like at least UI, API/daemon endpoint, desktop/Electron, i18n keys, and package/dependency/build packaging; for Validation, move the existing command/manual checks there so release review can see exactly what passed.

packages/diagnostics/src/redaction.ts
```diff
@@ -0,0 +1,67 @@
...
+// first so `access_token` wins over `token`.
+const BARE_SECRET_RE = /(^|[\s,;])(access_token|refresh_token|id_token|api[_-]?key|password|secret|token|auth(?:orization)?)(=|:\s*)([^\s,;"']+)/gi;
+
+// Bearer / Token / Basic auth schemes embed the secret after the scheme
+// keyword, so the BARE pattern above stops at the space before `Bearer` and
+// leaves the actual credential exposed. This pattern matches the scheme +
+// credential as one unit. Case-insensitive: HTTP auth schemes are
+// case-insensitive per RFC 7235 §2.1, and we routinely see lowercase
+// `authorization: bearer …` in proxy / curl-style logs.
+const HTTP_AUTH_SCHEME_RE = /\b(Bearer|Token|Basic)\s+([A-Za-z0-9._\-+/=:]{4,})/gi;
```
mrcfps: `HTTP_AUTH_SCHEME_RE` still excludes `~` from the bearer-token character class, so RFC 6750-valid tokens such as `Authorization: Bearer abcd~efgh` are only partially redacted here. On the current head, that example becomes `Authorization: [REDACTED] [REDACTED]~efgh`, which leaves part of the credential in the support bundle. Because this feature is explicitly packaging logs for external sharing, that is still a credential leak on the main path. Please widen the auth-scheme regex to the full bearer-token alphabet (for example `[A-Za-z0-9._\-+/=:~]{4,}`) and add a regression test alongside `packages/diagnostics/tests/redaction.test.ts:36-81` covering a `~`-containing bearer token so the redactor consumes the whole secret.


mrcfps (commented):
@lefarcen thanks for pushing the diagnostics export through and for tightening the earlier packaging/runtime fixes. I re-reviewed the current head, re-ran the diagnostics and tools-pack tests locally, and found one remaining credential-redaction gap in the new diagnostics package. I’m leaving it as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR, but I do think this case should be fixed before merge.




packages/diagnostics/src/redaction.ts
mrcfps: `HTTP_AUTH_SCHEME_RE` still excludes `~` from the bearer-token character class, so RFC 6750-valid headers like `Authorization: Bearer abcd~efgh` only redact the `abcd` prefix and leave `~efgh` in the exported zip. Because this feature is explicitly generating support bundles for external sharing, that is still a credential leak on the main path. Please widen the auth-scheme regex to the full bearer-token alphabet (for example `[A-Za-z0-9._\-+/=:~]{4,}`) and add a regression test in `packages/diagnostics/tests/redaction.test.ts` covering a `~`-containing bearer token so the whole secret is consumed.


mrcfps (commented):
@lefarcen thanks for pushing the diagnostics export through and for tightening the earlier packaging/runtime fixes. I re-reviewed the current head and found one remaining credential-redaction gap in the new diagnostics package. I'm leaving this as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR, but I do think this case should be fixed before merge.




packages/diagnostics/src/redaction.ts
mrcfps: `packages/diagnostics/src/redaction.ts:17` still excludes `~` from the bearer-token character class. RFC 6750 allows that character, so a header like `Authorization: Bearer abcd~efgh` currently becomes `Authorization: [REDACTED] [REDACTED]~efgh` with the changed regexes, which leaves part of the credential in the exported zip. Because this feature is explicitly generating support bundles for external sharing, that is still a main-path secret leak. Please widen `HTTP_AUTH_SCHEME_RE` to the full bearer-token alphabet (for example `[A-Za-z0-9._+/=:~-]{4,}`) and add a regression test alongside `packages/diagnostics/tests/redaction.test.ts:36-81` that covers a `~`-containing bearer token so the whole secret is consumed.


mrcfps (commented):
@lefarcen thanks for pushing the diagnostics export through and for tightening the earlier packaging/runtime fixes. I re-reviewed the current head and found one remaining credential-redaction gap in the new diagnostics package. I am leaving this as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR, but I do think this case should be fixed before merge.




apps/daemon/src/diagnostics-export.ts
```diff
@@ -0,0 +1,125 @@
...
+      runtimeRoot: namespaceRoot,
+    });
+    sources.push({
+      name: `logs/${app}/latest.log`,
+      absolutePath,
+      kind: 'text',
+      tailBytes: TAIL_BYTES_PER_LOG,
+    });
+    sources.push({
+      name: `logs/${app}/renderer.log`,
```
mrcfps: `buildSidecarLogSources` loops over daemon/web/desktop and unconditionally adds `logs/${app}/renderer.log`, but the only writer added in this PR is the desktop-side append path in `apps/desktop/src/main/runtime.ts:981-1015`. That means a normal export will always emit missing-file placeholders and manifest warnings for `logs/daemon/renderer.log` and `logs/web/renderer.log` even when the bundle is otherwise healthy, which makes the warnings noisy and hides genuine missing-log problems. Please only add the renderer log for `APP_KEYS.DESKTOP` (or make renderer logs opt-in per app) here and in the mirrored helper in `apps/desktop/src/main/diagnostics.ts`, then add a zip-level regression test that asserts a healthy export stays warning-free when the expected log set exists.


mrcfps (commented):
@lefarcen thanks for wiring the shared diagnostics export through both the browser and Electron paths. I re-reviewed the current head and found one follow-up issue in the new log-source builder; leaving it as a comment because it looks straightforward to tighten in place. 🙏




apps/desktop/src/main/diagnostics.ts
```diff
@@ -0,0 +1,151 @@
...
+      absolutePath,
+      kind: "text",
+      tailBytes: TAIL_BYTES_PER_LOG,
+    });
+    sources.push({
+      name: `logs/${appKey}/renderer.log`,
+      absolutePath: join(dirname(absolutePath), "renderer.log"),
+      kind: "text",
+      tailBytes: TAIL_BYTES_PER_LOG,
+    });
```
mrcfps: `apps/desktop/src/main/diagnostics.ts:62-67` still adds `renderer.log` for daemon, web, and desktop even though the only writer introduced in this PR is the Electron renderer append path in `apps/desktop/src/main/runtime.ts:981-1014`. On the current head, every desktop-side export will therefore emit missing-file placeholders and manifest warnings for `logs/daemon/renderer.log` and `logs/web/renderer.log` even when the bundle is otherwise healthy, which makes the warnings noisy and hides genuine missing-log problems. Please mirror the daemon helper here by only adding `renderer.log` for `APP_KEYS.DESKTOP` (or making renderer logs opt-in per app), and add a regression test around the export manifest so a healthy diagnostics bundle stays warning-free when the expected files exist.


mrcfps (commented):
@lefarcen thanks for wiring the diagnostics export through both the browser and Electron paths. I re-reviewed the current head and found one follow-up issue in the desktop export log-source list; leaving it as a comment because the rest of the changed ranges look in good shape.




mrcfps (commented):
@lefarcen thanks for carrying the diagnostics export through the browser and Electron paths. I re-reviewed the current head and found one follow-up issue in the desktop export log-source list; leaving it as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR, but I do think it should be tightened before merge.




apps/desktop/src/main/diagnostics.ts
mrcfps: `apps/desktop/src/main/diagnostics.ts:62-67` still adds `renderer.log` for daemon, web, and desktop even though the only renderer-log writer introduced in this PR is the Electron append path in `apps/desktop/src/main/runtime.ts:1037-1055`. Because `buildDiagnosticsZip()` turns unreadable files into placeholders and manifest warnings, every desktop-side export from this code path will report `logs/daemon/renderer.log` and `logs/web/renderer.log` as missing even when the bundle is otherwise healthy. That makes the diagnostics summary noisy and can hide a genuinely missing desktop renderer log behind expected false positives. Please mirror `apps/daemon/src/diagnostics-export.ts` here by only appending `renderer.log` when `appKey === APP_KEYS.DESKTOP` (or by making renderer logs opt-in per app), and add a regression test around the export manifest so a healthy desktop export stays warning-free when the expected files exist.


mrcfps (commented):
@lefarcen thanks for carrying the diagnostics export through the browser and Electron paths and for tightening the earlier packaging and runtime fixes. I re-reviewed the current head and found one follow-up gap around the repo capability-exposure rule for new user-facing features; leaving it as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR.



apps/daemon/src/server.ts
```diff
@@ -3454,6 +3469,17 @@ export async function startServer({
     composio: composioConnectorProvider,
   });
 
+  // Gate the diagnostics export behind requireLocalDaemonRequest so it stays
+  // unreachable when daemon binds to a non-loopback address (Tailscale,
+  // 0.0.0.0, etc.). The bundle contains daemon/web/desktop logs, host
+  // metadata, and crash reports — same threat tier as connector / live-
+  // artifact endpoints, which all use the same guard.
+  app.get(
+    DIAGNOSTICS_EXPORT_PATH,
```
mrcfps, line 3483: This PR wires the new diagnostics-export capability through the daemon route, Settings → About UI, and desktop menu, but it still does not add a matching `od` subcommand in `apps/daemon/src/cli.ts` / `SUBCOMMAND_MAP`. The repo capability-exposure contract requires every user-facing feature to ship on both surfaces because external agents only drive Open Design through the CLI; as written, browser and Electron users can export a bundle, but automation and shell users still cannot. Please add an `od diagnostics export` entry that reuses this daemon path (or the shared zip builder), accepts an explicit output path, and returns machine-readable metadata via `--json` so the feature is reachable outside the UI as well.

apps/daemon/src/server.ts
```diff
@@ -3459,6 +3474,17 @@ export async function startServer({
...
```
mrcfps, line 3483: This diff adds the new diagnostics-export capability to the daemon route, Settings → About UI, and the desktop Help menu, but the current head still never wires a matching `od` subcommand in `apps/daemon/src/cli.ts` / `SUBCOMMAND_MAP`. The repo's capability-exposure contract requires every user-facing feature to ship on both surfaces because external agents and shell workflows only drive Open Design through `od`; as written, browser/Electron users can export a bundle, but automation and headless callers still cannot. Please add an `od diagnostics export` command in this PR that reuses this route (or the shared zip builder), accepts an explicit output path plus `--json`, and returns machine-readable metadata so the feature is reachable outside the UI too.

mrcfps (commented):
@lefarcen thanks for carrying the diagnostics export through the browser and Electron flows and for tightening the earlier follow-ups. I re-reviewed the current head and found one remaining gap against the repo's dual-surface capability contract; leaving it as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR.



apps/daemon/src/diagnostics-export.ts
```diff
@@ -0,0 +1,131 @@
...
+  try {
+    const info = userInfo();
+    return info?.username && info.username.length > 0 ? info.username : undefined;
+  } catch {
+    return undefined;
+  }
+}
+
+function buildSidecarLogSources(runtime: SidecarRuntimeContext<SidecarStamp> | null): LogSource[] {
+  if (runtime == null) return [];
```
mrcfps, line 50: `buildSidecarLogSources()` returns `[]` whenever `runtime == null`. On the current head, the standard `od` launcher still takes that path because `runDaemonCliStartup()` -> `startDaemonRuntime()` calls `startServer()` without a `runtime`, so Settings → About and the new `od diagnostics export` command both report success while producing a bundle with only the summary JSONs when Open Design is started via plain `od`. That misses the feature promise of daemon, web, and desktop logs on one of the supported entry points. Please either add a null-runtime fallback that resolves the standalone daemon and web log files, or fail or disable diagnostics export when those log roots are unavailable, and add a regression test that covers the non-sidecar launch path.


mrcfps (commented):
@lefarcen thanks for carrying the diagnostics export through the browser, Electron, and CLI paths. I found one remaining gap in the daemon-side export path; I am leaving it as a comment because this reviewer has already used its REQUEST_CHANGES quota on this PR, but I do think it should be tightened before merge.




mrcfps (approved):
@lefarcen I re-reviewed the current head and verified the diagnostics export flow across the shared package, daemon route, CLI surface, desktop bridge, and packaging follow-ups. `pnpm guard`, `pnpm typecheck`, `pnpm --filter @open-design/diagnostics test`, and `pnpm --filter @open-design/tools-pack test -- workspace-build` all passed locally, and I did not find any new actionable issues in the changed ranges. Thanks for carrying this through. 🙏

PR #800, opened 2026-05-07
feat: add opt-in Langfuse telemetry

lefarcen:
## Summary

Adds opt-in Langfuse telemetry for completed agent turns, plus the privacy controls needed to keep the default behavior at zero data upload.

- **Daemon reporting** — records completed turns as Langfuse sessions/traces after the terminal assistant message has been persisted, so prompt/output metadata and produced-file manifests reflect the saved conversation state. Reporting remains fire-and-forget and cannot break the run path.
- **Langfuse shape** — follows the recommended chatbot model: `sessionId = conversationId`, one trace per user turn / agent run, and nested observations inside each turn trace. Each trace contains an `agent-run` span, an `llm` generation child, per-tool `tool:<name>` spans when tool calls occur, and summary events for artifacts or errors when present.
- **Tool waterfall** — run events now carry timestamps, and telemetry pairs `tool_use` with `tool_result` to emit individual tool spans under `agent-run`. Tool input/output obey the conversation-content privacy toggle and are redacted/truncated before send.
- **Observed fields** — includes model/agent/reasoning/skills/design-system context, token usage, error/cancel state, duration, and optional prompt/output/artifact manifest according to user prefs.
- **Privacy defaults** — first-run consent UI starts fully off. Accept and decline actions are equally prominent, and Settings -> Privacy exposes independent toggles for anonymous metrics, conversation content, and artifact manifests.
- **Single source of truth** — telemetry prefs, anonymous installation id, and `privacyDecisionAt` live in daemon app config rather than browser localStorage. `privacyDecisionAt` is separate from `installationId`, so deleting data can rotate the id without showing the first-run banner again.
- **Delete my data** — rotates the anonymous installation id immediately and disables all telemetry toggles locally.
- **Pre-send redaction** — strips common secrets/PII patterns before upload, including API keys, JWTs, Bearer tokens, emails, IPs, and Luhn-valid credit card numbers.
- **Resilience** — Langfuse POST timeout is now 20s by default, with one retry for network errors, 429, and 5xx responses. `LANGFUSE_TIMEOUT_MS` and `LANGFUSE_RETRIES` can override this.
- **Fork/dev safety** — no-op unless Langfuse env vars are present and the user has opted in.

## Current follow-ups

- Add `docs/privacy.md` covering collection scope, retention, and data residency.
- Wire production/CI secret injection for packaged builds.
- Add README first-start disclosure text.
- Rotate the dev Langfuse key that was exposed in prior local AI chat logs.

## Test plan

- [x] `pnpm install` after merging latest `origin/main`
- [x] `pnpm --filter @open-design/contracts run typecheck`
- [x] `pnpm --filter @open-design/daemon run typecheck`
- [x] `pnpm --filter @open-design/daemon exec vitest run tests/langfuse-trace.test.ts tests/app-config.test.ts tests/langfuse-bridge.test.ts`
- [x] `pnpm --filter @open-design/daemon exec vitest run tests/langfuse-trace.test.ts tests/langfuse-bridge.test.ts`
- [x] `pnpm --filter @open-design/web run typecheck`
- [x] `pnpm --filter @open-design/web exec vitest run tests/state/config.test.ts tests/components/SettingsDialog.test.ts`
- [x] Live smoke against `https://us.cloud.langfuse.com` using the daemon process env, without printing secrets: HTTP 207 with no event errors.
- [x] Live smoke for nested observations (`trace-create` + `span-create` + `generation-create` + `event-create` with `parentObservationId`): HTTP 207, 4 successes, 0 errors.
- [x] Live smoke for per-tool spans (`agent-run` + `llm` + two `tool:<name>` spans): HTTP 207, 5 successes, 0 errors.

Note: local validation emitted Node engine warnings because this machine is on Node v22.22.0 while the repo targets Node ~24.


apps/daemon/src/langfuse-bridge.ts
```diff
@@ -0,0 +1,211 @@
...
+        const messages = (
+          listMessages as (db: unknown, cid: string) => unknown[]
+        )(db, run.conversationId);
+        const m = (messages as Array<Record<string, unknown>>).find(
+          (x) => x.id === run.assistantMessageId,
+        );
+        if (m) {
+          messageContent = typeof m.content === 'string' ? m.content : '';
+          // listMessages returns producedFiles already parsed (db.ts:965).
+          producedFilesRaw = m.producedFiles;
```
mrcfps, line 302: The bridge reads `m.producedFiles` from the assistant message immediately when `onTerminate` fires. However, `finish()` emits the `end` SSE event before the web client's `onDone` handler refreshes the project files and persists `producedFiles` back to the assistant message (`apps/web/src/components/ProjectView.tsx` updates/persists that after receiving `end`). In a normal completed run, the row read here will therefore usually not include files produced by this just-finished turn, so an opted-in `artifactManifest` trace sends an empty or stale manifest even though this PR adds manifest forwarding. Please source the manifest from daemon-side state that is available before termination (for example, collect produced artifacts from run events/agent output or a daemon-side project-file diff), or delay/report after the produced-files persistence path is complete, and add coverage for a run that creates files during the just-finished turn.

mrcfps (commented):
@lefarcen Thanks for the daemon-side telemetry groundwork here — I reviewed the config contract, Langfuse payload construction, run-close hook, and tests. I found one follow-up around when artifact manifests are sourced; the rest of the trace gating/error-isolation path looks solid. 🙏



apps/daemon/src/server.ts
```diff
@@ -3522,7 +3523,21 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
   const design = {
-    runs: createChatRunService({ createSseResponse, createSseErrorPayload }),
+    runs: createChatRunService({
+      createSseResponse,
+      createSseErrorPayload,
+      // Forwards completed agent runs to Langfuse when LANGFUSE_PUBLIC_KEY /
+      // _SECRET_KEY are set in the daemon env AND the user has opted into
+      // metrics in Settings → Privacy. Always async, always wrapped in
+      // try/catch by the bridge — never blocks or fails a run.
+      onTerminate: (run) => {
```
lefarcen: **P1** — `onTerminate` fires immediately from the run `finish()` path, before the web client can consume the `end` SSE and persist the final assistant content / producedFiles via PUT. `langfuse-bridge` then reads SQLite once and posts, so completed traces can miss the final output and artifact manifest.

**Fix**: Move reporting to the terminal message persistence path (inside `PUT /api/.../messages/:mid` after SQLite write succeeds), or make the daemon own/persist the final content internally before calling the hook.

apps/daemon/src/langfuse-trace.ts
```diff
@@ -0,0 +1,267 @@
...
+    const response = await fetchImpl(`${config.baseUrl}/api/public/ingestion`, {
+      method: 'POST',
+      headers: {
+        Authorization: config.authHeader,
+        'Content-Type': 'application/json',
+      },
+      signal: AbortSignal.timeout(FETCH_TIMEOUT_MS),
+      body: JSON.stringify({ batch }),
+    });
+    if (!response.ok) {
```
lefarcen: **P2** — Langfuse legacy ingestion returns HTTP 207 for per-event validation errors, and `fetch` treats 207 as `ok`, so malformed trace/generation events are silently accepted locally while Langfuse drops them server-side.

**Fix**: Parse the 207 response body and warn when `errors` is non-empty. Add a test proving 207-with-errors surfaces a warning.

apps/daemon/src/langfuse-trace.ts
```diff
@@ -0,0 +1,267 @@
...
+  let batch: unknown[];
+  try {
+    batch = buildTracePayload(ctx);
+  } catch (error) {
+    console.warn(`[langfuse-trace] Payload build error: ${String(error)}`);
+    return;
+  }
+
+  const serialized = JSON.stringify({ batch });
+  if (serialized.length > HARD_BATCH_MAX_BYTES) {
```
lefarcen: **P2** — The hard cap is documented as bytes, but this compares UTF-16 string length, and `truncate()` also caps characters rather than UTF-8 bytes. Non-ASCII prompts/artifact names can exceed the intended byte budgets before network send.

**Fix**: Use `Buffer.byteLength(serialized, 'utf8')` for the hard cap check and byte-aware truncation for prompt/output.

lefarcen (commented):
Hey @lefarcen 👋

You've done solid work on the foundation — the spec is thorough, the privacy gates are correct, and the tests cover the key paths. However, the coding agent found a P1 timing issue with the `onTerminate` hook that will cause incomplete traces, plus a few other correctness gaps that need fixing before this can land safely.

## P1 — onTerminate timing race with message persistence

The `onTerminate` hook fires immediately from the run `finish()` path (server.ts:3533), but at that moment the web client hasn't consumed the `end` SSE and persisted the final assistant content / producedFiles via PUT yet. `langfuse-bridge` reads SQLite once and posts, so completed traces will frequently miss the final output and artifact manifest.

**Fix**: Move the reporting to the terminal message persistence path (inside `PUT /api/.../messages/:mid` after SQLite write succeeds), or make the daemon own/persist the final content internally before calling the hook.

## P2 — Missing 207 ingestion error detection

Langfuse legacy ingestion returns HTTP 207 for per-event validation errors. Your code treats 207 as `ok` (langfuse-trace.ts:228), so malformed trace/generation events are silently accepted locally while Langfuse drops them server-side.

**Fix**: Parse the 207 response body and warn when `errors` is non-empty. Add a test proving 207-with-errors surfaces a warning.

## P2 — Byte vs character confusion in caps

You cap prompt/output at 8 KB / 16 KB (langfuse-trace.ts:258), but the code compares UTF-16 string length, and `truncate()` also caps characters rather than UTF-8 bytes. Non-ASCII prompts can exceed the intended byte budgets before network send.

**Fix**: Use `Buffer.byteLength(serialized, 'utf8')` for the hard cap check and byte-aware truncation for prompt/output.

## P3 — Q4 open question already implemented

The spec (line 402) says Q4 "legacy vs OTel" is open and blocks step 2, but this PR **is** step 2 and implements legacy ingestion.

**Fix**: Resolve Q4 inline as a decision (which you already did de facto), or pull this PR out until you formally close Q4.

## P2 — Deletion endpoint in spec is wrong

Line 448 names `POST /api/public/trace/{id}` for deletion, but Langfuse's public API exposes it as `DELETE /api/public/traces/{traceId}`.

**Fix**: Correct the endpoint/method and replace the api-limits link with the actual trace deletion API reference.

---

Once these are fixed, this will be a clean foundation. Let me know when you've pushed the fixes and I'll re-review.

specs/change/20260507-langfuse-telemetry/spec.md
```diff
@@ -0,0 +1,471 @@
...
+
+## Open Questions
+
+These do not block steps 1–2 of the plan (contracts + module + tests with
+mock fetch); they need answers before later steps.
+
+| # | Question | Default proposal | Blocks step |
+|---|---|---|---|
+| Q2 | Should artifact bodies (HTML/JSX/PPTX JSON) ever be uploaded? | Out of scope for this spec; would require a follow-up object-storage spec. | n/a (excluded) |
+| Q4 | Legacy ingestion or OTel for the first version? | Legacy (L3 has no deprecation date; one fewer dep). | 2 |
```
lefarcen: **P3** — Q4 says legacy ingestion vs OTel is still open and blocks step 2, but this PR **implements** step 2 using legacy ingestion.

**Fix**: Resolve Q4 inline as a decision (which you already did de facto), or pull this PR out until you formally close Q4.

specs/change/20260507-langfuse-telemetry/spec.md
```diff
@@ -0,0 +1,471 @@
...
+  `docs/privacy.md` as an intentional privacy property; if sharper
+  cohorting becomes important, revisit Q5 with a hashed-machine-id
+  alternative behind an explicit consent string.
+- **R7** — *Q10 binds us to short retention*. The "rotate-and-expire"
+  deletion model (Q10 resolution) only satisfies GDPR Art. 17 if the
+  org's retention is short enough that "delete = stop sending + wait" is
+  defensible as undue-delay-free. Hobby's 30 days (L5) is borderline OK;
+  Pro's 3-year retention is **not**. Mitigation: if we ever upgrade past
+  Hobby for capacity reasons (Q7), Q10 must be re-opened to add an
+  active `POST /api/public/trace/{id}` deletion call. Add a tripwire
```
lefarcen: **P2** — The deletion mitigation names `POST /api/public/trace/{id}`, but Langfuse's public API exposes trace deletion as `DELETE /api/public/traces/{traceId}`.

**Fix**: Correct the endpoint/method and replace the api-limits link with the actual trace deletion API reference.

apps/daemon/src/server.ts
```diff
@@ -1936,6 +1937,32 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+      TERMINAL_RUN_STATUSES.has(saved.runStatus) &&
+      !reportedRuns.has(saved.runId)
+    ) {
+      const run = design.runs.get(saved.runId);
+      if (run) {
+        reportedRuns.add(saved.runId);
+        // Auto-evict so the Set doesn't accumulate forever in long-running
+        // daemons. Same TTL as the runs map cleanup in runs.ts.
+        setTimeout(() => reportedRuns.delete(saved.runId), 30 * 60 * 1000).unref?.();
+        void reportRunCompletedFromDaemon({
```
mrcfps, line 2839: This new persistence-trigger path can still report canceled turns as failed because the bridge derives the trace status from the in-memory `run` object passed here. The changed condition correctly keys off `saved.runStatus`, but the existing web cancel flow persists an assistant message with `runStatus: 'canceled'` immediately after aborting/canceling the request; at that point the daemon-side run may still be `active`/`cancelRequested` until the child exits. `langfuse-bridge.normalizeStatus()` maps any non-terminal run status to `failed`, so those user-initiated cancellations become failed Langfuse traces and skew the main telemetry metrics this PR is adding. Please pass the persisted terminal status (and, ideally, the persisted `endedAt`) into `reportRunCompletedFromDaemon`, or defer reporting until the in-memory run has reached the same terminal status, and add a fixture that simulates `saved.runStatus === 'canceled'` while the run is still non-terminal.

mrcfps (commented):
@lefarcen Thanks for the quick follow-up on the earlier timing and Langfuse response issues — I re-reviewed the updated persistence-trigger path, payload caps, spec updates, and focused tests. I found one remaining cancellation-status telemetry gap that looks merge-safe but worth tightening so the new traces match user intent. 🙏



lefarcen (commented):
New head pushed (`3fb4c7c4`), but the P1 timing race mrcfps identified is still present:

The `onTerminate` hook (server.ts:3533) fires before the web client's `onDone` handler persists `producedFiles` back to SQLite. Langfuse bridge reads the assistant message immediately, so completed traces will miss artifact manifests produced during this turn.

This doesn't block merging steps 1–3 (since telemetry defaults to off), but please track the timing fix for step 4+ when telemetry goes live.

lefarcen (commented):
Thanks for addressing most of the prior findings — the 207 detection, byte-aware caps, and spec clarifications are solid fixes. However, the P1 timing race is still present under the new trigger approach. The report now fires when `saved.runStatus` is terminal (server.ts:1946), but that happens before `producedFiles` are fully persisted.

## P1 — Timing race still present (different trigger, same root cause)

The hook was removed from `onTerminate`, which is good, but the new trigger on `PUT /api/.../messages/:mid` has the same problem:

- `consumeDaemonRun` calls `onRunStatus(endStatus)` when it sees the SSE `end` event, **before** `handlers.onDone()` runs
- Web daemon persists that terminal `runStatus` immediately, triggering the report
- But `handlers.onDone()` then flushes buffered text and `refreshProjectFiles()` later persists `producedFiles` — both **after** the report already fired
- Since `reportedRuns.add(saved.runId)` happens on first terminal PUT, subsequent PUTs are ignored, so Langfuse still receives incomplete output/manifests

**Fix**: Move the reporting to fire AFTER both conditions hold:
- `saved.runStatus` is terminal **AND**
- The message row has been updated by the web's `onDone` handler (check a content/producedFiles presence signal, or introduce a daemon-side "final persistence" flag)

Alternatively, let the bridge pull from daemon-owned state that is guaranteed complete at hook time (track producedFiles internally instead of relying on web-side SQLite writes).

---

Once this is fixed, the telemetry foundation will be solid.

lefarcen (commented):
Follow-up on the timing race fix attempt:

The P1 timing issue is partially addressed but not fully resolved. The new trigger (server.ts:1946) still reports too early:

**Root cause:** The trigger treats any persisted terminal `saved.runStatus` as proof that the assistant message is final. However, in the web daemon path:
1. `consumeDaemonRun` calls `onRunStatus(endStatus)` when it sees the SSE `end` event — **before** `handlers.onDone()` runs
2. ProjectView persists that terminal status immediately
3. But buffered text is only flushed in `onDone`, and `producedFiles` are persisted after a later `refreshProjectFiles()` promise
4. Since `reportedRuns.add(saved.runId)` happens on the first terminal `PUT`, the later fully-finalized `PUT` is ignored

**Result:** Langfuse can still receive incomplete output/artifact manifests.

**Positive changes from this push:**
- ✅ P2 207 detection addressed (errors are now parsed and warned)
- ✅ P2 byte vs character caps addressed (UTF-8 byte-aware truncation)
- ✅ P3 spec Q4 resolved (no longer blocks step 2)
- ✅ P2 deletion endpoint corrected in spec

The timing fix needs one more iteration — consider deferring the Langfuse report until after the ProjectView persistence cycle completes (e.g., listen for a project-files-refreshed event, or add a daemon-side "run fully finalized" hook).

apps/daemon/src/server.ts
```diff
@@ -2288,6 +2289,32 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+    // terminal shape (web sets runStatus + producedFiles in onDone /
+    // onError / cancel before calling PUT). Tying the trigger to the
+    // persistence path — instead of the daemon-internal run-close hook —
+    // guarantees the SQLite read inside the bridge sees the final
+    // assistant content and produced-file manifest.
+    if (
+      saved &&
+      saved.runId &&
+      typeof saved.runStatus === 'string' &&
+      TERMINAL_RUN_STATUSES.has(saved.runStatus) &&
```
mrcfps: `PUT /api/projects/:id/conversations/:cid/messages/:mid` still reports on the first persisted terminal `runStatus`, but the web daemon flow persists that terminal status before it finishes persisting the final assistant message. In `apps/web/src/providers/daemon.ts:305-338`, `consumeDaemonRun()` calls `onRunStatus(endStatus)` as soon as the SSE `end` frame arrives and only later runs `handlers.onDone(acc)`. `ProjectView` saves that terminal status immediately from `onRunStatus` via `updateMessageById(..., true)` (`apps/web/src/components/ProjectView.tsx:1281-1290`), while `onDone` is the code that flushes buffered text and only persists `producedFiles` after `refreshProjectFiles()` resolves (`apps/web/src/components/ProjectView.tsx:1179-1224`). Because `reportedRuns.add(saved.runId)` happens on this first terminal PUT, the later finalized PUTs are ignored, so Langfuse can still receive incomplete `output` / artifact manifests for completed turns. Please gate reporting on a signal that the final persistence pass has happened, or move the Langfuse snapshot to daemon-owned finalized state, and add a regression that simulates terminal `runStatus` arriving before the `onDone` persistence path.

mrcfps (changes requested):
@lefarcen Thanks for the follow-up here — I re-checked the updated persistence trigger, the web run lifecycle, and the focused tests. I still found one blocker on the current head: completed-run telemetry can be emitted before the final assistant content and produced-file manifest have actually been persisted, so the new Langfuse path can still ship incomplete payloads.



apps/daemon/src/server.ts
```diff
@@ -2288,6 +2289,32 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+      !reportedRuns.has(saved.runId)
```
mrcfps, line 2831: This branch can still forward a completed run before the final assistant snapshot has actually been persisted. In the daemon web flow, `consumeDaemonRun()` calls `onRunStatus(endStatus)` as soon as it sees the SSE `end` frame (`apps/web/src/providers/daemon.ts:305-310`), and `ProjectView` persists that terminal status immediately via `updateMessageById(..., true)` (`apps/web/src/components/ProjectView.tsx:1281-1290`). The final text flush and the `producedFiles` write happen later in `onDone()` / `refreshProjectFiles()` (`ProjectView.tsx:1179-1224`). Because `reportedRuns.add(saved.runId)` happens on this first terminal PUT, the later fully-finalized PUT is ignored, so Langfuse can still receive incomplete `output` / artifact manifests for successful turns. Please wait for a daemon-side “message fully finalized” signal before marking the run as reported (or move the snapshot to daemon-owned finalized state), and add a regression that simulates terminal `runStatus` arriving before the final assistant persistence pass.

apps/web/src/components/PrivacySection.tsx
```diff
@@ -0,0 +1,189 @@
...
+      ...c,
+      installationId: generateInstallationId(),
+      telemetry: { metrics: true, content: true, artifactManifest: false },
+    }));
+  }
+
+  function declineUsage(): void {
+    setCfg((c) => ({
+      ...c,
+      installationId: null,
```
mrcfps, line 56: After `declineUsage()` clears `installationId`, the three toggle rows can still re-enable telemetry by only patching booleans (`patchTelemetry(...)` at lines 28-29 and the toggle handlers at lines 61-78). That means a user who clicks **Delete my data** or **Don't share** and later turns metrics/content back on will start sending traces with `installationId: null`, even though this UI says the action rotates the anonymous ID. That matters because the deletion model in this PR depends on minting a fresh identifier for any future opted-in traffic; otherwise reenrolled sessions lose the stable `userId` the Langfuse grouping code expects. Please generate a new `installationId` whenever telemetry is re-enabled from the opted-out state (or force re-entry through the consent action), and add a UI/state regression for that path.

mrcfps (changes requested):
@lefarcen Thanks for the follow-up here — I re-checked the updated persistence trigger, the web run lifecycle, the new privacy controls, and the focused tests. I still found one blocker in the current Langfuse handoff path, plus one smaller follow-up in the new opt-in UI, before this is ready to merge safely.



apps/daemon/src/server.ts
```diff
@@ -2288,6 +2289,32 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+    ) {
+      const run = design.runs.get(saved.runId);
+      if (run) {
+        reportedRuns.add(saved.runId);
```
mrcfps, line 2835: This branch still marks the run as reported on the first terminal `PUT`, but in the daemon web flow that `PUT` happens before the final assistant snapshot is persisted. `consumeDaemonRun()` emits `onRunStatus(endStatus)` as soon as the SSE `end` frame arrives (`apps/web/src/providers/daemon.ts:305-310`), and `ProjectView` persists that terminal status immediately via `updateMessageById(..., true)` (`apps/web/src/components/ProjectView.tsx:1281-1290`). The final text flush plus the `producedFiles` write only happen later in `onDone()` / `refreshProjectFiles()` (`apps/web/src/components/ProjectView.tsx:1179-1224`). Because `reportedRuns.add(saved.runId)` happens here on that first terminal `PUT`, the later finalized write is ignored, so Langfuse can still receive incomplete `output` / artifact manifests for successful turns. Please gate reporting on a daemon-side signal that the final persistence pass has finished (or move the Langfuse snapshot to daemon-owned finalized state) and add a regression that simulates terminal `runStatus` arriving before `onDone` persists the final message.

mrcfps (changes requested):
@lefarcen Thanks for the follow-up here — I re-checked the current head against the daemon/web persistence flow and the new privacy UI. I still found one blocker in the Langfuse handoff path: the current trigger can fire before the final assistant snapshot has been persisted, so completed turns can still be exported without their final output/artifact manifest.



apps/daemon/src/server.ts
mrcfps, line 2835: reportedRuns.add(saved.runId) still runs on the first terminal PUT, but the daemon/web flow persists that terminal runStatus before the assistant message is actually finalized. In apps/web/src/providers/daemon.ts:305-310, onRunStatus(endStatus) fires as soon as the SSE end frame arrives; ProjectView immediately persists that status via updateMessageById(..., true) in apps/web/src/components/ProjectView.tsx:1281-1290. The final text flush and producedFiles write only happen later in onDone() / refreshProjectFiles() at apps/web/src/components/ProjectView.tsx:1179-1224. Because this set entry suppresses later terminal PUTs, completed turns can still be exported without their final output or artifact manifest. Please gate reporting on a daemon-side message-fully-finalized signal (or move the Langfuse snapshot onto daemon-owned finalized state), and add a regression that simulates terminal runStatus arriving before onDone persists the final message.

mrcfps (commented):
@lefarcen Thanks for the latest pass here — I re-checked the current head against the daemon/web persistence flow, the telemetry bridge, and the focused tests. I still found one follow-up in the Langfuse handoff path on this head, so I’ve left the detail inline for maintainers to weigh. 🙏



mrcfps:
@lefarcen I'm holding off on generating review comments for nexu-io/open-design#800 because this pull request has merge conflicts right now.

mrcfps:
@lefarcen I'm holding off on generating review comments for nexu-io/open-design#800 because this pull request has merge conflicts right now.

mrcfps:
@lefarcen I'm holding off on generating review comments for nexu-io/open-design#800 because this pull request has merge conflicts right now.

apps/web/src/index.css
```diff
@@ -14046,6 +14046,96 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
...
+   same typography + radii + shadow as the rest of the panels. */
+.privacy-consent-banner {
+  position: fixed;
+  right: 16px;
+  bottom: 16px;
+  /* Above .modal-backdrop (z-index: 100). On first run the Settings
+     welcome modal mounts at the same time as this banner; without a
+     higher z-index the banner would render behind the modal and be
+     unreachable until the user closed Settings. */
+  z-index: 200;
```
mrcfps: Raising the consent banner above `.modal-backdrop` makes it intercept pointer events for unrelated dialogs on a fresh install. The current `Validate workspace` failure is reproducible in three different flows (`e2e/ui/entry-configuration-flows.test.ts`, `project-management-flows.test.ts`, and `settings-api-protocol.test.ts`): each one times out because this banner subtree is sitting on top of the dialog button it is trying to click. That matters beyond the test suite because new users can hit the same state as soon as the welcome/settings modal or a bottom-right confirmation opens before they answer the consent prompt, leaving core UI actions unusable until they happen to click Share/Don't share first. Please either keep the consent UI inside the active modal, hide/suspend it while another dialog is open, or drop it below modal chrome so it cannot steal clicks, and add a first-run regression that exercises another dialog with the banner present.

mrcfps (commented):
@lefarcen Thanks for the latest pass here — the privacyDecisionAt migration and the Langfuse hardening both move this forward. I found one fresh first-run UI regression on this head, so I’ve left the detail inline for maintainers to weigh. 🙏



apps/web/src/App.tsx
```diff
@@ -785,6 +788,52 @@ export function App() {
...
         />
       ) : null}
+      {/* First-run privacy consent banner. Stays mounted in the bottom-right
+          until the user picks Share or Don't share (gating on
+          `privacyDecisionAt`). The banner sits above the
+          Settings welcome modal — first-run users on a fresh install
+          would otherwise see the welcome modal pop on top of the banner
+          and have no way to read or interact with the consent decision
+          until they closed Settings. */}
+      {config.privacyDecisionAt == null ? (
```
mrcfps: Because `saveConfig()` and `loadConfig()` now strip `privacyDecisionAt` out of localStorage (`apps/web/src/state/config.ts:257-263` and `362-367`), `config.privacyDecisionAt` is always `undefined` on the initial render after every reload. This new top-level gate therefore mounts `PrivacyConsentModal` for returning users until the async daemon config fetch finishes, and if the daemon is slow or offline it can stay visible indefinitely even though the user already made a decision. That re-opens the first-run consent surface this change is trying to suppress and can block startup interactions on every refresh. Please gate the banner on a daemon-config-hydrated flag, or persist only the resolved-needs-consent bit locally, so the modal only renders after the daemon has confirmed the decision is still missing. Adding a regression that boots with an existing daemon `privacyDecisionAt` but empty localStorage would lock this path down.

mrcfps (commented):
@lefarcen Thanks for the follow-up here — I re-checked the privacyDecisionAt migration, banner gating, and the daemon-owned config path. I found one merge-safe regression in the first-run consent flow on this head, so I left the detail inline for maintainers to weigh. 🙏



apps/daemon/src/server.ts
lefarcen: Fixed in `aec31bfa`: the web final persistence pass now sends a request-only `telemetryFinalized` marker, and the daemon only reports when a terminal message write includes that marker. The first terminal `runStatus` PUT no longer adds the run to `reportedRuns`. Added `tests/telemetry-message-finalization.test.ts` for terminal-before-finalized sequencing.

apps/daemon/src/server.ts
lefarcen: Fixed in `a4b62856`: the daemon now passes persisted `runStatus` / `endedAt` into the Langfuse bridge, so user-canceled turns remain `canceled` even if the in-memory run has not settled yet. Added bridge coverage for `saved.runStatus === 'canceled'` while the run object is still non-terminal.

apps/web/src/index.css
lefarcen: Fixed in `aec31bfa`: the consent banner now waits for daemon config hydration, hides while Settings is open, and its z-index is below modal backdrops. Added App-level regressions for both delayed hydration and opening another dialog while the first-run banner would otherwise be present.

apps/web/src/App.tsx
lefarcen: Fixed in `aec31bfa`: the banner is gated by `daemonConfigLoaded`, so returning users do not see the first-run consent UI before daemon-owned `privacyDecisionAt` has hydrated. Added a regression that keeps the banner hidden while `fetchDaemonConfig()` is unresolved.

apps/daemon/src/server.ts
```diff
@@ -4772,6 +4830,16 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     if (typeof clientRequestId === 'string' && clientRequestId)
       run.clientRequestId = clientRequestId;
     if (typeof agentId === 'string' && agentId) run.agentId = agentId;
+    // Stash the original user prompt + per-turn config so the
+    // langfuse-bridge report path can include them without reaching back
+    // into chatBody across the createChatRunService boundary. Each field
+    // is optional and only set when the chat body actually carried it.
+    if (typeof message === 'string' && message) run.userPrompt = message;
```
mrcfps: `run.userPrompt = message` is still capturing the collapsed daemon transcript, not the current user turn. In `apps/web/src/providers/daemon.ts:117-125`, the web client builds `message` by joining the entire `history` into one transcript before it calls `/api/runs`, so the value stashed here already contains earlier prompts and assistant replies from the same conversation. That means the first trace after a user opts in can upload pre-consent conversation content from earlier turns, and every later trace will keep duplicating the whole thread instead of representing a single turn as the PR summary/spec describe. Please pass the latest user turn separately (for example a dedicated `currentPrompt`/`latestUserMessage` field on `ChatRequest`) and store that here for telemetry, then add a regression that opts in mid-conversation and verifies only the post-consent turn text is reported.

mrcfps (commented):
@lefarcen Thanks for the latest pass here — I re-checked the daemon/web telemetry handoff, privacy gating, and the focused tests. I found one important follow-up around what the new Langfuse path records as the "prompt" for daemon turns; since this reviewer has already used its three REQUEST_CHANGES reviews on earlier heads, I’m leaving it as a COMMENT for maintainers to weigh. 🙏



apps/daemon/src/server.ts
```diff
@@ -4904,6 +4962,16 @@ export async function startServer({
...
```
mrcfps: `run.userPrompt = message` is still stashing the collapsed daemon transcript instead of the current user turn. In `apps/web/src/providers/daemon.ts:117-125`, the web client builds `message` by joining the entire `history` before it calls `/api/runs`, so the value stored here already contains earlier user prompts and assistant replies from the same conversation. That means the first trace after a user opts in can upload pre-consent conversation content from prior turns, and every later trace keeps duplicating the full thread instead of reporting the single turn described in the PR summary. Please pass the latest user turn separately on `ChatRequest` (for example `currentPrompt` / `latestUserMessage`) and stash that field here for telemetry, then add a regression that opts in mid-conversation and verifies only the post-consent turn text is reported.

mrcfps (commented):
@lefarcen Thanks for the latest pass here — I re-checked the finalized-message trigger, the privacy consent flow, and the new Langfuse prompt wiring. I found one remaining follow-up in the daemon prompt handoff on this head, so I’ve left it inline for maintainers to weigh. 🙏



lefarcen (commented):
@lefarcen I re-checked the current head (`aa3c854e`) per the manual review request. mrcfps's latest finding on the daemon prompt handoff is valid and still present:

**P2 — `run.userPrompt` captures collapsed transcript, not current turn**

In `apps/web/src/providers/daemon.ts:117-125`, the web client builds `message` by collapsing the entire `history` array into one transcript before calling `/api/runs`. This collapsed transcript is then stashed into `run.userPrompt` at `apps/daemon/src/server.ts:4969`.

**Impact:**
- First telemetry trace after a user opts in can upload pre-consent conversation content from earlier turns in the same conversation
- Every subsequent trace in that conversation duplicates the full thread instead of reporting just the current turn

This contradicts the PR summary's stated goal ("forward completed runs" — singular turn) and the spec's single-turn telemetry model.

**Suggested fix** (mrcfps's recommendation):
Pass the latest user turn separately on `ChatRequest` (e.g., `currentPrompt` / `latestUserMessage`) and stash that field instead of the collapsed `message`. Add a regression test that opts in mid-conversation and verifies only post-consent turn text is reported.

Given mrcfps marked this as non-blocking and CI is green, this is merge-safe as-is (telemetry defaults to off), but the fix should land before telemetry goes live in step 4+.

apps/daemon/src/server.ts
lefarcen: Fixed in `50181603`: `ChatRequest` now carries `currentPrompt` separately from the full CLI transcript, the web client sets it from the latest user turn, and the daemon stores that field as `run.userPrompt` for Langfuse. Added web/daemon regressions covering mid-conversation history and the reported trace input containing only the post-consent turn.

apps/web/src/App.tsx
```diff
@@ -171,6 +172,8 @@ export function App() {
   // {active:false} if this hasn't run.
   const activeProjectId = route.kind === 'project' ? route.projectId : null;
```
mrcfps, line 173: When `daemonIsLive()` returns false during bootstrap, this gate still flips on because the app sets `daemonConfigLoaded` to true even though it never fetched the daemon-owned privacy fields (`App.tsx:202-210`). Since `saveConfig()` now strips `installationId` / `telemetry` / `privacyDecisionAt` out of localStorage (`apps/web/src/state/config.ts:351-367`), the first-run banner can appear in offline mode, accept a choice the daemon cannot persist, and then forget that choice on the next reload. That is a correctness problem for the new consent flow because the UI is asking for a durable privacy decision in a state where the single source of truth is unavailable. Please gate `showPrivacyConsent` on a live daemon (or otherwise suppress the banner until `/api/app-config` has actually hydrated), and add an App-level regression for the `daemonIsLive() === false` path so offline startup never shows a non-persistable consent prompt.

mrcfps (commented):
@lefarcen Thanks for the latest pass here — I re-checked the finalized telemetry handoff, the daemon-owned privacy config flow, and the focused tests. I found one merge-safe follow-up in the first-run consent path on this head, so I’ve left it inline for maintainers to weigh. 🙏



mrcfps (approved):

PR #804, opened 2026-05-07
fix/Bug#772-DesignFilesSortButton-DesignFilesTableNowSortableColumns

### LINKED ISSUE #772: Fix Design Files sort arrow button doing nothing
## Summary

The upward arrow button in the Design Files list appears to be intended for sorting, but clicking it does not produce any visible effect.

This makes the control feel broken or unfinished because users expect it to change the file ordering.

## Current behavior

- The Design Files view shows an upward arrow control near the file list header.
- Clicking the button does not trigger any visible change.
- The file order remains the same.
- There is no indicator showing whether sorting is active, disabled, or unsupported.

## Expected behavior

If the upward arrow is a sort control, clicking it should visibly change the ordering of files and communicate the current sort direction.

Suggested improvements:

- Make the button actually toggle file ordering, such as ascending/descending by time or name.
- Show the active sort state clearly in the UI.
- If sorting is not supported yet, hide or disable the control instead of leaving it interactive without effect.
- Consider adding a tooltip or label to explain the sort behavior.

## Steps to reproduce

1. Open the Design Files view.
2. Locate the upward arrow button near the list header.
3. Click the button.
4. Observe that the list order does not visibly change.

## Notes

A screenshot is attached showing the Design Files list with the upward arrow control highlighted.

## Why this matters

Sort controls are basic file-management affordances. If the button appears actionable but does nothing, users lose trust in the file browser and cannot efficiently reorganize the list to find what they need.

Image

paulstean:
Closes #772

Disclaimer: Changes made using OpenCode with Big Pickle, an AI code assistant.

1. Removed the ↑ button from DesignFilesPanel.tsx (was a "back" button that closed the preview pane — confusingly placed in the header as if it were for sorting).

2. Converted the file list to a single table with sortable columns:
    - Replaced the section-based grouping (Pages, Sketches, Scripts, Images, Other) with a flat, sortable table
    - Added column headers: Name, Kind, Modified — all clickable to toggle sort direction
    - Default sort is by modified time descending (same as previous behavior)
    - Sort indicator arrows (↑/↓) show the active sort column and direction
    - Live artifacts remain as a separate section above the table

3. Added i18n keys designFiles.colName, designFiles.colKind, designFiles.colModified to all 17 locale files and the type definitions.

4. Updated CSS with table layout styles (.df-table, .df-file-row, column width classes, sortable header styles).

    Files modified:
    - apps/web/src/components/DesignFilesPanel.tsx
    - apps/web/src/index.css
    - apps/web/src/i18n/types.ts
    - apps/web/src/i18n/locales/en.ts


lefarcen:
Hi @paulstean! 🎉
Thanks for the contribution — this looks like a significant UX improvement, converting the file panel from sections to a sortable table.

— open-design team

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +259,137 @@ export function DesignFilesPanel({
...
+              {sortedFiles.length > 0 ? (
+                <table className="df-table">
+                  <thead>
+                    <tr>
+                      <th className="df-th-check" />
+                      <th className="df-th-icon" />
+                      <th
+                        className="df-th-name df-th-sortable"
+                        onClick={toggleSort('name')}
+                        aria-sort={sortKey === 'name' ? (sortDir === 'asc' ? 'ascending' : 'descending') : 'none'}
```
mrcfps: The new sortable headers attach the sort behavior directly to `<th>` elements with only `onClick`, so the sort controls are not focusable and cannot be operated with Enter/Space. This matters because the PR's primary new interaction (changing the Design Files sort order) works for pointer users only, while the previous file browser controls were built from focusable buttons. Please put a real `<button type="button">` inside each sortable header, or add equivalent `tabIndex`, role, and keyboard handling that invokes `toggleSort` for Enter and Space while preserving the `aria-sort` state on the column header.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +259,137 @@ export function DesignFilesPanel({
...
+                      const isHovered = hover === f.name;
+                      return (
+                        <tr
+                          key={f.name}
+                          data-testid={`design-file-row-${f.name}`}
+                          className={`df-file-row ${active ? 'active' : ''} ${selected.has(f.name) ? 'selected' : ''}`}
+                          onMouseEnter={() => setHover(f.name)}
+                          onMouseLeave={() => setHover((c) => (c === f.name ? null : c))}
+                          onClick={() => setPreview(f.name)}
+                          onDoubleClick={() => onOpenFile(f.name)}
```
mrcfps: Converting each file entry from a `<button>` to a clickable `<tr>` removes keyboard focus and activation for the main row action. In the changed lines, preview/open are only wired through `onClick` and `onDoubleClick`, so keyboard users can tab to the checkbox but cannot preview or open the file row itself. Please make the row action a focusable control again (for example, a button/link in the name cell that previews on click and supports Enter/Space, with a separate open action if double-click remains pointer-only), or give the row an appropriate role, `tabIndex`, and key handler that mirrors the pointer behavior.

mrcfps (commented):
@paulstean thanks for the contribution — the sortable table direction is a nice UX improvement. I found a couple of accessibility regressions in the new table interactions that are worth tightening up before relying on the new controls. 🙂


apps/web/src/components/DesignFilesPanel.tsx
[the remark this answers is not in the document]
paulstean: This is now fixed, and changes committed.

paulstean: Done, changes committed. Thanks.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +259,140 @@ export function DesignFilesPanel({
...
-                              toggleSelect(f.name);
-                            }
-                          }}
+                        <button type="button" className="df-th-btn" onClick={toggleSort('mtime')}>
+                          {t('designFiles.colModified')}
+                          {sortKey === 'mtime' ? <span className="df-sort-arrow">{sortDir === 'asc' ? ' \u2191' : ' \u2193'}</span> : null}
+                        </button>
+                      </th>
+                      <th className="df-th-menu" />
+                    </tr>
```
lefarcen: **P2 — Removed pagination exposes large-list performance regression**

The old code had per-section render caps (`INITIAL_SECTION_FILE_LIMIT` = 30 + "show more"). This PR renders ALL files at once. For projects with 500+ generated exports or images, the table will tank — every sort reorders the full DOM.

**Concrete fix:** Keep the table UI but restore pagination / virtualized scrolling / "show N more" button. Add a large-list regression test (mock 300-500 files, measure render time or row limit behavior).

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +259,140 @@ export function DesignFilesPanel({
...
-                          e.stopPropagation();
-                          clearSection(sectionFiles);
-                        }}
+              {sortedFiles.length > 0 ? (
+                <table className="df-table">
+                  <thead>
+                    <tr>
+                      <th className="df-th-check" />
+                      <th className="df-th-icon" />
+                      <th
```
lefarcen: **P2 — Sortable headers lost keyboard activation**

The old file rows were `<button>` elements (keyboard-focusable). The new `<th onClick>` + `<tr onClick>` are clickable by mouse but not activatable by keyboard (Enter / Space won't fire onClick without role + tabIndex + keydown handlers).

**Concrete fix:** Wrap sortable header text in `<button type="button">` + add explicit Enter/Space handlers to file rows, OR add `role="button"` + `tabIndex={0}` + `onKeyDown` to the `<th>` / `<tr>`. Same issue on file rows at line 300.

apps/web/src/i18n/locales/ar.ts
```diff
@@ -559,6 +559,9 @@ export const ar: Dict = {
   'designFiles.kindPresentation': 'عرض تقديمي',
   'designFiles.kindSpreadsheet': 'جدول بيانات',
   'designFiles.kindBinary': 'ثنائي',
+  'designFiles.colName': 'Name',
```
lefarcen: **P2 — Localization regression: all non-English locales hardcode English strings**

All 16 non-English locale files add `'designFiles.colName': 'Name'` etc. (English). This is a visible i18n bug in production for Arabic, German, French, Japanese, Chinese, etc. users.

**Concrete fix:** Translate these 3 labels per locale. If you need help with translations, tools like DeepL or community contributors on the i18n Discord can provide the strings. Also applies to de.ts:516, es-ES.ts:517, fa.ts:575, fr.ts:562, hu.ts:562, id.ts:563, ja.ts:515, ko.ts:562, pl.ts:562, pt-BR.ts:574, ru.ts:574, tr.ts:553, uk.ts:575, zh-CN.ts:564, zh-TW.ts:564.

apps/web/src/index.css
```diff
@@ -4752,6 +4752,95 @@ button.connector-action.is-loading {
   font-weight: 500;
   letter-spacing: 0;
 }
+/* --- file list table --- */
+.df-table {
+  width: 100%;
+  border-collapse: collapse;
+  table-layout: auto;
+}
+.df-thead { border-bottom: 1px solid var(--border); }
```
lefarcen: **P3 — `.df-thead` rule defined but never applied**

CSS defines `.df-thead { border-bottom: ... }` but the `<thead>` in DesignFilesPanel.tsx has no `className="df-thead"`, so the header border never renders.

**Concrete fix:** Either add `<thead className="df-thead">` or change the selector to `.df-table thead` so the rule actually applies.

lefarcen (dismissed):
Hey @paulstean, thanks for tackling issue #772! The conversion from confusing ↑ button to a proper sortable table is a major UX win. Found 4 issues (2 must-fix, 2 polish) — see inline below.

paulstean:
> Hi @paulstean! 🎉 Thanks for the contribution — this looks like a significant UX improvement, converting the file panel from sections to a sortable table. I will run a deep review and get back to you within 24h.
> 
> Thanks for making open-design better! — open-design team

Thanks. I want to do further improvements to the Design Files table UI, adding ability to add/remove additional columns that might be useful to users, such as path, created, lastupdated etc - there's quite a few columns available that look useful. Should I do that in this fork, or create a new issue to discuss first?

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +259,140 @@ export function DesignFilesPanel({
...
+                      const isHovered = hover === f.name;
+                      return (
+                        <tr
+                          key={f.name}
+                          data-testid={`design-file-row-${f.name}`}
+                          className={`df-file-row ${active ? 'active' : ''} ${selected.has(f.name) ? 'selected' : ''}`}
+                          onMouseEnter={() => setHover(f.name)}
+                          onMouseLeave={() => setHover((c) => (c === f.name ? null : c))}
+                          onClick={() => setPreview(f.name)}
+                          onDoubleClick={() => onOpenFile(f.name)}
```
mrcfps: Blocking: converting each file row from a `<button>` to a clickable `<tr>` leaves the primary row action pointer-only. In these changed lines the row previews via `onClick` and opens via `onDoubleClick`, but the `<tr>` is not focusable and has no Enter/Space key handling; the only keyboard-focusable control inside the row is the checkbox, so keyboard users can no longer preview or open regular design files from the table. Please make the row action a real focusable control again (for example, a button/link in the name cell that previews on activation and exposes a separate open action), or add an appropriate role, `tabIndex`, and key handler that mirrors the pointer behavior without breaking the nested checkbox/menu controls.

apps/web/src/i18n/locales/ar.ts
```diff
@@ -559,6 +559,9 @@ export const ar: Dict = {
...
+  'designFiles.colKind': 'Kind',
+  'designFiles.colModified': 'Modified',
```
mrcfps: Non-blocking: the new column labels are visible UI text, but the non-English locale files add the English strings verbatim (`Name`, `Kind`, `Modified`). The same pattern appears across the other changed locale files (for example German, French, Japanese, Korean, Chinese, and the rest), so users who selected those languages will see the new table headers regress to English while the surrounding Design Files strings remain localized. Please translate these three keys per locale, or intentionally route missing translations through the existing fallback mechanism instead of committing English text into every non-English dictionary.

mrcfps (changes requested):
@paulstean thanks for the thoughtful Design Files table work and for already tightening up the sortable header keyboard behavior. I found one remaining keyboard-accessibility regression that blocks the main file row flow, plus one visible localization follow-up to address while these new labels are still changing. 🙂


lefarcen:
Re: adding more columns — I'd suggest **opening a new issue first** to discuss which columns are most valuable (path / created / lastUpdated / etc.) and get maintainer sign-off on the design. That way this PR stays focused on fixing #772 (the confusing ↑ button + sortable table), and the additional columns become a follow-up enhancement.

Keeping each PR scoped to one clear goal makes review faster and reduces merge conflicts. 👍

apps/web/src/components/DesignFilesPanel.tsx
lefarcen: **P2 — Removed pagination still exposes large-list performance regression**

The old code had per-section render caps (`INITIAL_SECTION_FILE_LIMIT` = 30 + "show more"). This PR renders ALL files at once in `sortedFiles.map((f) => ...)`. For projects with 500+ generated exports or images, the table will tank — every sort reorders the full DOM.

**Concrete fix:** Restore pagination ("show 50 more" button) or virtualized scrolling (e.g., react-window). Add a large-list regression test (mock 300-500 files, measure render time or verify row limit behavior).

This is the same finding from my first review — it remains unfixed on the current head.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +259,140 @@ export function DesignFilesPanel({
...
+                      </th>
+                      <th className="df-th-menu" />
+                    </tr>
+                  </thead>
+                  <tbody>
+                    {sortedFiles.map((f) => {
+                      const active = preview === f.name;
+                      const isHovered = hover === f.name;
+                      return (
+                        <tr
```
lefarcen: **P2 — File rows still lack keyboard activation (half-fixed)**

You fixed the sortable **headers** (they now have `<button>` — great!), but the file **rows** (`<tr>` at line 303) still only have `onClick` and `onDoubleClick`. Keyboard users can tab to the checkbox, but cannot preview or open the file row itself via Enter/Space.

**Concrete fix:** Add `tabIndex={0}` + `onKeyDown` handler to the `<tr>` that calls `setPreview(f.name)` on Enter and `onOpenFile(f.name)` on Shift+Enter (or similar). Example:
```tsx
<tr
  tabIndex={0}
  onKeyDown={(e) => {
    if (e.key === 'Enter') {
      e.preventDefault();
      if (e.shiftKey) onOpenFile(f.name);
      else setPreview(f.name);
    }
  }}
  ...
>
```

apps/web/src/i18n/locales/ar.ts
lefarcen: **P2 — Localization regression: all 16 non-English locales still hardcode English strings**

All non-English locale files still have `'designFiles.colName': 'Name'` (English). This is a visible i18n bug for Arabic, German, French, Japanese, Chinese, etc. users in production.

**Concrete fix:** Translate these 3 labels (`colName`, `colKind`, `colModified`) per locale. If you need help, tools like DeepL or the i18n community can provide the strings. Also applies to: de.ts:516, es-ES.ts:517, fa.ts:575, fr.ts:562, hu.ts:562, id.ts:563, ja.ts:515, ko.ts:562, pl.ts:562, pt-BR.ts:574, ru.ts:574, tr.ts:553, uk.ts:575, zh-CN.ts:564, zh-TW.ts:564.

This is the same finding from my first review — it remains unfixed.

lefarcen (dismissed):
Hey @paulstean — great progress on the keyboard-accessible sortable headers! 🎉 The `<button>` wrappers you added address mrcfps's concern perfectly.

However, 3 of my original findings still need attention before this is production-ready. See inline comments below for concrete fixes.

lefarcen (dismissed):
Hey @paulstean! Thanks for adding the header keyboard controls — that addresses one part of the accessibility concern.

To your question about adding more columns (path, created, etc.): I'd recommend creating a **separate issue** first to discuss the design with maintainers. This PR already delivers the core table conversion + sorting, which is a solid standalone improvement. Additional columns would expand scope and might need UX decisions (which columns, toggleable visibility, defaults, etc.). Keeping PRs focused makes review faster and reduces merge conflicts if multiple features evolve in parallel.

For this PR, 3 must-fix items remain (2 are P2 blockers from the prior review that are still unaddressed on this head):

**P2 #1 — Performance regression (pagination removed)**
The old code rendered 30 files initially per section (`INITIAL_SECTION_FILE_LIMIT`) with a "show more" button. This PR renders ALL files at once in `sortedFiles.map()` (line 299) with no pagination or virtualization. For projects with 500+ files (generated exports, images, etc.), the table will freeze on every sort — full DOM reorder.

**Concrete fix:** Add back pagination ("Show 50 more" button below the table) OR virtual scrolling. The table UI can stay; just don't render all N files upfront. Add a large-list test (mock 300-500 files, verify render doesn't hang).

**P2 #2 — File rows still not keyboard-activatable**
You fixed the sortable headers by wrapping them in `<button>` elements (great!). But the file rows (`<tr>` at line 303) are still only clickable by mouse — they have `onClick` and `onDoubleClick` but no `tabIndex`, `role`, or `onKeyDown` handler for Enter/Space. Keyboard users can tab to the checkbox but can't preview or open the file row itself.

**Concrete fix:** Wrap the file name cell in a `<button type="button">` that handles preview on click + Enter/Space, OR add `tabIndex={0}`, `role="button"`, and `onKeyDown={(e) => { if (e.key === 'Enter' || e.key === ' ') { e.preventDefault(); setPreview(f.name); } }}` to the `<tr>`.

**P2 #3 — i18n regression (non-English locales hardcode English)**
All 16 non-English locale files add `'designFiles.colName': 'Name'`, `'designFiles.colKind': 'Kind'`, `'designFiles.colModified': 'Modified'` — English strings. This is a visible bug for Arabic, German, French, Japanese, Chinese, etc. users (surrounding UI is localized, but these 3 new labels regress to English).

**Concrete fix:** Translate these 3 keys per locale. Tools like DeepL or the i18n community can help. Applies to: ar.ts:562, de.ts:516, es-ES.ts:517, fa.ts:575, fr.ts:562, hu.ts:562, id.ts:563, ja.ts:515, ko.ts:562, pl.ts:562, pt-BR.ts:574, ru.ts:574, tr.ts:553, uk.ts:575, zh-CN.ts:564, zh-TW.ts:564.

(The P3 CSS issue — `.df-thead` rule defined but not applied — is non-blocking; you can fix it if time allows, but not required for merge.)

apps/web/src/components/DesignFilesPanel.tsx
paulstean: I've made it so keyboard users can navigate files with tab, and can focus with spacebar/enter, which will show the preview (or tick the box to add to download list, or click the ... to show the action menu per file). Double spacebar/return will launch the file to a new tab.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +260,155 @@ export function DesignFilesPanel({
...
+                        <button type="button" className="df-th-btn" onClick={toggleSort('mtime')}>
+                          {t('designFiles.colModified')}
+                          {sortKey === 'mtime' ? <span className="df-sort-arrow">{sortDir === 'asc' ? ' \u2191' : ' \u2193'}</span> : null}
+                        </button>
+                      </th>
+                      <th className="df-th-menu" />
+                    </tr>
+                  </thead>
+                  <tbody>
+                    {sortedFiles.map((f) => {
```
mrcfps: Blocking: this table now renders every project file at once via `sortedFiles.map(...)`, while the removed section renderer capped each section at 30 rows and exposed `designFiles.showMore`. For projects with hundreds of generated images/docs, opening Design Files or toggling a sort will mount/reorder the entire DOM list on the main thread, which can freeze the main path that this PR is changing. Please preserve the new table UI but reintroduce a render cap/“show more” pagination (or virtualization) and add a large-list fixture in `apps/web/tests` that verifies only the initial page renders and sorting remains bounded.

apps/web/src/i18n/locales/ar.ts
mrcfps: Non-blocking: the new column labels are visible UI, but the non-English dictionaries add English strings verbatim (`Name`, `Kind`, `Modified`); this Arabic example has the same pattern as the other changed locale files (de/es/fa/fr/hu/id/ja/ko/pl/pt-BR/ru/tr/uk/zh). Users who selected those locales will see the new table headers switch back to English while adjacent Design Files labels remain localized. Please translate these three keys per locale, or add an intentional fallback mechanism instead of committing English strings into each non-English dictionary.

mrcfps (changes requested):
@paulstean thanks for continuing to tighten up the Design Files table and keyboard behavior. I found one remaining main-flow regression around large file lists, plus a visible localization follow-up; details inline. 🙂


lefarcen (dismissed):
Hey @paulstean! Great progress on the keyboard access — the `tabIndex` + `onKeyDown` handler on file rows is exactly the right fix for P2 #2. 👏

Two must-fix items remain:

**P2 #1 — Performance regression (pagination removed)**
The old code rendered 30 files initially per section with a "show more" button. This PR renders ALL files at once in `sortedFiles.map()` (line 300) with no pagination. For projects with 500+ design files (common in large Figma imports), the table freezes on every sort — full DOM reorder.

Concrete fix: Add a "Show 50 more" button below the table OR virtual scrolling. The table UI can stay; just don't render all N files upfront.

**P2 #3 — i18n regression (non-English locales hardcode English)**
All 16 non-English locale files add `'designFiles.colName': 'Name'`, `'designFiles.colKind': 'Kind'`, `'designFiles.colModified': 'Modified'` — English strings. Chinese, Japanese, German, French, etc. users will see English headers while surrounding UI is localized.

Concrete fix: Translate these 3 keys per locale. DeepL or the i18n community can help. Applies to: ar.ts:562, de.ts:516, es-ES.ts:517, fa.ts:575, fr.ts:562, hu.ts:562, id.ts:563, ja.ts:515, ko.ts:562, pl.ts:562, pt-BR.ts:574, ru.ts:574, tr.ts:553, uk.ts:575, zh-CN.ts:564, zh-TW.ts:564.

lefarcen (dismissed):
Hey @paulstean! Thanks for the keyboard fix on the file rows — that addresses one of the prior findings. 👍

Two P2 items still need attention before this can merge:

**P2 #1 — Performance regression (pagination removed)**
The old code rendered 30 files initially per section with a "show more" button. This PR renders ALL files at once in `sortedFiles.map()` (line 299) with no pagination or virtualization. For projects with 500+ files (generated exports, images, etc.), the table will freeze on every sort — full DOM reorder.

**Concrete fix:** Add back pagination ("Show 50 more" button below the table) OR virtual scrolling. The table UI can stay; just don't render all N files upfront.

**P2 #2 — i18n regression (non-English locales hardcode English)**
All 16 non-English locale files add English strings verbatim: `'designFiles.colName': 'Name'`, `'designFiles.colKind': 'Kind'`, `'designFiles.colModified': 'Modified'`. This is a visible bug for Arabic, German, French, Japanese, Chinese, etc. users (surrounding UI is localized, but these 3 new labels regress to English).

**Concrete fix:** Translate these 3 keys per locale. Applies to: ar.ts, de.ts, es-ES.ts, fa.ts, fr.ts, hu.ts, id.ts, ja.ts, ko.ts, pl.ts, pt-BR.ts, ru.ts, tr.ts, uk.ts, zh-CN.ts, zh-TW.ts.

lefarcen (dismissed):
Hey @paulstean! 👍 Nice work on the keyboard navigation — the `tabIndex` + double-press pattern is exactly what was needed.

Two P2 items remain:

**1. Performance regression (removed pagination)** — Line 299 in DesignFilesPanel.tsx now renders ALL files at once (`sortedFiles.map()` with no cap), while the previous section renderer limited each section to 30 rows with a "show more" button. For projects with hundreds of generated images/docs, opening Design Files or toggling sort will mount/reorder the entire list on the main thread, freezing the UI. Please reintroduce a render cap (e.g., `sortedFiles.slice(0, visibleCount)`) + a "Show more" button, or use virtualization.

**2. i18n regression (non-English locales)** — The new column labels `colName`, `colKind`, `colModified` are visible UI text, but all 16 non-English locale files (ar, de, es-ES, fa, fr, hu, id, ja, ko, pl, pt-BR, ru, tr, uk, zh-CN, zh-TW) contain English strings (`'Name'`, `'Kind'`, `'Modified'`). Users who selected those locales will see the table headers regress to English. Please translate these three keys per locale, or use the existing fallback mechanism instead of committing English into non-English dictionaries.

apps/web/src/components/DesignFilesPanel.tsx
paulstean: Added pagination, keyboard friendly. Added drop down for how many files per page. Added test for 500 files. All tests passed, and manually tested.

apps/web/src/components/DesignFilesPanel.tsx
paulstean: Resolved.

lefarcen (dismissed):
Hey @paulstean! 🎉 Great progress — you've now fixed both the pagination regression and keyboard accessibility. I can see the table properly pages at 30 rows by default (addresses the large-list performance concern), and keyboard users can now navigate file rows with Tab + Enter/Space.

One remaining P2 item:

**P2 — i18n regression (non-English locales)**
The new column labels and pagination controls are visible UI text, but all 16 non-English locale files (ar, de, es-ES, fa, fr, hu, id, ja, ko, pl, pt-BR, ru, tr, uk, zh-CN, zh-TW) still contain English strings verbatim:
- `'designFiles.colName': 'Name'` (should be translated, e.g., 名前 in Japanese, 名称 in Chinese)
- `'designFiles.colKind': 'Kind'`
- `'designFiles.colModified': 'Modified'`
- `'designFiles.perPage': 'Show'`
- `'designFiles.all': 'All'`
- `'designFiles.prev': 'Previous'`
- `'designFiles.next': 'Next'`

Users who selected those languages will see these new table headers and pagination controls regress to English while the surrounding Design Files strings remain localized. Please translate these keys per locale, or add an intentional fallback mechanism instead of committing English strings into each non-English dictionary.

Once this is addressed, this PR is ready from my side. Thanks for the careful iteration!

lefarcen (dismissed):
Hi @paulstean! 👋 Great progress — pagination is working and keyboard navigation is solid now. One P2 item remains before this is ready:

**P2 — i18n regression (non-blocking, but needs fixing):**
The new table headers and pagination controls show English strings in all non-English locales. Users who selected Chinese, Japanese, German, etc. will see "Name", "Kind", "Modified", "Show", "Previous", "Next" in English, while the surrounding UI stays localized. Please translate `designFiles.colName`, `designFiles.colKind`, `designFiles.colModified`, `designFiles.perPage`, `designFiles.all`, `designFiles.prev`, `designFiles.next` in all 16 non-English locale files, or route missing keys through the existing fallback mechanism instead of committing English text.

Once this is addressed, looks good to merge!

apps/web/src/i18n/locales/ar.ts
```diff
@@ -559,6 +559,14 @@ export const ar: Dict = {
...
+  'designFiles.perPage': 'Show',
+  'designFiles.all': 'All',
+  'designFiles.prev': 'Previous',
+  'designFiles.next': 'Next',
+  'designFiles.jumpToPage': 'Go to page',
```
mrcfps: Blocking: these new Design Files table and pagination labels are visible UI strings, but the Arabic dictionary still adds them in English (`Name`, `Kind`, `Modified`, `Show`, `All`, `Previous`, `Next`, `Go to page`). The same English values were added to the other non-English locale files in this PR, so users who have selected those languages will see this new table regress back to English while the surrounding Design Files UI remains localized. Please translate these new keys in each non-English locale (or remove them and intentionally rely on the existing fallback path instead of committing English into localized dictionaries).

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +272,230 @@ export function DesignFilesPanel({
...
+                      >
+                        <option value={15}>15</option>
+                        <option value={30}>30</option>
+                        <option value={45}>45</option>
+                        <option value={60}>60</option>
+                        <option value="all">{t('designFiles.all')}</option>
+                      </select>
+                    </label>
+                    <span className="df-page-info">
+                      {safePage * effectivePageSize + 1}–{Math.min((safePage + 1) * effectivePageSize, sortedFiles.length)} of {sortedFiles.length}
```
mrcfps: Blocking: the page range text still hard-codes the English word `of` even after adding i18n keys for the surrounding pagination controls; the same string is rendered again in the bottom paginator. This means the pagination footer remains partially English in every non-English locale even if the dictionary values above are translated. Please route the whole range label through i18n, for example a `designFiles.pageInfo` key with `{start}`, `{end}`, and `{total}` placeholders, and use it for both page-info spans.

mrcfps (changes requested):
@paulstean thanks for the continued iteration here — the pagination and keyboard-row changes address the earlier main-flow concerns nicely. I found one remaining localization regression in the new table/pagination UI that should be fixed before merge. 🙂


lefarcen (dismissed):
Hey @paulstean! 🎉 Great progress — pagination is now implemented and the keyboard rows are properly focusable. One P2 item remains:

**P2 — i18n regression (must-fix)**: All 16 non-English locale files (`apps/web/src/i18n/locales/zh-CN.ts`, `ja.ts`, `de.ts`, etc.) still contain English strings for the new pagination keys (`designFiles.perPage`, `designFiles.all`, `designFiles.prev`, `designFiles.next`, `designFiles.jumpToPage`). Users who selected Chinese/Japanese/German will see "Show", "All", "Previous", "Next" in the new pagination controls while the rest of the UI remains localized.

Please translate these 5 keys for all 16 non-English locales, or add an intentional fallback mechanism.

lefarcen:
Hey @paulstean! For adding more columns (path, created, lastUpdated, etc.), I'd suggest opening a new issue first to discuss which columns are most valuable — that'll help prioritize and keep this PR focused on the sortable-table foundation. 🙂

Once this PR lands, the new issue can build on top of it.

apps/web/src/i18n/locales/ar.ts
paulstean: Translations updated for DesignFiles section. I noticed there's still plenty of english in other parts of the application in the i18n files.

apps/web/src/index.css
paulstean: Fixed

lefarcen (approved):
Thanks for fixing the i18n regression, @paulstean — all 16 non-English locales now have proper translations for the new pagination labels. 👍

mrcfps's finding about the hard-coded "of" in the pagination footer text (line 295 + 326: `{start}–{end} of {total}`) remains unaddressed. That's a non-English-locale blocker from mrcfps's review, not mine — I'm clearing my own review.

apps/web/src/components/DesignFilesPanel.tsx
lefarcen: P2: This page-info span hard-codes the English word "of" between the two numbers, so non-English users will see "1–30 of 500" even after selecting another language. Please create a new i18n key (e.g. `designFiles.pageInfo`) with placeholders and use it here:
```
{t('designFiles.pageInfo', { start: safePage * effectivePageSize + 1, end: Math.min((safePage + 1) * effectivePageSize, sortedFiles.length), total: sortedFiles.length })}
```
and add the translated strings to all locale files (en: `"{start}–{end} of {total}"`, zh-CN: `"{start}–{end} / 共{total}"`, etc.). The bottom paginator (line ~340) has the same pattern — both need the i18n key.

lefarcen (dismissed):
Hey @paulstean! Thanks for translating all the pagination labels — that resolves the localization regression for those keys.

One remaining P2 finding: the page-info text still hard-codes the English word "of" in two places (top and bottom paginators), so non-English users will still see partially English pagination. mrcfps's earlier review suggested routing the full page-info label through i18n with placeholders like `{start}`, `{end}`, `{total}` — please see the inline comment for details.

apps/web/src/components/DesignFilesPanel.tsx
lefarcen: 🔒 P1 (from mrcfps 11:18 review): Hard-coded English `of` here. Route the whole page-info string through i18n with placeholders:
```tsx
t('designFiles.pageInfo', { start: safePage * effectivePageSize + 1, end: Math.min((safePage + 1) * effectivePageSize, sortedFiles.length), total: sortedFiles.length })
```
Then add `'designFiles.pageInfo': '{start}–{end} of {total}'` to `en.ts` and translate it in the other 16 locales (e.g., zh-CN: `'{start}–{end} 共 {total}'`, ja: `'{start}–{end} / {total}'`, etc.).

lefarcen (dismissed):
Hey @paulstean! 👋 Great progress on the i18n fix and the CSS cleanup — both of those are now solid.

One blocker remains (from mrcfps's 11:18 review): the hard-coded English `of` in the page-info text at lines 295 + 494. This should route through i18n with placeholders like `{start}`, `{end}`, `{total}` so non-English users see a properly translated pagination footer.

Once that's in, we're clear. 🎉

lefarcen (dismissed):
Hey @paulstean! The i18n regression (all 16 non-English locales showing English pagination keys) is now fixed — verified zh-CN, ja, and spot-checked ar. Nice work on the translations.

mrcfps's finding about the hard-coded `of` in the page-info footer remains unaddressed — that's between you and mrcfps to resolve.

From my review: ✅ i18n clean, ✅ pagination UX preserved, ✅ keyboard accessibility restored, ✅ CSS applied properly.

Looks good to me.

mrcfps (commented):
@paulstean thanks for the steady iteration here — the sortable table, bounded pagination, keyboard row handling, translated new controls, and CSS cleanup are all moving this in the right direction. One remaining localization detail is still visible in the current diff; because this reviewer has already hit the PR's request-changes cap, I'm leaving it as a comment for maintainer follow-up rather than another blocking review. 🙂


apps/web/src/components/DesignFilesPanel.tsx
mrcfps: The page range still hard-codes the English word `of` in the top paginator (`{start}–{end} of {total}`), and the same pattern appears again in the bottom paginator below. Since the surrounding pagination labels now come from localized dictionaries, non-English users will still see a partially English footer in the new table UI. Please route the entire range label through i18n, for example with a `designFiles.pageInfo` string containing `{start}`, `{end}`, and `{total}` placeholders, then reuse that translated helper for both page-info spans and update the existing large-list test to assert the localized key instead of the literal English `of`.

apps/web/src/i18n/locales/ar.ts
paulstean: Resolved

apps/web/src/components/DesignFilesPanel.tsx
paulstean: Fixed

apps/web/src/components/DesignFilesPanel.tsx
paulstean: Fixed

apps/web/src/i18n/locales/ar.ts
paulstean: Fixed

apps/web/src/components/DesignFilesPanel.tsx
paulstean: Fixed

apps/web/src/i18n/locales/ar.ts
paulstean: Fixed

paulstean:
All done - all issues resolved and committed. I've learned something here - to resolve all open issues before committing/synching, as that triggers another automated review. First time contributing to a github repo for me, so I'm learning lots here, thanks for having me.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +297,215 @@ export function DesignFilesPanel({
...
                         type="button"
-                        className="df-select-all"
-                        title={t('designFiles.clearSelection')}
-                        onClick={(e) => {
-                          e.stopPropagation();
-                          clearSection(sectionFiles);
-                        }}
+                        className="df-page-btn"
+                        disabled={safePage <= 0}
+                        onClick={() => setPage((p) => Math.max(0, p - 1))}
```
mrcfps: Non-blocking: the visible page is clamped through `safePage`, but the underlying `page` state is not clamped when `files` shrinks after a refresh/delete/project update. If a user is on a high page and the list drops to fewer pages, this button still decrements from the stale `page` value, so the UI can keep showing the same clamped last page for several clicks before navigation appears to work. Please clamp `page` in an effect when `totalPages` changes, or base the prev/next updates on `safePage` (for example `setPage(Math.max(0, safePage - 1))`) so a single click always changes the rendered page.

mrcfps (commented):
@paulstean thanks for the continued iteration here — the sortable table now has bounded pagination, localized labels, and keyboard-accessible controls. I found one remaining pagination state edge case that is worth tightening up, but it should be safe to handle as a follow-up. 🙂


apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +297,215 @@ export function DesignFilesPanel({
...
+                        <option value={15}>15</option>
+                        <option value={30}>30</option>
+                        <option value={45}>45</option>
+                        <option value={60}>60</option>
+                        <option value="all">{t('designFiles.all')}</option>
+                      </select>
+                    </label>
+                    <span className="df-page-info">
+                      {safePage * effectivePageSize + 1}–{Math.min((safePage + 1) * effectivePageSize, sortedFiles.length)} {t('designFiles.of', { total: sortedFiles.length })} {sortedFiles.length}
+                    </span>
```
lefarcen, line 378: P2 — Duplicate total: the code renders `{t('designFiles.of', { total })} {sortedFiles.length}`, which will show as `1–30 of 100 100`. Either use `{t('designFiles.of')} {sortedFiles.length}` (remove the `total` param), OR create a full `designFiles.pageInfo` key with `{start}`, `{end}`, `{total}` placeholders per mrcfps's suggestion.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -300,154 +297,215 @@ export function DesignFilesPanel({
...
+                    <button
+                      type="button"
+                      className="df-page-btn"
+                      disabled={safePage >= totalPages - 1}
+                      onClick={() => setPage((p) => Math.min(totalPages - 1, p + 1))}
+                    >
+                      {t('designFiles.next')}
+                    </button>
+                    <span className="df-page-info">
+                      {safePage * effectivePageSize + 1}–{Math.min((safePage + 1) * effectivePageSize, sortedFiles.length)} {t('designFiles.of', { total: sortedFiles.length })} {sortedFiles.length}
```
lefarcen: P2 — Same duplicate-total bug as the top paginator. Apply the same fix here.

lefarcen (commented):
Hey @paulstean! 👋 You've addressed the hard-coded "of" issue (nice work!), but there's a small bug in the implementation:

The pagination footer renders the total **twice** — once inside the `t()` call and once after it:
```tsx
{t('designFiles.of', { total: sortedFiles.length })} {sortedFiles.length}
```
This will display as `1–30 of 100 100` (duplicate "100").

Two ways to fix:
1. **Simple**: Remove the duplicate — just `{t('designFiles.of')} {sortedFiles.length}`
2. **Full i18n** (what mrcfps suggested): Create a `designFiles.pageInfo` key with placeholders `{start}`, `{end}`, `{total}`, then use `{t('designFiles.pageInfo', {start, end, total})}`

Either approach works — pick the one that feels cleaner to you. The same fix applies to both pagination sections (top + bottom).

Otherwise, looking solid! The sortable table + pagination + keyboard handling are all in good shape. 🎉

lefarcen (changes requested):
Hey @paulstean! Thanks for the quick iteration. I found one implementation issue in the latest commit (8228c6ff) that's causing the pagination footer to still render duplicate totals.

**P2 — Duplicate total rendering persists**

Commit 8228c6ff removed the `{ total: sortedFiles.length }` interpolation parameter from `t('designFiles.of', ...)`, but kept the trailing `{sortedFiles.length}` literal on both lines 324 and 508.

Current output: `1–30 of 100 100` (the "of" translation + the literal total)

The fix should be **one** of:
- **Option A (mrcfps's recommendation)**: Create a single i18n key with placeholders:
  ```tsx
  {t('designFiles.pageInfo', { start: safePage * effectivePageSize + 1, end: Math.min(...), total: sortedFiles.length })}
  ```
  And in `en.ts`: `'designFiles.pageInfo': '{start}–{end} of {total}'`

- **Option B (minimal)**: Add the placeholder back to `designFiles.of`:
  In `en.ts`: `'designFiles.of': 'of {total}'`
  In the component: `{t('designFiles.of', { total: sortedFiles.length })}` (remove the trailing literal)

Either approach works — Option A is cleaner because it routes the whole label through one key (matching mrcfps's original suggestion).

lefarcen (approved):
Hey @paulstean! 🎉 All P2 items are now resolved:

✅ **Pagination + keyboard** — Both addressed in earlier commits.
✅ **i18n regression** — All 16 non-English locales now have proper translations (verified zh-CN, ja, and spot-checked ar).
✅ **Duplicate total rendering bug** — Fixed by using `designFiles.pageInfo` with placeholders `{start}`, `{end}`, `{total}` in both pagination sections (Option A from mrcfps's suggestion).

From my review: the sortable table, bounded pagination, keyboard accessibility, and full i18n coverage are all in good shape. Nice work on the careful iteration!

mrcfps's pagination state edge case (line 3201193002) is non-blocking per their review — they suggested handling it as a follow-up.

Looks good to me.

lefarcen (dismissed):
Thanks for fixing the duplicate total rendering, @paulstean! The pagination footer now properly routes the entire label through `t('designFiles.pageInfo', { start, end, total })` across all 17 locales. Verified:

- Both pagination sections (top + bottom) use the new `pageInfo` key
- All non-English locales have proper translations with placeholders
- The duplicate `{sortedFiles.length}` literal is removed

The sortable table + pagination + keyboard accessibility + i18n are all solid now. From my review: ✅ clean. Nice work on the steady iteration!

lefarcen (approved):
Hey @paulstean! 🎉

All findings resolved:

✅ **Keyboard accessibility** — file rows now have `tabIndex` + `onKeyDown` for Space/Enter
✅ **Pagination** — bounded rendering with page state + pageSize selector
✅ **i18n** — all table/pagination labels translated across 16 non-English locales
✅ **Hard-coded "of"** — now routes through `designFiles.pageInfo` i18n key
✅ **Pagination state edge case** — `setPage((p) => Math.min(p, totalPages - 1))` clamps correctly when files shrink

The sortable table is now production-ready. Thanks for the thoughtful iteration!
— open-design team

lefarcen (dismissed):
Hey @paulstean — all the pagination and i18n fixes are solid! 🎉

The pagination state now clamps correctly when the file list shrinks (`useEffect` on `totalPages` change), and the page-range label routes through `designFiles.pageInfo` with proper placeholders across all 17 locales. Clean work.

This is good to merge from my side. Thanks for the steady iteration and for making open-design better!

mrcfps (commented):
@paulstean thanks for the careful iteration here — the sortable table, pagination, keyboard handling, tests, and localization fixes are in much better shape now. I found one small pagination state edge case that is worth tightening up, but it should be safe to handle as a follow-up. 🙂


apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -66,30 +49,40 @@ export function DesignFilesPanel({
...
+      else if (sortKey === 'kind') cmp = kindSortPriority(a.kind) - kindSortPriority(b.kind);
+      else cmp = a.mtime - b.mtime;
+      return sortDir === 'asc' ? cmp : -cmp;
+    });
+  }, [files, sortKey, sortDir]);
+
+  const [page, setPage] = useState(0);
+  const [pageSize, setPageSize] = useState<number | 'all'>(30);
+
+  const effectivePageSize = pageSize === 'all' ? sortedFiles.length : pageSize;
```
mrcfps: Non-blocking: `pageSize === 'all'` makes `effectivePageSize` equal `sortedFiles.length`, so if the user has selected All and the file list becomes empty after a refresh/delete/project update, `effectivePageSize` becomes `0`. The next line then computes `Math.ceil(0 / 0)`, producing `NaN`, and the `totalPages` clamp effect can persist `page` as `NaN`; when files appear again, `safePage`, `rangeStart`, and `pageFiles` are still derived from `NaN`, so the table can show no rows/`NaN` page info despite having files. Please guard the empty-All case with a non-zero page size or finite total, for example `const effectivePageSize = pageSize === 'all' ? Math.max(1, sortedFiles.length) : pageSize`, and clamp with `Number.isFinite(totalPages)` so empty refreshes cannot poison pagination state.

lefarcen (dismissed):
Hi @paulstean! ✅

The pagination edge case is now fixed — `Math.max(1, sortedFiles.length)` ensures `effectivePageSize` is never 0, and `Number.isFinite(totalPages)` guards the clamp effect. This addresses mrcfps's final non-blocking finding.

I also see you changed the file row subtitle from kind to file size (`formatSize(f.size)`). That's a minor UX tweak (showing size instead of duplicating the Kind column in the subtitle) — looks reasonable.

All P2 findings from the review cycle are now resolved. Nice work on the iteration! 🎉

Deferring final merge decision to maintainer.
— open-design team

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -329,154 +332,215 @@ export function DesignFilesPanel({
...
-                            }
-                          }}
+                        {t('designFiles.next')}
+                      </button>
+                    </div>
+                  </div>
+                  <table className="df-table">
+                    <thead>
+                      <tr>
+                        <th className="df-th-check" />
```
mrcfps: Removing the old per-section `selectAllInSection` / `clearSection` affordances leaves the new table with no replacement bulk-select control: this header cell is empty, and the only remaining way to build a selection is toggling rows one by one. That regresses the existing batch download/delete flow for larger projects, because selecting a whole export set now requires clicking through every matching row across every page instead of using a single section-level action as before. Please add a table-level bulk selector here (for example a header checkbox or explicit `Select page` / `Select all` + `Clear selection` actions wired to `pageFiles` / `sortedFiles`) so the sortable-table rewrite preserves the previous multi-select capability.

mrcfps (commented):
@paulstean thanks for the careful iteration here — the sortable table, pagination, keyboard handling, and localization follow-ups are in much better shape now. I found one remaining regression in the new table UI that would be worth addressing before this pattern spreads further. 🙂


apps/web/src/i18n/locales/de.ts
```diff
@@ -488,6 +488,7 @@ export const de: Dict = {
   'designFiles.downloadSelected': 'Download {n} as ZIP',
   'designFiles.deleteSelected': '{n} löschen',
   'designFiles.clearSelection': 'Clear',
+  'designFiles.selectPage': 'Select all on page',
```
mrcfps: This new bulk-selection copy is still hard-coded in English here, and the same pattern remains in `apps/web/src/i18n/locales/es-ES.ts:489-493` and `apps/web/src/i18n/locales/ja.ts:487-491` (`Download {n} as ZIP`, `Clear`, `Select all on page`, `Select all`). That means German/Spanish/Japanese users will still see a partially English selection toolbar in the new table UI even though the pagination labels were localized. Please translate those four `designFiles.*` keys in the affected locale files before this i18n pass lands so the restored selection controls stay consistent with the rest of the panel.

mrcfps (commented):
@paulstean thanks for the steady iteration here — the sortable table, pagination follow-ups, and restored bulk-selection affordances are in much better shape now. I found one remaining non-blocking localization gap in the new selection toolbar.


lefarcen (approved):
Hi @paulstean! ✅

mrcfps's final i18n finding is now resolved — the bulk-selection toolbar strings (`downloadSelected`, `clearSelection`, `selectPage`, `selectAll`) are properly translated across all non-English locales.

The sortable table + pagination + keyboard accessibility + full localization coverage are all solid. Nice work on the thorough iteration! 🎉

Deferring final merge decision to maintainer.
— open-design team

lefarcen (dismissed):
Hey @paulstean! ✅

The bulk-selection toolbar strings (`downloadSelected`, `clearSelection`, `selectPage`, `selectAll`) are now properly localized across all 17 locale files — verified de/es-ES/ja no longer have English fallbacks. mrcfps's final i18n finding is addressed.

All P2 findings from the review cycle are now resolved. Nice work on the careful iteration! 🎉

— open-design team

mrcfps (commented):
@paulstean thanks for the careful iteration here — the sortable table, pagination, bulk-selection follow-ups, and locale fixes are in much better shape now. I found one remaining non-blocking cleanup in the Indonesian locale diff that looks like accidental carry-over from unrelated work. 🙂


Location: apps/web/src/i18n/locales/id.ts RIGHT lines 729-749

This hunk adds a second copy of the `fileViewer.deployProvider*` / `fileViewer.cloudflare*` translations even though the same keys are already defined again a few lines lower in this object (`fileViewer.cloudflareApiToken` starts at line 759 on the new side). Because object literals keep only the last duplicate, the entries added here are dead code, and this PR ends up carrying unrelated deploy-translation churn in a Design Files change. Please drop this duplicated provider block — and any other unrelated Indonesian locale edits in this PR — so the diff only contains the `designFiles.*` and workspace strings needed for the sortable-table work.



mrcfps (commented):
@paulstean thanks for the continued iteration here — the sortable table, pagination, bulk actions, and translated design-files strings are in much better shape now. I found one remaining non-blocking locale cleanup in the current diff.


Location: apps/web/src/i18n/locales/zh-CN.ts RIGHT line 740

These deploy-provider strings are unrelated to the Design Files work, and the added values here are still English (`fileViewer.cloudflareApiToken`, `fileViewer.cloudflareAccountId`). The same mixed-language carry-over appears in the matching changed hunks in `apps/web/src/i18n/locales/zh-TW.ts` and `apps/web/src/i18n/locales/id.ts`. That matters because this PR is no longer just the `#772` table/pagination follow-up: reviewers now have to validate deploy/settings copy too, and users on those locales will still see partially English labels in the touched sections. Please either drop the unrelated deploy/settings locale churn from this PR, or finish translating the added keys everywhere this PR modifies them so the change stays scoped and consistent.



apps/web/src/components/DesignFilesPanel.tsx
mrcfps: Looper checked this thread against head `1eb1f1ab47b2f80c425b464a5b81f45d73554abe`. The requested change appears objectively addressed by apps/web/src/components/DesignFilesPanel.tsx:155-173 adds bulk-selection handlers for the current page, all sorted files, and clearing selection; apps/web/src/components/DesignFilesPanel.tsx:379-393 renders corresponding Select all on page / Select everything / Clear actions, restoring multi-select without per-row clicking., so I’m resolving this thread. Reopen if this still needs discussion.

mrcfps (dismissed):
@paulstean I re-checked the latest changed ranges in `DesignFilesPanel` and the follow-up locale cleanup, and the remaining selection-toolbar / pagination / i18n issues from the earlier review cycle are addressed on this head. The current diff looks scoped and internally consistent, and I didn’t find any new actionable correctness, safety, or maintainability issues in the touched ranges. Thanks for sticking with the follow-ups and tightening this up. 🙂


lefarcen:
Hi @paulstean! 👋 This PR now has merge conflicts that need to be resolved before it can be merged. You can resolve them by:

```bash
git fetch origin main
git merge origin/main
# resolve conflicts
git commit
git push
```

Let me know if you need any help resolving the conflicts.

mrcfps (dismissed):
@paulstean I re-checked the current changed ranges in `apps/web/src/components/DesignFilesPanel.tsx`, the matching locale updates, and the `DesignFilesPanel` test wiring on head `bc70ee5`. The pagination, bulk-selection, and `pageInfo`/locale follow-ups look internally consistent on this head, and I did not find any new actionable correctness, safety, or maintainability issues in the touched ranges. Thanks for sticking with the follow-ups and tightening this up. 🙂


lefarcen:
Hi @paulstean! The merge conflicts are resolved ✅ and mrcfps has approved, but there's now a CI check failure that needs attention:

❌ **Validate workspace** check failed

You can see the failure details in the Checks tab. Once this is fixed, the PR should be ready to merge.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -689,6 +810,12 @@ function kindLabel(kind: ProjectFileKind, t: TranslateFn): string {
   return t('designFiles.kindBinary');
 }
 
+function formatSize(bytes: number): string {
```
mrcfps: formatSize() duplicates the same byte-formatting thresholds and output that humanBytes() already implements a few lines below in this file. That matters because any future change to size formatting now has to be kept in sync in two places, which makes it easy for the row subtitle and preview pane to drift apart again. Since the preview already calls humanBytes(file.size), please reuse that existing helper here (or rename it to a more generic shared name) instead of introducing a second formatter.

mrcfps (commented):
@paulstean thanks for the steady follow-ups here — the table rewrite, pagination fixes, restored bulk actions, locale cleanups, and the final e2e selector update all look much tighter on this head. I found one small maintainability cleanup in the new size-label helper that would be worth landing while this area is still open. 🙂


mrcfps (approved):
@paulstean I re-checked the changed ranges in `apps/web/src/components/DesignFilesPanel.tsx`, the matching locale/type updates, the new `apps/web/tests/components/DesignFilesPanel.test.tsx` coverage, and the `e2e/ui/app.test.ts` selector fix on this head. The pagination guards, bulk-selection flow, preview/open interactions, and the follow-up test updates look internally consistent here, and I didn’t find any new actionable correctness, safety, or maintainability issues in the touched ranges. Thanks for sticking with the iteration and tightening this up. 🙂


paulstean:
Closes #772

Issue #214, opened 2026-05-01
./src/better_sqlite3.lzz:2:10: fatal error: 'climits' file not found


alex-denisov:
When trying to do a fresh install on MacOS, using 
```
git clone https://github.com/nexu-io/open-design.git
cd open-design
corepack enable
corepack pnpm --version   # should print 10.33.2
pnpm install
pnpm tools-dev run web
```
It fails on `pnpm install` with following output:
```
username@local open-design % pnpm install
Scope: all 12 workspace projects
Lockfile is up to date, resolution step is skipped
Packages: +528
+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
Progress: resolved 528, reused 524, downloaded 0, added 528, done
node_modules/.pnpm/better-sqlite3@11.10.0/node_modules/better-sqlite3: Running install script, failed in 33.1s
.../node_modules/better-sqlite3 install$ prebuild-install || node-gyp rebuild --release
│ (node:7837) [DEP0176] DeprecationWarning: fs.R_OK is deprecated, use fs.constants.R_OK instead
│ (Use `node --trace-deprecation ...` to show where the warning was created)
│ prebuild-install warn install No prebuilt binaries found (target=24.14.1 runtime=node arch=x64 libc= platform=darwin)
│ gyp info it worked if it ends with ok
│ gyp info using node-gyp@11.5.0
│ gyp info using node@24.14.1 | darwin | x64
│ (node:7844) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead…
│ (Use `node --trace-deprecation ...` to show where the warning was created)
│ gyp info find Python using Python version 3.14.3 found at "/Library/Frameworks/Python.framework/Versions/3.14/bin/python3"
│ gyp WARN EACCES current user ("username") does not have permission to access the dev dir "/Users/username/Library/Caches/node-gyp/24.14.1"
│ gyp WARN EACCES attempting to reinstall using temporary dev dir "/var/folders/8h/b2_bzpmd26v96gdcv37d7z_w0000gn/T/.node-gyp"
│ gyp info spawn /Library/Frameworks/Python.framework/Versions/3.14/bin/python3
│ gyp info spawn args [
[...]

PR #813, opened 2026-05-07
fix(postinstall): auto-rebuild better-sqlite3 on Node.js ABI mismatch

VikingOwl91:
## Problem

`pnpm install` uses `prebuild-install` to fetch a prebuilt `better-sqlite3`
binary for the Node.js version active at install time. On systems where the
running Node ABI differs from Node 24 — Arch Linux system Node (currently 25),
`nodejs-lts-jod` (Node 22), or any environment where nvm/fnm isn't in use —
the addon silently installs but then fails to dlopen at daemon startup with an
ABI mismatch error. This also breaks AUR package builds under makepkg/fakeroot.

Related: #214

## Fix

`scripts/postinstall.mjs` now verifies the native addon loads after the
workspace builds complete. On failure it:

1. Resolves the pnpm symlink to the actual store path for `better-sqlite3`
2. Locates `node-gyp` from the pnpm virtual store (bundled as a direct dep of
   `better-sqlite3`, so no external tooling needed beyond a C++ compiler)
3. Runs `node-gyp rebuild` in the package directory against the current Node.js

`pnpm install` becomes self-healing across Node versions without requiring
nvm, fnm, or any other version manager.

A QUICKSTART troubleshooting entry is added for the edge case where
`ignore-scripts=true` prevents postinstall from running automatically.

## Changes

- `scripts/postinstall.mjs` — ABI check + conditional rebuild (~45 lines)
- `QUICKSTART.md` — one troubleshooting bullet for `ignore-scripts` users

### DIFF
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -214,7 +214,7 @@ open-design/
 
 ## Troubleshooting
 
-- **`better-sqlite3` fails to load / ABI mismatch after a Node.js version change** — `pnpm install` re-runs `postinstall` automatically and rebuilds the native addon. If you have `ignore-scripts=true` in your `.npmrc` (or used `pnpm install --ignore-scripts`), run it manually: `node scripts/postinstall.mjs`
+- **`better-sqlite3` fails to load / ABI mismatch after a Node.js version change** — `pnpm install` re-runs `postinstall` automatically and rebuilds the native addon for the current Node.js. To rebuild manually or verify the fix: `pnpm --filter @open-design/daemon rebuild better-sqlite3` then `node -e "require('better-sqlite3')"`. Requires build tools: `python3`, `make`, `g++` (or `clang++`). If you have `ignore-scripts=true` in your `.npmrc`, run `node scripts/postinstall.mjs` after `pnpm install`.
 - **"no agents found on PATH"** — install one of: `claude`, `codex`, `devin`, `gemini`, `opencode`, `cursor-agent`, `qwen`, `qodercli`, `copilot`. Or switch to API mode in Settings and paste a provider key.
 - **daemon 500 on /api/chat** — check the daemon terminal for the stderr tail; usually the CLI rejected its args. Different CLIs take different argv shapes; see `apps/daemon/src/agents.ts` `buildArgs` if you need to tweak.
 - **media generation says `OD_BIN` is missing or daemon URL is `:0`** — run the media dispatcher checks above. Do not resume the old CLI session; reopen the project from the Open Design app so the daemon can inject fresh `OD_*` variables.
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -214,7 +214,7 @@ open-design/
 
 ## Troubleshooting
 
-- **`better-sqlite3` fails to load / ABI mismatch after a Node.js version change** — `pnpm install` re-runs `postinstall` automatically and rebuilds the native addon for the current Node.js. To rebuild manually or verify the fix: `pnpm --filter @open-design/daemon rebuild better-sqlite3` then `node -e "require('better-sqlite3')"`. Requires build tools: `python3`, `make`, `g++` (or `clang++`). If you have `ignore-scripts=true` in your `.npmrc`, run `node scripts/postinstall.mjs` after `pnpm install`.
+- **`better-sqlite3` fails to load / ABI mismatch after a Node.js version change** — `pnpm install` re-runs `postinstall` automatically and rebuilds the native addon for the current Node.js. To rebuild manually or verify the fix: `pnpm --filter @open-design/daemon rebuild better-sqlite3` then `pnpm --filter @open-design/daemon exec node -e "require('better-sqlite3')"`. Requires build tools: `python3`, `make`, `g++` (or `clang++`). If you have `ignore-scripts=true` in your `.npmrc`, run `node scripts/postinstall.mjs` after `pnpm install`.
 - **"no agents found on PATH"** — install one of: `claude`, `codex`, `devin`, `gemini`, `opencode`, `cursor-agent`, `qwen`, `qodercli`, `copilot`. Or switch to API mode in Settings and paste a provider key.
 - **daemon 500 on /api/chat** — check the daemon terminal for the stderr tail; usually the CLI rejected its args. Different CLIs take different argv shapes; see `apps/daemon/src/agents.ts` `buildArgs` if you need to tweak.
 - **media generation says `OD_BIN` is missing or daemon URL is `:0`** — run the media dispatcher checks above. Do not resume the old CLI session; reopen the project from the Open Design app so the daemon can inject fresh `OD_*` variables.
--- a/QUICKSTART.md
+++ b/QUICKSTART.md
@@ -214,6 +214,7 @@ open-design/
 
 ## Troubleshooting
 
+- **`better-sqlite3` fails to load / ABI mismatch after a Node.js version change** — `pnpm install` re-runs `postinstall` automatically and rebuilds the native addon. If you have `ignore-scripts=true` in your `.npmrc` (or used `pnpm install --ignore-scripts`), run it manually: `node scripts/postinstall.mjs`
 - **"no agents found on PATH"** — install one of: `claude`, `codex`, `devin`, `gemini`, `opencode`, `cursor-agent`, `qwen`, `qodercli`, `copilot`. Or switch to API mode in Settings and paste a provider key.
 - **daemon 500 on /api/chat** — check the daemon terminal for the stderr tail; usually the CLI rejected its args. Different CLIs take different argv shapes; see `apps/daemon/src/agents.ts` `buildArgs` if you need to tweak.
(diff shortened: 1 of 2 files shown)


lefarcen:
Hi @VikingOwl91! 🎉
Thanks for the contribution — this is a real pain point with Node.js ABI mismatches across environments.

— open-design team

scripts/postinstall.mjs
```diff
@@ -48,3 +50,50 @@ for (const target of buildTargets) {
     process.exit(result.status ?? 1);
   }
 }
+
+// Verify the better-sqlite3 native addon loads under the current Node.js ABI.
+// prebuild-install may have fetched a prebuilt binary for a different ABI (e.g.
+// after switching between Node 22 / 24 / 25). When the addon fails to dlopen,
+// rebuild from source using the node-gyp bundled with better-sqlite3 in the
+// pnpm virtual store — no external tooling required beyond a C++ compiler.
+const b3Link = resolve(repoRoot, "node_modules", "better-sqlite3");
```
mrcfps: The load/rebuild check is currently looking for `node_modules/better-sqlite3` at the workspace root, but the changed behavior needs to repair the daemon dependency. In this pnpm workspace, `better-sqlite3` is declared by `apps/daemon/package.json` and the lockfile importer records it under that package, so a normal install links it at `apps/daemon/node_modules/better-sqlite3`, not as a root dependency. That means `existsSync(b3Link)` can be false in the exact ABI-mismatch scenario this PR is meant to fix, and postinstall will skip the verification/rebuild entirely before the daemon later fails to load SQLite. Please resolve the addon from the daemon package location (for example `resolve(repoRoot, "apps/daemon/node_modules/better-sqlite3")`, or use `createRequire(resolve(repoRoot, "apps/daemon/package.json")).resolve("better-sqlite3")`) so the check targets the installed daemon dependency.

scripts/postinstall.mjs
```diff
@@ -48,3 +50,50 @@ for (const target of buildTargets) {
...
+    req(b3Link);
+    b3Loads = true;
+  } catch {}
+
+  if (!b3Loads) {
+    // realpathSync resolves the pnpm symlink to the actual store path, e.g.
+    // node_modules/.pnpm/better-sqlite3@X.Y.Z/node_modules/better-sqlite3
+    // node-gyp is a direct dep and lives one level up in the same subtree.
+    const b3Dir = realpathSync(b3Link);
+    const nodeGypScript = resolve(b3Dir, "..", "node-gyp", "bin", "node-gyp.js");
```
mrcfps: This assumes `node-gyp` is installed as a sibling of the resolved `better-sqlite3` package, but the lockfile for `better-sqlite3@12.9.0` lists only `bindings` and `prebuild-install` as its dependencies. In this workspace `node-gyp` is present via `@electron/rebuild`, not as `.../better-sqlite3/../node-gyp/bin/node-gyp.js`, so an actual failed load will likely hit the `node-gyp not found` branch and exit instead of rebuilding. Please locate `node-gyp` from a real installed resolver context (for example via `createRequire` against a package that actually depends on it), add it as an explicit dependency where postinstall is allowed to use it, or invoke the package manager’s rebuild command for the daemon dependency so the recovery path does not depend on a nonexistent sibling layout.

mrcfps (changes requested):
@VikingOwl91 thanks for taking on the native-addon ABI mismatch path — this is a useful install-time recovery improvement. I found two correctness issues in the new postinstall recovery path that make the self-healing rebuild miss or fail in the workspace layout, so I’m requesting changes to keep the fix reliable.


scripts/postinstall.mjs
lefarcen: P1 — The `better-sqlite3` dependency lives in `apps/daemon/package.json`, so pnpm links it at `apps/daemon/node_modules/better-sqlite3`, not `repoRoot/node_modules/better-sqlite3`. This check will always be false in the default install, and the ABI verification/rebuild will silently never run.

Fix: resolve from the daemon package instead:
```js
const b3PkgPath = require.resolve("better-sqlite3/package.json", { paths: [resolve(repoRoot, "apps/daemon")] });
const b3Dir = dirname(b3PkgPath);
```
Then load/rebuild from that directory.

scripts/postinstall.mjs
lefarcen: P1 — `pnpm-lock.yaml` shows `better-sqlite3@12.9.0` depends only on `bindings` and `prebuild-install`, not `node-gyp` as a package-local direct dependency. `node-gyp` is supplied by the package-manager lifecycle, not bundled with the package.

Once the daemon path is fixed, ABI failures will hit "node-gyp not found" instead of rebuilding.

Fix options:
- Use pnpm's lifecycle-aware rebuild: `pnpm --filter @open-design/daemon rebuild better-sqlite3 --pending`
- Or explicitly add and resolve a supported `node-gyp` dependency

QUICKSTART.md
```diff
@@ -214,6 +214,7 @@ open-design/
 
 ## Troubleshooting
 
+- **`better-sqlite3` fails to load / ABI mismatch after a Node.js version change** — `pnpm install` re-runs `postinstall` automatically and rebuilds the native addon. If you have `ignore-scripts=true` in your `.npmrc` (or used `pnpm install --ignore-scripts`), run it manually: `node scripts/postinstall.mjs`
```
lefarcen: P2 — The new guidance gives only `pnpm install` and `node scripts/postinstall.mjs`, but omits the existing package-scoped recovery command already recommended by `tools-dev` diagnostics: `pnpm --filter @open-design/daemon rebuild better-sqlite3 --pending`.

Industry peers (Electron, n8n, etc.) generally document a direct native rebuild path plus build-tool prerequisites, not only a full install retry. Please add:
- The package-scoped rebuild command
- A verification command like `node -e "require('better-sqlite3')"`
- Platform build-tool notes (python3, make, g++/clang++)

lefarcen (changes requested):
Hey @VikingOwl91, thanks for tackling this ABI mismatch pain point! Found 2 must-fix issues that would prevent the rebuild from running — see inline below.

QUICKSTART.md
```diff
@@ -214,6 +214,7 @@ open-design/
 
 ## Troubleshooting
 
+- **`better-sqlite3` fails to load / ABI mismatch after a Node.js version change** — `pnpm install` re-runs `postinstall` automatically and rebuilds the native addon for the current Node.js. To rebuild manually or verify the fix: `pnpm --filter @open-design/daemon rebuild better-sqlite3` then `node -e "require('better-sqlite3')"`. Requires build tools: `python3`, `make`, `g++` (or `clang++`). If you have `ignore-scripts=true` in your `.npmrc`, run `node scripts/postinstall.mjs` after `pnpm install`.
```
mrcfps: The verification command here runs `require('better-sqlite3')` from the workspace root, but the changed postinstall code correctly notes that `better-sqlite3` is a daemon dependency rather than a root dependency. In a normal pnpm workspace install, running this exact root-level `node -e` check fails with `MODULE_NOT_FOUND` even when the daemon dependency is installed and healthy, which would make the troubleshooting guidance look broken after the rebuild. Please scope the verification to the daemon package, for example `pnpm --filter @open-design/daemon exec node -e "require('better-sqlite3')"`, or require the daemon package's installed module path explicitly.

mrcfps (commented):
@VikingOwl91 thanks for the quick follow-up on the daemon dependency resolution and pnpm-managed rebuild path — the implementation issue from the earlier pass looks addressed. I found one remaining docs accuracy issue in the new troubleshooting command that is worth fixing so users can verify from the right workspace context.


lefarcen (approved):
Hey @VikingOwl91 — great follow-up! 🎉

Both P1 issues from my earlier review are now fixed:
- ✅ Package resolution scoped to daemon context via `createRequire`
- ✅ Rebuild delegated to `pnpm rebuild` (no sibling node-gyp assumption)

mrcfps's latest comment points to a docs accuracy issue on the verification command (P2, non-blocking). Once that's addressed, this is ready from my side.

Thanks for the careful iteration — looking forward to seeing this land! 🚀

lefarcen (approved):
Both P1 issues from my previous review are fixed — nice work @VikingOwl91! 🎉

The rebuild path now correctly resolves better-sqlite3 from the daemon workspace and delegates to pnpm's own node-gyp lifecycle. This is ready from my side.

VikingOwl91:
@lefarcen my first proper PR on a project worth PR-ing for :100:

mrcfps (approved):
@VikingOwl91 I reviewed the current changed ranges for the postinstall recovery path and QUICKSTART troubleshooting update. The daemon-scoped `better-sqlite3` resolution, pnpm-managed rebuild command, and package-scoped verification guidance address the earlier correctness issues, and I did not find any remaining actionable implementation concerns. Thanks for the careful follow-up here — this should make the Node ABI mismatch path much friendlier for users. 🎉


leilei926524-tech:
Hi @VikingOwl91 — welcome to Open Design contributors. 🎉

Your first merged PR (#813) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

Issue #267, opened 2026-05-02
OpenCode is installed but it's not being detected


Artaherzadeh:
OpenCode is installed for the current user but even after adding it to the PATH, it's not being detected by OpenDesign.

lefarcen:
Hi @Artaherzadeh! 👋

Thanks for opening this issue — I can see OpenCode detection is causing problems even though you've added it to PATH.

## What I found

Looking at the agent detection code (`apps/daemon/src/agents.ts:576`), Open Design searches for the `opencode` binary on your PATH using the standard `resolveOnPath` function. This should work if OpenCode is correctly installed and PATH is set.

## Similar issues

This is related to:
- **#89** — "Indefinitely stuck at starting OpenCode" (spawn/handshake issues)
- **#163** — "OpenCode CLI exits with code 1" (session lifecycle)
- **#198** — "Session not found" error
- **#235** — OpenClaude (Claude Code fork) detection

All point to OpenCode integration edge cases that need better diagnostics.

## Diagnostic steps

Could you help narrow this down by running these checks in your terminal?

### 1. Verify OpenCode is actually on PATH
```bash
which opencode
# or on Windows:
where opencode
```

If this returns nothing → OpenCode isn't on PATH yet. You'll need to:
- Restart your terminal after installation
- Or manually add OpenCode's install directory to PATH

### 2. Check if OpenCode responds to version check
```bash
opencode --version
```

If this hangs or errors → OpenCode installation might be broken.

### 3. Test if OpenCode can list models
```bash
opencode models
```

This should print a list of available models. If it times out or returns "no models" → OpenCode needs configuration (API keys, provider setup).

### 4. Share your setup
- **OS**: (macOS / Windows / Linux?)
- **OpenCode version**: (`opencode --version`)
- **Installation method**: (npm / binary download / other?)

## Next steps

Once we have the diagnostic output, I can pinpoint whether this is:
1. PATH detection issue (Open Design not seeing the binary)
2. OpenCode installation/config issue
3. Daemon integration bug (Open Design sees OpenCode but can't spawn it)

Looking forward to your findings! 🔍

dabing1022:
Hi @Artaherzadeh, this issue should now be resolved in the latest version. Please pull the latest code or download the latest release and give it another try. If OpenCode still isn't detected, feel free to reopen or follow up here.

lefarcen:
Thanks for jumping in, @dabing1022! 👍

[...]

Issue #235, opened 2026-05-01
Support OpenClaude


GarryLai:
OpenClaude is a fork based on (leaked) Claude Code. The command line is no difference but change the name, but open-design cannot detect and choose it.
https://github.com/Gitlawb/openclaude

lefarcen:
Hi @GarryLai! Thanks for opening this issue — OpenClaude support would be a great addition given it's a Claude Code fork.

From looking at `apps/daemon/src/agents.ts`, our agent detection currently keys off the binary name (`claude` for Claude Code, `codex` for Codex, etc.). OpenClaude uses a different command name (likely `openclaude`), so open-design doesn't pick it up automatically.

Could you share:
1. What's the command-line name for OpenClaude? (Is it `openclaude`, `claude` with a wrapper, or something else?)
2. Does it support the same `--version`, `--help`, and stdin prompt input that Claude Code does?
3. Any flags/options that differ from vanilla Claude Code?

If the CLI interface is identical to Claude Code (just different binary name), adding support would be straightforward — we'd add a new entry to `AGENT_DEFS` in `agents.ts`. If there are differences in flags or behavior, we'd need to document those so the integration works smoothly.

Thanks for making open-design better!

GarryLai:
Just change `claude` to `openclaude`, but the difference is that openclaude do not have a binary executable. It using npm to install.
https://github.com/Gitlawb/openclaude/blob/main/README.md#quick-start

lefarcen:
Thanks for the quick response! 🙏 This is super helpful info.

So OpenClaude is **npm-based** (not a system binary) — that changes the detection approach. Looking at `apps/daemon/src/agents.ts`, all current agents probe for a binary in PATH (`claude`, `codex`, `gemini`, etc.), but OpenClaude would be `npx openclaude` or `npm exec openclaude`.

**Two paths forward:**

### Path A: Add as npx wrapper (simpler, no code change)
Since OpenClaude's CLI interface is identical to Claude Code (same `--version`, `--help`, stdin prompt input), you could create a shell wrapper:

```bash
#!/bin/bash
# Save as ~/bin/openclaude (or anywhere in PATH)
exec npx openclaude "$@"
```

Then open-design would detect it as a `claude`-compatible agent. You'd just need to pick "Claude Code" in the agent selector and it would work (since the wrapper is named `openclaude`, not `claude`, it won't collide with real Claude Code if you have both).

[...]

PR #818, opened 2026-05-07
fix(daemon): prefer opencode-cli over opencode binary so OpenCode Desktop installs resolve to the CLI (#814)

### LINKED ISSUE #814: OpenDesign launching the desktop version of OpenCode and not the CLI
Hello,
If you have the desktop version of OpenCode, OpenDesign launches that instead of using the CLI with the result of not working / interacting with OpenCode.
When you have the Desktop version of OpenCode you have 2 launchers / binaries:

opencode-cli
opencode

if you launch "opencode" you end up in the Desktop application. To launch the cli version you need to launch opencode-cli.

A custom config for execution modes would be able to get around this problem.
Thank you in advance.

Sid-Qin:
## Summary

- OpenCode Desktop installs **two** binaries on PATH: `opencode` (the GUI launcher — clicking it opens a desktop window, never reads stdin) and `opencode-cli` (the headless CLI that speaks `run --format json --dangerously-skip-permissions -`).
- The OpenCode `AGENT_DEF` set `bin: 'opencode'`, so [`resolveAgentExecutable`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/agents.ts#L875) resolved the GUI launcher for every user with the desktop install. The daemon happily spawned it, but the GUI launcher discards stdin and never emits the structured JSON stream `opencode-stream` parses — so the agent silently produced **no output** for the entire conversation. That's the symptom @giuliastro reports in #814.
- Fix: declare `bin: 'opencode-cli'` and `fallbackBins: ['opencode']` on the adapter. Same mechanism Claude Code already uses to fall back to `openclaude` (issue #235), so this is well-trodden ground in the codebase.

Fixes #814.

## Resolution semantics

| User shape | `bin` (`opencode-cli`) | `fallbackBins[0]` (`opencode`) | Resolved |
|---|---|---|---|
| OpenCode Desktop install (both on PATH) | ✓ | ✓ | `opencode-cli` (CLI) ← #814 fix path |
| Legacy CLI-only install (no desktop, no `-cli` suffix) | ✗ | ✓ | `opencode` (CLI in this shape) — no regression |
| Nothing installed | ✗ | ✗ | `null` — same as before |

## Diff shape

| File | What changes |
|---|---|
| `apps/daemon/src/agents.ts` | `bin: 'opencode'` → `bin: 'opencode-cli'`; add `fallbackBins: ['opencode']`; doc-comment block citing #814 + the #235 precedent. |
| `apps/daemon/tests/agents.test.ts` | One declarative test pins `opencode.bin === 'opencode-cli'` + `opencode.fallbackBins.includes('opencode')` (runs on every platform). Two filesystem-backed `fsTest`s pin the resolution priority: both-on-PATH picks `opencode-cli`; legacy-only falls back to `opencode`. Existing "most agents have a single binary" comment list updated to remove `opencode`. |

2 files, +97 / −2.

## What's verified

| Check | Result |
|---|---|
| Daemon vitest (full suite) | **993/993** |
| Daemon `tsc -p tsconfig.json --noEmit` | clean |
| Daemon `tsc -p tsconfig.tests.json --noEmit` (the CI killer that broke a previous PR) | clean |

## What's NOT verified

I don't have OpenCode Desktop installed locally, so I can't end-to-end exercise "click → open project → confirm CLI binary spawns instead of GUI". The structural argument:

1. The fix is the same `fallbackBins` mechanism Claude Code uses for OpenClaude (#235), already battle-tested in production.
2. The new filesystem-backed tests construct both PATH layouts (desktop with both binaries; legacy with only the bare name) and assert which file path `resolveAgentExecutable` returns — that's the exact decision the daemon makes at spawn time.
3. `opencode-cli` is documented by OpenCode Desktop as the headless CLI binary; the assumption that it speaks the same `run --format json` protocol as the legacy `opencode` follows from the user report (#814) and the OpenCode docs the issue thread cites.

If `opencode-cli --version` / `opencode-cli run --help` reveals subcommand differences vs the legacy `opencode` binary that @lefarcen flagged in the issue, this PR can be narrowed (e.g. version-gated `buildArgs`) — but for now the symptom #814 reports is "spawning the wrong binary entirely," not "spawning the right binary with the wrong flags," so the fallbackBins flip is the minimum fix.

## Test plan

- [x] Existing daemon test suite stays green (993/993).
- [x] One declarative test asserts adapter shape on every platform.
- [x] Two `fsTest`s pin the resolution priority for both PATH layouts.
- [x] Daemon typecheck (both `tsconfig.json` and `tsconfig.tests.json`) clean.
- [ ] Manual smoke from @giuliastro (or any OpenCode Desktop user): with the desktop install on PATH, an OpenCode-agent run actually produces output instead of the GUI launching silently.
- [ ] Smoke from a legacy `opencode`-only install (no `-cli` suffix) to confirm the fallback resolves and runs unchanged.

## Notes

- This is related to #267 ("OpenCode is installed but it's not being detected") but addresses a different shape: in #267 the desktop install appeared to provide no spawnable CLI at all, while #814 confirms the CLI exists under a different binary name. With this PR landed, #267 might be re-testable too.
- @lefarcen — your diagnosis on the issue is exactly what shaped this PR. Tagging in case you want to confirm the binary-rename hypothesis before merge, or wait on @giuliastro's `opencode-cli --version` output you asked for.

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -380,7 +380,24 @@ export const AGENT_DEFS = [
   {
     id: 'opencode',
     name: 'OpenCode',
-    bin: 'opencode',
+    // OpenCode Desktop (https://opencode.dev) ships two binaries when
+    // installed: `opencode` is the GUI launcher (clicking it opens a
+    // desktop app, not a stdin-driven CLI), and `opencode-cli` is the
+    // headless CLI that speaks the `run --format json …` protocol the
+    // daemon expects. Resolving `opencode` first ends up spawning the
+    // desktop launcher, which doesn't read stdin and never produces
+    // JSON events — so the agent silently does nothing for any user
+    // with the desktop install (issue #814).
+    //
+    // Resolve `opencode-cli` first, then fall back to bare `opencode`
+    // for the legacy CLI-only install (no desktop app), where there is
+    // no `-cli` suffix and the bare name is the real CLI.
+    // `resolveAgentExecutable` walks `bin` then `fallbackBins` in
+    // order, so this gives us "prefer the always-CLI binary, fall
+    // back to the historical name" — same mechanism Claude Code uses
+    // to fall back to `openclaude` (issue #235).
+    bin: 'opencode-cli',
+    fallbackBins: ['opencode'],
     versionArgs: ['--version'],
     // `opencode models` prints `provider/model` per line.
     listModels: {
--- a/apps/daemon/tests/agents.test.ts
+++ b/apps/daemon/tests/agents.test.ts
@@ -58,6 +58,7 @@ const pi = requireAgent('pi');
 const deepseek = requireAgent('deepseek');
 const gemini = requireAgent('gemini');
 const qoder = requireAgent('qoder');
+const opencode = requireAgent('opencode');
 const deepseekMaxPromptArgBytes = deepseek.maxPromptArgBytes;
 assert.ok(
   deepseekMaxPromptArgBytes !== undefined,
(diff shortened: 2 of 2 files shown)


lefarcen (commented):
Hi @Sid-Qin! 👋 Thanks for this fix — the desktop-vs-CLI distinction is exactly the right diagnosis for #814.

## Code correctness (Lens A)

The implementation is clean:

✅ `bin: 'opencode-cli'` + `fallbackBins: ['opencode']` follows the same precedent as Claude Code (#235) — already battle-tested.

✅ Three tests pin the resolution priority (declarative shape check, desktop-both-on-PATH, legacy-only), so the fallback logic is fully exercised.

✅ `resolveAgentExecutable` walks `bin` then `fallbackBins` in order, which is exactly what this needs.

No runtime bugs found.

## Minor suggestion (non-blocking)

**P3 docs drift** — `specs/current/runtime-adapter.md`, `docs/agent-adapters.md`, and the two README sections (CLI install, coding-agent comparison) still say the daemon runs/resolves `opencode`. With this PR landed, those docs will be stale: OpenCode Desktop users will have `opencode-cli` resolved first, but the public adapter matrix won't document it.

Concrete fix (can be a follow-up):
- Update the adapter docs/specs to say OpenCode resolves `opencode-cli` first and falls back to `opencode`.
- Adjust command examples to use the resolved OpenCode CLI binary rather than hard-coding only `opencode`.

This is a P3 because it doesn't affect correctness — just internal/external docs accuracy. If you want to knock it out in this PR, great; otherwise it's a small follow-up doc PR once this lands and the desktop path is confirmed working.

## Verification gap

You called out that you don't have OpenCode Desktop locally to end-to-end test. That's fine — the structural argument (same `fallbackBins` mechanism as Claude Code + filesystem-backed tests pinning both PATH layouts) is solid. If @giuliastro (the #814 reporter) or any OpenCode Desktop user can smoke-test after merge, that closes the loop.

---

**Overall**: LGTM from the bot's side. Deferring final approval to a maintainer for the usual sign-off.

Sid-Qin:
@lefarcen thanks for the review!

**On the CI red:** the `Validate workspace` failure isn't from this PR — it's a pre-existing typecheck regression on main. The Indonesian locale ([fcc37c6](https://github.com/nexu-io/open-design/commit/fcc37c6)) and the Cloudflare Pages deploy keys ([09eb88f](https://github.com/nexu-io/open-design/commit/09eb88f)) both landed around the same time but didn't conflict at merge — leaving 18 keys unset on `id.ts` and turning every open PR's typecheck red.

I opened **#822** with a focused fix that defaults `id.ts` to the English spread pattern the way 8 other locales already do, so future feature PRs adding new strings won't trigger the same race. Once #822 merges, this PR's CI should go green on the next re-run.

**On the P3 docs drift:** good catch. I'd prefer to land that as a follow-up after this merges so the OpenCode binary fix doesn't get blocked on README/spec edits — happy to open a separate PR for `specs/current/runtime-adapter.md`, `docs/agent-adapters.md`, and the README sections. Or if a maintainer prefers them in this PR, I'll squash them in.

Also noticed `AGENT_BIN_ENV_KEYS = new Map([['codex', 'CODEX_BIN']])` landed on main since I last looked. That's the env-var override mechanism you originally suggested for OpenCode in the issue thread — could add `['opencode', 'OPENCODE_BIN']` here as a belt-and-suspenders to the fallbackBins flip if you'd want it. Let me know.

mrcfps (dismissed):
@Sid-Qin I reviewed the OpenCode adapter change and the new resolver coverage. The primary binary now prefers `opencode-cli` while preserving the legacy `opencode` fallback, and the added tests exercise both the desktop two-binary layout and the legacy single-binary layout. Thanks for the focused fix and the clear regression coverage — this looks safe to merge once the unrelated workspace validation failure is handled. 🙂


lefarcen (commented):
Re-reviewed at head ed0986b3 (force-push after the prior 96a07f5c review).

The previous P3 finding (docs drift — test comment mentioning opencode in the 'most agents have single binary' list) is resolved in the current diff. The test comment now correctly notes that opencode used to be in that list but now declares fallbackBins for the desktop-vs-CLI case.

No new correctness, security, or documentation issues found. The fix remains sound: preferring `opencode-cli` with `opencode` fallback correctly handles both the OpenCode Desktop two-binary layout and the legacy CLI-only install.

LGTM from the bot's side; deferring final approval to a maintainer.

lefarcen:
@mrcfps 这个 PR 的最新 commit (b0dd4abc) 还在等你看一下。Author rebased 解决了 merge conflict，CI 已绿。

mrcfps (approved):
Approved.

PR #805, opened 2026-05-07
Migrate beta release publishing to R2

PerishCode:
## Summary

This PR finishes two related release-track items: it reduces the mac packaged release payload, then moves beta publishing away from GitHub Releases/tags and onto the R2 beta channel.

Core changes:

- Moves beta version state to R2 `beta/latest/metadata.json`; missing metadata now falls back to `${baseVersion}-beta.0` and no longer reads `latest-mac.yml` as a state source.
- Publishes beta assets only to R2 under `beta/latest/...` and `beta/versions/<beta>/...`.
- Keeps mac and Windows beta publishing enabled by default; Linux is opt-in and intentionally minimal for now, publishing AppImage/checksum to R2 without an updater feed.
- Adds independent `metadata.json` for GitHub Actions beta state and release information collection.
- Shrinks the mac packaged app by prebundling the standalone web runtime, prebundling daemon entrypoints, pruning Electron locales, excluding generated/dev residue, and using maximum release artifact compression in beta publishing.
- Keeps Windows tools-pack cache strategy and aligns mac with a high-ROI `mac.workspace-build` cache while leaving signing/notarization and final DMG/ZIP outputs uncached.
- Splits mac tools-pack packaging out of the deleted `tools/pack/src/mac.ts` into `tools/pack/src/mac/index.ts` plus focused build, lifecycle, builder, artifact, report, workspace, fs, paths, and type modules.
- Removes the unexpected `tools/launcher` directory and README references; guard now treats top-level `tools/` as an allowlist with only `AGENTS.md`, `dev/`, and `pack/`.
- Fixes mac afterPack portability for copied Next standalone pnpm symlinks by rewriting copied standalone symlinks inside the packaged resource and auditing for external/broken symlinks.

## Quantified Impact

- mac packaged app size after this PR's prebundle + locale pruning pass is roughly `586M -> 431M`, about `155M` smaller (`~26%`).
- `Open Design.app/Contents/Resources/app/node_modules` drops roughly `141M -> 2.6M`, removing about `138M` of packaged Node dependency payload (`~98%`).
- The earlier low-risk residue pass in the same size effort removed packaged sourcemaps and `better-sqlite3` build residue:
  - `.app`: `704.1 -> 610.3 MiB` (`-93.8 MiB`, `-13.3%`)
  - DMG: `744.3 -> 638.7 MiB` (`-105.6 MiB`, `-14.2%`)
[...]

PR #819, opened 2026-05-07
feat(daemon): Critique Theater Phase 6.1 (interrupt endpoint + run registry)

Nagendhra-web:
Phase 6.1 of the Critique Theater rollout per [specs/current/critique-theater-plan.md](https://github.com/nexu-io/open-design/blob/main/specs/current/critique-theater-plan.md). Scoped down from the original Phase 6 PR after lefarcen's review surfaced architectural issues with the rerun draft. Task 6.2 (rerun) is now a separate follow-up.

## What ships

`POST /api/projects/:projectId/critique/:runId/interrupt`. Cascades an `AbortController` to the orchestrator that owns the spawned CLI so the parser can flush best-so-far state and emit `critique.interrupted` before the process exits. Backed by a new in-process run registry that the orchestrator wiring registers each run into before `runOrchestrator` is invoked, and unregisters in a `finally` block.

## Lefarcen review fixes baked in

**Registry now keyed by `(projectId, runId)`, not just runId** (was P2). All registry methods (`get`, `interrupt`, `unregister`) take both arguments and use a composite internal key. A request for project p1's runId cannot match a handle from project p2 even if the ids collide. Defense in depth on top of the handler's own DB-row projectId check, with explicit cross-project tests at both the registry and HTTP layers.

**Interrupt is idempotent on already-interrupted rows** (was P1). A client that lost the first response and retries observes `202 prevStatus="interrupted"` rather than `409`. Other terminal statuses (shipped, failed, timed_out, degraded, below_threshold, legacy) still return 409 because those runs reached their real terminal state on their own. Test added for both the happy retry path and the stale-handle defense.

**Register-before-spawn race revisited** (was P2). After re-reading the actual server.ts wiring, `runRegistry.register` runs *before* the `await runOrchestrator(...)` call (which is the action that inserts the DB row). The interrupt handler does its DB lookup first and 404s if the row does not exist yet, so the only window where the registry exists but the DB does not is between `register` and the orchestrator's first DB insert. A request landing in that window 404s on the DB lookup, which is the correct outcome (the run is not visible yet from the public contract's perspective). No exploitable race.

## Why Task 6.2 (rerun) is not in this PR

The earlier draft inserted a `critique_runs` row in `rerun-handler.ts` and then handed control to a `startCritiqueRun` callback that called `runOrchestrator`. `runOrchestrator`'s first action is to insert the same runId — primary key conflict on the fire-and-forget promise that wasn't surfaced to the 202 response. The handler also did not actually spawn an agent or attach the prior artifact as context. Fixing rerun properly needs:

- A real chat-run path that spawns the agent with prior-art context.
- An `artifactId` format validator and SQL LIKE escaping for the row lookup.
- Atomic row creation that rolls back on spawn failure.

That is a scoped follow-up PR, not three lines on top of this one. Filing it separately keeps this PR small and the rerun design open for review.

## Process-tree teardown is still deferred

Abort/timeout/error paths in the orchestrator call `child.kill('SIGTERM')` on the immediate child only. Grandchildren spawned by the agent CLI (Claude Code, npx wrappers, ...) can survive. Wiring `collectProcessTreePids` from `@open-design/platform` is the same surface `tools-pack` uses today and belongs in the next PR alongside rerun. The interrupt endpoint contract here is "abort the orchestrator-level state machine and persist `interrupted`," which is what the tests assert.

## Files

- New: [run-registry.ts](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6-clean/apps/daemon/src/critique/run-registry.ts), [interrupt-handler.ts](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6-clean/apps/daemon/src/critique/interrupt-handler.ts).
- Modified: [server.ts](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6-clean/apps/daemon/src/server.ts) (registry wire-up around `runOrchestrator`, one route registration).
- Tests: [critique-run-registry.test.ts](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6-clean/apps/daemon/tests/critique-run-registry.test.ts) (17 cases), [critique-interrupt-endpoint.test.ts](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6-clean/apps/daemon/tests/critique-interrupt-endpoint.test.ts) (16 cases).

## Incidental

[apps/web/src/i18n/locales/id.ts](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6-clean/apps/web/src/i18n/locales/id.ts) was missing 18 `fileViewer.deploy*` / `fileViewer.cloudflare*` keys after #805 (R2 release publishing) landed deploy provider abstractions. The web Dict typecheck enforces locale parity, so without those keys the workspace typecheck fails on every PR. Added Indonesian translations for the missing keys to unblock CI.

Local: registry + interrupt suites 33/33, daemon and web typecheck, `pnpm guard` all clean.

### DIFF
--- a/apps/daemon/src/critique/interrupt-handler.ts
+++ b/apps/daemon/src/critique/interrupt-handler.ts
@@ -1,0 +1,129 @@
+import type { Request, Response } from 'express';
+import type Database from 'better-sqlite3';
+import {
+  getCritiqueRun,
+  markRunInterruptedRecovery,
+  type CritiqueRunStatus,
+} from './persistence.js';
+import type { RunRegistry } from './run-registry.js';
+
+/** HTTP status codes used by the interrupt endpoint. */
+const HTTP_BAD_REQUEST = 400;
+const HTTP_NOT_FOUND = 404;
+const HTTP_CONFLICT = 409;
+const HTTP_ACCEPTED = 202;
+
+/**
+ * POST /api/projects/:projectId/critique/:runId/interrupt
+ *
+ * Validates the run exists and belongs to the URL project, then signals the
+ * registered AbortController so the orchestrator can flush best-so-far state
+ * and emit critique.interrupted.
+ *
+ * Idempotency: if the row is already 'interrupted', the endpoint returns 202
+ * with prevStatus='interrupted' rather than 409, so a client that lost the
+ * first response and retries does not see the run flip from 202 to a hard
+ * conflict. Other terminal statuses (shipped, failed, timed_out, degraded,
+ * below_threshold, legacy) still return 409 because the run reached its real
+ * terminal state on its own and an interrupt is no longer meaningful.
+ *
+ * @see specs/current/critique-theater.md § interrupt endpoint (Task 6.1)
+ */
+export function handleCritiqueInterrupt(
+  db: Database.Database,
+  registry: RunRegistry,
+): (req: Request, res: Response) => void {
+  return function critiqueInterruptHandler(req: Request, res: Response): void {
+    const projectId =
+      typeof req.params['projectId'] === 'string'
+        ? req.params['projectId'].trim()
+        : '';
+    const runId =
+      typeof req.params['runId'] === 'string'
+        ? req.params['runId'].trim()
+        : '';
+
+    if (!projectId || !runId) {
+      res
+        .status(HTTP_BAD_REQUEST)
+        .json({ error: { code: 'BAD_REQUEST', message: 'projectId and runId are required' } });
+      return;
+    }
+
+    const row = getCritiqueRun(db, runId);
+
+    // Cross-project leak guard: a request to interrupt project p1's runId
+    // must NOT find a row that actually belongs to project p2. Returning 404
+    // (not 403) avoids leaking the existence of other projects' runs.
+    if (row === null || row.projectId !== projectId) {
+      res
+        .status(HTTP_NOT_FOUND)
+        .json({ error: { code: 'NOT_FOUND', message: 'critique run not found' } });
+      return;
+    }
+
+    const liveStatus = row.status as CritiqueRunStatus | 'running';
+
+    if (liveStatus === 'interrupted') {
+      // Idempotent retry path. The original interrupt already drove the run
+      // into the terminal 'interrupted' state; a duplicate request from a
+      // retrying client should observe the same accepted outcome rather than
+      // a 409.
+      res.status(HTTP_ACCEPTED).json({
+        runId,
+        accepted: true,
+        prevStatus: 'interrupted',
+      });
+      return;
+    }
+
+    if (liveStatus !== 'running') {
+      res
+        .status(HTTP_CONFLICT)
+        .json({
+          error: {
+            code: 'CONFLICT',
+            message: `run is already in terminal status: ${row.status}`,
+            currentStatus: row.status,
+          },
+        });
+      return;
+    }
+
+    // Project-keyed registry call: a request to interrupt project p1's runId
+    // cannot match a registry handle from project p2 even if a runId
+    // collision somehow occurred.
+    const aborted = registry.interrupt(projectId, runId, 'user_requested');
+
+    if (!aborted) {
+      // The DB row says 'running' but the in-process registry has no live
+      // AbortController for it. This happens after a daemon restart, in
+      // the window before reconcileStaleRuns considers the row old enough
+      // to flip to 'interrupted' on its own. Without this branch the
+      // endpoint would lie: 202 accepted, but no child is signaled, no
+      // critique.interrupted event is emitted, and the row stays
+      // 'running' until reconcileStaleRuns finally catches it.
+      //
+      // Recovery path: mark the row 'interrupted' directly with
+      // recoveryReason='no_live_handle' (mirroring how reconcileStaleRuns
+      // writes 'daemon_restart'), so the row's terminal state matches
+      // what the user asked for and the response carries the recovered
+      // flag for clients that want to distinguish the two paths.
+      const recovered = markRunInterruptedRecovery(db, runId, 'no_live_handle');
+      res.status(HTTP_ACCEPTED).json({
+        runId,
+        accepted: true,
+        prevStatus: 'running',
+        recovered: true,
+        ...(recovered ? {} : { recoveryFailed: true }),
+      });
+      return;
+    }
+
+    res.status(HTTP_ACCEPTED).json({
+      runId,
+      accepted: true,
+      prevStatus: 'running',
+    });
+  };
+}
--- a/apps/daemon/src/critique/persistence.ts
+++ b/apps/daemon/src/critique/persistence.ts
@@ -310,6 +310,43 @@ export function deleteCritiqueRun(db: Database.Database, id: string): void {
 }
 
 /**
+ * Marks a single 'running' row as 'interrupted' with the supplied
+ * recoveryReason embedded in rounds_json. Mirror of the per-row write in
+ * reconcileStaleRuns(), kept as its own function so the interrupt endpoint
+ * can use it when a request arrives for a row that has no live
+ * AbortController in the registry (the post-daemon-restart window before
+ * reconcileStaleRuns considers the row old enough). Atomic on the
+ * status='running' guard so a row that just transitioned to a different
+ * terminal state is not overwritten.
+ *
+ * Returns true when a row was mutated, false when the id was missing or
+ * not in 'running' status.
+ */
+export function markRunInterruptedRecovery(
+  db: Database.Database,
+  id: string,
+  recoveryReason: string,
+  now: number = Date.now(),
+): boolean {
+  const existing = db
+    .prepare(`SELECT ${COLS} FROM critique_runs WHERE id = ? AND status = 'running'`)
+    .get(id) as RawCritiqueRunRow | undefined;
+  if (existing === undefined) return false;
+  const { rounds } = parseRoundsPayload(existing.roundsJson);
+  const newPayload = serializeRoundsPayload(rounds, recoveryReason);
+  const result = db
+    .prepare(
+      `UPDATE critique_runs
+          SET status = 'interrupted',
+              rounds_json = ?,
+              updated_at = ?
+        WHERE id = ? AND status = 'running'`,
+    )
+    .run(newPayload, now, id);
+  return result.changes > 0;
+}
+
+/**
  * Recovery scan called on daemon boot: any run still in a non-terminal status
  * older than staleAfterMs is marked 'interrupted' with rounds_json.recoveryReason
  * = 'daemon_restart'. Returns the count of rows mutated.
--- a/apps/daemon/src/critique/run-registry.ts
+++ b/apps/daemon/src/critique/run-registry.ts
@@ -1,0 +1,108 @@
+/**
+ * In-process registry of in-flight critique runs. The daemon process is the
+ * single owner of all critique state; the registry exists so the interrupt
+ * endpoint can cascade an AbortController to the orchestrator that owns the
+ * spawned CLI. The registry is intentionally NOT persisted: a daemon restart
+ * mid-run is handled by reconcileStaleRuns on boot, not by recovering live
+ * AbortControllers.
+ *
+ * All lookup operations require BOTH projectId and runId. The composite key
+ * prevents a request to interrupt project p1's runId from accidentally
+ * aborting project p2's run that happens to share the same id (defense in
+ * depth on top of the HTTP handler's own DB-row projectId check).
+ *
+ * @see specs/current/critique-theater.md § Failure modes (interrupt)
+ */
+
+/** Handle for a single in-flight critique run. */
+export interface RunHandle {
+  runId: string;
+  projectId: string;
+  abort: AbortController;
+  startedAt: number;
+}
+
+/** Public surface of the in-process run registry. */
+export interface RunRegistry {
+  /**
+   * Register a new in-flight handle. Throws if a handle for the same
+   * (projectId, runId) is already registered (indicates a bug in the caller,
+   * not a user error).
+   */
+  register(handle: RunHandle): void;
+
+  /**
+   * Returns the handle if the (projectId, runId) pair is registered; null
+   * otherwise. A runId from a different project will not match.
+   */
+  get(projectId: string, runId: string): RunHandle | null;
+
+  /**
+   * Signals the AbortController for the given (projectId, runId).
+   * Returns true if the pair was found and aborted; false otherwise. A
+   * runId-only match against a different project does NOT abort.
+   */
+  interrupt(projectId: string, runId: string, reason?: string): boolean;
+
+  /**
+   * Removes the entry for the given (projectId, runId). Called by the server
+   * after the orchestrator settles. No-op if the pair is not registered.
+   */
+  unregister(projectId: string, runId: string): void;
+
+  /**
+   * Snapshot for diagnostics only. Returns a defensive copy so callers cannot
+   * mutate the registry's internal state.
+   */
+  list(): RunHandle[];
+}
+
+/**
+ * Builds the internal composite key for a (projectId, runId) pair. Pipe is
+ * not a legal character in either projectId or runId per the daemon's id
+ * generation rules, so collisions across pairs are impossible.
+ */
+function compositeKey(projectId: string, runId: string): string {
+  return `${projectId}|${runId}`;
+}
+
+/**
+ * Creates an in-memory RunRegistry backed by a Map.
+ * Node is single-threaded; no locking is needed.
+ *
+ * @see specs/current/critique-theater.md § interrupt endpoint (Task 6.1)
+ */
+export function createRunRegistry(): RunRegistry {
+  const store = new Map<string, RunHandle>();
+
+  return {
+    register(handle: RunHandle): void {
+      const key = compositeKey(handle.projectId, handle.runId);
+      if (store.has(key)) {
+        throw new Error(
+          `RunRegistry: duplicate (projectId="${handle.projectId}", runId="${handle.runId}"); unregister before re-registering`,
+        );
+      }
+      store.set(key, handle);
+    },
+
+    get(projectId: string, runId: string): RunHandle | null {
+      return store.get(compositeKey(projectId, runId)) ?? null;
+    },
+
+    interrupt(projectId: string, runId: string, reason?: string): boolean {
+      const handle = store.get(compositeKey(projectId, runId));
+      if (handle === undefined) return false;
+      handle.abort.abort(reason);
+      return true;
+    },
+
+    unregister(projectId: string, runId: string): void {
+      store.delete(compositeKey(projectId, runId));
+    },
+
+    list(): RunHandle[] {
+      return [...store.values()];
+    },
+  };
+}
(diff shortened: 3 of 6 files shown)


lefarcen (changes requested):
Hey @Nagendhra-web, thanks for pushing Phase 6 forward. I reviewed the interrupt/rerun paths and found **7 issues blocking merge** — mostly P1 atomicity/correctness gaps contradicting the PR's claims. The register-before-spawn race, rerun atomicity, and idempotent-interrupt guarantees aren't currently upheld.

## P1 findings

**1. Rerun path is broken and not atomic** (, )

The handler inserts a `critique_runs` row *before* `startCritiqueRun`, then the server's `startCritiqueRun` calls `runOrchestrator`, which immediately tries to insert the *same* `runId` again (primary-key failure on the fire-and-forget promise). That rejection is only followed by `.finally(...)`, not caught, so the caller already got `202` while a `running` orphan row is left behind.

The PR says "new artifact row committed only after spawn handle resolves" — opposite is true. It also doesn't spawn an agent, build a fresh artifact id, or attach prior-art context.

**Fix:** Make rerun start the same real chat/spawn path as initial critique, have the spawn handle resolve before committing the new row, OR make `runOrchestrator` support a pre-created row explicitly and handle spawn/orchestrator failure by rolling back or marking the row `failed`.

**2. Interrupt endpoint does not await persistence flush and is not idempotent across completed interrupt** (`apps/daemon/src/critique/interrupt-handler.ts:51-69`)

This calls `registry.interrupt(...)` and immediately returns `202`; there's no promise linking the endpoint to `runOrchestrator`'s final `updateCritiqueRun(...)`. If the first interrupt finishes quickly, a second call sees status `interrupted` and returns `409`, contradicting "calling twice returns 202" claim.

It also ignores `registry.interrupt`'s boolean, so a `running` DB row with no in-process handle still returns accepted without aborting anything.

**Fix:** Store an `interrupt(): Promise<final row>` or `settled` promise in the registry handle, await it (or at least await the orchestrator's flush boundary), and make repeated interrupts for the same interrupted run return the documented idempotent `202`.

## P2 findings

**3. Register-before-spawn race is not fixed** (`apps/daemon/src/server.ts:4090-4100`, `apps/daemon/src/server.ts:4174-4201`)

The child process is spawned and assigned to `run.child` around lines 4090-4100, but the critique registry entry isn't created until lines 4174-4180, and the DB `running` row isn't inserted until `runOrchestrator` starts.

An interrupt arriving in that window can 404 because `getCritiqueRun` cannot see the run yet, despite a live child already existing.

**Fix:** Create/register the critique handle and durable running state *before* spawning, or make the interrupt endpoint able to resolve the chat run and pending child during this pre-orchestrator window.

**4. Registry is not keyed by (projectId, runId) despite the PR claim** (`apps/daemon/src/critique/run-registry.ts:55-80`, `apps/daemon/src/critique/interrupt-handler.ts:42-67`)

`projectId` is stored on the handle but never participates in `get`, `interrupt`, or `unregister`; all lookups are run-id-only. The HTTP handler currently mitigates this by checking the DB row's `projectId` before interrupting, but the registry abstraction itself still permits future cross-project misuse and the tests encode run-id-only semantics.

**Fix:** Change registry methods to accept `projectId` and key by both values, then add tests showing identical run IDs in different projects do not collide and wrong-project interrupts do not abort.

**5. Deferred process-tree teardown leaves a real resource leak risk** (`apps/daemon/src/critique/orchestrator.ts:64-68`, `apps/daemon/src/critique/orchestrator.ts:141-142`, `apps/daemon/src/critique/orchestrator.ts:377-380`)

Abort/timeout/error paths only call `child.kill('SIGTERM')`; any grandchildren spawned by the CLI can continue running, consuming tokens/CPU or writing files after the daemon has persisted `interrupted`/`failed`.

Deferring full process-tree collection may be acceptable as a phase boundary, but this PR should not claim cancellation is sufficient for real adapter processes.

**Fix:** Either wire process-tree termination now via `@open-design/platform` OR explicitly mark the endpoint as best-effort and add monitoring/reconciliation for surviving descendants.

**6. Artifact lookup and rerun paths accept unescaped route data** (`apps/daemon/src/critique/persistence.ts:323-335`, `apps/daemon/src/critique/rerun-handler.ts:111-112`)

`getLatestCritiqueRunByArtifact` builds a `LIKE '%/${artifactId}/%'` pattern without escaping `%` or `_`, so crafted artifact IDs can match unrelated rows within the same project.

The rerun handler then builds `artifactDir` by string concatenation from the raw route param instead of validating an artifact-id grammar and using a fresh generated artifact directory.

**Fix:** Validate `artifactId` against the repository's artifact ID format, escape SQL `LIKE` wildcards or match normalized path segments, and generate a fresh rerun artifact id/path.

## P3 findings

**7. The new tests don't cover the PR's highest-risk claims** (`apps/daemon/tests/critique-interrupt-endpoint.test.ts:107-136`, `apps/daemon/tests/critique-rerun-endpoint.test.ts:131-207`, `apps/daemon/tests/critique-run-registry.test.ts:32-75`)

They test handler stubs and run-id-only registry behavior, but not:
- register-before-spawn ordering in the real server
- cross-project registry keying
- a second interrupt after persistence changes to `interrupted`
- rerun spawn failure atomicity
- endpoint-to-orchestrator abort propagation

The rerun tests also assert the current pre-inserted `running` row behavior, which is the opposite of the claimed atomicity.

**Fix:** Add integration-style tests around the real server spawn/orchestrator seam and failure injection tests where `startCritiqueRun` throws/rejects.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for pushing Phase 6 forward. I reviewed the interrupt/rerun paths and found **7 issues blocking merge** — mostly P1 atomicity/correctness gaps contradicting the PR's claims. The register-before-spawn race, rerun atomicity, and idempotent-interrupt guarantees aren't currently upheld.

## P1 findings

**1. Rerun path is broken and not atomic** (apps/daemon/src/critique/rerun-handler.ts:114-126, apps/daemon/src/server.ts:4440-4465)

The handler inserts a critique_runs row *before* startCritiqueRun, then the server's startCritiqueRun calls runOrchestrator, which immediately tries to insert the *same* runId again (primary-key failure on the fire-and-forget promise). That rejection is only followed by .finally(...), not caught, so the caller already got 202 while a running orphan row is left behind.

The PR says "new artifact row committed only after spawn handle resolves" — opposite is true. It also doesn't spawn an agent, build a fresh artifact id, or attach prior-art context.

Fix: Make rerun start the same real chat/spawn path as initial critique, have the spawn handle resolve before committing the new row, OR make runOrchestrator support a pre-created row explicitly and handle spawn/orchestrator failure by rolling back or marking the row failed.

**2. Interrupt endpoint does not await persistence flush and is not idempotent across completed interrupt** (apps/daemon/src/critique/interrupt-handler.ts:51-69)

This calls registry.interrupt(...) and immediately returns 202; there's no promise linking the endpoint to runOrchestrator's final updateCritiqueRun(...). If the first interrupt finishes quickly, a second call sees status interrupted and returns 409, contradicting "calling twice returns 202" claim.

It also ignores registry.interrupt's boolean, so a running DB row with no in-process handle still returns accepted without aborting anything.

Fix: Store an interrupt(): Promise<final row> or settled promise in the registry handle, await it (or at least await the orchestrator's flush boundary), and make repeated interrupts for the same interrupted run return the documented idempotent 202.

## P2 findings

**3. Register-before-spawn race is not fixed** (apps/daemon/src/server.ts:4090-4100, apps/daemon/src/server.ts:4174-4201)

The child process is spawned and assigned to run.child around lines 4090-4100, but the critique registry entry isn't created until lines 4174-4180, and the DB running row isn't inserted until runOrchestrator starts.

An interrupt arriving in that window can 404 because getCritiqueRun cannot see the run yet, despite a live child already existing.

Fix: Create/register the critique handle and durable running state *before* spawning, or make the interrupt endpoint able to resolve the chat run and pending child during this pre-orchestrator window.

**4. Registry is not keyed by (projectId, runId) despite the PR claim** (apps/daemon/src/critique/run-registry.ts:55-80, apps/daemon/src/critique/interrupt-handler.ts:42-67)

projectId is stored on the handle but never participates in get, interrupt, or unregister; all lookups are run-id-only. The HTTP handler currently mitigates this by checking the DB row's projectId before interrupting, but the registry abstraction itself still permits future cross-project misuse and the tests encode run-id-only semantics.

Fix: Change registry methods to accept projectId and key by both values, then add tests showing identical run IDs in different projects do not collide and wrong-project interrupts do not abort.

**5. Deferred process-tree teardown leaves a real resource leak risk** (apps/daemon/src/critique/orchestrator.ts:64-68, apps/daemon/src/critique/orchestrator.ts:141-142, apps/daemon/src/critique/orchestrator.ts:377-380)

Abort/timeout/error paths only call child.kill('SIGTERM'); any grandchildren spawned by the CLI can continue running, consuming tokens/CPU or writing files after the daemon has persisted interrupted/failed.

Deferring full process-tree collection may be acceptable as a phase boundary, but this PR should not claim cancellation is sufficient for real adapter processes.

Fix: Either wire process-tree termination now via @open-design/platform OR explicitly mark the endpoint as best-effort and add monitoring/reconciliation for surviving descendants.

**6. Artifact lookup and rerun paths accept unescaped route data** (apps/daemon/src/critique/persistence.ts:323-335, apps/daemon/src/critique/rerun-handler.ts:111-112)

getLatestCritiqueRunByArtifact builds a LIKE '%/artifactId/%' pattern without escaping % or _, so crafted artifact IDs can match unrelated rows within the same project.

The rerun handler then builds artifactDir by string concatenation from the raw route param instead of validating an artifact-id grammar and using a fresh generated artifact directory.

Fix: Validate artifactId against the repository's artifact ID format, escape SQL LIKE wildcards or match normalized path segments, and generate a fresh rerun artifact id/path.

## P3 findings

**7. The new tests don't cover the PR's highest-risk claims** (apps/daemon/tests/critique-interrupt-endpoint.test.ts:107-136, apps/daemon/tests/critique-rerun-endpoint.test.ts:131-207, apps/daemon/tests/critique-run-registry.test.ts:32-75)

They test handler stubs and run-id-only registry behavior, but not: register-before-spawn ordering in the real server, cross-project registry keying, a second interrupt after persistence changes to interrupted, rerun spawn failure atomicity, endpoint-to-orchestrator abort propagation.

The rerun tests also assert the current pre-inserted running row behavior, which is the opposite of the claimed atomicity.

Fix: Add integration-style tests around the real server spawn/orchestrator seam and failure injection tests where startCritiqueRun throws/rejects.

lefarcen (changes requested):
Hey @Nagendhra-web, I re-reviewed the current head () after your rebase. The **2 P1 findings are still blocking** — the core atomicity/idempotency claims remain unaddressed. CI also failed.

## Still-blocking P1 findings

**1. Rerun path is still broken and not atomic** (`apps/daemon/src/critique/rerun-handler.ts:115-125`)

Line 115 pre-inserts a `critique_runs` row with `status: 'running'` before calling `startCritiqueRun()` at line 125. If `runOrchestrator()` (called inside `startCritiqueRun`) tries to insert the same `runId` again, you get a primary-key failure on the fire-and-forget promise. The caller already got 202 while a `running` orphan row is left behind.

The comment at line 114 ("so the row is always visible") shows the intent, but the implementation doesn't match the PR's "new artifact row committed only after spawn handle resolves" claim — you're committing *before* spawn.

**Fix:** Either:
- Make `runOrchestrator` support a pre-created row explicitly (pass a flag `{ row_already_inserted: true }` so it skips the insert), OR
- Don't pre-insert the row; let `runOrchestrator` insert it, and only commit after spawn succeeds.

**2. Interrupt endpoint is still not idempotent** (`apps/daemon/src/critique/interrupt-handler.ts:54-63`)

Line 54 checks `liveStatus !== 'running'` and returns 409 CONFLICT (line 55-63) when the run is already in a terminal status (`interrupted` / `shipped` / `below_threshold` / `failed`).

This means if the first interrupt completes quickly and flushes the DB to `interrupted`, the second interrupt sees `status='interrupted'` and returns 409, **contradicting the PR's "calling twice returns 202" claim**.

**Fix:** Make the endpoint recognize `interrupted` as a valid "already handled" state that still returns 202 (idempotent success), not 409 (conflict). Example:
```typescript
if (liveStatus === 'interrupted') {
  res.status(HTTP_ACCEPTED).json({ runId, accepted: true, prevStatus: 'interrupted', note: 'already interrupted' });
  return;
}
if (liveStatus !== 'running') {
  res.status(HTTP_CONFLICT).json({ ... });
  return;
}
```

## CI failure

The workflow "Validate workspace" failed. Check the logs — might be typecheck / lint / test failures from these changes.

---

These two P1 issues are the same root concerns from my first review. The other 4 P2 findings (register-before-spawn race, registry keying, process-tree leak, SQL LIKE escaping) and 1 P3 (test coverage) are still relevant, but P1 blocks merge regardless. Let's get the atomicity and idempotency guarantees right first.

lefarcen (changes requested):
Hey @Nagendhra-web, I re-reviewed the current head (d2eb5fe8) after your rebase. The 2 P1 findings are still blocking — the core atomicity/idempotency claims remain unaddressed. CI also failed.

## Still-blocking P1 findings

**1. Rerun path is still broken and not atomic** (apps/daemon/src/critique/rerun-handler.ts:115-125)

Line 115 pre-inserts a critique_runs row with status: 'running' before calling startCritiqueRun() at line 125. If runOrchestrator() (called inside startCritiqueRun) tries to insert the same runId again, you get a primary-key failure on the fire-and-forget promise. The caller already got 202 while a running orphan row is left behind.

The comment at line 114 ("so the row is always visible") shows the intent, but the implementation doesn't match the PR's "new artifact row committed only after spawn handle resolves" claim — you're committing *before* spawn.

**Fix:** Either:
- Make runOrchestrator support a pre-created row explicitly (pass a flag { row_already_inserted: true } so it skips the insert), OR
- Don't pre-insert the row; let runOrchestrator insert it, and only commit after spawn succeeds.

**2. Interrupt endpoint is still not idempotent** (apps/daemon/src/critique/interrupt-handler.ts:54-63)

Line 54 checks liveStatus !== 'running' and returns 409 CONFLICT (line 55-63) when the run is already in a terminal status (interrupted / shipped / below_threshold / failed).

This means if the first interrupt completes quickly and flushes the DB to interrupted, the second interrupt sees status='interrupted' and returns 409, **contradicting the PR's "calling twice returns 202" claim**.

**Fix:** Make the endpoint recognize interrupted as a valid "already handled" state that still returns 202 (idempotent success), not 409 (conflict). Example code:

if (liveStatus === 'interrupted') { res.status(HTTP_ACCEPTED).json({ runId, accepted: true, prevStatus: 'interrupted', note: 'already interrupted' }); return; } if (liveStatus !== 'running') { res.status(HTTP_CONFLICT).json({ ... }); return; }

## CI failure

The workflow "Validate workspace" failed. Check the logs — might be typecheck / lint / test failures from these changes.

---

These two P1 issues are the same root concerns from my first review. The other 4 P2 findings (register-before-spawn race, registry keying, process-tree leak, SQL LIKE escaping) and 1 P3 (test coverage) are still relevant, but P1 blocks merge regardless. Let's get the atomicity and idempotency guarantees right first.

lefarcen (approved):
✅ All P1 blockers resolved. Nice work addressing the feedback!

## What changed (current head 4b358f90)

**Scope reduction:** The rerun handler has been removed entirely from this PR. This is now **Phase 6.1 (interrupt only)**. The prior P1 finding about rerun atomicity (pre-insert collision) is moot.

**P1 #2 (interrupt idempotency) — FIXED:** The handler now recognizes `liveStatus === 'interrupted'` at line 62 and returns 202 with `prevStatus: 'interrupted'` (idempotent retry path). A client that lost the first response and retries now gets the same 202 success, not 409.

**P2 #4 (registry composite-key) — FIXED:** All registry methods (`get`, `interrupt`, `unregister`) now use `compositeKey(projectId, runId)`. Cross-project leak guards are enforced at the registry layer.

**P2 #3 (register-before-spawn) — FIXED:** The new server.ts wiring (line 4297-4302) calls `critiqueRunRegistry.register(...)` BEFORE `await runOrchestrator(...)` (line 4324). The comment at line 4293-4296 explicitly documents this ordering.

**Test coverage is strong:** Idempotent interrupt 202 path (interrupt-endpoint.test.ts lines 221-246) and cross-project registry keying (run-registry.test.ts lines 46-66, 97-109) are both tested.

## Remaining notes

The prior P2 #5 (process-tree teardown) and P2 #6 (artifact lookup escaping) are still deferred, but those are outside the scope of this scoped-down PR. The rerun handler (which had the artifact-lookup concern) is now removed. Process-tree teardown is intentionally marked as a follow-up.

Ready to merge from my side.

lefarcen (dismissed):
All prior P1/P2 blockers are now addressed. The scope reduction to Phase 6.1 (interrupt only) was the right call — the rerun atomicity concern is moot. The three remaining fixes are clean:

**P1 #2 fixed**: Interrupt endpoint now returns 202 (not 409) for already-interrupted runs (apps/daemon/src/critique/interrupt-handler.ts:62-73), making retries idempotent as claimed.

**P2 #4 fixed**: Registry is now keyed by `(projectId, runId)` — all methods (`get`, `interrupt`, `unregister`) use `compositeKey(projectId, runId)`. Cross-project tests verify the isolation.

**P2 #3 correctly disputed**: The register-before-spawn ordering is correct. `critiqueRunRegistry.register` happens at server.ts:4299-4305, then `await runOrchestrator(...)` starts at 4324. The only window where the registry has an entry but the DB doesn't is between register and the orchestrator's first insert, which 404s correctly (the run isn't publicly visible yet). No exploitable race.

The Indonesian locale additions are a valid unblock (web typecheck enforces parity). Tests look comprehensive (17 registry + 16 endpoint cases). Approving.

Note: This clears my prior CHANGES_REQUESTED on the old commit — the current head (4b358f90) is clean. Deferring final merge decision to maintainer @lefarcen per project workflow.

apps/daemon/src/critique/interrupt-handler.ts
```diff
@@ -0,0 +1,100 @@
...
+            currentStatus: row.status,
+          },
+        });
+      return;
+    }
+
+    // Project-keyed registry call: a request to interrupt project p1's runId
+    // cannot match a registry handle from project p2 even if a runId
+    // collision somehow occurred.
+    registry.interrupt(projectId, runId, 'user_requested');
```
mrcfps: Non-blocking: please handle the `false` return from `registry.interrupt(...)` instead of always returning `202 accepted`. The changed line calls the registry, but `RunRegistry.interrupt()` explicitly returns `false` when the `(projectId, runId)` pair has no live `AbortController`. A `running` DB row can exist without a registry handle after a daemon restart that happens before `reconcileStaleRuns` considers the row old enough; in that case this endpoint tells the client the interrupt was accepted even though no child is signaled, no `critique.interrupted` event is emitted, and the row remains `running`. Consider treating `false` as a stale/missing live run (for example `404`/`409`) or marking the row `interrupted` as the recovery path, and add a regression test for a `running` row with no registered handle.

mrcfps (commented):
@Nagendhra-web Thanks for the careful Phase 6.1 cleanup here — the project-keyed registry and scoped interrupt endpoint are much easier to reason about now. I found one merge-safe edge case around stale running rows where the endpoint can report acceptance even though no live handle was actually signaled.


mrcfps (approved):
@Nagendhra-web I reviewed the current Phase 6.1 interrupt-only diff at the latest head, including the project-keyed run registry, interrupt handler recovery/idempotency paths, persistence helper, server wiring, and the new daemon test coverage. The prior stale-handle edge case is now covered by the recovery write and regression test, and I did not find any remaining actionable correctness, safety, or maintainability issues in the changed ranges. Thanks for the careful cleanup and for keeping this scoped and easy to review. 🙌

Issue #821, opened 2026-05-07
[BUG] BUG Open Design Cannot be Closed (When installing new update)


kutzki:
Here is an image. I have it closed, but if I also have to close my Codex/Claude/Antigravity, etc. then there's a serious issue: 

<img width="1228" height="765" alt="Image" src="https://github.com/user-attachments/assets/d619cadd-dac9-4e69-a11b-d8b74671c5d0" />

lefarcen:
Hi @kutzki! 👋

Thanks for reporting this. The quit-blocking behavior you're seeing (where closing the window doesn't actually quit the app when a dialog is showing) is related to a known macOS close-handler issue.

## What's happening

PR #270 intentionally changed the macOS red close button to hide the window instead of destroying the app (for proper Dock behavior), but this same close handler can interfere with **real quit flows** — including when modal dialogs like update prompts are showing. The result: clicking close or using Cmd-Q leaves the app running in the background, requiring a Force Quit.

## The fix

**PR #422** addresses this by adding a proper `before-quit` handler that:
- Routes through a unified shutdown path
- Cleanly tears down all windows and sidecars  
- Only allows the window close to proceed after cleanup completes

This preserves the hide-on-red-X behavior while making real quit (Cmd-Q / Quit menu / osascript quit) work correctly.

## Next steps

The fix is currently in review. Once **PR #422** merges and ships in the next release, the behavior should be:
- Update dialog showing → Cmd-Q or Quit menu → app cleanly shuts down (no Force Quit needed)
- Red close button on main window → hides window but keeps app in Dock (existing behavior, intentional)

If you're still seeing this issue after PR #422 lands, or if you're on Windows/Linux (where the root cause might be different), please reopen and let me know!

Related: #640

kutzki:
I am currently on Windows

lefarcen:
Thanks for clarifying that this is on Windows. My earlier macOS-specific explanation was too narrow.

On Windows this is likely a different quit path: the update prompt is a native/updater modal, and updater-driven quit/install flows do not behave the same as simply closing the last Electron window. In particular, relying on `window-all-closed` for cleanup can miss app-level quit/update flows, which can leave the desktop process or sidecar processes running.

[...]

PR #422, opened 2026-05-04
fix(desktop): cleanly quit macOS packaged app

fuyizheng3120:
## Problem

PR #270 intentionally makes the macOS red close button hide the window instead of destroying the app, which is the right Dock re-open behavior.

However, the same close handler can also intercept the real app quit path. In a packaged macOS app, using Quit / Cmd-Q can leave the Electron main process plus daemon/web sidecars alive, so the Dock still shows Open Design as running.

## Fix

- Handle Electron `before-quit` in desktop main.
- On the first real app quit, prevent the default quit and route through the existing unified `shutdown()` path.
- Once `shutdown()` calls `desktop.close()`, the runtime `stopped` flag allows the BrowserWindow close to proceed, and the existing `app.quit()` completes normally.

This keeps the PR #270 red-X behavior intact while making real Quit / Cmd-Q clean up the app.

## Test

- `pnpm install --frozen-lockfile`
- `pnpm --filter @open-design/desktop typecheck`

Local packaged verification in our macOS build:

- Build and install packaged app.
- Launch Open Design.
- Run `osascript -e 'tell application "Open Design" to quit'`.
- Verify `pgrep -fl 'Open Design|--od-stamp-namespace=release-stable'` returns no Open Design main/helper/sidecar processes.

PR #823, opened 2026-05-07
fix(pack/win): close detection gaps that let Open Design.exe stay locked at install time (#821)

Nagendhra-web:
Closes part of #821. Addresses the two installer-side gaps @lefarcen confirmed in [the issue thread](https://github.com/nexu-io/open-design/issues/821#issuecomment-4397357213) after their own code review.

## Background

The custom NSIS pre-install flow in [tools/pack/src/win/custom-installer.ts](https://github.com/nexu-io/open-design/blob/main/tools/pack/src/win/custom-installer.ts) detects and closes running OD processes before extraction. But the dialog kutzki sees ("Open Design cannot be closed. Please close it manually and click Retry to continue.") is **not** the custom `RunningInstancesCloseFailed` text. It's NSIS's native file-in-use Retry/Cancel prompt fired during `MUI_PAGE_INSTFILES`. That means detection and close ran cleanly but extraction still hit a locked `$INSTDIR\Open Design.exe`.

## Gap 1: WMI null `ExecutablePath`

`DetectRunningInstances` and `CloseRunningInstances` matched only on `Win32_Process.ExecutablePath` under the install root. WMI returns null `ExecutablePath` for processes the caller cannot fully introspect (insufficient access tokens, protected-process states, processes mid-spawn). A child process spawned right as the installer launched can hit this and slip past the filter.

**Fix:** added a fallback that, for null-`ExecutablePath` rows, matches against `CommandLine` containing the install-root prefix. The install path is OD-specific enough to avoid false positives without resorting to a broad `Name` match against arbitrary apps.

## Gap 2: race between `Stop-Process -Force` and NSIS extraction

`CloseRunningInstances` previously called `Stop-Process -Force` and returned. On Windows the kernel finalizes file-handle release asynchronously after process exit, so an overwrite right after the kill can race the handle GC. NSIS hits the file-in-use prompt even though the kill technically succeeded.

**Fix:** added `WaitForExit(5000)` per PID after the force-stop loop. Each killed process gets up to 5s to actually exit before the function returns and `RunningInstancesPageLeave` proceeds. By the time NSIS reaches extraction the OS handles are released.

## What's not in this PR

The third part of @lefarcen's proposed fix is the cross-platform `before-quit` cleanup in the Electron app so that real app-quit (and electron-updater's `quitAndInstall`) reliably tears down sidecars before the installer ever runs. That belongs with [#422](https://github.com/nexu-io/open-design/pull/422)'s direction and is intentionally out of scope here. This PR addresses only the installer-side gaps that fire when, for any reason, an OD process is still alive when the installer launches.

## Honest caveat

I do not have a Windows packaged-build environment to run `pnpm tools-pack win build --to nsis` and reproduce the locked-file dialog end-to-end. The PowerShell edits are textual and follow the patterns already in the template; the file does not have any workspace test surface. A verifying install pass on a real Windows host (previous OD installed and running, then run new installer over it) is recommended before merge. Happy to iterate on whatever the maintainer's local repro reveals.

`pnpm guard` clean. The `tools-pack typecheck` fails on a pre-existing missing `@electron/rebuild` devDep in `src/win/app.ts` on current main (reproducible by checking out main directly without my edit), unrelated to this change.

### DIFF
--- a/tools/pack/src/win/custom-installer.ts
+++ b/tools/pack/src/win/custom-installer.ts
@@ -258,7 +258,13 @@ FunctionEnd
 Function DetectRunningInstances
   Push $0
   Push $1
-  nsExec::ExecToStack 'powershell.exe -NoLogo -NoProfile -NonInteractive -ExecutionPolicy Bypass -Command "& { param($$install, $$registered); $$roots = @($$install, $$registered) | Where-Object { $$_ } | ForEach-Object { $$root = $$_.TrimEnd([char]92).ToLowerInvariant(); [pscustomobject]@{ Exact = $$root; Prefix = ($$root + [char]92) } } | Select-Object -Unique Exact, Prefix; $$matches = Get-CimInstance Win32_Process | Where-Object { $$matched = $$false; $$exe = $$_.ExecutablePath; if ($$null -ne $$exe) { $$exe = $$exe.ToLowerInvariant(); foreach ($$root in $$roots) { if ($$root.Exact -and (($$exe -eq $$root.Exact) -or $$exe.StartsWith($$root.Prefix))) { $$matched = $$true; break } } }; $$matched }; if ($$matches) { $$matches | ForEach-Object { [string]$$_.ProcessId + [char]32 + $$_.Name } } }" "$INSTDIR" "$ExistingInstallLocation"'
+  ; Match by ExecutablePath under the install root when available. WMI returns
+  ; null ExecutablePath for processes the caller cannot fully introspect
+  ; (insufficient access, processes mid-spawn). For those rows, fall back to
+  ; CommandLine containing the install-root prefix, which is OD-specific
+  ; enough to avoid false positives without resorting to global Name matching.
+  ; Issue #821.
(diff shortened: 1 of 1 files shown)


lefarcen (commented):
Hey @Nagendhra-web! 👋 Thanks for diving into this Windows installer gap — the approach looks solid for the two failure modes you identified in #821.

I've reviewed the PowerShell changes and have a few suggestions to tighten the edge-case handling. None block merge, but they'd make the fix more robust:

**Lens A — Code correctness:**

1. **CommandLine fallback scope** (P2): The `CommandLine.Contains($root.Prefix)` match can catch unrelated processes that happen to reference the install directory in their arguments (e.g. a text editor opening a file under `$INSTDIR`). Consider narrowing the fallback to match only when the command line *starts* with the install-root prefix (quoted or unquoted), or require the known basename `Open Design.exe` within that prefix before force-killing.

2. **WaitForExit timeout handling** (P2): `WaitForExit(5000)` returns a boolean, but the result is ignored. If a process doesn't exit within 5s, the installer proceeds anyway, relying on a second WMI pass that has the same gaps and can race file-lock release. Collect PIDs where `WaitForExit` returns false, and either exit non-zero to trigger the existing close-failed dialog, or report them back to NSIS so the installer blocks explicitly rather than hoping the race resolves.

3. **Both-null case** (P3 / informational): If both `ExecutablePath` and `CommandLine` are null (Windows commonly withholds both for protected/higher-integrity processes from unelevated callers), the process is silently skipped. This is fine for the "insufficient access" case you named, but it means the fix doesn't fully cover that scenario — it only catches processes where `CommandLine` is visible. Consider either:
   - Adjusting the PR description to say "partial coverage for insufficient-access rows" (honest caveat), or
   - Adding a separate Restart Manager / file-lock probe for `$INSTDIR\Open Design.exe` to detect lock ownership when WMI fields are unavailable.

The core changes (CommandLine fallback + WaitForExit race fix) are architecturally correct for the two gaps you targeted. These suggestions just polish the edge cases. Let me know if you want to iterate on any of them!

mrcfps (approved):
@Nagendhra-web I reviewed the Windows installer template changes in `tools/pack/src/win/custom-installer.ts`, including the WMI `ExecutablePath`/`CommandLine` matching path and the new post-kill `WaitForExit(5000)` delay before NSIS extraction. I also checked the current PR metadata, diff, and prior review context for this head SHA, and I did not find any additional actionable issues in the changed ranges. Thanks for tackling this installer lockup path and documenting the Windows validation caveat clearly. 🚀


lefarcen (approved):
Approved.

PR #827, opened 2026-05-07
fix(web): scope settings save validation to the active sidebar section (#739)

Nagendhra-web:
Closes #739.

## Problem

The footer Save button's enabled state was computed purely from execution-mode completeness (BYOK requires `apiKey` + `model` + valid `baseUrl`; Local CLI requires a selected available agent). That check ran regardless of which sidebar section the user was on, so a draft mode toggle on the execution section that left required fields empty would lock the Save button across every other section.

After clicking BYOK without filling fields and navigating to language or appearance, the user could not save unrelated changes in those sections even though they had nothing to do with execution mode. Exactly the symptom @shangxinyu1 described.

## Fix

Extracted the gating into a pure helper `shouldEnableSettingsSave(cfg, activeSection, agents, isBaseUrlValid)` in [apps/web/src/components/SettingsDialog.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SettingsDialog.tsx).

- On any sidebar section other than `'execution'` (language, appearance, composio, media, integrations, notifications, pet, library, about): returns `true`. The mode-completeness check is meaningless on those sections.
- On `'execution'`: keeps the existing mode-completeness check unchanged. BYOK-incomplete still disables Save where the user is actually editing the mode fields.

The component's `canSave` is now a one-liner that delegates to the helper.

## Tests

`shouldEnableSettingsSave` has 4 new test groups in [apps/web/tests/components/SettingsDialog.test.ts](https://github.com/nexu-io/open-design/blob/main/apps/web/tests/components/SettingsDialog.test.ts):

- The fix itself: incomplete BYOK on execution does NOT block save when the user navigates to any other section.
- `'execution' + daemon`: returns true only when an available agent is selected.
- `'execution' + api`: returns true only when apiKey, model, and baseUrl are all valid.
- Regression guard: incomplete BYOK on `'execution'` still disables save (the within-section invariant the fix must NOT loosen).

Local: web tests 28/28, web typecheck, `pnpm guard` all clean.

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -452,6 +452,78 @@ export function agentRefreshOptionsForConfig(cfg: AppConfig): AgentRefreshOption
   };
 }
 
+/**
+ * Returns whether the modal's footer Save button should be enabled for the
+ * currently active sidebar section.
+ *
+ * The mode-completeness check (BYOK requires apiKey + model + valid baseUrl;
+ * Local CLI requires a selected available agent) is only meaningful on the
+ * execution-mode section, where the user is actively editing those fields.
+ * On every other sidebar section (language, appearance, composio, media,
+ * integrations, notifications, pet, library, about), partial state from a
+ * draft mode toggle (e.g. user clicked BYOK on the execution section without
+ * filling in fields, then navigated to language) must NOT block saving
+ * changes the user is making in those unrelated sections. Issue #739.
+ */
+export function shouldEnableSettingsSave(
+  cfg: AppConfig,
+  activeSection: SettingsSection,
+  agents: ReadonlyArray<{ id: string; available: boolean }>,
+  isBaseUrlValid: boolean,
+): boolean {
+  if (activeSection !== 'execution') return true;
+  if (cfg.mode === 'daemon') {
+    return Boolean(
+      cfg.agentId && agents.find((a) => a.id === cfg.agentId)?.available,
+    );
+  }
+  return Boolean(cfg.apiKey.trim() && cfg.model.trim() && isBaseUrlValid);
+}
+
+/**
+ * Returns the config that should actually be persisted by `onSave`.
+ *
+ * Counterpart to {@link shouldEnableSettingsSave}: when Save is enabled on a
+ * non-execution sidebar section but the user's draft execution config is
+ * incomplete (e.g. they toggled BYOK on the execution section, never filled
+ * in apiKey, then navigated to Language and clicked Save), the raw `cfg`
+ * still carries that broken draft. Persisting it would leave the app in an
+ * unusable execution state after the modal closes. This helper reverts the
+ * execution-related fields to their `initial` values in that case, so saving
+ * an unrelated section change never silently commits an incomplete execution
+ * mode.
+ *
+ * Within the execution section, or when execution is already valid, the
+ * config passes through unchanged. Issue #739.
+ */
+export function sanitizeSettingsSavePayload(
+  cfg: AppConfig,
+  initial: AppConfig,
+  activeSection: SettingsSection,
+  agents: ReadonlyArray<{ id: string; available: boolean }>,
+  isBaseUrlValid: boolean,
+): AppConfig {
+  if (activeSection === 'execution') return cfg;
+  // Reuse the existing execution-section validity gate so the two helpers
+  // share one source of truth for "execution config is complete enough."
+  const executionValid = shouldEnableSettingsSave(cfg, 'execution', agents, isBaseUrlValid);
+  if (executionValid) return cfg;
+  return {
+    ...cfg,
+    mode: initial.mode,
+    apiKey: initial.apiKey,
+    apiProtocol: initial.apiProtocol,
+    apiVersion: initial.apiVersion,
+    apiProtocolConfigs: initial.apiProtocolConfigs,
+    apiProviderBaseUrl: initial.apiProviderBaseUrl,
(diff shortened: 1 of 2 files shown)


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -452,6 +452,34 @@ export function agentRefreshOptionsForConfig(cfg: AppConfig): AgentRefreshOption
...
+ * filling in fields, then navigated to language) must NOT block saving
+ * changes the user is making in those unrelated sections. Issue #739.
+ */
+export function shouldEnableSettingsSave(
+  cfg: AppConfig,
+  activeSection: SettingsSection,
+  agents: ReadonlyArray<{ id: string; available: boolean }>,
+  isBaseUrlValid: boolean,
+): boolean {
+  if (activeSection !== 'execution') return true;
```
lefarcen, line 474: 🔒 **P1 — Save persists invalid execution config**

The helper correctly enables Save on non-execution sections, but the footer still calls `onSave(cfg, ...)` with the **entire draft config** at L1710, not a section-scoped patch.

The exact issue scenario can now persist incomplete execution state:
1. User toggles BYOK (mode='api'), leaves `apiKey`/`model` empty
2. Navigates to Language/Appearance
3. Clicks Save (now enabled per this PR)
4. `App.handleConfigSave` receives `mode: 'api'` + empty provider fields and writes it

User's execution mode is now broken — they can't run queries until they go back to execution section and fill in the fields again.

**Concrete fix**: when `activeSection !== 'execution'` and execution config is incomplete, merge the active section's changes onto `initial` (or last valid execution config) instead of persisting the invalid execution draft. E.g.:
```ts
const sanitizedCfg = activeSection !== 'execution' && !isExecutionModeComplete(cfg)
  ? { ...initial, [activeSection]: cfg[activeSection] }
  : cfg;
onSave(sanitizedCfg, ...);
```
Alternatively, skip execution fields altogether when saving from a non-execution section.

apps/web/tests/components/SettingsDialog.test.ts
```diff
@@ -375,3 +376,75 @@ describe('deriveComposioCredentialState', () => {
...
+  const validDaemonCfg: AppConfig = {
+    ...baseConfig,
+    mode: 'daemon',
+    agentId: 'claude-code',
+  };
+
+  const availableAgent = { id: 'claude-code', available: true };
+  const unavailableAgent = { id: 'claude-code', available: false };
+
+  it('returns true on any non-execution section regardless of mode completeness (the fix for #739)', () => {
```
lefarcen, line 410: **P2 — Missing end-to-end save payload test**

The new tests only verify `shouldEnableSettingsSave(...) === true` in isolation. They don't prove the **save payload** is safe — the main regression risk (incomplete execution config persisting from another section).

**Concrete fix**: add a component-level test that renders `SettingsDialog`, toggles BYOK incomplete, switches to a non-execution section, clicks Save, and asserts `onSave` receives a config that does NOT commit the incomplete execution mode. This would have caught the P1 above.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for tackling #739! The approach is clean — extracting `shouldEnableSettingsSave` and scoping the gate to `activeSection === 'execution'` is the right direction. However, there's a critical behavioral gap the fix introduces. See inline.

P1 must be addressed before merge. The P2 is optional but strongly recommended to catch this class of issue in CI.

apps/web/src/components/SettingsDialog.tsx
mrcfps, line 474: 🔒 **Blocking — prevent saving invalid execution drafts from other sections**

This new non-execution branch enables the footer Save button even when the current draft execution config is incomplete. The footer still calls `onSave(cfg, activeSection !== 'composio')` with the entire draft config (`SettingsDialog.tsx:1710`), so a user can switch to BYOK, leave `apiKey`/`model` blank, navigate to Language/Appearance, and then save that unrelated change while persisting `mode: 'api'` with empty credentials. That leaves the app in a broken execution mode after the modal closes, which is worse than the original lockout. Please either save only the fields owned by the active non-execution section, or restore the last valid/initial execution fields before calling `onSave` whenever `activeSection !== 'execution'` and the execution draft fails the same completeness check.

mrcfps (changes requested):
@Nagendhra-web thanks for jumping on #739 and for adding focused helper coverage. I found one main-path persistence issue: Save is now enabled from unrelated sections, but the click still saves the full draft config, so incomplete execution edits can be committed from another section. Details inline.


lefarcen (approved):
Fixed! 🎉

The `sanitizeSettingsSavePayload` helper + wiring at the `onSave` call site completely addresses the P1 blocker both @mrcfps and I flagged. When the user saves from a non-execution section with an incomplete BYOK draft, the execution-mode fields now revert to `initial` while preserving the non-execution change. The broken `mode='api'` + empty credentials never reach localStorage.

Test coverage is thorough: the revert path, the no-op when execution is valid, the no-op on the execution section itself, every non-execution section, and the unavailable-agent edge case.

Clean fix — approved from my side.

mrcfps (approved):
@Nagendhra-web I rechecked the changed save path in `SettingsDialog`: non-execution saves now sanitize incomplete execution drafts before `handleConfigSave`, and the added helper coverage exercises both the revert path and the valid-execution pass-through path. I could not rerun the package checks in this checkout because `node_modules` are not installed here, but the changed ranges themselves look correct to me. Thanks for the quick follow-up on the earlier blocker.

PR #829, opened 2026-05-07
fix(web): wrap comment-popover action row so the Save/Sending button cannot exceed the popover edge (#779)

Nagendhra-web:
Closes #779.

## Problem

`.comment-popover-actions` in [apps/web/src/index.css](https://github.com/nexu-io/open-design/blob/main/apps/web/src/index.css) lays out four buttons (Remove, Add note, Save comment, Send to chat / Sending...) with `display: flex`, `justify-content: space-between`, no `flex-wrap`, and no per-child shrink ceiling. The popover itself is `width: min(320px, calc(100% - 28px))` with 10px padding on each side, leaving roughly 300px of inner room.

The combined natural width of the labels (`Add note` + `Save comment` + `Send to chat`) exceeds 300px before the gaps are even counted. With nowrap and no shrink, the rightmost button visibly spills past the popover's right edge, exactly the symptom @shangxinyu1 reported. The `Sending...` text in the screenshot is just where the user happened to notice it; the underlying overflow is independent of the specific button label.

## Fix

Two single-property additions:

```css
.comment-popover-actions {
  /* existing */
  flex-wrap: wrap;
}
.comment-popover-actions > * {
  max-width: 100%;
}
```

`flex-wrap: wrap` lets the row break onto a second line when the labels do not fit. `max-width: 100%` on direct children stops a single oversized button from pushing the row out instead of wrapping. Wider screens where the buttons already fit see no layout change.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -5876,8 +5876,17 @@ button.connector-action.is-loading {
   display: flex;
   justify-content: space-between;
   align-items: center;
+  flex-wrap: wrap;
   gap: 8px;
   margin-top: 8px;
+}
+.comment-popover-actions > * {
+  /* Buttons may carry labels longer than their natural fair-share of the
+     320px popover width (Save comment, Send to chat, Sending..., Add note,
+     Remove). Without an explicit shrink ceiling and wrap, the row pushes
+     past the popover's right edge instead of breaking onto a new line.
+     Issue #779. */
+  max-width: 100%;
 }
 .comment-popover-remove {
   color: var(--red);


lefarcen (commented):
Hi @Nagendhra-web! 👋

Thanks for the fix — clean root-cause analysis and the minimal CSS change is exactly right.

**Code review (Lens A)**

No blocking issues. The fix correctly uses `flex-wrap: wrap` to allow the button row to break onto a second line when labels exceed the ~300px inner width, and `max-width: 100%` prevents any single oversized button from forcing overflow. The only usage of `.comment-popover-actions` is in `FileViewer.tsx:1357`, so this change has the intended narrow scope.

`justify-content: space-between` still works per flex line after wrapping (a single item on a wrapped final line will align left, which is standard flex behavior and not a correctness issue).

**Residual notes (non-blocking)**

- A visual check in the browser would be useful but isn't required for this small CSS fix.
- For future extreme localization cases with very long unbroken labels, `overflow-wrap: anywhere` could be considered, but current labels don't need it.

— open-design team

mrcfps (approved):
@Nagendhra-web I reviewed the scoped CSS change in `apps/web/src/index.css` and verified it stays narrowly focused on the comment popover action row: `flex-wrap: wrap` lets the existing action buttons break instead of overflowing, and the direct-child `max-width: 100%` rule prevents a single oversized control from forcing the popover wider. I also checked the only usage in `FileViewer.tsx` and did not find any correctness, safety, or maintainability issues in the changed range. Thanks for the thoughtful root-cause writeup and the minimal fix — nice work! 🙌



lefarcen (approved):
Approved.

PR #830, opened 2026-05-07
fix(web): wrap long note text inside picker/comment popover (#782)

Nagendhra-web:
Closes #782.

## Problem

`.board-note-item` is a flex row inside the comment popover holding a `<span>` with the note text and a `Remove` button. The span had no width hints. When the user added a note containing a very long unbroken string (URL, hash, base64), the span tried to fit it on one line, pushing the row wider than the popover's 320px boundary and distorting the overlay's right edge plus the surrounding picker UI.

## Fix

Three properties on `.board-note-item span`:

```css
flex: 1;            /* take remaining width next to the Remove button */
min-width: 0;       /* lift the default flex-item min-content floor so shrinking works */
overflow-wrap: anywhere; /* break long unbroken strings at any character when needed */
```

`flex: 1` and `min-width: 0` together let the span size to the available width. `overflow-wrap: anywhere` allows breaking at any character when natural word boundaries can't fit the string. No layout change for normal-length notes.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -5865,7 +5865,15 @@ button.connector-action.is-loading {
   font-size: 12px;
 }
 .board-note-item span {
+  /* The note span sits next to a Remove button inside a fixed-width
+     ~300px popover. Without flex: 1 + min-width: 0 + overflow-wrap an
+     unbroken long string (URL, hash, base64) refuses to break and
+     pushes the popover layout sideways past its 320px boundary.
+     Issue #782. */
+  flex: 1;
+  min-width: 0;
   color: var(--text);
+  overflow-wrap: anywhere;
 }
 .comment-popover textarea {
   min-height: 78px;


lefarcen (commented):
Hi @Nagendhra-web! 👋

Clean fix for the layout overflow bug. The `flex: 1` + `min-width: 0` + `overflow-wrap: anywhere` combination is the correct modern CSS pattern for flex-item text wrapping in a constrained container — addresses the common flex min-content floor issue without forcing normal text to break prematurely.

Grep shows `.board-note-item` is narrowly scoped to one component (note text span + Remove button), so no side-effect risk. `overflow-wrap: anywhere` has broad modern-browser support and is preferable to older `word-break: break-all` because it preserves word boundaries when possible.

mrcfps (approved):
@Nagendhra-web I reviewed the focused CSS change for the board note/comment popover. The added `flex: 1`, `min-width: 0`, and `overflow-wrap: anywhere` on `.board-note-item span` match the constrained flex-row layout and should keep long unbroken note text inside the 320px popover without changing normal note layout. Thanks for the targeted fix — this is a clean, maintainable way to address #782. 🙌



lefarcen (approved):
Approved.

PR #832, opened 2026-05-07
feat(daemon): finalize design package endpoint (closes #450)

### LINKED ISSUE #450: feat(daemon): finalize design package — generate DESIGN.md from project source + chat transcript

bankielewicz:
## Summary
- Adds `POST /api/projects/:id/finalize/anthropic`: composes `.transcript.jsonl` (truncated to 384 KiB head+tail when oversized) + active design system + current artifact → Claude synthesis → atomic write of `DESIGN.md` → structured response with byte counts and token usage.
- Mirrors the existing `connectionTest.ts` fetch pattern for Anthropic calls (raw `fetch` + `x-api-key` + `anthropic-version: '2023-06-01'`); no new npm dependencies.
- Reuses the `exportProjectTranscript` primitive from #493 for the transcript half of the input.

## Design decisions (per maintainer guidance on #450)
- **Source files included:** transcript + active design system + current artifact (per @lefarcen answer (a) at https://github.com/nexu-io/open-design/issues/450#issuecomment-4396862609).
- **Response model:** blocking 200 (per @lefarcen answer (b)). Streaming deferred to follow-up.
- **Failure handling:** 1 retry on transient (429/5xx), 502 on terminal failure, redacted upstream text in `details` (per @lefarcen answer (c)).
- **Current artifact resolution:** active artifact tab → newest `.artifact.json` by `manifest.updatedAt` → null. "Active tab" requires a `.artifact.json` sidecar on disk (an active tab pointing at a `.md` falls through). Returns `artifact: null` in payload when no artifact is in scope; the synthesis prompt's "Current artifact" section reads "none".
- **Path shape:** `/api/projects/:id/finalize/anthropic` — project-scoped (matches every sibling `/api/projects/:id/*` route in `server.ts`) + provider-namespaced (leaves room for future `/openai`, `/bedrock`).
- **Body shape:** field names match `ProxyStreamRequest` (`packages/contracts/src/api/proxy.ts:8-19`) — `apiKey`, `baseUrl?`, `model`, `maxTokens?`. `baseUrl` is optional (intentional divergence from the proxy, which requires it) with default `https://api.anthropic.com` so standard Anthropic users don't need to set it; Bedrock / self-hosted users still can.
- **Transcript truncation policy:** real-world signal at authoring was a 3.95 MB local transcript, well past Anthropic's ~200K-token context. The `truncateTranscriptForPrompt` helper hard-caps the prompt body at 384 KiB via head+tail with a single `{"kind":"truncated","reason":"size","omittedBytes":N}` sentinel between them; the on-disk `.transcript.jsonl` from #493 stays lossless.

## Test plan
- 26 integration tests in `apps/daemon/tests/finalize-design.test.ts` covering: truncation policy (small/large), current-artifact resolver (active-tab beats newest, fallthrough on non-artifact tab, null-when-empty), prompt construction (verbatim transcript inclusion, "none" fallbacks for missing design system / missing artifact), Anthropic call retry (no retry on 4xx, retry on 429, exhaust on 5xx, abort propagation), DESIGN.md extraction (concat / malformed shape / no text blocks), full pipeline (atomic write happy path, response shape, no design system path, lockfile contention, second-finalize replacement, tmp+lockfile cleanup on error, default baseUrl), HTTP-layer validation (invalid baseUrl, forbidden private IP, missing apiKey, unsafe :id).
- Full daemon suite: 1078/1078 pass.
- `pnpm --filter @open-design/daemon typecheck`: clean. The new module drops `@ts-nocheck` per #493 review precedent (catches drift inside `transcript-export.ts` itself; `db.ts` retains it per existing convention so DB-helper return types still flow through as `any` at the boundary — same caveat as #493).

## Honest deviations from spec
- Test count is 26 (spec target 21). The retry+extract describe blocks split into more granular unit tests than originally enumerated (4 retry cases, 3 extract cases). All real, none skipped; the +5 reflects more thorough unit coverage of the same surface area.
- One commit beyond the planned 7-commit stack (`fix(daemon): resolve noUncheckedIndexedAccess in truncateTranscriptForPrompt`) recovers from a typecheck miss in the prior commit. Honest stack rather than amend per the project's commit policy.

## Out of scope (deliberate)
- Web button + toast — that's PR 3 (or part of #451).
- SSE streaming of synthesis progress — follow-up if @lefarcen requests it.
- 202+poll background-job pattern — @lefarcen answer (b) explicitly chose blocking.
- Multi-provider support (`/openai`, `/bedrock`, etc.) — provider-namespaced URL leaves room but not in this PR.
- Prompt caching headers — minimal first; optimization follow-up.
- `.history/` versioning of past `DESIGN.md` files — second finalize overwrites atomically.
- `packages/contracts` exposure of `FinalizeAnthropicRequest`/`FinalizeAnthropicResponse` — types stay inlined per scope rules; defer to a follow-up if PR 3 needs the shared types.

Closes #450.

### DIFF
--- a/apps/daemon/src/finalize-design.ts
+++ b/apps/daemon/src/finalize-design.ts
@@ -198,6 +198,83 @@ export async function finalizeDesignPackage(
   throw new Error('finalizeDesignPackage not yet implemented (phase D scaffold)');
 }
 
+const SYSTEM_PROMPT = `You are a senior product designer synthesizing a finalized design package
+from a multi-turn design session. Your output is a single Markdown document
+named DESIGN.md that captures the durable design intent of the work so a
+fresh contributor (human or LLM) can reconstruct context without replaying
+the full chat.
+
+Output structure (Markdown headings exactly as below):
+# DESIGN.md
+## Summary
+## Brand & Voice
+## Information Architecture
+## Components & Patterns
+## Visual System
+## Open Questions
+## Provenance
+
+The Provenance section MUST list:
+- Project ID
+- Design system (or "none" if not selected)
+- Current artifact (file name, or "none" if not in scope)
+- Transcript message count
+- Generated UTC timestamp
+
+Output the Markdown body only. No preamble, no chat-style framing, no
+"Here's your DESIGN.md" prefix. Do not invent facts not supported by the
+inputs; if an input is missing or empty, the corresponding section should
+say so explicitly rather than fabricating content.`;
+
+export interface SynthesisPromptInput {
+  projectId: string;
+  transcriptJsonl: string;
+  transcriptMessageCount: number;
+  designSystemId: string | null;
+  designSystemBody: string | null;
+  artifact: ResolvedArtifact | null;
+  now: Date;
+}
+
+export interface SynthesisPromptOutput {
+  systemPrompt: string;
+  userPrompt: string;
+}
+
+/**
+ * Build the system + user prompts for the Anthropic Messages API call.
+ * Inputs are verbatim except for the transcript (which the caller has
+ * already passed through `truncateTranscriptForPrompt` — this function
+ * does not re-truncate). Missing inputs (no design system selected, no
+ * artifact in scope) produce explicit "none"/parenthetical placeholders
+ * so Claude does not hallucinate content for absent sections.
+ */
+export function buildSynthesisPrompt(input: SynthesisPromptInput): SynthesisPromptOutput {
+  const designSystemHeader = input.designSystemId ?? 'none';
+  const designSystemBody =
+    input.designSystemBody && input.designSystemBody.trim().length > 0
+      ? input.designSystemBody
+      : '(no design system selected for this project)';
+
+  const artifactHeader = input.artifact ? input.artifact.name : 'none';
+  const artifactBody = input.artifact
+    ? input.artifact.body
+    : '(no artifact in scope for this finalize)';
+
+  const userPrompt =
+    `The following inputs describe the design session for project ${input.projectId}.\n\n` +
+    `## Transcript (JSONL)\n${input.transcriptJsonl}\n\n` +
+    `## Active design system: ${designSystemHeader}\n${designSystemBody}\n\n` +
+    `## Current artifact: ${artifactHeader}\n${artifactBody}\n\n` +
+    `## Generation context\n` +
+    `- Generated at: ${input.now.toISOString()}\n` +
+    `- Project ID: ${input.projectId}\n` +
+    `- Transcript message count: ${input.transcriptMessageCount}\n\n` +
+    `Synthesize DESIGN.md per the system instructions.`;
+
+  return { systemPrompt: SYSTEM_PROMPT, userPrompt };
+}
+
 /**
  * Truncate a JSONL transcript body so it fits inside Claude's context
  * window when fed into a synthesis prompt. The on-disk transcript stays
--- a/apps/daemon/src/finalize-design.ts
+++ b/apps/daemon/src/finalize-design.ts
@@ -23,13 +23,18 @@
 // export — verified during PR #493). Schema-mismatch tests in the test
 // file would catch any drift between this restated union and the contract.
 
-import { existsSync } from 'node:fs';
+import { randomBytes } from 'node:crypto';
+import fs from 'node:fs';
 import * as path from 'node:path';
+import Database from 'better-sqlite3';
+import { getProject } from './db.js';
+import { readDesignSystem } from './design-systems.js';
 import {
   listFiles,
   projectDir,
   readProjectFile,
 } from './projects.js';
+import { exportProjectTranscript } from './transcript-export.js';
 
 const DEFAULT_BASE_URL = 'https://api.anthropic.com';
 const DEFAULT_MAX_TOKENS = 16000;
@@ -93,13 +98,7 @@ export class FinalizeUpstreamError extends Error {
   }
 }
 
-interface Db {
-  prepare(sql: string): {
-    all(...args: unknown[]): unknown[];
-    get(...args: unknown[]): unknown;
-    run(...args: unknown[]): unknown;
-  };
-}
+type Db = Database.Database;
 
 interface ResolvedArtifact {
   name: string;
@@ -142,7 +141,7 @@ export async function resolveCurrentArtifact(
 
   if (activeTabName) {
     const sidecarPath = path.join(dir, `${activeTabName}.artifact.json`);
-    if (existsSync(sidecarPath)) {
+    if (fs.existsSync(sidecarPath)) {
       const file = await readProjectFile(projectsRoot, projectId, activeTabName);
       return {
         name: file.name,
@@ -157,7 +156,7 @@ export async function resolveCurrentArtifact(
   const candidates = files
     .filter((f: { name: string; artifactManifest?: { updatedAt?: string } | null }) => {
       // Require a real sidecar on disk; an inferred manifest does not count.
-      return existsSync(path.join(dir, `${f.name}.artifact.json`));
+      return fs.existsSync(path.join(dir, `${f.name}.artifact.json`));
     })
     .map((f: { name: string; artifactManifest?: { updatedAt?: string } | null }) => ({
       name: f.name,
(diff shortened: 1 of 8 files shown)


lefarcen:
Hi @bankielewicz! 🎉

Thanks for the contribution — this implements the design package endpoint spec from #450 with comprehensive test coverage and proper error handling.

— open-design team

apps/daemon/src/finalize-design.ts
```diff
@@ -0,0 +1,602 @@
...
+  options: FinalizeOptions,
+): Promise<FinalizeAnthropicResponse> {
+  const project = getProject(db, projectId);
+  if (!project) {
+    // Defensive — the route handler validates this and returns 404 before
+    // reaching here. Kept for direct (non-HTTP) callers, e.g. CLI scripts.
+    throw new Error(`project not found: ${projectId}`);
+  }
+
+  const dir = projectDir(projectsRoot, projectId);
```
lefarcen: **P1 — Path traversal via encoded project-id**

The `unsafe :id` test only covers `bad!id`, but encoded dots like `%2e%2e` can bypass the guard and make `projectDir(PROJECTS_DIR, projectId)` resolve outside `.od/projects/`. An attacker could create a project row with encoded id, then finalize writes `DESIGN.md` under `.od/` or `projects/` rather than the intended subfolder.

**Fix:** Add a centralized project-id validator that rejects `.`, `..`, and any path-significant segment (decoded). Use it in project creation, `projectDir`, and this route. Add explicit tests for `%2e` and `%2e%2e`.

apps/daemon/src/finalize-design.ts
```diff
@@ -0,0 +1,602 @@
...
+ *
+ * Sidecar presence is checked via `existsSync` on the on-disk path so
+ * the resolver does not depend on `inferLegacyManifest`'s heuristic.
+ */
+export async function resolveCurrentArtifact(
+  db: Db,
+  projectsRoot: string,
+  projectId: string,
+): Promise<ResolvedArtifact | null> {
+  const dir = projectDir(projectsRoot, projectId);
```
lefarcen: **P1 — Imported-folder projects broken**

This ignores `project.metadata.baseDir`, so imported-folder projects (created via `/api/import/folder`) get a successful 200 but `DESIGN.md` is written to the hidden `.od/projects/<id>` data dir instead of the user's actual folder. The current artifact resolution also misses the real files.

**Fix:** Thread project metadata into `resolveCurrentArtifact` and `finalizeDesignPackage`, use `resolveProjectDir`/`ensureProject` semantics (same pattern as other project-file writes in the daemon). Add an integration test for an imported-folder project.

apps/daemon/src/finalize-design.ts
```diff
@@ -0,0 +1,602 @@
...
+    max_tokens: params.maxTokens,
+    system: params.systemPrompt,
+    messages: [{ role: 'user', content: params.userPrompt }],
+    stream: false,
+  });
+
+  for (let attempt = 0; attempt <= 1; attempt += 1) {
+    const init: RequestInit = { method: 'POST', headers, body };
+    if (params.signal) init.signal = params.signal;
+    const response = await fetchImpl(url, init);
```
lefarcen, line 479: **P1 — Network failures return 500 INTERNAL instead of 502 UPSTREAM_FAILED**

DNS errors, connection resets, and `response.json()` parse failures fall through as generic 500s rather than the documented `502 UPSTREAM_FAILED` with redacted details. Only `fetch()` returning a non-ok HTTP status is wrapped in `FinalizeUpstreamError`.

**Fix:** Wrap fetch rejections and JSON parse errors as `FinalizeUpstreamError(502, ...)`. Add tests for fetch rejection and 200-with-non-JSON.

apps/daemon/src/finalize-design.ts
```diff
@@ -0,0 +1,602 @@
...
+import { exportProjectTranscript } from './transcript-export.js';
+
+const DEFAULT_BASE_URL = 'https://api.anthropic.com';
+const DEFAULT_MAX_TOKENS = 16000;
+const INPUT_BODY_CAP_BYTES = 384 * 1024;
+const LOCK_FILENAME = '.finalize.lock';
+const OUTPUT_FILENAME = 'DESIGN.md';
+const DEFAULT_TIMEOUT_MS = 120_000;
+
+export interface FinalizeAnthropicRequest {
```
lefarcen: **P2 — Error codes + DTOs not in packages/contracts**

`FINALIZE_IN_PROGRESS`, `AUTH_FAILED`, `UPSTREAM_FAILED`, `TIMEOUT`, `INTERNAL` are daemon-local and not in the shared `ApiErrorCode` union. The finalize request/response types are also daemon-local, causing drift risk.

**Fix:** Add finalize DTOs to `packages/contracts` and either add the new error codes or reuse existing ones (`CONFLICT`, `UNAUTHORIZED`, `UPSTREAM_UNAVAILABLE`, `INTERNAL_ERROR`).

apps/daemon/src/finalize-design.ts
```diff
@@ -0,0 +1,602 @@
...
+): Promise<ResolvedArtifact | null> {
+  const dir = projectDir(projectsRoot, projectId);
+
+  const activeTabRow = db
+    .prepare(`SELECT name FROM tabs WHERE project_id = ? AND is_active = 1 LIMIT 1`)
+    .get(projectId) as { name?: unknown } | undefined;
+  const activeTabName =
+    activeTabRow && typeof activeTabRow.name === 'string' ? activeTabRow.name : null;
+
+  if (activeTabName) {
```
lefarcen, line 142: **P3 — Active-tab name not validated before sidecar probe**

The raw tab name goes into `path.join(dir, \`${activeTabName}.artifact.json\`)` before `readProjectFile` validates it. A malformed tab (e.g., `../../../etc/passwd.artifact.json`) could make finalize probe outside the project dir and then fail during `readProjectFile`, instead of safely falling through to newest artifact/null.

**Fix:** Validate/sanitize the tab name before `path.join`. Add a test for an unsafe active-tab name.

lefarcen (changes requested):
Thanks for the comprehensive implementation, @bankielewicz! The endpoint structure, test coverage, and atomic-write pattern are solid. I found 3 critical issues (path traversal + imported-folder incompatibility + upstream error mapping) that need fixing before merge.

## Summary

P1 × 3 (must-fix):
1. **Path traversal via encoded project-id** — `%2e%2e` bypasses the `!id` guard
2. **Imported-folder projects broken** — writes go to `.od/projects/<id>` instead of `metadata.baseDir`
3. **Network failures mapped as 500 INTERNAL** — should be 502 UPSTREAM_FAILED

P2 × 2 (should-fix):
4. Error codes + DTOs not in `packages/contracts` — drift risk
5. Active-tab name not validated before sidecar probe — could probe outside project

See inline comments for details. Happy to answer questions!

mrcfps (changes requested):
@bankielewicz thanks for the thorough endpoint implementation and the broad daemon test coverage here — the atomic-write path and prompt assembly are easy to follow. I completed a full pass over the changed ranges and am requesting changes because the latest head still has main-path/security blockers already called out in unresolved resolvable threads, so I am not opening duplicate inline threads for the same locations.

Blocking items still present on `d692bab2e4db3766920eb866cc9b0e7206a76815`:

1. `apps/daemon/src/finalize-design.ts:196` still resolves the finalize output directory with `projectDir(projectsRoot, projectId)`, and `resolveCurrentArtifact` does the same at `apps/daemon/src/finalize-design.ts:134`. For folder-import projects, the daemon's documented behavior is that project files live under `project.metadata.baseDir` (`apps/daemon/src/server.ts:1901-1904`), but this path ignores metadata, so finalize writes `DESIGN.md` and reads artifacts from `.od/projects/<id>` instead of the user's imported folder. Please thread the project metadata through this pipeline and use the same `resolveProjectDir` / `metadata`-aware file helpers as the existing project-file APIs, with an imported-folder integration test.
2. `apps/daemon/src/server.ts:3190` mirrors the existing permissive id regex, which allows decoded `.` / `..` segments. Because this route then calls `finalizeDesignPackage` and `projectDir`, an encoded project id such as `%2e%2e` is not rejected by this guard after Express decoding. Please centralize project-id validation so `.` / `..` and other path-significant segments are rejected before DB/filesystem work, and add explicit HTTP tests for encoded dot segments.
3. `apps/daemon/src/finalize-design.ts:404` and `apps/daemon/src/finalize-design.ts:279` only wrap non-OK HTTP responses as `FinalizeUpstreamError`; fetch rejections and 200-with-malformed/non-JSON bodies fall through to the route's generic 500 handler at `apps/daemon/src/server.ts:3268`. The PR's failure contract says terminal upstream failures should surface as redacted `502 UPSTREAM_FAILED`, so please wrap network failures and response JSON parse failures as `FinalizeUpstreamError(502, ...)` and cover both cases.

The rest of the implementation looks promising once those blockers are addressed. 🙏


lefarcen (dismissed):
All 3 P1 findings from the previous review are fixed in this commit:

1. **Path traversal** — `isSafeId` now rejects pure-dot ids (`.`, `..`) with an explicit `/^\.+$/` guard
2. **Imported-folder projects** — `resolveCurrentArtifact` and `finalizeDesignPackage` now thread `metadata` through and use `resolveProjectDir` instead of bare `projectDir`
3. **Network failures as 500** — fetch rejections and JSON parse errors are now rewrapped as `FinalizeUpstreamError(502)` inside the function

The P2 items are also addressed:
- Error codes/DTOs moved to `packages/contracts/src/api/finalize.ts` per review feedback
- Active-tab name validated via `validateProjectPath` before `path.join`

The implementation now correctly handles the security and imported-folder edge cases. CI failure appears to be a workspace validation issue unrelated to the finalize logic.

Great work addressing all the findings! 🎉

packages/contracts/package.json
```diff
@@ -18,6 +18,9 @@
       "types": "./dist/api/connectionTest.d.ts",
       "default": "./dist/api/connectionTest.mjs"
     },
+    "./api/finalize": {
```
mrcfps, line 21: This new subpath export is currently type-only, which makes the PR fail the shared contracts runtime-shape checks on the current head. `packages/contracts/tests/package-runtime.test.ts:38-47` expects every entry in `package.json#exports` to expose both a generated `.mjs` runtime target and a `.d.ts` target, and CI is already red with `target.default` being `undefined` for `./api/finalize`. On top of that, `packages/contracts/esbuild.config.mjs:6` does not build `src/api/finalize.ts` as a standalone runtime entry, so even adding `default` here would still point at a missing file. Please either remove the `./api/finalize` subpath and import these types from the package root, or add a real runtime export by generating `dist/api/finalize.mjs` and wiring both `default` and `types` here.

mrcfps (changes requested):
@bankielewicz thanks for addressing the earlier finalize blockers and tightening the daemon-side coverage here. I found one remaining blocker on the current head that is still failing the contracts package checks in CI; details are inline.


lefarcen:
My earlier APPROVE verified the daemon-side fixes (path traversal, imported-folder, network errors) — those are solid. @mrcfps found a new blocker in the contracts package that needs fixing before merge. Author, please address the contracts subpath export issue in the latest review thread.

packages/contracts/package.json
bankielewicz: Fixed in 29dc717 — took your path B (real runtime entrypoint).

- Added `./src/api/finalize.ts` to esbuild's `entryPoints` so `dist/api/finalize.mjs` is now generated as a standalone module rather than only inlined into the bundled root.
- Re-added `./api/finalize` to `package.json#exports` with both `default: ./dist/api/finalize.mjs` and `types: ./dist/api/finalize.d.ts`, mirroring `./api/connectionTest`'s shape.
- Kept `FINALIZE_SCHEMA_VERSION = 1` as a real value export so the standalone module has a runtime symbol to emit beyond the type-only interfaces.
- Daemon now imports via `import type { ... } from '@open-design/contracts/api/finalize'` (mirroring how `connectionTest` is consumed).

Sanity checks:
- `node --input-type=module -e "import('@open-design/contracts/api/finalize').then(m=>console.log(JSON.stringify(Object.keys(m))))"` prints `["FINALIZE_SCHEMA_VERSION"]`.
- `pnpm --filter @open-design/contracts test`: 6/6 (both `package-runtime.test.ts` cases pass on the rebuilt exports map).
- `pnpm --filter @open-design/daemon typecheck`: exits 0.
- `pnpm --filter @open-design/daemon test`: 1089/1089.

Honest deviation: I first attempted path A (remove subpath, import from root). That hit a separate `tsc/NodeNext` resolution issue where the daemon couldn't see the re-exported names through `@open-design/contracts` even though `dist/index.d.ts` had `export * from './api/finalize'`. Path B sidesteps it by treating `./api/finalize` as a first-class published module identity (which is also the philosophy your `package-runtime.test.ts` encodes). CI is now green.

lefarcen (approved):
mrcfps's contracts package blocker is now fixed on this head ():
- `packages/contracts/package.json` now wires both `types` and `default` for `./api/finalize`
- `packages/contracts/esbuild.config.mjs` builds `src/api/finalize.ts` as a standalone runtime entry
- CI: **SUCCESS** ✅

The implementation correctly addresses the missing runtime export issue. All prior daemon-side fixes (path traversal, imported-folder, network error wrapping) remain solid.

Great work!

lefarcen (dismissed):
mrcfps's contracts package blocker is now fixed on this head (29dc717):
- packages/contracts/package.json now wires both types and default for ./api/finalize
- packages/contracts/esbuild.config.mjs builds src/api/finalize.ts as a standalone runtime entry
- CI: SUCCESS ✅

The implementation correctly addresses the missing runtime export issue. All prior daemon-side fixes (path traversal, imported-folder, network error wrapping) remain solid.

Great work!

bankielewicz:
Pushed `29dc717` addressing @mrcfps's CI-red feedback on the `./api/finalize` subpath export. Surfacing here as a top-level comment because the inline reply landed in a now-resolved/collapsed thread.

Took **path B** from your two options: real runtime entrypoint, not a type-only re-export.

- `packages/contracts/esbuild.config.mjs` — added `./src/api/finalize.ts` to `entryPoints` so `dist/api/finalize.mjs` is generated as a standalone module rather than only inlined into the bundled root.
- `packages/contracts/package.json` — re-added `./api/finalize` with both `default` and `types` (mirrors `./api/connectionTest`'s shape).
- `packages/contracts/src/api/finalize.ts` — kept `FINALIZE_SCHEMA_VERSION = 1` as the standalone module's runtime symbol.
- `apps/daemon/src/finalize-design.ts` — switched to `import type { ... } from '@open-design/contracts/api/finalize'`.

CI: ✅ Validate workspace SUCCESS (run `25516118828` at 2026-05-07T19:05 UTC).

**Honest deviation:** I first attempted path A (remove subpath, import from package root). That hit a separate `tsc/NodeNext` resolution issue where the daemon couldn't see the re-exported names through `@open-design/contracts` even though `dist/index.d.ts` had the `export * from './api/finalize'`. Path B sidesteps it by treating `./api/finalize` as a first-class published module identity (the philosophy your `package-runtime.test.ts` encodes).

Inline thread reply: https://github.com/nexu-io/open-design/pull/832#discussion_r3204037943

mrcfps (dismissed):
@bankielewicz thanks for tightening the finalize endpoint and the shared contracts export on this head. I reviewed the updated changed ranges, confirmed the daemon, tests, and contracts wiring now line up, and the live `Validate workspace` check is passing on `29dc7178446a85cb88e228d9f5bae0b9f4043b8d`. Nice follow-through on the earlier feedback. 🙏


lefarcen:
@bankielewicz This PR has merge conflicts that need to be resolved. Please rebase against main and resolve the conflicts, then we can proceed with the merge. Thanks!

bankielewicz:
Rebased onto current `main` (`d592f608`). New head: `c558438f`.

Conflicts were both in `packages/contracts/`:

- `package.json` exports map — your `./api/research` (#898) lined up next to my `./api/finalize`. Combined alphabetically; took the `0.5.0` version bump.
- `esbuild.config.mjs` entryPoints — same shape; both new entries kept, multiline form.

`apps/daemon/src/server.ts` and `packages/contracts/src/index.ts` auto-merged cleanly.

Local verification before push:

- `pnpm --filter @open-design/contracts build` ✓
- `pnpm --filter @open-design/contracts test` — 6/6 ✓
- `pnpm --filter @open-design/daemon test` — 1325/1325 ✓
- `pnpm guard` ✓
- `pnpm typecheck` ✓ (full repo)

CI re-running. The force-push dismissed prior approvals — when you have a moment, would appreciate the re-approve. Thanks!

mrcfps (approved):
@bankielewicz I reviewed the rebased changed ranges on `c558438f`, verified the finalize flow, imported-folder handling, shared contracts export, and the passing workspace check all line up on this head. Thanks for tightening the follow-up fixes here — this looks ready to merge. 🙏

PR #833, opened 2026-05-07
fix: prevent comment popover header overflow when label is too long

nmsn:
## Summary
When the comment popover header contains a long label, the layout breaks and the close button may be pushed out of view.

## Changes
- Add `min-width: 0` and `overflow: hidden` to `.comment-popover-head div`
- Add `text-overflow: ellipsis` and `white-space: nowrap` to `strong` and `span` elements
- Add `flex: 0 0 auto` to close button to keep it fixed width
- Add `title` attribute to header div and close button for hover tooltip

## Files
- `apps/web/src/index.css`
- `apps/web/src/components/FileViewer.tsx`

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -1313,12 +1313,12 @@ function BoardComposerPopover({
       }}
     >
       <div className="comment-popover-head">
-        <div>
+        <div title={target.elementId}>
           <strong id={titleId}>{target.elementId}</strong>
           <span>{target.label}</span>
           <span>{selectionKindLabel(target.selectionKind, target.memberCount)}</span>
         </div>
-        <button type="button" className="ghost" onClick={onClose}>
+        <button type="button" className="ghost" onClick={onClose} title={t('common.close')}>
           {t('common.close')}
         </button>
       </div>
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -5817,11 +5817,25 @@ button.connector-action.is-loading {
 .comment-popover-head div {
   display: grid;
   gap: 1px;
+  min-width: 0;
+  overflow: hidden;
 }
-.comment-popover-head strong { font-size: 13px; }
+.comment-popover-head strong {
+  font-size: 13px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
 .comment-popover-head span {
   color: var(--text-muted);
   font-size: 12px;
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
+}
+.comment-popover-head button.ghost {
+  flex: 0 0 auto;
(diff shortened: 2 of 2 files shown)


lefarcen:
Hi @nmsn! 🎉
Thanks for the contribution — nice catch on the header overflow issue.

— open-design team

nmsn:
Fixes #785

apps/web/src/components/FileViewer.tsx
```diff
@@ -1313,12 +1313,12 @@ function BoardComposerPopover({
       }}
     >
       <div className="comment-popover-head">
-        <div>
+        <div title={target.elementId}>
```
lefarcen, line 1316: **P3 (nit):** The `title` here only shows `target.elementId`, but CSS now truncates three separate values: elementId, label, and selectionKindLabel. If the overflow is caused by a long label, hovering the header shows the wrong/incomplete value.

**Concrete fix:** Either (a) add `title` on each truncated `<strong>`/`<span>` individually with its own value, or (b) compose a full header title like `title={`${target.elementId} · ${target.label} · ${selectionKindLabel(...)}`}` so users can see all three parts.

apps/web/src/index.css
```diff
@@ -5817,11 +5817,25 @@ button.connector-action.is-loading {
 .comment-popover-head div {
   display: grid;
   gap: 1px;
+  min-width: 0;
```
lefarcen, line 5820: **P3 (non-blocking):** This CSS fix itself looks solid, but the visual regression (close button pushed out under long label) isn't covered by tests. Existing Playwright tests only assert the popover appears, not layout stability.

**Concrete fix:** Add a Playwright test with a very long `data-od-id` or label + constrained viewport, then assert the close button bounding box stays within the popover. Or add a visual-regression screenshot if that's the project's pattern.

lefarcen (commented):
Hi @nmsn! Thanks for this UI fix. The flexbox approach (min-width:0 + non-shrinking close button) is directionally sound and should prevent the overflow.

I noticed two areas for polish (both non-blocking P3 suggestions — see inline below):

1. The `title` attribute on the header div only exposes `elementId`, but the three truncated texts include label and selection-kind. If the long value is the label, users won't see the full text on hover.
2. This visual regression isn't currently covered by tests — a Playwright test or screenshot assertion would catch future regressions.

**Related context:** PR #830 (merged 45 min ago) also fixed comment-popover overflow for long note text in the same .index.css area. This PR complements it nicely by addressing the header row. No conflicts expected.

Looks good overall! 🎉

mrcfps (approved):
@nmsn I reviewed the changed ranges in `apps/web/src/components/FileViewer.tsx` and `apps/web/src/index.css`, and verified the flex/min-width/ellipsis update addresses the header overflow path without introducing obvious regressions in the popover structure. Thanks for tightening up this UI edge case — nice focused fix.

PR #838, opened 2026-05-07
fix(web): truncate long inspect-panel labels so they cannot spill past the panel edge (#780)

Nagendhra-web:
Closes #780.

## Problem

The inspect panel renders the selected component's label as a `<strong>` inside `.inspect-panel-title`. The grid container had `min-width: 0` (allowing it to shrink) but the inner `<strong>` rule in [apps/web/src/index.css](https://github.com/nexu-io/open-design/blob/main/apps/web/src/index.css) only set `font-size` with no overflow constraints. A deeply-nested component with a long generated selector path produced a label longer than the 296px panel width, and the text spilled out past the panel's right edge instead of clipping inside the title's background frame.

## Fix

Three properties on `.inspect-panel-title strong`:

```css
white-space: nowrap;
overflow: hidden;
text-overflow: ellipsis;
```

The label now truncates with ellipsis within the panel boundary. The full string remains accessible to users via the `title` attribute already present on the sibling `<code>` element that renders the same selector context.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -5957,8 +5957,8 @@ button.connector-action.is-loading {
   /* Long target.label values (deeply-nested component selectors) used
      to spill past the title container and out of the panel because
      this rule had no overflow constraint. Truncate with ellipsis;
-     the full string remains accessible via the title attribute on
-     the sibling <code>. Issue #780. */
+     the <strong> carries title={label} so the full string is still
+     visible on hover. Issue #780. */
   font-size: 13px;
   white-space: nowrap;
   overflow: hidden;
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -5954,7 +5954,15 @@ button.connector-action.is-loading {
   min-width: 0;
 }
 .inspect-panel-title strong {
+  /* Long target.label values (deeply-nested component selectors) used
+     to spill past the title container and out of the panel because
+     this rule had no overflow constraint. Truncate with ellipsis;
+     the full string remains accessible via the title attribute on
+     the sibling <code>. Issue #780. */
   font-size: 13px;
+  white-space: nowrap;
+  overflow: hidden;
+  text-overflow: ellipsis;
 }
 .inspect-panel-title code {
   color: var(--text-muted);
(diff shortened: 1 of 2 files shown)


lefarcen (commented):
Hi @Nagendhra-web! 👋

This is a clean fix for the overflow issue. The `white-space: nowrap; overflow: hidden; text-overflow: ellipsis;` approach is standard and should work well since `.inspect-panel-title` is already a shrinkable flex child with `min-width: 0`.

**One minor clarification needed:**

The CSS comment states that "the full string remains accessible via the title attribute on the sibling `<code>`" — but looking at the code, the sibling `<code>` has `title={target.selector}`, while the truncated text is `target.label || target.elementId`. These aren't guaranteed to be the same.

**Two options:**
1. Remove that accessibility claim from the comment (just say "truncate with ellipsis"), OR
2. Add `title={target.label || target.elementId}` to the `<strong>` wrapper so the exact truncated label is available on hover.

Either way, the fix itself is solid. Great work! 🎉

apps/web/src/index.css
```diff
@@ -5954,7 +5954,15 @@ button.connector-action.is-loading {
   min-width: 0;
 }
 .inspect-panel-title strong {
+  /* Long target.label values (deeply-nested component selectors) used
+     to spill past the title container and out of the panel because
+     this rule had no overflow constraint. Truncate with ellipsis;
+     the full string remains accessible via the title attribute on
```
lefarcen: P3: The comment claims the full string is accessible via the sibling `<code>` title attribute, but that `<code>` shows `target.selector` while the truncated text is `target.label || target.elementId` (not guaranteed to be the same). Either remove the accessibility claim from this comment or add `title={target.label || target.elementId}` to the `<strong>` wrapper in FileViewer.tsx:1510.

mrcfps (approved):
@Nagendhra-web I reviewed the FileViewer title update and the CSS truncation changes in `apps/web/src/components/FileViewer.tsx` and `apps/web/src/index.css`; the hover title now matches the truncated label, and I didn't find any remaining correctness or maintainability issues in the changed ranges. Thanks for tightening up both the overflow fix and the follow-up accessibility detail. 🎉

PR #840, opened 2026-05-07
fix(web): give MCP server Copy button a solid surface so it reads against the code block (#742)

### LINKED ISSUE #742: Fix MCP server Copy button becoming invisible or hard to see
## Summary

On the MCP server page, the Copy button becomes effectively invisible in some configurations.

In the reported case, the button is still present in the layout, but its styling is so faint or visually collapsed against the code block that users can barely see it or may think it is missing entirely.

## Current behavior

- Open Settings > MCP server.
- View the generated config/code block.
- The Copy button is not clearly visible in some variants of the page.
- It can appear hidden, extremely low-contrast, or visually merged into the background.
- This makes the primary copy action hard to discover.

## Expected behavior

The Copy button should remain clearly visible and discoverable across all MCP server config variants.

Suggested improvements:

- Ensure the Copy button has sufficient contrast against the code block and surrounding container.
- Reserve a stable visible area for the button regardless of config type or content length.
- Prevent the button from blending into the background or clipping against the edge of the block.
- Keep the copy affordance visually consistent across Codex, VS Code, and other MCP config modes.

## Steps to reproduce

1. Open Settings.
2. Go to MCP server.
3. Switch between different config targets if applicable.
4. Inspect the code/config block area.
5. Observe that the Copy button may appear missing or extremely hard to see.

## Notes

- one MCP server state where the Copy button is barely visible / effectively hidden
- another MCP server state where the Copy button is clearly visible as expected

## Why this matters

Copying the MCP config is the main action on this page. If the button is hard to see, users may assume the feature is broken or that there is no copy support at all.

Image

Nagendhra-web:
Closes #742.

## Problem

The Copy button in the MCP server section is positioned absolute over a syntax-highlighted `<pre>` code block. `button.ghost`'s default `background: transparent` let the dark code surface bleed through, so on some themes the button rendered nearly invisible against the snippet backdrop. Users could miss the primary copy affordance entirely, exactly the symptom reported.

## Fix

Three additions to the inline style on the Copy button in [apps/web/src/components/SettingsDialog.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SettingsDialog.tsx):

```tsx
background: 'var(--bg-panel)',
border: '1px solid var(--border)',
boxShadow: '0 1px 2px rgba(0, 0, 0, 0.08)',
```

The button now floats as a visible chip above the code block in both light and dark themes. Hover/disabled behavior remains delegated to the existing `.ghost` class rules so the visual contract elsewhere in the app stays unchanged. The fix is scoped to this one button via inline style; no new shared CSS rules.

Local: web typecheck and `pnpm guard` clean.

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -2423,7 +2423,7 @@ function IntegrationsSection() {
           </pre>
           <button
             type="button"
-            className="ghost"
+            className="ghost mcp-copy-btn"
             onClick={onCopy}
             disabled={!snippet}
             style={{
@@ -2432,15 +2432,6 @@ function IntegrationsSection() {
               right: 8,
               padding: '4px 10px',
               fontSize: 12,
-              // The button floats above the syntax-highlighted code block.
-              // button.ghost's default transparent background let the dark
-              // pre surface bleed through and made the Copy affordance
-              // effectively invisible. Pin a solid panel-colored background
-              // and a visible border so the button reads against either
-              // theme. Issue #742.
-              background: 'var(--bg-panel)',
-              border: '1px solid var(--border)',
-              boxShadow: '0 1px 2px rgba(0, 0, 0, 0.08)',
             }}
             aria-label="Copy MCP configuration snippet"
           >
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -2432,6 +2432,15 @@ function IntegrationsSection() {
               right: 8,
               padding: '4px 10px',
               fontSize: 12,
+              // The button floats above the syntax-highlighted code block.
+              // button.ghost's default transparent background let the dark
+              // pre surface bleed through and made the Copy affordance
+              // effectively invisible. Pin a solid panel-colored background
+              // and a visible border so the button reads against either
+              // theme. Issue #742.
+              background: 'var(--bg-panel)',
+              border: '1px solid var(--border)',
+              boxShadow: '0 1px 2px rgba(0, 0, 0, 0.08)',
             }}
             aria-label="Copy MCP configuration snippet"
           >
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -5963,6 +5963,21 @@ button.connector-action.is-loading {
   overflow: hidden;
   text-overflow: ellipsis;
 }
+
+/* MCP Copy button floats above a syntax-highlighted <pre>; .ghost's
+   transparent background let the dark code surface bleed through and
+   made the affordance effectively invisible. Pin a solid panel
+   background here (not inline) so button.ghost:hover:not(:disabled)
+   still wins the cascade. Issue #742. */
+button.ghost.mcp-copy-btn {
+  background: var(--bg-panel);
+  border: 1px solid var(--border);
+  box-shadow: 0 1px 2px rgba(0, 0, 0, 0.08);
+}
+button.ghost.mcp-copy-btn:hover:not(:disabled) {
+  background: var(--bg-subtle);
+  border-color: var(--border-strong);
+}
 .inspect-section {
   display: flex;
   flex-direction: column;


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2432,6 +2432,15 @@ function IntegrationsSection() {
               right: 8,
               padding: '4px 10px',
               fontSize: 12,
+              // The button floats above the syntax-highlighted code block.
+              // button.ghost's default transparent background let the dark
+              // pre surface bleed through and made the Copy affordance
+              // effectively invisible. Pin a solid panel-colored background
+              // and a visible border so the button reads against either
+              // theme. Issue #742.
+              background: 'var(--bg-panel)',
```
lefarcen: P2 — Inline `background` and `border` override `button.ghost:hover:not(:disabled)` from `index.css`, breaking the hover state change. Move these to a CSS class with hover rules or use a different selector strategy.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for tackling the invisible Copy button! The solid background approach is on the right track, but found one must-fix hover regression and two smaller suggestions:

**P2 — Inline styles break hover behavior:** The `background` and `border` inline styles override `button.ghost:hover:not(:disabled)` from `index.css`, so the hover state change no longer works for this button (disabled/focus still work). The PR body says "hover behavior remains delegated to the existing .ghost class", but that's not true once you set inline background/border. **Fix:** use a scoped CSS class with both base and hover rules instead of inline styles, or use `!important` in the stylesheet hover rule (less ideal).

**P3 — Button may cover snippet text on narrow screens:** Making the button opaque can obstruct the bash one-liner when the viewport is narrow, since the `<pre>` has no top/right padding reserved for the overlay. The old transparent button had the same footprint but was less visible. **Fix:** add padding to the code block or move the copy action to a small header above the `<pre>`.

**P3 — Testing verification:** The "both light and dark themes" claim is good, but no evidence is shown (screenshots/UI tests). Consider adding a screenshot pair in the PR body for future readers.

Once the hover fix lands, this will be solid. Thanks for the clear problem statement and clean scoped approach!

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2432,6 +2432,15 @@ function IntegrationsSection() {
...
+              border: '1px solid var(--border)',
```
mrcfps: These inline `background` and `border` values make the Copy button readable at rest, but they also outrank `button.ghost:hover:not(:disabled)` in `apps/web/src/index.css:216-221`, so this instance no longer gets the hover background/border change that every other ghost button uses. That matters because the PR description and code comment both say hover behavior stays delegated to `.ghost`, but on this code path the button now looks static on hover. Please move the surface styling into a scoped CSS selector (or another non-inline styling path) and define matching hover rules there so the button stays visible without regressing the hover affordance.

mrcfps (commented):
@Nagendhra-web thanks for the focused visibility fix here. I rechecked the changed range in `apps/web/src/components/SettingsDialog.tsx` and found one follow-up issue inline: the new inline surface styles stop this button from inheriting the existing `.ghost` hover affordance, so the visual behavior no longer matches the rest of the UI.


lefarcen (approved):
Nice work @Nagendhra-web! 🎉

The hover regression is now fixed — moving the surface styles to `.mcp-copy-btn` with a dedicated hover rule preserves the ghost button's hover affordance while keeping the Copy button visible against the code block. The approach is clean and doesn't leak into other buttons.

The two P3 suggestions (narrow screen padding + testing evidence) are optional polish, but the must-fix is resolved.

apps/web/src/components/SettingsDialog.tsx
mrcfps: Looper checked this thread against head `f90525d71de3cf84d9ea6c3da0e429fd87287e42`. The requested change appears objectively addressed by Current HEAD matches f90525d71de3cf84d9ea6c3da0e429fd87287e42. The button in apps/web/src/components/SettingsDialog.tsx:2424-2435 now uses className="ghost mcp-copy-btn" with only positioning/padding/fontSize inline, and apps/web/src/index.css:5972-5980 defines scoped button.ghost.mcp-copy-btn base styles plus a matching :hover:not(:disabled) background/border rule., so I’m resolving this thread. Reopen if this still needs discussion.

mrcfps (approved):
@Nagendhra-web I rechecked the changed ranges in `apps/web/src/components/SettingsDialog.tsx` and `apps/web/src/index.css`. Moving the MCP Copy button surface styling into `button.ghost.mcp-copy-btn` preserves the shared ghost hover behavior while keeping the control readable against the code block, and I did not find any remaining correctness, safety, or maintainability issues in this scoped fix. Thanks for the quick follow-up here. 🙏

PR #841, opened 2026-05-07
fix(web): preserve Chat scroll position across Chat/Comments tab switches (#790)

Nagendhra-web:
Closes #790.

## Problem

In [apps/web/src/components/ChatPane.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/ChatPane.tsx) the chat-log container is conditionally rendered (`{tab === 'chat' ? <>...</> : null}`). When the user switches to Comments and back, the chat-log is unmounted and remounted; the remounted element starts at `scrollTop: 0`, and the initial-bottom-scroll effect skips because `didInitialScrollRef.current` is already `true` from the original mount. The conversation view jumps to the top of the long conversation instead of preserving the user's reading position.

## Fix

Replaced the empty-deps scroll listener with a tab-keyed effect that:

1. Captures `scrollTop` inside the existing `onScroll` handler so the saved position is always current.
2. On every chat-log mount (when `tab` becomes `'chat'`), restores the saved `scrollTop` on the next animation frame so layout finishes before the scroll write lands.

```ts
const savedChatScrollTopRef = useRef<number | null>(null);
useEffect(() => {
  if (tab !== 'chat') return;
  const el = logRef.current;
  if (!el) return;
  const saved = savedChatScrollTopRef.current;
  if (saved !== null) {
    requestAnimationFrame(() => {
      const target = logRef.current;
      if (target) target.scrollTop = saved;
    });
  }
  function onScroll() {
    const target = logRef.current;
    if (!target) return;
    savedChatScrollTopRef.current = target.scrollTop;
    const distance = target.scrollHeight - target.scrollTop - target.clientHeight;
    setScrolledFromBottom(distance > 120);
  }
  el.addEventListener('scroll', onScroll);
  return () => el.removeEventListener('scroll', onScroll);
}, [tab]);
```

The existing `scrolledFromBottom` signal that drives the jump-to-bottom button is folded into the same handler and now correctly re-attaches on every chat-log remount, incidentally fixing a secondary issue where that listener silently stopped firing after a tab toggle.

Local: web typecheck and `pnpm guard` clean.

### DIFF
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -156,6 +156,9 @@ export function ChatPane({
 
   useEffect(() => {
     didInitialScrollRef.current = false;
+    // A new conversation should land at the bottom (its own initial
+    // scroll), not inherit the previous conversation's saved position.
+    savedChatScrollRef.current = null;
   }, [activeConversationId]);
 
   useEffect(() => {
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -169,7 +169,13 @@ export function ChatPane({
       el.scrollTop = el.scrollHeight;
       setScrolledFromBottom(false);
     });
-  }, [activeConversationId, messages.length]);
+    // `tab` is in the deps so that switching conversations while
+    // Comments is open doesn't strand the new conversation at scrollTop:
+    // 0. The activeConversationId-reset effect above clears
+    // didInitialScrollRef while the chat-log is unmounted; this effect
+    // then re-runs when the user returns to Chat and the element is
+    // available, scrolling the new conversation to its initial bottom.
+  }, [activeConversationId, messages.length, tab]);
 
   useEffect(() => {
     const el = logRef.current;
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -180,19 +180,42 @@ export function ChatPane({
     }
   }, [messages, error]);
 
+  // Saved chat-log scrollTop, preserved across tab switches.
+  // The chat-log <div> is conditionally rendered (line 331:
+  // `{tab === 'chat' ? ... : null}`) so it unmounts when the user
+  // switches to Comments. On remount it would default to scrollTop: 0
+  // (top of conversation), and the initial-bottom-scroll effect skips
+  // because didInitialScrollRef is already true. Save scrollTop while
+  // Chat is visible and restore it on remount so reading position
+  // survives tab toggles. Issue #790.
+  const savedChatScrollTopRef = useRef<number | null>(null);
   useEffect(() => {
+    if (tab !== 'chat') return;
     const el = logRef.current;
     if (!el) return;
+
+    // Restore previously-saved position on remount. Defer to the next
+    // frame so the conditional <> contents finish layout before the
+    // scrollTop write lands.
+    const saved = savedChatScrollTopRef.current;
+    if (saved !== null) {
+      requestAnimationFrame(() => {
+        const target = logRef.current;
+        if (target) target.scrollTop = saved;
+      });
+    }
+
     function onScroll() {
       const target = logRef.current;
       if (!target) return;
+      savedChatScrollTopRef.current = target.scrollTop;
       const distance =
         target.scrollHeight - target.scrollTop - target.clientHeight;
       setScrolledFromBottom(distance > 120);
     }
     el.addEventListener('scroll', onScroll);
     return () => el.removeEventListener('scroll', onScroll);
-  }, []);
+  }, [tab]);
 
   // Close the conversation history dropdown on outside click / Escape.
   useEffect(() => {
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -218,6 +218,14 @@ export function ChatPane({
         } else {
           target.scrollTop = saved.scrollTop;
         }
+        // Resync the jump-to-latest affordance with the restored
+        // position. Without this, a user who left Chat ~60px from the
+        // bottom and returns to find new messages stacked underneath
+        // would land hundreds of pixels above the latest turn while
+        // scrolledFromBottom remained false until they scrolled.
+        const distance =
+          target.scrollHeight - target.scrollTop - target.clientHeight;
+        setScrolledFromBottom(distance > 120);
       });
     }
 
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -180,15 +180,17 @@ export function ChatPane({
     }
   }, [messages, error]);
 
-  // Saved chat-log scrollTop, preserved across tab switches.
-  // The chat-log <div> is conditionally rendered (line 331:
-  // `{tab === 'chat' ? ... : null}`) so it unmounts when the user
-  // switches to Comments. On remount it would default to scrollTop: 0
-  // (top of conversation), and the initial-bottom-scroll effect skips
-  // because didInitialScrollRef is already true. Save scrollTop while
-  // Chat is visible and restore it on remount so reading position
-  // survives tab toggles. Issue #790.
-  const savedChatScrollTopRef = useRef<number | null>(null);
+  // Saved chat-log scroll state, preserved across tab switches. The
+  // chat-log <div> is conditionally rendered so it unmounts when the
+  // user switches to Comments. On remount it would default to
+  // scrollTop: 0 and the initial-bottom-scroll effect skips because
+  // didInitialScrollRef is already true. We capture either the absolute
+  // scrollTop or a "pinned to bottom" flag while Chat is visible, so
+  // bottom-followers stay pinned even when new messages stream in
+  // off-tab. Issue #790.
+  const savedChatScrollRef = useRef<
+    { pinnedToBottom: true } | { pinnedToBottom: false; scrollTop: number } | null
+  >(null);
   useEffect(() => {
     if (tab !== 'chat') return;
     const el = logRef.current;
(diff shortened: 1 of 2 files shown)


apps/web/src/components/ChatPane.tsx
```diff
@@ -180,19 +180,42 @@ export function ChatPane({
...
+  const savedChatScrollTopRef = useRef<number | null>(null);
   useEffect(() => {
+    if (tab !== 'chat') return;
     const el = logRef.current;
     if (!el) return;
+
+    // Restore previously-saved position on remount. Defer to the next
+    // frame so the conditional <> contents finish layout before the
+    // scrollTop write lands.
+    const saved = savedChatScrollTopRef.current;
```
lefarcen: **P2 — Bottom-pinned state lost when messages arrive off-tab**

The restore uses an absolute `scrollTop` pixel offset. If messages/errors are appended while the user is viewing Comments, returning to Chat restores the old offset, placing them above the new output. A user who left while at the latest message comes back scrolled-up instead of bottom-pinned.

**Suggested fix:** Save distance-from-bottom or a "was pinned" flag. On restore, check if the user was near bottom (distance < 50px?), and if so, set `scrollTop = scrollHeight` to maintain bottom-pin; otherwise restore relative to the new `scrollHeight`.

apps/web/src/components/ChatPane.tsx
```diff
@@ -180,19 +180,42 @@ export function ChatPane({
...
     function onScroll() {
       const target = logRef.current;
       if (!target) return;
+      savedChatScrollTopRef.current = target.scrollTop;
       const distance =
         target.scrollHeight - target.scrollTop - target.clientHeight;
       setScrolledFromBottom(distance > 120);
     }
     el.addEventListener('scroll', onScroll);
     return () => el.removeEventListener('scroll', onScroll);
```
lefarcen: **P3 — Cleanup doesn't snapshot final scroll state**

The cleanup only removes the scroll listener and doesn't snapshot the current `el.scrollTop`. The ref usually updates via scroll events, but it can be stale around programmatic/smooth scrolls or final layout/scroll changes just before the chat log unmounts.

**Suggested fix:** Capture `savedChatScrollTopRef.current = el.scrollTop` in the cleanup before removing the listener.

apps/web/src/components/ChatPane.tsx
```diff
@@ -180,19 +180,42 @@ export function ChatPane({
...
+  // Saved chat-log scrollTop, preserved across tab switches.
+  // The chat-log <div> is conditionally rendered (line 331:
+  // `{tab === 'chat' ? ... : null}`) so it unmounts when the user
+  // switches to Comments. On remount it would default to scrollTop: 0
+  // (top of conversation), and the initial-bottom-scroll effect skips
+  // because didInitialScrollRef is already true. Save scrollTop while
+  // Chat is visible and restore it on remount so reading position
+  // survives tab toggles. Issue #790.
+  const savedChatScrollTopRef = useRef<number | null>(null);
   useEffect(() => {
```
lefarcen, line 203: **P3 — No test coverage for tab-switch scroll restoration**

The only `ChatPane` test currently covers timestamps. Regressions in tab-switch scroll restoration, bottom-pinned restoration, and RAF timing would slip through.

**Suggested fix:** Add a jsdom/RTL test that scrolls `.chat-log`, switches to Comments and back, and asserts restoration. Include the "new message while hidden and previously pinned to bottom" case if you address the P2 above.

lefarcen (commented):
Hey @Nagendhra-web, nice fix — the tab-keyed effect solves the unmount/remount scroll reset cleanly. Found one edge case to consider and two polish suggestions:

**Main concern (P2):** When new messages arrive while the user is on Comments and was previously pinned to bottom, restoring the absolute scrollTop offset brings them back above the new output instead of staying pinned to bottom. See inline comment for a suggested approach.

**Polish (P3):** Cleanup could snapshot scrollTop before removing the listener, and the behavior lacks test coverage for regressions. Details inline.

mrcfps (commented):
@Nagendhra-web thanks for tightening up the remount handling here — the tab-keyed listener/restore flow addresses the basic tab-switch reset cleanly. I found one non-blocking correctness gap around bottom-pinned chats receiving new messages while the Comments tab is open; details are inline.

apps/web/src/components/ChatPane.tsx
```diff
@@ -180,19 +180,42 @@ export function ChatPane({
...
     if (!el) return;
+
+    // Restore previously-saved position on remount. Defer to the next
+    // frame so the conditional <> contents finish layout before the
+    // scrollTop write lands.
+    const saved = savedChatScrollTopRef.current;
+    if (saved !== null) {
+      requestAnimationFrame(() => {
+        const target = logRef.current;
+        if (target) target.scrollTop = saved;
```
mrcfps: When the user leaves Chat while pinned near the bottom, the `[messages, error]` auto-scroll effect above (`ChatPane.tsx:171-181`) stops running because `logRef.current` is `null` while the Comments tab is mounted. That means `saved` here is the old absolute `scrollTop`; if new messages stream in before the user comes back, restoring that pixel offset lands them above the newest output instead of keeping the conversation bottom-pinned. This still breaks the main "follow the latest message" path after a tab switch. Could we persist either a `wasPinnedToBottom` flag or `distanceFromBottom`, then on remount snap to `scrollHeight` when the saved state was near-bottom (or restore relative distance for scrolled-up readers)?

apps/web/src/components/ChatPane.tsx
```diff
@@ -180,19 +180,64 @@ export function ChatPane({
...
 
+  // Saved chat-log scroll state, preserved across tab switches. The
+  // chat-log <div> is conditionally rendered so it unmounts when the
+  // user switches to Comments. On remount it would default to
+  // scrollTop: 0 and the initial-bottom-scroll effect skips because
+  // didInitialScrollRef is already true. We capture either the absolute
+  // scrollTop or a "pinned to bottom" flag while Chat is visible, so
+  // bottom-followers stay pinned even when new messages stream in
+  // off-tab. Issue #790.
+  const savedChatScrollRef = useRef<
```
lefarcen, line 200: **P3 — Scroll state not scoped to conversation**

`savedChatScrollRef` persists across conversation switches. If the user switches conversations while on the Comments tab, returning to Chat restores the previous conversation's scroll state instead of starting fresh at the bottom (the expected initial state for a new conversation).

**Suggested fix:** Reset the ref when `activeConversationId` changes, or key the saved state per conversation if users expect to preserve scroll position when revisiting conversations.

lefarcen (commented):
Great iteration! The bottom-pinned state preservation, cleanup snapshot, and test coverage are all addressed. Found one minor edge case:

**Polish (P3):** The saved scroll state persists across conversation switches, so returning to Chat after switching conversations can restore the previous conversation's scroll position instead of starting fresh. See inline for a quick fix.

lefarcen (commented):
Nice progress! The bottom-pinned preservation and test coverage are solid, and the conversation-switch reset prevents cross-conversation bleed. Found one subtle edge case:

**Conversation-switch while off-tab (P3):** When the user switches conversations while on the Comments tab, returning to Chat doesn't land at the new conversation's intended initial bottom. See inline.

apps/web/src/components/ChatPane.tsx
```diff
@@ -156,6 +156,9 @@ export function ChatPane({
 
   useEffect(() => {
     didInitialScrollRef.current = false;
+    // A new conversation should land at the bottom (its own initial
+    // scroll), not inherit the previous conversation's saved position.
+    savedChatScrollRef.current = null;
   }, [activeConversationId]);
 
   useEffect(() => {
```
lefarcen, line 164: **P3 — Conversation-switch while off-tab doesn't get initial scroll**

When `activeConversationId` changes while on the Comments tab, this effect resets `savedChatScrollRef` (correct), but the initial-bottom-scroll effect at lines 171-181 runs with `logRef.current === null` (Chat is unmounted). When the user later returns to Chat, the restore branch at line 205 sees `saved === null` and skips, leaving the new conversation at default `scrollTop: 0` instead of its initial bottom.

**Suggested fix:** Move the initial-scroll logic into the tab-keyed effect (lines 197-243), or add `activeConversationId` to the initial-scroll deps so it reruns when the conversation changes during Chat tab's mount.

apps/web/tests/components/chat-scroll-preservation.test.tsx
```diff
@@ -0,0 +1,212 @@
...
+      set: (v: number) => {
+        remountedTop = v;
+      },
+    });
+
+    await flushFrame();
+
+    // Saved state was cleared on the conversation switch, so the rAF
+    // restore branch is skipped and scrollTop stays at its initial 0.
+    expect(remountedTop).toBe(0);
```
mrcfps: This assertion is locking in the remaining regression. When `activeConversationId` changes while the Comments tab is mounted, `ChatPane.tsx:157-172` resets the saved state but the initial-bottom-scroll effect bails out because `logRef.current` is `null`, so returning to Chat leaves the new conversation at the browser default `scrollTop = 0`. That matters because a fresh conversation now opens above its latest messages instead of at the bottom path the component comment says should be the initial position. Could we make the tab-keyed mount effect fall back to the initial bottom scroll when `savedChatScrollRef.current === null` and `didInitialScrollRef.current` is still false (or otherwise rerun the initial-scroll logic on Chat remount), then update this test to assert the new conversation lands at its bottom instead?

mrcfps (commented):
@Nagendhra-web thanks for iterating on the tab-switch preservation fix — the pinned-bottom handling and added coverage helped narrow this down. I found one remaining non-blocking correctness gap around switching conversations while the Comments tab is active; details are inline.

lefarcen (commented):
@Nagendhra-web Great iteration! The conversation-switch fix is solid — clearing `savedChatScrollRef` on `activeConversationId` change plus adding `tab` to the initial-scroll deps closes the edge case mrcfps spotted. The new test case locks in that path, and the fix handles both the happy case (returning to a fresh conversation lands at bottom) and the residual saved-position logic (scrolled-up readers keep their place).

apps/web/src/components/ChatPane.tsx
```diff
@@ -180,19 +189,64 @@ export function ChatPane({
...
+    if (tab !== 'chat') return;
     const el = logRef.current;
     if (!el) return;
+
+    // Restore previously-saved position on remount. Defer to the next
+    // frame so the conditional <> contents finish layout before the
+    // scrollTop write lands.
+    const saved = savedChatScrollRef.current;
+    if (saved !== null) {
```
mrcfps, line 212: When `saved !== null`, this restore branch updates `scrollTop` but never recomputes `scrolledFromBottom`. That leaves the jump-to-latest affordance stale after a tab switch: if someone leaves Chat ~60px from the bottom, new messages arrive while Comments is open, and this branch restores the old absolute offset, they can come back hundreds of pixels above the latest turn while `setScrolledFromBottom` is still `false` until they manually scroll. That matters because this PR is explicitly trying to keep the post-remount scroll state accurate, and this path still hides the recovery button in a real changed-range case. Could we recalculate the distance inside this `requestAnimationFrame` (or invoke the same update once after attaching the listener) so `scrolledFromBottom` matches the restored position immediately?

mrcfps (commented):
@Nagendhra-web thanks for iterating on the tab-switch preservation fix — the pinned-bottom path, conversation-switch reset, and new coverage all helped tighten this up. I found one remaining non-blocking correctness gap around the jump-to-latest state after restoring a saved scroll position; details are inline.

lefarcen (commented):
Excellent iteration! The scroll-restoration logic is now complete:

- **Tab-switch preservation** with discriminated-union state (pinnedToBottom vs absolute scrollTop) — fixes the core #790 regression
- **Conversation-switch reset** via `activeConversationId` effect clearing the saved ref — new conversations land at their own bottom
- **Jump-to-latest button** stays synchronized with restored scroll position via the rAF recalculation — fixes the last off-by-one UX gap
- **Test coverage** for all four scenarios (scrolled-up restore, pinned restore with off-tab messages, jump-to-latest visibility after restore, conversation-switch off-tab)

Clean pass from my review. Looks good to me; deferring final approval to a maintainer.

mrcfps (approved):
@Nagendhra-web I re-checked the changed `ChatPane` scroll-restore paths and the new regression coverage for pinned-bottom restores, off-tab conversation switches, and jump-to-latest state. The changed-range issues raised earlier on this head look addressed now, and the final behavior reads consistent with the PR goal.

Thanks for sticking with the iteration here — this was a careful usability fix and the end state looks solid. <sub>Generated by Looper 0.6.2 · runner=reviewer · agent=opencode</sub>

PR #842, opened 2026-05-07
fix(web): keep Design Files tab visible when workspace tabs scroll

### LINKED ISSUE #775: Improve crowded tab strip behavior to keep Design Files and the latest tab visible
## Summary

When many files are open, the tab strip does not handle visibility priorities well.

The `Design Files` entry should remain visible, and the most recently opened file tab should also remain visible. Older tabs in the middle can be compressed, overflowed, or scrolled, similar to how Chrome handles crowded tab bars.

## Current behavior

- Opening many files fills the tab strip quickly.
- The `Design Files` entry can become difficult to see or effectively disappear.
- Newly opened file tabs are not guaranteed to remain clearly visible.
- The tab strip does not appear to prioritize key navigation anchors versus older background tabs.

## Expected behavior

The tab strip should preserve visibility for the most important navigation targets.

Suggested improvements:

- Keep the `Design Files` entry persistently visible.
- Keep the most recently opened or currently active file tab visible.
- Allow older tabs in the middle to compress, scroll, or overflow first.
- Consider a Chrome-like tab handling model where important edge tabs remain reachable while middle tabs absorb the crowding.
- Make the overflow behavior predictable so users can still find both the file list entry and the latest working tab.

## Steps to reproduce

1. Open the Design Files view.
2. Open many files into tabs.
3. Continue until the tab strip becomes crowded.
4. Observe that the `Design Files` entry and newly opened files are not handled with clear visibility priority.

## Notes

Screenshots are attached showing crowded tab-strip states. The desired interaction is closer to a browser-style tab model:
- `Design Files` remains visible
- the latest opened tab remains visible
- middle tabs can be compressed or overflowed

## Why this matters

This is a navigation usability issue, not just a visual one. Users need constant access to both the file list entry and the newest working tab. If the strip does not preserve those anchors, file browsing and recent-context switching become much harder.

Image

Image

Nagendhra-web:
Closes #775.

When enough tabs are open, `.ws-tabs-bar` scrolls horizontally (`overflow-x: auto; flex-wrap: nowrap`) and the Design Files entry slides off the left edge with everything else, so there's no obvious way back to the file list.

This pins the Design Files button at the left with `position: sticky; left: 0` and a small right-side shadow so it stays anchored while the rest of the strip scrolls behind it. Active state gets the existing `--bg-subtle` so it still reads as selected.

Touches only [`apps/web/src/index.css`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/index.css), no JSX changes.

Validated locally: `pnpm --filter @open-design/web typecheck` and `pnpm guard` both clean.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -4642,6 +4642,26 @@ button.connector-action.is-loading {
 .ws-tab.design-files-tab {
   font-weight: 500;
   color: var(--text);
+  /* Pin the Design Files entry to the left edge of the horizontally
+     scrollable tab strip so the user always has a way back to the file
+     list even when many tabs overflow. Without sticky positioning,
+     Design Files sits at flex index 0 and scrolls off the left edge
+     once the strip is wide enough to require horizontal scrolling.
+     Issue #775. */
+  position: sticky;
+  left: 0;
+  z-index: 1;
+  /* Solid panel surface so other tabs that scroll past underneath the
+     sticky element do not visibly bleed through. */
+  background: var(--bg-panel);
+  /* Subtle right-edge shadow signals the visual separation when other
+     tabs are scrolled beneath it. */
+  box-shadow: 4px 0 6px -4px rgba(0, 0, 0, 0.12);
+}
+.ws-tab.design-files-tab.active {
+  /* Active state should still read as selected; layer the active tint
+     over the sticky panel surface. */
+  background: var(--bg-subtle);
 }
 
 .ws-tabs-actions { display: inline-flex; gap: 4px; align-items: center; }


apps/web/src/index.css
```diff
@@ -4642,6 +4642,26 @@ button.connector-action.is-loading {
...
+     Design Files sits at flex index 0 and scrolls off the left edge
+     once the strip is wide enough to require horizontal scrolling.
+     Issue #775. */
+  position: sticky;
+  left: 0;
+  z-index: 1;
+  /* Solid panel surface so other tabs that scroll past underneath the
+     sticky element do not visibly bleed through. */
+  background: var(--bg-panel);
+  /* Subtle right-edge shadow signals the visual separation when other
```
lefarcen, line 4657: P3 The new base `background: var(--bg-panel)` rule has equal specificity to `.ws-tab:hover` (both single-class selectors) and appears later in the cascade, so it suppresses the existing hover background for the non-active Design Files tab. This makes the Design Files tab behave differently from the rest of the tab strip and removes a visual affordance for clickability.

Consider adding:
```css
.ws-tab.design-files-tab:hover {
  background: var(--bg-subtle);
}
```
This preserves the established hover pattern while keeping the sticky tab opaque.

lefarcen (commented):
Hi @Nagendhra-web! Thanks for the PR — this is a clean solution to the sticky Design Files problem.

The `position: sticky` approach is solid: it keeps the tab anchored while the rest of the strip scrolls, and the right-edge shadow nicely signals the separation. The active state override works correctly.

mrcfps (approved):
@Nagendhra-web I verified the CSS-only change in `apps/web/src/index.css`: the Design Files tab now stays pinned via `position: sticky`, the active state still preserves the selected styling, and the update stays isolated to the tab strip without touching the JSX wiring. Thanks for the thoughtful fix here — keeping the file-list entry reachable when tabs overflow is a nice usability improvement. ✨

PR #845, opened 2026-05-07
fix(web): differentiate recent and your designs sorting

Priyanshudotdev:
## Summary

This PR resolves issue #808 by introducing distinct sorting behavior for the "Recent" and "Your Designs" tabs.

Previously, both tabs appeared nearly identical because they relied on similar ordering logic. This update improves usability by making each tab reflect a different browsing purpose:

* "Recent" is now sorted by `updatedAt` (most recently updated first)
* "My Designs" is now sorted by creation time (newest first)

### `apps/web/src/components/DesignsTab.tsx`

* Added separate sorting logic for each tab
* Updated ordering behavior to better reflect user intent
* Preserved existing search and filtering functionality

```typescript
       // separate sorting for recent tab
	if (sub === "recent") {
			list = [...list].sort((a, b) => b.updatedAt - a.updatedAt);
		}
       // separate for yours tab
	if (sub === "yours") {
			list = [...list].sort((a, b) => a.updatedAt - b.updatedAt);
	}
```

## Testing

image

---

image

Verified:

* Correct ordering in both tabs
* Proper tab switching behavior
* No regressions in grid or kanban views

Fixes #808

### DIFF
--- a/apps/web/src/components/DesignsTab.tsx
+++ b/apps/web/src/components/DesignsTab.tsx
@@ -1,373 +1,498 @@
-import { useEffect, useMemo, useState } from 'react';
-import { useT } from '../i18n';
-import { deleteLiveArtifact, fetchLiveArtifacts } from '../providers/registry';
+import { useEffect, useMemo, useState } from "react";
+import { useT } from "../i18n";
+import { deleteLiveArtifact, fetchLiveArtifacts } from "../providers/registry";
 import type {
-  DesignSystemSummary,
-  LiveArtifactSummary,
-  Project,
-  ProjectDisplayStatus,
-  SkillSummary,
-} from '../types';
-import { Icon } from './Icon';
-import { LiveArtifactBadges } from './LiveArtifactBadges';
+	DesignSystemSummary,
+	LiveArtifactSummary,
+	Project,
+	ProjectDisplayStatus,
+	SkillSummary,
+} from "../types";
+import { Icon } from "./Icon";
+import { LiveArtifactBadges } from "./LiveArtifactBadges";
 
-type SubTab = 'recent' | 'yours';
-type ViewMode = 'grid' | 'kanban';
+type SubTab = "recent" | "yours";
+type ViewMode = "grid" | "kanban";
 
 type DesignListItem =
-  | { type: 'project'; project: Project; updatedAt: number }
-  | { type: 'live-artifact'; project: Project; liveArtifact: LiveArtifactSummary; updatedAt: number };
+	| { type: "project"; project: Project; updatedAt: number }
+	| {
+			type: "live-artifact";
+			project: Project;
+			liveArtifact: LiveArtifactSummary;
+			updatedAt: number;
+	  };
 
-const DESIGNS_VIEW_STORAGE_KEY = 'od:designs:view';
+const DESIGNS_VIEW_STORAGE_KEY = "od:designs:view";
 
 export const STATUS_ORDER = [
-  'not_started',
-  'running',
-  'awaiting_input',
-  'succeeded',
-  'failed',
-  'canceled',
+	"not_started",
+	"running",
+	"awaiting_input",
+	"succeeded",
+	"failed",
+	"canceled",
 ] as const satisfies readonly ProjectDisplayStatus[];
 
 export const STATUS_LABEL_KEYS = {
-  not_started: 'designs.status.notStarted',
-  queued: 'designs.status.queued',
-  running: 'designs.status.running',
-  awaiting_input: 'designs.status.awaitingInput',
-  succeeded: 'designs.status.succeeded',
-  failed: 'designs.status.failed',
-  canceled: 'designs.status.canceled',
-} as const satisfies Record<ProjectDisplayStatus, Parameters<ReturnType<typeof useT>>[0]>;
+	not_started: "designs.status.notStarted",
+	queued: "designs.status.queued",
+	running: "designs.status.running",
+	awaiting_input: "designs.status.awaitingInput",
+	succeeded: "designs.status.succeeded",
+	failed: "designs.status.failed",
+	canceled: "designs.status.canceled",
+} as const satisfies Record<
+	ProjectDisplayStatus,
+	Parameters<ReturnType<typeof useT>>[0]
+>;
 
 interface Props {
-  projects: Project[];
-  skills: SkillSummary[];
-  designSystems: DesignSystemSummary[];
-  onOpen: (id: string) => void;
-  onOpenLiveArtifact: (projectId: string, artifactId: string) => void;
-  onDelete: (id: string) => void;
+	projects: Project[];
+	skills: SkillSummary[];
+	designSystems: DesignSystemSummary[];
+	onOpen: (id: string) => void;
+	onOpenLiveArtifact: (projectId: string, artifactId: string) => void;
+	onDelete: (id: string) => void;
 }
 
 export function DesignsTab({
-  projects,
-  skills,
-  designSystems,
-  onOpen,
-  onOpenLiveArtifact,
-  onDelete,
+	projects,
+	skills,
+	designSystems,
+	onOpen,
+	onOpenLiveArtifact,
+	onDelete,
 }: Props) {
-  const t = useT();
-  const [filter, setFilter] = useState('');
-  const [sub, setSub] = useState<SubTab>('recent');
-  const [liveArtifactsByProject, setLiveArtifactsByProject] = useState<Record<string, LiveArtifactSummary[]>>({});
-  const [view, setView] = useState<ViewMode>(() => {
-    if (typeof window === 'undefined') return 'grid';
-    try {
-      const storedView = window.localStorage.getItem(DESIGNS_VIEW_STORAGE_KEY);
-      return storedView === 'grid' || storedView === 'kanban' ? storedView : 'grid';
-    } catch {
-      return 'grid';
-    }
-  });
+	const t = useT();
+	const [filter, setFilter] = useState("");
+	const [sub, setSub] = useState<SubTab>("recent");
+	const [liveArtifactsByProject, setLiveArtifactsByProject] = useState<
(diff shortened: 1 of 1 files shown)


lefarcen:
Hi @Priyanshudotdev! 🎉

Thanks for this PR — the distinct sorting behavior for Recent vs Your Designs is exactly the kind of UX polish that makes the difference.

I'll run a deep review and get back to you within 24h.

— open-design team

apps/web/src/components/DesignsTab.tsx
```diff
@@ -1,373 +1,496 @@
...
+				updatedAt: Date.parse(liveArtifact.updatedAt) || project.updatedAt,
+			})),
+		);
+		list = [...list, ...liveItems];
 
-  const filteredProjects = useMemo(
-    () => filtered.filter((item): item is Extract<DesignListItem, { type: 'project' }> => item.type === 'project'),
-    [filtered],
-  );
+		if (sub === "recent") {
```
lefarcen, line 136: **P1 — Search filter is now unreachable**

Both `sub === "recent"` and `sub === "yours"` branches sort the list and immediately `return list`, so the code below (lines 137-145) that filters by search query `q` is never executed.

When users type in the search box, they'll see no results no matter what they search for.

**Fix**: Apply sorting first, then run the `q` filter before returning:

```typescript
if (sub === "recent") {
  list = [...list].sort((a, b) => b.updatedAt - a.updatedAt);
}
if (sub === "yours") {
  list = [...list].sort((a, b) => a.updatedAt - b.updatedAt); // also fix sort direction below
}

// THEN filter by q
if (!q) return list;
return list.filter(item => ...);
```

apps/web/src/components/DesignsTab.tsx
```diff
@@ -1,373 +1,496 @@
...
+			list = [...list].sort((a, b) => b.updatedAt - a.updatedAt);
+		}
+		if (sub === "yours") {
+			list = [...list].sort((a, b) => a.updatedAt - b.updatedAt);
```
lefarcen: **P1 — Wrong field + wrong direction for "Your Designs"**

Your Designs should show **newest created first** (per #808), but this sorts by `updatedAt` ascending (oldest-updated first).

The `DesignListItem` type doesn't track `createdAt` yet, so you'll need to:

1. Add a `createdAt` field to the item type (pull from `project.createdAt` when available)
2. Change this branch to:

```typescript
if (sub === "yours") {
  list = [...list].sort((a, b) => b.createdAt - a.createdAt);
  ...
}
```

If projects don't have an explicit `createdAt` field, let me know and we'll figure out the best fallback.

apps/web/src/components/DesignsTab.tsx
```diff
@@ -1,373 +1,496 @@
...
+										{skill ? ` · ${skill}` : ""}
+										{" · "}
+										<span
+											className={`design-card-status design-card-status-${status}`}
+										>
+											{statusLabel(status, t)}
+										</span>
+										{p.status?.updatedAt
+											? ` · ${relativeTime(p.status.updatedAt, t)}`
+											: ""}
```
lefarcen: **P2 — Card timestamp doesn't match active tab's sort dimension**

The cards still display `p.status.updatedAt` regardless of which tab is active.

When the user is on **Recent** (sorted by `updatedAt`), the timestamp makes sense. But on **Your Designs** (sorted by creation time), showing the last-updated time is confusing — it doesn't explain the ordering.

**Suggestion**: Conditionally show the timestamp that matches the active sort:

```typescript
{sub === "recent" && p.status?.updatedAt
  ? ` · ${relativeTime(p.status.updatedAt, t)}`
  : sub === "yours" && p.createdAt
  ? ` · ${relativeTime(p.createdAt, t)}`
  : ""}
```

Apply the same logic to kanban cards (line 444) and live-artifact cards (line 306) for consistency.

lefarcen (changes requested):
Hey @Priyanshudotdev! Thanks for tackling this UX improvement — the distinct tab behaviors will definitely make browsing clearer.

I found 3 issues in the sorting logic that need fixing before this ships:

apps/web/src/components/DesignsTab.tsx
Priyanshudotdev: Okay ! I will work no it.

Priyanshudotdev: Done

apps/web/src/components/DesignsTab.tsx
```diff
@@ -1,373 +1,503 @@
...
-    const ok = await deleteLiveArtifact(projectId, artifact.id);
-    if (!ok) return;
-    setLiveArtifactsByProject((current) => ({
-      ...current,
-      [projectId]: (current[projectId] ?? []).filter((candidate) => candidate.id !== artifact.id),
-    }));
-  };
+		if (sub === "yours") {
+			list = [...list].sort(
+				(a, b) => b.project.createdAt - a.project.createdAt,
```
lefarcen: **P1 — Live artifacts not sorted by creation time**

Your Designs now sorts projects by `project.createdAt` (correct!), but live artifacts still use the parent project's `createdAt`. A live artifact created today inside an old project will be sorted with that old project, not at the top.

**Fix**: Add a `createdAt` field to the `DesignListItem` type and use:
- `project.createdAt` for projects
- `Date.parse(liveArtifact.createdAt)` for live artifacts (if available)

Or exclude live artifacts from "Your Designs" if that's the intended product decision.

apps/web/src/components/DesignsTab.tsx
```diff
@@ -1,373 +1,503 @@
...
+										<div className="design-card-meta">
+											<span className="ds">{p.name}</span>
+											{" · "}
+											{artifactStatusLabel(
+												artifact.status,
+												artifact.refreshStatus,
+												t,
+											)}
+											{" · "}
+											{relativeTime(item.updatedAt, t)}
```
lefarcen: **P2 — Live-artifact timestamp doesn't match tab**

Live-artifact cards always show `item.updatedAt`, even when the user is on "Your Designs" (where the list is sorted by creation time). The visible timestamp should match the selected tab's ordering.

Once you add creation timestamps to list items (per the P1 above), render the mode-appropriate timestamp here too.

apps/web/src/components/DesignsTab.tsx
```diff
@@ -1,373 +1,503 @@
...
+											<span>{t("designs.cardFreeform")}</span>
+										)}
+										{skill ? ` · ${skill}` : ""}
+										{" · "}
+										<span
+											className={`design-card-status design-card-status-${status}`}
+										>
+											{statusLabel(status, t)}
+										</span>
+										{sub === "recent" && p.status?.updatedAt
```
lefarcen: **P2 — Recent project timestamp uses wrong field**

Recent is sorted by `project.updatedAt`, but cards display `p.status.updatedAt` (which can diverge — e.g., project renames don't update status).

**Fix**: Use `item.updatedAt` or `p.updatedAt` for Recent cards. Apply the same fix to kanban timestamps at line 447.

lefarcen (changes requested):
Great progress @Priyanshudotdev! ✅ The search filter bug is fixed.

Still need 3 adjustments to fully match the intended behavior:

apps/web/src/components/DesignsTab.tsx
mrcfps: `Your Designs` still is not ordering live artifacts by their own creation time. In this diff, the list entries only carry `updatedAt` (lines 116-127), then this branch sorts every item by `project.createdAt`, so an artifact created today inside an older project is still grouped with that older project instead of rising to the top. That means the new tab behavior does not fully match the PR goal of "newest created first" for all items shown in this view.

Please add a `createdAt` field to `DesignListItem`, populate it from `project.createdAt` for projects and `Date.parse(liveArtifact.createdAt)` for live artifacts, then sort/render the `yours` view from that field so both ordering and timestamps stay aligned.

apps/web/src/components/DesignsTab.tsx
mrcfps: Recent is sorted by `project.updatedAt` (`updatedAt: project.updatedAt` above and the `recent` sort at line 132), but this timestamp still renders `p.status.updatedAt`. Those values can diverge whenever a project changes without a status transition, so the card can show a time that does not explain the visible ordering.

Please render `item.updatedAt` or `p.updatedAt` for the recent view here, and make the same change in the matching kanban branch at lines 447-448, so the metadata matches the actual sort key.

mrcfps (commented):
Thanks @Priyanshudotdev — I left two follow-ups around keeping the Designs tab sort keys and displayed timestamps aligned for projects versus live artifacts.


lefarcen (approved):
Thanks @Priyanshudotdev — all three findings from my last review are now resolved! 🎉

✅ **Your Designs** now correctly sorts all items (projects + live artifacts) by their individual creation time
✅ **Recent tab** timestamps now match the sort key ( for both grid and kanban)
✅ **Your Designs tab** timestamps now match the sort key ( for both grid and kanban)

The changes are clean and consistent across both view modes. Deferring final merge decision to a maintainer.

lefarcen (approved):
Thanks @Priyanshudotdev — all three findings from my last review are now resolved! 🎉

✅ **Your Designs** now correctly sorts all items (projects + live artifacts) by their individual creation time
✅ **Recent tab** timestamps now match the sort key (updatedAt for both grid and kanban)
✅ **Your Designs tab** timestamps now match the sort key (createdAt for both grid and kanban)

The changes are clean and consistent across both view modes. Deferring final merge decision to a maintainer.

mrcfps (approved):
@Priyanshudotdev I re-checked the latest head and reviewed the updated  diff end-to-end. The current changes keep search filtering intact, sort Recent by , and align Your Designs ordering plus displayed timestamps to each item's  across grid and kanban.

Thanks for iterating on this so quickly — the final pass looks consistent and much easier to reason about. 🎉

PR #846, opened 2026-05-07
fix(daemon): make MCP install snippet survive daemon port changes

emilneander:
**Why:** Install snippet bakes `--daemon-url http://127.0.0.1:<port>`,
but tools-dev and packaged both pass DAEMON_PORT=0 so the daemon
picks a fresh port every restart. Pasted client configs break until
edited.

**How:** `od mcp` now discovers the live URL via the daemon's sidecar
IPC status socket on each spawn. Resolution order: `--daemon-url`,
`OD_DAEMON_URL`, IPC discovery, default `http://127.0.0.1:7456`.
Snippet drops `--daemon-url`.

**To test:**
- Open Settings → MCP server: snippet args have no `--daemon-url`.
- Update your client's MCP config (Claude Code, Cursor, etc.) with
  the new snippet, restart the client, run a tool call.
- Restart Open Design (port changes), open a new chat in your
  client, tool calls still work.

**Notes:**
- An already-spawned `od mcp` caches the discovered URL for its
  lifetime; if the daemon restarts mid-session, restart the MCP
  client too.
- Direct `od --port <X>` launches do not open an IPC socket; set
  `OD_DAEMON_URL` or pass `--daemon-url` for non-default ports.

### DIFF
--- a/apps/daemon/src/cli.ts
+++ b/apps/daemon/src/cli.ts
@@ -516,8 +516,8 @@ async function runMcp(args) {
     return;
   }
 
-  const daemonUrl =
-    flags['daemon-url'] || process.env.OD_DAEMON_URL || 'http://127.0.0.1:7456';
+  const { resolveMcpDaemonUrl } = await import('./mcp-daemon-url.js');
+  const daemonUrl = await resolveMcpDaemonUrl({ flagUrl: flags['daemon-url'] });
 
   const { runMcpStdio } = await import('./mcp.js');
   await runMcpStdio({ daemonUrl });
@@ -532,8 +532,15 @@ in another repo so the agent can pull files from a local Open Design
 project without exporting a zip every iteration.
 
 Options:
-  --daemon-url <url>   Open Design daemon HTTP base URL (default: env
-                       OD_DAEMON_URL, falling back to http://127.0.0.1:7456).
+  --daemon-url <url>   Open Design daemon HTTP base URL. Resolution
+                       order: this flag, OD_DAEMON_URL, the running
+                       daemon's sidecar IPC status socket
+                       (/tmp/open-design/ipc/<namespace>/daemon.sock),
+                       then http://127.0.0.1:7456. With IPC discovery
+                       the spawned MCP server tracks the live daemon
+                       URL across restarts even when the daemon binds
+                       an ephemeral port, so MCP client configs do not
+                       need to be re-pasted on each port change.
 
 Tools exposed:
   list_projects                  list every Open Design project
--- a/apps/daemon/src/cli.ts
+++ b/apps/daemon/src/cli.ts
@@ -536,11 +536,13 @@ Options:
                        order: this flag, OD_DAEMON_URL, the running
                        daemon's sidecar IPC status socket
                        (/tmp/open-design/ipc/<namespace>/daemon.sock),
-                       then http://127.0.0.1:7456. With IPC discovery
-                       the spawned MCP server tracks the live daemon
-                       URL across restarts even when the daemon binds
-                       an ephemeral port, so MCP client configs do not
-                       need to be re-pasted on each port change.
+                       then http://127.0.0.1:7456. Each new MCP spawn
+                       discovers the live daemon URL at startup, so
+                       MCP client configs stay valid across daemon
+                       restarts even when the port is ephemeral. A
+                       running MCP server caches the URL; restart the
+                       MCP client after a daemon restart to pick up a
+                       new port.
 
 Tools exposed:
   list_projects                  list every Open Design project
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -1667,9 +1667,15 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     const commandEnv = process.env.ELECTRON_RUN_AS_NODE === '1'
       ? { ELECTRON_RUN_AS_NODE: '1' }
       : null;
+    // The snippet intentionally omits --daemon-url. `od mcp` discovers
+    // the live daemon URL via the sidecar IPC status socket on every
+    // spawn, so the client config stays valid across restarts even
+    // when the daemon binds an ephemeral port (tools-dev, packaged).
+    // OD_DAEMON_URL or --daemon-url remain available as escape hatches
+    // for users who run the daemon outside the sidecar runtime.
     const payload = {
       command: process.execPath,
-      args: [cliPath, 'mcp', '--daemon-url', `http://127.0.0.1:${resolvedPort}`],
+      args: [cliPath, 'mcp'],
       ...(commandEnv == null ? {} : { env: commandEnv }),
       daemonUrl: `http://127.0.0.1:${resolvedPort}`,
       // Surface platform so the install panel can localize path hints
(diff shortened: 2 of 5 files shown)


lefarcen:
Hi @emilneander! 🎉

Thanks for the contribution — this looks like a solid fix for the dynamic daemon port discovery issue.

— open-design team

apps/daemon/src/server.ts
```diff
@@ -1667,9 +1667,15 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
     const commandEnv = process.env.ELECTRON_RUN_AS_NODE === '1'
       ? { ELECTRON_RUN_AS_NODE: '1' }
       : null;
+    // The snippet intentionally omits --daemon-url. `od mcp` discovers
+    // the live daemon URL via the sidecar IPC status socket on every
+    // spawn, so the client config stays valid across restarts even
+    // when the daemon binds an ephemeral port (tools-dev, packaged).
+    // OD_DAEMON_URL or --daemon-url remain available as escape hatches
+    // for users who run the daemon outside the sidecar runtime.
     const payload = {
```
lefarcen, line 1701: P1 — MCP config must carry sidecar namespace.

The generated snippet no longer includes `--daemon-url`, relying on IPC discovery. But MCP clients (Claude Code, Cursor, etc.) don't inherit the daemon's environment variables. When the daemon runs with a non-default namespace (`tools-dev --namespace beta`, or multiple packaged Open Design installs with different namespaces), the spawned `od mcp` process defaults to `OD_SIDECAR_NAMESPACE=default` and looks for `/tmp/open-design/ipc/default/daemon.sock`, which doesn't exist. Discovery fails silently, falls back to `7456`, and the MCP client can't reach the actual daemon.

Concrete fix:
```typescript
const payload = {
  command: process.execPath,
  args: [cliPath, 'mcp'],
  env: {
    ...(commandEnv ?? {}),
    // Pass sidecar discovery env so non-default namespaces work
    ...(process.env.OD_SIDECAR_NAMESPACE && {
      OD_SIDECAR_NAMESPACE: process.env.OD_SIDECAR_NAMESPACE,
    }),
    ...(process.env.OD_SIDECAR_IPC_BASE && {
      OD_SIDECAR_IPC_BASE: process.env.OD_SIDECAR_IPC_BASE,
    }),
  },
  daemonUrl: `http://127.0.0.1:${resolvedPort}`,
  // ...
};
```
Add a test where `OD_SIDECAR_NAMESPACE=custom` and verify the snippet config discovers the right socket.

apps/daemon/src/server.ts
```diff
@@ -1667,9 +1667,15 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
       command: process.execPath,
```
lefarcen, line 1702: P2 — Direct daemon launches (non-sidecar mode) regress.

The PR notes correctly say "direct `od --port <X>` launches do not open an IPC socket", but the generated snippet now *always* omits `--daemon-url`. That means a user who runs `od --port 9999` for debugging gets an MCP config that points at `7456` (the fallback default), not `9999`.

The old behavior was correct for this case: it baked `--daemon-url http://127.0.0.1:${resolvedPort}` into the snippet. The new behavior only works when the daemon is running as a sidecar.

Concrete fix: detect sidecar mode and conditionally include `--daemon-url` for non-sidecar launches:
```typescript
const isSidecarMode = /* check if daemon was started as a sidecar with IPC socket */;
const payload = {
  command: process.execPath,
  args: isSidecarMode
    ? [cliPath, 'mcp']
    : [cliPath, 'mcp', '--daemon-url', `http://127.0.0.1:${resolvedPort}`],
  // ...
};
```
Or document this as a known limitation and require users of `od --port X` to manually set `OD_DAEMON_URL` (but that's a UX regression).

apps/daemon/src/cli.ts
```diff
@@ -532,8 +532,15 @@ in another repo so the agent can pull files from a local Open Design
...
 
 Options:
-  --daemon-url <url>   Open Design daemon HTTP base URL (default: env
-                       OD_DAEMON_URL, falling back to http://127.0.0.1:7456).
+  --daemon-url <url>   Open Design daemon HTTP base URL. Resolution
+                       order: this flag, OD_DAEMON_URL, the running
+                       daemon's sidecar IPC status socket
+                       (/tmp/open-design/ipc/<namespace>/daemon.sock),
+                       then http://127.0.0.1:7456. With IPC discovery
+                       the spawned MCP server tracks the live daemon
```
lefarcen: P3 — Help text incorrectly implies live URL tracking.

The help says "the spawned MCP server tracks the live daemon URL across restarts", but that's not what happens. `runMcp` calls `resolveMcpDaemonUrl` once at process startup and passes the resulting `baseUrl` to `runMcpStdio`, which caches it for the lifetime of the MCP server process. If the daemon restarts on a new port, the already-running MCP server keeps using the stale cached URL until the MCP client restarts that server.

This isn't necessarily wrong behavior (requiring MCP client restart after daemon restart is reasonable), but the help text is misleading.

Concrete fix: reword to clarify the timing:
```
With IPC discovery each *new* MCP spawn discovers the live daemon
URL at startup, so MCP client configs remain valid across daemon
restarts even when the port is ephemeral. An already-running MCP
server caches the URL; restart the MCP client after a daemon restart.
```

lefarcen (changes requested):
Hey @emilneander! Thanks for tackling the dynamic port discovery issue. The IPC discovery approach is the right direction, and I appreciate the thorough test coverage.

However, I found some critical gaps that would break real-world usage:

1. **Non-default namespaces** break discovery — MCP clients launched from Claude/Cursor don't inherit sidecar env, so tools-dev with custom `--namespace` or parallel packaged builds will silently fall back to port 7456.
2. **Direct daemon launches** regress — `od --port X` no longer works because the snippet doesn't adapt to sidecar vs non-sidecar modes.
3. **Help text** incorrectly implies live tracking across restarts, but the MCP server caches the URL at spawn time.

See inline comments for concrete fixes.

apps/daemon/src/cli.ts
emilneander: Hey @lefarcen, thanks for the great feedback! See new changes :)

lefarcen (approved):
Hey @emilneander! 🎉 All three findings are now resolved:

1. ✅ **Non-default namespaces fixed** — sidecar snippets now propagate `OD_SIDECAR_NAMESPACE` and `OD_SIDECAR_IPC_BASE` when non-default, so custom-namespace daemons stay reachable.
2. ✅ **Direct daemon launches fixed** — non-sidecar mode (direct `od --port X`) correctly bakes `--daemon-url` into the snippet.
3. ✅ **Help text clarified** — now accurately says "each new MCP spawn discovers" and explicitly mentions MCP client restart after daemon restart.

The test coverage is comprehensive (default namespace, custom namespace, custom IPC base, non-sidecar). Looks good to me; deferring final merge decision to a maintainer.

mrcfps (approved):
@emilneander I reviewed the updated daemon URL resolution path, the sidecar/non-sidecar MCP snippet generation, and the new install-info/url-resolution tests. The namespace propagation, direct-launch fallback, and restart messaging now line up with the behavior in the changed code. Thanks for tightening this up and for following through on the earlier review feedback. 🎉


leilei926524-tech:
Hi @emilneander — quick note from the Open Design team.

What you get there:

PR #847, opened 2026-05-07
fix(web): reserve clearance for MCP Copy button so it stops overlapping the snippet

Nagendhra-web:
Closes #632.

The Copy button in **Settings > MCP server** is absolutely positioned at `top: 8 right: 8` over the snippet `<pre>` ([SettingsDialog.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SettingsDialog.tsx)), but the `<pre>` only had `padding: '12px 14px'`. Result: the first line of the command sat directly under the button, and a wrapped bash one-liner kept growing right and slid behind the button. The reporter's screenshot shows exactly that overlap.

Rather than restructure the layout, this reserves the clearance inside the `<pre>`'s own padding:

```ts
padding: '40px 80px 12px 14px'
```

- 40px top → first line renders below the button.
- 80px right → wrapped bash content stops short of the button column.
- 12px bottom / 14px left unchanged.

Button stays anchored where it is, no JSX/structure changes, no new CSS rules.

Validated locally: `pnpm --filter @open-design/web typecheck`, `pnpm guard`, and the existing SettingsDialog vitest suite all clean.

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -2399,7 +2399,12 @@ function IntegrationsSection() {
             style={{
               background: 'var(--surface-2, #11141a)',
               color: 'var(--fg-1, #e6e6e6)',
-              padding: '12px 14px',
+              // Reserve top clearance for the absolutely-positioned
+              // Copy button so the first line of the snippet does not
+              // sit underneath it, and reserve right clearance so a
+              // wrapped bash one-liner stops short of the button rather
+              // than scrolling behind it. Issue #632.
+              padding: '40px 80px 12px 14px',
               borderRadius: 8,
               overflowX: 'auto',
               fontFamily:
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -2403,8 +2403,11 @@ function IntegrationsSection() {
               // Copy button so the first line of the snippet does not
               // sit underneath it, and reserve right clearance so a
               // wrapped bash one-liner stops short of the button rather
-              // than scrolling behind it. Issue #632.
-              padding: '40px 80px 12px 14px',
+              // than scrolling behind it. The right padding is sized
+              // for the wider "Copied" post-click state (icon + text +
+              // button padding + the 8px right offset) with a few px
+              // of buffer for elevated font sizes / zoom. Issue #632.
+              padding: '40px 104px 12px 14px',
               borderRadius: 8,
               overflowX: 'auto',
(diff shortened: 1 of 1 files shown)


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2399,7 +2399,12 @@ function IntegrationsSection() {
             style={{
               background: 'var(--surface-2, #11141a)',
               color: 'var(--fg-1, #e6e6e6)',
-              padding: '12px 14px',
+              // Reserve top clearance for the absolutely-positioned
+              // Copy button so the first line of the snippet does not
+              // sit underneath it, and reserve right clearance so a
+              // wrapped bash one-liner stops short of the button rather
+              // than scrolling behind it. Issue #632.
+              padding: '40px 80px 12px 14px',
```
lefarcen: P2 — The 80px right clearance may not be enough for the "Copied" state. The button is `right: 8`, and "Copied" + icon can be ~80-90px wide depending on font/zoom, putting its left edge around `width - 88/90px` while text is allowed through `width - 80px`. Overlap can still occur after clicking.

Concrete fix: bump to at least `padding-right: 104px`, or move the button above the snippet (not absolutely positioned over text) to remove the magic-number dependency.

lefarcen (commented):
Thanks @Nagendhra-web — the approach is sound (reserve clearance in padding rather than restructure), and the comment is clear about the intent.

One edge-case observation on the clearance sizing:

The Copy button switches between "Copy" and "Copied" states. The wider "Copied" text + icon can reach ~80-90px depending on font/zoom. With the button at `right: 8px` and `padding-right: 80px`, the button's left edge and the text's right boundary are close enough that overlap can still occur after clicking, especially if the user's font size or zoom is elevated.

Two options:
- Safer: bump `padding-right` to at least `104px` to accommodate the "Copied" state plus some margin.
- Cleaner long-term: move the button above the snippet (not absolutely positioned over selectable code), which removes the magic-number dependency entirely.

The current change is definitely an improvement over the status quo, so this is non-blocking. Up to you whether to tweak the clearance now or keep it as-is.

— open-design team

mrcfps (approved):
@Nagendhra-web I reviewed the changed `SettingsDialog` range and the follow-up bump to `padding: "40px 104px 12px 14px"` closes the original MCP snippet overlap while also leaving enough room for the wider post-click `Copied` state, without adding layout churn around the button. I did not find any additional correctness or maintainability issues in the scoped diff.

Thanks for tightening this up and for following through on the earlier feedback. 🙂

PR #852, opened 2026-05-08
feat(skills): teach hyperframes skill the HTML-in-Canvas API

pftom:
## Summary

Vendored `skills/hyperframes/` predates upstream `v0.5.1`, which shipped the `drawElementImage` HTML-in-Canvas guide and the `vfx-iphone-device` / `vfx-liquid-glass` / `vfx-portal` family of catalog blocks. The skill that gets injected into every OD agent's system prompt has no mention of any of it, so when a user asks for "live HTML on a 3D phone screen" or "shader-warped dashboard", agents either skip the canvas integration entirely or invent the wrong API. The render itself succeeds — but the device screen comes out blank or frozen, which is the "no effect" symptom we kept seeing.

This PR closes the doc gap. No code changes, no version pin moves.

- Add `skills/hyperframes/references/html-in-canvas.md`, adapted from the upstream `docs/guides/html-in-canvas.mdx` guide. Includes feature detection, per-frame re-capture (the most common cause of dead-looking output), the catalog block table, and a short OD-specific note that our render path already auto-enables `--enable-features=CanvasDrawElement`.
- Surface the new reference from `SKILL.md` and add triggers (`html in canvas`, `drawElementImage`, `html shader`, `vfx-iphone-device`, `vfx-liquid-glass`).

## Why not bump a hyperframes version?

We don't pin one. `apps/daemon/src/media.ts` shells out to `npx hyperframes render`, which always pulls the latest published version (currently `v0.5.3`). The Chrome flag is already auto-enabled by the upstream CLI during render. The only reason the capability looked "missing" was that the agent didn't know how to author code that uses it.

## Test plan

- [x] `pnpm guard`
- [x] `pnpm --filter @open-design/e2e test -- tests/localized-content.test.ts` (skill discovery still aligns with localized copy)
- [ ] Spot-check on a real composition: ask an agent for a `vfx-iphone-device` clip and confirm the screen renders live HTML rather than going blank

### DIFF
--- a/skills/hyperframes/SKILL.md
+++ b/skills/hyperframes/SKILL.md
@@ -14,6 +14,7 @@ triggers:
   - "html shader"
   - "vfx-iphone-device"
   - "vfx-liquid-glass"
+  - "vfx-portal"
 od:
   mode: video
   surface: video
--- a/skills/hyperframes/SKILL.md
+++ b/skills/hyperframes/SKILL.md
@@ -9,6 +9,11 @@ triggers:
   - "captions"
   - "tts video"
   - "kinetic typography"
+  - "html in canvas"
+  - "drawElementImage"
+  - "html shader"
+  - "vfx-iphone-device"
+  - "vfx-liquid-glass"
 od:
   mode: video
   surface: video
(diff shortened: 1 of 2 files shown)


skills/hyperframes/SKILL.md
```diff
@@ -9,6 +9,11 @@ triggers:
   - "captions"
   - "tts video"
   - "kinetic typography"
+  - "html in canvas"
+  - "drawElementImage"
+  - "html shader"
+  - "vfx-iphone-device"
+  - "vfx-liquid-glass"
```
mrcfps, line 16: `skills/hyperframes/SKILL.md` now tells agents to read `references/html-in-canvas.md` when the user asks for `vfx-portal`, but the trigger list added in this hunk still omits that phrase. That means a prompt like "make a `vfx-portal` clip" can miss auto-loading the updated reference, so the exact effect called out in the new docs may still route without the HTML-in-Canvas guidance. Please add `"vfx-portal"` to `triggers:` here, or remove it from the reference callout if it is intentionally unsupported, so the trigger surface matches the documented entry points.

mrcfps (commented):
@pftom Thanks for tightening the HTML-in-Canvas reference and following up on the paint-event issue. I found one small routing follow-up so agents consistently load this guidance for the portal variant too.


mrcfps (approved):
@pftom I re-checked the changed ranges on the current head, including the paint-event guidance in `skills/hyperframes/references/html-in-canvas.md` and the trigger routing in `skills/hyperframes/SKILL.md`. The docs now line up with the follow-up feedback: the examples explain the `onpaint`/`requestPaint()` requirement, and `vfx-portal` is included in the trigger surface so the new HTML-in-Canvas reference is reachable for the documented entry points. Thanks for tightening this up and for following through on the portal routing fix. 🙏

PR #854, opened 2026-05-08
feat(web): wire design-system gallery to create panel and prompt templates

pftom:
## Summary

Picking a design system from the gallery now flows through to the rest of the entry experience instead of just stamping the workspace default and stopping. One coherent thread:

1. **DS gallery** → click a card's new **Use for new project** button.
2. **Create panel** picker mirrors the choice; if the DS surface is image/video, the create tab snaps to image/video.
3. **Prompt templates tab** narrows visible cards to categories implied by the DS metadata, with a dismissible hint and a one-click "Show all templates" escape hatch.

Inspired by the way design-tool entry surfaces typically waste the moment after brand selection — the user picks something, then has to manually re-discover where to use it. This collapses those steps.

## Files changed

- `apps/web/src/components/DesignSystemsTab.tsx` — replace whole-card click-to-select with an explicit `Use for new project` button per card. Cursor only becomes `pointer` on the thumb (still opens preview) and the action button.
- `apps/web/src/components/EntryView.tsx` — new `selectDesignSystemFromGallery` that updates the workspace default DS and, for image/video DS, jumps to the matching prompt-template top tab. Pass `designSystemId` + `designSystems` into `PromptTemplatesTab`.
- `apps/web/src/components/NewProjectPanel.tsx` — when the workspace default DS changes from the gallery, mirror it into the picker primary slot, drop multi-select, and reset the existing `dsSelectionTouched` flag so the prior default-sync effect keeps things aligned. Media-aligned DS additionally switches the create tab to image/video.
- `apps/web/src/components/PromptTemplatesTab.tsx` — accept `designSystemId` / `designSystems`; narrow the visible card set to categories implied by DS metadata when no explicit category filter is set; show a dismissible hint banner with a `Show all templates` button. Templates themselves stay surface-grouped — only the visible card set narrows.
- `apps/web/src/utils/promptTemplateDsCategories.ts` (new) — maps DS title/category/summary → curated prompt-template categories. Kept here so `PromptTemplatesTab` does not couple to per-template schema.
- `apps/web/src/i18n/{types,locales/*}.ts` — `promptTemplates.dsNarrowedHint`, `promptTemplates.showAllTemplates`, `ds.selectForProject` across all 20 locales (incl. `id.ts`).
- `apps/web/src/index.css` — styles for `.ds-card-actions` / `.ds-card-select-btn` and `.prompt-templates-ds-hint` / `-clear`; switch `.ds-card` cursor from `pointer` to `default` and move pointer onto `.ds-card-thumb`.

## Test plan

- [x] `pnpm guard`
- [x] `pnpm --filter @open-design/web test` — 392/392 pass (45 files), incl. the locale-key parity test
- [ ] Manual: in entry view, switch to **Design Systems** tab → click `Use for new project` on a media-aligned brand → verify (a) sidebar tab snaps to image/video, (b) the corresponding prompt-template tab shows the narrowing hint and only relevant categories, (c) clicking `Show all templates` reveals everything, (d) clicking the thumb still opens the preview modal.

### DIFF
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -294,6 +294,27 @@ export function NewProjectPanel({
     setSelectedDsIds(ids);
   }
 
+  // Selecting a design system from the gallery is an explicit intent that
+  // should win over any earlier in-panel picker tweaks: mirror it into the
+  // primary slot, drop multi-select, and clear the touched flag so the
+  // existing default-sync effect above keeps things aligned afterwards.
+  useEffect(() => {
+    if (!defaultDesignSystemId) return;
+    setSelectedDsIds([defaultDesignSystemId]);
+    setDsMulti(false);
+    setDsSelectionTouched(false);
+  }, [defaultDesignSystemId]);
+
+  // Media-oriented design systems pair with their create tabs so the sidebar
+  // matches what the image/video template galleries imply.
+  useEffect(() => {
+    if (!defaultDesignSystemId) return;
+    const ds = designSystems.find((d) => d.id === defaultDesignSystemId);
+    const surf = ds?.surface ?? 'web';
+    if (surf === 'image') setTab('image');
+    else if (surf === 'video') setTab('video');
+  }, [defaultDesignSystemId, designSystems]);
+
   useEffect(() => {
     const el = tabsRef.current;
     if (!el) return;
--- a/apps/web/src/components/PromptTemplatesTab.tsx
+++ b/apps/web/src/components/PromptTemplatesTab.tsx
@@ -4,7 +4,12 @@ import {
   localizePromptTemplateCategory,
   localizePromptTemplateSummary,
 } from '../i18n/content';
-import type { PromptTemplateSource, PromptTemplateSummary } from '../types';
+import type {
+  DesignSystemSummary,
+  PromptTemplateSource,
+  PromptTemplateSummary,
+} from '../types';
+import { inferPromptTemplateCategoriesForDs } from '../utils/promptTemplateDsCategories';
 import { Icon } from './Icon';
 
 // Stable, human-readable provider name used by the source filter and the
@@ -27,6 +32,11 @@ function providerLabel(source: PromptTemplateSource): string {
 interface Props {
   surface: 'image' | 'video';
   templates: PromptTemplateSummary[];
+  // When the workspace default design system is media-aligned, pass it in to
+  // narrow the gallery to categories implied by the DS metadata. Templates
+  // themselves stay surface-grouped; only the visible card set narrows.
+  designSystemId?: string | null;
+  designSystems?: DesignSystemSummary[];
   onPreview: (tpl: PromptTemplateSummary) => void;
 }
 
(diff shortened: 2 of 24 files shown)


apps/web/src/components/PromptTemplatesTab.tsx
```diff
@@ -35,11 +45,34 @@ interface Props {
...
+  }, [designSystemId, designSystems]);
+
+  const dsNarrowCategories = useMemo(() => {
+    if (!activeDs) return null;
+    const inferred = inferPromptTemplateCategoriesForDs(activeDs);
+    if (!inferred || inferred.length === 0) return null;
+    return new Set(inferred.map((c) => c.toLowerCase()));
+  }, [activeDs]);
+
+  const dsNarrowActive = !!dsNarrowCategories && !showAll && category === 'All';
```
lefarcen, line 75: P2 — `dsNarrowActive` doesn't check `activeDs.surface`. Most existing design systems default to `web`, but their category/title text can still match the regex mapper, so visiting prompt templates with a web DS like fintech/developer/e-commerce can unexpectedly hide templates and show a DS-related hint. Gate narrowing to media-aligned systems, ideally the current `surface`, or only pass `designSystemId` when the selected DS surface matches the tab.

apps/web/src/components/PromptTemplatesTab.tsx
```diff
@@ -67,6 +100,13 @@ export function PromptTemplatesTab({ surface, templates, onPreview }: Props) {
       if (category !== 'All' && (tpl.category || 'General') !== category) {
         return false;
       }
+      if (
+        dsNarrowActive
+        && dsNarrowCategories
+        && !dsNarrowCategories.has((tpl.category || 'General').toLowerCase())
```
lefarcen, line 106: P2 — The narrowing activates even if the inferred category set has no intersection with categories available for the current surface. That can turn a populated tab into `emptyNoMatch` while the hint still claims templates are being shown for the DS. Intersect inferred categories with `surfaceScoped` categories and fall back to all templates, or show a specific no-related-templates state with an obvious reset.

lefarcen (commented):
Hey @pftom! The gallery-to-create wiring looks clean — nice explicit button vs whole-card click. Found 2 edge cases in the template narrowing that could hide templates unexpectedly:

**P2 — Surface mismatch**: `dsNarrowActive` doesn't check `activeDs.surface`, so a web DS (fintech/developer/e-commerce) with category text matching the regex can hide templates on image/video tabs even though it's not media-aligned. Gate narrowing to media-aligned systems or only pass `designSystemId` when the selected DS surface matches the tab.

**P2 — Empty intersection**: Narrowing activates even if inferred categories have no templates for the current surface, turning a populated tab into `emptyNoMatch` while the hint still claims templates are shown. Intersect inferred categories with available surface categories and fall back to all templates, or show a specific no-related-templates state.

Also: the PR description doesn't document what happens when DS metadata maps to zero categories or when the user dismisses the hint and then changes DS. Would help reviewers/future-you to spell those out.

Minor — the test plan shows manual tests unchecked; worth running through the scenarios above before merge.

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -294,6 +294,27 @@ export function NewProjectPanel({
...
 
+  // Selecting a design system from the gallery is an explicit intent that
+  // should win over any earlier in-panel picker tweaks: mirror it into the
+  // primary slot, drop multi-select, and clear the touched flag so the
+  // existing default-sync effect above keeps things aligned afterwards.
+  useEffect(() => {
+    if (!defaultDesignSystemId) return;
+    setSelectedDsIds([defaultDesignSystemId]);
+    setDsMulti(false);
+    setDsSelectionTouched(false);
```
mrcfps, line 305: This sync runs whenever `defaultDesignSystemId` is truthy, not only when the selection came from the gallery. On this head that regresses the existing multi-select create flow: CI is already failing `e2e/ui/project-management-flows.test.ts:135` because after picking three systems the trigger still renders `None — freeform+3` instead of keeping a real primary design system, so the create request can no longer reliably persist `designSystemId` plus inspiration metadata. Please gate this reset behind an explicit “selected from gallery” signal (or recompute through `defaultDesignSystemSelection(defaultDesignSystemId, designSystems)` and preserve user-touched multi picks) instead of forcing `[defaultDesignSystemId]` and `dsMulti = false` for every default update.

mrcfps (changes requested):
@pftom Thanks for wiring the gallery selection through the create flow — the explicit CTA and template handoff are headed the right way. I found one blocking regression before merge: the new default-design-system sync is breaking an existing create-panel path, and `Validate workspace` is already red on this head because `e2e/ui/project-management-flows.test.ts` now fails in the multi-select DS flow.


mrcfps:
@pftom Thanks again for the contribution here 💛

PR #857, opened 2026-05-08
fix(daemon): pin OD_DATA_DIR in /api/mcp/install-info env so the macOS-packaged MCP server does not EPERM on .od/projects

### LINKED ISSUE #848: Antigravity MCP Config Error
OS: Mac OS
Antigravity Version: 1.23.2
VSCode OSS Version: 1.107.0
Open Design Version: 0.5.0

Antigravity MCP Config:
`"open-design": {
      "command": "/Applications/Open Design.app/Contents/Resources/open-design/bin/node",
      "args": [
        "/Applications/Open Design.app/Contents/Resources/app/prebundled/daemon/daemon-cli.mjs",
        "mcp",
        "--daemon-url",
        "http://127.0.0.1:54078"
      ]
    }`

It creates following error:

`EPERM: operation not permitted, mkdir '/Applications/Open Design.app/Contents/Resources/app/prebundled/.od/projects' at Object.mkdirSync (node:fs:1350:26) at file:///Applications/Open%20Design.app/Contents/Resources/app/prebundled/daemon/chunks/chunk-RMCVM4DA.mjs:89020:5 at ModuleJob.run (node:internal/modules/esm/module_job:437:25) at async node:internal/modules/esm/loader:639:26 at async file:///Applications/Open%20Design.app/Contents/Resources/app/prebundled/daemon/daemon-cli.mjs:8:1 { errno: -1, code: 'EPERM', syscall: 'mkdir', path: '/Applications/Open Design.app/Contents/Resources/app/prebundled/.od/projects' } Node.js v24.15.0 : calling "initialize": EOF`

I have fixed it again using AI(Gemini 3.1 Pro High) via simple prompting and here is the trace and suggestion:

> The error occurs because the Open Design MCP server is trying to create a .od/projects data directory inside the application bundle at /Applications/Open Design.app/..., which is read-only for security reasons on macOS.
> 
> To fix this, I've modified your MCP server configuration file (mcp_config.json) to include an OD_DATA_DIR environment variable. This variable tells Open Design to store its data in your user directory instead (e.g., ~/.od), avoiding any permissions issues.
> 
> The updated block now looks like this:
```
"open-design": {
      "command": "/Applications/Open Design.app/Contents/Resources/open-design/bin/node",
      "args": [
        "/Applications/Open Design.app/Contents/Resources/app/prebundled/daemon/daemon-cli.mjs",
        "mcp",
        "--daemon-url",
        "http://127.0.0.1:54078"
      ],
      "env": {
        "OD_DATA_DIR": "/Users/<username>/.od"
      }
    }
```
After this environment addition to MCP, it worked as expected.

Nagendhra-web:
Closes #848.

@ozcanoguz reported on packaged Open Design 0.5.0 (macOS, Antigravity 1.23.2) that pointing the IDE's MCP config at the bundled `daemon-cli.mjs` crashed with:

```
EPERM: operation not permitted, mkdir
'/Applications/Open Design.app/Contents/Resources/app/prebundled/.od/projects'
```

@lefarcen confirmed the diagnosis on the issue thread:
> The packaged `od mcp` command imports part of the daemon server module before it dispatches into MCP mode. That module creates the daemon data directories on import. Because the MCP process is launched by Antigravity and does not inherit the packaged app's `OD_DATA_DIR`, it falls back to creating `.od/projects` relative to the packaged bundle.

The reporter's workaround (`env.OD_DATA_DIR = ~/.od`) is exactly the right fix path — it just shouldn't require manual editing of the IDE's MCP config.

## Fix

[`/api/mcp/install-info`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/server.ts) already serializes an `env` object into every client snippet (Cursor / Claude Code / VS Code / Zed / Windsurf / Antigravity / Codex via [`buildMcpStdioServerConfig`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SettingsDialog.tsx)). Previously it only emitted `ELECTRON_RUN_AS_NODE: '1'` when the daemon itself was running inside Electron-as-Node. This adds `OD_DATA_DIR: RUNTIME_DATA_DIR` unconditionally, so the snippet pins the daemon's already-resolved data root and the spawned MCP process writes to the same directory the daemon uses regardless of how the IDE launched it.

`RUNTIME_DATA_DIR` is computed once at daemon startup via the existing `resolveDataDir(process.env.OD_DATA_DIR, PROJECT_ROOT)` helper, so:
- Dev runs from source → `<projectRoot>/.od` (no behavior change vs today, just made explicit in the snippet).
- Packaged macOS / Linux / Windows → whatever the launcher set (`~/Library/Application Support/Open Design`, etc.) — exactly what was missing in the EPERM case.

## Test

`apps/daemon/tests/mcp-install-info.test.ts` already mirrored the handler shape; extended it to construct the env block the same way and added an assertion that `body.env.OD_DATA_DIR === dataDir`. The pre-existing tests (cross-origin reject, loopback accept, cache hit) still cover their original paths.

## Validation

- `pnpm guard` — clean.
- `pnpm --filter @open-design/daemon typecheck` — clean.
- `pnpm --filter @open-design/daemon test -- --run tests/mcp-install-info.test.ts` — fails locally on `Failed to load url undici` from `apps/daemon/src/media.ts`. Reproducible on `origin/main` without my edits, so it's a pre-existing local-environment issue, not something this PR introduces. CI will be the source of truth.

### DIFF
--- a/apps/daemon/tests/mcp-install-info.test.ts
+++ b/apps/daemon/tests/mcp-install-info.test.ts
@@ -7,12 +7,15 @@ import express from 'express';
 import { afterAll, beforeAll, describe, expect, it } from 'vitest';
 import { SIDECAR_DEFAULTS, SIDECAR_ENV } from '@open-design/sidecar-proto';
 import { isLocalSameOrigin } from '../src/server.js';
+import { buildMcpInstallPayload } from '../src/mcp-install-info.js';
 
 // The install-info endpoint is a self-contained handler that resolves
 // absolute paths to node + cli.js so the Settings → MCP server panel
 // can render snippets that work regardless of PATH. We re-build a
-// minimal Express app with the same handler shape rather than booting
-// the full daemon (which needs SQLite, sidecar, fs scaffolding).
+// minimal Express app rather than booting the full daemon (which needs
+// SQLite, sidecar, fs scaffolding), but the payload itself is built by
+// the same exported helper the production handler uses, so any shape
+// divergence (env, args, buildHint) would fail this test.
 
 interface InstallInfoOpts {
   cliPath: string;
@@ -43,19 +46,10 @@ function makeInstallInfoApp({ cliPath, port, env = {}, dataDir }: InstallInfoOpt
       return res.json(cache.payload);
     }
     resolveCalls += 1;
-    const cliExists = fs.existsSync(cliPath);
-    const nodeExists = fs.existsSync(process.execPath);
-    const hints: string[] = [];
-    if (!cliExists) hints.push('cli missing');
-    if (!nodeExists) hints.push('node missing');
 
-    // Mirror the production handler in apps/daemon/src/server.ts:
-    // sidecar-bootstrapped daemons rely on IPC discovery in `od mcp`
-    // and propagate namespace / IPC base through the snippet env;
-    // non-sidecar (direct `od --port X`) launches bake the URL.
-    // OD_DATA_DIR is pinned unconditionally so the spawned process
-    // does not fall back to <cwd>/.od and EPERM inside a packaged
-    // app bundle. Issue #848.
+    // Mirror the production handler's sidecar detection so this test
+    // exercises the same path; the helper below is the same one
+    // server.ts calls.
     const sidecarIpcPath = env[SIDECAR_ENV.IPC_PATH];
     const isSidecarMode = sidecarIpcPath != null && sidecarIpcPath.length > 0;
     const sidecarEnv: Record<string, string> = {};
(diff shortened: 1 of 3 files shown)


lefarcen (commented):
Hey @Nagendhra-web, nice work on the EPERM fix! The core logic is sound: unconditionally pinning `OD_DATA_DIR` in the MCP snippet's `env` block solves the packaged-macOS case where the IDE doesn't inherit the daemon's environment.

✅ **Code correctness (Lens A):**
- Root cause diagnosis is accurate — MCP process launched by Antigravity doesn't inherit `OD_DATA_DIR` from the packaged app.
- Fix is minimal and surgical — adds `OD_DATA_DIR: RUNTIME_DATA_DIR` to the env block without changing other behavior.
- Test coverage added for the new `env.OD_DATA_DIR` field.

**Minor suggestion (non-blocking):** see inline below.

apps/daemon/tests/mcp-install-info.test.ts
```diff
@@ -107,6 +116,13 @@ describe('GET /api/mcp/install-info', () => {
     expect(body.buildHint).toBeNull();
   });
 
+  it('pins OD_DATA_DIR in the env so IDE-spawned MCP processes write to the daemon data dir (issue #848)', async () => {
+    const res = await fetch(`http://127.0.0.1:${port}/api/mcp/install-info`);
+    const body = await res.json();
+    expect(body.env).toBeDefined();
+    expect(body.env.OD_DATA_DIR).toBe(dataDir);
```
lefarcen, line 180: **P3 — Test coupling:** This assertion only verifies the copied test handler (`makeInstallInfoApp`), not the production `/api/mcp/install-info` implementation in `server.ts`. If the real handler later omits `OD_DATA_DIR`, uses the wrong resolved data root, or diverges from this fixture, this test can still pass.

Concrete fix: extract the install-info payload construction into a small exported helper (e.g., `buildMcpInstallPayload({ cliPath, port, dataDir })`) used by both `server.ts` and the test fixture. Then this test verifies the actual production code path, not just a mirror.

Non-blocking — the current test still catches regressions in the env block shape; refactoring to a shared helper is a future cleanup opportunity.

lefarcen (commented):
Perfect, @Nagendhra-web — the extracted helper solves the test-coupling issue cleanly. Now any regression in the env block shape (`OD_DATA_DIR` missing, wrong format, lost `ELECTRON_RUN_AS_NODE`) will actually fail the test instead of passing with a stale fixture mirror.

✅ **Fix complete:**
- `buildMcpInstallPayload` extracted as pure function
- Production handler + test fixture both wired through it
- Test verifies real production code path

Looks good to merge from my side. CI green, fix addresses the root cause, and the test coverage is now solid.

mrcfps (dismissed):
@Nagendhra-web I reviewed the extracted `buildMcpInstallPayload` path in `apps/daemon/src/mcp-install-info.ts`, its wiring in `apps/daemon/src/server.ts`, and the updated daemon test. The fix keeps the packaged Electron-as-Node env handling intact, now always pins `OD_DATA_DIR` for IDE-launched MCP processes, and the live GitHub workspace validation check is green. Thanks for the careful follow-up refactor here — this is in good shape to merge.


lefarcen (approved):

PR #863, opened 2026-05-08
fix(web): show explicit error/retry state when example preview HTML fails to load

### LINKED ISSUE #860: 0.5.0 BUG
Image
The buttons in the upper right corner of the 0.5.0 version example preview are all disabled, and only restarting the program is possible.

Nagendhra-web:
Closes #860 (issue was closed by the reporter, looks like an accidental hit on Close-with-comment when answering @lefarcen's diagnostic ask; happy to have the issue re-opened to track this PR if maintainers prefer).

@lock-down on macOS 26.4.1 packaged 0.5.0 reported the toolbar buttons in the example preview modal greyed out and only the program restart restoring them. @lefarcen pinned the cause on the issue thread:

> These buttons are disabled when the skill example HTML fails to load. The Share menu specifically checks `if (!activeHtml)` before enabling ([PreviewModal.tsx:306](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/PreviewModal.tsx#L306)).

I confirmed by reading the path: [`fetchSkillExample`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/providers/registry.ts) returned `null` on every failure, and [`ExamplesTab.loadPreview`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/ExamplesTab.tsx) cached that `null` so subsequent retries short-circuited. The modal then sat at `preview.loading` forever with no recovery affordance.

## Fix

Three coordinated changes:

1. **`fetchSkillExample`**: returns a discriminated `{ html: string } | { error: string }` instead of collapsing every failure into `null`. Callers can now tell a real fetch failure apart from a successful load.

2. **`PreviewView` gains an optional `error` field.** When set, `PreviewModal` renders a stacked title/body/Retry affordance instead of the indefinite "Loading…" placeholder. The Retry button re-fires the existing `onView` callback for the active view id; parents are responsible for clearing the error and re-running the fetch.

3. **`ExamplesTab` tracks per-skill errors** alongside per-skill html, clears the in-flight value before each fetch (so a retry from the error UI doesn't briefly flash the prior error), and wires `onView={() => loadPreview(...)}` so the Retry button actually retries.

## Scope

Touches only the example-preview path. The other `PreviewModal` consumers (DesignSystemPreviewModal, PromptTemplatePreviewModal, FileViewer, EntryView) keep working unchanged because the new `error` field is optional and they never set it, so the loading-vs-loaded behaviour is identical for them.

## i18n

Three new keys (`preview.errorTitle`, `preview.errorBody`, `preview.retry`) defined in `types.ts` and translated across all 17 locales. The `locales-aligned` test in [`tests/i18n/locales.test.ts`](https://github.com/nexu-io/open-design/blob/main/apps/web/tests/i18n/locales.test.ts) stays green.

## CSS

`.ds-modal-error` extends `.ds-modal-empty` (already absolute-positioned over the iframe stage) with a stacked column layout for title, body, and the Retry button. No other modals styled.

## Validation

- `pnpm guard` clean.
- `pnpm --filter @open-design/web test` clean (392/392 passing, including the i18n alignment check).

Live verification: I don't have packaged 0.5.0 on macOS, so I haven't reproduced the original symptom end-to-end. But the path is straightforward — when `/api/skills/:id/example` returns non-2xx, the modal now renders the error state with a Retry button that actually re-runs the fetch.

### DIFF
--- a/apps/web/src/components/ExamplesTab.tsx
+++ b/apps/web/src/components/ExamplesTab.tsx
@@ -108,6 +108,10 @@ export function ExamplesTab({ skills, onUsePrompt }: Props) {
   const { locale, t } = useI18n();
   // Hold preview HTML per skill across re-renders so cards never re-flicker.
   const [previews, setPreviews] = useState<Record<string, string | null>>({});
+  // Track per-skill fetch failures separately so the preview modal can show
+  // an actionable error / retry state instead of staying stuck at "loading".
+  // Issue #860.
+  const [previewErrors, setPreviewErrors] = useState<Record<string, string>>({});
   const [surfaceFilter, setSurfaceFilter] = useState<SurfaceFilter>('all');
   const [modeFilter, setModeFilter] = useState<ModeFilter>('all');
   const [scenarioFilter, setScenarioFilter] = useState<ScenarioFilter>('all');
@@ -119,11 +123,32 @@ export function ExamplesTab({ skills, onUsePrompt }: Props) {
 
   const loadPreview = useCallback(
     async (id: string) => {
-      if (previews[id] !== undefined) return;
-      const html = await fetchSkillExample(id);
-      setPreviews((prev) => ({ ...prev, [id]: html }));
+      // Skip the fetch only when we already hold a successful html result.
+      // A prior error must not short-circuit a retry; a prior success can.
+      if (previews[id] !== undefined && previewErrors[id] === undefined) return;
+      // Reset both branches before firing so a retry from the error UI
+      // immediately swaps to "loading" instead of flashing the old error.
+      setPreviewErrors((prev) => {
+        if (prev[id] === undefined) return prev;
+        const next = { ...prev };
+        delete next[id];
+        return next;
+      });
+      setPreviews((prev) => ({ ...prev, [id]: null }));
+      const result = await fetchSkillExample(id);
+      if ('html' in result) {
+        setPreviews((prev) => ({ ...prev, [id]: result.html }));
+      } else {
+        setPreviewErrors((prev) => ({ ...prev, [id]: result.error }));
+        setPreviews((prev) => {
+          if (prev[id] === undefined) return prev;
+          const next = { ...prev };
+          delete next[id];
+          return next;
+        });
+      }
     },
-    [previews],
+    [previews, previewErrors],
   );
 
   // Open the modal for a card. We always trigger a preview fetch even if
@@ -334,9 +359,13 @@ export function ExamplesTab({ skills, onUsePrompt }: Props) {
               id: 'preview',
               label: t('examples.previewLabel'),
               html: previews[previewSkill.id],
+              error: previewErrors[previewSkill.id] ?? null,
               deck: previewSkill.mode === 'deck',
             },
           ]}
+          // Wire the modal's Retry button back into loadPreview so a
+          // failure followed by a click re-fires the fetch. Issue #860.
+          onView={() => void loadPreview(previewSkill.id)}
           exportTitleFor={() => previewSkill.name}
           onClose={() => setPreviewSkillId(null)}
         />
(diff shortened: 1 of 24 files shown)


lefarcen:
Hi @Nagendhra-web! 🎉

Thanks for the PR — very clean fix approach turning invisible fetch failures into explicit error states with recovery UI. I will run a deep review and get back to you within 24h.

— open-design team

lefarcen (commented):
Hi @Nagendhra-web,

Thanks for the clean fix! The discriminated union approach is exactly right for turning invisible failures into actionable UI. I have two suggestions for robustness:

## Test coverage (P2)

The new error path isn't exercised by any test. Could you add a test case that:
- Mocks `fetchSkillExample` to return `{error: 'simulated failure'}`
- Verifies `PreviewModal` renders the error state with Retry button
- Verifies clicking Retry re-fires `onView`

This would catch regressions if someone later changes the discriminated union shape or the modal's conditional branches.

## Race condition (P2 — edge case, low-probability)

`ExamplesTab.loadPreview` line 126:
```ts
if (previews[id] !== undefined && previewErrors[id] === undefined) return;
```

If two parallel calls happen before either updates state (e.g. user rapidly clicks two cards), both could pass the check and fire duplicate fetches. React doesn't guarantee atomic read-check-write across two separate state objects.

**Fix:** Track in-flight fetches with a `Set` or `Map` so the second caller sees the first one already running:
```ts
const [inFlight, setInFlight] = useState<Set<string>>(new Set());
// ...
if (inFlight.has(id) || (previews[id] !== undefined && previewErrors[id] === undefined)) return;
setInFlight(prev => new Set(prev).add(id));
// ... fetch ...
setInFlight(prev => { const next = new Set(prev); next.delete(id); return next; });
```

Low-priority (users don't multi-click cards rapidly in practice), but eliminates the theoretical issue.

---

Everything else looks good: i18n keys aligned, CSS scoped, retry UX clear. Deferring final approval to a maintainer.

apps/web/src/components/ExamplesTab.tsx
```diff
@@ -334,9 +387,15 @@ export function ExamplesTab({ skills, onUsePrompt }: Props) {
...
               html: previews[previewSkill.id],
+              error: previewErrors[previewSkill.id] ?? null,
               deck: previewSkill.mode === 'deck',
             },
           ]}
+          // Stable identity (see onPreviewView definition) so PreviewModal's
+          // mount-time onView effect doesn't re-fire on every state update;
+          // the Retry button reaches loadPreview through the same handler.
+          // Issue #860.
+          onView={onPreviewView}
```
mrcfps, line 409: `PreviewModal` calls `onView(activeId)`, where `activeId` is the modal view id (`preview` in this component), but this handler now forwards that argument straight into `loadPreviewRef.current(id)`. That means the mount effect and the Retry button fetch `/api/skills/preview/example` instead of `/api/skills/${previewSkill.id}/example`. When the initial skill fetch fails, clicking Retry never re-requests the selected skill, so the new error state cannot recover and the PR's main fix stays broken. Please keep the callback stable, but close over `previewSkill.id` (or another stable reference to the active skill id) instead of using the modal view id, and add an `ExamplesTab`-level regression test that proves Retry refetches the active skill.

mrcfps (changes requested):
@Nagendhra-web Thanks for the quick follow-up here. I found one blocking regression in the new `onPreviewView` wiring: the modal now forwards its internal view id instead of the active skill id, so the new retry path does not actually refetch the selected example.


Nagendhra-web:
@lefarcen @mrcfps round-3 fix in 0855a1a: `onPreviewView` now closes over the active skill id, Retry test asserts the second fetch hits `live-dashboard` (not `'preview'`), CI green at 4m2s, ready for another pass.

mrcfps (approved):
@Nagendhra-web I reviewed the updated `ExamplesTab`/`PreviewModal` retry wiring and the new regression coverage, and the latest head now closes Retry over the active skill id while keeping the callback stable so the error state can recover without reintroducing the auto-retry loop. The added tests cover both the modal error UI and the end-to-end retry path for the selected skill. Thanks for turning a stuck loading state into a clear recovery flow 🙏

PR #867, opened 2026-05-08
feat: Add a toggle to reveal media provider API keys

nettee:
## Summary
- Mask media provider API key inputs by default in Settings.
- Add an inline visibility toggle with matching eye / eye-off states.
- Cover the hidden-to-visible toggle behavior in SettingsDialog tests.

## Validation
- pnpm --filter @open-design/web test -- SettingsDialog.execution.test.tsx

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -1883,6 +1883,9 @@ function MediaProvidersSection({
   setCfg: Dispatch<SetStateAction<AppConfig>>;
 }) {
   const { t } = useI18n();
+  const [visibleApiKeys, setVisibleApiKeys] = useState<ReadonlySet<string>>(
+    () => new Set(),
+  );
   const providers = MEDIA_PROVIDERS
     .filter((p) => p.settingsVisible !== false)
     .slice()
@@ -1911,6 +1914,17 @@ function MediaProvidersSection({
       return { ...curr, mediaProviders: map };
     });
   };
+  const toggleApiKeyVisibility = (providerId: string) => {
+    setVisibleApiKeys((current) => {
+      const next = new Set(current);
+      if (next.has(providerId)) {
+        next.delete(providerId);
+      } else {
+        next.add(providerId);
+      }
+      return next;
+    });
+  };
 
   return (
     <section className="settings-section">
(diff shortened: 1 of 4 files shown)


lefarcen:
Clean implementation — scoped state, accessible toggle, proper disabled handling, solid test coverage. No security/correctness concerns. 👍

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1911,6 +1914,17 @@ function MediaProvidersSection({
...
+    setVisibleApiKeys((current) => {
+      const next = new Set(current);
+      if (next.has(providerId)) {
+        next.delete(providerId);
+      } else {
+        next.add(providerId);
+      }
+      return next;
+    });
+  };
```
mrcfps, line 1937: In `MediaProvidersSection`, once a provider ID gets added to `visibleApiKeys`, nothing in this new state ever removes it again. If someone reveals a key, clicks **Clear**, and then pastes a replacement key for the same provider in the same dialog session, the input stays `type="text"` instead of returning to masked-by-default. That weakens the security goal of this PR because a newly entered secret can remain exposed after the old credential has been cleared.

Could we remove the provider from `visibleApiKeys` when its config is cleared (or whenever `apiKey` becomes empty) so replacement keys start hidden again?

mrcfps (commented):
Thanks for the security hardening here — the default masking/toggle flow looks good overall. I found one follow-up in the new visibility state handling that seems worth addressing before relying on this as complete masking coverage.


lefarcen:
Good catch on the state lifecycle concern! The implementation already handles this via the `useEffect` at line 1886-1896 — it syncs `visibleApiKeys` with the actual `cfg.mediaProviders` state, removing provider IDs when their `apiKey` becomes empty. The test at line 1752-1760 verifies this: after Clear is clicked, `apiKeyInput.type` returns to `'password'`, and the replacement key stays masked.

So replacement keys do start hidden again — the visibility state resets automatically when the old credential is cleared. 👍

mrcfps (approved):
@nettee I re-reviewed the updated visibility-state sync in `MediaProvidersSection`, the new eye / eye-off toggle wiring, and the replacement-key regression test. The `useEffect` now clears revealed-provider state when an API key is removed, so newly entered replacement keys return to masked-by-default as intended. Thanks for tightening up the security behavior here 🙏

PR #875, opened 2026-05-08
fix(web): confirm before clearing a saved Media provider API key

Nagendhra-web:
Closes #737.

@lefarcen pinned the pattern on the issue thread:

> The codebase already uses \`window.confirm()\` for similar actions (conversation/design deletion), so adding confirmation here would be consistent. Related: #734 has the same pattern for Composio. Both could be fixed together with the same approach.

This PR covers the Media providers Clear button in [SettingsDialog.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SettingsDialog.tsx). The Composio sibling lands as a follow-up PR (per the per-issue rule).

## Fix

Wrap the existing onClick in \`window.confirm()\`, matching the existing destructive-action pattern used in:

- [ChatPane.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/ChatPane.tsx) (conversation delete)
- [DesignsTab.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/DesignsTab.tsx) (design delete)
- [FileWorkspace.tsx](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/FileWorkspace.tsx) (file delete)

```ts
if (!confirm(t('settings.mediaProviderClearConfirm', { name: provider.label }))) {
  return;
}
updateProvider(provider, { apiKey: '', baseUrl: '', model: '' });
```

The prompt names the provider so the user knows exactly which key is about to go: e.g. *"Clear saved OpenAI API key? You'll need to enter it again to use OpenAI."*

## i18n

New key \`settings.mediaProviderClearConfirm\` defined in [types.ts](https://github.com/nexu-io/open-design/blob/main/apps/web/src/i18n/types.ts) and translated across all 17 locales. The \`locales-aligned\` test in [tests/i18n/locales.test.ts](https://github.com/nexu-io/open-design/blob/main/apps/web/tests/i18n/locales.test.ts) stays green (Hungarian needed both \`{name}\` placeholders to match the English shape).

## Tests

- Updated the existing \`SettingsDialog media providers interactions > clears an existing provider config\` test to auto-accept the prompt via \`vi.spyOn(window, 'confirm').mockReturnValue(true)\` so it still asserts the cleared-payload path.
- Added a sibling test \`cancels Clear when the confirmation is dismissed (issue #737)\` that returns false from \`confirm\` and verifies both the form fields stay populated and the saved payload still carries the original config.

## Validation

- \`pnpm guard\` clean.
- \`pnpm --filter @open-design/web exec vitest run tests/components/SettingsDialog.execution.test.tsx tests/i18n/locales.test.ts\` — 56/56 pass (including the new test and the i18n alignment check).

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -3355,7 +3355,23 @@ function MediaProvidersSection({
                   type="button"
                   className="ghost"
                   disabled={!clearable}
-                  onClick={() => updateProvider(provider, { apiKey: '', baseUrl: '', model: '' })}
+                  onClick={() => {
+                    // Match the existing window.confirm guard the rest of
+                    // the app uses for destructive actions (conversation
+                    // delete, design delete, file delete in FileWorkspace).
+                    // Without this a stray click on the row's Clear button
+                    // wipes the saved key with no recovery. Issue #737.
+                    if (
+                      !confirm(
+                        t('settings.mediaProviderClearConfirm', {
+                          name: provider.label,
+                        }),
+                      )
+                    ) {
+                      return;
+                    }
+                    updateProvider(provider, { apiKey: '', baseUrl: '', model: '' });
+                  }}
                 >
                   {t('settings.mediaProviderClear')}
                 </button>
--- a/apps/web/src/i18n/locales/ar.ts
+++ b/apps/web/src/i18n/locales/ar.ts
@@ -146,6 +146,7 @@ export const ar: Dict = {
   'settings.mediaProviderConfigured': 'تم التكوين',
   'settings.mediaProviderUnset': 'غير محدد',
   'settings.mediaProviderClear': 'مسح',
+  'settings.mediaProviderClearConfirm': 'مسح إعدادات {name} المحفوظة؟ ستحتاج إلى إدخالها مرة أخرى لاستخدام {name}.',
   'settings.mediaProviderPlaceholder': 'الصق مفتاح API',
   'settings.mediaProviderBaseUrlPlaceholder': 'تجاوز رابط القاعدة الافتراضي',
   'settings.about': 'حول',
--- a/apps/web/src/i18n/locales/de.ts
+++ b/apps/web/src/i18n/locales/de.ts
@@ -146,6 +146,7 @@ export const de: Dict = {
   'settings.mediaProviderConfigured': 'Konfiguriert',
   'settings.mediaProviderUnset': 'Nicht gesetzt',
   'settings.mediaProviderClear': 'Leeren',
+  'settings.mediaProviderClearConfirm': 'Gespeicherte {name}-Einstellungen löschen? Du musst sie erneut eingeben, um {name} zu nutzen.',
   'settings.mediaProviderPlaceholder': 'API-Key einfügen',
   'settings.mediaProviderBaseUrlPlaceholder': 'Standard-Base-URL überschreiben',
   'settings.about': 'Info',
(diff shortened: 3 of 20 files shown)


lefarcen:
Hi @Nagendhra-web! 🎉
Thanks for the contribution — this adds a safety confirmation step for Media provider API key clearing, matching the existing destructive-action pattern.

— open-design team

mrcfps (dismissed):
@Nagendhra-web I verified the new confirmation guard in `SettingsDialog`, the added i18n key wiring, and the confirm accept/dismiss test coverage for the media-provider clear flow. I did not find any actionable issues in the changed ranges—thanks for tightening this destructive action flow. 🙌


lefarcen (commented):
Hi @Nagendhra-web! The confirm guard is solid — it preserves state on dismiss and clears only after acceptance. The i18n coverage across all 17 locales is complete, and both test paths (accept + dismiss) are covered.

No blocking issues found. Looks good to me; deferring final approval to a maintainer.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2015,7 +2015,23 @@ function MediaProvidersSection({
...
-                  onClick={() => updateProvider(provider, { apiKey: '', baseUrl: '', model: '' })}
+                  onClick={() => {
+                    // Match the existing window.confirm guard the rest of
+                    // the app uses for destructive actions (conversation
+                    // delete, design delete, file delete in FileWorkspace).
+                    // Without this a stray click on the row's Clear button
+                    // wipes the saved key with no recovery. Issue #737.
+                    if (
+                      !confirm(
+                        t('settings.mediaProviderClearConfirm', {
```
mrcfps, line 3366: `clearable` still enables this button when only `baseUrl` or `model` is populated, but the new confirmation string always says we are clearing a saved API key and that the user will need to re-enter it. That means a row with only a custom base URL/model override now shows inaccurate destructive copy in the exact path this PR adds. Since line 1953 treats any populated provider field as clearable, could we either generalize `settings.mediaProviderClearConfirm` to talk about clearing the provider settings/config, or branch the message based on which fields are actually populated? Adding a small test for the base-URL/model-only case would keep that wording honest too.

mrcfps (commented):
@Nagendhra-web Thanks for tightening this destructive-action flow and covering both confirm paths in tests. I found one small follow-up in the new confirmation copy that would make the prompt match the exact settings being cleared.


mrcfps (dismissed):
@Nagendhra-web I re-checked the SettingsDialog confirm guard, the accept/dismiss test coverage around the Clear flow, and the locale/fallback wiring that landed in this head. I did not find any actionable issues in the changed ranges—thanks for tightening this destructive-action path and cleaning up the locale coverage. 🙌


lefarcen (commented):
Hey @Nagendhra-web! 👋

**mrcfps's inline finding resolved** — the confirm copy now says "Clear saved {name} settings?" which accurately matches what  clears (any of //).

**Drive-by localized-content fix looks correct** —  added alphabetically to the three EN fallback arrays, which is the right mechanism for .

**CI failure is unrelated** — the failing test () is in a different area from your SettingsDialog/i18n changes. This looks like a flake in the scroll restoration logic.

lefarcen (commented):
Hey @Nagendhra-web! 👋

**mrcfps's inline finding resolved** — the confirm copy now says "Clear saved {name} settings?" which accurately matches what clearable clears (any of apiKey/baseUrl/model).

**Drive-by localized-content fix looks correct** — mission-control added alphabetically to the three EN fallback arrays, which is the right mechanism for localized-content.test.ts.

**CI failure is unrelated** — the failing test (chat-scroll-preservation.test.tsx) is in a different area from your SettingsDialog/i18n changes. This looks like a flake in the scroll restoration logic.

mrcfps (approved):
@Nagendhra-web I verified the new confirmation guard in `SettingsDialog`, the updated `settings.mediaProviderClearConfirm` locale wiring, and the accept/dismiss coverage in `SettingsDialog.execution.test.tsx`. I did not find any actionable issues in the changed ranges—thanks for tightening this destructive-action flow and carrying the follow-up cleanup through. 🙌

PR #882, opened 2026-05-08
feat: vector PDF export for HTML decks (+ python3 fix in PPTX prompt)

yalkongs:
## Summary

- Adds `POST /api/projects/:id/export-pdf` and a Playwright-based exporter that emits one **vector** PDF page per slide, then merges with `pdf-lib`. Text, SVG, and CSS borders stay sharp at any print resolution; canvas/WebGL is rasterized by Chromium internally.
- Surfaces the new export as an **\"Export PDF — high quality\"** share-menu item in `FileViewer` (gated to deck-shaped HTML).
- Fixes one PPTX verify-layout prompt to use `python3` instead of `python` so dev machines without a `python` symlink (modern macOS, several Linux distros) don't fail with `ModuleNotFoundError: No module named 'pptx'`.

## Why

The existing client-side \"Export as PDF\" path in `FileViewer` produces raster PDFs (each slide is a screenshot), so printed text and lines are noticeably soft even at 300 DPI. Several real decks already declare proper `@media print` rules, so a server-side `page.pdf()` pass yields a vector PDF essentially for free.

## Capture strategy

The exporter tries two strategies in order:

1. **Trust the deck's `@media print` rules.** Single `page.pdf()` call with `preferCSSPageSize: true`. If the resulting page count equals the slide count, that's the result. Most well-built decks land here.
2. **Fall back to per-slide screen-mode isolation.** Navigate via the deck's own JS conventions (toggling `active` and `is-active` classes, resetting `#deck` transform/width), capture each slide at 1920×1080, merge with `pdf-lib`.

Verified on three deck shapes (sc-deck-final horizontal-scroll, im-stablecoin-manual display-toggle, and an `is-active` deck).

## Why root `scripts/`

`e2e/scripts/` is now restricted to `playwright.ts` only by `pnpm guard`. Putting the exporter at root `scripts/` keeps Node module resolution simple (root `node_modules` carries `@playwright/test` and `pdf-lib` via the new root devDependencies entries).

## Test plan

- [ ] `pnpm guard` passes
- [ ] `pnpm typecheck` passes  
- [ ] Manual: trigger \"Export PDF — high quality\" on a deck — file downloads, opens in a viewer, vector text is selectable in PDF reader
- [ ] Verify with `pdftotext` that text is extractable (proves vector glyphs, not raster)
- [ ] Verify with `pdfinfo` that the page size matches the deck's @page declaration
- [ ] Print preview at 300+ DPI to confirm sharp glyphs and rules

## Out of scope (intentionally not included)

These are personal/local customizations the author maintains separately and are NOT pushed here:
- Hardcoded \"slide-resource directory\" injection (`~/code/slideresource`)
- A `start-dev.sh` convenience launcher
- A slide-agent `--append-system-prompt` directive that protects against parent `CLAUDE.md` / `AGENTS.md` leakage
- Default-model fallback resolution from app config
- Model registry refresh (Opus 4.7 / Sonnet 4.6) — separate concern, separate PR if wanted


lefarcen:
Hi @yalkongs! 🎉

Thanks for this PR — vector PDF export is a high-quality improvement over raster screenshots, and the python3 fix addresses a real cross-platform issue.

— open-design team

apps/daemon/src/server.ts
```diff
@@ -3923,6 +3923,64 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+    // Reject path traversal attempts
+    if (filename.includes('/') || filename.includes('\\') || filename.startsWith('.')) {
+      return res.status(400).json({ error: 'invalid filename' });
+    }
+
+    const htmlPath = path.join(PROJECTS_DIR, req.params.id, filename);
+    if (!fs.existsSync(htmlPath)) return res.status(404).json({ error: 'html file not found' });
+
+    const pdfPath = path.join(os.tmpdir(), `od-pdf-${randomUUID()}.pdf`);
+    const scriptPath = path.join(PROJECT_ROOT, 'scripts', 'export-slides-pdf.ts');
```
lefarcen, line 3950: **P1** Runtime packaging blocker: This endpoint spawns `node` from PATH to run `scripts/export-slides-pdf.ts`, which depends on root `devDependencies` (`@playwright/test`, `pdf-lib`). In packaged builds:
- `PROJECT_ROOT/scripts/export-slides-pdf.ts` may not exist (only bundled daemon code ships)
- Root devDeps aren't bundled with the daemon runtime
- Users may not have `node` on PATH even though the daemon is running

Result: The FileViewer button ships but fails with 501/MODULE_NOT_FOUND/ENOENT in release installs.

**Fix:** Move the exporter into daemon-owned runtime code (e.g., `apps/daemon/src/export-pdf/`), declare the runtime deps in the daemon's `package.json` (not root devDeps), and spawn via `process.execPath` or the packaged Node path.

apps/daemon/src/server.ts
```diff
@@ -3923,6 +3923,64 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+
+    const filename = typeof req.body?.file === 'string' ? req.body.file.trim() : '';
+    if (!filename) return res.status(400).json({ error: 'file required' });
+
+    // Reject path traversal attempts
+    if (filename.includes('/') || filename.includes('\\') || filename.startsWith('.')) {
+      return res.status(400).json({ error: 'invalid filename' });
+    }
+
+    const htmlPath = path.join(PROJECTS_DIR, req.params.id, filename);
```
lefarcen, line 3946: **P2** This route bypasses the existing project-file abstraction. It builds `.od/projects/:id/:filename` directly and rejects any slash, so:
- Linked-folder projects that work through `readProjectFile()`/`resolveProjectDir()` can't be exported
- Nested project files are unreachable
- Misses the symlink-aware `resolveSafeReal()` path validation used by other daemon file APIs

This is both a correctness regression for supported project shapes and weaker security hygiene.

**Suggestion:** Use the existing project-file resolution helpers (`readProjectFile`, `resolveProjectDir`) to match the daemon's file-access patterns.

apps/daemon/src/server.ts
```diff
@@ -3923,6 +3923,64 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+
+    const pdfPath = path.join(os.tmpdir(), `od-pdf-${randomUUID()}.pdf`);
+    const scriptPath = path.join(PROJECT_ROOT, 'scripts', 'export-slides-pdf.ts');
+
+    if (!fs.existsSync(scriptPath)) {
+      return res.status(501).json({ error: 'export script not found' });
+    }
+
+    try {
+      await new Promise((resolve, reject) => {
```
lefarcen, line 3957: **P2** The export child has no timeout, abort handling, or cleanup. If Playwright hangs (networkidle, font loading, GPU/browser startup, or a very large deck), the HTTP request waits indefinitely. If the client disconnects, the Chromium/export child keeps running.

Since the PR expects 30-60s for large decks, this should either:
- Be an async task with polling/progress (like existing media tasks), OR
- Have a bounded timeout, `req/res close` cancellation, and child kill cleanup

**Suggestion:** Add `setTimeout` + `child.kill()` on timeout/abort, and clean up Chromium processes on client disconnect.

apps/daemon/src/server.ts
```diff
@@ -3923,6 +3923,64 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+        const child = spawn(
+          'node',
+          ['--experimental-strip-types', scriptPath, htmlPath, pdfPath],
+          { cwd: PROJECT_ROOT, env: { ...process.env, INIT_CWD: PROJECT_ROOT }, stdio: ['ignore', 'pipe', 'pipe'] },
```
lefarcen, line 3961: **P2** Child stdout/stderr are piped but never consumed. Chromium/Playwright failures can be noisy; if either pipe fills, the child blocks and the request hangs.

**Fix:** Use `stdio: ['ignore', 'ignore', 'pipe']` (ignore stdout, collect stderr), or actively consume bounded stderr for the 500 response.

apps/daemon/src/server.ts
```diff
@@ -3923,6 +3923,64 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+
+      const stat = fs.statSync(pdfPath);
+      const pdfName = filename.replace(/\.html$/i, '') + '.pdf';
+      res.setHeader('Content-Type', 'application/pdf');
+      res.setHeader('Content-Length', stat.size);
+      res.setHeader(
+        'Content-Disposition',
+        `attachment; filename*=UTF-8''${encodeURIComponent(pdfName)}`,
+      );
+      const stream = fs.createReadStream(pdfPath);
```
lefarcen, line 3974: **P2** PDF read/stream errors are not handled. `fs.statSync()` is caught, but `createReadStream(pdfPath)` failures happen asynchronously. Without an `error` listener, a missing/unreadable temp file after `stat` can crash the daemon or leave the response hanging after headers.

**Fix:** Add stream error handling:
```ts
stream.on('error', (err) => {
  fs.unlink(pdfPath, () => {});
  if (!res.headersSent) res.status(500).json({ error: `Stream failed: ${err.message}` });
});
```

scripts/export-slides-pdf.ts
```diff
@@ -0,0 +1,164 @@
...
+    }
+    slides.forEach((s, si) => {
+      s.style.display = si === idx ? '' : 'none';
+    });
+  }, i);
+
+  // One frame of paint time
+  await page.waitForTimeout(120);
+
+  const pdfBytes = await page.pdf({
```
lefarcen, line 143: **P2** The fallback silently truncates overflow pages. `pageRanges: '1'` + `copyPages(slidePdf, [0])` force one output page even if the isolated slide paginates to multiple pages (unexpected sizing/margins/content overflow). This violates the "one page per slide" guarantee without reporting.

**Fix:** Check `slidePdf.getPageCount()` before copying:
```ts
if (slidePdf.getPageCount() > 1) {
  console.warn(`Slide ${i+1} produced ${slidePdf.getPageCount()} pages — truncating to first page`);
}
```
And document this fallback failure mode in the script's top comment.

scripts/export-slides-pdf.ts
```diff
@@ -0,0 +1,164 @@
...
+const baseCwd = process.env['INIT_CWD'] ?? process.cwd();
+const htmlPath = resolve(baseCwd, htmlArg);
+const outputPath = process.argv[3]
+  ? resolve(baseCwd, process.argv[3])
+  : join(dirname(htmlPath), basename(htmlPath, '.html') + '.pdf');
+
+console.log(`Source : ${htmlPath}`);
+console.log(`Output : ${outputPath}`);
+
+const browser = await chromium.launch({ headless: true });
```
lefarcen, line 47: **P3** Browser cleanup is not in a `finally`. Several failure paths after launch (e.g., `page.goto`, `page.pdf`, `PDFDocument.load`, `writeFileSync`) can exit before `browser.close()`, potentially orphaning Chromium processes.

**Fix:** Wrap in try/finally:
```ts
try {
  // ... existing logic
} finally {
  await browser.close();
}
```

apps/web/src/components/FileViewer.tsx
```diff
@@ -4328,6 +4329,57 @@ function HtmlViewer({
                         : t('fileViewer.exportPdf')}
                     </span>
                   </button>
+                  {effectiveDeck ? (
+                    <button
+                      type="button"
+                      className="share-menu-item"
+                      role="menuitem"
+                      disabled={exportingServerPdf}
+                      onClick={async () => {
```
lefarcen, line 4338: **P3** The long-running export gives little visible feedback. The menu closes immediately, and "Exporting PDF…" is only visible inside the closed menu, so a 30-60s sync export looks like nothing is happening until a download or alert.

**Suggestion:** Surface a visible toast/status (and ideally cancellation). Async task progress (like existing media tasks) would be even better for long-running operations.

lefarcen (changes requested):
Hey @yalkongs, thanks for this PR! The vector PDF export idea is solid and the two-strategy fallback is well-designed. However, there are several implementation-boundary issues that would cause failures in packaged/release installs.

**Key blockers:**
- The daemon endpoint runs a root TypeScript script via `node` from PATH, which won't work in packaged builds
- Runtime depends on root `devDependencies` that aren't bundled with the daemon
- Missing timeout/cancellation for long-running Playwright operations

See inline comments for specifics. Once these are addressed (particularly the runtime packaging issue), this will be ready to ship.

scripts/export-slides-pdf.ts
```diff
@@ -0,0 +1,164 @@
...
+const page = await browser.newPage();
+await page.setViewportSize({ width: 1920, height: 1080 });
+await page.goto(`file://${htmlPath}`, { waitUntil: 'networkidle' });
```
mrcfps, line 50: The new `page.goto(\`file://${htmlPath}\`)` call builds a raw file URL, so deck filenames that contain URL-significant characters from linked/imported folders (for example `deck#1.html` or `deck?final.html`) get parsed as a fragment/query and the exporter opens the wrong path. That breaks a valid project file even though the rest of the app already treats arbitrary on-disk names as supported input — `projectRawUrl()` explicitly encodes each segment for that reason (`apps/web/src/providers/registry.ts:998-1003`). Please convert the path with `pathToFileURL(htmlPath).href` before calling Playwright so special characters are escaped consistently.

mrcfps (commented):
@yalkongs Thanks for shipping the vector PDF export path — I found one additional correctness issue in the new exporter script that looks worth fixing before this is fully robust.


lefarcen:
@yalkongs Thanks for this PR contribution. Closing due to 72h inactivity after review feedback (no commits or comments since 2026-05-08).

The PR idea (vector PDF export) is valuable, but the implementation needs fixes:
- **P1 blocker:** Runtime packaging — daemon endpoint spawns root TypeScript script via `node` from PATH with root `devDependencies`, which won't work in packaged builds
- **P2 issues:** Missing timeout/cancellation for long-running Playwright operations, path validation bypass, file URL escaping for special characters in deck filenames

Feel free to reopen if you'd like to address the review feedback. We'd be happy to revisit once the blockers are fixed.

— open-design team

Issue #648, opened 2026-05-06
Bug: Media generation task state is volatile and lost on daemon restart


pmasadali20776-ui:
The daemon application currently manages media generation tasks using an in-memory Map (mediaTasks). While the actual
  media generation process may create files on disk, the metadata and status tracking for these tasks reside entirely in
   RAM.

  Impact:
  If the daemon process restarts (due to a crash, update, or manual restart), all currently active and recently
  completed task metadata are wiped. This leads to:
  1. Orphaned Processes/Files: Tasks that were in progress may continue to run or leave partial files on disk with no
  corresponding record in the system.
  2. Loss of Visibility: Users lose the ability to track the progress or retrieve the status of tasks that were
  initiated before the restart.
  3. Inconsistent State: The filesystem may contain generated media that the application no longer "knows" about.

  Suggested Fix:
  Persist the mediaTasks state to a lightweight database (like SQLite) or a JSON state file in the RUNTIME_DATA_DIR so
  that task state can be recovered upon daemon startup.

  Exact Location of Problem:
  File: apps/daemon/src/server.ts
  Line: 511
  Code: const mediaTasks = new Map<string, MediaTask>();

lefarcen:
This looks like a good first issue for someone new to the project.

If you'd like to work on it, comment `/claim` below. The first `/claim` after the latest release gets it. Any comment, linked PR, or progress note keeps the claim active.

lefarcen:
Thanks for the report. I checked the current daemon implementation and this is confirmed.

The line number has moved in the current tree, but the issue still exists: `apps/daemon/src/server.ts` currently declares `const mediaTasks = new Map()` around line 945. Media generation creates a task in that map, `/api/media/tasks/:id/wait` reads from that map, and `/api/projects/:id/media/tasks` lists from that map. There is no corresponding `media_tasks` table, JSON state file, or boot recovery path.

The generated media bytes themselves are written under the project data directory (`RUNTIME_DATA_DIR/projects/<projectId>/...`), but the task metadata is process-local. That means a daemon restart loses the task id, progress, status, terminal error/file metadata, and any waiters. After restart, polling the old task id will return `404 task not found`.

This affects the media dispatcher generally, especially long-running video/HyperFrames/provider-polling tasks where the CLI is expected to keep calling `media wait <taskId>`. The repo already has a good nearby pattern for this in Critique Theater: it persists `critique_runs` in SQLite and reconciles stale `running` rows on daemon boot.

[...]

PR #884, opened 2026-05-08
Bug FIx: Media generation task state is volatile and lost on daemon restart #648

pmasadali20776-ui:
Bug has been fixed from Issue #648


lefarcen:
Hi @pmasadali20776-ui! 🎉

Thanks for the contribution — this addresses a real data loss risk at daemon restart. I will run a deep review and get back to you within 24h.

— open-design team

apps/daemon/src/server.ts
```diff
@@ -1599,6 +1679,19 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+    terminalTtlMs: TASK_TTL_AFTER_DONE_MS,
+  });
+  if (mediaReconcile.interrupted > 0 || mediaReconcile.deleted > 0) {
+    console.warn(
+      `[media] reconcileMediaTasksOnBoot interrupted ${mediaReconcile.interrupted} task(s), ` +
+        `deleted ${mediaReconcile.deleted} expired terminal task(s)`,
+    );
+  }
+  mediaTasks.clear();
+  for (const row of listRecentMediaTasks(db, { terminalTtlMs: TASK_TTL_AFTER_DONE_MS })) {
```
lefarcen, line 1906: **P2** Terminal tasks hydrated after restart never get a TTL cleanup timer. Boot only deletes rows already expired at boot, and runtime deletion is only scheduled by `notifyTaskWaiters()` when a task transitions terminal in the current process.

Scenario: a task completes before shutdown, daemon restarts within the 10-minute TTL, then stays up — that task remains in memory/SQLite indefinitely.

Fix: schedule GC when hydrating terminal rows using `endedAt/updatedAt + TASK_TTL_AFTER_DONE_MS`, or centralize TTL cleanup in route/read paths. Add a restart-within-TTL-then-expire test to catch this.

apps/daemon/src/server.ts
```diff
@@ -3878,7 +3974,7 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
       return res.status(403).json({ error: 'cross-origin request rejected' });
     }
     const taskId = req.params.id;
-    const task = mediaTasks.get(taskId);
+    const task = getLiveMediaTask(db, taskId);
```
lefarcen, line 4591: **P2** `/api/media/tasks/:id/wait` hydrates any persisted task by ID without checking terminal TTL. If an expired terminal row survived boot or missed cleanup, `/wait` keeps returning `200` instead of expiring to `404`, breaking the intended retention semantics.

Fix: make `getLiveMediaTask()` reject/delete expired terminal rows before hydrating.

apps/daemon/src/server.ts
```diff
@@ -1442,15 +1477,41 @@ function createMediaTask(taskId, projectId, info = {}) {
     waiters: new Set(),
   };
   mediaTasks.set(taskId, task);
```
lefarcen, line 1678: **P3** `createMediaTask()` inserts into the in-memory map before the SQLite insert succeeds. If the insert fails, the request returns an error but leaves a stale queued task in memory.

Fix: move `mediaTasks.set()` after `insertMediaTask()` succeeds, or remove the map entry in a catch.

lefarcen (commented):
Hi @pmasadali20776-ui! Thanks for tackling the media task persistence issue.

I've reviewed the implementation with our coding agent. The core approach (SQLite persistence + boot reconciliation) is sound and addresses the data loss at restart. The tests cover key scenarios.

I found 2 non-blocking concerns around terminal task TTL cleanup after restart — see inline comments below. These won't break the PR's primary goal (recover interrupted tasks), but they affect resource cleanup over time.

Once you've had a chance to look at these, I'll bring this to the maintainers for final review.

mrcfps (dismissed):
@pmasadali20776-ui I reviewed the SQLite-backed media task persistence changes, including the boot-time reconciliation path, the recovered `/api/media/tasks/:id/wait` flow, and the new persistence/route coverage added here. The interrupted-task recovery behavior and CLI handling line up with the restart-loss bug this PR is fixing, and the overall structure reads cleanly.

Thanks for pushing this through — this should make media generation much less fragile after daemon restarts. ✨


apps/daemon/src/server.ts
```diff
@@ -1813,6 +1872,19 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
+    terminalTtlMs: TASK_TTL_AFTER_DONE_MS,
+  });
+  if (mediaReconcile.interrupted > 0 || mediaReconcile.deleted > 0) {
+    console.warn(
+      `[media] reconcileMediaTasksOnBoot interrupted ${mediaReconcile.interrupted} task(s), ` +
+        `deleted ${mediaReconcile.deleted} expired terminal task(s)`,
+    );
+  }
+  mediaTasks.clear();
+  for (const row of listRecentMediaTasks(db, { terminalTtlMs: TASK_TTL_AFTER_DONE_MS })) {
```
mrcfps, line 1906: Hydrating rows here repopulates terminal tasks, but it never restores their TTL cleanup deadline. `notifyTaskWaiters()` is the only path that schedules `deleteMediaTask()`, and boot-recovered terminal rows never pass through it, so a task that finished shortly before shutdown can stay in memory, keep showing up in `includeDone=1`, and keep returning `200` from `/api/media/tasks/:id/wait` indefinitely until the next daemon restart. That breaks the 10-minute retention contract this PR is introducing. Please schedule GC when hydrating a terminal row (for example from `endedAt ?? updatedAt`) and have `getLiveMediaTask()`/the list path drop rows whose TTL has already expired so recovered terminal tasks age out the same way as tasks completed in the current process.

mrcfps (commented):
@pmasadali20776-ui Thanks for pushing the restart recovery work. I rechecked the persistence, boot reconciliation, and recovered `/api/media/tasks/:id/wait` path on the current head, and the main recovery flow looks good overall. I left one non-blocking follow-up on terminal-task TTL cleanup after boot so recovered rows keep the same retention behavior as tasks completed in the current process.


apps/daemon/src/server.ts
```diff
@@ -1813,6 +1893,19 @@ export async function startServer({ port = 7456, host = process.env.OD_BIND_HOST
...
```
mrcfps, line 1906: Hydrating rows here still brings terminal tasks back into memory without restoring their retention deadline. `notifyTaskWaiters()` is the only place that schedules `deleteMediaTask()`, and recovered `done`/`failed`/`interrupted` rows never pass through it, so a task that finished shortly before shutdown can keep showing up in `includeDone=1` and keep returning `200` from `/api/media/tasks/:id/wait` indefinitely until the next restart. That breaks the 10-minute TTL this PR adds. Please schedule GC when hydrating a terminal row from `endedAt ?? updatedAt`, and have `getLiveMediaTask()`/the list path drop rows whose TTL has already expired so recovered terminal tasks age out the same way as tasks completed in the current process.

mrcfps (commented):
@pmasadali20776-ui Thanks for the follow-up here. I rechecked the restart recovery path on this head; the persistence, interruption reconciliation, and recovered `/api/media/tasks/:id/wait` flow look solid overall. I left one non-blocking follow-up on terminal-task TTL cleanup after boot so recovered rows keep the same retention behavior as tasks completed in the current process.


leilei926524-tech:
Hi @pmasadali20776-ui — welcome to Open Design contributors. 🎉

Your first merged PR (#884) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #886, opened 2026-05-08
test(web): de-flake chat-scroll-preservation tab-switch tests

lefarcen:
## Summary

The `chat-scroll-preservation.test.tsx` suite has been failing intermittently on CI (most recently on PRs #666 and #714, but the test file lives on main and goes red on main runs too — most main runs go through but 1–3 of the 4 cases fail nondeterministically). The product code in `ChatPane.tsx` is not at fault; the test fixture has a timing race against React 18 + jsdom rAF.

## Root cause

The earlier shape installed instance-level `Object.defineProperty(remounted, 'scrollTop', { get/set })` mocks on the *remounted* `.chat-log` only **after** `await switchTab('Chat')`. Inside that `act()` the component's tab-restore `useEffect` schedules a `requestAnimationFrame` that writes `scrollTop` on the new element:

```ts
// ChatPane.tsx
useEffect(() => {
  if (tab !== 'chat') return;
  const el = logRef.current;
  if (!el) return;
  const saved = savedChatScrollRef.current;
  if (saved !== null) {
    requestAnimationFrame(() => {
      const target = logRef.current;
      if (saved.pinnedToBottom) target.scrollTop = target.scrollHeight;
      else target.scrollTop = saved.scrollTop;
      // ...
    });
  }
  // ...
}, [tab]);
```

Depending on whether jsdom's rAF polyfill flushed before the `await switchTab` resolved, the write either:

- landed on the still-default prototype setter (jsdom stores it internally but the test's instance-level mock getter — installed on the next line — has its own closure-captured initial `0` and cannot see it), or
- landed on the not-yet-installed instance setter (no setter at all, also lost).

Either way the assertion `expect(remountedTop).toBe(1500)` saw `0` and failed. The race is timing-dependent so the test went green or red across CI runs without any product-code change.

## Fix

Patch `scrollTop`/`scrollHeight`/`clientHeight` at `HTMLElement.prototype` level in `beforeEach`, routed through a single per-test `geom` object. Any `.chat-log` React mounts later automatically reads/writes through this storage. The component's restore rAF can now fire at any point and still write to the same place the assertion reads from. Restored prototype descriptors in `afterEach`.

The test surface stayed the same — same four cases, same expected behavior. The internal helper changed from `mockScrollGeometry(el)` (instance-level) to a global `setGeom`/`setUserScroll` API, dropping ~60 lines of per-test `Object.defineProperty` boilerplate.

## Validation

- `pnpm --filter @open-design/web test -- chat-scroll-preservation` × 8 consecutive runs: 488/488 passed each time, no flakes.
- `pnpm --filter @open-design/web typecheck`: clean.

## Why this fix and not a retry

`vitest`'s `retry: N` would mask the race instead of fixing it, and would still cost a real run on every flake. The race is not a product bug — `ChatPane.tsx` behaves correctly in real browsers — so the right place to fix is the test fixture's interaction with jsdom.

### DIFF
--- a/apps/web/tests/components/chat-scroll-preservation.test.tsx
+++ b/apps/web/tests/components/chat-scroll-preservation.test.tsx
@@ -1,49 +1,87 @@
 // @vitest-environment jsdom
 
 import { act, cleanup, fireEvent, render, screen } from '@testing-library/react';
-import { afterEach, describe, expect, it } from 'vitest';
+import { afterEach, beforeEach, describe, expect, it } from 'vitest';
 import { ChatPane } from '../../src/components/ChatPane';
 import type { ChatMessage } from '../../src/types';
 
-// jsdom does not run a layout engine, so scrollHeight, clientHeight, and
-// scrollTop are zero by default. The scroll-preservation effect derives
-// "near bottom" from those, so we drive them explicitly per test.
-function mockScrollGeometry(
-  el: HTMLElement,
-  geom: { scrollHeight: number; clientHeight: number; scrollTop: number },
-) {
-  Object.defineProperty(el, 'scrollHeight', {
-    configurable: true,
-    get: () => geom.scrollHeight,
-  });
-  Object.defineProperty(el, 'clientHeight', {
-    configurable: true,
-    get: () => geom.clientHeight,
-  });
-  let scrollTop = geom.scrollTop;
-  Object.defineProperty(el, 'scrollTop', {
-    configurable: true,
-    get: () => scrollTop,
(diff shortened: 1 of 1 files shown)


PerishCode (approved):

mrcfps (approved):
@lefarcen I reviewed the updated chat-scroll preservation test changes and the prototype-level geometry shim matches the ChatPane restore flow cleanly. I didn't find any correctness, safety, or maintainability issues in the changed ranges—thanks for tightening up a flaky test without touching the product behavior.

PR #894, opened 2026-05-08
feat: add design file rename support

### LINKED ISSUE #855: paste的文本忘记取名了，希望能增加修改文本文件名的功能。
Image

CIoudherd:
## Why

Fixes #855.

Design Files allowed pasted files to be saved without a custom filename, but there was no way to rename those files later.

## What changed

- Added a project file rename contract.
- Added a daemon rename API with safe path validation and conflict handling.
- Added Design Files row-menu rename UI.
- Kept the file list and open workspace tabs in sync after rename.
- Covered conflict, invalid path, missing file, artifact metadata, folder import, and web tab sync behavior.

## Validation

- pnpm guard
- pnpm typecheck
- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/project-file-rename.test.ts
- pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/FileWorkspace.test.tsx
- Manual UI check with `pnpm tools-dev run web --daemon-port 17456 --web-port 17573`
- Container verification with install, targeted tests, guard, typecheck, daemon build, and local web/daemon smoke check


apps/daemon/src/projects.ts
```diff
@@ -426,6 +426,116 @@ export async function deleteProjectFile(projectsRoot, projectId, name, metadata?
...
+        size: sourceStat.size,
+        mtime: sourceStat.mtimeMs,
+        kind: kindFor(oldName),
+        mime: mimeFor(oldName),
+        artifactKind: manifest?.kind,
+        artifactManifest: manifest,
+      },
+      oldName,
+      newName: oldName,
+    };
```
mrcfps: On the default macOS/Windows case-insensitive filesystem, `resolveSafeReal(dir, oldName)` and `resolveSafeReal(dir, newName)` both resolve `foo.txt` -> `Foo.txt` to the same real path, so this branch reports success and returns `newName: oldName` without ever changing the filename. That means users still cannot fix filename casing from the new rename UI on the platforms most contributors use.

A concrete fix here is to treat `oldName !== newName` plus `source === target` as a real case-only rename path instead of a no-op—for example by renaming through a temporary sibling name (and covering it with a targeted daemon test such as `paste-1.txt` -> `Paste-1.txt`).

mrcfps (commented):
@CIoudherd Thanks for wiring file rename through the contracts, daemon, and Design Files UI. I found one follow-up around case-only renames on case-insensitive filesystems.


CIoudherd:
That's right, I have fixed this bug by treating oldName !== newName with the same resolved realpath as a real case-only rename. The daemon now renames through a temporary sibling path and includes a targeted paste-1.txt -> Paste-1.txt test.
Thank you for your suggestion!

apps/web/src/components/FileWorkspace.tsx
```diff
@@ -335,6 +336,28 @@ export function FileWorkspace({
     }
   }
 
+  async function handleRename(oldName: string, nextName: string): Promise<ProjectFile | null> {
+    const result = await renameProjectFile(projectId, oldName, nextName);
+    const renamed = result.file;
+    await onRefreshFiles();
+
+    const nextTabs = persistedTabs.map((name) => (name === oldName ? renamed.name : name));
```
mrcfps, line 422: If the user has a pending sketch tab open and renames a persisted file to that same sketch filename, this replacement turns the persisted tab list into  without checking for an existing non-persisted tab of the same name.  later de-duplicates names with  (), so the pending sketch disappears from the tab strip even though its unsaved state is still sitting in . From the user's perspective that looks like lost work.

A concrete fix here is to guard  against collisions with pending sketch names before calling —for example, reject the rename client-side when , or re-key the pending sketch state so both tabs remain addressable.

mrcfps (commented):
@CIoudherd Thanks for following up on the case-only rename fix and the added coverage. I found one merge-safe web follow-up around tab-state collisions when a rename lands on an unsaved sketch name.


lefarcen:
Hi @CIoudherd! 🎉

Thanks for the contribution — this adds a much-needed rename UI for Design Files with solid validation.

— open-design team

apps/web/src/components/FileWorkspace.tsx
```diff
@@ -335,6 +336,28 @@ export function FileWorkspace({
...
+    const nextActive = tabsState.active === oldName ? renamed.name : tabsState.active;
+    onTabsStateChange({ tabs: nextTabs, active: nextActive });
+    if (activeTab === oldName) setActiveTab(renamed.name);
+
+    setSketches((curr) => {
```
lefarcen, line 427: **P2 — Unsaved sketch clobber**

If a saved sketch is renamed to the same name as an unsaved pending sketch tab, this state move clobbers the in-memory pending sketch entry with the persisted one. The result is silent loss of unsaved work and a tab/file mismatch.

The rename path needs an explicit collision check against `sketches[newName]` before updating tabs/state, and this needs a regression test.

apps/daemon/src/projects.ts
```diff
@@ -426,6 +426,158 @@ export async function deleteProjectFile(projectsRoot, projectId, name, metadata?
...
+  }
+
+  const target = await resolveSafeReal(dir, newName);
+  const targetPath = source === target ? resolveSafe(dir, newName) : target;
+
+  if (source !== target) {
```
lefarcen, line 465: **P2 — TOCTOU overwrite window**

This is still a TOCTOU overwrite window. You `stat(target)` and then later do `rename(temp, target)` via `renameFilePath()`. On POSIX, if another process creates `target` after the existence check but before the second rename, the final rename will replace that file instead of returning `EEXIST`, so the API's no-overwrite contract is violated.

The implementation needs an atomic no-clobber strategy for the second step of the case-only rename flow.

apps/daemon/tests/project-file-rename.test.ts
```diff
@@ -0,0 +1,167 @@
...
+    await writeText(projectId, 'b.txt', 'second');
+
+    const resp = await renameFile(projectId, 'a.txt', 'b.txt');
+    expect(resp.status).toBe(409);
+
+    const existing = await fetch(`${baseUrl}/api/projects/${projectId}/raw/b.txt`);
+    expect(await existing.text()).toBe('second');
+  });
+
+  it('rejects invalid and escaping paths', async () => {
```
lefarcen, line 134: **P3 — Missing symlink-escape test coverage**

The new API tests cover `..`, absolute paths, and reserved segments, but they do not cover the symlink-escape case that `resolveSafeReal()` is specifically defending.

Please add a regression that attempts a rename through a symlinked subdirectory (both source and target variants if possible) so future refactors do not silently reopen traversal.

lefarcen (changes requested):
Hey @CIoudherd, thanks for following up on mrcfps's initial findings! I ran a deeper review and found two remaining issues around data integrity and TOCTOU races:

**Overall take:**
The rename flow is directionally sound and mrcfps's case-insensitive feedback is properly addressed. Two remaining issues can cause data loss or silent overwrites under concurrent file operations or sketch state collisions—both need fixing before merge.

**Findings:**
1. **P2 — Unsaved sketch clobber risk** (FileWorkspace.tsx:349): see inline
2. **P2 — TOCTOU overwrite window** (projects.ts:462): see inline
3. **P3 — Missing symlink-escape test coverage**: see inline

These are all Lens A (code correctness) findings. No docs/specs changes in the rename-specific commits, so Lens B has no comments.

CIoudherd:
Thanks! I've fixed bug and add new test in these new commits.

## Detail

## P2 — Unsaved sketch clobber risk
Rename is now rejected client-side if the target name collides with an open non-persisted sketch tab, before calling the daemon. Added a `FileWorkspace` regression test to verify the API is not called and the pending sketch tab remains visible.

## P2 — TOCTOU overwrite window
The daemon now commits rename targets with an atomic no-clobber step using `link(temp, target)`, so a concurrently-created target returns `EEXIST` instead of being overwritten. Added a regression test for this race.

## P3 — Symlink-escape coverage
Added source and target symlink-escape tests for the rename API, covering symlinked project paths that resolve outside the project root.

## Additional Self-Found
Source filenames are now validated without being sanitized, so folder-import files like `my note.txt` can be renamed correctly. Added a folder-import regression test for source names with spaces.

## Validation
Ran Docker-isolated validation: install, builds, daemon/web targeted tests, package typechecks, `pnpm guard`, root `pnpm typecheck`, daemon build, and web/daemon smoke check. All passed.

lefarcen (dismissed):
Hey @CIoudherd, nice work on the fixes! All three P2 issues are properly addressed:

**Verified fixes:**
✅ P2 unsaved sketch clobber — UI now rejects renames that collide with pending sketch tabs before calling the daemon (FileWorkspace.tsx:339-347 + test coverage)
✅ P2 TOCTOU overwrite — daemon uses atomic `link(temp, target)` for no-clobber publish, returns EEXIST on concurrent creation instead of overwriting (projects.ts:478-524 + race test)
✅ P3 symlink-escape coverage — added both source and target symlink regression tests (project-file-rename.test.ts:203-230)

**Bonus fix:**
✅ Folder-import source validation now preserves spaces correctly (validateProjectPath instead of sanitizePath)

All code changes are real and tests execute the new paths. Clean pass from my side; deferring final approval to a maintainer.

mrcfps:
@CIoudherd  Thanks — the code looks good from my side now. The remaining blocker is that the branch is currently conflicting with `main`, so please rebase or otherwise resolve the merge conflicts first. Once that’s done, I’ll merge it right away.

CIoudherd:
@mrcfps Thanks, I rebased the branch onto the latest `main` and force-pushed the updated PR branch with `--force-with-lease`.

I also re-ran the focused validation after the rebase:
- `pnpm guard`
- daemon rename regression test: `tests/project-file-rename.test.ts`
- web rename UI tests: `FileWorkspace.test.tsx` and `DesignFilesPanel.test.tsx`

All passed. I also ran the full Docker validation earlier before the final rebase, including workspace tests, guard, typecheck, daemon build, and a web/daemon startup smoke check.

lefarcen (commented):
Rebase onto main looks clean — the P2 fixes from prior review (unsaved sketch clobber guard + TOCTOU atomic rename) are intact at head a4079e8f, and author confirmed tests pass.

Ready from code-review perspective; deferring final approval to maintainer.

mrcfps (approved):

lefarcen:
Thread resolution needed: 2 review threads (both from earlier mrcfps findings) are still unresolved, which is blocking the merge. Since mrcfps's latest APPROVED review (just now at 13:12) said the code looks good after the rebase, those findings appear addressed — just need the threads manually resolved to clear the gate.

leilei926524-tech:
Hi @CIoudherd — welcome to Open Design contributors. 🎉

Your first merged PR (#894) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #900, opened 2026-05-08
fix(web): unbreak Create button on plain HTTP / LAN-IP deployments (#849)

### LINKED ISSUE #849: Create button silently does nothing when served over HTTP on a LAN IP
## Bug

Clicking **+ Create** on the new project panel does nothing — no navigation, no error, no network request. The button appears enabled and clickable.

## Root cause

`crypto.randomUUID()` is only available in [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS or `localhost`). When Open Design is served over plain HTTP on a LAN IP (e.g. `http://192.168.1.10:17573`, which is the standard Docker self-hosted setup), Chrome/Chromium silently makes `crypto.randomUUID` unavailable.

In `apps/web/src/state/projects.ts`, `createProject()` calls `crypto.randomUUID()` to generate the project ID before the `fetch`. This throws `TypeError: crypto.randomUUID is not a function`, which is caught by the surrounding `try/catch` and returned as `null`. `handleCreateProject` in `App.tsx` sees `!result` and exits silently — no fetch is made, no navigation happens.

The same issue affects `apps/web/src/components/ProjectView.tsx` (3 further calls), which would break agent interactions once a project is reached.

## Steps to reproduce

1. Run Open Design via Docker Compose with the default config
2. Access the UI at `http://<LAN-IP>:<port>` (not `localhost`)
3. Type a project name and click **+ Create**
4. Nothing happens — no navigation, no toast, no console error

## Expected behaviour

Project is created and UI navigates to the project view.

## Fix

Add a polyfill that falls back to a `Math.random`-based v4 UUID when `crypto.randomUUID` is unavailable:

**`apps/web/src/utils/uuid.ts`** (new file):
```ts
// crypto.randomUUID is only available in secure contexts (HTTPS/localhost).
// Fall back to a Math.random-based v4 UUID for plain-HTTP LAN deployments.
export function randomUUID(): string {
  if (typeof crypto !== 'undefined' && typeof crypto.randomUUID === 'function') {
    return crypto.randomUUID();
  }
  return 'xxxxxxxx-xxxx-4xxx-yxxx-xxxxxxxxxxxx'.replace(/[xy]/g, (c) => {
    const r = (Math.random() * 16) | 0;
    return (c === 'x' ? r : (r & 0x3) | 0x8).toString(16);
  });
}

Sid-Qin:
## Summary

- `crypto.randomUUID()` is restricted to [secure contexts](https://developer.mozilla.org/en-US/docs/Web/Security/Secure_Contexts) (HTTPS or `localhost`). When Open Design is served over plain HTTP on a LAN IP — the standard Docker / unRAID / NAS self-hosted setup, e.g. `http://192.168.1.10:17573` — Chromium silently makes the function undefined. Calls throw `TypeError: crypto.randomUUID is not a function`, which the `try/catch` around `createProject()` swallows by returning `null`, which the click handler reads as "no project, do nothing".
- Net effect: the **Create button is a silent no-op for every LAN-IP user**. Same root cause shows up across [#849](https://github.com/nexu-io/open-design/issues/849) (this PR) and [#394](https://github.com/nexu-io/open-design/issues/394).
- Fix: centralize UUID generation into a new `apps/web/src/utils/uuid.ts` helper with a 3-tier fallback (per @lefarcen's [review suggestion](https://github.com/nexu-io/open-design/issues/849#issuecomment-4404371657)), and route all four affected callsites through it.

Fixes #849. Same root cause as #394 — landing this should close that one too.

Big credit to @socratestherefore for the gold-standard root-cause writeup and proposed polyfill, and @lefarcen for the `getRandomValues` upgrade path that keeps the non-secure tier crypto-random.

## Three-tier fallback

| Tier | Path | When it fires |
|---|---|---|
| 1 | `crypto.randomUUID()` | Secure context (HTTPS or `localhost`). Native, fast, crypto-random. |
| 2 | `crypto.getRandomValues()` + RFC 4122 §4.4 byte layout | Non-secure context (the issue #849 LAN-IP case). The Web Crypto API itself isn't gated by `isSecureContext`, so `getRandomValues` is still there — gives us a real v4 UUID with crypto-quality entropy. |
| 3 | `Math.random()` | Environments missing both. Project / message / client-request ids are scoped to a single user's browser session — collision avoidance is enough; cryptographic uniqueness isn't required. |

## Diff shape

| File | What changes |
|---|---|
| `apps/web/src/utils/uuid.ts` | New helper. ~60 lines, standalone — no other imports beyond `globalThis.crypto`. |
| `apps/web/src/state/projects.ts` | Import the helper; replace `crypto.randomUUID()` at line 48 with `randomUUID()`. |
| `apps/web/src/components/ProjectView.tsx` | Import the helper; replace 3 `crypto.randomUUID()` callsites (user msg id, assistant msg id, client request id). |
| `apps/web/tests/utils/uuid.test.ts` | New: 6 tests covering tier 1 / tier 2 / tier 3 / format / "doesn't throw on missing randomUUID" / 1000-iteration uniqueness on tier 2. |

4 files, +184 / −4.

## Why a single helper instead of just polyfilling each callsite

Centralizing means:
- One place to evolve the fallback strategy if a new tier becomes preferred.
- Future callsites can `import { randomUUID } from '../utils/uuid'` and not re-litigate the secure-context question.
- Sweep test: `grep -rn "crypto.randomUUID" apps/web/src/` returns zero hits in app code after this PR (only the helper file's own implementation), which is the regression guard for "did anyone reintroduce a direct call".

## What's verified

| Check | Result |
|---|---|
| `pnpm --filter @open-design/web vitest` | **522/522** (was 516, +6) |
| `pnpm --filter @open-design/web typecheck` (`tsc -b --noEmit`) | clean |
| `grep -rn "crypto.randomUUID" apps/web/src/` | only the helper's own implementation; zero direct callsites in app code |

## What's NOT verified

I don't have a Docker / LAN-IP self-hosted setup locally, so I haven't reproduced the original symptom end-to-end. The structural argument:

1. The 6 new unit tests stub `globalThis.crypto = { randomUUID: undefined, getRandomValues }` — the exact shape Chromium presents in non-secure contexts — and assert the helper still returns a valid v4 UUID without throwing. That's the literal #849 root cause being pinned.
2. Format validation uses an RFC 4122 v4 regex (`8-4-4-4-12` hex with version `4` and variant `10xx`) plus explicit version/variant nibble checks, so a future refactor that accidentally returns a v1/v3/v5 UUID under tier 2 fails with a pointed error.
3. 1000-iteration uniqueness check on tier 2 catches any entropy regression.

If @socratestherefore can pull this branch onto their unRAID Docker setup and confirm Create works over `http://<LAN-IP>`, that closes the loop end-to-end.

## Test plan

- [x] 6 new helper unit tests pass (tier 1 / tier 2 / tier 3 / format / no-throw / uniqueness).
- [x] Existing web test suite (522 tests) stays green.
- [x] Web typecheck + i18n-check clean.
- [x] No residual direct `crypto.randomUUID()` callsite in `apps/web/src/`.
- [ ] Manual smoke on plain-HTTP LAN-IP (e.g. Docker self-hosted): Create button creates a project and navigates, no silent failure.
- [ ] Manual smoke on localhost / HTTPS: no regression on the secure-context path (tier 1 still wins).

## Notes

- Cross-link: #394 reports the same root cause and was confirmed in that thread by multiple users. This fix should close it.
- Considered the alternative @socratestherefore mentioned (server-side UUID generation by making the daemon's POST id field optional). Stuck with the client-side polyfill because it's smaller scope, doesn't change the wire contract, and doesn't require backward-compat handling for clients that still send their own ids. Trade-off noted at the issue level — happy to iterate if a maintainer prefers the server-side approach.

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -38,6 +38,7 @@ import {
   apiProtocolModelLabel,
 } from '../utils/apiProtocol';
 import { playSound, showCompletionNotification } from '../utils/notifications';
+import { randomUUID } from '../utils/uuid';
 import { DEFAULT_NOTIFICATIONS } from '../state/config';
 import type { TodoItem } from '../runtime/todos';
 import { appendErrorStatusEvent } from '../runtime/chat-events';
@@ -983,7 +984,7 @@ export function ProjectView({
       setError(null);
       const startedAt = Date.now();
       const userMsg: ChatMessage = {
-        id: crypto.randomUUID(),
+        id: randomUUID(),
         role: 'user',
         content: prompt,
         createdAt: startedAt,
@@ -1010,7 +1011,7 @@ export function ProjectView({
               selectedAgentChoice?.model,
             )
           : apiProtocolModelLabel(config.apiProtocol, config.model);
-      const assistantId = crypto.randomUUID();
+      const assistantId = randomUUID();
       const assistantMsg: ChatMessage = {
         id: assistantId,
         role: 'assistant',
@@ -1260,7 +1261,7 @@ export function ProjectView({
           projectId: project.id,
           conversationId: activeConversationId,
           assistantMessageId: assistantId,
-          clientRequestId: crypto.randomUUID(),
+          clientRequestId: randomUUID(),
           skillId: project.skillId ?? null,
           designSystemId: project.designSystemId ?? null,
           attachments: attachments.map((a) => a.path),
--- a/apps/web/src/state/projects.ts
+++ b/apps/web/src/state/projects.ts
@@ -6,6 +6,7 @@
 // the UI can stay rendered when the daemon is briefly unreachable.
 
 import type { ImportFolderRequest, ImportFolderResponse } from '@open-design/contracts';
+import { randomUUID } from '../utils/uuid';
 import type {
   ChatMessage,
   Conversation,
@@ -45,7 +46,13 @@ export async function createProject(input: {
   metadata?: ProjectMetadata;
 }): Promise<{ project: Project; conversationId: string } | null> {
   try {
-    const id = crypto.randomUUID();
+    // `randomUUID` falls back to `crypto.getRandomValues` / `Math.random`
+    // when `crypto.randomUUID` is unavailable. Open Design served over
+    // plain HTTP on a LAN IP (Docker / unRAID self-hosting) is a
+    // non-secure context, where `crypto.randomUUID` is undefined and
+    // calling it directly throws — the surrounding try/catch then turns
+    // the Create button into a silent no-op (issue #849).
+    const id = randomUUID();
     const resp = await fetch('/api/projects', {
       method: 'POST',
       headers: { 'Content-Type': 'application/json' },
--- a/apps/web/src/utils/uuid.ts
+++ b/apps/web/src/utils/uuid.ts
@@ -1,0 +1,56 @@
+// Tiered v4-UUID generator that survives non-secure contexts.
+//
+// `crypto.randomUUID()` is restricted to secure contexts — HTTPS or
+// `localhost`. When Open Design is served over plain HTTP on a LAN
+// IP (the standard Docker / unRAID / NAS self-hosted setup, e.g.
+// `http://192.168.1.10:17573`), Chromium silently makes
+// `crypto.randomUUID` undefined. Calls then throw
+// `TypeError: crypto.randomUUID is not a function`, which the surrounding
+// try/catch in `state/projects.ts` swallows — the Create button
+// effectively becomes a no-op for every LAN-IP user (issue #849, also
+// reported as #394).
+//
+// Three-tier fallback, preferred in order:
+//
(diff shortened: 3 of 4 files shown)


mrcfps (approved):
@Sid-Qin I reviewed the changed ranges in the web UUID fallback and the updated call sites. Centralizing the non-secure-context fallback in `apps/web/src/utils/uuid.ts` and covering the `randomUUID`-missing path in `apps/web/tests/utils/uuid.test.ts` addresses the LAN-IP/plain-HTTP regression without changing the existing API contracts. Thanks for tracking this down and for adding focused regression coverage. 🙏


joeylee12629-star:
Hey @Sid-Qin  — Joey from Open Design core team. 👋 Just sent you an email about a community Maintainer invite — if you missed it, our Discord is the easiest way to chat:

👉 https://discord.com/invite/qhbcCH8Am4

Once you're in, ping me with your Discord username — here, or by email at lijinwei@open-design.ai — and we'll find you.

Cheers,
Joey - Open Design team

PR #906, opened 2026-05-08
fix(packaged): swallow harmless setTypeOfService EINVAL from undici (#895)

Sid-Qin:
## Summary

- Opening Settings → Pets → Community in the packaged desktop app surfaces a native **"JavaScript error in main process"** dialog with `Uncaught Exception: Error: setTypeOfService EINVAL`. Root cause: undici's socket setup tries to set the IP_TOS byte (QoS / DSCP marking) on outbound sockets, and the macOS kernel refuses with `EINVAL` on some configurations (VPNs, IPv6-only sockets, certain firewall postures). The byte is purely advisory — the socket itself is healthy — so the rejection should not crash the app.
- The flow that triggers it isn't actually Pets-specific. Every renderer page load and API call in the packaged build flows through the `od://` Electron protocol handler in [`apps/packaged/src/protocol.ts`](https://github.com/nexu-io/open-design/blob/main/apps/packaged/src/protocol.ts), which forwards via Node's global `fetch` (undici under the hood). Pets > Community happens to be the page that surfaces it for #895's reporter, but anything that fetches could trip it.
- Two cooperating defensive layers:
  - **Layer 1 — `protocol.ts`**: extract the inner handler as `handleOdRequest()` (testable) and wrap the `await fetch()` in a try/catch that returns a 502 Response on failure. Without this, every undici rejection — not just `setTypeOfService` — escapes to Electron's default uncaught-exception path. Now the renderer sees a normal error response and the main process keeps running.
  - **Layer 2 — `logging.ts`**: add a `process.on('uncaughtException')` handler with a narrow `isHarmlessSocketOptionError()` filter that matches the canonical undici shape only (message contains `setTypeOfService` AND code is `EINVAL` or message contains `EINVAL`). Anything unrecognised re-throws via `setImmediate` so Node's default crash + Electron's crash dialog still fire end-to-end for real bugs.

Fixes #895.

## Why two layers

| Layer | Catches | Why we need it |
|---|---|---|
| `protocol.ts` try/catch | undici rejections that propagate as a rejected `fetch()` promise | The common case — once the socket setup throws, the promise rejects, `await` re-throws, our catch converts to a 502. Already enough to fix the reported scenario. |
| `logging.ts` uncaughtException filter | undici/socket errors that escape promise rejection (e.g. errors fired from socket-level callbacks that don't have a Promise to wire into) | Defence in depth. Without it, any future undici code path that emits the error out-of-band would still crash the main process. The filter is intentionally narrow — see negative tests below. |

## Negative guard against over-broad swallowing

The filter at `logging.ts:isHarmlessSocketOptionError` matches **only** the canonical undici shape. Tests explicitly assert it does NOT match:

- `write EINVAL` (a generic EINVAL — also raised by real bugs)
- `setTypeOfService EACCES` (a setTypeOfService failure with a different errno — could indicate a real permissions issue worth investigating)
- Non-`Error` rejection values (preserves fail-fast for primitive rejections)
- An `Error` with empty message even when `code === 'EINVAL'`

A future regression that broadens the filter to "every EINVAL" would trip these tests.

## Diff shape

| File | What changes |
|---|---|
| `apps/packaged/src/protocol.ts` | Pull inner handler out as `handleOdRequest()`, accept an optional `fetchImpl` for tests, wrap fetch in try/catch, build a 502 Response with structured error body on failure. |
| `apps/packaged/src/logging.ts` | Add `isHarmlessSocketOptionError()` (exported for tests) and a `process.on('uncaughtException')` handler that uses it. Existing `uncaughtExceptionMonitor` log call kept intact. `unhandledRejection` also gains the filter for symmetry. |
| `apps/packaged/tests/protocol.test.ts` | New: 5 tests — happy path proxy, URL rewriting preserves path/search, **fetch rejecting with the canonical EINVAL shape returns a 502 (the #895 regression guard)**, never-throw guarantee, non-Error rejection handling. |
| `apps/packaged/tests/logging.test.ts` | New: 8 tests — canonical match, message-only / code-only variants, **negative guard against generic EINVAL**, negative guard against setTypeOfService with a non-EINVAL errno, primitive rejections, empty-message Errors. |

4 files, +320 / −2.

## What's verified

| Check | Result |
|---|---|
| `pnpm --filter @open-design/packaged vitest tests/protocol.test.ts tests/logging.test.ts` | **13/13** |
| packaged `tsc -p tsconfig.json --noEmit` | clean |
| packaged `tsc -p tsconfig.tests.json --noEmit` (the CI killer) | clean |

There is one pre-existing failure in `tests/sidecars.test.ts` (`adds custom VP_HOME/bin to the packaged PATH builder`) that is unrelated to this PR — confirmed by stashing the change and re-running on the bare branch base. Not in scope to fix here.

## What's NOT verified

I don't have a macOS / VPN configuration locally that reproduces the original `setTypeOfService EINVAL`. The structural argument:

1. The handler-side test stubs the exact undici error shape (`Error('setTypeOfService EINVAL')` with `code: 'EINVAL'`, `syscall: 'setTypeOfService'`) and asserts the protocol handler returns a 502 Response instead of throwing.
2. The filter-side tests pin both the positive match (canonical shape) and the negative guards (generic EINVAL, different syscall, non-Error rejections), so a future contributor can't accidentally regress either direction.
3. If the affected user (cc the #895 reporter) can pull this branch in a packaged build and confirm the dialog no longer appears, that closes the loop end-to-end.

## Test plan

- [x] 13 new tests pass (5 protocol + 8 logging).
- [x] Both packaged typecheck configs clean.
- [x] No new failures introduced (pre-existing `sidecars.test.ts` flake noted but out of scope).
- [ ] Manual smoke on packaged build with the affected configuration: open Settings → Pets → Community and observe no native error dialog. Renderer should see a normal "Pets community feed unavailable" error if the underlying network actually is broken — never a main-process crash.
- [ ] Manual smoke on a healthy macOS / Linux: no regression on the `od://` proxy happy path; the web app continues to load and API calls flow normally.

## Notes

- `od://` scheme handler is the central network-proxy hop in the packaged build. Hardening it has the side benefit of making *every* renderer fetch failure gracefully degrade — not just the `setTypeOfService EINVAL` case the issue reports. Future undici / socket / DNS errors that would have crashed main now surface as 502s the renderer can handle.
- Did not bump up to a global per-process Electron crash handler; the targeted filter is enough for the reported scenario and keeps fail-fast for unknown errors.

### DIFF
--- a/apps/packaged/src/logging.ts
+++ b/apps/packaged/src/logging.ts
@@ -38,27 +38,80 @@ function normalizeError(error: unknown): unknown {
  * still connects and serves traffic — so the right behaviour is to
  * log + ignore, not to crash.
  *
- * Match strategy is intentionally narrow: name the syscall (the
- * `setTypeOfService` literal), and verify the error code is the
- * expected `EINVAL`. We avoid swallowing every `EINVAL` because that
- * code is also raised by genuine bugs (bad config values, malformed
- * arguments to other syscalls). Exported so a unit test can pin the
- * exact shape this branch matches.
+ * Match strategy is intentionally narrow:
+ *   1. The error message must name the `setTypeOfService` syscall.
+ *   2. The structured `code` property is **authoritative** when
+ *      present: it must equal `EINVAL` for the error to qualify.
+ *      A `code` of anything else (e.g. `EACCES`) is treated as a
+ *      contradicting signal and the filter rejects the match —
+ *      otherwise an `EACCES` permission failure with a stale or
+ *      copy-pasted `EINVAL` substring in the message would slip
+ *      through.
+ *   3. Only when `code` is absent (some libuv builds don't populate
+ *      it on raw thrown Errors) do we fall back to looking for the
+ *      `EINVAL` token in the message.
+ *
+ * We never swallow every `EINVAL`: that code is raised by plenty of
+ * real bugs (bad config values, malformed arguments to other
+ * syscalls). Exported so a unit test can pin the exact shape this
+ * branch matches.
  */
 export function isHarmlessSocketOptionError(value: unknown): boolean {
   if (!(value instanceof Error)) return false;
   const message = typeof value.message === "string" ? value.message : "";
   if (!message) return false;
-  const code =
-    typeof (value as NodeJS.ErrnoException).code === "string"
-      ? (value as NodeJS.ErrnoException).code
-      : "";
-  // Primary shape: undici / Node socket initialiser. The error message
-  // string is constructed by libuv as `<syscall> <errcode>`.
-  if (message.includes("setTypeOfService") && (code === "EINVAL" || message.includes("EINVAL"))) {
-    return true;
+  if (!message.includes("setTypeOfService")) return false;
+  const code = (value as NodeJS.ErrnoException).code;
+  if (typeof code === "string" && code.length > 0) {
+    // Structured code present — it has to be EINVAL. Anything else
+    // (EACCES, EPERM, ECONNRESET, …) is a contradicting signal and
+    // we let it crash so real bugs don't get hidden.
+    return code === "EINVAL";
   }
-  return false;
+  // No structured code: fall back to message-based detection. The
+  // libuv error string is `<syscall> <errcode>` so the EINVAL token
+  // appears alongside the syscall name.
+  return message.includes("EINVAL");
+}
+
+/**
+ * Build the named `uncaughtException` handler used by
+ * `attachPackagedDesktopProcessLogging`. Exposed as its own factory
+ * so a unit test can drive it without bringing up the full logging
+ * pipeline.
+ *
+ * Behaviour contract (issue #906 review):
+ *   - Harmless `setTypeOfService EINVAL` shapes from undici socket
+ *     internals are logged at warn level and the function returns
+ *     silently. The process continues running.
+ *   - Anything else is logged at error level, then the handler
+ *     **removes itself from `process.uncaughtException` listeners**
+ *     and re-throws via `setImmediate`. Removing the listener first
+ *     is critical: without it, the re-throw re-enters the same
+ *     handler, schedules another setImmediate, and the packaged
+ *     main process spins forever instead of terminating —
+ *     reproduced and called out by mrcfps + lefarcen on the first
+ *     #906 review pass. With the listener gone the next throw has
+ *     no `uncaughtException` listener to land in, Node's default
+ *     crash path takes over, and Electron's native "JavaScript
+ *     error in main process" dialog renders as it did before this
+ *     code existed.
+ */
+export function createFatalUncaughtExceptionHandler(
+  logger: PackagedDesktopLogger,
+): (error: unknown) => void {
+  const handler = (error: unknown): void => {
+    if (isHarmlessSocketOptionError(error)) {
+      logger.warn("packaged desktop swallowed harmless socket option error", { error });
+      return;
+    }
+    logger.error("packaged desktop fatal uncaught exception", { error });
+    process.removeListener("uncaughtException", handler);
+    setImmediate(() => {
+      throw error;
+    });
+  };
+  return handler;
 }
 
 function normalizeMeta(meta: Record<string, unknown> | undefined): Record<string, unknown> | undefined {
@@ -168,20 +221,15 @@ export function attachPackagedDesktopProcessLogging(options: {
   // error in main process" dialog the next time anything in the
   // renderer does a fetch (e.g. opening Settings → Pets → Community).
   //
-  // For unknown errors we re-throw via `setImmediate` so Node's
-  // default uncaughtException behaviour (process death + crash dialog)
-  // is preserved end-to-end. Adding the handler at all does not
-  // suppress that path — it only short-circuits the specific harmless
-  // shapes we recognise.
-  process.on("uncaughtException", (error) => {
-    if (isHarmlessSocketOptionError(error)) {
-      logger.warn("packaged desktop swallowed harmless socket option error", { error });
-      return;
-    }
-    setImmediate(() => {
-      throw error;
-    });
-  });
+  // For non-harmless errors the handler restores Node's default
+  // uncaughtException behaviour: it removes itself from the listener
+  // list and re-throws via `setImmediate`. With no `uncaughtException`
+  // handlers registered, Node's default crash path takes over (stack
+  // trace + non-zero exit) and Electron's own "JavaScript error in
+  // main process" dialog renders as it would have before this code
+  // existed. See `createFatalUncaughtExceptionHandler` for the
+  // unit-tested factory.
+  process.on("uncaughtException", createFatalUncaughtExceptionHandler(logger));
   process.on("unhandledRejection", (reason) => {
     if (isHarmlessSocketOptionError(reason)) {
       logger.warn("packaged desktop swallowed harmless socket option rejection", { reason });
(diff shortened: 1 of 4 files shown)


apps/packaged/src/logging.ts
```diff
@@ -123,7 +158,35 @@ export function attachPackagedDesktopProcessLogging(options: {
...
+  // suppress that path — it only short-circuits the specific harmless
+  // shapes we recognise.
+  process.on("uncaughtException", (error) => {
+    if (isHarmlessSocketOptionError(error)) {
+      logger.warn("packaged desktop swallowed harmless socket option error", { error });
+      return;
+    }
+    setImmediate(() => {
+      throw error;
+    });
```
mrcfps: Re-throwing through `setImmediate` while this same `uncaughtException` listener remains installed never restores Node's default crash path. The next throw re-enters `uncaughtException`, this branch schedules another `setImmediate(() => { throw error; })`, and the packaged main process loops forever instead of terminating. I reproduced that behavior with a minimal Node script using this exact pattern: it repeatedly logged the same handler invocation until the process was killed.

That means any real uncaught exception that does **not** match `isHarmlessSocketOptionError()` will now spin the app and spam logs/CPU rather than fail fast, which is the opposite of the fallback behavior described in this file.

Please switch to a crash path that bypasses this listener after logging — for example, temporarily remove the handler before re-throwing, or set an exit code and terminate the process after flushing logs. I'd also add a regression test for the non-harmless branch so future refactors keep the intended crash semantics.

mrcfps (changes requested):
@Sid-Qin thanks for digging into the `setTypeOfService EINVAL` crash and adding focused regression coverage here. The protocol-side catch looks good, but I found one blocker in the new `uncaughtException` fallback that changes real crash behavior for non-harmless errors.

lefarcen:
Hi @Sid-Qin! 🎉

Thanks for the contribution — this is a thorough defense-in-depth fix for the setTypeOfService EINVAL crash in the packaged build. The two-layer approach (protocol.ts try/catch + logging.ts narrow filter) is solid, and the test coverage gives me confidence in the negative guards.

— open-design team

apps/packaged/src/logging.ts
```diff
@@ -123,7 +158,35 @@ export function attachPackagedDesktopProcessLogging(options: {
...
+  // Electron's default handler and surfaces as a native "JavaScript
+  // error in main process" dialog the next time anything in the
+  // renderer does a fetch (e.g. opening Settings → Pets → Community).
+  //
+  // For unknown errors we re-throw via `setImmediate` so Node's
+  // default uncaughtException behaviour (process death + crash dialog)
+  // is preserved end-to-end. Adding the handler at all does not
+  // suppress that path — it only short-circuits the specific harmless
+  // shapes we recognise.
+  process.on("uncaughtException", (error) => {
```
lefarcen: **P1** — The non-harmless path re-throws inside `setImmediate`, which re-enters this same still-registered `uncaughtException` handler → the real crash can spin/log forever instead of terminating or showing Electron's default dialog.

Fix: use a named handler and remove it before rethrowing, or explicitly terminate after logging. Add a regression test for a non-harmless uncaught exception to verify the process actually exits / shows the dialog.

apps/packaged/src/logging.ts
```diff
@@ -26,6 +26,41 @@ function normalizeError(error: unknown): unknown {
...
+  if (!(value instanceof Error)) return false;
+  const message = typeof value.message === "string" ? value.message : "";
+  if (!message) return false;
+  const code =
+    typeof (value as NodeJS.ErrnoException).code === "string"
+      ? (value as NodeJS.ErrnoException).code
+      : "";
+  // Primary shape: undici / Node socket initialiser. The error message
+  // string is constructed by libuv as `<syscall> <errcode>`.
+  if (message.includes("setTypeOfService") && (code === "EINVAL" || message.includes("EINVAL"))) {
```
lefarcen: **P2** — The filter is narrower than "every EINVAL", but it still trusts the message over a contradictory structured errno. If `code` is present and not `EINVAL`, a message containing both `setTypeOfService` and `EINVAL` is swallowed anyway — this weakens the claimed negative guard for a process-wide exception handler.

Fix: make `code` authoritative when present. Require `code === "EINVAL"`, and only fall back to message `EINVAL` when `code` is absent. Add a test for `code = "EACCES"` with a message containing `setTypeOfService EINVAL` to pin that it does NOT match.

lefarcen (dismissed):
Hey @Sid-Qin, thorough defensive work here! The two-layer approach + negative-guard tests show solid engineering discipline. Found one must-fix (the `setImmediate` re-throw path) and a few sharpening opportunities below.

**Code review (Lens A):**

The protocol.ts try/catch is clean — extracts the handler for testability, returns a structured 502, and the tests pin the #895 regression path. The logging.ts filter is intentionally narrow (syscall + errno), which is the right direction.

**Reasoning review (Lens B):**

The "Why two layers" argument is clear for the protocol layer, but the global `uncaughtException` handler is justified as defense-in-depth for "socket-level callbacks that don't have a Promise to wire into" without citing an observed undici/Electron path or upstream issue. For a high-scope packaged main-process change, that weakens the case.

See inline comments for specifics.

Sid-Qin:
@mrcfps @lefarcen both findings addressed in `018dc5c`. Thanks for catching them — the recursive rethrow loop especially: I had the right end goal ("preserve fail-fast") but the wrong mechanism, and your minimal Node repro made it obvious.

**P1 — recursive rethrow.** Extracted the handler as a named factory `createFatalUncaughtExceptionHandler(logger)` so the closure can capture itself by reference. Non-harmless branch now does `process.removeListener('uncaughtException', self)` **before** scheduling the rethrow. With no listener registered, the next throw lands in Node's default crash path — which is exactly the behaviour we wanted.

**P2 — code is authoritative.** Filter rewritten so `code` (when present) is checked first and must equal `EINVAL`; only when `code` is genuinely absent (some libuv builds don't populate it) does message-based detection take over. Per @lefarcen's example: `code: 'EACCES'` paired with `'setTypeOfService EINVAL'` in the message now slips through to the crash path instead of being swallowed.

**3 new tests** pin both fixes:
- `does NOT match when code contradicts the message` + `EPERM` variant — P2 regression guard.
- `removes itself from uncaughtException listeners before scheduling the rethrow` — uses `vi.spyOn(process, 'removeListener')` and a stubbed `setImmediate` to assert removeListener fires *before* the rethrow is scheduled (via `mock.invocationCallOrder`).
- `does NOT re-enter itself when invoked twice` — belt-and-suspenders loop guard. If a future refactor drops the removeListener call, this test trips.

Verified locally:
- packaged vitest: **18/18** (was 13, +5)
- packaged `tsc -p tsconfig.json --noEmit`: clean
- packaged `tsc -p tsconfig.tests.json --noEmit`: clean

@lefarcen on the Lens B "weakens the case" point — fair feedback. Layer 2 isn't justified by an observed undici path, just defence-in-depth for socket-callback errors that escape promise rejection (e.g. an idle pool ping firing from a setImmediate-style context). If you'd prefer Layer 2 dropped entirely and we rely on the protocol-handler try/catch alone, I'm happy to do that — Layer 1 covers the reported #895 reproducer cleanly. Let me know.

mrcfps (approved):
@Sid-Qin thanks for the fast follow-up here. I re-reviewed `apps/packaged/src/logging.ts` and `apps/packaged/src/protocol.ts` on `018dc5c`, and the previous crash-path concerns look addressed: the fatal handler now unregisters itself before rethrowing, the harmless-error filter treats `code` as authoritative, and the added regression coverage exercises both fixes. I also re-ran the packaged tests locally (`pnpm --filter @open-design/packaged test`) and they passed on this head. Appreciate the careful tightening on the second pass.

PR #907, opened 2026-05-08
[codex] Optimize Composio connector previews

alchemistklk:
## Summary

Optimizes the Composio connector flow so first connect, connector counts, OAuth pending state, and tool preview no longer block on full Composio tool discovery.

## Changes

- Persist and reuse Composio `auth_config_id` values in `.od/connectors/composio-config.json`.
- Avoid full Composio discovery on Connect; use static connector metadata plus scoped auth config lookup/create.
- Add auth-config latency logging and metrics.
- Make duplicate successful OAuth callbacks idempotent when the connected account is already stored.
- Split lightweight toolkit metadata discovery from full tool hydration.
- Add display-only `toolCount` plus paginated single-connector tool preview in the drawer.
- Add OAuth pending UI state with polling, persistence, and cancel affordance.

## Validation

- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/connectors-routes.test.ts`
- `pnpm --filter @open-design/web exec vitest run tests/providers/registry.test.ts tests/components/EntryView.test.ts`
- `git diff --check`


mrcfps:
@alchemistklk I'm holding off on generating review comments for nexu-io/open-design#907 because this pull request has merge conflicts right now.

apps/web/src/components/ConnectorsBrowser.tsx
```diff
@@ -525,6 +697,38 @@ export function ConnectorsBrowser({
...
+  useEffect(() => {
+    if (!detailConnector) return;
+    if (hasLoadedAllAdvertisedConnectorTools(detailConnector)) return;
+    if (toolPreviewAttemptedIds[detailConnector.id]) return;
+    if (toolPreviewLoadingIds[detailConnector.id]) return;
+    void hydrateToolPreview(detailConnector.id);
+  }, [detailConnector, toolPreviewAttemptedIds, toolPreviewLoadingIds]);
+
+  function cancelConnectorAuthorization(connectorId: string) {
+    setConnectorAuthorizationPending((curr) => clearConnectorAuthorizationPending(curr, connectorId));
```
mrcfps: This cancel handler only removes the local `connectorAuthorizationPending` flag, but the daemon still keeps the OAuth `state` alive in `ComposioConnectorProvider.pendingConnections`. If the user clicks Cancel and then finishes the still-open provider tab, `/api/connectors/oauth/callback/:connectorId` will still accept that callback and connect the account anyway, so the new cancel affordance does not actually cancel the authorization flow. Please pair this UI action with a daemon endpoint that invalidates the pending connection (or otherwise marks the state as canceled) and add a route/integration test that proves a callback received after cancel is rejected.

mrcfps (commented):
Thanks for the optimization work here — I found one follow-up around the new authorization-cancel flow that still needs server-side cleanup.


mrcfps (dismissed):
@alchemistklk I reviewed the latest head focusing on the partial tool-preview hydration fix, the daemon-backed authorization cancel flow, and the persisted `toolCount` rehydration path. The updated daemon route/test coverage and the web-side preview/cancel wiring line up with the earlier regressions I saw, and I did not find any new actionable issues in the changed ranges. Thanks for pushing these follow-ups through. 🙌


mrcfps (commented):
@alchemistklk Thanks for pushing the follow-ups here — I found one more drawer-preview case worth addressing before this optimization is fully robust.


apps/web/src/components/ConnectorsBrowser.tsx
```diff
@@ -210,6 +224,137 @@ function mergeConnectors(current: ConnectorDetail[], incoming: ConnectorDetail[]
...
+  return next;
+}
+
+export function getConnectorDisplayToolCount(connector: ConnectorDetail): number {
+  return connector.toolCount ?? connector.tools.length;
+}
+
+export function hasLoadedAllAdvertisedConnectorTools(connector: ConnectorDetail): boolean {
+  if (connector.toolsNextCursor) return false;
+  if (connector.toolCount === undefined) return true;
```
mrcfps: When `toolCount` is missing, this branch now treats the connector as fully hydrated before we have fetched any preview data. Most documented Composio connectors are created in `createComposioCatalogDefinition()` with `tools: []` and no `toolCount` unless `composio-descriptions.ts` has a curated count, so connectors like Bitbucket or Box will skip `fetchConnectorDetail(..., { hydrateTools: true })` entirely and the drawer falls back to "No tools available" even though the provider can still return tools. Please keep the previous `connector.tools.length > 0` behavior here, or otherwise trigger hydration whenever the advertised count is unknown and the preview list is still empty.

mrcfps (commented):
@alchemistklk Thanks for pushing these follow-ups through — the drawer hydration fixes and daemon-backed cancel flow addressed several of the earlier regressions, but I found two remaining changed-range cases that are still worth tightening before this is fully robust.


apps/daemon/src/tools/connectors.ts
```diff
@@ -70,7 +70,14 @@ export async function listConnectorTools(context: ConnectorToolContext & { useCa
     const fastDefinition = fastDefinitionsById.get(connectorId);
     return !fastDefinition || fastDefinition.tools.length === 0;
```
mrcfps: When a connector is already connected and `fastDefinition.tools` only contains the curated preview subset, this guard still skips `listHydratedDefinitions()`. For connectors like Notion, the fast catalog now advertises `toolCount: 48` but only ships two static tools, so `/api/tools/connectors/list` will keep returning just those two tools to agents even after the account is connected. That defeats the new lightweight-preview flow on the agent-facing path. Please treat `tools.length < toolCount` as needing hydration here and add a route test that covers a connected connector with partial preview tools.

apps/web/src/components/ConnectorsBrowser.tsx
```diff
@@ -525,6 +700,63 @@ export function ConnectorsBrowser({
...
+      delete next[connectorId];
+      return next;
+    });
+    setDetailConnectorId(connectorId);
+  }
+
+  async function cancelConnectorAuthorization(connectorId: string) {
+    const connector = await cancelConnectorAuthorizationRequest(connectorId);
+    if (connector) updateConnector(connector);
+    setConnectorAuthorizationPending((curr) => clearConnectorAuthorizationPending(curr, connectorId));
```
mrcfps: `cancelConnectorAuthorizationRequest()` now returns `null` on any non-200 response or fetch error, but this branch clears `connectorAuthorizationPending` unconditionally. If the daemon rejects or misses the cancel request, the UI stops showing the pending/cancel state even though the server-side OAuth state is still live, so finishing the already-open provider tab can still connect the account after the user thought they canceled. Please only clear the local pending flag after the daemon confirms cancellation, or restore it and surface an error when the POST fails.

lefarcen (approved):

PR #916, opened 2026-05-08
Add Quranic Arabic E2E tests

Radmanded:
## Summary
- Add E2E tests for Quranic Arabic UI functionality
- Add startup script for launching the app

## Test plan
- Run `pnpm --filter @open-design/e2e test` to verify tests pass
- Tests cover startup, font rendering, and basic UI functionality

### DIFF
--- a/e2e/ui/quranic-arabic.test.ts
+++ b/e2e/ui/quranic-arabic.test.ts
@@ -2,7 +2,7 @@ import { test, expect } from '@playwright/test';
 
 test.describe('Quranic Arabic Learning App', () => {
   test.beforeEach(async ({ page }) => {
-    await page.goto('http://localhost:8765/index.html');
+    await page.goto('http://localhost:17573/index.html');
   });
 
   test('should load the app and display dashboard', async ({ page }) => {
--- a/e2e/ui/quranic-arabic.test.ts
+++ b/e2e/ui/quranic-arabic.test.ts
@@ -1,0 +1,142 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Quranic Arabic Learning App', () => {
+  test.beforeEach(async ({ page }) => {
+    await page.goto('http://localhost:8765/index.html');
+  });
+
+  test('should load the app and display dashboard', async ({ page }) => {
+    // Check that the app loads
+    await expect(page).toHaveTitle(/Quranic Arabic/);
+    
+    // Dashboard should be visible initially
+    const dashboard = page.locator('#dashboard');
+    await expect(dashboard).toBeVisible();
+    
+    // Exercise and Complete screens should be hidden
+    await expect(page.locator('#exercise')).toBeHidden();
+    await expect(page.locator('#complete')).toBeHidden();
+  });
+
+  test('should display vocabulary options on dashboard', async ({ page }) => {
+    // Wait for dashboard to be visible
+    await expect(page.locator('#dashboard')).toBeVisible();
+    
+    // Check that vocabulary buttons exist
+    const vocabButtons = page.locator('.vocabulary-item');
+    const count = await vocabButtons.count();
+    expect(count).toBeGreaterThan(0);
+  });
+
+  test('should navigate to exercise when vocabulary is selected', async ({ page }) => {
+    // Wait for dashboard
+    await expect(page.locator('#dashboard')).toBeVisible();
+    
+    // Click first vocabulary item
+    const firstVocab = page.locator('.vocabulary-item').first();
+    await firstVocab.click();
+    
+    // Exercise screen should now be visible
+    await expect(page.locator('#exercise')).toBeVisible();
+    
+    // Dashboard should be hidden
+    await expect(page.locator('#dashboard')).toBeHidden();
+  });
+
+  test('should display Arabic text on exercise screen', async ({ page }) => {
+    // Navigate to exercise
+    await page.locator('.vocabulary-item').first().click();
+    await expect(page.locator('#exercise')).toBeVisible();
+    
+    // Check for Arabic text display
+    const arabicWord = page.locator('.arabic-word');
+    await expect(arabicWord).toBeVisible();
+    
+    // Arabic text should contain actual content
+    const text = await arabicWord.textContent();
+    expect(text).toBeTruthy();
+    expect(text!.trim().length).toBeGreaterThan(0);
+  });
+
+  test('should show progress indicator', async ({ page }) => {
+    // Navigate to exercise
+    await page.locator('.vocabulary-item').first().click();
+    await expect(page.locator('#exercise')).toBeVisible();
+    
+    // Progress should be visible
+    const progress = page.locator('.progress');
+    await expect(progress).toBeVisible();
+  });
+
+  test('should complete exercise and show completion screen', async ({ page }) => {
+    // Navigate to exercise
+    await page.locator('.vocabulary-item').first().click();
+    await expect(page.locator('#exercise')).toBeVisible();
+    
+    // Find and click the complete/reveal button
+    const revealButton = page.locator('.reveal-button');
+    if (await revealButton.count() > 0) {
+      await revealButton.click();
+      
+      // After completing all items, should show completion screen
+      const complete = page.locator('#complete');
+      await expect(complete).toBeVisible({ timeout: 10000 });
+    }
+  });
+
+  test('should have functioning audio controls', async ({ page }) => {
+    await expect(page.locator('#dashboard')).toBeVisible();
+    
+    // Look for audio play buttons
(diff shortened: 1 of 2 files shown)


lefarcen:
Hi @Radmanded! 🎉
Thanks for the contribution — adding E2E test coverage for Quranic Arabic rendering is valuable for catching UI regressions.

— open-design team

e2e/ui/quranic-arabic.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    // Check that the app loads
+    await expect(page).toHaveTitle(/Quranic Arabic/);
+    
+    // Dashboard should be visible initially
+    const dashboard = page.locator('#dashboard');
+    await expect(dashboard).toBeVisible();
+    
+    // Exercise and Complete screens should be hidden
+    await expect(page.locator('#exercise')).toBeHidden();
+    await expect(page.locator('#complete')).toBeHidden();
```
mrcfps, line 18: This opening scenario is written against a different UI contract than the one shipped by this repo.  still sets the page title to , and a repo-wide search outside this new file does not find , , , , , or  anywhere under , so Playwright will fail on the first assertion instead of covering a real Open Design flow. Because this file lives under , that leaves the PR adding a permanently red test suite on the current head. Please rewrite the suite to target the actual Open Design entry/project shell selectors and flows, or move these Quranic Arabic checks to the repository that owns that app.

mrcfps (changes requested):
@Radmanded Thanks for adding E2E coverage here. I found one blocker to fix before this can merge: the new Playwright suite is currently asserting a Quranic Arabic app contract that does not match the Open Design web app this repo actually serves, so the added coverage fails on the current head.


e2e/ui/quranic-arabic.test.ts
```diff
@@ -0,0 +1,142 @@
+import { test, expect } from '@playwright/test';
+
+test.describe('Quranic Arabic Learning App', () => {
+  test.beforeEach(async ({ page }) => {
```
lefarcen, line 4: **P1 — Wrong app entirely**

This test targets `http://localhost:17573/index.html` and expects a title `/Quranic Arabic/` with UI elements like `#dashboard`, `#exercise`, `#complete`, `.vocabulary-item`. None of these exist in Open Design.

**Why it matters**: This will always fail in CI and for other contributors. It looks like this test was written for a different project and accidentally submitted here.

**Concrete fix**: Remove this file, or rewrite it to test Open Design's actual UI (e.g. artifact editor with Arabic text, RTL layout in design systems, PPTX export with Arabic fonts). Use relative URLs (`await page.goto('/')`) and the repo's configured `baseURL`.

start.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+# Open Design launcher - activates Node 24 and starts the app
+cd ~/projects/open-design
```
lefarcen, line 3: **P2 — Machine-specific launcher**

This script hardcodes `~/projects/open-design`, assumes Homebrew's `fnm` at `/opt/homebrew/opt/fnm/bin`, and uses `eval "$(fnm env ...)"` which executes arbitrary shell output.

**Why it matters**: Won't work in CI, for Windows contributors, or anyone whose repo isn't at that exact path. The repo already has documented setup via Corepack + pinned pnpm.

**Concrete fix**: Don't commit personal launcher scripts. If you need a helper, keep it local (add to `.gitignore`), make it script-relative (`cd "$(dirname "$0")"` instead of hardcoded path), and document the canonical setup in `CONTRIBUTING.md`.

e2e/ui/quranic-arabic.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    await expect(progress).toBeVisible();
+  });
+
+  test('should complete exercise and show completion screen', async ({ page }) => {
+    // Navigate to exercise
+    await page.locator('.vocabulary-item').first().click();
+    await expect(page.locator('#exercise')).toBeVisible();
+    
+    // Find and click the complete/reveal button
+    const revealButton = page.locator('.reveal-button');
```
lefarcen, line 77: **P2 — Silent no-op when button missing**

The test only asserts inside `if (await revealButton.count() > 0)`, so it silently passes if `.reveal-button` doesn't exist — exactly the failure mode you'd want to catch.

**Concrete fix**: Replace the `if` with `await expect(revealButton).toBeVisible()` to require the button, then perform the full interaction deterministically.

e2e/ui/quranic-arabic.test.ts
```diff
@@ -0,0 +1,142 @@
...
+      const complete = page.locator('#complete');
+      await expect(complete).toBeVisible({ timeout: 10000 });
+    }
+  });
+
+  test('should have functioning audio controls', async ({ page }) => {
+    await expect(page.locator('#dashboard')).toBeVisible();
+    
+    // Look for audio play buttons
+    const audioButtons = page.locator('.audio-button, .play-audio, [aria-label*="play"], [aria-label*="audio"]');
```
lefarcen, line 91: **P2 — Audio controls test cannot fail**

`expect(count + exerciseCount).toBeGreaterThanOrEqual(0)` is always true (sum of two non-negative numbers is ≥ 0).

**Concrete fix**: If audio controls are required, assert `> 0`. If optional, remove the test.

e2e/ui/quranic-arabic.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    // Navigate to exercise
+    await page.locator('.vocabulary-item').first().click();
+    await expect(page.locator('#exercise')).toBeVisible();
+    
+    // Reload the page
+    await page.reload();
+    
+    // Progress might be stored in localStorage
+    // Check if there's saved progress
+    const storage = await page.evaluate(() => {
```
lefarcen, line 131: **P2 — Progress persistence test always passes**

Asserting `expect(storage).toBeTruthy()` on an object literal is always true, even when `hasProgress` is false.

**Concrete fix**: Assert the actual persisted key/value: `expect(storage.hasProgress).toBe(true)`, and verify the post-reload UI resumes the expected progress state.

e2e/ui/quranic-arabic.test.ts
```diff
@@ -0,0 +1,142 @@
...
+    await expect(page.locator('#exercise')).toBeHidden();
+    await expect(page.locator('#complete')).toBeHidden();
+  });
+
+  test('should display vocabulary options on dashboard', async ({ page }) => {
+    // Wait for dashboard to be visible
+    await expect(page.locator('#dashboard')).toBeVisible();
+    
+    // Check that vocabulary buttons exist
+    const vocabButtons = page.locator('.vocabulary-item');
```
lefarcen, line 26: **P3 — Fragile class-based selectors**

Locators like `.vocabulary-item`, `.arabic-word`, `.progress` are fragile under styling refactors and provide weak signal (checking a class exists ≠ checking behavior works).

**Concrete fix**: Use user-facing roles/text or stable `data-testid`s (consistent with existing e2e tests). Assert behavior-specific outcomes: selected vocabulary content, Arabic-script text presence, RTL `dir` attribute, progress value changes.

lefarcen (changes requested):
Hi @Radmanded — thanks for thinking about E2E test coverage! However, this PR has a fundamental scope issue that needs to be fixed before it can merge.

**Core problem**: The test file references a standalone "Quranic Arabic Learning App" with UI elements (`#dashboard`, `#exercise`, `#complete`, `.vocabulary-item`, etc.) that don't exist anywhere in the Open Design codebase. The hardcoded URL `http://localhost:17573/index.html` and title assertion `/Quranic Arabic/` suggest this test was written for a different project.

Open Design is a design tool (Figma/Canva alternative) — it doesn't have a "Quranic Arabic Learning App" mode. If you want to add E2E coverage for Arabic RTL rendering in Open Design's actual UI, you'd need to:
- Use the repo's existing Playwright setup (see `e2e/` folder)
- Test the real Open Design routes/UI (artifacts editor, design systems, export flows)
- Verify Arabic text rendering, RTL layout, font handling in those contexts

See inline comments for additional issues. Let me know if you'd like guidance on writing E2E tests for the actual app.

mrcfps:
Thanks for the contribution. We’re closing this PR because it targets a Quranic Arabic learning app surface that does not exist in this repository. Our E2E tests run through the existing tools-dev / Playwright harness and exercise real Open Design flows, while this patch hardcodes a separate app shape (localhost:17573/index.html, custom dashboard/exercise selectors, Quranic-specific storage assumptions). The added start.sh is also environment-specific and does not match this repo’s lifecycle conventions. If you want to contribute Arabic/RTL coverage, please open a new PR that tests existing Open Design behavior using the current E2E harness.

PR #924, opened 2026-05-08
fix: settle completed runs and clean up shutdown children

### LINKED ISSUE #891: Assistant turn stuck showing "Working / elapsed timer" after the run has actually completed (daemon mode)
### Observed behavior
A chat turn that actually finished (the agent has produced output, and the usage row clearly shows tokens + cost, e.g. `1785 输出 · $1.1275`) stays visually in the **Working** state with the live elapsed timer still ticking (2m 29s and counting), instead of flipping to a terminal state.

The input area also continues to show a **Stop** button even though there's nothing actively streaming.

Refreshing / closing & reopening the conversation eventually makes the message render as completed, so the run state appears to be persisted correctly on the daemon side — only the in-memory UI state is stale.

### Expected behavior
Once the run's terminal event arrives (or the final `usage` event lands and no further events follow for ~N seconds), the assistant footer should flip to the final state (`doneLabel` or `unfinishedLabel`) and the **Stop** button should revert to **Send**.

### Reproduction
1. macOS packaged build 0.5.0, daemon mode.
2. Agent: Claude-Code-compatible CLI.
3. Ask the agent to edit an existing HTML artifact (e.g. "revert the header back, keep only the original layout").
4. Wait for the agent to finish producing output. The usage / cost row appears.
5. Observe: the footer stays at **Working …** with a running timer. Timer never stops until the conversation is re-entered.

Screenshot of the stuck state:

*(will attach in a follow-up comment if the inline image above doesn't render)*

### Hypothesis (for context)
In `apps/web/src/components/ProjectView.tsx`, the `onDone` handler writes:

\`\`\`ts
updateAssistant((prev) => ({
  ...prev,
  endedAt: Date.now(),
  runStatus: config.mode === 'api' || prev.runId ? 'succeeded' : prev.runStatus,
}));
\`\`\`

If daemon mode somehow reaches `onDone` without `prev.runId` being set (e.g. `onRunCreated` fired but React batching hasn't flushed yet, or certain code paths around critique theater skip `onRunCreated` entirely), the ternary falls through to `prev.runStatus` — which is still `'running'` — and the footer never flips.

In `apps/web/src/providers/daemon.ts`, `onRunStatus?.('succeeded')` is only called when the SSE stream delivers an explicit `event === 'end'` frame. If the stream disconnects before that frame lands (network hiccup, proxy buffering, timing around `child.on('close')` → `finish(run, …)`), the UI never receives the terminal signal.

### Defensive alternatives worth considering
- Always downgrade `runStatus` to `'succeeded'` once `onDone` fires and `endedAt` is written, regardless of `prev.runId` (trust the daemon's `handlers.onDone` as authoritative — it's only called when the run actually ended on the server).
- When the `usage` event lands and no further events are received for ~5s while `streaming === true`, proactively call `fetchChatRunStatus(runId)` and reconcile.
- Ensure `onRunCreated` is guaranteed to fire before `onDone` on every code path (including critique-theater).

### Environment
- Open Design 0.5.0 (packaged macOS arm64)
- Agent: Claude-Code-compatible CLI
- macOS 15.x (Sequoia)

Siri-Ray:
## Summary

Closes #891.

- Mark successful daemon/API `onDone` paths as `succeeded` even if `runId` has not reached the assistant message state yet, preventing completed turns from staying visually stuck in `Working`.
- Add daemon run-service shutdown cleanup for all active runs: mark them canceled, emit terminal `end`, abort ACP sessions when available, send `SIGTERM` to child processes, and escalate to `SIGKILL` after a grace window.
- Wire shutdown cleanup into daemon sidecar stop and direct `od` CLI `SIGINT` / `SIGTERM` handling.
- Add focused web and daemon regression coverage.

## Test plan

- [x] `pnpm guard`
- [x] `pnpm typecheck`
- [x] `pnpm --filter @open-design/web test -- tests/components/ProjectView.run-cleanup.test.tsx`
- [x] `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/runs.test.ts tests/chat-route.test.ts`

## Notes

I initially tried `pnpm --filter @open-design/daemon test -- tests/runs.test.ts`, but this package script passes through in a way that ran the full daemon suite; one unrelated timing-sensitive chat-route test failed in that broad run. Running the relevant daemon files directly with Vitest passed, including `tests/chat-route.test.ts`.

### DIFF
--- a/apps/daemon/src/cli.ts
+++ b/apps/daemon/src/cli.ts
@@ -134,7 +134,19 @@ for (let i = 0; i < argv.length; i++) {
   }
 }
 
-startServer({ port, host }).then(url => {
+startServer({ port, host, returnServer: true }).then((started) => {
+  const { url, server, shutdown } = started;
+  let shuttingDown = false;
+  const stop = () => {
+    if (shuttingDown) return;
+    shuttingDown = true;
+    void Promise.resolve()
+      .then(() => shutdown?.())
+      .then(() => new Promise((resolve) => server.close(() => resolve())))
+      .finally(() => process.exit(0));
+  };
+  process.on('SIGINT', stop);
+  process.on('SIGTERM', stop);
   console.log(`[od] listening on ${url}`);
   if (open) {
     const opener = process.platform === 'darwin' ? 'open'
--- a/apps/daemon/src/cli.ts
+++ b/apps/daemon/src/cli.ts
@@ -136,14 +136,41 @@ for (let i = 0; i < argv.length; i++) {
 
 startServer({ port, host, returnServer: true }).then((started) => {
   const { url, server, shutdown } = started;
+  const closeTimeoutMs = 5_000;
   let shuttingDown = false;
   const stop = () => {
-    if (shuttingDown) return;
+    if (shuttingDown) {
+      process.exit(0);
+    }
     shuttingDown = true;
+    let closeTimedOut = false;
+    let closeTimeout;
     void Promise.resolve()
       .then(() => shutdown?.())
-      .then(() => new Promise((resolve) => server.close(() => resolve())))
-      .finally(() => process.exit(0));
+      .then(() => new Promise((resolve) => {
+        let resolved = false;
+        const resolveOnce = () => {
+          if (resolved) return;
+          resolved = true;
+          resolve();
+        };
+        closeTimeout = setTimeout(() => {
+          closeTimedOut = true;
+          server.closeAllConnections?.();
+          resolveOnce();
+        }, closeTimeoutMs);
+        closeTimeout.unref?.();
+        server.close(() => resolveOnce());
+      }))
+      .finally(() => {
+        if (closeTimeout) {
+          clearTimeout(closeTimeout);
+        }
+        if (!closeTimedOut) {
+          server.closeIdleConnections?.();
+        }
+        process.exit(0);
+      });
   };
   process.on('SIGINT', stop);
   process.on('SIGTERM', stop);
--- a/apps/daemon/src/cli.ts
+++ b/apps/daemon/src/cli.ts
@@ -137,40 +137,37 @@ for (let i = 0; i < argv.length; i++) {
 startServer({ port, host, returnServer: true }).then((started) => {
   const { url, server, shutdown } = started;
   const closeTimeoutMs = 5_000;
+  const closeServer = () => new Promise((resolve) => {
+    let resolved = false;
+    const resolveOnce = () => {
+      if (resolved) return;
+      resolved = true;
+      resolve();
+    };
+    const idleTimer = setTimeout(() => {
+      server.closeIdleConnections?.();
+    }, Math.min(1_000, closeTimeoutMs));
+    const hardTimer = setTimeout(() => {
+      server.closeAllConnections?.();
+      resolveOnce();
+    }, closeTimeoutMs);
+    idleTimer.unref?.();
+    hardTimer.unref?.();
+    server.close(() => resolveOnce());
+  }).finally(() => {
+    server.closeIdleConnections?.();
+  });
   let shuttingDown = false;
   const stop = () => {
     if (shuttingDown) {
       process.exit(0);
     }
     shuttingDown = true;
-    let closeTimedOut = false;
-    let closeTimeout;
+    const closePromise = closeServer();
+    const shutdownPromise = Promise.resolve().then(() => shutdown?.());
     void Promise.resolve()
-      .then(() => shutdown?.())
-      .then(() => new Promise((resolve) => {
-        let resolved = false;
-        const resolveOnce = () => {
-          if (resolved) return;
-          resolved = true;
-          resolve();
-        };
-        closeTimeout = setTimeout(() => {
-          closeTimedOut = true;
-          server.closeAllConnections?.();
-          resolveOnce();
-        }, closeTimeoutMs);
-        closeTimeout.unref?.();
-        server.close(() => resolveOnce());
-      }))
-      .finally(() => {
-        if (closeTimeout) {
-          clearTimeout(closeTimeout);
-        }
-        if (!closeTimedOut) {
-          server.closeIdleConnections?.();
-        }
-        process.exit(0);
-      });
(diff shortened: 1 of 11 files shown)


apps/daemon/src/cli.ts
```diff
@@ -134,7 +134,46 @@ for (let i = 0; i < argv.length; i++) {
...
 }
 
-startServer({ port, host }).then(url => {
+startServer({ port, host, returnServer: true }).then((started) => {
+  const { url, server, shutdown } = started;
+  const closeTimeoutMs = 5_000;
+  let shuttingDown = false;
+  const stop = () => {
+    if (shuttingDown) {
+      process.exit(0);
```
lefarcen, line 163: **P1 — New-run race during shutdown**

Shutdown snapshots and cancels active runs *before* the HTTP server stops accepting connections. During the `shutdownActive()` grace window (3s), a new `/api/chat` or `/api/runs` request can still create a run, and that child won't be SIGTERM/SIGKILLed before process exit — potential orphaned agent process.

**Fix:** Either (a) call `server.close()` *before* `shutdown()` to stop accepting new work first, OR (b) set a daemon-level `shuttingDown` gate that makes run-creation routes return 503. Ideally do both and add a test where a request races shutdown.

apps/daemon/src/cli.ts
```diff
@@ -134,7 +134,46 @@ for (let i = 0; i < argv.length; i++) {
...
+    }
+    shuttingDown = true;
```
lefarcen, line 165: **P2 — Unbounded server.close() + second-SIGINT no-op**

`server.close()` is unbounded. Any long-lived request/SSE connection can keep it from firing, preventing cleanup from starting (because `server.once('close')` is the fallback that runs `shutdownDaemonRuns()`). And a second SIGINT currently just returns instead of forcing exit.

**Fix:** Use a bounded close helper with `closeIdleConnections()` / `closeAllConnections()` escalation. Ensure run shutdown starts *before* waiting for the close event, and make the second SIGINT a hard exit.

apps/daemon/src/runs.ts
```diff
@@ -143,6 +174,27 @@ export function createChatRunService({
     }
   };
 
+  const shutdownActive = async ({ graceMs = shutdownGraceMs } = {}) => {
```
lefarcen, line 177: **P2 — ACP session abort undefined**

The new shutdown path calls `run.acpSession?.abort()`, but real `attachAcpSession()` (from `apps/daemon/src/acp.ts:485`) returns only `hasFatalError()` — no `abort()` method. The test uses a fake `{ abort }` session, so it proves the optional call but not actual ACP cleanup.

**Fix:** Add an `abort()` implementation to the ACP session adapter, or narrow the behavior/test wording to pi-rpc only and remove the `acpSession.abort()` call. Then cover the real adapter shape in the test.

lefarcen (changes requested):
Hi @Siri-Ray! Thanks for tackling the shutdown cleanup issue — the SIGTERM→SIGKILL escalation and comprehensive test coverage are solid additions. 👏

However, I found 3 blocking race conditions around the shutdown sequence that need attention before merge:

apps/daemon/src/sidecar/server.ts
```diff
@@ -88,6 +94,9 @@ export async function startDaemonSidecar(runtime: SidecarRuntimeContext<SidecarS
     stopped = true;
     state.state = "stopped";
     state.updatedAt = new Date().toISOString();
+    await serverHandle.shutdown?.().catch((error: unknown) => {
```
mrcfps: `stop()` awaits `serverHandle.shutdown()` before the HTTP listener is closed here, and `shutdownActive()` only walks the runs that already exist in its initial snapshot (`apps/daemon/src/runs.ts:177-195`). While that await is in flight, `/api/runs` and `/api/chat` are still live (`apps/daemon/src/server.ts:5167-5207`), so a request that lands during the grace window can create a brand-new child process that never gets SIGTERM/SIGKILLed before the daemon exits. That breaks the shutdown-cleanup guarantee from this PR and can leave orphaned agent processes behind. Please stop accepting new work before awaiting run cleanup (for example, close or gate the server first and/or make new run-creation requests return 503 once shutdown begins), then add a race test proving a late `/api/chat` or `/api/runs` request cannot outlive shutdown.

apps/daemon/src/sidecar/server.ts
```diff
@@ -88,6 +94,9 @@ export async function startDaemonSidecar(runtime: SidecarRuntimeContext<SidecarS
...
+      console.error("daemon shutdown cleanup failed", error);
+    });
     await ipcServer?.close().catch(() => undefined);
     await closeHttpServer(serverHandle.server).catch(() => undefined);
```
mrcfps: `closeHttpServer()` is still an unbounded wrapper around `server.close()` (`apps/daemon/src/sidecar/server.ts:42-47,101`). This sidecar also serves long-lived SSE endpoints like `/api/projects/:id/events` (`apps/daemon/src/server.ts:2190-2205`), so if the UI is connected when `stop()` runs, `server.close()` can wait forever and the sidecar shutdown never completes. The CLI path in this PR fixed the same class of hang with a bounded timeout plus `closeAllConnections()`, but this stop path still blocks indefinitely under the same conditions. Please reuse a bounded close helper here as well (including `closeIdleConnections()` / `closeAllConnections()` escalation) so sidecar shutdown always finishes even with live SSE clients.

mrcfps (changes requested):
@Siri-Ray Thanks for the shutdown cleanup work — the run teardown coverage and the bounded CLI exit path both move this in the right direction. I found two remaining shutdown blockers in the sidecar/stop sequencing that still need attention before merge.


mrcfps (dismissed):
@Siri-Ray I re-checked the shutdown flow end to end: the daemon now rejects new `/api/runs` and `/api/chat` work as shutdown begins, both the CLI and sidecar paths bound long-lived HTTP shutdown, and the added ACP/run cleanup coverage matches those lifecycle changes. Thanks for tightening up the teardown sequencing and backing it with focused regression coverage. 🙌


lefarcen (approved):

PR #738, opened 2026-05-07
fix(web): replace SketchEditor window.prompt() text tool with in-app modal

Sid-Qin:
## Summary

The sketch editor's `text` tool silently does nothing in the desktop app. Clicking the canvas with the text tool active calls `window.prompt()`, which Electron 28+ disables by default — the call returns null synchronously and the handler's `if (text)` guard exits before any UI appears.

This is the same root cause as the FileViewer Save-as-template regression fixed in [#723](https://github.com/nexu-io/open-design/issues/723) (precedent branch `fix/save-as-template-modal-723`). I'm not claiming to close #723 — that PR fixes the FileViewer side. This PR is the second of the two `window.prompt()` callsites in the web app (the third lives inside FileViewer and is part of the #723 fix). After both land, there are no remaining `window.prompt()` calls in `apps/web`.

## What changed

- `apps/web/src/components/SketchEditor.tsx:129` — replace the `window.prompt(t('sketch.textPrompt'))` call with an in-app modal:
  - Stash the click position in a ref when the user clicks the canvas with the text tool active.
  - Open a modal with a single text input (`autoFocus`, Enter-to-submit when non-empty, Escape-to-cancel).
  - Append the `TextItem` only on confirm using the stashed anchor; Cancel and Escape clear state.
- Reuses the existing `.modal-backdrop` / `.modal` / `.modal-head` / `.modal-foot` classes from `apps/web/src/index.css` — no new CSS.
- Adds one i18n key, `sketch.textModalTitle` ("Add text"), to the `Dict` interface and all 16 locale files. The existing `sketch.textPrompt` is reused as the field label so no other translator churn.

Strictly scoped to `SketchEditor.tsx` + i18n; `FileViewer.tsx` is intentionally untouched (#723's territory).

## Test plan

- [x] `pnpm --filter @open-design/web typecheck` — clean
- [x] `pnpm --filter @open-design/web test` — 320 tests across 44 files pass
- [x] `tsx scripts/i18n-check.ts` — passes (locale registration, README switchers, core doc links consistent)
- [ ] Manual: in the desktop app, open a sketch file, pick the text tool, click on the canvas — modal opens, Enter places the text at the click point, Cancel/Escape leave the canvas unchanged.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #933, opened 2026-05-08
fix(desktop): allow od:// URLs in setWindowOpenHandler so live artifact previews open in child window (#911)

Sid-Qin:
## Summary

- The Orbit panel's "Open artifact" button is an `<a target="_blank" href="/api/live-artifacts/.../preview?projectId=...">` rendered by [`apps/web/src/components/SettingsDialog.tsx:3274`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SettingsDialog.tsx#L3274). In **packaged Electron builds** the renderer lives at `od://app/`, so that relative `href` resolves to `od://app/api/live-artifacts/.../preview?projectId=...` by the time `setWindowOpenHandler` sees it.
- The existing handler in [`apps/desktop/src/main/runtime.ts:300`](https://github.com/nexu-io/open-design/blob/main/apps/desktop/src/main/runtime.ts#L300) ran three checks: `isAllowedChildWindowUrl(url)` (only `blob:`), `isHttpUrl(url)` (only `http(s):`), then fall through to `{ action: "deny" }`. The packaged `od://` scheme matched neither, so the click was silently dropped — Orbit's "Open artifact" became a no-op for every desktop user.
- Dev mode (`http://127.0.0.1:port/`) was unaffected: its links resolved through the http branch and opened in the user's external browser via `shell.openExternal`.

Fix: extend `isAllowedChildWindowUrl` to also accept the packaged `od:` scheme. Electron then opens a child BrowserWindow that inherits the same protocol registration + preload, so the live artifact preview HTML (served via the `od://` proxy in `apps/packaged/src/protocol.ts`) renders in the new window. 1-line core change in `runtime.ts`, plus a re-export and test.

Fixes #911.

## Behaviour delta

| Mode | Before | After |
|---|---|---|
| Packaged (Electron, `od://app/` renderer) | click → `od://app/api/.../preview` → silently denied | click → child BrowserWindow renders the artifact preview |
| Dev mode (`http://127.0.0.1:port/`) | link → `http://127.0.0.1:.../api/.../preview` → `shell.openExternal` opens in user's default browser | unchanged |
| Other links (github.com, external docs, etc.) inside packaged | open in user's default browser via `shell.openExternal` | unchanged — `od:` is not a wildcard, only the packaged scheme passes the allowlist |

## Why a child window is the right target (not `shell.openExternal`)

Routing `od://` through `shell.openExternal` would fail because `od:` is a privileged Electron protocol, not an OS-level URL handler — the OS doesn't know what to do with it. The only way to render the preview is inside the same Electron runtime that registered the scheme, which means a child BrowserWindow. The child inherits the parent's `webPreferences` (sandbox, contextIsolation, preload), so it lands at the same security posture as the main window.

## Diff shape

| File | What changes |
|---|---|
| `apps/desktop/src/main/runtime.ts` | Extend `isAllowedChildWindowUrl` to accept `od:` alongside `blob:`. Mark `isHttpUrl` and `isAllowedChildWindowUrl` `export` so the packaged workspace's vitest can pin them. Heavy doc comment on the new branch citing the Orbit no-op. |
| `apps/desktop/src/main/index.ts` | Re-export the two helpers through `@open-design/desktop/main` so cross-package tests can import them via the package boundary (avoids `rootDir` violation in `tsc -p tsconfig.tests.json`). |
| `apps/packaged/tests/desktop-url-allowlist.test.ts` | New: 8 tests pinning both helpers — `od:` allowed (the #911 regression guard), `blob:` still allowed (existing), `http(s):` NOT allowed by this branch (handled by sibling `isHttpUrl` path), `file:` / `javascript:` / `data:` explicitly NOT allowed (security boundary). |

3 files, +124 / −3.

## What's verified

| Check | Result |
|---|---|
| `pnpm --filter @open-design/packaged vitest tests/desktop-url-allowlist.test.ts` | **8/8** |
| `pnpm --filter @open-design/desktop typecheck` | clean |
| packaged `tsc -p tsconfig.json --noEmit` | clean |
| packaged `tsc -p tsconfig.tests.json --noEmit` (the CI killer) | clean |

The pre-existing `apps/packaged/tests/sidecars.test.ts > adds custom VP_HOME/bin to the packaged PATH builder` failure is unrelated to this PR — confirmed by stashing the change and re-running on the bare branch base. Out of scope here.

## What's NOT verified

I don't have a packaged build pipeline locally, so the actual click-through ("press Open artifact → child window pops with live artifact preview") needs a desktop tester. The structural argument:

1. The new test asserts `isAllowedChildWindowUrl('od://app/api/live-artifacts/abc/preview?projectId=p1') === true` — the exact URL shape that the orbit panel produces in packaged builds.
2. `setWindowOpenHandler` returning `{ action: "allow" }` is Electron's documented "open a child BrowserWindow" path — same mechanism `blob:` already uses for in-renderer download URLs.
3. The child window inherits the parent's `webPreferences` and the app-level `od://` protocol registration, so loading `od://app/api/.../preview` in the child runs through the same proxy in `apps/packaged/src/protocol.ts` that already serves the main window's API calls.

If @shangxinyu1 or any desktop user can pull `Sid-Qin:fix/orbit-open-artifact-desktop-911` against a packaged build and confirm Settings → Orbit → Open artifact now opens the preview, that closes the loop.

## Test plan

- [x] 8 new tests covering the URL allowlist surface.
- [x] Both packaged typecheck configs clean.
- [x] Desktop typecheck clean.
- [ ] Manual smoke on packaged build: open Settings → Orbit, click **Open artifact**, expect a child window with the live artifact preview rendered.
- [ ] Verify other `target="_blank"` links inside the packaged app still route to `shell.openExternal` (e.g. clicking a github.com link in the readme/credits panel should still open the user's default browser, not pop a child window).

## Notes

- This is the third desktop-side fix in a row that turns out to be an Electron-policy gap (#738 `window.prompt` disabled, #895 undici socket option, now #911 child-window allowlist). The pattern is consistent: web code that works fine in a browser quietly stops working in the packaged Electron host because of a default-deny policy that nobody has noticed yet. Worth flagging in case the team wants to invest in a packaged-mode QA sweep.
- The exported helpers are pure functions; making them part of `@open-design/desktop/main`'s public API doesn't add runtime coupling — only test reachability.

### DIFF
--- a/apps/desktop/src/main/index.ts
+++ b/apps/desktop/src/main/index.ts
@@ -27,6 +27,13 @@ import { readProcessStamp } from "@open-design/platform";
 
 import { createDesktopRuntime } from "./runtime.js";
 
+// Re-export pure URL-policy helpers so the packaged workspace's
+// vitest can pin their behaviour without spinning up a full Electron
+// runtime. They are part of the security boundary for child-window
+// navigation (see `setWindowOpenHandler` in `runtime.ts`), so
+// pinning them is worth the small extra surface.
+export { isAllowedChildWindowUrl, isHttpUrl } from "./runtime.js";
+
 const TOOLS_DEV_PARENT_PID_ENV = SIDECAR_ENV.TOOLS_DEV_PARENT_PID;
 
 export type DesktopMainOptions = {
--- a/apps/desktop/src/main/runtime.ts
+++ b/apps/desktop/src/main/runtime.ts
@@ -182,7 +182,11 @@ async function applyWindowChromeCss(window: BrowserWindow): Promise<void> {
   await window.webContents.insertCSS(MAC_WINDOW_CHROME_CSS, { cssOrigin: "user" });
 }
 
-function isHttpUrl(url: string): boolean {
+// Exported for unit tests in `apps/packaged/tests/desktop-url-allowlist.test.ts`
+// — these are pure URL-policy helpers and `apps/desktop` itself has no
+// vitest setup, so the packaged workspace hosts the coverage. Keep them
+// pure and side-effect-free.
+export function isHttpUrl(url: string): boolean {
   try {
     const parsed = new URL(url);
     return parsed.protocol === "http:" || parsed.protocol === "https:";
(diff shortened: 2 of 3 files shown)


mrcfps (approved):
@Sid-Qin I reviewed the desktop URL-policy change against the packaged `od://` protocol flow and the new regression coverage. The `setWindowOpenHandler` update lines up with how `apps/packaged/src/protocol.ts` proxies `od://app/...`, and the added tests pin the main allow/deny scheme boundaries for this behavior. Thanks for tightening this regression guard and documenting the intent here. 🙌

PR #937, opened 2026-05-08
fix: sync Orbit last run with selected prompt template

### LINKED ISSUE #910: Orbit Last run does not update when switching prompt templates
## Summary

In Orbit settings, the **Last run** section always shows the most recent run overall instead of the latest run for the currently selected prompt template.

## Current behavior

After running Orbit with multiple design or prompt templates, the **Last run** timestamp and summary stay tied to the newest run globally. When switching the **Prompt template** dropdown, the **Last run** panel does not update to match the selected template.

## Expected behavior

The **Last run** section should reflect the latest run associated with the currently selected prompt template. Changing the **Prompt template** selection should immediately sync the timestamp and run summary shown in the panel.

## Steps to reproduce

1. Run Orbit with one prompt template.
2. Run Orbit again with a different prompt template.
3. Open Settings > Orbit.
4. Switch the **Prompt template** dropdown between the templates that have run history.
5. Check the **Last run** timestamp and summary.

## Notes

- This was observed after Orbit ran multiple design templates / prompt templates.
- The UI only displays the latest run overall.
- Switching the prompt template does not refresh the **Last run** panel.
- A screenshot was provided showing the Orbit settings panel and the stale **Last run** section.

## Why this matters

This makes the Orbit history misleading and breaks the mental model that the settings panel is showing status for the currently selected template. Users can easily think a template ran more recently than it actually did, or that switching templates had no effect.

Image

Derrick-xn:
## Summary
- persist Orbit last-run history per template instead of only keeping a single global summary
- return template-scoped Orbit status from the daemon and reuse a shared Orbit status contract
- update Orbit settings so the Last run panel follows the currently selected prompt template

## Test Plan
- pnpm guard
- pnpm typecheck
- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/orbit.test.ts
- pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/SettingsDialog.orbit.test.tsx tests/components/SettingsDialog.test.ts
- pnpm --filter @open-design/web build

Closes #910


Derrick-xn:
Root cause was on both sides of the boundary, not just in the settings UI. The daemon only persisted a single global `lastRun`, and the Orbit settings panel rendered that value directly, so switching the prompt-template dropdown could never show template-specific history.

This PR fixes it by:
- storing Orbit run summaries with `templateSkillId`
- returning `lastRunsByTemplate` from `/api/orbit/status` while keeping the legacy global `lastRun`
- updating the Orbit settings panel to prefer the currently selected template's last run, with legacy fallback only when the daemon does not yet expose template-scoped history
- adding daemon + web regression coverage for template switching

Validation run for this branch:
- `pnpm guard`
- `pnpm typecheck`
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/orbit.test.ts`
- `pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/SettingsDialog.orbit.test.tsx tests/components/SettingsDialog.test.ts`

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2715,7 +2696,12 @@ function OrbitSection({
     })();
   };
 
-  const lastRun = status?.lastRun ?? null;
+  const templateScopedLastRun = effectiveTemplateSkillId
+    ? status?.lastRunsByTemplate?.[effectiveTemplateSkillId] ?? null
+    : null;
+  const lastRun = supportsTemplateScopedHistory
+    ? templateScopedLastRun
+    : status?.lastRun ?? null;
```
mrcfps, line 2655: When `status.lastRunsByTemplate` is present but does not contain the currently selected template, this branch now resolves `lastRun` to `null` instead of falling back to `status.lastRun`. That happens for users upgrading from the old single-summary `activity-summary.json` format: `normalizeSummaryStore()` keeps the legacy `lastRun`, but it cannot synthesize a template entry when the stored summary has no `templateSkillId`. The result is that the Last run panel goes blank after upgrade until the user triggers a fresh Orbit run. Could we fall back to `status.lastRun` whenever `templateScopedLastRun` is missing, and add a regression test for the migrated payload?

mrcfps (commented):
Thanks for the Orbit history sync work — I found one migration follow-up before merge. The inline note covers the legacy-summary fallback so existing users keep seeing their previous Last run data after upgrading.


Derrick-xn:
Follow-up for the migration feedback from @chatgpt-codex-connector and @mrcfps: fixed in `9f455eb1`.

The concern was valid: with the first version of this PR, users upgrading from the old single-summary Orbit payload could temporarily lose the Last run panel because `lastRunsByTemplate` existed but did not yet contain an entry for the selected template.

I did **not** switch to a broad `templateScopedLastRun ?? status.lastRun` fallback, because that would reintroduce the original bug and let templates with no history display the newest global run again.

The updated behavior is narrower:
- preserve a legacy unscoped `lastRun` only for the template that was selected when the old payload was first loaded
- clear that fallback after the user switches templates
- keep template-specific history authoritative once it exists

Added a regression test for the migrated `lastRun` + empty `lastRunsByTemplate` payload as part of the follow-up.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2678,6 +2659,17 @@ function OrbitSection({
...
+
+  useEffect(() => {
+    const hasTemplateScopedHistory = Object.keys(status?.lastRunsByTemplate ?? {}).length > 0;
+    const hasLegacyUnscopedLastRun = Boolean(status?.lastRun && !status.lastRun.templateSkillId);
+    if (hasLegacyUnscopedLastRun && !hasTemplateScopedHistory) {
+      setLegacyLastRunTemplateSkillId(effectiveTemplateSkillId || null);
+      return;
+    }
+    setLegacyLastRunTemplateSkillId(null);
+  }, [status]);
```
mrcfps: When `/api/orbit/status` keeps returning the migrated legacy shape (`lastRun` without `templateSkillId`, `lastRunsByTemplate: {}`), this effect reassigns `legacyLastRunTemplateSkillId` from the *current* dropdown selection on every status refresh. After a user switches away from the original template, any later `refreshStatus()` call will stamp the legacy run onto the newly selected template again, so the UI can start showing the old global summary under the wrong template after the next poll/refresh. Could we initialize this fallback mapping only once for a given legacy payload (for example, only when `legacyLastRunTemplateSkillId` is still null, and otherwise preserve the original template until scoped history arrives), and add a regression test that switches templates and then simulates another `/api/orbit/status` refresh?

mrcfps (commented):
Thanks for the quick follow-up here — the template-scoped history shape and the migration guard both look close, but I found one remaining refresh-path regression to tighten up before this is fully done.


Derrick-xn:
Follow-up for the remaining refresh-path regression from @mrcfps: fixed in `b1093fbc`.

The issue was that the legacy fallback mapping could be rebound on every later `/api/orbit/status` refresh. After switching templates, a subsequent poll could stamp the old unscoped summary onto the newly selected template again.

The fix now pins the legacy fallback to the template that was active when that legacy payload was first seen, and preserves that mapping across subsequent status refreshes until real template-scoped history arrives or the legacy payload disappears.

Added regression coverage for:
- switching away from the original template under a legacy payload
- a later status refresh/poll still not reassigning that legacy summary to the newly selected template

Validation:
- `pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/SettingsDialog.orbit.test.tsx`

mrcfps:
@Derrick-xn I'm holding off on generating review comments for nexu-io/open-design#937 because this pull request has merge conflicts right now.

lefarcen:
Hi @Derrick-xn! 🎉
Thanks for the contribution — fixing the Orbit template history sync is a key usability improvement.

— open-design team

apps/daemon/src/orbit.ts
```diff
@@ -369,6 +449,7 @@ export class OrbitService {
           startedAt,
           completedAt,
           trigger,
+          templateSkillId: template?.id ?? this.config.templateSkillId ?? null,
```
lefarcen: **P2** — This reads `this.config.templateSkillId` at completion time, so a run whose template could not be resolved can be written under whatever template is selected later while the agent is still running.

Capture the configured template id at run start, use that value for resolver lookup and the summary fallback, and add a test that changes config before `completion` resolves.

lefarcen (changes requested):
Hi @Derrick-xn — thanks for the Orbit template history sync work. The daemon migration path and Settings UI fallback logic look solid, and the test coverage is focused.

I found one race condition where an in-flight run can be attributed to the wrong template if config changes before completion — see inline below.

lefarcen:
Agent-to-agent announce: review of #937 found one request-changes correctness issue. In apps/daemon/src/orbit.ts, the summary fallback reads this.config.templateSkillId when the run completes, so an unresolved-template run can be attributed to a later selected template if config changes while the agent is in flight. Suggested fix: capture the configured template id at run start, use that captured value for resolver lookup and summary fallback, and add a regression test that changes config before completion resolves.

Derrick-xn:
Follow-up for the daemon race reported by @lefarcen: fixed in `37d18ff2`.

The issue was valid: when template resolution returned `null`, the completion path still read `this.config.templateSkillId` at write time, so changing the selected template while the run was in flight could attribute that summary to the wrong template.

The fix now captures the configured template id at run start and reuses that pinned value for both:
- template resolver lookup
- summary `templateSkillId` fallback when resolution stays null

I also added a regression test that starts a run, changes the config before completion resolves, and verifies the stored summary still stays on the original template. While wiring that, I tightened the existing multi-run daemon test to wait for each prior run to fully settle before starting the next one so the assertions are deterministic.

Validation run for this follow-up:
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/orbit.test.ts`

mrcfps:
@Derrick-xn I'm holding off on generating review comments for nexu-io/open-design#937 because this pull request has merge conflicts right now.

Derrick-xn:
Conflicts with `main` are resolved on the branch now.

Follow-up commits after the daemon race fix:
- `654f7612` merges the branch forward and keeps both `@open-design/contracts` API exports (`orbit` and `finalize`) in the contracts build/package exports.
- `98066992` syncs the Orbit settings test fixtures with the new `SkillSummary.aggregatesExamples` field from `main`.

Validation after the merge-forward:
- `pnpm install`
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/orbit.test.ts`
- `pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/SettingsDialog.orbit.test.tsx tests/components/SettingsDialog.test.ts`

mrcfps (approved):
@Derrick-xn I re-reviewed the latest head and verified the template-scoped Orbit history flow end to end: the daemon now persists both the global last run and per-template history, the settings panel follows the selected template with the legacy fallback/race fixes in place, and the focused daemon/web regressions plus the contracts build all pass locally.

Thanks for pushing through the follow-up fixes here — this looks ready to merge. ✨


lefarcen (approved):
Thanks @Derrick-xn — the daemon race fix is solid: the template id is now captured at run start and reused consistently for resolver lookup and the summary fallback. The mid-flight config-change regression test covers the exact scenario from my earlier review.

The PR is clean on my side. ✅

lefarcen:
Merged! 🎉 Thanks @Derrick-xn for the contribution to open-design. Looking forward to the next one!

PR #938, opened 2026-05-08
fix(web): correct srcdoc injection and deck bridge for JS strings con…

### LINKED ISSUE #889: App  自带的预览功能异常
使用浏览器查看没有问题，但是使用内置预览功能时，如下图所示
Image
预览文件：
[index.html](https://github.com/user-attachments/files/27509176/index.html)

Morzorz:
Closes #889 
image

…taining HTML tags

- Use indexOf for </head> (real tag precedes JS string occurrences)
- Use lastIndexOf for </body> (real tag follows JS string occurrences)
- Scope deck bridge slide selector to direct children of deck containers to avoid counting cloned overview thumbnails
- Update .slide-number data attributes and .progress-bar width from deck bridge's updateDeckChrome so page counter and progress track with deck navigation

### DIFF
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -105,13 +105,17 @@ function injectManualEditBridge(doc: string): string {
 }
 
 function injectBeforeHeadEnd(doc: string, payload: string): string {
-  if (/<\/head>/i.test(doc)) return doc.replace(/<\/head>/i, `${payload}</head>`);
+  const tag = '</head>';
+  const idx = doc.toLowerCase().indexOf(tag);
+  if (idx >= 0) return doc.slice(0, idx) + payload + doc.slice(idx);
   if (/<head[^>]*>/i.test(doc)) return doc.replace(/<head[^>]*>/i, (m) => `${m}${payload}`);
   return payload + doc;
 }
 
 function injectBeforeBodyEnd(doc: string, payload: string): string {
-  if (/<\/body>/i.test(doc)) return doc.replace(/<\/body>/i, `${payload}</body>`);
+  const tag = '</body>';
+  const idx = doc.toLowerCase().lastIndexOf(tag);
+  if (idx >= 0) return doc.slice(0, idx) + payload + doc.slice(idx);
   return doc + payload;
 }
 
@@ -631,12 +635,16 @@ html[data-od-comment-mode] body * { cursor: crosshair !important; }
 html[data-od-inspect-mode] body * { cursor: crosshair !important; }
 html[data-od-comment-mode][data-od-comment-mode-kind="pod"] body * { cursor: cell !important; }
 </style>`;
-  const withStyle = /<\/head>/i.test(doc)
-    ? doc.replace(/<\/head>/i, style + '</head>')
+  const headEndTag = '</head>';
+  const headEndIdx = doc.toLowerCase().indexOf(headEndTag);
+  const withStyle = headEndIdx >= 0
+    ? doc.slice(0, headEndIdx) + style + doc.slice(headEndIdx)
     : /<head[^>]*>/i.test(doc)
       ? doc.replace(/<head[^>]*>/i, (m) => m + style)
       : style + doc;
-  if (/<\/body>/i.test(withStyle)) return withStyle.replace(/<\/body>/i, script + '</body>');
+  const bodyEndTag = '</body>';
+  const bodyEndIdx = withStyle.toLowerCase().lastIndexOf(bodyEndTag);
+  if (bodyEndIdx >= 0) return withStyle.slice(0, bodyEndIdx) + script + withStyle.slice(bodyEndIdx);
   return withStyle + script;
 }
 
@@ -670,8 +678,10 @@ function injectDeckBridge(doc: string, initialSlideIndex = 0): string {
   const styleFix = `<style data-od-deck-fix>
 .stage, .deck-stage, .deck-shell { place-content: center !important; }
 </style>`;
-  const docWithStyle = /<\/head>/i.test(doc)
-    ? doc.replace(/<\/head>/i, styleFix + "</head>")
+  const headEndTag = '</head>';
+  const headEndIdx = doc.toLowerCase().indexOf(headEndTag);
+  const docWithStyle = headEndIdx >= 0
+    ? doc.slice(0, headEndIdx) + styleFix + doc.slice(headEndIdx)
     : /<head[^>]*>/i.test(doc)
     ? doc.replace(/<head[^>]*>/i, (m) => m + styleFix)
     : styleFix + doc;
@@ -679,7 +689,7 @@ function injectDeckBridge(doc: string, initialSlideIndex = 0): string {
   const script = `<script data-od-deck-bridge>(function(){
   var initialSlideIndex = ${safeInitialSlideIndex};
   var didRestoreInitialSlide = initialSlideIndex <= 0;
-  function slides(){ return document.querySelectorAll('.slide'); }
+  function slides(){ return document.querySelectorAll('.deck > .slide, .deck-stage > .slide, .deck-shell > .slide, body > .slide'); }
   function scroller(){
     if (document.body && document.body.scrollWidth > document.body.clientWidth + 1) return document.body;
     return document.scrollingElement || document.documentElement;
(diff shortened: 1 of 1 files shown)


apps/web/src/runtime/srcdoc.ts
```diff
@@ -754,6 +764,12 @@ function injectDeckBridge(doc: string, initialSlideIndex = 0): string {
     if (total) total.textContent = pad2(count);
     if (prev) prev.toggleAttribute('disabled', i <= 0);
     if (next) next.toggleAttribute('disabled', i >= count - 1);
+    document.querySelectorAll('.slide-number').forEach(function(el){
+      el.setAttribute('data-current',i+1); el.setAttribute('data-total',count);
+    });
+    document.querySelectorAll('.progress-bar>span').forEach(function(el){
+      el.style.width=(count?((i+1)/count*100)+'%':'0');
```
mrcfps: `updateDeckChrome()` only runs from `setActive()`, but the horizontal-scroll path below goes through `go()` → `scrollGo()` → `report()` without ever calling `setActive()`. That means these new `.slide-number` / `.progress-bar` updates still never run for scroll decks, so the simple-deck-style counter and progress bar stay stale after navigation even though this PR is trying to fix that behavior. Please move this chrome sync into `report()` (or another path shared by both scroll and class-toggle decks) so every supported deck mode updates the in-preview footer consistently.

mrcfps (changes requested):
@Morzorz Thanks for chasing down the srcdoc tag-injection and deck bridge edge cases. I found one remaining issue that still leaves the in-preview slide counter/progress stale for one of the supported deck modes.


apps/web/src/runtime/srcdoc.ts
Morzorz: Got it, thanks for the correction. I’ll update it.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -105,13 +105,17 @@ function injectManualEditBridge(doc: string): string {
 }
 
 function injectBeforeHeadEnd(doc: string, payload: string): string {
-  if (/<\/head>/i.test(doc)) return doc.replace(/<\/head>/i, `${payload}</head>`);
+  const tag = '</head>';
+  const idx = doc.toLowerCase().indexOf(tag);
```
mrcfps: `injectBeforeHeadEnd()` still breaks the PR's target case when a script inside `<head>` contains the literal `</head>`. Because `indexOf('</head>')` returns the first textual match, HTML like `<head><script>const s = "</head>"</script></head>` gets the payload inserted into the string literal instead of before the real closing tag, which corrupts the iframe HTML. The same raw search is duplicated in `injectSelectionBridge()` and `injectDeckBridge()` below, so this still leaves srcdoc injection incorrect for templates that load JS in `<head>`. Please resolve the closing tag relative to the actual `<head>` element (for example by parsing and appending through the DOM, or by searching for `</head>` only after the matched `<head...>` opener) and then reuse that logic at the two duplicated call sites.

mrcfps (changes requested):
@Morzorz Thanks for addressing the earlier deck bridge issue. I rechecked the updated scroll-deck path and the srcdoc injection changes, and I found one remaining `</head>` case that can still corrupt the injected HTML when preview scripts live inside `<head>`.


apps/web/src/runtime/srcdoc.ts
lefarcen: **P1** — mrcfps's finding still applies here.

`indexOf('</head>')` returns the **first** textual match in the lowercased string, regardless of HTML structure. When a script inside `<head>` contains the literal `</head>` (e.g., `<script>const s = "</head>"</script>`), this code injects the payload **into the string literal** instead of before the real closing tag, corrupting the iframe HTML.

The same pattern is duplicated in `injectSelectionBridge()` (line 638) and `injectDeckBridge()` (line 681), so all three call sites have the same vulnerability.

**Fix approach:**
- Parse the HTML structure (e.g., use DOMParser in a detached document) to locate the **actual** `</head>` element, then inject relative to that element node.
- OR: search for `</head>` only **after** the matched `<head...>` opener (find `<head`, then search for `</head>` starting from that offset + opener length). This avoids hitting string literals that appear lexically before the `<head>` block.
- Reuse the fix at all three call sites (don't leave duplicated vulnerable code).

**Test case to verify the fix:**
```html
<head><script>const sentinel = "</head>"</script></head>
<body>content</body>
```
Payload injection should produce:
```html
<head><script>const sentinel = "</head>"</script>[PAYLOAD]</head>
```
Not:
```html
<head><script>const sentinel = "[PAYLOAD]</head>"</script></head>
```

lefarcen (dismissed):
Hey @Morzorz, thanks for tackling the srcdoc injection edge cases! I verified mrcfps's finding against the current head (1a4fd19), and the core issue he identified is still present. See inline comment below.

apps/web/src/runtime/srcdoc.ts
Morzorz: String matching for injection is indeed error-prone. I'll use DOMParser + DOM operations to replace all string-based injection logic.

Morzorz: I've switched to using DOMParser and already committed the changes.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -104,15 +104,31 @@ function injectManualEditBridge(doc: string): string {
   return injectBeforeBodyEnd(withStyle, buildManualEditBridge(true));
 }
 
+// DOM-based HTML mutation: parse with DOMParser, apply mutations, serialize back.
+// Avoids string-matching pitfalls where </head>/<body> appear inside <script>
+// or <style> text — DOMParser correctly separates raw-text element content from
+// structural markup.
+function domMutate(doc: string, mutate: (parsed: Document) => void): string {
+  if (typeof DOMParser === 'undefined') return doc;
```
mrcfps: `domMutate()` now returns the original document unchanged whenever `DOMParser` is missing. In this repo, `apps/web/tests/runtime/srcdoc.test.ts` runs under `apps/web/vitest.config.ts`'s `environment: 'node'`, and Node 24 does not provide a global `DOMParser`, so every bridge that now funnels through this helper (`injectBeforeHeadEnd`, `injectBeforeBodyEnd`, `injectSelectionBridge`, and `injectDeckBridge`) becomes a no-op there. That breaks the existing `buildSrcdoc()` expectations for injected deck/selection/edit markup even though this PR only meant to change *where* the payload is inserted. Please keep a non-DOM fallback here (or create a local parser inside the helper) so `buildSrcdoc()` still injects these bridges in Node-based tests and any other non-browser callers.

mrcfps (changes requested):
@Morzorz Thanks for replacing the raw tag matching with a DOM-based injection path — that closes the original `</head>` / `</body>` corruption case. I found one blocker on the current head: the new helper silently drops all of these injections in the repo's Node-based test/runtime path.


lefarcen (dismissed):
Hey @Morzorz, excellent work on the DOMParser fix! 🎉

The string-matching vulnerability is completely resolved. Using `DOMParser` + DOM operations properly separates script/style text content from structural markup, eliminating the edge case where `</head>` or `</body>` inside JS strings would corrupt injection.

The implementation is clean:
- Proper error handling with fallback
- Reused across all three call sites (no duplication)
- Documented the rationale

apps/web/src/runtime/srcdoc.ts
Morzorz: Sure, I’ve added a string fallback for non-browser environments and covered edge cases accordingly.

mrcfps (approved):
@Morzorz I rechecked the current head's srcdoc injection path, the non-browser fallback, and the deck bridge counter/progress updates. The DOMParser path plus the structural string fallback now cover the earlier `</head>` / `</body>` corruption case without regressing the Node-based runtime path. Thanks for iterating on this — the final shape looks solid.


lefarcen (approved):
The DOMParser + string fallback approach addresses both the indexOf vulnerability and the Node test environment. The deck bridge fixes look solid. Ready to ship.

leilei926524-tech:
Hi @Morzorz — welcome to Open Design contributors. 🎉

Your first merged PR (#938) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #943, opened 2026-05-08
fix(daemon): typecheck leaf modules

nettee:
## Summary
- Typecheck daemon leaf modules and supporting helpers without touching the core server hot path.
- Update matching daemon tests for the stricter types.
- Refresh the maintainability roadmap drift entry included in this split.

## Validation
- Not run in this split step.


PerishCode (approved):

apps/daemon/src/artifact-manifest.ts
```diff
@@ -178,18 +194,18 @@ export function validateArtifactManifestInput(manifest, entry) {
   return { ok: true, value: sanitizeManifest(manifest, safeEntry) };
 }
 
-export function sanitizeManifest(manifest, entry) {
+export function sanitizeManifest(manifest: JsonRecord, entry: string): JsonRecord {
```
kpatrasubham, line 197: Consider adding a validation layer for the `renderer` and `exports` properties to prevent runtime errors from malformed input. Additionally, you should replace the `now` constant with a `lastModified` check if you want to avoid updating `updatedAt` when no actual data changes occur.

kpatrasubham (commented):
submited

Issue #61, opened 2026-04-29
Python files in Design Files cannot be previewed inline


shangxinyu1:
## Summary
Python files in the project file list currently do not support inline preview.

When a generated or exported `.py` file appears in Design Files, the side panel does not show a readable code preview. Users can only rely on actions like open or download, which makes it harder to inspect generated scripts quickly inside the workspace.

## Expected behavior
Selecting a `.py` file should show an inline code/text preview in the side panel, similar to how text-based project artifacts are expected to be inspectable without leaving the current workspace.

## Actual behavior
When selecting a Python file:
- the file appears in the list
- the side panel does not render a readable code preview
- the file effectively behaves like a non-previewable attachment
- users need to open or download it instead of inspecting it inline

## Why this is a problem
- Python files are text files and are reasonable candidates for inline preview.
- Users may want to quickly verify generated export scripts before opening them externally.
- This creates an inconsistent file experience between previewable text artifacts and source-code files.
- It slows down review and debugging workflows.

## Steps to reproduce
1. Generate or add a `.py` file to the project.
2. Open the Design Files / project file list.
3. Select the Python file.
4. Observe that there is no useful inline preview for the file contents.

## Expected fix direction
Possible approaches:
- Treat `.py` as a previewable text/code file in the file viewer.
- Render the file content in a read-only code/text preview panel.
- Reuse the same preview path used for other plain-text or code-like artifacts.

## Priority
Medium — not as blocking as failed exports, but it weakens inspection and debugging workflows for generated code artifacts.

## Screenshot

<img width="5120" height="2638" alt="Image" src="https://github.com/user-attachments/assets/fb3f7b88-e81b-45bd-bb9c-c789c4143c35" />

lefarcen:
你好 @shangxinyu1，感谢提 issue 🙏 已转给 @Siri-Ray 做初步分流（文件预览/渲染属于 desktop UI 部分）。cc @lefarcen

lefarcen:
Routing to @nettee for first-pass triage (daemon file parsing). cc @lefarcen

lefarcen:
@shangxinyu1 你之前提的这个问题应该被刚合并的 PR #261 修了。请 `git pull && pnpm install && pnpm dev:all` 后重试，看看是否还能复现。如果还有问题，在这条 issue 下回复就行，我们继续看。感谢反馈！

PR #261, opened 2026-05-02
fix(daemon): treat .py files as previewable code in Design Files

Sid-Qin:
## Summary

Python (`.py`) files in Design Files currently fall back to `BinaryViewer` and cannot be previewed inline, even though they are plain text and the `'code'` viewer (`TextViewer`) is already wired up for similar source files.

`Closes #61`.

## Root cause

[`kindFor()`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/projects.ts#L278) maps a filename to a coarse \`kind\` bucket the frontend uses to pick a viewer. The \`'code'\` branch covers \`.js / .mjs / .cjs / .ts / .tsx / .json / .css\` — \`.py\` is not in the list, so Python files fall through to \`'binary'\` and [`FileViewer.tsx`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/FileViewer.tsx) renders them with `BinaryViewer` (download / open externally only).

The frontend already renders \`'code'\` via \`TextViewer\` (\`file.kind === 'text' || file.kind === 'code'\`), so once a file lands in the \`'code'\` bucket it gets the existing read-only code preview for free — no frontend change is needed.

## Fix

```diff
-  if (['.js', '.mjs', '.cjs', '.ts', '.tsx', '.json', '.css'].includes(ext)) {
+  if (['.js', '.mjs', '.cjs', '.ts', '.tsx', '.json', '.css', '.py'].includes(ext)) {
     return 'code';
   }
```

## Scope

Kept intentionally minimal to match the issue title (`.py` only). Other commonly-requested text-like extensions are easy follow-ups if you want them as part of this PR or a later one — happy to extend either way:

- shell / config: `.sh`, `.yaml`, `.yml`, `.toml`, `.env.example`, `.ini`
- other languages: `.rb`, `.go`, `.java`, `.kt`, `.rs`, `.c`, `.cpp`, `.lua`, `.php`, `.swift`

I left those out so the diff stays single-concern; ping me if you'd prefer them bundled.

## Verification

- Single-line array extension; all callers (\`projects.ts\` itself, \`media.ts\`, \`document-preview.ts\`) only branch on the \`kind\` returned, and none gate on \`'code'\` in a way that would regress (e.g. \`document-preview.ts\` only handles \`pdf / document / presentation / spreadsheet\`).
- Frontend already handles \`'code'\` via \`TextViewer\` — no UI changes required.
[...]

PR #947, opened 2026-05-08
fix: serve python files as text

leprincep35700:
## Summary
- Serve `.py` files as text from the daemon MIME map
- Add regression coverage for lowercase and uppercase Python extensions

## Why
Python files are classified as code elsewhere in the project, but `mimeFor()` fell back to `application/octet-stream`. Serving them as text keeps previews and direct file responses consistent with code-file handling.

## Testing
- `corepack pnpm install`
- `corepack pnpm --filter @open-design/daemon test -- project-classifiers`
- `corepack pnpm --filter @open-design/daemon typecheck`

Note: local environment is Node v20.19.2 while the repo declares Node ~24, but the targeted suite and typecheck passed.

### DIFF
--- a/apps/daemon/src/projects.ts
+++ b/apps/daemon/src/projects.ts
@@ -566,6 +566,7 @@ const EXT_MIME = {
   '.cjs': 'text/javascript; charset=utf-8',
   '.jsx': 'text/javascript; charset=utf-8',
   '.ts': 'text/typescript; charset=utf-8',
+  '.py': 'text/x-python; charset=utf-8',
   // `.tsx` previously served as `text/typescript`, which browser module
   // loaders and strict CSPs do not accept as a JavaScript MIME. Multi-file
   // React prototypes that load `.tsx` via Babel-standalone (`<script
--- a/apps/daemon/tests/project-classifiers.test.ts
+++ b/apps/daemon/tests/project-classifiers.test.ts
@@ -105,6 +105,7 @@ describe('mimeFor', () => {
     expect(mimeFor('a.jsx')).toBe('text/javascript; charset=utf-8');
     expect(mimeFor('a.tsx')).toBe('text/javascript; charset=utf-8');
     expect(mimeFor('a.ts')).toBe('text/typescript; charset=utf-8');
+    expect(mimeFor('a.py')).toBe('text/x-python; charset=utf-8');
     expect(mimeFor('a.json')).toBe('application/json; charset=utf-8');
     expect(mimeFor('a.md')).toBe('text/markdown; charset=utf-8');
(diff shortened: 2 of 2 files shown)


lefarcen:
Hi @leprincep35700! 🎉
Thanks for the contribution — nice catch on the Python file MIME type inconsistency.

— open-design team

lefarcen:
## Related context

This PR completes the Python file support work started in PR #261 (which fixed issues #61 and #62).

**Background:**
- #61/#62: Python files couldn't be previewed inline in Design Files or main workspace
- PR #261 (merged 2026-05-02): Added `.py` to `kindFor()` so the frontend could choose the text viewer
- **This PR**: Adds `.py` to `mimeFor()` MIME map so the daemon serves it as `text/x-python` instead of `application/octet-stream`

The two changes work together:
- `kindFor()` controls which **viewer component** the frontend picks (text vs binary)
- `mimeFor()` controls the **HTTP Content-Type** header when serving files directly

Without this MIME fix, raw file responses would still serve Python files with an octet-stream header, which breaks browser text rendering and direct preview links.

**Confidence:** High — file path overlap with #261 + same symptom domain (Python preview) + this PR explicitly references the MIME gap in the PR body.

lefarcen (commented):
Hi @leprincep35700! Nice work on this fix.

## Review summary

The change is correct and completes the Python file preview support:
- ✅ Adds `.py` to `EXT_MIME` map with `text/x-python; charset=utf-8` — industry-standard MIME type
- ✅ Test coverage for both lowercase (`.py`) and uppercase (`.PY`) cases
- ✅ Consistent with existing patterns (`.ts`, `.jsx`, `.tsx` mapping)
- ✅ Works with PR #261's `kindFor()` change to provide full Python preview support

## What I checked

**Lens A (code correctness):**
- Read full `projects.ts` and verified `mimeFor()` callers in `server.ts`, `media.ts`, `mcp.ts`
- Confirmed `.py` is already in the `'code'` bucket via `kindFor()` (PR #261)
- Verified `text/*` MIME checks will accept `text/x-python` for textual handling
- Test regression coverage is appropriate — two new assertions match the existing test structure

**No findings.** The change is minimal, well-scoped, and fills a real gap (Python files previously fell back to `application/octet-stream`).

mrcfps (approved):
@leprincep35700 I reviewed the `.py` MIME mapping and the new classifier coverage in `apps/daemon/src/projects.ts` and `apps/daemon/tests/project-classifiers.test.ts`. The change lines up with the existing text handling paths, and the lowercase/uppercase regression checks cover the behavior this PR is fixing. Thanks for tightening up Python preview support here. 🙌

PR #948, opened 2026-05-08
test: cover model option rendering

leprincep35700:
## Summary
- Add focused regression coverage for `renderModelOptions`
- Cover flat options, provider optgroups, default/custom entries, leading-slash IDs, label prefix stripping, and `isCustomModel`

## Why
The model picker has nuanced rendering behavior. These tests protect that behavior without changing production code.

## Testing
- `corepack pnpm install`
- `corepack pnpm --filter @open-design/web test -- modelOptions`

Note: local environment is Node v20.19.2 while the repo declares Node ~24, but the targeted suite and typecheck passed.

### DIFF
--- a/apps/web/tests/components/modelOptions.test.tsx
+++ b/apps/web/tests/components/modelOptions.test.tsx
@@ -1,0 +1,72 @@
+import { renderToStaticMarkup } from 'react-dom/server';
+import { describe, expect, it } from 'vitest';
+
+import {
+  CUSTOM_MODEL_SENTINEL,
+  isCustomModel,
+  renderModelOptions,
+} from '../../src/components/modelOptions';
+import type { AgentModelOption } from '../../src/types';
+
+function renderOptions(models: AgentModelOption[]): string {
+  return renderToStaticMarkup(<select>{renderModelOptions(models)}</select>);
+}
+
+describe('renderModelOptions', () => {
+  it('renders flat model lists as ungrouped options in input order', () => {
+    expect(
+      renderOptions([
+        { id: 'default', label: 'Default' },
+        { id: 'sonnet', label: 'Claude Sonnet' },
+        { id: 'opus', label: 'Claude Opus' },
+      ]),
+    ).toBe(
+      '<select><option value="default">Default</option><option value="sonnet">Claude Sonnet</option><option value="opus">Claude Opus</option></select>',
+    );
+  });
+
+  it('pins default and other flat options before provider optgroups', () => {
+    expect(
+      renderOptions([
+        { id: 'openai/gpt-5.1', label: 'openai/gpt-5.1' },
+        { id: 'default', label: 'Default' },
+        { id: 'anthropic/claude-sonnet-4.5', label: 'anthropic/claude-sonnet-4.5' },
+        { id: 'custom-local', label: 'Custom local' },
+        { id: 'openai/o3', label: 'openai/o3' },
+      ]),
+    ).toBe(
+      '<select><option value="default">Default</option><option value="custom-local">Custom local</option><optgroup label="openai"><option value="openai/gpt-5.1">gpt-5.1</option><option value="openai/o3">o3</option></optgroup><optgroup label="anthropic"><option value="anthropic/claude-sonnet-4.5">claude-sonnet-4.5</option></optgroup></select>',
+    );
+  });
+
+  it('treats leading-slash ids as flat and only strips matching provider label prefixes', () => {
+    expect(
+      renderOptions([
+        { id: '/missing-provider', label: '/missing-provider' },
+        { id: 'openai/gpt-5.1', label: 'GPT 5.1' },
+        { id: 'openai/o3', label: 'openai/o3' },
+      ]),
+    ).toBe(
+      '<select><option value="/missing-provider">/missing-provider</option><optgroup label="openai"><option value="openai/gpt-5.1">GPT 5.1</option><option value="openai/o3">o3</option></optgroup></select>',
+    );
+  });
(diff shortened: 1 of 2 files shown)


lefarcen:
Hi @leprincep35700! 🎉
Thanks for the contribution — adding comprehensive test coverage for the model options rendering logic is valuable.

— open-design team

apps/web/tests/components/modelOptions.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+        { id: 'default', label: 'Default' },
+        { id: 'sonnet', label: 'Claude Sonnet' },
+        { id: 'opus', label: 'Claude Opus' },
+      ]),
+    ).toBe(
+      '<select><option value="default">Default</option><option value="sonnet">Claude Sonnet</option><option value="opus">Claude Opus</option></select>',
+    );
+  });
+
+  it('pins default and other flat options before provider optgroups', () => {
```
lefarcen, line 32: **P2** — The test name says "pins default...before provider optgroups," but the fixture doesn't actually prove `default` is pinned ahead of other flat options. In your input, `custom-local` appears *after* `default`, so the expected output would still pass if `renderModelOptions` merely preserved flat-option insertion order.

Given the production code's comment says `default` is always pinned first, **add a flat option *before* `default` in the input** (e.g. move `custom-local` to the top), and verify `default` still appears first in the output. Otherwise, rename the test to "preserves flat option insertion order" to avoid asserting behavior it doesn't cover.

apps/web/tests/components/modelOptions.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  });
+});
+
+describe('isCustomModel', () => {
+  const models: AgentModelOption[] = [
+    { id: 'default', label: 'Default' },
+    { id: 'openai/gpt-5.1', label: 'openai/gpt-5.1' },
+  ];
+
+  it('returns false for empty selections and listed model ids', () => {
```
lefarcen, line 65: **P3** — The test says "empty selections" but only covers `null` and `undefined`. It misses the **empty string** `''`, which is the common DOM `<select>` empty value.

Consider adding `expect(isCustomModel('', models)).toBe(false)` if empty string should be treated as no selection, or rename the test to "nullish selections" if the current behavior is intentional.

apps/web/tests/components/modelOptions.test.tsx
```diff
@@ -0,0 +1,72 @@
...
+  isCustomModel,
+  renderModelOptions,
+} from '../../src/components/modelOptions';
+import type { AgentModelOption } from '../../src/types';
+
+function renderOptions(models: AgentModelOption[]): string {
+  return renderToStaticMarkup(<select>{renderModelOptions(models)}</select>);
+}
+
+describe('renderModelOptions', () => {
```
lefarcen, line 15: **P3** — `renderModelOptions` has no **empty-list coverage**. That's a simple edge case for a renderer helper and would document the expected empty output.

Consider adding `expect(renderOptions([])).toBe('<select></select>')` to cover this scenario.

lefarcen (commented):
Hi @leprincep35700! Thanks for the comprehensive test coverage. I reviewed the test assertions against the actual `renderModelOptions` and `isCustomModel` implementation. The tests are well-structured and pass correctly, but I found a few opportunities to strengthen the coverage and clarify intent:

**Summary:** 3 findings (all non-blocking suggestions to improve test robustness)

mrcfps (dismissed):
@leprincep35700 I reviewed the added regression coverage against `apps/web/src/components/modelOptions.tsx` and confirmed the new cases exercise the current flat-option ordering, provider optgroup rendering, label prefix stripping, leading-slash IDs, and `isCustomModel` behavior. Thanks for tightening up this model-picker surface — the focused coverage here should make future refactors much safer.


leprincep35700:
Addressed the review suggestions in `97c3816`:

- Added empty-list coverage for `renderModelOptions([])`.
- Strengthened the default-pinning test by placing a flat custom option before `default` in the fixture.
- Updated `renderModelOptions` to actually pin `default` before other flat options, matching the production comment.
- Added explicit empty-string coverage for `isCustomModel(\"\", models)` and treat it as no selection.

Validation rerun:
- `corepack pnpm --filter @open-design/web test -- modelOptions`

lefarcen (approved):
Excellent work, @leprincep35700! 🎉

You not only addressed all three review suggestions but also improved the production code to match the documented behavior:

1. ✅ Added empty-list coverage
2. ✅ Strengthened the default-pinning test with proper fixture order
3. ✅ **Implemented actual default-pinning logic** in `renderModelOptions` (line 24)
4. ✅ Added empty-string coverage
5. ✅ **Fixed `isCustomModel` to treat empty string as falsy** — consistent with DOM `<select>` behavior

The production code changes are safe and correct. The sort logic correctly pins `default` first while preserving relative order of other flat options, and the empty-string fix aligns with standard form behavior.

Looks good to me; deferring final merge decision to a maintainer.

mrcfps (approved):
@leprincep35700 I reviewed the updated `apps/web/src/components/modelOptions.tsx` behavior against the new `apps/web/tests/components/modelOptions.test.tsx` coverage and the current callers in `SettingsDialog` and `AvatarMenu`. The diff now actually pins the `default` flat option ahead of provider groups, keeps the custom-model empty-string path aligned with the settings select flow, and adds focused regression coverage for the empty-list, grouping, label-stripping, leading-slash, and custom-id cases. Thanks for tightening up this model-picker path — this looks solid and should make future refactors much safer.

Issue #383, opened 2026-05-03
[Windows] Next.js dev mode 404 + daemon crash on browser open


Sapientropic:
## Environment

- **OS:** Windows 11 Pro (10.0.26100), native (not WSL2)
- **Node:** v24.12.0
- **pnpm:** 10.33.2
- **Next.js:** 16.2.4 (Turbopack)
- **Commit:** 4de93fe

## Bug 1: `pnpm tools-dev run web` returns 404

Next.js 16.2.4 dev mode with Turbopack on Windows native fails to match the `[[...slug]]` catch-all route at `/`. The response body contains both the 404 error page *and* the actual app shell (title="Open Design" + all React chunks), but the browser renders the 404.

```bash
$ pnpm tools-dev run web
# Web: http://127.0.0.1:62969/
# Daemon: http://127.0.0.1:55054/

$ curl -s http://127.0.0.1:62969/ | grep '<title>'
<title>404: This page could not be found.</title>
<title>Open Design</title>
```

Both titles appear in the same response — Next.js renders the not-found boundary instead of the catch-all page, but the app layout is still present underneath.

**Workaround:** Build the static export and run daemon in production mode (see Bug 2 workaround).

## Bug 2: Daemon production mode crashes with `spawn start ENOENT`

When running the daemon with `OD_WEB_PROD=1`, it tries to open the browser using `spawn('start', [url])` on Windows. This fails with an unhandled `'error'` event that kills the process entirely:

```
$ OD_WEB_PROD=1 node apps/daemon/dist/cli.js
[od] daemon listening on http://127.0.0.1:7456
[od] listening on http://127.0.0.1:7456
node:events:486
      throw er; // Unhandled 'error' event
      ^
Error: spawn start ENOENT
    at ChildProcess._handle.onexit (node:internal/child_process:286:19)
    ...
  errno: -4058,
  code: 'ENOENT',
  syscall: 'spawn start',
  path: 'start',
  spawnargs: [ 'http://127.0.0.1:7456' ]
```

**Workaround:** Use `--no-open` flag:
```bash
OD_WEB_PROD=1 node apps/daemon/dist/cli.js --no-open
```

This works correctly — daemon serves the static build at `http://127.0.0.1:7456/` and returns 200 with the proper app.

## Full workaround for Windows native

```bash
pnpm install
pnpm build
pnpm --filter @open-design/daemon build
OD_WEB_PROD=1 node apps/daemon/dist/cli.js --no-open
[...]

Issue #573, opened 2026-05-05
OpenDesign 0.4.0 is not running in windows


AceVentura:
I've tried a clean install of the 0.4.0 version (just released) in Windows 11, but it's not doing anything. It does show up for a few seconds in Task Manager, but that's it.
I've checked Event Viewer, but I can't see any logs related with it. Is there anything else I can try to figure out the issue?

Version 0.3.0 worked.

<img width="948" height="721" alt="Image" src="https://github.com/user-attachments/assets/03551057-eca0-4e43-8a8d-053745fe5b47" />

nilabhragit90:
Same thing is happening with me

lefarcen:
Hi @AceVentura! 👋

Thanks for the report — you hit a **Windows startup crash** in 0.4.0 that prevents the daemon from launching. The good news: the root cause is identified and there's an open PR with the fix.

## What's happening

The daemon tries to open your browser with:
```javascript
spawn('start', [url])  // apps/daemon/src/cli.ts:120
```

But on Windows, `start` is a `cmd.exe` builtin, not a standalone executable → `ENOENT` → daemon crashes before the web server even starts. That's why you see the process appear briefly in Task Manager and then vanish.

## The fix

**PR #331** changes it to:
```javascript
spawn('cmd', ['/c', 'start', url])
```

This PR is open and tested on Windows 11. It shipped **after** the 0.4.0 freeze, so 0.4.0 doesn't include it yet.

## Workaround until the fix merges

Run the daemon with `--no-open`:
```bash
open-design.exe --no-open
```
Then manually open `http://127.0.0.1:7456` in your browser. The daemon itself works fine — it's only the auto-open that crashes.

## Next steps

I'll link this issue to PR #331 so you're notified when it merges. Once that lands (likely in 0.4.1), a clean reinstall should work out of the box.

Thanks for making Open Design better! 🚀

AceVentura:
@lefarcen thanks for the quick reply. I've tried the workaround, but still not working. 

<img width="1496" height="917" alt="Image" src="https://github.com/user-attachments/assets/3f613fc2-4b40-4040-8461-5ad9c226f5ba" />

lefarcen:
Thanks for trying the workaround — this is useful. `--no-open` did bypass the confirmed Windows `spawn('start')` crash, but the browser failure is a separate symptom.

[...]

PR #339, opened 2026-05-03
fix(platform): wrap cmd.exe shim invocations to survive /s /c quote stripping

Sid-Qin:
## Summary

Fixes [#315](https://github.com/nexu-io/open-design/issues/315) — on Windows, when an agent CLI is installed at a path containing a space (default for many user accounts) and the binary is a `.CMD` shim (e.g. npm-installed Codex CLI), the daemon spawns it with mangled quoting and the agent fails immediately with `'Ethical' is not recognized as an internal or external command`.

This is a quoting regression in PR #258, which standardized agent spawning through `createCommandInvocation` and routed Windows `.cmd` / `.bat` paths through `cmd.exe /d /s /c <line>`. PR #232's earlier `shell:true` fix didn't survive that refactor, and the new shape has its own quoting bug on argv-style spawn.

## Root cause

Two layers of quoting collide:

1. **`cmd.exe /s /c`** strips exactly one leading and one trailing `"` from the rest of the command line.
2. **Node's default Windows arg escaping** (CommandLineToArgvW rules) turns `"path with space"` into `\"path with space\"` for each argv element. cmd.exe doesn't understand backslash-escaped quotes.

So `[\"/d\", \"/s\", \"/c\", '\"C:\\\\Users\\\\Ethical Byte\\\\...\\\\codex.CMD\" --help']` → cmd.exe sees the `\"` Node wrote, fails to parse the quoting, and after `/s` strips the visible outer `\"`, what's left is `C:\\Users\\Ethical Byte\\...\\codex.CMD --help` with no quotes at all. The first space wins, and `Ethical` becomes the command.

## Fix

Mirror what Node's own `child_process.spawn({ shell: true })` does internally:

1. Wrap the entire joined command line in an extra `\"…\"` so cmd.exe's `/s /c` strip eats the outer pair and leaves inner per-arg quoting intact.
2. Set `windowsVerbatimArguments: true` so Node forwards argv to CreateProcess unchanged, without re-applying CommandLineToArgvW escaping.

This is the only shape that survives both layers.

## Changes

- **`packages/platform/src/index.ts`** — extend `CommandInvocation` with optional `windowsVerbatimArguments`; extract `buildCmdShimInvocation` helper that applies the outer wrap + verbatim flag in both `createCommandInvocation` and `createPackageManagerInvocation`; forward the flag through `spawnBackgroundProcess` / `spawnLoggedProcess`.
[...]

PR #331, opened 2026-05-02
fix(windows): daemon startup crash and tools-dev hang on Windows

joseperezlara:
## Problem

  Two bugs that prevent Open Design from starting at all on Windows:

  ### 1. Daemon crashes immediately on startup (`apps/daemon/src/cli.ts`)

  `spawn('start', [url])` throws `ENOENT` because `start` is a `cmd.exe`
  built-in, not a standalone executable. The daemon exits before the web
  server even starts.

  ### 2. `pnpm tools-dev run web` hangs forever (`packages/platform/src/index.ts`)

  `listWindowsProcessSnapshots()` runs `Get-CimInstance Win32_Process` via
  PowerShell with no timeout. On some Windows systems WMI is slow or
  unresponsive, so `tools-dev` hangs indefinitely with no output or error.

  ## Fix

  1. Use `spawn('cmd', ['/c', 'start', url])` on Windows.
  2. Add a 5 s timeout to the PowerShell child process — `listProcessSnapshots()`
     already catches rejections and returns `[]`, so tools-dev continues normally.

  ## Testing

  Reproduced and fixed on Windows 11 (build 26200). Both bugs are
  hit on a clean install before any design is generated.

PR #953, opened 2026-05-08
fix daemon browser opener on Windows

Sapientropic:
## Summary

- Add a small daemon browser opener helper with platform-specific invocation details.
- Open Windows URLs through `cmd.exe /d /s /c start "" "<url>"` instead of spawning the `start` shell builtin directly.
- Attach an opener `error` handler so a missing platform opener warns instead of crashing the already-started daemon.

## Why

Refs #383.

The daemon CLI currently uses `spawn('start', [url])` on Windows. `start` is a `cmd.exe` builtin, not a standalone executable, so native Windows throws `spawn start ENOENT`. Because the detached child had no `error` listener, that best-effort browser-open failure could terminate the daemon after it had already started serving.

I updated my local checkout to latest `main` first. The Next.js dev-mode 404 described in #383 did not reproduce on latest `main` in my Windows native smoke (`/` returned 200 with only the Open Design title), so this PR is scoped to the still-relevant daemon browser opener crash.

## Validation

- `pnpm install`
- `pnpm tools-dev run web --daemon-port 17601 --web-port 17602`, then fetched `http://127.0.0.1:17602/`: HTTP 200, `<title>Open Design</title>`, no Next 404 title
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/browser-open.test.ts`
- `pnpm --filter @open-design/desktop build` followed by `pnpm typecheck`
- Production smoke on Windows native: `OD_WEB_PROD=1 node apps/daemon/dist/cli.js --port 17604` without `--no-open`; daemon stayed alive and `http://127.0.0.1:17604/` returned HTTP 200 with the Open Design title

## Known unrelated local test failures

`pnpm --filter @open-design/daemon test` still fails on latest `main` in this Windows native checkout outside this PR's opener path. The failures I saw are in existing agent executable override tests, Windows path normalization expectations in `server-paths.test.ts` / `mcp-config.test.ts`, two folder-import symlink status-code expectations, and one chat-route stalled-stream expectation. The focused opener regression test, daemon typecheck, daemon build, and root `pnpm typecheck` pass after building `@open-design/desktop`.

### DIFF
--- a/apps/daemon/src/browser-open.ts
+++ b/apps/daemon/src/browser-open.ts
@@ -1,0 +1,78 @@
+import { spawn as nodeSpawn } from 'node:child_process';
+import type { ChildProcess, SpawnOptions } from 'node:child_process';
+
+type SupportedPlatform = NodeJS.Platform;
+
+export type BrowserOpenInvocation = {
+  command: string;
+  args: string[];
+  options: SpawnOptions;
+};
+
+type OpenBrowserDeps = {
+  platform?: SupportedPlatform;
+  spawn?: (command: string, args: string[], options: SpawnOptions) => ChildProcess;
+  warn?: (message: string) => void;
+  env?: NodeJS.ProcessEnv;
+};
+
+function quoteWindowsCommandArg(value: string, { force = false }: { force?: boolean } = {}): string {
+  if (value.length === 0) return '""';
+  if (!force && !/[\s"&<>|^%]/.test(value)) return value;
+  const escaped = value.replace(/"/g, '""').replace(/%/g, '"^%"');
+  return `"${escaped}"`;
+}
+
+export function createBrowserOpenInvocation(
+  platform: SupportedPlatform,
+  url: string,
+  env: NodeJS.ProcessEnv = process.env,
+): BrowserOpenInvocation {
+  if (platform === 'win32') {
+    const comspec = env.ComSpec || env.COMSPEC || 'cmd.exe';
+    // `start` is a cmd.exe builtin on Windows, not a real executable. The empty
+    // title argument keeps cmd from treating the URL itself as the window title.
+    // Match @open-design/platform's cmd.exe shim shape: Node's default Windows
+    // argv quoting uses backslash escapes that cmd.exe does not understand, so
+    // the inner command must be wrapped for `/s /c` and passed verbatim.
+    const inner = [
+      'start',
+      quoteWindowsCommandArg(''),
+      quoteWindowsCommandArg(url, { force: true }),
+    ].join(' ');
+    return {
+      command: comspec,
+      args: ['/d', '/s', '/c', `"${inner}"`],
+      options: { detached: true, stdio: 'ignore', windowsHide: true, windowsVerbatimArguments: true },
+    };
+  }
+
+  return {
+    command: platform === 'darwin' ? 'open' : 'xdg-open',
+    args: [url],
+    options: { detached: true, stdio: 'ignore' },
+  };
+}
+
+export function openBrowser(url: string, deps: OpenBrowserDeps = {}): ChildProcess | null {
+  const platform = deps.platform ?? process.platform;
+  const spawn = deps.spawn ?? nodeSpawn;
+  const warn = deps.warn ?? ((message: string) => console.warn(message));
+  const invocation = createBrowserOpenInvocation(platform, url, deps.env);
+
+  try {
+    const child = spawn(invocation.command, invocation.args, invocation.options);
+    // Browser opening is best-effort. A missing opener must not crash the daemon
+    // after the server has already started and printed its URL.
+    child.on('error', (error) => {
+      const detail = error instanceof Error ? error.message : String(error);
+      warn(`[od] failed to open browser: ${detail}`);
+    });
+    child.unref();
+    return child;
+  } catch (error) {
+    const detail = error instanceof Error ? error.message : String(error);
+    warn(`[od] failed to open browser: ${detail}`);
+    return null;
+  }
+}
--- a/apps/daemon/src/cli.ts
+++ b/apps/daemon/src/cli.ts
@@ -5,6 +5,7 @@ import { runLiveArtifactsMcpServer } from './mcp-live-artifacts-server.js';
 import { runConnectorsToolCli } from './tools-connectors-cli.js';
 import { runLiveArtifactsToolCli } from './tools-live-artifacts-cli.js';
 import { splitResearchSubcommand } from './research/cli-args.js';
+import { openBrowser } from './browser-open.js';
 
 const argv = process.argv.slice(2);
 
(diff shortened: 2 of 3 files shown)


Sapientropic:
Related prior work: #331 already covers the same Windows `spawn('start')` opener root cause as part of a larger Windows startup PR.

This PR is intentionally narrower:

- it only addresses the daemon browser-open crash path from #383 / #573;
- it is rebased on current `main` after the desktop/runtime churn;
- it adds a focused regression test for Windows `cmd.exe` wrapping and opener `error` handling;
- it leaves the WMI/process-snapshot timeout concern from #331 out of scope because that path has separate review feedback and CI history.

Happy to close this if maintainers prefer to land #331 wholesale, but this should be a smaller alternative for the opener crash specifically.

lefarcen:
Hi @Sapientropic! 🎉

Thanks for the contribution — this looks like a solid fix for the Windows daemon browser opener crash issue. The error-handler addition and the proper cmd.exe invocation through shell are the right approach.

— open-design team

apps/daemon/src/browser-open.ts
```diff
@@ -0,0 +1,67 @@
...
+  url: string,
+  env: NodeJS.ProcessEnv = process.env,
+): BrowserOpenInvocation {
+  if (platform === 'win32') {
+    const comspec = env.ComSpec || env.COMSPEC || 'cmd.exe';
+    // `start` is a cmd.exe builtin on Windows, not a real executable. The empty
+    // title argument keeps cmd from treating the URL itself as the window title.
+    return {
+      command: comspec,
+      args: ['/d', '/s', '/c', `start "" ${quoteForCmd(url)}`],
```
lefarcen: **P1** — Windows command quoting mismatch. This builds a `cmd.exe /d /s /c` command string with embedded quotes, but the spawn options do not set `windowsVerbatimArguments: true`. Node/libuv will backslash-escape the embedded quotes in the final arg (producing something like `\"\"`/`\"url\"`), but `cmd.exe` doesn't use backslash as quote escaping — it only recognizes doubled quotes (`""`). So `start` can receive malformed arguments.

Reuse/export the platform helper from `packages/platform/src/index.ts` (which already handles this correctly with `windowsVerbatimArguments: true` + outer quotes), or mirror its `/s /c \"<inner>\"` approach. Then assert that option in the test.

lefarcen (changes requested):
Hey @Sapientropic! Thanks for tackling the Windows daemon crash. The approach is solid — moving browser opening to a helper and adding error handling — but there's a critical Windows command-line quoting issue that needs fixing.

## The problem

The Windows `cmd.exe /d /s /c` invocation you've built doesn't match Node/Windows quoting semantics. When Node.js spawns a process on Windows without `windowsVerbatimArguments: true`, it backslash-escapes embedded quotes. But `cmd.exe` doesn't use backslash as a quote escape character — it only recognizes doubled quotes (`""`). So the current code can produce malformed command lines that `start` won't parse correctly.

This repo's existing Windows `cmd.exe` helper in `packages/platform/src/index.ts` already handles this correctly with `windowsVerbatimArguments: true` + outer quotes. You should reuse that pattern (or export that helper) rather than rebuilding it.

## What to do

1. Check `packages/platform/src/index.ts` for the `cmd.exe /s /c` helper (it's in the Windows agent spawn path)
2. Either:
   - Export that helper and call it from your `browser-open.ts`, or
   - Mirror its `windowsVerbatimArguments: true` + `/s /c \"<inner>\"` approach
3. Add a test assertion that the Windows `options` include `windowsVerbatimArguments: true`

See the inline comment for the specific line where this needs fixing.

lefarcen:
## Related context

**Concurrent PR:** #331 by @joseperezlara is also OPEN and touches `apps/daemon/src/cli.ts` with a similar Windows daemon fix approach. The two PRs have overlapping scope — you might want to coordinate to avoid merge conflicts or redundant fixes.

**Windows cmd.exe quoting precedent:** This repo has existing Windows command-line helpers in `packages/platform/src/index.ts` that handle the `windowsVerbatimArguments` + outer-quote pattern correctly (see PR #339, PR #232 for the history). Your fix should follow that established pattern rather than rebuilding it.

apps/daemon/src/browser-open.ts
```diff
@@ -0,0 +1,67 @@
...
+      options: { detached: true, stdio: 'ignore', windowsHide: true },
```
mrcfps: This still misses the Windows `cmd.exe` quoting shape we already rely on in `packages/platform/src/index.ts:173-193`. Here the inner command is passed as `start "" ${quoteForCmd(url)}`, but the spawn options on line 35 do not set `windowsVerbatimArguments: true`. On Windows, Node re-quotes that argv element before launching `cmd.exe`, and `cmd.exe /s /c` does not treat those backslash-escaped quotes as grouping, so the opener can still misparse/fail even though the direct `spawn('start')` ENOENT path is gone. Please mirror or reuse the existing helper pattern here: wrap the full inner command in outer quotes, set `windowsVerbatimArguments: true`, and extend `apps/daemon/tests/browser-open.test.ts` to assert that flag so this Windows regression stays covered.

mrcfps (changes requested):
@Sapientropic Thanks for the Windows opener fix and the focused regression coverage. I found one blocking issue in the Windows `cmd.exe` invocation that still needs to be aligned with the repo's existing verbatim-argument pattern; details are inline.


lefarcen (approved):
The Windows cmd.exe quoting issue is fixed! ✅

You added `windowsVerbatimArguments: true` and properly wrapped the inner command with outer quotes (`/s /c "start ..."`), which matches the repo's established pattern in `packages/platform/src/index.ts`. The `quoteWindowsCommandArg` helper correctly escapes cmd.exe metacharacters (`%` → `"^%"`, `"` → `""`).

The regression tests now assert both the `windowsVerbatimArguments` option and the metacharacter escaping, so this won't regress.

Nice work addressing the feedback so quickly — the fix looks solid. Deferring final merge approval to a maintainer.

mrcfps (approved):
@Sapientropic I reviewed the Windows opener changes in `apps/daemon/src/browser-open.ts`, `apps/daemon/src/cli.ts`, and `apps/daemon/tests/browser-open.test.ts`. The new helper now follows the repo's existing `cmd.exe /d /s /c` verbatim-argument pattern, keeps browser-open failures best-effort instead of crashing the daemon, and adds focused regression coverage for the Windows quoting path. Thanks for tightening this up and adding the targeted test coverage. 🙌


leilei926524-tech:
Hi @Sapientropic — welcome to Open Design contributors. 🎉

Your first merged PR (#953) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #958, opened 2026-05-08
Fix Cloudflare Pages custom domain lookup

bulai0408:
## Summary
- replace Cloudflare Pages custom-domain pagination lookup with direct hostname lookup
- keep 404 as "not yet bound" so custom-domain creation can continue
- update daemon route/unit regressions to ensure the deploy path no longer sends unsupported `page` / `per_page` query options for Pages domains

## Validation
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/deploy.test.ts tests/deploy-routes.test.ts`
- `pnpm guard`
- `pnpm typecheck`


mrcfps (approved):
@bulai0408 I verified the Cloudflare Pages custom-domain path now uses the per-hostname lookup, still treats a 404 as "not yet bound", and updates the daemon route/unit coverage around that flow. Thanks for tightening up this deploy regression and covering the behavior clearly.


lefarcen (commented):
Hey @bulai0408! 🎉

Reviewed the fix — the direct hostname lookup matches Cloudflare's documented endpoint shape, the 404→create fallback is appropriate, and the test coverage exercises the regression. Clean implementation.

PR #960, opened 2026-05-08
fix: prevent design system filter popover from shifting position on reopen

xxiaoxiong:
## Summary

Fixes #921 - Prevents the design system filter popover from repositioning incorrectly when reopened after filtering.

## Problem

The design system filter popover was appearing too high and becoming partially hidden at the top of the viewport when reopened after applying filters.

**Root cause:**
- The popover uses `position: absolute` with `top: calc(100% + 6px)`
- When filtering reduces the number of items, the popover height shrinks
- On reopen, the reduced height can cause the popover to appear higher than expected
- This is especially noticeable when the trigger button is near the top of the viewport

## Solution

Added `min-height: 120px` to `.ds-picker-list` in `apps/web/src/index.css`:

```css
.ds-picker-list {
  display: flex;
  flex-direction: column;
  min-height: 120px;  /* ← Added */
  max-height: 320px;
  overflow-y: auto;
  padding: 4px;
}
```

**Why this works:**
- Ensures the popover maintains a consistent minimum height
- Prevents position shifts when content is filtered
- The popover stays anchored correctly to its trigger
- The 120px minimum provides enough space for ~3-4 items

## Impact

This change affects all design system picker popovers:
1. ✅ Design system filter in NewProjectPanel
2. ✅ Design system filter in SettingsDialog
3. ✅ Prompt template picker (uses same CSS class)

All will benefit from more stable positioning across filter state changes.

## Testing

- [x] Verified CSS change is minimal and conservative
- [x] Confirmed the min-height value (120px) is reasonable
- [x] Checked all usages of `.ds-picker-list` class
- [x] Ensured backward compatibility

## Checklist

- [x] Code follows project style guidelines
- [x] Change is minimal and focused on the reported issue
- [x] No breaking changes
- [x] Commit message follows conventional commits format

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -4338,10 +4338,12 @@ code {
 .ds-picker-list {
   display: flex;
   flex-direction: column;
-  min-height: 120px;
   max-height: 320px;
   overflow-y: auto;
   padding: 4px;
+}
+.ds-picker-list-design-systems {
+  min-height: 120px;
 }
 .ds-picker-empty {
   padding: 16px 12px;
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -4338,6 +4338,7 @@ code {
 .ds-picker-list {
   display: flex;
   flex-direction: column;
+  min-height: 120px;
   max-height: 320px;
   overflow-y: auto;
   padding: 4px;
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -1430,7 +1430,7 @@ function DesignSystemPicker({
               </button>
             </div>
           </div>
-          <div className="ds-picker-list">
+          <div className="ds-picker-list ds-picker-list-design-systems">
             <DsPickerItem
               active={selectedIds.length === 0}
               multi={multi}


mrcfps (commented):
@xxiaoxiong Thanks for the targeted fix — I found one merge-safe follow-up around the shared picker styling.


apps/web/src/index.css
```diff
@@ -4338,6 +4338,7 @@ code {
 .ds-picker-list {
   display: flex;
   flex-direction: column;
+  min-height: 120px;
```
mrcfps: This selector is shared by more than the design-system filter this PR is trying to fix: `NewProjectPanel.tsx` also uses it for the prompt-template picker (`apps/web/src/components/NewProjectPanel.tsx:1056`), and `SettingsDialog.tsx` uses it for the MCP client picker (`apps/web/src/components/SettingsDialog.tsx:3781`). Adding a global `min-height: 120px` here means those unrelated popovers now always render with ~120px of body height even when they only have 2 options, which introduces extra blank space without being tied to the reported bug. Can we scope the minimum height to the design-system picker only (for example with a dedicated modifier class on that popover/list) so the fix does not change every `.ds-picker-list` consumer?

lefarcen:
Hi @xxiaoxiong! 🎉

Thanks for the contribution — this is a clean, focused fix for the popover positioning issue.

— open-design team

apps/web/src/index.css
lefarcen: **Minor clarification (P3):** The PR body mentions "~3-4 items" for the 120px minimum, but actual `.ds-picker-item` rows are roughly 48-50px tall (with padding/avatars), so 120px is closer to about 2 visible rows.

This doesn't affect the fix itself — 120px works fine as a minimum to prevent position jumps. Just noting the math for future reference if you update the PR description.

lefarcen (commented):
Hey @xxiaoxiong! 👋

I checked the `.ds-picker-list` usages across the codebase (NewProjectPanel, SettingsDialog, prompt template picker) — the `min-height: 120px` change looks safe. It won't affect scrolling since it's well below the `max-height: 320px`, and the fix correctly addresses the height-collapse issue you described.

One small observation below (non-blocking):

xxiaoxiong:
Thanks for the detailed review! You're absolutely right — applying `min-height` globally to `.ds-picker-list` would affect the prompt-template picker and MCP client picker unnecessarily.

I've updated the PR to:
1. Add a dedicated `.ds-picker-list-design-systems` modifier class
2. Apply the `min-height: 120px` only to the design-system picker
3. Keep other pickers unaffected

Now the fix is scoped to only the design-system filter popover that has the position-shift bug.

Regarding the P3 note about row height: you're right that 120px is closer to 2-3 visible rows. I'll update the PR description to reflect the accurate math. Thanks for catching that!

lefarcen (commented):
Perfect fix! 🎯

The scoped modifier class addresses mrcfps's concern completely:
- `.ds-picker-list-design-systems` now isolates the `min-height` to only the design-system picker
- Other pickers (prompt-template, MCP client) are unaffected
- The positioning fix stays targeted to the reported bug

mrcfps (approved):
@xxiaoxiong I reviewed the updated head and verified the new `.ds-picker-list-design-systems` modifier keeps the minimum-height behavior scoped to the design-system picker while leaving the prompt-template and MCP pickers on the shared `.ds-picker-list` styling. Thanks for tightening the fix quickly — this looks ready to merge.


lefarcen:
Merged! 🎉 Thanks @xxiaoxiong for the quick fix and responsive iteration. The scoped CSS approach is exactly right — keeps the fix minimal and safe.

Looking forward to the next contribution!

PR #965, opened 2026-05-08
fix: hide surface filter tabs with zero count in Design Systems view (#922)

### LINKED ISSUE #922: Hide empty surface tabs when no matching templates are returned
## Summary

In the Design systems view, surface tabs such as **Image**, **Video**, and **Audio** are still shown even when their counts are `0` and no matching templates exist.

## Current behavior

The UI renders all surface tabs regardless of whether the backend returns any matching templates for that surface. As a result, users see tabs like **Image 0**, **Video 0**, and **Audio 0**, which lead to empty states with no useful content.

## Expected behavior

Only show surface tabs that have matching templates returned by the API. If a surface has no matching templates, hide that tab instead of rendering a `0` count tab.

## Steps to reproduce

1. Open the **Design systems** view.
2. Look at the **Surface** filter row.
3. Observe that surfaces with `0` matching results are still rendered as selectable tabs.

## Notes

- A screenshot was provided showing **Image 0**, **Video 0**, and **Audio 0** still visible in the Surface row.
- The request is to determine visibility from the API response rather than always rendering every possible surface.
- This likely applies to other template-related views as well if they share the same surface-tab logic.

## Why this matters

Showing empty tabs adds noise and invites users into dead-end states. Hiding surfaces with no matching results would make the UI cleaner and help users focus on surfaces that actually have available content.

Image

nmsn:
Fixes #922

Hide surface filter tabs (Image, Video, Audio) when their count is 0, keeping only tabs that have matching templates. The "All" tab is always shown.

### DIFF
--- a/apps/web/src/components/DesignSystemsTab.tsx
+++ b/apps/web/src/components/DesignSystemsTab.tsx
@@ -63,6 +63,15 @@ export function DesignSystemsTab({ systems, selectedId, onSelect, onPreview }: P
     return counts;
   }, [systems]);
 
+  const categories = useMemo(() => {
+    const cats = new Set<string>();
+    for (const s of surfaceScoped) cats.add(s.category || 'Uncategorized');
+    const ordered: string[] = [];
+    for (const c of CATEGORY_ORDER) if (cats.has(c)) ordered.push(c);
+    for (const c of [...cats].sort()) if (!ordered.includes(c)) ordered.push(c);
+    return ['All', ...ordered];
+  }, [surfaceScoped]);
+
   // Keep surfaceFilter and category in sync when systems changes dynamically.
   // If the currently selected surface has zero items, fall back to 'all'.
   // If the current category is no longer present in the filtered list, fall back to 'All'.
@@ -74,15 +83,6 @@ export function DesignSystemsTab({ systems, selectedId, onSelect, onPreview }: P
       setCategory('All');
     }
   }, [systems, surfaceFilter, surfaceCounts, category, categories]);
-
-  const categories = useMemo(() => {
-    const cats = new Set<string>();
-    for (const s of surfaceScoped) cats.add(s.category || 'Uncategorized');
-    const ordered: string[] = [];
-    for (const c of CATEGORY_ORDER) if (cats.has(c)) ordered.push(c);
-    for (const c of [...cats].sort()) if (!ordered.includes(c)) ordered.push(c);
-    return ['All', ...ordered];
-  }, [surfaceScoped]);
 
   const filtered = useMemo(() => {
     const q = filter.trim().toLowerCase();
--- a/apps/web/src/components/DesignSystemsTab.tsx
+++ b/apps/web/src/components/DesignSystemsTab.tsx
@@ -131,7 +131,7 @@ export function DesignSystemsTab({ systems, selectedId, onSelect, onPreview }: P
         aria-label={t('ds.surfaceLabel')}
       >
         <span className="examples-filter-label">{t('ds.surfaceLabel')}</span>
-        {SURFACE_PILLS.map((p) => (
+        {SURFACE_PILLS.filter((p) => p.value === 'all' || surfaceCounts[p.value] > 0).map((p) => (
           <button
             key={p.value}
             type="button"
--- a/apps/web/src/components/DesignSystemsTab.tsx
+++ b/apps/web/src/components/DesignSystemsTab.tsx
@@ -63,6 +63,18 @@ export function DesignSystemsTab({ systems, selectedId, onSelect, onPreview }: P
     return counts;
   }, [systems]);
 
+  // Keep surfaceFilter and category in sync when systems changes dynamically.
+  // If the currently selected surface has zero items, fall back to 'all'.
+  // If the current category is no longer present in the filtered list, fall back to 'All'.
+  useEffect(() => {
+    if (surfaceFilter !== 'all' && surfaceCounts[surfaceFilter] === 0) {
+      setSurfaceFilter('all');
+      setCategory('All');
+    } else if (category !== 'All' && !categories.includes(category)) {
+      setCategory('All');
+    }
+  }, [systems, surfaceFilter, surfaceCounts, category, categories]);
+
   const categories = useMemo(() => {
     const cats = new Set<string>();
     for (const s of surfaceScoped) cats.add(s.category || 'Uncategorized');
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -1313,12 +1313,12 @@ function BoardComposerPopover({
       }}
     >
       <div className="comment-popover-head">
-        <div>
+        <div title={target.elementId}>
           <strong id={titleId}>{target.elementId}</strong>
           <span>{target.label}</span>
           <span>{selectionKindLabel(target.selectionKind, target.memberCount)}</span>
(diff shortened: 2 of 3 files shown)


apps/web/src/components/DesignSystemsTab.tsx
```diff
@@ -131,7 +131,7 @@ export function DesignSystemsTab({ systems, selectedId, onSelect, onPreview }: P
         aria-label={t('ds.surfaceLabel')}
       >
         <span className="examples-filter-label">{t('ds.surfaceLabel')}</span>
-        {SURFACE_PILLS.map((p) => (
+        {SURFACE_PILLS.filter((p) => p.value === 'all' || surfaceCounts[p.value] > 0).map((p) => (
```
mrcfps, line 146: If `systems` changes after the user has selected a non-`all` surface, this filter can remove the active tab without resetting `surfaceFilter`. In that state `surfaceScoped` still uses the stale value from lines 55-57, so the grid falls back to `ds.emptyNoMatch` even when other design systems still exist, and there is no visible selected tab for the user to click back out of. Please add a fallback when the current surface count drops to zero (for example, a `useEffect` that resets `surfaceFilter` and `category` back to `'all'`) so the rendered pills and the active filter state stay in sync.

mrcfps (commented):
Thanks for the fixes here — the overflow cleanup looks good, and I verified the new surface-pill filtering logic in `DesignSystemsTab`. I did find one state-sync case worth tightening before this ships.


lefarcen:
Hi @nmsn! 🎉

Thanks for the contribution — this looks like a clean fix for the zero-count surface filter tabs issue.

— open-design team

apps/web/src/components/DesignSystemsTab.tsx
lefarcen, line 146: **P2** — Filtering out zero-count surface pills can remove the currently selected `surfaceFilter` when `systems` changes dynamically. `surfaceScoped` still filters by the stale surface value, so the grid can remain empty while no visible surface tab is selected; `category` can also remain set to a category no longer valid for the new surface/system set.

Please add an effect keyed on `systems`, `surfaceFilter`, `surfaceCounts`, and `category` that resets `surfaceFilter` to `'all'` when its count drops to 0 and resets `category` to `'All'` when it is no longer present.

lefarcen (changes requested):
Hi @nmsn, thanks for tackling this! The zero-count filter logic looks good for the initial render, and the FileViewer overflow fixes prevent layout breakage.

However, I agree with @mrcfps on the state-sync issue: when `systems` changes after the user has selected a non-`all` surface, this filter can remove the active tab without resetting `surfaceFilter`. The grid then filters by a stale surface value with no visible selected tab.

Please add a `useEffect` (keyed on `systems`, `surfaceFilter`, `surfaceCounts`, and `category`) that resets `surfaceFilter` to `'all'` and `category` to `'All'` when the current surface count drops to zero, so the UI state and the visible pills stay in sync.

apps/web/src/components/DesignSystemsTab.tsx
```diff
@@ -63,6 +63,18 @@ export function DesignSystemsTab({ systems, selectedId, onSelect, onPreview }: P
...
+  // If the currently selected surface has zero items, fall back to 'all'.
+  // If the current category is no longer present in the filtered list, fall back to 'All'.
+  useEffect(() => {
+    if (surfaceFilter !== 'all' && surfaceCounts[surfaceFilter] === 0) {
+      setSurfaceFilter('all');
+      setCategory('All');
+    } else if (category !== 'All' && !categories.includes(category)) {
+      setCategory('All');
+    }
+  }, [systems, surfaceFilter, surfaceCounts, category, categories]);
```
mrcfps, line 85: `apps/web/src/components/DesignSystemsTab.tsx`: this new effect now references `categories` inside the body and dependency array before the `const categories = useMemo(...)` declaration below. Because the dependency array is evaluated during render, `categories` is still in the temporal dead zone here, so opening the Design Systems tab will throw a `ReferenceError` before the reset-to-`'all'` logic ever runs. Please move the `categories` memo above this effect, or compute the category-validity check from a value declared earlier, so the sync fix does not break the tab.

mrcfps (changes requested):
@nmsn thanks for following up on the earlier state-sync feedback. I reviewed the latest head and found one blocker in the new `DesignSystemsTab` sync effect: it reads `categories` before that memo is declared, so this tab will throw before the fallback logic can run.


apps/web/src/components/DesignSystemsTab.tsx
lefarcen, line 85: P1 — `categories` is referenced in the dependency array here, but it's not declared until line 78. This creates a Temporal Dead Zone error. Move the `categories` memo above this effect.

lefarcen (changes requested):
Hi @nmsn — thanks for following up on the state-sync issue. However, the TDZ (Temporal Dead Zone) blocker from mrcfps's last review is still present on this head.

The `useEffect` at line 69 (and its dependency array at line 76) reference `categories` before it's declared at line 78. Because React evaluates the dependency array during render, this will throw a `ReferenceError` before the reset-to-`'all'` logic can run.

**Fix:** Move the `const categories = useMemo(...)` declaration above the new `useEffect`, or compute the category-validity check without referencing `categories` directly.

apps/web/src/components/FileViewer.tsx
```diff
@@ -1313,12 +1313,12 @@ function BoardComposerPopover({
       }}
     >
       <div className="comment-popover-head">
-        <div>
+        <div title={target.elementId}>
```
mrcfps, line 1316: `apps/web/src/components/FileViewer.tsx`: this new tooltip only exposes `target.elementId`, but the text that now gets ellipsized in this header also includes `target.label` and the selection summary on the next lines. When the label is the long value that overflowed before, hovering still does not reveal the truncated content, so the layout no longer breaks but the user still cannot read the hidden label. Please set the tooltip to the full visible header content (for example the element id plus label plus selection summary, or at least `target.label` when present) so the ellipsis fallback remains accessible.

mrcfps (commented):
@nmsn thanks for following up on the state-sync fix — I rechecked the latest head, and the `DesignSystemsTab` changes look good now. I found one small follow-up in the comment-popover overflow fallback that is worth tightening.


lefarcen (approved):
Hi @nmsn — thanks for the fix! The TDZ blocker is now resolved: `categories` is declared before the `useEffect` reads it.

The state-sync logic looks good: when the selected surface count drops to zero or the current category becomes invalid, the UI resets to 'all'/'All'. The FileViewer overflow fixes and CSS adjustments are solid.

lefarcen (approved):
Hey @nmsn, thanks for the quick iteration! 🎉

The TDZ fix is correct — `categories` memo is now declared before the `useEffect` that depends on it, so the component will render without throwing.

The state-sync reset logic also looks solid:
- Resets `surfaceFilter` to `'all'` when the current surface count drops to zero
- Resets `category` to `'All'` when the selected category disappears from the filtered list
- Proper dependency array includes all the values the effect reads

CI is mostly green (Windows smoke still running, but Mac smoke passed, and the TDZ was caught at compile-time in the earlier head, so this should be fine).

**Ready to merge from my side** once Windows smoke finishes and a maintainer gives the final call. Nice work addressing the review feedback! 👍

mrcfps (approved):

PR #971, opened 2026-05-08
fix: ensure Settings close button is always clickable

xxiaoxiong:
## Summary

Fixes the Settings close button becoming unclickable after PR #681 introduced the new Settings layout with Orbit summaries.

## Root Cause

The issue was caused by a combination of:

1. **Low z-index**: `.settings-chrome` had `z-index: 3`, which could be overridden by modal content in certain stacking contexts
2. **Pointer events pattern**: The `pointer-events: none` on the container with `pointer-events: auto` on children may not work reliably in all scenarios

## Changes

- Increased `.settings-chrome` z-index from `3` to `10` to ensure the close button always stays above all modal content
- Removed `pointer-events: none` from `.settings-chrome` container to avoid potential click interception issues
- `.settings-autosave` already has `pointer-events: none`, so it won't block clicks

## Testing

- Verified the CSS changes are minimal and focused
- The close button should now be reliably clickable across all Settings tabs and window sizes
- No visual changes to the UI

## Notes

This is a conservative fix that addresses the z-index stacking and pointer-events issues without touching the layout or structure introduced in PR #681.

Fixes #905

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -1204,11 +1204,10 @@ code {
   position: absolute;
   top: 14px;
   right: 14px;
-  z-index: 3;
+  z-index: 10;
   display: flex;
   align-items: center;
   gap: 8px;
-  pointer-events: none; /* children opt back in */
 }
 .settings-chrome > * {
   pointer-events: auto;


lefarcen:
Hi @xxiaoxiong! 🎉

Thanks for the contribution — nice focused fix targeting the z-index stacking issue!

— open-design team

mrcfps (approved):
@xxiaoxiong I reviewed the `apps/web/src/index.css` change around the Settings chrome stacking and pointer-event handling. The fix stays narrowly scoped to the top-right chrome strip, and the autosave pill remains non-interactive, so I did not find any changed-range correctness or maintainability issues here. Thanks for chasing down this Settings regression 🙏

lefarcen (commented):
Hey @xxiaoxiong! 👋

Nice focused fix. The z-index bump to 10 keeps the close button above Settings-local content (which uses z-index 2-6) while staying well below global overlays (900-1600), and removing the container-level `pointer-events: none` is safe since the only passive child (`.settings-autosave`) already has its own `pointer-events: none`.

The change is narrowly scoped and addresses the regression from #681 without side effects.

PR #972, opened 2026-05-08
fix(daemon): write SSE events atomically in createSseResponse.send

jnalv414:
## Summary
- Combine the three `res.write()` calls in `createSseResponse.send` into one atomic write so each SSE event lands as a single TCP frame for sub-MTU events.
- Fixes four chat-route SSE tests in `apps/daemon/tests/chat-route.test.ts` (`surfaces Qoder assistant error records...`, `fails Qoder runs when the result reports is_error...`, `fails stalled json-stream runs after the inactivity timeout elapses`, `marks stalled runs failed even when the child ignores SIGTERM`) which were racing against the multi-write chunk split inside the `readSseUntil` helper.

## Why this is a real bug, not just a test fix
The three separate writes (`id:`, `event:`, `data:`) get flushed as three TCP chunks on a fast localhost connection. Any consumer that reads chunks and looks for the `event: <type>` marker can return between writes — the test helper hits that exact race, but production consumers that don't fully buffer events are vulnerable to the same partial-event read. Atomic-per-event writes match how other SSE implementations (rxjs/SSE, EventSource polyfills, Express SSE middlewares) emit events.

## Test plan
- [x] `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts -t "surfaces Qoder|is_error with exit code 0|fails stalled json-stream|marks stalled runs failed even when"` — 4/4 pass (was 0/4)
- [x] `pnpm --filter @open-design/daemon test` — 86/86 test files pass
- [x] `pnpm typecheck` — clean across all 12 workspaces
- [x] Public API of `createSseResponse` unchanged; no callers needed updating

## Diff

```diff
     send(event, data, id: string | number | null | undefined = null) {
       if (!canWrite()) return false;
-      if (id !== null && id !== undefined) res.write(`id: ${id}\n`);
-      res.write(`event: ${event}\n`);
-      res.write(`data: ${JSON.stringify(data)}\n\n`);
+      // Assemble the full SSE event into a single write so id/event/data land
+      // in one TCP chunk. Three separate writes would let `event: <type>` flush
+      // ahead of the `data:` payload, which produces partial events for
+      // consumers that read chunk-by-chunk (e.g. tests using a Response body
+      // reader with a substring marker).
+      const idLine = id !== null && id !== undefined ? `id: ${id}\n` : '';
+      res.write(`${idLine}event: ${event}\ndata: ${JSON.stringify(data)}\n\n`);
       return true;
     },
```

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -1790,9 +1790,13 @@ export function createSseResponse(
     /** @param {ChatSseEvent['event'] | ProxySseEvent['event'] | string} event */
     send(event, data, id: string | number | null | undefined = null) {
       if (!canWrite()) return false;
-      if (id !== null && id !== undefined) res.write(`id: ${id}\n`);
-      res.write(`event: ${event}\n`);
-      res.write(`data: ${JSON.stringify(data)}\n\n`);
+      // Assemble the full SSE event into a single write so id/event/data land
+      // in one TCP chunk. Three separate writes would let `event: <type>` flush
+      // ahead of the `data:` payload, which produces partial events for
+      // consumers that read chunk-by-chunk (e.g. tests using a Response body
+      // reader with a substring marker).
+      const idLine = id !== null && id !== undefined ? `id: ${id}\n` : '';
+      res.write(`${idLine}event: ${event}\ndata: ${JSON.stringify(data)}\n\n`);
       return true;
     },
     writeKeepAlive,


apps/daemon/src/server.ts
```diff
@@ -1790,9 +1790,13 @@ export function createSseResponse(
...
-      if (id !== null && id !== undefined) res.write(`id: ${id}\n`);
-      res.write(`event: ${event}\n`);
-      res.write(`data: ${JSON.stringify(data)}\n\n`);
+      // Assemble the full SSE event into a single write so id/event/data land
+      // in one TCP chunk. Three separate writes would let `event: <type>` flush
+      // ahead of the `data:` payload, which produces partial events for
+      // consumers that read chunk-by-chunk (e.g. tests using a Response body
+      // reader with a substring marker).
+      const idLine = id !== null && id !== undefined ? `id: ${id}\n` : '';
+      res.write(`${idLine}event: ${event}\ndata: ${JSON.stringify(data)}\n\n`);
```
mrcfps, line 1799: `createSseResponse.send` is safer with one `write()`, but this still doesn't guarantee that `event:` and `data:` arrive in the same network/read chunk. Node can split a single `res.write(...)` across chunked-transfer boundaries, so consumers like `readSseUntil(..., 'event: error')` can still observe `event:` before the matching `data:` if they stop on a substring instead of waiting for the SSE record terminator. That matters because the new comment and PR rationale describe this as an atomic delivery fix, which can leave the flaky test helper in place and overstate the production guarantee. Suggested change: either update the affected SSE reader/test helper to buffer until `\n\n` before parsing an event, or soften this code comment to say this only reduces chunk-split reads and add a regression test in `apps/daemon/tests/sse-response.test.ts` that asserts `send()` performs one `write()` call if that stronger contract is intentional.

mrcfps (commented):
Thanks for the fix here — I found one follow-up worth addressing before we rely on this as an SSE atomicity guarantee.


lefarcen:
Hi @jnalv414! 🎉
Thanks for the contribution — this is a solid fix for the SSE test race condition.

— open-design team

apps/daemon/src/server.ts
```diff
@@ -1790,9 +1790,13 @@ export function createSseResponse(
     /** @param {ChatSseEvent['event'] | ProxySseEvent['event'] | string} event */
     send(event, data, id: string | number | null | undefined = null) {
       if (!canWrite()) return false;
-      if (id !== null && id !== undefined) res.write(`id: ${id}\n`);
-      res.write(`event: ${event}\n`);
-      res.write(`data: ${JSON.stringify(data)}\n\n`);
+      // Assemble the full SSE event into a single write so id/event/data land
```
lefarcen, line 1793: **P3** The comment says "one TCP chunk," but a single `res.write()` is one *application-level write* — Node/kernel/TLS/proxies may still split or coalesce bytes downstream. SSE consumers must still buffer until the blank-line terminator (`\n\n`).

Consider rewording as "single application-level write" for accuracy. The fix is correct either way — you've eliminated the local multi-write race.

lefarcen (commented):
Hi @jnalv414! Clean fix. The atomic write approach matches standard SSE implementations and correctly addresses the test race.

One minor comment on phrasing:

lefarcen (approved):

jnalv414:
Happy I could be of service!!!

leilei926524-tech:
Hi @jnalv414 — welcome to Open Design contributors. 🎉

Your first merged PR (#972) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

Issue #918, opened 2026-05-08
Codex agent stalled


Liridonrexhaj:
Codex 5.5 Agent stalled without emitting any new output for 120s.

<img width="440" height="229" alt="Image" src="https://github.com/user-attachments/assets/9e426e84-37de-4e1c-b70a-3d679b388b15" />

lefarcen:
Hi @Liridonrexhaj, thanks for the report.

That message is coming from Open Design's daemon inactivity watchdog: it fails a run if the agent process emits no stdout/stderr/structured events for 120 seconds. This is separate from the ACP 180s timeout being handled in #765.

Could you share the Open Design version/commit you're running and the exact execution agent/model selected in Settings? That will tell us whether this was the Codex CLI adapter itself or an ACP-backed agent using a Codex 5.5 model.

As a temporary workaround for testing, you can raise the watchdog with:

```bash
OD_CHAT_RUN_INACTIVITY_TIMEOUT_MS=600000
```

That sets the inactivity window to 10 minutes.

shangxinyu1:
Thanks for reporting this.

The message in the screenshot means the run hit the new inactivity watchdog: the Codex 5.5 agent did not emit any new output for 120s, so Open Design marked the run as stalled instead of leaving it hanging indefinitely.

To make this actionable, could you share a bit more context?

- What exact prompt/task were you running when this happened?
- Was this in the desktop app or web?
- Does it reproduce consistently, or was it a one-off?
- What was the last visible output before it stalled?
- If you switch to another model/agent, does the same task complete normally?

If you can share a reproducible prompt or steps, that would help narrow down whether this is specific to Codex 5.5, a certain tool path, or a broader run-lifecycle issue.

Liridonrexhaj:
Hi, yes here is some context i hope this helps. i pulled it out with codex.

pulled the local run metadata from `.od/app.sqlite` and the tools-dev logs.

Open Design version/commit:
- Version: `0.5.0`
- Commit: `8fee22d3586fe6c584e34c43f1a0d571be7387e1`
- Commit title: `Fix stuck chat runs and unintended cancels (#896)`

Environment:
- Running via local web, not desktop: `pnpm tools-dev start web`
- Web was running at `127.0.0.1:61549`
- Daemon was running at `127.0.0.1:61548`
- Desktop status was idle
- Codex CLI path: `/Applications/Codex.app/Contents/Resources/codex`
[...]

PR #976, opened 2026-05-08
fix: reset inactivity watchdog on raw stdout bytes, not just parsed events (#918)

ButterHost69:
Summary
---
Add a universal raw stdout activity listener so the agent inactivity watchdog resets on every byte written to the pipe, not just on complete parsed JSON events. Fixes false "Agent stalled without emitting any new output for 120s" errors across all seven structured stream formats.

Changes
---
apps/daemon/src/server.ts (1 file, +6 lines)
- Added child.stdout.on('data', () => noteAgentActivity()) at server.ts:5557, registered immediately after child.stdout.setEncoding('utf8') and before any format-specific stream dispatch
- Comment documents the affected adapters: Codex item.completed batching, pi-rpc session/prompt, ACP agent messages, and models with long non-streamed reasoning phases

Context
---
Structured stream handlers (Claude, Qoder, Copilot, Codex, OpenCode, pi-rpc, ACP) all use createJsonLineStream-style buffering — they accumulate bytes, split on newlines, and only emit events when a complete JSON line is parsed. noteAgentActivity() was only called on those parsed events.
During long model reasoning or buffered artifact generation — notably Codex 5.5 writing a full landing page HTML as a single item.completed event, or DeepSeek V4 / Kimi 2.6 spending >120s in non-streamed reasoning — the stdout pipe stays silent even though the agent is working. The watchdog fires and kills the run with a false stall error.
The fix adds a raw 'data' listener before any format-specific wiring. Node.js supports multiple 'data' listeners on readable streams — the raw listener fires first and resets the timer, then the format parser processes the chunk as before. Stderr was already covered (line 5805); stdout was the gap.

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -5550,6 +5550,12 @@ export async function startServer({
     child.stdout.setEncoding('utf8');
     child.stderr.setEncoding('utf8');
 
+    // Reset the inactivity watchdog on every raw stdout byte so that
+    // structured adapters that buffer partial lines (Codex item.completed,
+    // pi-rpc session/prompt, ACP agent messages) and models that spend a
+    // long time in non-streamed reasoning still keep the run alive.
+    child.stdout.on('data', () => noteAgentActivity());
+
     // Critique Theater branch (M0 dark launch, default disabled).
     // Only plain-stream adapters are routed through runOrchestrator in v1.
     // Adapters that emit structured wrappers (claude-stream-json,


mrcfps (approved):
@ButterHost69 I reviewed the daemon-side watchdog change around `child.stdout` and verified the new raw stdout activity listener is registered before the format-specific handlers, so buffered structured adapters now keep the run alive without changing the existing stderr/error/close lifecycle. Thanks for tightening up this false-stall path across the shared stream handling.


apps/daemon/src/server.ts
```diff
@@ -5550,6 +5550,12 @@ export async function startServer({
     child.stdout.setEncoding('utf8');
     child.stderr.setEncoding('utf8');
 
+    // Reset the inactivity watchdog on every raw stdout byte so that
+    // structured adapters that buffer partial lines (Codex item.completed,
+    // pi-rpc session/prompt, ACP agent messages) and models that spend a
+    // long time in non-streamed reasoning still keep the run alive.
+    child.stdout.on('data', () => noteAgentActivity());
```
lefarcen, line 5557: **P3** Test coverage gap: the fix solves the #918 case (raw stdout chunks arrive for >120s while structured parser buffers incomplete JSON), but existing tests don't exercise this path. Consider adding a daemon chat-route test with a fake structured agent that:
1. Writes partial JSON/stdout chunks below the timeout interval
2. Delays the newline/complete event until after the old watchdog would have fired
3. Exits successfully

That would fail before this PR and pass with the raw listener. Not blocking — just strengthens confidence that future refactors won't reintroduce the false-stall.

lefarcen (commented):
Hey @ButterHost69! 🎉 Thanks for this fix — addresses the false-stall watchdog issue cleanly by resetting on raw stdout bytes before JSON parsing.

Reviewed the daemon lifecycle and verified the new listener is registered at the right point (after spawn/encoding, before adapter-specific handlers). Multiple `data` listeners work correctly here — no race or interference with existing stream handlers.

One non-blocking suggestion below on test coverage.

lefarcen:
Agent-to-agent review note: Applied Lens A to `apps/daemon/src/server.ts`; Lens B not applicable. No P1/P2 correctness, lifecycle, stream-interference, security, or architecture issues found with the raw stdout activity listener. One P3 follow-up: add a regression test for the exact #918 case where raw stdout chunks arrive under the inactivity timeout but no complete structured JSON event is emitted until after the old watchdog would have fired.

leilei926524-tech:
Hi @ButterHost69 — welcome to Open Design contributors. 🎉

Your first merged PR (#976) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #987, opened 2026-05-08
fix(web): differentiate selected, hover, and focus states in the language switcher

Nagendhra-web:
Closes #628.

The Settings → Language menu used \`--bg-subtle\` for both \`:hover\` and \`.active\`, so the currently-selected language was hard to tell apart from a row the pointer was over, and \`:focus-visible\` had no treatment at all. The reporter's screenshot shows exactly that: pick a language, hover another row, and the active state effectively disappears.

CSS-only change in [\`apps/web/src/index.css\`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/index.css):

- Selected row → \`--accent-tint\` background + accent text + \`font-weight: 500\` so the active choice reads even when nothing is hovered.
- Selected + hover → lifts to \`--accent-soft\` so the cursor still gives feedback over the active row.
- \`:focus-visible\` → 2px accent outline (inset) for keyboard users.

The existing \`.active\` className wiring in the React picker already drives this; no JSX changes needed.

Validated locally: \`pnpm guard\` and \`pnpm --filter @open-design/web typecheck\` clean.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -3151,9 +3151,20 @@ code {
   text-align: left;
   color: var(--text);
 }
-.settings-language-option:hover,
-.settings-language-option.active { background: var(--bg-subtle); }
-.settings-language-option.active { color: var(--accent); }
+/* Differentiate hover, selected, and keyboard-focus states so the
+   currently-selected language is visually distinct even when the
+   pointer is hovering a different row. Issue #628. */
+.settings-language-option:hover { background: var(--bg-subtle); }
+.settings-language-option.active {
+  background: var(--accent-tint);
+  color: var(--accent);
+  font-weight: 500;
+}
+.settings-language-option.active:hover { background: var(--accent-soft); }
+.settings-language-option:focus-visible {
+  outline: 2px solid var(--accent);
+  outline-offset: -2px;
+}
 .empty-card {
   padding: 16px;
   border: 1px dashed var(--border);


mrcfps (approved):
@Nagendhra-web I verified the language switcher CSS change in `apps/web/src/index.css`: the selected row now stays visually distinct from plain hover, the active row still gives pointer feedback on hover, and keyboard navigation gets its own `:focus-visible` treatment without changing the existing React wiring. Thanks for tightening up this UX/accessibility fix 🙌


lefarcen:
Hi @Nagendhra-web! 🎉

Thanks for the contribution — this CSS fix directly addresses the hover/selected/focus confusion reported in #628.

— open-design team

lefarcen (commented):
Hey @Nagendhra-web, nice work on this! 👍 The CSS change is solid and directly addresses the issue.

- Theme variables (`--accent-tint`, `--accent-soft`) exist and are used appropriately ✅
- State model (hover → selected → selected+hover) is correct ✅
- No specificity conflicts detected ✅

1. **Accent text color**: The `color: var(--accent)` on line 3160 doesn't affect the visible language title/code because `.settings-language-option-title` and `.settings-language-option-code` child rules override it. If accent-colored text is intended for the active state, those child spans need explicit color rules.

2. **Focus contrast**: In light theme, the accent outline is about 2.89:1 against the `--accent-soft` background (line 3165), slightly below WCAG's 3:1 focus-indicator target. Consider a stronger focus color or avoiding `--accent-soft` directly under the inset outline.

Both are subtle polish points — the core fix is production-ready. Looks good to me; deferring final approval to a maintainer.

PR #988, opened 2026-05-08
fix(web): confirm before closing a dirty sketch so unsaved strokes aren't lost

### LINKED ISSUE #139: Closing a sketch discards unsaved changes without auto-save or a save confirmation
Summary:
When editing a sketch, clicking Close immediately discards the current file changes. There is no auto-save behavior and no confirmation asking whether the user wants to save before closing.

Steps to reproduce:
1. Open or create a sketch.
2. Draw or edit content in the sketch.
3. Click the Close button without manually clicking Save first.
4. Reopen the sketch or check the saved result.

Expected behavior:
If the sketch has unsaved changes, the system should either:
- auto-save before closing, or
- show a confirmation dialog asking whether to save, discard, or cancel.

Actual behavior:
- Clicking Close immediately exits the sketch.
- The current changes are discarded.
- There is no auto-save and no secondary confirmation.

Why this is a problem:
- Users can lose work very easily.
- Close behaves like a destructive action without warning.
- The current behavior does not match common editing expectations.
- This weakens trust in the sketch editing workflow, especially during quick iteration.

Expected fix direction:
- Detect unsaved changes in the sketch editor.
- Prompt the user before closing when there are unsaved edits.
- Alternatively support reliable auto-save, with visible feedback so users know their work is preserved.
- Make discard behavior explicit instead of silent.

Priority:
High. This can directly cause user work loss in a core editing workflow.

Image

Nagendhra-web:
Closes #139.

Clicking **Close** in the sketch editor went straight to \`onCancel()\` even when the toolbar already showed the unsaved-changes bullet (\`{dirty ? ' •' : ''}\` next to the file name). One stray click and the work was gone.

Wrap the existing onClick in [\`apps/web/src/components/SketchEditor.tsx\`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SketchEditor.tsx) with a \`window.confirm()\` guard, gated on the existing \`dirty\` prop:

```tsx
onClick={() => {
  if (dirty && !confirm(t('sketch.closeConfirm'))) return;
  onCancel();
}}
```

Clean sketches still close in one click; dirty ones now ask first. Same pattern the rest of the codebase uses for destructive actions (conversation delete, design delete, file delete in FileWorkspace, the media-provider and Composio Clear buttons).

## i18n

New key \`sketch.closeConfirm\` defined in [\`types.ts\`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/i18n/types.ts) and translated across all 17 locales. Locale alignment test stays green.

## Validation

- \`pnpm guard\` clean.
- \`pnpm --filter @open-design/web exec vitest run tests/i18n/locales.test.ts\` — 2/2 pass.

### DIFF
--- a/apps/web/src/components/FileWorkspace.tsx
+++ b/apps/web/src/components/FileWorkspace.tsx
@@ -168,7 +168,10 @@ export function FileWorkspace({
   }
 
   function closeTab(name: string) {
-    const isPending = sketches[name] && !sketches[name]!.persisted;
+    const sketchEntry = sketches[name];
+    const isPending = sketchEntry && !sketchEntry.persisted;
+    const hasUnsavedStrokes = sketchEntry && (sketchEntry.dirty || !sketchEntry.persisted);
+    if (hasUnsavedStrokes && !confirm(t('sketch.closeConfirm'))) return;
     if (isPending) {
       setSketches((curr) => {
         const next = { ...curr };
--- a/apps/web/src/i18n/locales/ar.ts
+++ b/apps/web/src/i18n/locales/ar.ts
@@ -928,6 +928,7 @@ export const ar: Dict = {
   'sketch.undo': 'تراجع',
   'sketch.clear': 'مسح',
   'sketch.close': 'إغلاق',
+  'sketch.closeConfirm': 'إغلاق الرسم وتجاهل التغييرات غير المحفوظة؟',
   'sketch.textPrompt': 'النص:',
   'sketch.textModalTitle': 'إضافة نص',
 
--- a/apps/web/src/i18n/locales/de.ts
+++ b/apps/web/src/i18n/locales/de.ts
@@ -816,6 +816,7 @@ export const de: Dict = {
   'sketch.undo': 'Rückgängig',
   'sketch.clear': 'Leeren',
   'sketch.close': 'Schließen',
+  'sketch.closeConfirm': 'Skizze schließen und ungespeicherte Änderungen verwerfen?',
   'sketch.textPrompt': 'Text:',
   'sketch.textModalTitle': 'Text hinzufügen',
 
--- a/apps/web/src/i18n/locales/en.ts
+++ b/apps/web/src/i18n/locales/en.ts
@@ -974,6 +974,7 @@ export const en: Dict = {
   'sketch.undo': 'Undo',
   'sketch.clear': 'Clear',
   'sketch.close': 'Close',
+  'sketch.closeConfirm': 'Close the sketch and discard your unsaved changes?',
   'sketch.textPrompt': 'Text:',
   'sketch.textModalTitle': 'Add text',
 
--- a/apps/web/src/i18n/locales/es-ES.ts
+++ b/apps/web/src/i18n/locales/es-ES.ts
@@ -817,6 +817,7 @@ export const esES: Dict = {
   'sketch.undo': 'Deshacer',
   'sketch.clear': 'Limpiar',
   'sketch.close': 'Cerrar',
+  'sketch.closeConfirm': '¿Cerrar el boceto y descartar los cambios sin guardar?',
   'sketch.textPrompt': 'Texto:',
   'sketch.textModalTitle': 'Añadir texto',
 
--- a/apps/web/src/i18n/locales/fa.ts
+++ b/apps/web/src/i18n/locales/fa.ts
@@ -960,6 +960,7 @@ export const fa: Dict = {
   'sketch.undo': 'واگرد',
   'sketch.clear': 'پاک کردن',
   'sketch.close': 'بستن',
+  'sketch.closeConfirm': 'طرح را ببندم و تغییرات ذخیره‌نشده را نادیده بگیرم؟',
   'sketch.textPrompt': 'متن:',
   'sketch.textModalTitle': 'افزودن متن',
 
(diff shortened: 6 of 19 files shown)


apps/web/src/components/SketchEditor.tsx
```diff
@@ -255,7 +255,16 @@ export function SketchEditor({
...
         {onCancel ? (
-          <button className="ghost" onClick={onCancel}>
+          <button
+            className="ghost"
+            onClick={() => {
+              // Close immediately when there's nothing to lose. With unsaved
+              // strokes, gate behind window.confirm so a stray click doesn't
+              // discard a sketch the user was iterating on. Issue #139.
+              if (dirty && !confirm(t('sketch.closeConfirm'))) return;
+              onCancel();
```
mrcfps: `apps/web/src/components/FileWorkspace.tsx:157-183` still closes a sketch tab immediately via `Tab.onClose={() => closeTab(name)}`. For dirty or not-yet-persisted sketches, `closeTab()` deletes the in-memory entry or removes the tab without any confirmation, so the same unsaved strokes can still be lost through the tab close icon even after this toolbar button guard lands. Because the PR goal is to stop accidental sketch loss on close, the confirmation needs to live on the shared close path that both the toolbar button and tab close affordance use, or `closeTab()` needs its own dirty-sketch check before it mutates state.

mrcfps (changes requested):
@Nagendhra-web Thanks for tightening the toolbar close flow and wiring the new locale key through every language. I found one remaining unsaved-work path on the current head that still needs the same guard before this is ready to merge.


lefarcen (commented):
Clean implementation — the dirty-gated confirm follows the same pattern used elsewhere in the codebase (conversation/design/file delete). The i18n coverage is complete (17 locales + type definition), and the logic is minimal (no save/undo/clear changes).

No blocking issues found. Deferring final approval to a maintainer.

apps/web/src/components/SketchEditor.tsx
Nagendhra-web: Good catch, fixed in 6dbe2c0. Moved the guard into closeTab so the same prompt fires whether you hit the toolbar Close button or the tab close icon. The check uses sketchEntry.dirty || !sketchEntry.persisted, which mirrors the dirty-mark dot we already render on the tab label, so anything visually flagged as unsaved now triggers the confirm before state is mutated.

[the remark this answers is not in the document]
Nagendhra-web: Addressed in 6dbe2c0 by moving the confirm into closeTab, so the tab close icon now goes through the same dirty/unpersisted check as the toolbar button.

lefarcen (commented):
Verified the fix — the guard now lives in `closeTab()` (FileWorkspace.tsx:173), so both the toolbar Close button and the tab close icon check for unsaved work before closing. The logic uses `sketchEntry.dirty || !sketchEntry.persisted`, which matches the dirty-mark dot on the tab label.

No new issues found. CI is green. mrcfps's finding is addressed on this head (6dbe2c0), though mrcfps hasn't re-reviewed the current head yet. Looks ready to me; deferring final approval to a maintainer.

apps/web/src/components/SketchEditor.tsx
```diff
@@ -255,7 +255,16 @@ export function SketchEditor({
...
         </span>
         {onCancel ? (
-          <button className="ghost" onClick={onCancel}>
+          <button
+            className="ghost"
+            onClick={() => {
+              // Close immediately when there's nothing to lose. With unsaved
+              // strokes, gate behind window.confirm so a stray click doesn't
+              // discard a sketch the user was iterating on. Issue #139.
+              if (dirty && !confirm(t('sketch.closeConfirm'))) return;
```
mrcfps: Clicking the toolbar Close button now shows two confirmations for the same dirty sketch. `SketchEditor` asks on `dirty` here, then `onCancel()` immediately calls `FileWorkspace.closeTab()` (`apps/web/src/components/FileWorkspace.tsx:170-175`), which asks again for `sketchEntry.dirty || !sketchEntry.persisted`. That makes the main close flow require two accepts for unsaved sketches, while the tab-close icon only asks once. Please keep the confirm in the shared close path only—e.g. let this button call `onCancel()` directly and rely on `closeTab()` for the prompt, or thread an explicit already-confirmed flag through `onCancel` if you want the editor to own the dialog.

mrcfps (commented):
@Nagendhra-web Thanks for moving the dirty-sketch guard onto the shared close path and wiring the new locale key through every language. I found one follow-up UX regression on the current head: the toolbar Close path now hits the same confirm twice, so I left an inline suggestion to keep the shared guard in one place.


apps/web/src/components/SketchEditor.tsx
Nagendhra-web: Good catch, fixed in 7acdd87. Reverted the SketchEditor toolbar Close button to a plain onClick={onCancel} so the prompt only fires from closeTab() now. Toolbar Close and the tab close icon both flow through the same single guard, no more double accept on dirty sketches.

mrcfps (approved):
@Nagendhra-web I reviewed the changed range on `7acdd878` and verified the close confirmation now lives on `FileWorkspace.closeTab()`, so both the toolbar Close action and the tab close affordance share the same unsaved-sketch guard. I also checked that `sketch.closeConfirm` is typed in `apps/web/src/i18n/types.ts`, translated across the locale files in this PR, and covered by the existing locale-alignment test. Thanks for tightening this flow and closing the remaining accidental-loss path 🙌

PR #989, opened 2026-05-08
fix(web): keep chat auto-scroll glued to bottom across streaming chunks

### LINKED ISSUE #983: Chat does not automatically scroll down to show latest responses/questions
The main chat window on the left panel does not automatically scroll down to show latest chat responses or questions.

Tested on Ubuntu 24.04 / latest Chrome

Nagendhra-web:
Closes #983.

The chat panel stopped following new assistant turns even when the user was sitting at the bottom of the conversation. The auto-scroll effect in [\`apps/web/src/components/ChatPane.tsx\`](https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/ChatPane.tsx) gated on a *post-content* distance check:

```ts
const distance = el.scrollHeight - el.scrollTop - el.clientHeight;
if (distance < 80) el.scrollTop = el.scrollHeight;
```

By the time that effect runs, \`scrollHeight\` has already grown by the new chunk. A tool-use card or markdown render typically adds 100+ px in one tick, so the post-content distance is suddenly \`> 80\` even though the user was glued to the bottom a moment before. The auto-scroll skips, and successive chunks compound the gap.

## Fix

Switch the gate to the existing \`scrolledFromBottom\` state. That flag is maintained by the user-driven scroll listener and only flips on a real scroll event (not on content growth), so it carries the user's pre-content intent into the effect:

```ts
if (!scrolledFromBottom) {
  el.scrollTop = el.scrollHeight;
}
```

A user reading earlier output stays put; a user pinned to the latest message stays pinned across every streamed chunk.

## Validation

- \`pnpm guard\` clean.
- \`pnpm --filter @open-design/web exec vitest run tests/components/chat-scroll-preservation.test.tsx tests/components/conversation-timestamps.test.tsx\` — 6/6 pass. The prior-state behavior the chat-scroll-preservation tests cover is tab-switch restore (bottom-pinned vs absolute scrollTop), which this change does not affect.

### DIFF
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -142,6 +142,13 @@ export function ChatPane({
   const historyWrapRef = useRef<HTMLDivElement | null>(null);
   const composerRef = useRef<ChatComposerHandle | null>(null);
   const didInitialScrollRef = useRef(false);
+  // Tracks whether the user is glued close enough to the bottom that
+  // streamed content should auto-follow. Distinct from the jump-button
+  // state below, which uses a wider threshold (120px) so the affordance
+  // stays visible for short scroll-ups. Auto-follow needs the tighter
+  // 80px cutoff: scrolling ~90px up is an intentional pause that
+  // shouldn't be yanked back the moment the next chunk streams in.
+  const pinnedToBottomRef = useRef(true);
   const [tab, setTab] = useState<Tab>('chat');
   const [showConvList, setShowConvList] = useState(false);
   const [scrolledFromBottom, setScrolledFromBottom] = useState(false);
@@ -178,6 +185,7 @@ export function ChatPane({
     requestAnimationFrame(() => {
       el.scrollTop = el.scrollHeight;
       setScrolledFromBottom(false);
+      pinnedToBottomRef.current = true;
     });
     // `tab` is in the deps so that switching conversations while
     // Comments is open doesn't strand the new conversation at scrollTop:
@@ -190,11 +198,18 @@ export function ChatPane({
   useEffect(() => {
     const el = logRef.current;
     if (!el) return;
-    // Auto-scroll only when we're already pinned near the bottom — preserves
-    // a user's scrollback position when they're reading earlier output while
-    // a new turn streams in.
-    const distance = el.scrollHeight - el.scrollTop - el.clientHeight;
-    if (distance < 80) {
+    // Auto-scroll only when the user was already pinned near the bottom,
+    // so a scrollback session reading earlier output isn't yanked to the
+    // latest message. We key off the pre-content `pinnedToBottomRef`
+    // (a ref so it doesn't itself re-fire this effect on scroll) instead
+    // of recomputing distance from the just-grown scrollHeight: a single
+    // streamed chunk can add 100+ px in one render, which made the
+    // post-content distance check skip auto-scroll even when the user
+    // was glued to the bottom. We deliberately use the tighter 80px
+    // cutoff tracked by the ref (not the wider 120px jump-button
+    // threshold) so a deliberate ~90px scroll-up isn't snapped back the
+    // next time content streams in. Issue #983.
+    if (pinnedToBottomRef.current) {
       el.scrollTop = el.scrollHeight;
     }
   }, [messages, error]);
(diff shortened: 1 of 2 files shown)


apps/web/src/components/ChatPane.tsx
```diff
@@ -190,14 +190,18 @@ export function ChatPane({
...
+    // Auto-scroll only when the user was already pinned near the bottom,
+    // so a scrollback session reading earlier output isn't yanked to the
+    // latest message. We key off the pre-content `scrolledFromBottom`
+    // signal (maintained by the scroll listener and only flipped by a
+    // real user scroll) instead of recomputing distance from the just-
+    // grown scrollHeight: a single streamed chunk can add 100+ px in
+    // one render, which made the post-content distance check skip
+    // auto-scroll even when the user was glued to the bottom. Issue #983.
+    if (!scrolledFromBottom) {
       el.scrollTop = el.scrollHeight;
```
mrcfps: `scrolledFromBottom` is the jump-button state (`distance > 120` in `onScroll`), not the stricter "user is pinned close enough that streaming may auto-follow" signal that this effect needs. With this gate, if someone intentionally scrolls a short distance up — e.g. they stop ~90px from the bottom — `onScroll` still leaves `scrolledFromBottom` false, so the next streamed chunk immediately forces `scrollTop = scrollHeight` and yanks them away from what they were reading. That regresses the scrollback-preservation behavior this PR is trying to restore. Please keep a separate pre-content pinned/ref threshold for the auto-scroll decision (or otherwise preserve the old near-bottom cutoff for this effect) and add a regression test covering the short-scrollback case in `apps/web/tests/components/chat-scroll-preservation.test.tsx`.

mrcfps (changes requested):
@Nagendhra-web Thanks for the fix here — I verified the new auto-scroll path and found one correctness regression in the updated gate that should be addressed before merge.


lefarcen (commented):
Hey @Nagendhra-web! The diagnosis is excellent — tracking the root cause to the post-content distance check and switching to the pre-content scroll state is the right direction. Found one non-blocking concern below that's worth addressing before merge.

**Overall**: The fix correctly identifies that `scrolledFromBottom` is a better signal than recalculating distance from the already-grown `scrollHeight`. The concern is about when this effect fires, not what it checks.

apps/web/src/components/ChatPane.tsx
```diff
@@ -190,14 +190,18 @@ export function ChatPane({
...
     }
-  }, [messages, error]);
+  }, [messages, error, scrolledFromBottom]);
```
lefarcen: **P2** — Adding `scrolledFromBottom` to the effect dependency list makes this auto-scroll effect run on user scroll-state changes, not only on content changes. If a user has scrolled up and then manually scrolls back to within the "near bottom" threshold, `setScrolledFromBottom(false)` will re-render and this effect will immediately force `scrollTop = scrollHeight` even when no new message/error arrived. That's a behavioral regression from "preserve user-controlled scrollback" to "snap when crossing the threshold."

Keep the content-growth trigger as `[messages, error]` and read the latest pinned state from a ref updated by the scroll listener, or gate the effect so it only scrolls in response to message/error growth.

Alternatively, if the current behavior is intentional (snap-on-threshold-cross), clarify that in the comment so future readers understand why `scrolledFromBottom` is in the deps.

Nagendhra-web: Yep, that was a real concern. In 4880905 the effect's deps are back to [messages, error] only, with the pinned-to-bottom signal tracked through a ref instead of the scrolledFromBottom state. Crossing the threshold no longer re-fires the effect; it only runs when content actually grows.

mrcfps (approved):
@Nagendhra-web Thanks for the quick follow-up here — I rechecked the updated `pinnedToBottomRef` auto-follow gate in `apps/web/src/components/ChatPane.tsx` and the new short-scrollback regression coverage in `apps/web/tests/components/chat-scroll-preservation.test.tsx`, and the current head cleanly addresses the earlier scrollback and effect-trigger regressions. Nice catch on separating the tighter auto-follow threshold from the wider jump-button state.


lefarcen (commented):
Nice fix, @Nagendhra-web! The separate `pinnedToBottomRef` with the tighter 80px cutoff addresses both the original issue and the regression concern. The effect deps staying as `[messages, error]` ensures scroll-state changes don't re-trigger the auto-follow, and the new test covers the ~90px short-scrollback case.

Both my P2 concern and mrcfps's CHANGES_REQUESTED finding are resolved. Looks good to me; deferring final approval to a maintainer.

PR #957, opened 2026-05-08
fix: increase top padding in modal footer for better visual balance

xxiaoxiong:
## Summary

Fixes #913 - Increases top padding in modal footers to provide better visual balance between the primary action button and the divider line above it.

## Problem

In modal footers (e.g., Deploy to Cloudflare Pages, Save Template), the primary action buttons were sitting too close to the top border, making the footer area feel visually cramped.

## Solution

Modified `.modal-foot` CSS class in `apps/web/src/index.css`:
- **Before:** `padding: 12px var(--modal-padding);` (uniform 12px padding)
- **After:** `padding: 16px var(--modal-padding) 12px;` (16px top, 12px bottom)

This increases the top padding by 4px while maintaining the existing bottom spacing and left/right padding.

## Impact

This change affects three modal footers:
1. ✅ Deploy modal (FileViewer.tsx:4949) - **Primary target**
2. ✅ Template save modal (FileViewer.tsx:4692) - Also benefits from improved spacing
3. ✅ Sketch text editor modal (SketchEditor.tsx:304) - Also benefits from improved spacing

All three modals will have better visual balance with no negative side effects.

## Testing

- [x] Verified CSS change is minimal and conservative (+4px top padding)
- [x] Confirmed no other modal styles are affected
- [x] Checked all usages of `.modal-foot` class (3 locations)
- [x] Ensured backward compatibility with existing modal layouts

## Checklist

- [x] Code follows project style guidelines
- [x] Change is minimal and focused on the reported issue
- [x] No breaking changes
- [x] Commit message follows conventional commits format

PR #992, opened 2026-05-08
fix(web): give the deploy modal primary action a bit more breathing room

### LINKED ISSUE #913: Deploy modal footer button sits too close to the top border
## Summary

In the deploy modal, the primary action button in the footer sits too close to the top divider line.

## Current behavior

The footer action area feels visually cramped because the button is positioned too close to the border above it.

## Expected behavior

The footer actions should have a bit more top spacing so the primary button sits slightly lower and feels better balanced.

## Steps to reproduce

1. Open a deploy flow such as **Deploy to Cloudflare Pages**.
2. Look at the footer area at the bottom of the modal.
3. Compare the primary action button against the divider line above it.

## Notes

- A screenshot was provided showing the button placement.
- The issue is visual spacing, not button sizing or alignment logic.
- The request was to move the button slightly downward.

## Why this matters

Small spacing issues in modal footers make primary actions feel cramped and reduce overall UI polish, especially in prominent flows like deploy.

Image

Nagendhra-web:
Closes #913.

The deploy modal's primary action button (Cloudflare Pages, etc.) was sitting flush against the divider that separates the dialog body from the footer. The shared `.modal-foot` rule uses 12px vertical padding, which feels fine in the shorter dialogs that share it (sketch close confirm, template save) but reads as cramped in deploy because the form above ends in dense token / domain config rows that push content right up to the divider.

Added a deploy-modal-scoped override at https://github.com/nexu-io/open-design/blob/main/apps/web/src/index.css to nudge the footer top padding to 18px. Bottom padding stays on the shared 12px so the overall dialog height does not jump and we are not making any global change to other modals' rhythm.

### What I checked
- `pnpm --filter @open-design/web test` (567 passed).
- Visual check that the deploy modal primary button no longer crowds the divider while the cancel button alignment, modal height, and surrounding spacing are unchanged.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -9368,6 +9368,19 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
   max-height: calc(100vh - 32px);
   overflow: auto;
 }
+/* Add a few extra pixels above the deploy dialog's primary action so
+   it does not crowd the divider. The shared .modal-foot uses 12px
+   vertical padding which is fine for shorter dialogs but reads as
+   cramped here because the deploy form ends in dense token / domain
+   config rows that push content right up to the border. Scoped to the
+   deploy dialog only (the template-save dialog also carries the
+   .deploy-modal class on its root, which is why we key off the
+   purpose-specific .deploy-flow-modal hook instead). 16px matches the
+   bump proposed for the global rule in #957 so the rhythm stays close
+   to the rest of the app. Issue #913. */
+.deploy-flow-modal .modal-foot {
+  padding-top: 16px;
+}
 .deploy-form {
   display: flex;
   flex-direction: column;
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -4737,7 +4737,7 @@ function HtmlViewer({
       ) : null}
       {deployModalOpen ? (
         <div className="modal-backdrop" role="presentation">
-          <div className="modal deploy-modal" role="dialog" aria-modal="true">
+          <div className="modal deploy-modal deploy-flow-modal" role="dialog" aria-modal="true">
             <div className="modal-head">
               <div className="kicker">{deployProviderLabel}</div>
               <h2>{t('fileViewer.deployToProvider', { provider: deployProviderLabel })}</h2>


apps/web/src/index.css
```diff
@@ -9368,6 +9368,13 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
   max-height: calc(100vh - 32px);
   overflow: auto;
 }
+/* Bump footer top padding so the primary action button doesn't crowd
+   the divider above it. The shared .modal-foot 12px reads cramped in
+   the deploy flow because the dialog body sits flush against the
+   border. Issue #913. */
+.deploy-modal .modal-foot {
+  padding-top: 18px;
```
mrcfps: This selector also matches the template-save dialog, not just the deploy flow. In `apps/web/src/components/FileViewer.tsx:4682` and `:4740`, both modals render as `className="modal deploy-modal"`, so `.deploy-modal .modal-foot` increases the footer spacing in the template modal too. That matters because the PR description explicitly calls out template save as one of the shorter dialogs that should keep the shared 12px rhythm, and it makes future modal-specific spacing tweaks harder to reason about. Please narrow this to a deploy-only hook — for example, add a dedicated class on the actual deploy dialog or its footer and target that class here instead of the shared `deploy-modal` wrapper.

mrcfps (commented):
@Nagendhra-web Thanks for tightening the deploy modal spacing — I found one small selector-scope issue, so I left an inline suggestion below.


apps/web/src/index.css
```diff
@@ -9368,6 +9368,13 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
   max-height: calc(100vh - 32px);
   overflow: auto;
 }
+/* Bump footer top padding so the primary action button doesn't crowd
+   the divider above it. The shared .modal-foot 12px reads cramped in
+   the deploy flow because the dialog body sits flush against the
+   border. Issue #913. */
+.deploy-modal .modal-foot {
```
lefarcen: **P1 (Lens A — Code correctness)**: This selector is not deploy-modal-only in practice.

`FileViewer.tsx` also uses `className="modal deploy-modal"` for the save-template modal, so this CSS change affects **both** the deploy modal **and** the template-save modal.

Since your PR reasoning emphasizes avoiding the global modal-wide change from #957, the implementation should either:
1. Add a deploy-specific class to the actual deploy dialog (e.g., `deploy-flow-modal`), or
2. Explicitly acknowledge in the comment that template-save is intentionally affected too.

Right now the comment says "deploy flow" but the code changes two modals.

apps/web/src/index.css
```diff
@@ -9368,6 +9368,13 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
   max-height: calc(100vh - 32px);
   overflow: auto;
 }
+/* Bump footer top padding so the primary action button doesn't crowd
```
lefarcen: **P2 (Lens B — Reasoning)**: The comment references `--modal-padding` but that CSS variable appears to be defined only under `.modal-settings`. Outside settings, `padding: 12px var(--modal-padding)` computes to `12px` (the var falls back to nothing).

This override adds `padding-top: 18px` but doesn't preserve the "shared bottom/horizontal padding model" described in the PR body. The actual shared `.modal-foot` rule is just `padding: 12px`, not `12px var(--modal-padding, 22px) 12px`.

Please update the comment to match what the code actually does.

apps/web/src/index.css
lefarcen: **P2 (Lens B — Reasoning)**: The 18px value is not justified against PR #957's alternative (16px).

If the goal is a scoped version of the same visual fix, why does deploy need **18px** instead of the **16px** proposed globally in #957? A sentence explaining this choice (e.g., "deploy modal has denser content so needs slightly more breathing room") would help maintainers understand the reasoning.

Or, if 16px is sufficient, using the smaller value keeps the change closer to the existing spacing rhythm.

lefarcen (changes requested):
Hi @Nagendhra-web! Thanks for tackling this spacing issue. 🎉

The CSS itself is simple and browser-safe, and the scoped approach is reasonable. However, I found a few issues with the implementation and reasoning that need addressing before merge:

apps/web/src/index.css
Nagendhra-web: Right, that was a real bug. In c192bba the deploy dialog at FileViewer.tsx:4740 now also carries .deploy-flow-modal, and the CSS keys off that class instead of the shared .deploy-modal wrapper. The template-save dialog at :4682 is untouched.

apps/web/src/index.css
```diff
@@ -9368,6 +9368,19 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
...
+   vertical padding which is fine for shorter dialogs but reads as
+   cramped here because the deploy form ends in dense token / domain
+   config rows that push content right up to the border. Scoped to the
+   deploy dialog only (the template-save dialog also carries the
+   .deploy-modal class on its root, which is why we key off the
+   purpose-specific .deploy-flow-modal hook instead). 16px matches the
+   bump proposed for the global rule in #957 so the rhythm stays close
+   to the rest of the app. Issue #913. */
+.deploy-flow-modal .modal-foot {
+  padding-top: 16px;
```
mrcfps, line 9382: `apps/web/src/index.css:1190` still defines the shared footer spacing as `padding: 12px var(--modal-padding)`, but this dialog never provides `--modal-padding` — the only definition in the stylesheet is on `.modal-settings` at `apps/web/src/index.css:1130`, and the deploy dialog here is just `className="modal deploy-modal deploy-flow-modal"` (`apps/web/src/components/FileViewer.tsx:4740`). In browsers that makes the shared shorthand invalid, so this new rule is the only footer padding that survives: top becomes 16px, but left/right/bottom fall back to `0` instead of keeping the shared 12px spacing described in the comment and PR summary. That matters because the footer box model is still under-defined on the deploy modal, so the action row can remain flush on three sides even after this fix. Please either give this dialog a concrete `--modal-padding` value or replace the footer padding here with an explicit full declaration such as `padding: 16px 22px 12px` so the deploy footer spacing is fully specified.

mrcfps (commented):
@Nagendhra-web Thanks for tightening the deploy modal spacing — I verified the selector scoping fix on the current head and found one remaining footer-padding issue in the CSS cascade.


lefarcen (approved):
Great work addressing all the feedback, @Nagendhra-web! 🎉

All three issues from the previous review are now fixed:

✅ **Selector scope**: Added `deploy-flow-modal` class to scope the CSS to only the deploy modal
✅ **Comment accuracy**: Removed the `--modal-padding` reference and clarified the actual behavior  
✅ **Value justification**: Changed to 16px with clear rationale (matches #957's proposed rhythm)

The approach is clean and well-documented. The CSS change is simple, browser-safe, and the scoping is now correct. Deferring final merge decision to a maintainer.

Thanks for the quick iteration!

PR #993, opened 2026-05-09
fix(web): hide the Save comment button on Pods (issue #792)

Nagendhra-web:
Closes #792.

The board comment popover already gated `Save comment` behind `disabled={target.selectionKind === 'pod' || !draft.trim()}`, which meant on a Pods selection the button was visible but greyed out. That reads like a normal "needs input" disabled state and gives users no signal that the operation is unsupported here, which is exactly the misleading-affordance pattern #792 asked us to clean up.

Now we drop rendering the button when `selectionKind === 'pod'` so the popover footer collapses to `Add note` + `Send to chat`, which are the affordances Pods actually uses. The disabled prop on the remaining call site only carries the draft-empty constraint, since the pod check is no longer needed.

Source: https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/FileViewer.tsx

### What I checked
- `pnpm guard` clean.
- `pnpm --filter @open-design/web test` (568 passed).
- Visual check that an Inspect comment popover still shows Save comment, and a Pods popover only shows Add note + Send to chat with the same layout rhythm as before.

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -1419,14 +1419,16 @@ function BoardComposerPopover({
         >
           Add note
         </button>
-        <button
-          type="button"
-          className="ghost"
-          disabled={target.selectionKind === 'pod' || !draft.trim()}
-          onClick={() => void onSaveComment()}
-        >
-          Save comment
-        </button>
+        {target.selectionKind === 'pod' ? null : (
+          <button
+            type="button"
+            className="ghost"
+            disabled={!draft.trim()}
+            onClick={() => void onSaveComment()}
+          >
+            Save comment
+          </button>
+        )}
         <button
           type="button"
           className="primary"


mrcfps (approved):
@Nagendhra-web I reviewed the `BoardComposerPopover` change in `apps/web/src/components/FileViewer.tsx` and verified that Pods selections now remove the unsupported `Save comment` action while the non-Pods save path and shared draft gating stay intact. Thanks for tightening this affordance and making the Pods flow clearer. 🙌


apps/web/src/components/FileViewer.tsx
```diff
@@ -1419,14 +1419,16 @@ function BoardComposerPopover({
...
         </button>
-        <button
-          type="button"
-          className="ghost"
-          disabled={target.selectionKind === 'pod' || !draft.trim()}
-          onClick={() => void onSaveComment()}
-        >
-          Save comment
-        </button>
+        {target.selectionKind === 'pod' ? null : (
```
lefarcen, line 1422: **P3 — Reasoning completeness (optional):** The change is correct, but the PR body doesn't explicitly say why hiding is better than showing a disabled button with a tooltip explaining "not supported in Pods." #792 asked to remove the misleading affordance, which this does. If you want to round out the rationale, one sentence like: "Hiding the button (rather than disabling with tooltip) avoids advertising an unavailable workflow" would close the reasoning loop.

Not blocking — the code change is clean.

lefarcen (commented):
Hey @Nagendhra-web! 👋

Clean fix — the change correctly gates the `Save comment` button behind `selectionKind !== 'pod'`, removing the misleading affordance #792 reported. The disabled prop now only carries the draft-empty constraint, which is exactly the right scope.

**Code review (Lens A):**
- Correctness ✅: `target.selectionKind === 'pod'` is the right discriminator; `PreviewCommentSelectionKind` only has `'element' | 'pod'`.
- Architecture ✅: Fits existing pod-specific handling pattern in this component.
- Tests: Your PR description covers validation (`pnpm guard`, typecheck, web tests, visual check). A focused regression test for this conditional would be nice-to-have but not blocking for an XS bugfix.

**Reasoning completeness (Lens B):**
- One optional clarification: the PR body doesn't explicitly compare hiding vs. disabled-with-tooltip. #792's expected behavior asks to remove the unsupported action, which this does. If you want to add one sentence like "I chose hiding over a disabled tooltip because Pods has no save-comment capability, so showing the action still advertises an unavailable workflow," that rounds out the rationale.

No blocking issues. Ship as-is or with the optional reasoning polish.

— open-design team

PR #994, opened 2026-05-09
fix(web): clear stale upload error banner when previewing existing files

### LINKED ISSUE #786: Clear stale upload failure banner when previewing existing files in Design Files

Nagendhra-web:
Closes #786.

The upload-failure banner in FileWorkspace was rendered unconditionally inside `ws-body`, so once a failed upload set `uploadError`, the message stayed pinned above whatever file the user opened next. The banner is feedback about the Design Files upload surface, not about the active file viewer, so it reads as stale state once the user clicks an existing file in the list.

Now we scope the banner to `activeTab === DESIGN_FILES_TAB`. Previewing any file hides it, and returning to Design Files brings it back: the failed upload is still unresolved and we do not want to silently drop that signal entirely. Source: https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/FileWorkspace.tsx

### What I checked
- `pnpm --filter @open-design/web test` (568 passed).
- Visual check: trigger a failed upload in Design Files, confirm the banner appears, then click an existing file from the list and confirm the banner disappears above the preview. Click back to Design Files and the banner is still there.

### DIFF
--- a/apps/web/src/components/FileWorkspace.tsx
+++ b/apps/web/src/components/FileWorkspace.tsx
@@ -591,7 +591,28 @@ export function FileWorkspace({
         ) : null}
       </div>
       <div className="ws-body">
-        {uploadError ? <div className="viewer-empty">{uploadError}</div> : null}
+        {/* Keep the failure banner visible across tab switches so the
+            partial-success case (some files succeed and auto-open while
+            others fail) doesn't silently drop the failure signal. The
+            banner now carries an explicit dismiss button so the user
+            can clear the stale message themselves, which is what #786
+            was really asking for: a way to drop the message rather than
+            have it pinned above an unrelated file preview forever. The
+            next upload also clears it via setUploadError(null) at the
+            top of uploadFiles(). */}
+        {uploadError ? (
+          <div className="viewer-empty viewer-empty-dismissible">
+            <span>{uploadError}</span>
+            <button
+              type="button"
+              className="ghost"
+              aria-label={t('common.close')}
+              onClick={() => setUploadError(null)}
+            >
+              {t('common.close')}
+            </button>
+          </div>
+        ) : null}
         {activeTab === DESIGN_FILES_TAB ? (
           <DesignFilesPanel
             key={projectId}
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -8514,6 +8514,17 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
   color: var(--text-muted);
   font-size: 13px;
 }
+/* Compact horizontal variant for the upload-failure banner: the message
+   sits inline next to a dismiss button so the user can clear the stale
+   notice without changing tabs. Issue #786. */
+.viewer-empty.viewer-empty-dismissible {
+  padding: 12px 16px;
+  display: flex;
+  align-items: center;
+  justify-content: center;
+  gap: 12px;
+  text-align: left;
+}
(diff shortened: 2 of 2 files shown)


apps/web/src/components/FileWorkspace.tsx
```diff
@@ -591,7 +591,14 @@ export function FileWorkspace({
         ) : null}
       </div>
       <div className="ws-body">
-        {uploadError ? <div className="viewer-empty">{uploadError}</div> : null}
+        {/* Scope the upload-failure banner to the Design Files surface so
+            it does not hang above an unrelated file preview after the
+            user navigates to an existing file from the list. The banner
+            reappears when they return to Design Files since the failed
+            upload is still unaddressed state. Issue #786. */}
+        {uploadError && activeTab === DESIGN_FILES_TAB ? (
```
mrcfps: Because `uploadFiles()` auto-opens the last successful file before it sets `uploadError` (`apps/web/src/components/FileWorkspace.tsx:235-249`), this new `activeTab === DESIGN_FILES_TAB` guard suppresses the failure banner for the common partial-success case. If I upload N files and 1+ succeed while others fail, the UI navigates to the uploaded file and the user never sees the only feedback that some files were rejected unless they manually click back to Design Files. That silently drops the upload failure signal in the main path this PR is trying to preserve. Could we keep the failure visible for the upload that just finished — for example by using a toast/dismissible banner that survives the auto-open, or by only hiding the banner when the user manually switches away from Design Files instead of when `openFile()` does it for them?

mrcfps (changes requested):
@Nagendhra-web Thanks for tightening up the stale-banner behavior — I found one blocking regression in the new gating logic that looks worth fixing before merge.


mrcfps (approved):
@Nagendhra-web I reviewed the updated FileWorkspace banner changes and the current changed ranges look solid: the upload failure state stays visible through partial-success auto-open flows, and the new dismiss action gives users a direct way to clear stale banner state without losing the next-upload reset path. Thanks for iterating on the regression quickly — this looks ready to merge.

PR #995, opened 2026-05-09
fix(daemon): make the connector Close window button always give feedback

Nagendhra-web:
Closes #669.

When the connector callback page is reached via direct navigation rather than the daemon's popup-driven flow (linked from a different tab, opened in a new window by the OAuth provider, etc.) there is no `window.opener` and any `window.close()` call is silently rejected by the browser. The previous handler called close anyway and only flipped the manual-close hint after a 250ms timer gated on `visibilityState === 'visible'`. If the visibility check ever read `'hidden'` (the popup losing focus during the close attempt is enough), the hint never updated and the button looked completely dead.

The new code detects the no-opener case up front. With no live opener, the click skips the no-op close call and shows the manual-close instructions immediately, so the user sees instant feedback. When there is an opener, the hint update is scheduled unconditionally after the close attempt: if close actually worked the page is unloading and the timer never runs, otherwise the hint reliably surfaces.

Source: https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/connectors/routes.ts

### What I checked
- `pnpm guard` and `pnpm --filter @open-design/daemon typecheck` clean.
- `pnpm --filter @open-design/daemon exec vitest run tests/connectors` (75 passed).
- Walked the script logic by hand for the three observable states: popup with live opener (auto + manual close attempt), popup whose opener has been closed (manual hint immediately), direct tab nav (manual hint immediately).

### DIFF
--- a/apps/daemon/src/connectors/routes.ts
+++ b/apps/daemon/src/connectors/routes.ts
@@ -476,17 +476,34 @@ function renderConnectorConnectedHtml(connectorId: string): string {
           closeButton.textContent = 'Close this tab manually';
           hint.textContent = 'Your browser blocked automatic closing. You can close this tab and return to Open Design.';
         }
+        function hasLiveOpener() {
+          try {
+            return Boolean(window.opener) && !window.opener.closed;
+          } catch {
+            return false;
+          }
+        }
         function requestClose() {
+          // window.close() is silently rejected by browsers when the tab
+          // was not opened by a script (no opener), so trying it from a
+          // direct navigation always looks like the button "did nothing".
+          // Skip the no-op call and surface the manual-close instructions
+          // immediately so the click visibly produces feedback. Issue #669.
+          if (!hasLiveOpener()) {
+            showManualCloseHint();
+            return;
+          }
           try {
             window.close();
           } finally {
-            window.setTimeout(() => {
-              if (document.visibilityState === 'visible') showManualCloseHint();
-            }, 250);
+            // If the page is still alive after the close attempt, the
+            // browser blocked it. Update the hint unconditionally; if
+            // close did succeed the page is unloading and this never runs.
+            window.setTimeout(showManualCloseHint, 400);
           }
         }
         try {
-          if (window.opener && !window.opener.closed) {
+          if (hasLiveOpener()) {
             window.opener.postMessage(message, '*');
             window.setTimeout(requestClose, 900);
           } else {


mrcfps (approved):
@Nagendhra-web I reviewed the connector callback close-window update in `apps/daemon/src/connectors/routes.ts` and verified the new opener guard/manual-close fallback matches the direct-tab, closed-opener, and popup flows described in the PR. I didn't find any correctness, safety, or maintainability issues in the changed range. Thanks for tightening up this UX edge case 🙂


lefarcen (commented):
Hey @Nagendhra-web! Thanks for tackling #669. The core fix — detecting no-opener cases up front and showing manual-close instructions immediately — is solid and addresses the reported "button does nothing" UX gap.

**Code review (Lens A):**

✅ The new `hasLiveOpener()` helper safely wraps `window.opener` access with try/catch, handling cross-origin restrictions and closed/replaced opener states.

✅ The `visibilityState` gate removal is the right move — that conditional timer was the root of the "button appears dead" symptom.

✅ Security: `postMessage(message, '*')` only sends connector id/label (pre-existing), no OAuth tokens/state leak. Using a concrete expected origin instead of `'*'` would be marginally stronger if the daemon knows it, but not a blocker for this fix.

**P3** — `hasLiveOpener()` as the gate for attempting `window.close()` may be slightly too strict: a script-opened popup can still be closeable even if `window.opener` is now closed, and some `noopener`/browser-specific flows may have no readable opener but still permit close. Consider: use opener liveness only for `postMessage`, but still attempt `window.close()` in all cases and let the unconditional 400ms fallback handle browsers that reject the call. That way you preserve the "close when browser permissions allow it" part of #669 without over-filtering.

**P3** — The new inline callback-page JS logic is not meaningfully exercised by the daemon's vitest suite (existing route tests assert returned HTML contains strings but don't execute `hasLiveOpener()` / `requestClose()` under live/closed/no-opener cases). A small jsdom or extracted script-behavior test would better pin the #669 regression, but this is a nice-to-have for future hardening, not a merge blocker.

**Verdict:**

The fix is production-ready as-is. The P3 suggestions above are optional refinements. Nice work on the thorough testing notes in the PR body — that level of verification is exactly what we want to see. 🙌

(Note: PR #682 is also open for #669 with a similar approach. Coordinating with @nmsn may avoid duplicate merges or conflicts.)

lefarcen:
Merged! 🎉 Thanks @Nagendhra-web for the contribution to open-design. Looking forward to the next one!

PR #997, opened 2026-05-09
fix(web): reset Settings content scroll on section change

Nagendhra-web:
Closes #634.

Switching settings sections (for example clicking About after scrolling around in Library or Pets) kept the previous `scrollTop` on the right-hand content panel, so the new section's header often landed below the fold and the panel read as half-loaded. The complaint in #634 was specifically about reaching About from a sidebar position near the bottom and finding the About content not fully visible.

Added a ref on `.settings-content` and a small `useEffect` that resets `scrollTop` to 0 whenever `activeSection` changes. The sidebar's own scroll position is left alone, so users who had to scroll the sidebar to reach a tab still see that tab in view after clicking it. Source: https://github.com/nexu-io/open-design/blob/main/apps/web/src/components/SettingsDialog.tsx

### What I checked
- `pnpm --filter @open-design/web test` (568 passed; SettingsDialog suite covers section switching).
- Visual check: open Settings, scroll Library, click About, confirm About lands at the top of the content panel. Repeat with Pets and Notifications.

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -535,6 +535,12 @@ export function SettingsDialog({
   const [showApiKey, setShowApiKey] = useState(false);
   const [languageOpen, setLanguageOpen] = useState(false);
   const [activeSection, setActiveSection] = useState<SettingsSection>(initialSection);
+  // Scroll the right-hand content pane back to the top whenever the user
+  // picks a different settings section. Without this, switching from a
+  // long section the user had scrolled (e.g. Library) into a short one
+  // (About) keeps the previous scrollTop, so the new section's header
+  // can land out of view and the panel reads as half-loaded. Issue #634.
+  const settingsContentRef = useRef<HTMLDivElement | null>(null);
   const [languageMenuRect, setLanguageMenuRect] = useState<DOMRect | null>(null);
   const [agentRescanRunning, setAgentRescanRunning] = useState(false);
   const [agentRescanNotice, setAgentRescanNotice] =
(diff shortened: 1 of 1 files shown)


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -560,6 +566,10 @@ export function SettingsDialog({
   useEffect(() => {
     setActiveSection(initialSection);
   }, [initialSection]);
+  useEffect(() => {
+    const el = settingsContentRef.current;
+    if (el) el.scrollTop = 0;
```
mrcfps, line 571: The new `scrollTop = 0` effect fixes the runtime behavior, but this PR does not add a regression test for the exact bug it is addressing. Because the bug depends on imperative DOM state surviving a section switch, a later refactor to `activeSection` handling or the `.settings-content` wrapper could silently reintroduce it while the existing About/Library rendering assertions still pass. Please add a focused `SettingsDialog` test in `apps/web/tests/components/SettingsDialog.execution.test.tsx` that seeds `.settings-content.scrollTop`, switches from a long section like Library to About, and asserts the content pane resets to `0` after the click.

mrcfps (commented):
@Nagendhra-web Thanks for fixing the settings-pane scroll reset. I found one follow-up that would make this regression safer to maintain.


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -560,6 +566,10 @@ export function SettingsDialog({
   useEffect(() => {
     setActiveSection(initialSection);
   }, [initialSection]);
+  useEffect(() => {
```
lefarcen, line 569: **P3** (nit) — The existing SettingsDialog suite covers section switching in general, but I don't see a regression assertion for this new imperative scroll side effect.

A small test that sets `.settings-content.scrollTop`, switches to another section (e.g., About), and asserts it returns to `0` would protect the exact #634 fix from silently regressing.

Not a blocker — just a nice-to-have for long-term stability.

lefarcen (commented):
Hey @Nagendhra-web!

The implementation is narrowly scoped and should fix #634 without security or performance concerns. The ref + useEffect pattern is standard React practice for imperative DOM manipulation.

One small suggestion below to strengthen test coverage for this exact scroll behavior — non-blocking.

Looks good overall! 🎉

lefarcen:
@mrcfps This PR's current head 633b634d is waiting on your review.

mrcfps (approved):

PR #999, opened 2026-05-09
feat(memory): auto-memory store with chat-protocol-aware extraction

pftom:
## Summary

- New auto-memory feature: markdown store at `<dataDir>/memory/` (one `.md` file per fact + a hand-edited `MEMORY.md` index), folded into the system prompt so cross-chat preferences, role, and ongoing-work context survive restarts.
- Two extractors share the surface: a heuristic regex pack (covers explicit `remember:` / `我是 X` markers in EN + zh) runs synchronously in the chat route; a small-model LLM extractor runs in the background after each turn. Both write through the same store and emit SSE events for live UI updates.
- LLM extractor speaks four protocols (anthropic / openai / azure / google). The chat handler now forwards the chat agent id to `extractWithLLM`, and `pickProvider` constrains its auto-pick to the chat's protocol family — a Claude Code chat will only consider `ANTHROPIC_API_KEY`, never silently fall through to an OpenAI key parked in media-config. When no matching key is configured the attempt records as `skipped: no-provider` instead of quietly switching vendors and producing surprising `openai/gpt-4o-mini` failures.
- Settings UI gets a Memory tab (entries, hand-edited index, extraction history with per-attempt phase / skip / failure / delete) and an inline `MemoryModelInline` picker that sits next to the chat model picker in both CLI and BYOK modes. The picker reuses the same `SUGGESTED_MODELS_BY_PROTOCOL` table and `Custom…` pattern as the chat picker (extracted into `apps/web/src/state/apiProtocols.ts`); the user only chooses the model id, the surrounding chat protocol/key/baseUrl/apiVersion ride along automatically.
- All 17 locales updated; new contract types live in `packages/contracts/src/api/memory.ts`.

## Test plan

- [x] `pnpm guard`
- [x] `pnpm --filter @open-design/web test` (45 files, 392 tests pass — includes the locales-parity test that enforces every key + placeholder is mirrored across all 17 locales)
- [ ] Manual smoke: in CLI mode with Claude Code, a chat turn with no `ANTHROPIC_API_KEY` records `Skipped: no-provider` (not the previous `Failed: openai/gpt-4o-mini · media settings`); switching the inline picker to a specific model commits an override that survives a daemon restart.
- [ ] Manual smoke: in BYOK mode, picking a memory model copies the chat protocol's key/baseUrl/apiVersion into the saved override; the daemon calls the configured provider directly.


mrcfps:
@pftom I'm holding off on generating review comments for nexu-io/open-design#999 because this pull request has merge conflicts right now.

mrcfps:
@pftom I'm holding off on generating review comments for nexu-io/open-design#999 because this pull request has merge conflicts right now.

apps/web/src/components/MemoryModelInline.tsx
```diff
@@ -0,0 +1,394 @@
...
+  }, [customDraft, persist, buildOverride]);
+
+  // CLI mode with no models advertised by the agent — fall back to a
+  // simple "Same as chat" / "Custom..." pair so the picker is still
+  // usable. (Some CLIs don't expose a `models` command; the chat
+  // picker shows the same fallback there.)
+  const showSuggestedOptions = modelOptions.length > 0;
+
+  return (
+    <label className="field">
```
mrcfps: Wrapping the entire control in one `<label>` makes the select's accessible name absorb the hint text too, so Playwright's existing `getByLabel('Model')` and even `getByLabel('API key')` locators now resolve to the memory-model select as well. That's the concrete reason `ui/settings-api-protocol.test.ts` is red in CI on this head, and it also makes the settings form harder to target correctly for assistive tech. Please switch this to a non-label wrapper and associate only the short title with the select/input via `htmlFor`/`id` or `aria-labelledby`, leaving the hint outside the labelling element.

apps/daemon/src/server.ts
```diff
@@ -5002,6 +5244,21 @@ export async function startServer({
...
+    // Auto-memory hook. Pulls explicit "remember:" / "我是 X" / "I prefer Y"
+    // markers out of the just-arrived user message and writes them as MD
+    // files under <dataDir>/memory/. We await so the very next
+    // composeSystemPrompt() call (a few lines below) re-reads memory from
+    // disk and a marker inside this turn's message is reflected in this
+    // turn's prompt. Failures are swallowed — memory is best-effort and
+    // must never block the agent run.
+    if (typeof message === 'string' && message.trim().length > 0) {
+      try {
+        await extractFromMessage(RUNTIME_DATA_DIR, message);
```
mrcfps: This hook only runs in the daemon chat path, but API/BYOK chats still compose prompts and stream entirely in `ProjectView` without ever calling `/api/memory/extract` or receiving `memoryBody`. That leaves the new Memory tab/model picker as a no-op for BYOK users even though the PR summary and test plan explicitly say that mode is supported. Please route API-mode sends through the same memory injection/extraction flow, or plumb the equivalent logic into the web-side path before merging.

apps/daemon/src/memory.ts
```diff
@@ -0,0 +1,707 @@
...
+// ----- System-prompt body -------------------------------------------------
+
+// Build the markdown block that the prompt composer folds into every
+// run. Returns `''` when memory is disabled, missing, or empty so the
+// composer can drop the block without an extra `if`.
+export async function composeMemoryBody(dataDir) {
+  const cfg = await readMemoryConfig(dataDir);
+  if (!cfg.enabled) return '';
+  const entries = await listMemoryEntries(dataDir);
+  if (entries.length === 0) return '';
```
mrcfps: The feature copy says removing a link from `MEMORY.md` should stop that fact from being injected while keeping the file on disk, but `composeMemoryBody()` still starts from `listMemoryEntries()`, i.e. every `.md` file in the directory. That means edited-out memories continue to influence every new chat, so the index editor cannot actually disable a fact. Please derive the active set from the parsed `MEMORY.md` links here and ignore unlinked files when building the prompt body.

mrcfps (changes requested):
@pftom Thanks for shipping such a large memory pass — I found a few blockers that need to be addressed before this is safe to merge.


apps/daemon/src/memory-llm.ts
```diff
@@ -0,0 +1,767 @@
...
+  }
+  const json = await resp.json();
+  const block = (json?.content || []).find((b) => b?.type === 'text');
+  return block?.text ?? '';
+}
+
+async function callOpenAI(provider, system, user) {
+  let resp;
+  try {
+    resp = await fetch(`${provider.baseUrl}/v1/chat/completions`, {
```
mrcfps: This still hard-codes another `/v1` onto every OpenAI-compatible base URL. In this PR, the BYOK memory picker persists `chatBaseUrl` verbatim, and the existing settings flow/tests already use versioned values like `https://api.openai.com/v1` and `https://api.deepseek.com/v1`. With those inputs this call becomes `/v1/v1/chat/completions`, so the new background memory extraction fails even though the same provider config works for chat. Please normalize the base URL the same way the main OpenAI-compatible request path does: strip or detect a trailing `/v1` once, then append only `/chat/completions` (or reuse the shared URL helper) so both versioned and unversioned endpoints work.

mrcfps (changes requested):
@pftom Thanks for the follow-up here — the accessibility/test fix looks good, but this head still has the two unresolved blockers in `apps/daemon/src/server.ts` (BYOK chats still bypass memory injection/extraction) and `apps/daemon/src/memory.ts` (`MEMORY.md` edits still do not control what gets injected), and there is one additional provider-path regression below.


apps/web/src/components/MemoryModelInline.tsx
```diff
@@ -0,0 +1,414 @@
...
+    (modelId: string): MemoryExtractionConfigShape => {
+      const trimmedModel = modelId.trim();
+      if (mode === 'api') {
+        return {
+          provider: apiProtocol,
+          model: trimmedModel,
+          baseUrl: chatBaseUrl.trim(),
+          apiKey: chatApiKey,
+          apiVersion: apiProtocol === 'azure' ? chatApiVersion.trim() : '',
+        };
```
mrcfps, line 237: `buildOverride()` snapshots `apiProtocol`, `chatApiKey`, `chatBaseUrl`, and `chatApiVersion` into the saved memory override only when the user touches this picker. If they pick a memory model first and then change the surrounding BYOK provider, rotate the API key, or edit the base URL later in the same settings flow, `/api/memory/config` is never patched again, so background extraction keeps calling the stale vendor/credential even though this component promises to "borrow the surrounding chat picker's protocol, key, base URL, and api-version automatically." The concrete evidence in this diff is that the only writes happen in `onSelectChange`/`onSaveCustom`, while there is no effect that re-persists an existing API-mode override when those chat props change. Please either resync the saved override whenever the BYOK chat config changes while a memory override is active, or store only the model choice and let the daemon derive provider/key/baseUrl/apiVersion from the live chat config at extraction time.

mrcfps (commented):
@pftom Thanks for tightening up the memory follow-ups — I found one merge-safe correctness issue in the BYOK memory-model flow worth addressing before this behavior settles in.


apps/daemon/src/server.ts
```diff
@@ -3187,6 +3207,270 @@ export async function startServer({
...
+        : await extractFromMessage(RUNTIME_DATA_DIR, userMessage);
+      let attemptedLLM = false;
+      if (userMessage.trim().length > 0 && hasAssistant) {
+        attemptedLLM = true;
+        void import('./memory-llm.js')
+          .then(({ extractWithLLM }) =>
+            extractWithLLM(
+              RUNTIME_DATA_DIR,
+              { userMessage, assistantMessage },
+              { projectRoot: PROJECT_ROOT, chatAgentId: null },
```
mrcfps: Leaving the BYOK picker on "Same as chat" still breaks the default LLM extraction path here. In the new web flow, `MemoryModelInline` clears the override for that option, and both `/api/memory/extract` calls in `ProjectView` only send the messages, so this branch always reaches `extractWithLLM(..., { chatAgentId: null })` with no BYOK provider/key/baseUrl/apiVersion context. `pickProvider()` then falls back to env/media-config instead of the live chat settings, which means a BYOK chat that works for inference can still skip memory extraction or hit the wrong vendor unless the user manually picks a separate memory model. Please thread the active BYOK provider config through this endpoint (or persist "Same as chat" as a concrete override instead of `null`) so the default path actually follows the chat configuration.

mrcfps (commented):
@pftom Thanks for tightening up the memory follow-ups — I found one merge-safe BYOK default-path issue worth addressing before this behavior settles in.


pftom:
@mrcfps Pinging for re-review — all four review threads should be addressed on the latest head (`545a0225`):

1. **MemoryModelInline `<label>` wrapper** — switched to `<div>` + `aria-labelledby` so only the short title is the labelling element; hint and flash sit outside.
2. **`composeMemoryBody` ignoring `MEMORY.md`** — now derives the active set from parsed `INDEX_LINK_RE` bullets and ignores unlinked files.
3. **`callOpenAI` hard-coded `/v1`** — both `callOpenAI` and `callAnthropic` go through the same conditional `appendVersionedApiPath` helper the chat proxy / connection-test routes use, so versioned base URLs no longer become `/v1/v1/…`.
4. **BYOK memory wiring** — three layers:
   - `composeSystemPrompt` (contracts) takes `memoryBody`; daemon exposes `GET /api/memory/system-prompt`; `ProjectView` folds the body into the BYOK system prompt before each turn.
   - `POST /api/memory/extract` is now two-phase (heuristic pre-turn, LLM post-turn, no double records); `ProjectView` calls both phases on BYOK turns.
   - **"Same as chat" default**: `ExtractMemoryRequest.chatProvider` threads the live BYOK provider/apiKey/baseUrl/apiVersion through to `pickProvider()` on every API-mode call, so the default extraction path follows the chat config (and uses the per-protocol fast-model default — `gpt-4o-mini` / `claude-haiku-4-5` — instead of paying chat-tier rates). Recorded as `credentialSource: 'chat-byok'` in extraction history.
   - **Override drift**: when an explicit memory model is set, a debounced effect in `MemoryModelInline` re-PATCHes `/api/memory/config` whenever `apiProtocol` / `apiKey` / `baseUrl` / `apiVersion` change, using `apiKeyTail` to detect rotation without round-tripping the secret.

Mergeable, all CI green, `pnpm guard` + `pnpm --filter @open-design/web typecheck` + `pnpm --filter @open-design/daemon typecheck` + `pnpm --filter @open-design/web test` (596/596) all pass locally.

apps/daemon/src/server.ts
```diff
@@ -3187,6 +3207,301 @@ export async function startServer({
...
+      if (Object.prototype.hasOwnProperty.call(body, 'extraction')) {
+        if (body.extraction === null) {
+          patch.extraction = null;
+        } else if (body.extraction && typeof body.extraction === 'object') {
+          const incoming = body.extraction;
+          const current = await readMemoryConfig(RUNTIME_DATA_DIR);
+          const reuseKey =
+            current.extraction
+            && current.extraction.provider === incoming.provider
+            && (incoming.apiKey === undefined || incoming.apiKey === '')
```
mrcfps: Because `MemoryModelInline` now re-PATCHes `/api/memory/config` whenever the BYOK chat creds drift, clearing the surrounding chat API key sends `apiKey: ""` back through this handler. The new reuse branch treats that explicit empty string the same as an omitted field, so the old `current.extraction.apiKey` is kept in `.config.json` and background memory extraction keeps calling the provider even after the user removed the key from the chat settings. That breaks the new "borrow the surrounding chat picker's key automatically" contract and can keep generating requests on a stale credential unexpectedly. Please only reuse the stored key when `apiKey` is absent, and treat an explicit empty string as a real clear (or add a separate preserve/clear signal from the client).

mrcfps (commented):
@pftom Thanks for tightening up the memory follow-ups — I found one merge-safe credential-sync issue in the new BYOK memory override flow.


apps/daemon/src/server.ts
pftom: Good catch — fixed in 88eb886c.

The PATCH handler now distinguishes four apiKey states:

- **field absent** → preserve the stored secret (settings re-save without re-typing)
- **`apiKey === ''`** → CLEAR the stored secret (this is exactly the drift-resync case you flagged: when the user clears the BYOK chat key, `MemoryModelInline` sends `apiKey: ''`, and the daemon now respects that as an intentional clear)
- **`apiKey === 'sk-...'`** → replace
- **provider differs** → ignore the stored key entirely

Added `apps/daemon/tests/memory-config-route.test.ts` with regression coverage for all four cases (preserve / clear / replace / provider-change).

pftom:
@mrcfps Pushed `88eb886c` for the credential-sync issue you flagged on `apps/daemon/src/server.ts:3268`.

**What changed**

The PATCH `/api/memory/config` handler now distinguishes four apiKey states instead of conflating absence with empty string:

| Incoming | Behavior |
|---|---|
| field absent | preserve stored secret (settings re-save without re-typing) |
| `apiKey: ''` | **clear** the stored secret — fires when the user removes the BYOK chat key from the picker, `MemoryModelInline`'s drift-resync effect sends `''`, and the daemon now stops calling the provider |
| `apiKey: 'sk-...'` | replace |
| provider differs | ignore stored key entirely (no cross-provider key reuse) |

**Tests**

Added `apps/daemon/tests/memory-config-route.test.ts` with 4 cases pinning the contract:

- preserves the stored apiKey when the patch omits the field entirely
- clears the stored apiKey when the patch sends an explicit empty string
- replaces the stored apiKey when the patch sends a new value
- does not reuse the stored apiKey when the provider changes

`pnpm guard` and `pnpm typecheck` both green; new tests pass locally.

Could you take another look when you have a moment?

apps/daemon/src/server.ts
mrcfps: Looper checked this thread against head `88eb886c90cc2c16ff883c684a9207e9b4dd4251`. The requested change appears objectively addressed by apps/daemon/src/server.ts:3278-3289 now preserves the stored key only when extraction.apiKey is omitted; an explicit empty string leaves nextApiKey as '', clearing the stored credential. Regression tests in apps/daemon/tests/memory-config-route.test.ts:63-156 cover omit/preserve, explicit empty-string clear, replace, and provider-change cases., so I’m resolving this thread. Reopen if this still needs discussion.



mrcfps (commented):
@pftom Thanks for tightening up the memory follow-ups — I found one merge-safe contract mismatch worth fixing before more callers build on this API.


Location: packages/contracts/src/api/memory.ts RIGHT line 120

`MemoryConfigResponse.extraction` is typed as `MemoryExtractionConfig | null`, but the route this PR adds at `apps/daemon/src/server.ts:3311-3315` actually returns `maskMemoryExtractionConfig(next.extraction)`, i.e. `{ apiKeyTail, apiKeyConfigured }` rather than the raw `apiKey` fields from `MemoryExtractionConfig`. That leaves the new contracts package out of sync with the live API: any future caller using `MemoryConfigResponse` will either miss the masked fields the daemon really sends or assume secrets round-trip when they do not. Please change this response contract to `MemoryExtractionMaskedConfig | null` (or export a dedicated PATCH-response type that matches the daemon) so `packages/contracts` stays authoritative for the memory config endpoint.

mrcfps:
@pftom I'm holding off on generating review comments for nexu-io/open-design#999 because this pull request has merge conflicts right now.

mrcfps:
@pftom I'm holding off on generating review comments for nexu-io/open-design#999 because this pull request has merge conflicts right now.

pftom:
@mrcfps friendly ping — resolved the latest conflicts with `main` (incl. #1205 / #1203 / #1140). Merge commit `2d3cad15`. Local `pnpm guard` + `pnpm typecheck` + the affected SettingsDialog test suite all pass. Could you take another look when you have a moment? 🙏

mrcfps (commented):
@pftom Thanks for pushing the memory follow-ups — I found one merge-safe contract mismatch worth fixing before more callers build on this API.


Location: packages/contracts/src/api/memory.ts RIGHT lines 125-127

`MemoryConfigResponse.extraction` is still typed as `MemoryExtractionConfig | null`, but the PATCH handler this PR adds at `apps/daemon/src/server.ts:3661-3715` returns `maskMemoryExtractionConfig(next.extraction)`, i.e. `{ apiKeyTail, apiKeyConfigured }` rather than the raw `apiKey` field described by `MemoryExtractionConfig`. That leaves the contracts package out of sync with the live API: any caller that trusts this type will either miss the masked fields the daemon really sends or assume secrets round-trip when they do not. Please change this response contract to `MemoryExtractionMaskedConfig | null` (or introduce a dedicated PATCH-response type that matches the masked server payload) so `packages/contracts` stays authoritative for `/api/memory/config`.

pftom:
@mrcfps another conflict surfaced after #1043 (daemon route-registration split) landed — re-merged in `f3fb1479`. Resolution: dropped the inline route block from pie-lens since #1043 now wires routes via `registerXxxRoutes(app, deps)`, but kept the new `/api/memory*` handlers inline (no `memory-routes.ts` module exists yet — happy to extract them in a follow-up if you'd prefer). `pnpm guard` / `pnpm typecheck` / `pnpm --filter @open-design/daemon build` all clean.

lefarcen (approved):

mrcfps (approved):

Issue #890, opened 2026-05-08
Inspect and Picker silently do nothing on generated artifacts without data-od-id


fuyizheng3120:
### Observed behavior
After asking the agent to turn a plain PRD markdown into an HTML mockup (no specific skill selected — just free-form generation via a Claude-Code-compatible CLI), the resulting HTML preview renders fine. However:

- Clicking the **Inspect** button, then clicking any element in the sandbox, does nothing. The inspect panel never opens.
- Toggling **Tweaks** mode and clicking **Picker**, then clicking any element, also does nothing.

No error in devtools, no visible feedback; the cursor does turn into crosshair in Picker mode but clicks are ignored.

### Expected behavior
Either:
- Inspect / Picker should work on any element, falling back to a synthetic id when `data-od-id` / `data-screen-label` is missing, OR
- The UI should surface a clear hint (e.g. "this artifact has no inspectable elements — ask the agent to add `data-od-id`") so the user doesn't think the feature is broken.

### Root cause (for context)
`apps/web/src/runtime/srcdoc.ts` — `targetFrom(el)` bails out silently if the element has neither `data-od-id` nor `data-screen-label`:

\`\`\`js
function targetFrom(el){
  var id = el.getAttribute('data-od-id') || el.getAttribute('data-screen-label');
  if (!id) return null;
  ...
}
\`\`\`

Several built-in skills (web-prototype, mobile-app, magazine-poster, gamified-app, etc.) already instruct the model to add `data-od-id` per-section. But when the user doesn't go through a skill — or the skill forgot to enforce it — the contract silently breaks from the user's POV.

### Suggested fixes (non-exclusive)
1. Surface an empty-state hint in the viewer when `allTargets().length === 0` and the user just activated Inspect / Picker.
2. Fall back to a computed id based on CSS selector path so any element becomes inspectable (with a warning that overrides won't persist if the DOM shifts).
3. Lift the `data-od-id` requirement into the base system prompt (`apps/daemon/src/prompts/official-system.ts`) so every artifact, including ad-hoc ones, gets it by default.

### Environment
- Open Design 0.5.0 (packaged macOS arm64)
- Agent: Claude-Code-compatible CLI
- macOS 15.x (Sequoia)

fuyizheng3120:
Follow-up — a related scenario that shares the same root cause is worth flagging, because it looks like a *different* bug at first sight but isn't:

### Scenario: parent element has `data-od-id`, target child does not

[...]

PR #1005, opened 2026-05-09
fix(web): surface empty-annotation state for Inspect/Picker (#890)

Sid-Qin:
## Summary

Addresses the **zero-annotation** scenario from #890: when the agent emits an HTML artifact with no \`data-od-id\` / \`data-screen-label\` annotations (a freeform PRD → HTML pass through a Claude-Code-compatible CLI without going through a skill), the existing Inspect / Picker affordances no-oped silently and the static \"Click any element with \`data-od-id\`\" copy was misleading.

The **partial-annotation** scenario also documented in #890 (parent has \`data-od-id\`, child doesn't, click resolves to the wrapper instead of the actual click target) is tracked as **#1022** as a follow-up — fixing it cleanly needs either a new bridge signal or a synthetic-id fallback that touches save-to-source persistence, both bigger changes than the empty-state UX patch this PR ships.

## What changed

Two surgical fixes in \`FileViewer.tsx\`:

1. **Dedicated \`od:comment-targets\` listener** that installs whenever Inspect OR Comments mode is active, mirroring the bridge's broadcast into \`liveCommentTargets\`. The comment-mode-only listener still owns its hover / click / pod events; this new listener only handles the targets list.
2. **Inspect-empty-hint dispatch** keyed off a new \`targetsState: 'unknown' | 'empty' | 'present'\` tri-state:
   - \`'empty'\` (after the iframe authoritatively reported zero targets): \`This artifact has no <code>data-od-id</code> annotations yet — ask the agent to add them to the sections you want to inspect/comment on.\` (mode-aware verb)
   - \`'unknown'\` (initial / on srcDoc rebuild) or \`'present'\`: existing \`Click any element with <code>data-od-id</code> to tune its style / leave a comment.\`

Mirrored across Inspect AND Picker modes; \`activateBoard()\` resets the hint dismissal so a user who dismissed the Inspect hint earlier still sees the Picker hint on a later visit.

Bridge contract is unchanged — \`targetFrom()\` still requires \`data-od-id\` / \`data-screen-label\`. Synthetic-id fallback would change save-to-source semantics: the persisted CSS in \`<style data-od-inspect-overrides>\` keys off \`elementId\`, and a synthetic selector path (\`#root > div:nth-child(2) > h1\`) drifts the moment the agent regenerates the artifact or the user manually edits the markup. Without new validation/migration logic at save-to-source time, overrides written against synthetic ids would silently break on the next mutation. That tradeoff belongs in a separate PR (tracked at #1022).

## Why this cut

Option 1 from the issue (\"surface an empty-state hint\") gives the user the missing signal without changing the underlying contract. Options 2 and 3 (synthetic-id fallback / lift \`data-od-id\` into the base system prompt) are bigger changes that touch save-to-source persistence and per-skill authoring conventions; flagging in the design comment so they're tracked as follow-ups, not blocked here.

## Test plan

- [x] \`pnpm i18n:check\` — passes.
- [x] \`pnpm typecheck\` (web, with rebuilt contracts) — zero errors.
- [x] \`pnpm test\` (web) — 585 / 585 tests pass across 65 files. New tests:
  - \`tests/runtime/srcdoc-bridge-empty-targets.test.ts\` (3 cases): runs the bridge IIFE in jsdom and pins (a) \`allTargets()\` posts an empty list for unannotated DOM, (b) clicks on unannotated elements do NOT post \`od:comment-target\` (regression pin against future synthetic-id fallbacks), (c) clicks DO still resolve to an annotated ancestor when one exists.
  - \`tests/components/FileViewer.inspect-empty-hint.test.tsx\` (6 cases): empty-state copy in Inspect, switch to instructive copy when targets show up, empty-state mirror in Picker mode, **targetsState race pin** (no flash of empty-state on annotated artifacts before the bridge broadcasts), **Picker dismissal pin** (Inspect dismissal doesn't suppress Picker hint), **mode-symmetric copy pin** (entering Tweaks while Inspect is still active doesn't leak Inspect verbs into the Picker banner).

## Manual repro

1. \`pnpm --filter @open-design/daemon dev\`, \`pnpm --filter @open-design/web dev\`.
2. Open a project, paste a plain HTML mockup with no \`data-od-id\` (the issue's PRD-style example works).
3. Click **Inspect** in the toolbar. Old behavior: \"Click any element with \`data-od-id\` to tune its style.\" + clicks no-op. New: \"This artifact has no \`data-od-id\` annotations yet — ask the agent to add them to the sections you want to inspect.\"
4. Toggle off Inspect, click **Tweaks** (Picker is the default sub-tool). Same banner with mode-appropriate copy (\"comment on\" instead of \"inspect\").
5. Now paste a mockup with \`data-od-id=\"hero\"\` somewhere. The banner switches to the existing instructive copy and clicking the annotated element opens the Inspect / Comment composer as before.

## Out of scope (flagged for follow-ups)

- **Partial-annotation scenario** — the second flow from #890 (parent annotated, child clicked → wrapper edited). Tracked at **#1022**. Needs either a bridge signal for ancestor-fallback or a synthetic-id fallback that doesn't break save-to-source persistence.
- i18n: the existing inspect-empty-hint copy is hardcoded English; rolling it into the 17-locale Dict is a separate cleanup.


mrcfps (approved):
@Sid-Qin I reviewed the FileViewer listener split and the new regression coverage around empty annotation targets, and the changed ranges look consistent with the existing inspect/comment bridge behavior. Thanks for tightening the empty-state UX here and pinning the contract with focused tests.


lefarcen (commented):
Hi @Sid-Qin! Thanks for this PR — the empty-state calibration signal is a real usability win for freeform HTML generation, and the test coverage is thorough. 👍

I've run a deep review (code correctness + reasoning completeness) and found 5 items to tighten before merge. Two are **P2 gaps** that can leave the user confused in edge cases; the rest are refinements.

## Code correctness (Lens A)

### P2 — Picker hint dismissal not persisting

**Location**: `apps/web/src/components/FileViewer.tsx` L4704 (the new `(inspectMode || (boardMode && boardTool === 'inspect'))` condition)

**Issue**: Picker reuses `openHintBox`, but that state is only reset when enabling Inspect (`setOpenHintBox(true)` at L4286-4293). If a user dismisses the Inspect hint once, then later enters Tweaks/Picker on an unannotated artifact, the new empty-state hint never appears, so #890 still reproduces for Picker.

**Fix**: Reset the hint when entering `boardMode` with `boardTool === 'inspect'`, or track dismissal per mode/file instead of one global boolean. Add a test that closes the hint in Inspect, switches to Picker, and confirms the empty-state copy still appears.

---

### P2 — "No targets" shown before bridge broadcasts

**Location**: `apps/web/src/components/FileViewer.tsx` L3200-3250 (the new useEffect that mirrors `od:comment-targets`)

**Issue**: `liveCommentTargets.size === 0` conflates "the current iframe reported zero targets" with "we have not heard from this iframe yet." The map starts empty and is not keyed to `srcDoc`/source generation, so annotated artifacts can briefly show the "no annotations" error before the bridge broadcasts, and source changes can show stale target state until the next iframe message.

**Fix**: Track an explicit status like `targetsState: 'unknown' | 'empty' | 'present'` plus a frame/source generation, reset to `unknown` on `srcDoc` changes, and only render the no-target empty state after an `od:comment-targets` message from the current frame.

---

### P3 — Mode wording mismatch when Inspect + Tweaks overlap

**Location**: `apps/web/src/components/FileViewer.tsx` L4219-4290 (Tweaks activation)

**Issue**: Entering Tweaks while Inspect is active does not clear `inspectMode`, but the new shared hint branches on `inspectMode ? 'inspect' : 'comment on'`. That can leave both bridge modes active and show Inspect wording while the user is in Picker.

**Fix**: Make mode activation symmetric by clearing `inspectMode` inside the board/Tweaks activation path, or derive the hint copy from the visible active surface instead of the raw boolean.

---

## Reasoning completeness (Lens B)

### P2 — Partial-annotation scenario not covered

**Location**: PR body "Closes #890"

**Issue**: The PR body says this "Closes #890", but issue #890's comments also document the **partial-annotation** scenario (some elements have `data-od-id`, clicked child does not → click edits the wrong ancestor). This implementation only fixes the **all-targets-empty** case; partial annotations will still show "Click any element" (because `liveCommentTargets.size > 0`) while the user clicks an unannotated child and gets a no-op or unintended parent edit.

**Fix**: Either (a) narrow the Closes claim to the zero-annotation bug and open/link a follow-up for partial annotations, or (b) add a minimal bridge signal for click-miss/ancestor-fallback so the UI can explain that case too.

---

### P3 — Synthetic-id rationale missing detail

**Location**: PR body "Why this cut" section

**Issue**: The section says synthetic-id fallback "would change save-to-source semantics" but doesn't spell out the invariant, making the scope cut hard to review.

**Fix**: Add 1-2 sentences explaining that Inspect persistence serializes CSS against stable `[data-od-id]` / `[data-screen-label]` selectors, and synthetic selector paths can drift after DOM edits (requiring source mutation or new validation rules). This makes the cut reviewable instead of relying on unstated assumptions.

---

Once the P2 items are addressed, this will be ready to ship. Let me know if any of the suggestions need clarification!

— open-design team

lefarcen:

Sid-Qin:
Pushed fixes for the 5 review items.

**P2 — Picker hint dismissal:** `activateBoard()` now calls `setOpenHintBox(true)`, mirroring what the Inspect-mode toggle already did. Dismiss-in-Inspect → enter-Picker-on-unannotated now re-renders the empty-state hint. New regression test pins this.

**P2 — "No targets" race:** new `targetsState: 'unknown' | 'empty' | 'present'` tri-state. Defaults to `'unknown'`, set authoritatively only after an `od:comment-targets` broadcast from the current iframe, reset to `'unknown'` on every `srcDoc` change. Empty-state copy now renders only when `targetsState === 'empty'`. Annotated artifacts no longer flash the wrong copy in the microsecond gap between mode-on and the bridge's first post; source/file changes no longer leave stale verdicts in place. New test pins the instructive default surviving until the iframe broadcasts.

**P3 — Mode wording overlap:** `activateBoard()` now calls `setInspectMode(false)` for symmetry with the Inspect activation handler that already calls `setBoardMode(false)`. Entering Tweaks while Inspect is still active no longer leaks Inspect verbs into the Picker banner. New test pins this — banner says "comment on", asserts `/\binspect\b/` is NOT in the body.

**P2 — Partial-annotation gap:** Narrowed the PR body's Closes claim to call out only the zero-annotation scenario this PR fixes, and opened **#1022** to track the partial-annotation flow separately. The follow-up writeup walks through three viable approaches (bridge signal for click-miss / synthetic-id fallback / UI affordance to descend) and notes the save-to-source persistence tradeoff that makes synthetic-id fallback a bigger change than the empty-state UX patch.

**P3 — Synthetic-id rationale detail:** Filled in the "Why this cut" section with the missing invariant — Inspect override CSS in `<style data-od-inspect-overrides>` keys off `elementId`, so a synthetic selector path (`#root > div:nth-child(2) > h1`) drifts the moment the agent regenerates the artifact or the user edits the markup. Either we accept the drift (overrides float off after edits) or we add validation/migration logic at save-to-source time. That belongs in a separate PR (the one tracked at #1022).

`pnpm test` 585/585 (added 3 regression tests on top of the original 6); typecheck clean; `pnpm i18n:check` passes.

lefarcen:
Nice work @Sid-Qin! 🎉 All 5 items comprehensively addressed:

✅ **P2 — Picker hint dismissal**: `activateBoard()` now mirrors the Inspect path (`setOpenHintBox(true)`), so dismissing in Inspect → entering Picker on unannotated artifacts re-renders the empty-state hint. Regression test pins this.

✅ **P2 — Stale targets race**: New `targetsState` tri-state (`'unknown' | 'empty' | 'present'`) defaults to `'unknown'`, set authoritatively only after an `od:comment-targets` broadcast from the current iframe, reset to `'unknown'` on every `srcDoc` change. Empty-state copy now renders only when `targetsState === 'empty'`. Annotated artifacts no longer flash wrong copy; source/file changes no longer leave stale verdicts. Test pins the instructive default surviving until iframe broadcasts.

✅ **P3 — Mode wording overlap**: `activateBoard()` now calls `setInspectMode(false)` for symmetry, so entering Tweaks while Inspect is active no longer leaks Inspect verbs into Picker banner. Test asserts banner says "comment on" with no `/\binspect\b/` in body.

✅ **P2 — Partial-annotation gap**: Narrowed PR body's Closes claim to the zero-annotation scenario this PR fixes, opened **#1022** to track partial-annotation flow separately with three viable approaches + save-to-source tradeoff analysis.

✅ **P3 — Synthetic-id rationale detail**: Filled "Why this cut" with the missing invariant — Inspect override CSS keys off `elementId`, so synthetic selector paths drift after agent regeneration/user edits. Requires validation/migration at save-to-source time → belongs in separate PR tracked at #1022.

3 new regression tests on top of original 6 → 585/585 pass; typecheck clean; i18n:check passes. The fixes tighten the empty-state signal and prevent future mode-overlap bugs. Ready from my side. 🚀

Issue #10, opened 2026-04-28
Windows: spawn agent fails with ENOENT despite successful PATH detection


Sun-sunshine06:
## Bug Description

On Windows, when the daemon successfully detects a local CLI agent (e.g., `claude`) via `/api/agents`, the subsequent `/api/chat` request fails with `spawn claude ENOENT`. This happens because the daemon probes for the full executable path during detection but then spawns the agent using only the bare binary name, which fails on Windows when the binary is not in the system's default PATH for child processes.

## Reproduction Steps

1. Clone the repo on Windows.
2. Install a supported agent CLI globally (e.g., `npm install -g @anthropic-ai/claude-code`).
3. Ensure `claude` is on your user PATH.
4. Start the daemon and frontend (`npm run dev:all`).
5. Open the UI and confirm the agent picker shows **"Claude Code" as available** (green / selectable).
6. Type any prompt and hit **Send**.
7. Observe the error: `spawn claude ENOENT`.

## Expected Behavior

If `/api/agents` reports an agent as `available: true` with a resolved `path`, `/api/chat` should successfully spawn that exact executable.

## Actual Behavior

`/api/chat` crashes immediately with `spawn claude ENOENT` because it passes the bare string `def.bin` (`'claude'`) to `spawn()` instead of the resolved absolute path.

## Root Cause

In `daemon/agents.js`, the `probe()` function resolves the full path and returns it as `path`:

```js
const resolved = resolveOnPath(def.bin);
// ...
return { ...stripFns(def), available: true, path: resolved, version };
```

However, in `daemon/server.js`, the `/api/chat` handler ignores this resolved `path` and spawns using only `def.bin`:

```js
child = spawn(def.bin, args, {
  env: { ...process.env },
  stdio: ['ignore', 'pipe', 'pipe'],
  cwd: cwd || undefined,
});
```

On Unix-like systems, `spawn('claude', ...)` often works because the shell inherits the user's PATH. On Windows, this is less reliable—especially when the daemon is launched from a context (e.g., IDE, `concurrently`, or a different shell) where the user's PATH modifications are not propagated.

## Suggested Fix

Cache the resolved agent paths returned by `detectAgents()` and use the cached absolute path when calling `spawn()` in `/api/chat`. For example:

1. In `daemon/server.js`, store a map of `agentId → resolvedPath` after the first `/api/agents` call (or eagerly at startup).
[...]

PR #1007, opened 2026-05-09
fix(daemon): version-aware Gemini binary resolution (#978)

### LINKED ISSUE #978: [Gemini CLI] DMG app picks up old gemini binary on PATH → `Unknown arguments: output-format, outputFormat`
## Summary

Open Design (DMG, v0.5.0) on macOS fails to start the Gemini CLI agent immediately with:

```
agent exited with code 1
... [help output truncated] ...
-v, --version   Show version number
-h, --help      Show help
Unknown arguments: output-format, outputFormat
```

This is a **PATH resolution problem**, not a Gemini config issue: the GUI app spawns `gemini` from a stripped macOS GUI environment, which resolves to an old binary on `/usr/local/bin` that predates the `--output-format stream-json` flag.

## Repro

1. Install `gemini-cli` via multiple package managers over time (e.g., a very old `npm i -g @google/gemini-cli` left in `/usr/local/bin`, plus a current Homebrew/nvm install).
2. Confirm the modern `gemini` works in a Terminal session: `gemini --output-format stream-json --yolo` — succeeds.
3. Open the **DMG-installed** Open Design app and pick the **Gemini** agent.
4. Send any prompt → agent exits with code 1 and the yargs help/`Unknown arguments` error above.

## Root cause

`apps/daemon/src/agents.ts:311` builds:

```ts
const args = ['--output-format', 'stream-json', '--yolo'];
```

The flags are correct for **modern** Gemini CLI (≥ ~0.30.x), but on macOS the GUI-launched app inherits a minimal PATH (`/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin`). It does **not** include `/opt/homebrew/bin` or any nvm shim. So when the user has an ancient `gemini` left over in `/usr/local/bin` (e.g., 0.1.12 from an old global npm install), `which gemini` from the GUI app resolves to that old binary, which doesn't know `--output-format`.

In my case:

| PATH entry | Version | Supports `--output-format`? |
|---|---|---|
| `/usr/local/bin/gemini` | **0.1.12** | ❌ |
| `/opt/homebrew/bin/gemini` | 0.37.1 | ✅ |
| `~/.nvm/.../bin/gemini` | 0.40.1 | ✅ |

Terminal-launched Gemini works because the user's shell PATH puts nvm/Homebrew first. The GUI app does not see those.

## Environment

- macOS Tahoe 26.x, Apple Silicon
- Open Design **v0.5.0** (DMG install)
- `gemini --version` (terminal) → 0.40.1 (works)
- `/usr/local/bin/gemini --version` → 0.1.12 (the one the app actually picks up)

## Suggested fixes

A few options, roughly in increasing order of effort:

1. **Detect Gemini CLI version on agent spawn.** Run `gemini --version` first; if below the minimum that supports `--output-format stream-json`, surface a clear error in the UI ("Gemini CLI ≥ X.Y required; found 0.1.12 at /usr/local/bin/gemini") instead of letting yargs spit out raw help text. Today the failure mode is opaque.
2. **Resolve `gemini` against an enriched PATH on macOS GUI launches.** When the daemon spawns CLI agents on macOS, prepend the common shell paths (`/opt/homebrew/bin`, `/usr/local/bin`, `$HOME/.nvm/versions/node/*/bin`, `$HOME/.volta/bin`, `$HOME/.asdf/shims`, etc.) — or run resolution through the user's login shell (`zsh -ilc 'which gemini'`) once at startup and cache it. This is what other Electron-based dev tools do (VS Code's "Resolve shell environment").
3. **Allow the user to pin a binary path** in settings (per-agent `binPath` override), similar to how Claude/Copilot agents are configured. Useful as an escape hatch when auto-resolution still picks the wrong binary.
4. **Document a minimum Gemini CLI version** in the agent docs and link to it from the error.

(1) + (2) together would fix the silent breakage for the most common macOS multi-install scenario.

## Workaround

Until this is fixed, removing or replacing the stale `/usr/local/bin/gemini` so the app resolves to the modern binary is enough:

```bash
ls -la $(which -a gemini)
sudo rm /usr/local/bin/gemini   # if it points to an old @google/gemini-cli
# then symlink the modern one, or rely on $PATH
```

## Related

- #736 — also "agent exited with code 1" for Gemini, but a different root cause (Node.js loader error).
- #836 — same family: incompatibility between bundled CLI args (`--yolo`) and the user's installed Gemini CLI version.

Sid-Qin:
## Summary

Closes #978. \`resolveAgentExecutable\` walks PATH then \`userToolchainDirs()\` and returns the first match. On macOS GUI launches, system PATH runs first (\`/usr/bin:/bin:/usr/sbin:/sbin:/usr/local/bin\`), toolchain dirs come second. A stale \`/usr/local/bin/gemini=0.1.12\` left over from an old \`npm i -g @google/gemini-cli\` shadows a modern Homebrew/nvm install, and the daemon spawn lands on yargs \`Unknown arguments: output-format, outputFormat\` because \`--output-format stream-json\` didn't exist in that vintage. Documented \`GEMINI_BIN\` is the existing escape hatch but the user has to know to set it.

## What changed

Opt-in version-aware resolution via a new \`def.minVersion\` field. Currently set only on Gemini (\`'0.30.0'\` — the cited baseline in #978 where \`--output-format stream-json\` shipped stable). Other agents skip the version walk and keep first-match resolution unchanged.

When \`def.minVersion\` is set, \`chooseExecutableByMinVersion\`:

1. Honors the explicit \`<AGENT>_BIN\` configured override (Gemini's \`GEMINI_BIN\`) without running version probes — the user opted out of auto-pick.
2. Enumerates every path on PATH + toolchain dirs where \`def.bin\` resolves.
3. Runs \`--version\` on each in parallel with a 1.5s timeout.
4. Picks the first whose version meets \`minVersion\`.
5. Falls through to first-found when nothing meets the floor — existing detection error path still fires (cryptic, but visible) instead of silently dropping the agent.

\`probe()\` runs the smart resolver and caches the chosen path. The synchronous chat-time \`resolveAgentBin\` (used by the chat handler so spawn() lands on the same executable detection chose, fix for #10) consults the cache first.

## Why this cut

The user's diagnosis was correct: it IS PATH resolution. Just enriching PATH (already partially done in #614) doesn't help when the stale binary is on the system PATH that runs first. Just clearer error messaging (option 1 alone in the issue) leaves the user to manually \`sudo rm /usr/local/bin/gemini\`. The smart resolver actually fixes the multi-install scenario without user action while keeping the explicit override as the escape hatch. \`minVersion\` is opt-in per agent so nothing changes for claude / codex / opencode / etc.

## Test plan

- [x] \`pnpm i18n:check\` — passes.
- [x] \`pnpm typecheck\` (daemon) — my touched files contribute zero new errors (5 pre-existing errors on \`origin/main\` from the \`DesktopExportPdfInput\` sidecar drift, unrelated to this PR).
- [x] \`pnpm test\` (daemon) — 1074 / 1074 of the runnable tests pass; 5 pre-existing failures on origin/main (vp-cli-probe scenarios, resolveCurrentArtifact) unchanged. New tests:
  - \`compareSemver\` (3 cases): basic compare, v-prefix / pre-release / build tolerance, returns 0-on-unparseable so the fall-through fires correctly.
  - \`enumerateOnPath\` (2 cases): PATH order preserved across multiple matches, empty list when bin missing.
  - \`chooseExecutableByMinVersion\` (4 cases): picks newer when older is on PATH first, falls through to first-found when nothing meets the floor (so the existing error surface still fires), honors \`GEMINI_BIN\` override without running version probes, returns null resolved when no candidate exists.
  - \`detectAgents + resolveAgentBin\` end-to-end: pins the cache wiring — without it, chat-time spawn would re-resolve via first-match \`resolveAgentExecutable\` and land on the stale binary detection skipped.
  - Regression guard pinning \`gemini.minVersion\` (so a future cleanup that strips the field can't silently put the GUI app back into the failure mode).

## Manual repro

The issue's setup, simulated locally with two fake \`gemini\` shell scripts:

\`\`\`bash
mkdir -p /tmp/old/bin /tmp/new/bin
cat > /tmp/old/bin/gemini <<'EOSH'
#!/bin/sh
[ \"\$1\" = \"--version\" ] && echo \"0.1.12\"
EOSH
cat > /tmp/new/bin/gemini <<'EOSH'
#!/bin/sh
[ \"\$1\" = \"--version\" ] && echo \"0.40.1\"
EOSH
chmod +x /tmp/old/bin/gemini /tmp/new/bin/gemini

PATH=\"/tmp/old/bin:/tmp/new/bin\" pnpm --filter @open-design/daemon dev
\`\`\`

Before: \`/api/agents\` reports \`gemini\` at \`/tmp/old/bin/gemini\` with version \`0.1.12\`. Sending a chat → \`agent exited with code 1\` + yargs help.

After: \`/api/agents\` reports \`gemini\` at \`/tmp/new/bin/gemini\` with version \`0.40.1\`. Chat lands on the modern binary and the spawn succeeds.

Setting \`GEMINI_BIN=/tmp/old/bin/gemini\` in the daemon's env still wins (escape hatch preserved) — the user has explicitly opted out of auto-pick.

## Out of scope (flagged for follow-ups)

- Surfacing a structured \`disabledReason\` to the API contract / Settings UI when no candidate met \`minVersion\`. Useful for the \"no modern install anywhere\" case (the existing error stays cryptic) but layers over the contract.
- Documenting the minimum Gemini CLI version in agent docs (option 4 in the issue).

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -1160,21 +1160,29 @@ export function enumerateOnPath(bin) {
 // otherwise-valid Gemini CLI versions just because the upstream tags
 // shipped that way.
 //
-// Returns 0 (equal) when either side is unparseable — the caller
-// already has a fall-through path to "use first found", so an
-// unparseable version shouldn't accidentally read as "lower than
-// minVersion" and disqualify a fine binary.
+// Returns `null` when either side is unparseable. The caller MUST
+// distinguish that from `0` (equal) — folding both into "passes the
+// floor" let an old `gemini` that prints a banner like
+// `gemini version 0.1.12` (instead of bare `0.1.12`) sneak past the
+// minVersion check and re-introduce #978. Only callers that have
+// their own fall-through path can safely treat null as "skip this
+// candidate".
 export function compareSemver(a, b) {
   const parse = (raw) => {
     if (typeof raw !== 'string') return null;
     const trimmed = raw.trim().replace(/^v/i, '');
+    // Anchor `^` so a banner like `gemini version 0.1.12` no longer
+    // reads as the bare semver `0.1.12` — that misread was the P1 in
+    // mrcfps + lefarcen's review of #1007: an old Gemini whose
+    // `--version` prints prose with embedded numbers would otherwise
+    // pass the minVersion floor and shadow a modern install.
     const match = trimmed.match(/^(\d+)\.(\d+)\.(\d+)/);
     if (!match) return null;
     return [Number(match[1]), Number(match[2]), Number(match[3])];
   };
   const aParts = parse(a);
   const bParts = parse(b);
-  if (!aParts || !bParts) return 0;
+  if (!aParts || !bParts) return null;
   for (let i = 0; i < 3; i++) {
     if (aParts[i] !== bParts[i]) return aParts[i] > bParts[i] ? 1 : -1;
   }
@@ -1244,7 +1252,15 @@ export async function chooseExecutableByMinVersion(def, configuredEnv, env) {
   );
   for (const result of probes) {
     if (!result.version) continue;
-    if (compareSemver(result.version, def.minVersion) >= 0) {
+    // `compareSemver` returns `null` for unparseable inputs (see
+    // its comment). Treat that as "this candidate's version is not
+    // confidently comparable with the floor" — skip it so a banner
+    // like `gemini version 0.1.12` doesn't accidentally pass the
+    // floor and re-introduce #978. The fall-through below still
+    // runs the spawn against the first candidate so the existing
+    // detection error path stays visible.
+    const cmp = compareSemver(result.version, def.minVersion);
+    if (cmp !== null && cmp >= 0) {
       return { resolved: result.path, version: result.version };
     }
   }
@@ -1320,6 +1336,16 @@ async function probe(def, configuredEnv = {}) {
       // would re-resolve via the first-match `resolveAgentExecutable`
       // and pick the stale binary even though detection skipped it.
       resolvedAgentBinCache.set(def.id, { resolved, version });
+    } else {
+      // Detection couldn't resolve a binary this pass — clear the
+      // cache so a previously-cached path doesn't survive an
+      // uninstall / PATH change / override clear. Without this,
+      // `detectAgents()` would correctly report Gemini unavailable
+      // while `resolveAgentBin()` still returned the stale cached
+      // path, and chat / connection-test would keep spawning the
+      // binary that no longer exists. mrcfps + codex review of
+      // #1007.
+      resolvedAgentBinCache.delete(def.id);
     }
   } else {
     resolved = resolveAgentExecutable(def, configuredEnv);
@@ -1644,10 +1670,19 @@ export function checkWindowsDirectExeCommandLineBudget(def, resolvedBin, args) {
 // first-match `resolveAgentExecutable` and pick the same stale binary
 // detection skipped, putting the user back into the cryptic
 // `Unknown arguments: output-format, outputFormat` failure.
+//
+// IMPORTANT: an explicit `<AGENT>_BIN` configured override (passed
+// per-request from `/api/chat` and `/api/test/connection`) always
+// wins over the cache. Without this guard, a cached auto-pick from
+// an earlier detect would shadow a request-scoped override the user
+// changed afterwards — re-introducing the problem #978's escape
+// hatch is supposed to solve. mrcfps + lefarcen review of #1007.
 export function resolveAgentBin(id, configuredEnv = {}) {
   const def = getAgentDef(id);
   if (!def?.bin) return null;
   if (def.minVersion) {
+    const configured = configuredExecutableOverride(def, configuredEnv);
+    if (configured) return configured;
     const cached = resolvedAgentBinCache.get(id);
     if (cached?.resolved) return cached.resolved;
   }
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -440,6 +440,19 @@ export const AGENT_DEFS = [
     name: 'Gemini CLI',
     bin: 'gemini',
     versionArgs: ['--version'],
+    // Issue #978: Gemini CLI builds before the `--output-format
+    // stream-json` flag stabilized (~0.30) reject our buildArgs with
+    // a yargs `Unknown arguments: output-format, outputFormat` error.
+    // On macOS the GUI app's PATH order means a stale
+    // `/usr/local/bin/gemini` (left over from an old `npm i -g
+    // @google/gemini-cli`) wins over a modern Homebrew/nvm install,
+    // and the spawn fails with cryptic help-output. Setting
+    // `minVersion` opts this entry into the version-aware resolver
+    // (`chooseExecutableByMinVersion`) which walks every `gemini` on
+    // PATH + toolchain dirs, picks the first that meets this floor,
+    // and caches the choice so the chat-time spawn lands on the same
+    // binary detection chose.
+    minVersion: '0.30.0',
     fallbackModels: [
       DEFAULT_MODEL_OPTION,
       // Gemini 3 (May 2026): top-tier reasoning + fast frontier-class.
@@ -1109,6 +1122,144 @@ export function resolveAgentExecutable(def, configuredEnv = {}) {
   return inspectAgentExecutableResolution(def, configuredEnv).selectedPath;
 }
 
+// Enumerate every absolute path on PATH + user toolchain dirs where
+// `bin` resolves. Used by the version-aware resolver to skip stale
+// installs (#978): a user with both `/usr/local/bin/gemini=0.1.12`
+// (legacy `npm i -g`) and `/opt/homebrew/bin/gemini=0.40.1` (current)
+// would otherwise have the GUI app pick the older one because the
+// system PATH segment runs ahead of the toolchain segment, even
+// though the modern binary is also resolvable.
+//
+// Mirrors `resolveOnPath` semantics — same dir order, same Windows
+// PATHEXT walk, same de-dupe — but returns every match rather than
+// the first.
+export function enumerateOnPath(bin) {
+  if (!bin) return [];
+  const exts =
+    process.platform === 'win32'
+      ? (process.env.PATHEXT || '.EXE;.CMD;.BAT').split(';')
+      : [''];
+  const dirs = resolvePathDirs();
+  const out = [];
+  const seen = new Set();
+  for (const dir of dirs) {
+    for (const ext of exts) {
+      const full = path.join(dir, bin + ext);
+      if (!full || seen.has(full)) continue;
+      seen.add(full);
+      if (existsSync(full)) out.push(full);
+    }
+  }
+  return out;
+}
+
+// Compare two version strings by their leading `major.minor.patch`
+// tuple. Returns -1 / 0 / 1 in the usual semver-compare convention.
+// Tolerates pre-release / build / `v`-prefix noise (`'v0.40.1'`,
+// `'0.40.1-rc.2'`, `'0.40.1+meta'`) so the resolver doesn't reject
+// otherwise-valid Gemini CLI versions just because the upstream tags
+// shipped that way.
+//
+// Returns 0 (equal) when either side is unparseable — the caller
+// already has a fall-through path to "use first found", so an
+// unparseable version shouldn't accidentally read as "lower than
+// minVersion" and disqualify a fine binary.
+export function compareSemver(a, b) {
+  const parse = (raw) => {
+    if (typeof raw !== 'string') return null;
+    const trimmed = raw.trim().replace(/^v/i, '');
+    const match = trimmed.match(/^(\d+)\.(\d+)\.(\d+)/);
+    if (!match) return null;
+    return [Number(match[1]), Number(match[2]), Number(match[3])];
+  };
+  const aParts = parse(a);
+  const bParts = parse(b);
+  if (!aParts || !bParts) return 0;
+  for (let i = 0; i < 3; i++) {
+    if (aParts[i] !== bParts[i]) return aParts[i] > bParts[i] ? 1 : -1;
+  }
+  return 0;
+}
+
+// Module-level cache keyed by agent id. Populated by `probe()` when
+// the def opts into version-aware resolution; consulted by
+// `resolveAgentBin()` so the synchronous chat-time spawn picks the
+// same binary the async detection chose.
+const resolvedAgentBinCache = new Map();
+
+// Strip the cache for a given agent id (or all when called bare).
+// Exposed so tests don't have to fight cross-test bleed when they
+// shadow PATH for `chooseExecutableByMinVersion` runs.
+export function clearResolvedAgentBinCache(id) {
+  if (id) resolvedAgentBinCache.delete(id);
+  else resolvedAgentBinCache.clear();
+}
+
+// Walk every candidate binary on PATH + toolchain dirs, run
+// `--version` on each in parallel, and return the first absolute
+// path whose version is at or above `def.minVersion`. Falls through
+// to the first-found path when no candidate meets the minimum so the
+// existing detection error path (cryptic, but visible) still fires
+// instead of silently swallowing the spawn intent. When the explicit
+// `<AGENT>_BIN` override is configured, the override wins
+// unconditionally — the user has already opted out of auto-pick.
+//
+// Returns `{ resolved: string|null, version: string|null }` so the
+// caller can short-circuit the second `--version` exec in `probe()`
+// — the version we just probed here is exactly the one the API
+// response should ship.
+export async function chooseExecutableByMinVersion(def, configuredEnv, env) {
+  const configured = configuredExecutableOverride(def, configuredEnv);
+  if (configured) return { resolved: configured, version: null };
+  const bins = [
+    def.bin,
+    ...(Array.isArray(def.fallbackBins) ? def.fallbackBins : []),
+  ].filter((bin) => typeof bin === 'string' && bin.length > 0);
+  const candidates = [];
+  const seen = new Set();
+  for (const bin of bins) {
+    for (const found of enumerateOnPath(bin)) {
+      if (seen.has(found)) continue;
+      seen.add(found);
+      candidates.push(found);
+    }
+  }
+  if (candidates.length === 0) return { resolved: null, version: null };
+  // Probe versions in parallel with a short timeout so a hung old
+  // binary doesn't block detection. 1500ms is generous for a
+  // local CLI's `--version` path which is typically <50ms.
+  const probes = await Promise.all(
+    candidates.map(async (candidate) => {
+      try {
+        const { stdout } = await execAgentFile(candidate, def.versionArgs, {
+          env,
+          timeout: 1500,
+        });
+        const version = String(stdout || '').trim().split('\n')[0] || '';
+        return { path: candidate, version };
+      } catch {
+        return { path: candidate, version: null };
+      }
+    }),
+  );
+  for (const result of probes) {
+    if (!result.version) continue;
+    if (compareSemver(result.version, def.minVersion) >= 0) {
+      return { resolved: result.path, version: result.version };
+    }
+  }
+  // No candidate met the minimum — fall through to the first-found
+  // path so the spawn still happens (and the existing yargs/argv
+  // mismatch error surfaces). Returning null here would silently
+  // drop the agent from detection even though a binary IS present;
+  // worse UX than the cryptic-but-visible failure.
+  const fallback = probes[0];
+  return {
+    resolved: fallback?.path ?? candidates[0],
+    version: fallback?.version ?? null,
+  };
+}
+
 async function fetchModels(def, resolvedBin, env) {
   if (typeof def.fetchModels === 'function') {
     try {
@@ -1141,15 +1292,6 @@ async function fetchModels(def, resolvedBin, env) {
 }
 
 async function probe(def, configuredEnv = {}) {
-  const resolved = resolveAgentExecutable(def, configuredEnv);
-  if (!resolved) {
-    return {
-      ...stripFns(def),
-      models: def.fallbackModels ?? [DEFAULT_MODEL_OPTION],
-      available: false,
-      ...installMetaForAgent(def.id),
-    };
-  }
   const probeEnv = spawnEnvForAgent(
     def.id,
     {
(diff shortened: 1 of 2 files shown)


apps/daemon/src/agents.ts
```diff
@@ -982,6 +995,144 @@ export function resolveAgentExecutable(def, configuredEnv = {}) {
...
+        const version = String(stdout || '').trim().split('\n')[0] || '';
+        return { path: candidate, version };
+      } catch {
+        return { path: candidate, version: null };
+      }
+    }),
+  );
+  for (const result of probes) {
+    if (!result.version) continue;
+    if (compareSemver(result.version, def.minVersion) >= 0) {
```
mrcfps: `compareSemver()` intentionally returns `0` when either version string is unparseable, but this `>= 0` check treats that case as satisfying the floor. In practice that means a candidate whose `--version` output is something like `gemini 0.1.12` (or any other non-bare-semver string) will still be selected and cached, which puts us back on the stale binary even though `minVersion` was supposed to filter it out. Please distinguish parse failure from equality here — e.g. reject non-matching outputs before the comparison, or have `compareSemver()` surface whether parsing succeeded so only genuinely comparable versions can pass this branch.

apps/daemon/src/agents.ts
```diff
@@ -1030,15 +1173,47 @@ async function probe(def, configuredEnv = {}) {
...
+    // shadowing a modern Homebrew/nvm install on macOS GUI launches).
+    const picked = await chooseExecutableByMinVersion(def, configuredEnv, probeEnv);
+    resolved = picked.resolved;
+    version = picked.version;
+    if (resolved) {
+      // Cache so the synchronous chat-time `resolveAgentBin` returns
+      // the same path the async detection chose. Without this, chat
+      // would re-resolve via the first-match `resolveAgentExecutable`
+      // and pick the stale binary even though detection skipped it.
+      resolvedAgentBinCache.set(def.id, { resolved, version });
```
mrcfps, line 1342: This cache only gets updated on a successful pick, so when a later detect pass returns `resolved = null` (binary removed, PATH changed, override cleared, etc.) the old entry survives. `detectAgents()` will report Gemini unavailable, but `resolveAgentBin()` still consults the stale cached path first and can keep spawning the removed executable. Please clear `resolvedAgentBinCache` before returning `available: false` for `minVersion` agents, and add a regression test that exercises detect → uninstall/path change → detect again.

apps/daemon/src/agents.ts
```diff
@@ -1333,9 +1508,20 @@ export function checkWindowsDirectExeCommandLineBudget(def, resolvedBin, args) {
...
+// first-match `resolveAgentExecutable` and pick the same stale binary
+// detection skipped, putting the user back into the cryptic
+// `Unknown arguments: output-format, outputFormat` failure.
 export function resolveAgentBin(id, configuredEnv = {}) {
   const def = getAgentDef(id);
   if (!def?.bin) return null;
+  if (def.minVersion) {
+    const cached = resolvedAgentBinCache.get(id);
+    if (cached?.resolved) return cached.resolved;
+  }
```
mrcfps, line 1701: Because `resolveAgentBin()` returns the cached path before it looks at `configuredEnv`, any request-scoped `GEMINI_BIN` override passed from `/api/chat` or `/api/test/connection` is ignored once an earlier detect pass has filled this cache. That breaks the explicit override/escape-hatch behavior described in the PR and can keep chat or connection tests pinned to the stale binary even after the user selects a different executable. Please let `configuredExecutableOverride(def, configuredEnv)` win before consulting the cache, or bypass the cache whenever the agent-bin env var is present, and cover that with an end-to-end override test.

mrcfps (commented):
@Sid-Qin Thanks for pushing on the Gemini PATH/version selection fix — I walked the changed resolver/cache paths and found a few follow-ups worth tightening before we rely on this as the escape hatch for multi-install setups.

apps/daemon/src/agents.ts
lefarcen: **P1 — Unparseable version treated as satisfying minVersion**

`compareSemver()` returns `0` for unparseable versions (line 1047: `if (!aParts || !bParts) return 0`), but `chooseExecutableByMinVersion()` treats `>= 0` as satisfying the floor (line 1120: `if (compareSemver(result.version, def.minVersion) >= 0)`). That means a first PATH candidate printing something like `gemini version 0.1.12` (banner text, not clean semver) will pass through and shadow a later valid modern install, reintroducing #978.

**Fix:** Make `compareSemver()` return `null` for unparseable input (instead of `0`), then only accept candidates when the comparison is non-null AND `>= 0`. Add a resolver test with an unparseable first candidate and a valid newer second candidate to guard this.

apps/daemon/src/agents.ts
```diff
@@ -1333,9 +1508,20 @@ export function checkWindowsDirectExeCommandLineBudget(def, resolvedBin, args) {
...
 // detection reported as available — fixes Windows ENOENT when the bare
 // bin name isn't on the child process's PATH (issue #10).
+//
+// For agents that opted into version-aware resolution via `def.minVersion`
+// (currently just gemini, see #978), prefer the path the async detection
+// already chose. Without this, chat would re-resolve via the
+// first-match `resolveAgentExecutable` and pick the same stale binary
+// detection skipped, putting the user back into the cryptic
+// `Unknown arguments: output-format, outputFormat` failure.
 export function resolveAgentBin(id, configuredEnv = {}) {
```
lefarcen, line 1693: **P2 — Cache ignores GEMINI_BIN override**

`resolveAgentBin()` checks `resolvedAgentBinCache` before considering `configuredEnv` (line 1518-1520), and the cache is keyed only by agent id. If detection cached an auto-selected Gemini path, then the user later sets or changes `GEMINI_BIN`, chat-time spawn can ignore the explicit override and keep using the stale cached binary.

**Fix:** Either check the explicit configured override before cache lookup (mirror the `configuredExecutableOverride()` call in `chooseExecutableByMinVersion`), or include the configured binary/env fingerprint in the cache key and clear when detection finds no resolved path.

apps/daemon/tests/agents.test.ts
```diff
@@ -2107,3 +2112,274 @@ test('spawnEnvForAgent does not mutate the input env', () => {
...
+  'enumerateOnPath returns every directory containing the binary, in PATH order',
+  () => {
+    const dirA = mkdtempSync(join(tmpdir(), 'od-enumerate-a-'));
+    const dirB = mkdtempSync(join(tmpdir(), 'od-enumerate-b-'));
+    try {
+      writeFileSync(join(dirA, 'gemini'), '');
+      writeFileSync(join(dirB, 'gemini'), '');
+      chmodSync(join(dirA, 'gemini'), 0o755);
+      chmodSync(join(dirB, 'gemini'), 0o755);
+      process.env.PATH = `${dirA}:${dirB}`;
```
lefarcen, line 2283: **P2 — Test PATH isolation incomplete**

The new PATH/enumeration tests set `process.env.PATH` to temp dirs but mostly leave `OD_AGENT_HOME` unset, so `enumerateOnPath()` still appends real well-known toolchain dirs from the developer/CI machine via `userToolchainDirs()`. On a machine with a real `gemini` under Homebrew/nvm/asdf/npm-global, the "missing binary" and exact-list assertions can fail or mask extra candidates.

**Fix:** Set `OD_AGENT_HOME` to an empty temp home for every #978 resolver test (not just the end-to-end cache test at line 2223), so toolchain dirs resolve to empty and PATH is the only source.

lefarcen (changes requested):
Hey @Sid-Qin! 🎉 Thanks for tackling #978 with this comprehensive version-aware resolver. The approach is sound — walking all PATH candidates and picking by version floor is the right fix for the stale-binary shadowing scenario. The test coverage is excellent (274 lines covering compare/enumerate/choose/cache/regression guards).

Found 3 issues that need addressing before merge:

Sid-Qin:
Pushed fixes for the four review items + the CI red.

**P1 (compareSemver banner-style pass-through, both reviewers + codex):** `compareSemver` now returns `null` for unparseable inputs (including prose-wrapped versions like `gemini version 0.1.12` — anchored the regex to `^` so prose with embedded numbers no longer reads as a clean semver). `chooseExecutableByMinVersion` gates on `cmp !== null && cmp >= 0`, so banner-style candidates fall through to the next on PATH. New regression test pins both `compareSemver` returning null on the banner shape and the resolver skipping the bannered candidate to accept the modern one.

**P2 (cache-on-failure, mrcfps + codex):** Probe now `delete`s the cache entry when it can't resolve a binary this pass. New end-to-end test exercises detect → uninstall (PATH cleared) → re-detect → `resolveAgentBin` returns null instead of the stale cached path.

**P2 (override-beats-cache, mrcfps + lefarcen + codex):** `resolveAgentBin` now calls `configuredExecutableOverride` before the cache lookup, so request-scoped `GEMINI_BIN` always wins. New test covers detect (caches auto-pick) → `resolveAgentBin` with `GEMINI_BIN` override (returns the override path) → `resolveAgentBin` without override (cache still serves the auto-pick, so the override stays request-scoped).

**P2 (test PATH isolation, lefarcen):** New `isolateAgentHome()` helper pins `OD_AGENT_HOME` to an empty temp directory for every #978 / #1007 test so `userToolchainDirs()` resolves to nothing and PATH is the only candidate source. Wired into all the new `enumerateOnPath` / `chooseExecutableByMinVersion` / e2e tests, not just the cache one.

**CI red:** `gemini.minVersion` was tripping `TS18048` (`possibly 'undefined'`) under `tsconfig.tests.json` strict mode. Narrowed via a local `const minVersion = gemini.minVersion` + `typeof === 'string' && length > 0` guard before the comparison.

Daemon `pnpm typecheck` clean (only the pre-existing `DesktopExportPdfInput` / `blake3-wasm` errors from main remain, neither in agent code). All 21 of my #978 / #1007 tests pass; full `agents.test.ts` is 107 / 109 with the same 2 pre-existing vp-cli-probe failures unchanged from main.

apps/daemon/src/agents.ts
```diff
@@ -1030,15 +1189,57 @@ async function probe(def, configuredEnv = {}) {
...
+    // shadowing a modern Homebrew/nvm install on macOS GUI launches).
+    const picked = await chooseExecutableByMinVersion(def, configuredEnv, probeEnv);
+    resolved = picked.resolved;
+    version = picked.version;
+    if (resolved) {
+      // Cache so the synchronous chat-time `resolveAgentBin` returns
+      // the same path the async detection chose. Without this, chat
+      // would re-resolve via the first-match `resolveAgentExecutable`
+      // and pick the stale binary even though detection skipped it.
+      resolvedAgentBinCache.set(def.id, { resolved, version });
```
mrcfps, line 1342: When `chooseExecutableByMinVersion()` resolves Gemini through `GEMINI_BIN`, this branch still caches that pinned path as the canonical resolver result. Because `/api/app-config` only persists the new settings and `/api/chat` / `/api/test/connection` later call `resolveAgentBin()` without forcing another `detectAgents()` pass, clearing `GEMINI_BIN` can leave chat pinned to the old override until `/api/agents` happens to run again. That makes the escape hatch non-reversible in the main flow even though the current config no longer requests it. Please either skip caching configured overrides here, or record whether the cache entry came from an explicit override and ignore that cached entry once the current request has no `GEMINI_BIN`. A regression test that detects with `GEMINI_BIN` set, clears it, and then calls `resolveAgentBin("gemini", {})` before another detect would lock this down.

mrcfps (commented):
@Sid-Qin Thanks for tightening the resolver after the first review pass — I re-walked the latest cache/override flow and found one more follow-up worth addressing.

lefarcen (dismissed):
All three findings addressed! 🎉

**P1 (compareSemver unparseable pass-through):** Fixed — now returns `null` for unparseable inputs, regex anchored to reject prose-wrapped versions. `chooseExecutableByMinVersion` correctly gates on `cmp !== null && cmp >= 0`.

**P2 (cache-on-failure):** Fixed — `probe()` now deletes the cache entry when `resolved = null`, preventing stale paths from surviving uninstall/PATH changes.

**P2 (override-beats-cache):** Fixed — `resolveAgentBin` checks `configuredExecutableOverride` before cache lookup, so request-scoped `GEMINI_BIN` always wins.

**P2 (test isolation):** Fixed — `isolateAgentHome()` helper wired into all #978/#1007 tests.

Test coverage is comprehensive, CI green. This is production-ready.

lefarcen:
@mrcfps This PR's current head 33182de7 is waiting on your review.

Sid-Qin:
Rebased onto current `main` (`32fa0c23`) to resolve the merge conflict with #985 (`installMetaForAgent` add to unavailable-agent responses).

**Conflict shape:** PR #985 added an early `resolveAgentExecutable` short-circuit at the top of `probe()` that returns `{ available: false, ...installMetaForAgent(def.id) }` when no binary is found. This PR replaces that short-circuit with the `minVersion`-aware path (`chooseExecutableByMinVersion` walks every PATH candidate and picks the first one meeting `def.minVersion`), so the short-circuit had to go — but the `installMetaForAgent(def.id)` field from #985 must still appear in the new "no resolved binary" return.

**Resolution:**
- Dropped the conflicting HEAD-side early return in `probe()` (its job is now covered by the `minVersion`-aware path or the `else { resolved = resolveAgentExecutable(...) }` branch).
- Added `...installMetaForAgent(def.id)` to the consolidated `if (!resolved) return …` block downstream, so the unavailable-agent response still carries `installUrl` / `docsUrl` for the Settings onboarding links shipped in #985.
- Auto-merge in `apps/daemon/tests/agents.test.ts` was clean — the new imports (`compareSemver`, `chooseExecutableByMinVersion`, `clearResolvedAgentBinCache`, `enumerateOnPath`, `resolveAgentBin`) and the #978 test block landed without conflicts and were not modified.

**No logic change vs the previously approved head** (`33182de7`) beyond the merge resolution itself.

**Local verification on the rebased head (`634b0833`):**
- `pnpm --filter @open-design/daemon typecheck` — clean (exit 0)
- `cd apps/daemon && pnpm exec vitest run tests/agents.test.ts` — **112/112 passed**, including the three #978/#1007 regression pins (`chooseExecutableByMinVersion picks the newer candidate even when an older one is on PATH first`, `chooseExecutableByMinVersion skips a banner-style version output and accepts the next candidate`, `detectAgents + resolveAgentBin returns the modern gemini even when an older one shadows it on PATH`)
- The four daemon test failures that remain in the broader suite (`tests/finalize-design.test.ts` ×3, `tests/project-file-rename.test.ts` ×1) reproduce verbatim on bare `origin/main` — they are pre-existing and unrelated to this PR.

Ready for re-review when you have a moment, @lefarcen — happy to address anything else that surfaces.

lefarcen (commented):
Hey @Sid-Qin, thanks for the iteration on the cache + override interaction. I re-walked the current head and found one more follow-up worth addressing before this ships.

**Override-cache pollution (P2):**

`chooseExecutableByMinVersion()` still caches the `GEMINI_BIN` override path (line 1321). After the user clears `GEMINI_BIN`, `resolveAgentBin('gemini', {})` finds no `configured` override at the top of the function, falls back to the cache lookup, and returns the stale pinned path—even though detection just ran and the user's intent was to go back to auto-pick.

The fix in `resolveAgentBin()` (checking `configuredExecutableOverride` before the cache) addresses the opposite direction: request-scoped override now correctly wins over an existing auto-pick cache. It doesn't fix the reported stale-pin concern because the cache itself is being populated with the override path at detect-time.

**Suggested fix:**

Skip caching when `chooseExecutableByMinVersion()` returns an override path. Only cache the result when the resolver actually walked candidates and picked one via version comparison. That way, clearing the override lets the next request fall through to the sync first-match resolver (which will also respect the cleared override), keeping detection and chat in sync.

**Test gap:**

The new test at line 2574 covers override-before-cache when the cache contains an auto-picked binary. Add a complementary test that detects with `GEMINI_BIN` set, clears it, and verifies `resolveAgentBin('gemini', {})` returns a fresh auto-pick (not the stale pinned path).

---

This is the last piece needed to make the escape hatch reversible in the main flow. Once this lands, the PR will address all the review findings. Looking forward to seeing this ship!

Sid-Qin:
@lefarcen Rebased onto current `main` (`1dc02245`) and pushed `5f70e919` addressing the override-cache pollution P2. Three commits on the branch now: original (`275f6bf5` after rebase), the previous review tighten (`19cf54ce`), and this commit on top.

**P2 — Override-cache pollution** ([thread](https://github.com/nexu-io/open-design/pull/1007#issuecomment-4417263918))

You were right that the previous override-beats-cache fix only handled one direction. The reverse — clearing the override after detection cached the pin — still left a stale pin in the cache.

Approach: thread a `fromOverride` flag through `chooseExecutableByMinVersion`'s return so `probe()` can keep its existing version-aware vs unavailable cache shape and just skip the write on the override branch. That keeps the previous mrcfps/codex evict-on-failure path intact and doesn't touch `resolveAgentBin`'s precedence (which is already correct).

- `chooseExecutableByMinVersion` now returns `{ resolved, version, fromOverride }`. Three return sites:
  - `configured` early-return → `fromOverride: true`
  - version-pick (`compareSemver` ≥ floor) → `fromOverride: false`
  - first-found fallback (no candidate met floor) → `fromOverride: false`
- `probe()` cache write is now gated on `resolved && !picked.fromOverride`. The `else if (!resolved)` evict-on-failure branch is unchanged. There is a new trailing comment that spells out *why* a `fromOverride: true` result intentionally leaves the cache as-is (so a future contributor doesn't \"helpfully\" re-add the write).

**Test gap** (lefarcen)

Added `detectAgents with GEMINI_BIN does not poison the cache; clearing the override returns to the auto-pick (#1007 P2 lefarcen)` covering the canonical reversible-escape-hatch scenario:

1. detect without override → cache = `autoPath` (a 0.40.1 binary on PATH)
2. detect with `GEMINI_BIN: pinPath` (a different 0.40.2 binary) → API result returns the pin, cache **untouched**
3. `resolveAgentBin('gemini', {})` → returns `autoPath`, not the pin

Verified RED before the fix (probe wrote `picked.resolved` unconditionally so step 3 returned the pin) and GREEN after. The complementary test `resolveAgentBin honors a request-scoped GEMINI_BIN override even when the cache is populated` still passes — together they pin both directions of the override interaction.

**Rebase note** — `main` between our last head (`32fa0c23`) and now (`1dc02245`) added `#1205 Improve Codex CLI path fallback UX`, which refactored `resolveAgentExecutable` into a thin wrapper around the new `inspectAgentExecutableResolution`. That refactor is orthogonal to the `minVersion` path so `agents.ts` auto-merged cleanly. The only conflict was an import-list collision in `tests/agents.test.ts` (their `inspectAgentExecutableResolution` next to our `enumerateOnPath` + `resolveAgentBin`); merged alphabetically with no semantic change.

**Local verification on `5f70e919`:**
- `pnpm --filter @open-design/daemon typecheck` — clean (exit 0)
- `cd apps/daemon && pnpm exec vitest run tests/agents.test.ts` — **114/114 passed**, including both #1007 P2 tests (the previous round's override-beats-cache and the new poison-the-cache one)

Ready for re-review.

lefarcen (approved):
P2 fixed — the `fromOverride` flag correctly gates the cache write so clearing `GEMINI_BIN` now falls back to the version-aware auto-pick instead of staying pinned to the override. Test coverage confirms the reversible-escape-hatch scenario works. LGTM. 🎉

lefarcen:
@mrcfps This PR's current head 5f70e919 is waiting on your review.

lefarcen:
Hey @Sid-Qin — quick note: the PR is still in **draft** state. All the review findings (P1 compareSemver pass-through, P2 cache-on-failure, P2 override-beats-cache) have been addressed, and the current head (`5f70e919`) is approved from my side.

If you consider the changes complete, marking the PR ready for review will move it into the normal merge queue. Let us know if you had more changes in mind first!

lefarcen:
Hey @Sid-Qin — following up: `main` has moved on since your last push and the PR now has a **merge conflict**. You'll need to rebase onto current `main` before it can be reviewed or merged.

Once you've rebased (and un-drafted if you're happy with the changes), just drop a note here and we'll pick up the review! Let us know if you run into anything.

lefarcen:
Heads-up: PR #2797 by @YOMXXX is also open against the same area — same title, closes #978, and overlaps `apps/daemon/src/agents.ts`. It was opened on 2026-05-23 after this PR went quiet, isn't in draft, and @mrcfps approved it yesterday on the current head.

The maintainer team will decide which version lands. Sharing so nothing is wasted — the approach you prototyped here clearly carried through into the solution in #2797. Thank you for the groundwork. 🙏

PerishCode:
@Sid-Qin — quick check-in: this PR has been quiet since the 5/30 note about the overlapping #2797. Current GitHub state is still draft + conflicted against `main`, and #2797 is now closed, so this is the remaining open PR for #978.

Are you still planning to carry this one forward? If yes, the next visible state change would be a rebase onto current `main` and marking it ready once you are happy with it. If priorities shifted, just say so and we can close it cleanly.

PR #1016, opened 2026-05-09
feat(contracts): share critique round / status types via the contracts package

Nagendhra-web:
First step of the Phase 6.2 unblock plan we discussed on [#1006](https://github.com/nexu-io/open-design/pull/1006). This PR moves the three pieces of public critique state out of daemon-private source into the shared contracts package so the web layer can consume them without duplicating shapes or reaching across the apps/web → apps/daemon boundary.

### What moves

[\`packages/contracts/src/critique.ts\`](https://github.com/nexu-io/open-design/blob/feat/critique-contracts-shared-types/packages/contracts/src/critique.ts) gains:

- \`CritiqueRoundSummary\` (record persisted with each run; mirrors what the orchestrator writes in \`decideRound\`).
- \`CritiqueRunStatus\` (terminal-only union: \`ShipStatus | 'degraded' | 'failed' | 'legacy'\`).
- \`CRITIQUE_RUN_STATUSES\` (canonical display order: ship outcomes → quality / lifecycle exits → \`legacy\`).

### What stays daemon-side

[\`apps/daemon/src/critique/persistence.ts\`](https://github.com/nexu-io/open-design/blob/feat/critique-contracts-shared-types/apps/daemon/src/critique/persistence.ts) re-exports all three so every existing daemon import (orchestrator, server, run-registry, tests) keeps working unchanged.

The CHECK-constraint-aware \`ALL_VALID_STATUSES\` set that includes the in-flight \`'running'\` value stays daemon-private because it's strictly a DB invariant, not a public contract.

### Why now

This is the foundation step that codex-bot P2 and lefarcen P2 on [#1006](https://github.com/nexu-io/open-design/pull/1006) both asked for: the rerun and history endpoints will need a shared shape, and Phase 7 (web reducer) starts consuming them as soon as the artifact-extraction work lands. Splitting it out as its own no-behavior-change PR keeps it small and reviewable.

### Validation

- \`pnpm guard\` and \`pnpm --filter @open-design/contracts typecheck\` clean.
- \`pnpm --filter @open-design/daemon build\` clean (the same step CI runs for tsc).
- \`pnpm --filter @open-design/web typecheck\` clean (sanity check that web doesn't break on the contracts version bump).
- \`pnpm --filter @open-design/daemon exec vitest run tests/critique\` (12 files, 150 passed; existing \`critique-persistence.test.ts\` continues to import \`CRITIQUE_RUN_STATUSES\` through the daemon re-export and asserts the array still covers every status).

### DIFF
--- a/apps/daemon/src/critique/persistence.ts
+++ b/apps/daemon/src/critique/persistence.ts
@@ -1,27 +1,20 @@
 import type Database from 'better-sqlite3';
-import type { ShipStatus } from '@open-design/contracts/critique';
+import {
+  CRITIQUE_RUN_STATUSES,
+  type CritiqueRoundSummary,
+  type CritiqueRunStatus,
+} from '@open-design/contracts/critique';
 
 /**
- * Final critique status persisted with each run. Mirrors the spec's CHECK
- * constraint on critique_status. 'failed' covers orchestrator-level errors,
- * 'legacy' marks rows produced before the feature shipped (reserved for the
- * artifacts-on-disk backfill in Phase 15).
+ * Re-export the public contract types and enumeration so existing
+ * daemon-side imports (`./persistence.js`) keep working unchanged. The
+ * canonical definitions live in `@open-design/contracts/critique` so the
+ * web layer can consume the same shapes and the same display order
+ * through the rerun / history endpoints (AGENTS.md requirement that
+ * shared API DTOs live in packages/contracts).
  */
-export type CritiqueRunStatus =
-  | ShipStatus
-  | 'degraded'
-  | 'failed'
-  | 'legacy';
-
-export const CRITIQUE_RUN_STATUSES: readonly CritiqueRunStatus[] = [
-  'shipped',
-  'below_threshold',
-  'timed_out',
-  'interrupted',
-  'degraded',
-  'failed',
-  'legacy',
-];
+export { CRITIQUE_RUN_STATUSES };
+export type { CritiqueRoundSummary, CritiqueRunStatus };
 
 // All values accepted by the DB CHECK constraint, including the in-flight value
 // that the public type union deliberately omits.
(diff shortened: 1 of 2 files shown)


mrcfps (approved):
@Nagendhra-web I verified the critique round summary and terminal status types now live in `@open-design/contracts/critique`, while `apps/daemon/src/critique/persistence.ts` keeps the daemon-side re-exports and DB-only `running` invariant intact. The moved contract surface and the daemon compatibility shim both look consistent with the repo boundary rules here—thanks for splitting this foundation step into a small, easy-to-review change.


lefarcen:
Hi @Nagendhra-web! 🎉

Thanks for the contribution — this is a clean architectural move that extracts critique types to the contracts layer where they belong.

— open-design team

lefarcen (commented):
Thanks @Nagendhra-web — this is a clean extraction that moves critique types to their correct home in contracts. `RoundDecision` and `ShipStatus` are already exported there, so the `CritiqueRoundSummary.decision` type change is backward-compatible. The re-export pattern in daemon preserves the existing API surface, and your validation coverage (150 passing tests) gives confidence.

## Minor design considerations (non-blocking)

**P2 — Internal 'running' vs terminal-only contract** (`apps/daemon/src/critique/persistence.ts`)

`CritiqueRunRow.status` is typed as terminal-only `CritiqueRunStatus`, but the DB persists `'running'` too (visible in `ALL_VALID_STATUSES`), and `normalizeRow` can return it. Now that `CritiqueRunStatus` lives in contracts where history/rerun DTOs may reuse it, the mismatch becomes more important: those API responses could accidentally expose in-flight states.

Consider introducing an internal daemon type like `CritiquePersistedStatus = CritiqueRunStatus | 'running'` for DB rows, while keeping API DTOs explicitly terminal-only. That way the type system catches if someone tries to return a `'running'` row through a public endpoint.

**P3 — Exhaustiveness of CRITIQUE_RUN_STATUSES** (`packages/contracts/src/critique.ts`)

`CRITIQUE_RUN_STATUSES` uses `satisfies` to prove every listed value is valid, but doesn't prove the list is exhaustive. If `ShipStatus` grows later (e.g. a new ship outcome like `'partial'`), this array and the display/filter logic can silently miss it.

Small safeguard: add an exhaustiveness type test (mapped type over `CritiqueRunStatus` → ensure every variant appears in the array), or derive the union from the canonical constant so upstream changes fail loudly.

lefarcen (approved):
Approved.

PR #1031, opened 2026-05-09
fix(web): make privacy consent choices explicit

lefarcen:
## Summary

- Replace the privacy consent segmented control with two equal-weight plain buttons, so neither choice appears pre-selected.
- Apply the same explicit button treatment to the Settings -> Privacy consent card.
- Add a regression asserting the first-run banner no longer renders a selected segmented control.

## Test plan

- [x] `pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/App.connectors.test.tsx`

Note: local validation emitted Node engine warnings because this machine is on Node v22.22.0 while the repo targets Node ~24.

### DIFF
--- a/apps/web/src/components/PrivacyConsentModal.tsx
+++ b/apps/web/src/components/PrivacyConsentModal.tsx
@@ -1,4 +1,3 @@
-import type { CSSProperties } from 'react';
 import { useT } from '../i18n';
 
 interface Props {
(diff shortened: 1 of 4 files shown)


mrcfps (approved):
@lefarcen I reviewed the updated privacy consent flow and verified the banner and Settings card now present two equal-weight, unselected actions without changing the underlying share/decline behavior. I also checked the added regression coverage around the first-run banner state. Thanks for tightening up the consent UX here 🙌


lefarcen (commented):
Hey @lefarcen! Clean privacy consent UX fix. Two minor housekeeping suggestions:

**Stale reasoning comment (P3):**
The comment in `PrivacyConsentModal.tsx:21` still says telemetry decisions "key off whether installationId is set", but the current model uses `privacyDecisionAt` to record consent state. In privacy-critical code this stale reasoning can mislead future changes. Consider updating it to say `privacyDecisionAt` gates whether consent has been decided, and `installationId` is only generated when needed.

**Missing Settings regression (P3):**
The Settings privacy consent card received the same segmented-control-to-buttons treatment, but the new regression only asserts the first-run banner. A future change could reintroduce `.seg-control` / `.seg-btn.active` in Settings without failing tests. Adding a small assertion to `PrivacySection.test.tsx` for the undecided state (no `.seg-control`, no `.seg-btn.active`, both buttons use `.privacy-consent-action`) would close the gap.

No blocking correctness, security, or performance findings. Nice explicit-choice design direction — this mirrors the good pattern from GDPR banners where neither option is visually pre-selected.

PR #1032, opened 2026-05-09
fix(desktop): keep modal controls clickable in drag regions

mrcfps:
## Summary
- Mark settings and design-system preview modals as `no-drag` so controls remain clickable when overlapping the Electron chrome drag strip.
- Broaden the macOS injected chrome CSS from individual DS modal controls to whole modal/backdrop surfaces.

## Validation
- `pnpm --filter @open-design/desktop typecheck`

### DIFF
--- a/apps/desktop/src/main/runtime.ts
+++ b/apps/desktop/src/main/runtime.ts
@@ -101,6 +101,16 @@ const MAC_WINDOW_CHROME_CSS = `
   .app-chrome-drag {
     -webkit-app-region: drag;
   }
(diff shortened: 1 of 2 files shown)


nettee (approved):

lefarcen (commented):
The parent-level no-drag approach looks correct and preserves #864's protected DS modal controls through broader `.ds-modal *` coverage. Checked for unintended side effects — found no current modal descendant that should remain draggable.

Nice cleanup of the child-selector approach 👍

PR #1033, opened 2026-05-09
feat: add scheduled routines for unattended agent runs

pftom:
## Summary

Generalizes Orbit's single hard-coded daily-digest scheduler into user-defined **routines**: each one fires on a schedule (Hourly / Daily / Weekdays / Weekly with IANA timezone) and starts a fresh agent conversation, either inside an existing project or in a new project minted on the spot.

- **Backend (apps/daemon):** new `RoutineService` with timezone-aware `nextRunAt` computed via `Intl.DateTimeFormat` (no new dependency); two-pass `tzWallToUtc` so DST transitions stay correct. Each fire chains `rescheduleOne` in `finally()` so the cadence keeps going. New `routines` + `routine_runs` SQLite tables; `schedule_json` is the authoritative form, `schedule_kind`/`value` kept populated for query convenience. `/api/routines` CRUD + `/api/routines/:id/run` + `/api/routines/:id/runs`.
- **UI (Settings → Routines):** pill-chip schedule kind picker, time + timezone fields, weekday picker for Weekly, live preview line ("Runs daily at 9:00 AM GMT+8"). Routine list with status pill + next/last meta + expandable run history; each history row links into the project the run wrote to via the existing router primitive.
- **Custom cron** intentionally deferred — the discriminated union leaves room to add it later without breaking changes.

## Test plan

- [x] `pnpm guard`
- [x] `pnpm --filter @open-design/web test` — 596/596 pass
- [x] Rebased cleanly onto latest `origin/main`
- [ ] Manual smoke: open Settings → Routines, create a `Hourly` and a `Daily` routine, verify the live preview line, "Run now" spawns a conversation in the right project, "History" lists the run with an "Open project" link that navigates correctly
- [ ] Manual smoke: create a `Weekly` routine for tomorrow's weekday at HH:MM in a non-UTC timezone, restart the daemon, confirm it still fires on schedule

### DIFF
--- a/apps/daemon/src/routines.ts
+++ b/apps/daemon/src/routines.ts
@@ -452,9 +452,16 @@ export class RoutineService {
       return handlerStart;
     })();
     this.inflight.set(routineId, promise);
-    promise.finally(() => {
-      this.inflight.delete(routineId);
-    });
+    // The trailing `finally(...)` returns a new promise that mirrors the
+    // original rejection; without `.catch` it would surface as an
+    // unhandled rejection (fatal in modern Node) when the handler rejects
+    // before producing a start handle. The original `promise` is still
+    // returned to callers, who handle the rejection there.
+    promise
+      .finally(() => {
+        this.inflight.delete(routineId);
+      })
+      .catch(() => {});
     return promise;
   }
 }
--- a/apps/daemon/src/routines.ts
+++ b/apps/daemon/src/routines.ts
@@ -1,0 +1,460 @@
+// RoutineService — multi-routine scheduler. Generalizes the single-routine
+// pattern in OrbitService: a list of user-defined routines, each with its
+// own schedule, that fires the registered run handler. Schedule kinds
+// covered: hourly (every hour at minute M), daily (HH:MM in timezone),
+// weekdays (Mon-Fri at HH:MM in timezone), weekly (one weekday at HH:MM in
+// timezone). The run handler (wired by server.ts) is responsible for
+// project/conversation creation and dispatch into startChatRun.
+
+import { randomUUID } from 'node:crypto';
+
+// Local mirror of the @open-design/contracts routine types. Kept here so
+// this service typechecks under NodeNext (the contracts dist re-exports are
+// extension-less, which only works under bundler-mode resolution). The
+// shapes must stay aligned with packages/contracts/src/api/routines.ts.
+
+export type RoutineRunStatus =
+  | 'queued'
+  | 'running'
+  | 'succeeded'
+  | 'failed'
+  | 'canceled';
+
+export type RoutineRunTrigger = 'manual' | 'scheduled';
+
+export type Weekday = 0 | 1 | 2 | 3 | 4 | 5 | 6;
+
+export type RoutineSchedule =
+  | { kind: 'hourly'; minute: number }
+  | { kind: 'daily'; time: string; timezone: string }
+  | { kind: 'weekdays'; time: string; timezone: string }
+  | { kind: 'weekly'; time: string; timezone: string; weekday: Weekday };
+
+export type RoutineProjectTarget =
+  | { mode: 'create_each_run' }
+  | { mode: 'reuse'; projectId: string };
+
+export interface Routine {
+  id: string;
+  name: string;
+  prompt: string;
+  schedule: RoutineSchedule;
+  target: RoutineProjectTarget;
+  skillId: string | null;
+  agentId: string | null;
+  enabled: boolean;
+  nextRunAt: number | null;
+  lastRun: unknown;
+  createdAt: number;
+  updatedAt: number;
+}
+
+export interface RoutineRun {
+  id: string;
+  routineId: string;
+  trigger: RoutineRunTrigger;
+  status: RoutineRunStatus;
+  projectId: string;
+  conversationId: string;
+  agentRunId: string;
+  startedAt: number;
+  completedAt: number | null;
+  summary: string | null;
+  error: string | null;
+}
+
+export interface RoutineRunHandlerStart {
+  projectId: string;
+  conversationId: string;
+  agentRunId: string;
+  completion: Promise<RoutineRunCompletion>;
+}
+
+export interface RoutineRunCompletion {
+  status: RoutineRunStatus;
+  summary?: string;
+  error?: string;
+}
+
+export type RoutineRunHandler = (input: {
+  routine: Routine;
+  trigger: RoutineRunTrigger;
+  startedAt: number;
+  runId: string;
+}) => Promise<RoutineRunHandlerStart>;
+
+export interface RoutinePersistence {
+  list(): Routine[];
+  insertRun(run: RoutineRun): void;
+  updateRun(id: string, patch: Partial<RoutineRun>): void;
+  getLatestRun(routineId: string): RoutineRun | null;
+}
+
+interface ScheduledTimer {
+  routineId: string;
+  timer: NodeJS.Timeout;
+  fireAt: Date;
+}
+
+// ---------- timezone math ----------
+
+// Returns the wall-clock parts of `atUtc` rendered in `timezone`. Uses
+// Intl.DateTimeFormat which Node ships with full-icu by default. If the
+// timezone is invalid the formatter throws — we catch upstream.
+function partsInTimezone(timezone: string, atUtc: Date): {
+  year: number;
+  month: number;
+  day: number;
+  hour: number;
+  minute: number;
+  second: number;
+  weekday: Weekday;
+} {
+  const dtf = new Intl.DateTimeFormat('en-US', {
+    timeZone: timezone,
+    hour12: false,
+    year: 'numeric',
+    month: '2-digit',
+    day: '2-digit',
+    hour: '2-digit',
+    minute: '2-digit',
+    second: '2-digit',
+    weekday: 'short',
+  });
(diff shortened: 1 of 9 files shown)


lefarcen (commented):
Solid implementation overall. The timezone math looks correct (two-pass approach handles DST properly), and the service architecture follows existing patterns.

**Code review (Lens A)**:

The core scheduling logic is well-designed:
- Two-pass timezone conversion correctly handles DST transitions
- Intl.DateTimeFormat usage avoids third-party dependencies
- setTimeout overflow protection (2^31ms cap + rechaining) prevents edge-case crashes
- Inflight map prevents duplicate concurrent runs

A few considerations for future hardening:

**P2 — Timezone validation timing**: isValidTimezone validates on routine creation, but if a timezone becomes invalid after creation (e.g., IANA data changes), nextRunAtForSchedule will return null and the routine silently will not fire. Consider adding a validation step in scheduleRoutine and logging a warning when fireAt is null.

**P2 — Race on rescheduleOne**: rescheduleOne clears the timer, refetches from persistence.list(), then schedules. If a routine is deleted between clear and refetch, it reschedules a deleted routine until the next rescheduleAll. Minor edge case, but worth noting.

**P3 — Test coverage**: No unit tests for nextRunAtForSchedule or DST edge cases (spring-forward, fall-back). The timezone math is correct, but DST boundary tests would future-proof against regressions.

**P3 — Error handling in UI**: RoutinesSection catches fetch errors but does not distinguish between network failures and validation errors. Users see a generic error message even when the backend returns a specific validation error (e.g., Invalid timezone). Consider parsing res.json().error and surfacing it.

**Minor observations**:
- nextWallTimeMatching 14-day walk-forward is a reasonable safety bound for weekly patterns
- timer.unref() correctly prevents blocking daemon shutdown
- UI timezone picker prepending local zone is a nice UX touch

**Reasoning review (Lens B)**: N/A (no design docs in this PR)

Ready from my side. 🎉

apps/daemon/src/routines.ts
```diff
@@ -0,0 +1,467 @@
...
+  day: number,
+  hour: number,
+  minute: number,
+): Date | null {
+  try {
+    const tentative = Date.UTC(year, month - 1, day, hour, minute, 0);
+    const t1 = tzOffsetMinutes(timezone, new Date(tentative));
+    const adjusted = tentative - t1 * 60_000;
+    const t2 = tzOffsetMinutes(timezone, new Date(adjusted));
+    return new Date(tentative - t2 * 60_000);
```
mrcfps: For spring-forward DST gaps, this conversion can return an instant that no longer round-trips to the requested wall clock. For example, `tzWallToUtc('America/New_York', 2026, 3, 8, 2, 30)` currently produces `2026-03-08T06:30:00.000Z`, which renders back as **01:30** local, so a `02:30` routine would fire an hour early on the transition day. Because routines are defined as local wall-clock schedules, firing before the requested time breaks the core scheduling contract. Please re-render the computed candidate in `timezone` and verify that the `year/month/day/hour/minute` still match the requested wall time; if they do not, either advance to the first valid post-gap instant or skip that day explicitly, and add a DST-gap fixture so this behavior stays covered.

mrcfps (commented):
@pftom Thanks for shipping the routines flow — I found one DST scheduling edge case that's worth tightening up before people rely on local-time delivery.


apps/daemon/src/routines.ts
```diff
@@ -0,0 +1,498 @@
...
+    const candidate1 = new Date(tentative - t1 * 60_000);
+    const t2 = tzOffsetMinutes(timezone, candidate1);
+    const candidate2 = new Date(tentative - t2 * 60_000);
+    if (matchesWallClock(timezone, candidate2, year, month, day, hour, minute)) {
+      return candidate2;
+    }
+    if (matchesWallClock(timezone, candidate1, year, month, day, hour, minute)) {
+      return candidate1;
+    }
+    return candidate1.getTime() > candidate2.getTime() ? candidate1 : candidate2;
```
mrcfps: When the scheduled wall time lands in the repeated hour on a fall-back day, `tzWallToUtc()` collapses the overlap down to a single instant before `nextWallTimeMatching()` compares it to `now`. That means a daily routine like `01:30 America/New_York` will skip the second 01:30 entirely if the daemon checks at `2026-11-01T05:45Z`: the helper resolves today's `01:30` to the first occurrence (`05:30Z`), sees it is already in the past, and then jumps straight to the next day even though there is still a valid `01:30` at `06:30Z`. Because routines are defined as local wall-clock schedules, this makes overlap-hour runs fire a day late once per DST fall-back. Please preserve both matching candidates for ambiguous wall times (or probe the later overlap candidate when the first one is already `<= now`) and add a fixture for the repeated-hour case so the intended behavior stays locked in.

mrcfps (commented):
@pftom Thanks for tightening the DST gap handling here — I found one remaining fall-back overlap case in the same scheduler path, so I left an inline note below.


apps/web/src/components/RoutinesSection.tsx
```diff
@@ -0,0 +1,680 @@
...
+  const [showForm, setShowForm] = useState(false);
+  const [form, setForm] = useState<FormState>(emptyForm);
+  const [submitting, setSubmitting] = useState(false);
+  const [busyId, setBusyId] = useState<string | null>(null);
+  const [expandedId, setExpandedId] = useState<string | null>(null);
+  const [historyTick, setHistoryTick] = useState(0);
+
+  const timezones = useMemo(() => {
+    const local = detectLocalTimezone();
+    const set = new Set<string>([local, ...COMMON_TIMEZONES]);
```
mrcfps: The picker is currently built from the local timezone plus a short hard-coded allowlist, but the backend validator and the public contract both accept any IANA timezone. That means users can schedule `America/New_York` or `Europe/Paris`, but they cannot pick other valid zones like `America/Phoenix` or `Africa/Johannesburg` unless one of those happens to be their local zone. Because the feature is advertised and validated as an IANA-timezone scheduler, this UI restriction quietly makes part of the supported API surface unreachable from Settings. Please either source the options from the platform (`Intl.supportedValuesOf("timeZone")` with a fallback) or explicitly narrow the backend/contract to the same curated set so the product surface stays consistent.

apps/web/src/components/RoutinesSection.tsx
```diff
@@ -0,0 +1,680 @@
...
+  try {
+    return Intl.DateTimeFormat().resolvedOptions().timeZone || 'UTC';
+  } catch {
+    return 'UTC';
+  }
+}
+
+// "GMT+8", "GMT-5:30", "GMT" — short label that mirrors the screenshot's
+// "Shanghai (GMT+8)" pattern for legibility.
+function gmtLabel(timezone: string, at = new Date()): string {
```
mrcfps, line 89: `gmtLabel()` always derives the offset from `new Date()`, and both the live preview line and the timezone dropdown reuse that value. For DST-observing zones this drifts seasonally: for example, a New York routine created in winter renders as `GMT-5`, but the same recurring schedule will run on `GMT-4` for the next several months once DST starts. Since this label is the main confirmation surface for when a routine will fire, showing today's offset can mislead users about the actual next occurrence. Please compute the displayed offset from the next scheduled fire time (or omit the GMT suffix for recurring schedules) so the UI stays truthful across DST boundaries.

mrcfps (commented):
@pftom Thanks for tightening the DST scheduling path here — the new overlap/gap fixtures make the backend behavior much easier to trust. I left two non-blocking follow-ups in the new Routines UI so the timezone surface stays aligned with what the daemon actually supports.


apps/web/src/components/RoutinesSection.tsx
```diff
@@ -0,0 +1,714 @@
...
+// Returns every IANA zone the platform recognizes, so the picker stays in
+// sync with the backend validator (which accepts any IANA timezone). Falls
+// back to a curated subset on older runtimes that lack `supportedValuesOf`.
+function listSupportedTimezones(): string[] {
+  try {
+    const fn = (Intl as { supportedValuesOf?: (key: string) => string[] })
+      .supportedValuesOf;
+    if (typeof fn === 'function') {
+      const list = fn('timeZone');
+      if (Array.isArray(list) && list.length > 0) return list;
```
mrcfps: `Intl.supportedValuesOf('timeZone')` does not include `UTC` on current runtimes (for example Node 24 returns only canonical region names), so this `return list` path drops the most common non-local zone unless the browser itself already runs in UTC. The backend validator and the new contract examples still accept `UTC`, which means a user on a non-UTC machine can no longer create a documented UTC routine from Settings even though the API supports it. Please prepend `UTC` alongside the detected local zone before deduping, or append it inside `listSupportedTimezones()`, so the picker stays aligned with the supported schedule surface.

mrcfps (commented):
@pftom Thanks for tightening the DST scheduling path and bringing the timezone picker closer to the backend contract. I found one remaining non-blocking gap in the new `supportedValuesOf` source, so I left an inline suggestion below.


mrcfps (approved):
@pftom I re-reviewed the latest head and verified the follow-up fixes around DST scheduling plus the timezone-picker contract: the daemon now covers spring-forward gaps and repeated-hour fall-back cases with focused fixtures, and the web picker now exposes the supported IANA zones while keeping `UTC` available. Thanks for iterating on the routines flow — this looks solid to me.

Issue #441, opened 2026-05-04
feat: Add Kilo CLI as a supported coding agent (ACP)


thirdeye-media:
## Summary

Proposing to add **Kilo** (`kilo`) as a supported coding agent CLI in Open Design. Kilo supports the **Agent Client Protocol (ACP)** via `kilo acp`, making it a drop-in addition alongside existing ACP agents like Hermes, Devin, and Kimi CLI.

## What Kilo Is

- **Binary**: `kilo` (installable via Homebrew, npm)
- **ACP support**: `kilo acp` command for programmatic integration over JSON-RPC/stdio
- **Model discovery**: Exposes 500+ models dynamically through its ACP interface
- **Non-interactive mode**: `kilo run --auto` or `kilo run --dangerously-skip-permissions`
- **Version**: currently at 7.2.34
- **Gateway**: Kilo Gateway supports 50+ providers (Anthropic, OpenAI, Google, DeepSeek, Mistral, NVIDIA, Cohere, Meta, etc.)

## Proposed Code Change

Adding one entry to `apps/daemon/src/agents.ts` (inserted alphabetically after `kimi`):

```typescript
{
  id: 'kilo',
  name: 'Kilo',
  bin: 'kilo',
  versionArgs: ['--version'],
  fetchModels: async (resolvedBin) =>
    detectAcpModels({
      bin: resolvedBin,
      args: ['acp'],
      timeoutMs: 15_000,
      defaultModelOption: DEFAULT_MODEL_OPTION,
    }),
  // Fallback models from common Kilo model aliases observed via `kilo models`.
  fallbackModels: [
    DEFAULT_MODEL_OPTION,
    { id: 'anthropic/claude-sonnet-latest', label: 'Claude Sonnet (latest)' },
    { id: 'anthropic/claude-opus-latest', label: 'Claude Opus (latest)' },
    { id: 'anthropic/claude-haiku-latest', label: 'Claude Haiku (latest)' },
    { id: 'openai/gpt-latest', label: 'GPT (latest)' },
    { id: 'openai/gpt-mini-latest', label: 'GPT Mini (latest)' },
    { id: 'google/gemini-pro-latest', label: 'Gemini Pro (latest)' },
    { id: 'google/gemini-flash-latest', label: 'Gemini Flash (latest)' },
  ],
  buildArgs: () => ['acp'],
  streamFormat: 'acp-json-rpc',
},
```

## Proof of Working Integration

Tested end-to-end on the local daemon:

[...]

PR #1034, opened 2026-05-09
feat(settings): add fetch models button for BYOK providers

Demoniooo:
**Motivation**
The BYOK settings currently only expose a static list of preset models. When users switch baseUrl or change provider, they have to look up the docs and type the model name by hand, which is error-prone. This PR lets users pull the actual list of available models from the provider using the key they already filled in.

**Summary**
- Added a "Fetch models" button to the Settings → BYOK form.
- First version supports OpenAI-compatible, Anthropic, and Google Gemini. Azure OpenAI still requires manually entering the deployment name.
- Fetched results are merged and deduplicated with the existing static suggestions. The currently filled `model` value is never overwritten, and the user's final selection continues to be saved through the existing config flow.

**Key Changes**
- New provider models contract: request includes `protocol`, `baseUrl`, `apiKey`, and optional `apiVersion`. Response is either a success with `models: { id, label }[]` or a classified failure status.
- New daemon endpoint `POST /api/provider/models`:
  - Reuses existing baseUrl safety validation, timeout handling, error classification, and key redaction.
  - OpenAI-compatible: `GET <baseUrl>/models`, parses `data[].id`.
  - Anthropic: `GET <baseUrl>/models?limit=1000` with `x-api-key` + `anthropic-version`, parses `data[].id` and `display_name`.
  - Google: `GET <baseUrl>/v1beta/models?key=...`, keeps only models that support `generateContent`, prefers `baseModelId` and falls back to extracting the id from `models/<id>`.
- Web layer adds a fetch wrapper and Settings state:
  - Caches fetched results within the current Settings dialog session, keyed by protocol + baseUrl + apiKey.
  - The "Fetch models" button is enabled when key/baseUrl are valid and the provider is not Azure. Shows a spinner while loading and renders inline status on failure.
  - Model dropdown source is now: remote models + provider preset/static suggestions + current model, deduplicated while preserving order.
- i18n: added English and Chinese strings for fetch button, loading state, fetched count, no models available, auth failure, invalid base URL, and the Azure manual-entry hint.

**Test Plan**
- Daemon tests:
  - OpenAI-compatible `/models` parses successfully and returns the model list.
  - Anthropic parses `display_name` correctly and covers the `limit=1000` request.
  - Google returns only models that support `generateContent`.
  - 401, 429, 5xx, timeout, and invalid/private baseUrl all return a classified failure, and the detail does not leak the API key.
- Web tests:
  - Button disabled/enabled state matches key/baseUrl/protocol.
  - After a successful fetch, the dropdown merges remote and static suggestions while preserving the current model.
  - When the current model is not in the remote list, the custom input is still shown.
  - Field changes clear stale fetch state, and failure states render correctly.
- Validation commands:
  - `pnpm --filter @open-design/daemon test -- connection-test`
  - `pnpm --filter @open-design/web test -- SettingsDialog`
  - `pnpm guard`
  - `pnpm typecheck`

**Assumptions**
- The API key is never sent automatically. The daemon is only contacted when the user explicitly clicks "Fetch models".
- Remote model lists are not persisted. They are cached only for the current Settings dialog session. The selected model continues to be persisted through the existing config flow.
- Azure deployment discovery is intentionally out of scope. The current BYOK form only collects an inference endpoint + api-key, while Azure deployment listing is a control-plane concern.
- External API references: OpenAI [List models](https://developers.openai.com/api/reference/resources/models/methods/list), Anthropic [List Models](https://platform.claude.com/docs/en/api/models/list), Google Gemini [models.list](https://ai.google.dev/api/models#v1beta.models.list), Azure OpenAI [deployment inference shape](https://learn.microsoft.com/en-us/azure/foundry/openai/reference).


mrcfps:
@Demoniooo I'm holding off on generating review comments for nexu-io/open-design#1034 because this pull request has merge conflicts right now.

lefarcen:
Hi @Demoniooo! 🎉

Thanks for the contribution — adding a fetch models button for BYOK providers is a solid UX improvement that removes manual lookup friction.

— open-design team

lefarcen (commented):
Hi @Demoniooo! Solid implementation of the model-discovery UX. The security discipline (API key redaction, baseUrl validation reuse, timeout handling) and test coverage are excellent. A few architectural observations:

**Code Review (Lens A):**

✅ **Security**: API key redaction is properly applied across all error paths (lines 237, 251, 285 in providerModels.ts). The existing `validateBaseUrl` prevents SSRF, and the 12s timeout prevents hangs.

✅ **Error handling**: The error classification (auth_failed / invalid_base_url / rate_limited / upstream_unavailable / timeout) matches the existing connection-test contract, so UI can render categorized inline status.

✅ **Testing**: Comprehensive — daemon tests for OpenAI/Anthropic/Google model parsing + auth/network failures, web tests for button enable/disable, cache invalidation, failure rendering, and custom model preservation. The `limit=1000` for Anthropic and `supportedGenerationMethods` filter for Google are captured in tests.

✅ **Model deduplication**: `uniqueModels` (line 69-79) handles duplicate model ids from providers.

**Lens B — Design completeness:**

✅ **Azure exclusion reasoning**: PR body explains Azure deployment discovery is control-plane, not inference-endpoint. The BYOK form only collects inference endpoint + api-key, so out-of-scope is justified.

✅ **Why not persist remote models**: PR body states "cached only for the current Settings dialog session" — this avoids stale model lists when providers add/remove models. The cache key (protocol+baseUrl+apiKey) ensures switching providers clears stale results. This is a defensible trade-off.

**Minor observations (non-blocking):**

- **Performance**: The Google API can return 500+ models (Kilo use case from issue #441). The current implementation fetches all in one request (`limit=1000` for Anthropic, no pagination for OpenAI/Google). If a provider returns 2000+ models, the 12s timeout may fire before the response completes. Consider: (a) document the limitation, or (b) add pagination if/when a real provider hits this limit.

- **Cache lifetime**: The session-scoped cache is cleared when baseUrl/apiKey/protocol changes (test verifies this). If user switches between two OpenAI-compatible providers (e.g., api.openai.com ↔ proxy.example.com) multiple times within one Settings dialog session, each switch re-fetches. This is correct (avoids showing stale Provider A models after switching to Provider B), just noting the trade-off.

- **Model id vs label**: For OpenAI-compatible providers, `label` defaults to `id`. For Anthropic, `display_name` is used when available. For Google, `displayName` is preferred over `baseModelId`. This heterogeneity matches provider APIs, so no action needed — just documenting the pattern.

**Verdict**: This is production-ready. The core security/correctness concerns are handled, tests are thorough, and the UX friction (manually typing model names after switching providers) is solved. The minor observations above are future optimizations, not blockers.

Deferring final approval to a maintainer. Nice work! 🎉

Demoniooo:
Conflicts resolved and rebased onto latest main. Ready for another look.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -276,6 +284,49 @@ export function canRunProviderConnectionTest(
   );
 }
 
+export function canFetchProviderModels(
+  config: Pick<AppConfig, 'apiKey' | 'baseUrl'>,
+  protocol: ApiProtocol,
+): boolean {
+  return (
+    protocol !== 'azure' &&
```
mrcfps, line 292: `canFetchProviderModels` only excludes Azure here, so the new button still enables on the Ollama Cloud tab as soon as the user fills in a valid base URL and API key. The first click then posts `protocol: "ollama"`, but the new daemon route below still rejects anything outside `anthropic|openai|azure|google` with HTTP 400. In practice that means the new feature always fails for Ollama users even though the UI presents it as available. Please either gate Ollama out in this helper/UI (with the same kind of explicit unsupported hint you added for Azure) or add daemon-side Ollama model listing support, and cover that behavior with a regression test.

mrcfps (commented):
@Demoniooo Thanks for the thoughtful fetch-models work — I walked through the new daemon/web flow and found one merge-safe follow-up that would make the UX line up with the supported-provider set a bit better.


lefarcen (commented):
@Demoniooo Nice rebase — conflicts resolved cleanly. @mrcfps caught the Ollama UI/daemon mismatch above (the button enables but daemon rejects `protocol: "ollama"` with 400). That's the main follow-up.

The rest of the implementation (security, tests, design trade-offs) remains solid from my prior review. Once you gate Ollama out in `canFetchProviderModels` or add daemon-side Ollama support (per mrcfps's comment), this is ready from my side.

Deferring final approval to a maintainer. 🎉

Demoniooo:
Fixed in 829df9ae. Excluded Ollama from `canFetchProviderModels` and added a manual-entry hint mirroring the Azure pattern, plus a regression test in `SettingsDialog.execution.test.tsx` and `SettingsDialog.test.ts`.

lefarcen (commented):
@Demoniooo Perfect — Ollama gating is now in place with the manual-entry hint and regression test coverage. mrcfps's finding is fully addressed.

The PR is ready from my side. Deferring final approval to a maintainer. 🎉

apps/daemon/src/providerModels.ts
```diff
@@ -0,0 +1,305 @@
...
+      return {
+        ok: false,
+        kind: 'unknown',
+        latencyMs,
+        status: response.status,
+        detail: redactSecrets(
+          err instanceof Error ? err.message : String(err),
+          [input.apiKey],
+        ),
+      };
```
mrcfps, line 258: On provider/gateway failures that come back as HTML or plaintext instead of JSON, this `JSON.parse` branch returns `kind: "unknown"` immediately and never reaches the `statusToKind(response.status)` handling below. In practice that means common 401/429/5xx fetch-model failures lose the auth/rate-limit/upstream classification that the rest of the Settings flow already relies on, so users see the generic fallback copy even though the HTTP status is enough to categorize the error. Please keep non-2xx parse failures on the status-mapping path here — for example, if parsing fails, fall back to a redacted `rawText`/parse-error detail but still return `kind: statusToKind(response.status)` — and add a regression test that exercises a non-JSON 401/429/5xx response.

mrcfps (commented):
@Demoniooo Thanks for the follow-up here — I re-checked the latest head after the Ollama fix and found one merge-safe error-classification gap in the new provider-model discovery path.


Demoniooo:
Fixed in 76b05a86. The catch block now records the parse error but lets control fall through to the existing `if (!response.ok)` branch, so non-JSON 401/429/5xx responses get classified by `statusToKind(response.status)` like JSON ones do. Added regression coverage in `tests/connection-test.test.ts` for all three statuses with mixed body shapes (JSON 401, plain text 429, HTML 503).

Note: 3 unrelated tests in `finalize-design.test.ts` fail locally on macOS due to `/var` vs `/private/var` symlink resolution in temp paths — they don't touch providerModels and aren't on this PR's path. The connection-test suite (47 tests) and the provider-models tests all pass.

lefarcen (commented):
@Demoniooo Excellent — the parse-error handling now preserves status-based classification for non-JSON failures, and the regression coverage exercises all three paths (401/429/5xx × JSON/plain/HTML). mrcfps's finding is fully addressed.

The PR is ready from my side. Deferring final approval to a maintainer. 🎉

mrcfps (commented):
@Demoniooo Thanks for the follow-up here — I re-checked the latest head after the Ollama and error-classification fixes, and the fetch-model flow itself looks solid. I found one merge-safe i18n follow-up around the newly added settings copy for non-English locales.


Location: apps/web/src/i18n/locales/de.ts RIGHT lines 99-114

The new `settings.fetchModels*` strings and `settings.azureModelFetchHint` are still hard-coded in English here, and the same pattern repeats in the other locale files touched by this PR (`ar`, `es-ES`, `fa`, `fr`, `hu`, `id`, `ja`, `ko`, `pl`, `pt-BR`, `ru`, `tr`, `uk`, `zh-TW`). Because the surrounding settings keys in these files are already localized, users who selected those locales will now see a mixed-language Settings panel specifically around the new fetch-model flow. Please either add translated copy for the shipped locales before merge, or drop these per-locale overrides and let untranslated locales fall back to the base English strings until the translations are ready.



Demoniooo:
Fixed in f4f35246.

**Removed English `fetchModels*` and `azureModelFetchHint` overrides** from non-English locales. Kept only in `en` and `zh-CN` (which has real translations). The other 14 locales now fall through to English automatically instead of presenting hard-coded English under a translated key

Validated:
- `pnpm --filter @open-design/web test -- i18n` — all 66 files / 602 tests pass
- `pnpm typecheck` — clean

apps/daemon/src/server.ts
```diff
@@ -6155,6 +6156,62 @@ export async function startServer({
...
+    const abortIfResponseClosed = () => {
+      if (!res.writableEnded) controller.abort();
+    };
+    req.on('close', abortIfRequestAborted);
+    res.on('close', abortIfResponseClosed);
+    const body = req.body || {};
+    const protocol = body.protocol;
+    if (
+      typeof protocol !== 'string' ||
+      !['anthropic', 'openai', 'azure', 'google'].includes(protocol)
```
mrcfps: `ProviderModelsRequest.protocol` is still typed as `ConnectionTestProtocol`, which includes `ollama`, and this route already has an `unsupported_protocol` result path in `listProviderModels()`. Because the guard here rejects anything outside `anthropic|openai|azure|google` with HTTP 400 first, a valid contract-shaped `protocol: "ollama"` request never reaches the classified response path and callers fall back to the generic `unknown` transport error instead of the explicit unsupported hint. Please either widen this allowlist to every `ConnectionTestProtocol` value and let `listProviderModels()` return `unsupported_protocol`, or narrow the exported request type so the contract only advertises the subset this endpoint actually accepts.

mrcfps (commented):
@Demoniooo Thanks for the fetch-models work — the provider coverage, redaction, and UI wiring look solid overall. I found one merge-safe follow-up around the new provider-models endpoint contract.


lefarcen (commented):
@Demoniooo Perfect — the i18n fix is clean. Removed English overrides from 14 non-English locales while keeping real translations in `zh-CN`, and the fallback to English happens safely at runtime. mrcfps's finding is fully addressed.

The PR is ready from my side. Deferring final approval to a maintainer. 🎉

apps/daemon/src/server.ts
Demoniooo: Fixed in c8d9458c. Went with widening the allowlist:

The `/api/provider/models` route now accepts all 5 `ConnectionTestProtocol` values past the 400 input check. `ollama` flows through to `listProviderModels()`, which returns:

```json
{ "ok": false, "kind": "unsupported_protocol" }
```

Callers now see an explicit "model discovery not supported for this protocol" hint instead of a generic transport/unknown error.

Regression test added:
- `protocol: "ollama"` request returns HTTP 200
- response body is `{ ok: false, kind: "unsupported_protocol" }`
- no upstream fetch is actually issued

Validated: `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/connection-test.test.ts` — 48 tests passed.

apps/daemon/src/providerModels.ts
```diff
@@ -0,0 +1,310 @@
...
+
+function providerModelsUrl(protocol: ConnectionTestProtocol, baseUrl: string, apiKey: string): string {
+  if (protocol === 'openai') return appendVersionedApiPath(baseUrl, '/models');
+  if (protocol === 'anthropic') {
+    const url = new URL(appendVersionedApiPath(baseUrl, '/models'));
+    url.searchParams.set('limit', '1000');
+    return url.toString();
+  }
+  if (protocol === 'google') {
+    const url = new URL(`${baseUrl.replace(/\/+$/, '')}/v1beta/models`);
```
mrcfps, line 159: Google's `models.list` API is paginated: if `pageSize` is omitted it returns 50 models per page and a `nextPageToken` for the rest. This branch always builds a single `/v1beta/models` request and `listProviderModels()` only consumes that first response, so any account or endpoint with more than 50 Gemini models will silently get a truncated picker list even though this feature is meant to fetch the available models. Please either request a larger page (`pageSize=1000`) and/or loop on `nextPageToken` until exhausted, then add a regression in `apps/daemon/tests/connection-test.test.ts` that stubs a two-page Gemini response so we keep returning the full set.

mrcfps (commented):
@Demoniooo Thanks for the fetch-models work — I re-checked the latest head and found one merge-safe follow-up in the new Google model-discovery path.


lefarcen (commented):
@Demoniooo Perfect — the protocol-contract finding is fixed. `ollama` now flows through the daemon route, hits `listProviderModels()`, and returns explicit `unsupported_protocol` instead of the generic `unknown` transport error. Regression test is solid.

The PR is ready from my side. Deferring final approval to a maintainer. 🎉

mrcfps (approved):

PR #1036, opened 2026-05-09
fix(connectors): preserve OAuth state and advertised tool counts

alchemistklk:
## Summary
- stop clearing daemon OAuth pending state when the browser/system launch cannot be confirmed
- keep launch failures out of local pending UI state so users can retry immediately
- show advertised connector tool counts from `toolCount` instead of the curated preview subset

## Tests
- pnpm --filter @open-design/web exec vitest run tests/providers/registry.test.ts tests/components/EntryView.test.ts

### DIFF
--- a/apps/web/src/components/ConnectorsBrowser.tsx
+++ b/apps/web/src/components/ConnectorsBrowser.tsx
@@ -312,15 +312,6 @@ export function clearConnectorAuthorizationPending(
 }
 
 export function getConnectorDisplayToolCount(connector: ConnectorDetail): number {
-  // Prefer the hand-curated catalog subset so the card/header badge stays
-  // stable across Composio hydration: `toolCount` is sourced from the
-  // provider's totalItems (a GitHub-style toolkit jumps from 2 to 48 once an
-  // API key is configured), and `tools.length` shifts the same way after
-  // hydration. `curatedToolNames` is fixed at the catalog and is what the
-  // "N tools" badge is meant to reflect (issue #748).
-  if (connector.curatedToolNames && connector.curatedToolNames.length > 0) {
-    return connector.curatedToolNames.length;
-  }
   return connector.toolCount ?? connector.tools.length;
 }
 
--- a/packages/contracts/src/api/connectors.ts
+++ b/packages/contracts/src/api/connectors.ts
@@ -45,8 +45,8 @@ export interface ConnectorDetail {
    * read-only with auto approval. Used by the agent layer to gate
    * which tools are invocable. Note: this list **grows** on Composio
    * hydration (a GitHub-style provider can add tens of read tools to
-   * the catalog baseline of 2), so it is not the right anchor for a
-   * stable UI summary count — see `curatedToolNames` for that.
+   * the catalog baseline of 2), so it is not the right anchor for the
+   * advertised provider tool count — see `toolCount` for that.
    *
    * Optional in the type only to keep test fixtures terse — the daemon
    * always populates this from `connectorDefinitionToDetail` so wire
@@ -56,11 +56,10 @@ export interface ConnectorDetail {
   /**
    * Hand-curated catalog subset. Stable across hydration: never
    * extended by provider discovery, only the static catalog names.
-   * UIs surfacing a single "N tools" summary (the connector card and
-   * drawer header badges) should read this so the displayed count
-   * doesn't lurch when an API key flips on (issue #748). The full
-   * provider inventory is still discoverable in the drawer's tools
-   * section, which renders `tools` directly.
+   * This is used for featured/curated preview ordering and does not
+   * represent the advertised provider inventory. UI summary badges
+   * should use `toolCount` when present, while the drawer's rendered
+   * tool rows still come from `tools` directly.
    *
    * Optional in the type only for fixture brevity, see `allowedToolNames`.
    */
--- a/packages/contracts/src/api/connectors.ts
+++ b/packages/contracts/src/api/connectors.ts
@@ -56,10 +56,11 @@ export interface ConnectorDetail {
   /**
    * Hand-curated catalog subset. Stable across hydration: never
    * extended by provider discovery, only the static catalog names.
-   * This is used for featured/curated preview ordering and does not
-   * represent the advertised provider inventory. UI summary badges
-   * should use `toolCount` when present, while the drawer's rendered
-   * tool rows still come from `tools` directly.
+   * This preserves the static catalog baseline for consumers that
+   * need that curated subset, but it does not represent the
+   * advertised provider inventory. UI summary badges should use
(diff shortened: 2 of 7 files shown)


apps/web/src/components/ConnectorsBrowser.tsx
```diff
@@ -312,15 +312,6 @@ export function clearConnectorAuthorizationPending(
 }
 
 export function getConnectorDisplayToolCount(connector: ConnectorDetail): number {
```
lefarcen, line 314: **P1** — This change reintroduces the count-jump regression that `curatedToolNames` was added to prevent in #748.

The daemon still documents `toolCount` as provider inventory derived from `toolPage.totalItems`, so connectors like GitHub jump from a catalog baseline (2 curated tools in catalog) to the advertised provider count (846) after hydration. That makes the "N tools" badge unstable and misleading.

The original #748 comment explained: *"Prefer the hand-curated catalog subset so the card/header badge stays stable across Composio hydration."*

**Recommendation:**
- Keep `curatedToolNames.length` for card/drawer header badges, OR
- If the product intent is now "show advertised provider inventory", change the UI label/copy to make that explicit (e.g., "846 tools available") and update contract comments/tests that currently say badges should use `curatedToolNames`.

Which direction did you intend?

lefarcen (changes requested):
@alchemistklk — thanks for the quick follow-up on connector state management.

Found one P1 regression: removing the `curatedToolNames` logic reintroduces the count-jump problem that #748 fixed. See inline.

OAuth cancel removal looks correct — matches the "browser launch cannot be confirmed" intent from #907.

packages/contracts/src/api/connectors.ts
```diff
@@ -56,11 +56,10 @@ export interface ConnectorDetail {
   /**
    * Hand-curated catalog subset. Stable across hydration: never
```
lefarcen, line 57: **P3** — This comment says `curatedToolNames` is "used for featured/curated preview ordering", but the current web code (`ConnectorsBrowser.tsx`) doesn't use it for that purpose. Either:
- Wire `curatedToolNames` into the claimed preview-ordering behavior, OR
- Update this comment to reflect its actual current purpose (if any)

Otherwise the contract doc is misleading.

lefarcen (changes requested):
@alchemistklk — thanks for the doc updates, but the P1 is not yet resolved.

Your new commits reframe `curatedToolNames` as "for preview ordering" and say "UI summary badges should use toolCount". However:

1. **This is a breaking UX change from #748** — the original intent was badge stability (2 tools → stays 2 after hydration). Switching to `toolCount` means the badge now jumps from 2 → 846, which is the exact regression #748 fixed. Updating comments to match the new implementation doesn't close the regression without **explicit maintainer/product sign-off**.

2. **New P3**: The updated contracts say `curatedToolNames` is "used for featured/curated preview ordering", but the current web code doesn't use it for that. This makes the contract doc misleading.

**Recommendation:**
Before merging, confirm with @lefarcen (or product owner) whether the badge should show:
- **Curated catalog baseline (2 tools, stable across hydration)** — original #748 behavior, OR
- **Advertised provider inventory (846 tools, jumps after hydration)** — your new implementation

If the latter is the product decision, document that as a **breaking change** and update the #748 linked issue. If the former, revert `ConnectorsBrowser.tsx:314` to the `curatedToolNames.length` logic.

See inline for the new P3.

mrcfps (dismissed):
@alchemistklk I verified the OAuth launch path now preserves daemon-side pending authorization while still clearing the local pending UI state so users can retry immediately, and I checked the tool-count contract update so the card/drawer badges consistently read the advertised `toolCount`. I also ran `pnpm install` to hydrate the workspace and `pnpm --filter @open-design/web test -- tests/providers/registry.test.ts tests/components/EntryView.test.ts`, which passed.

Thanks for tightening up both the retry flow and the connector count semantics. 🙏


lefarcen (dismissed):
P1 resolved — @mrcfps's approval on the tool-count contract update addresses the product sign-off needed for the breaking UX change.

OAuth state management looks correct. Tests pass. Ship it. 🚀

apps/web/src/components/ConnectorsBrowser.tsx
```diff
@@ -312,15 +312,6 @@ export function clearConnectorAuthorizationPending(
...
-  // Prefer the hand-curated catalog subset so the card/header badge stays
-  // stable across Composio hydration: `toolCount` is sourced from the
-  // provider's totalItems (a GitHub-style toolkit jumps from 2 to 48 once an
-  // API key is configured), and `tools.length` shifts the same way after
-  // hydration. `curatedToolNames` is fixed at the catalog and is what the
-  // "N tools" badge is meant to reflect (issue #748).
-  if (connector.curatedToolNames && connector.curatedToolNames.length > 0) {
-    return connector.curatedToolNames.length;
-  }
   return connector.toolCount ?? connector.tools.length;
```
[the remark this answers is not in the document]
alchemistklk: Thanks for flagging this. We are intentionally keeping the new behavior: the card/drawer summary should show the advertised provider inventory count from `toolCount`, not the curated static subset. This matches the current Composio optimization goal: show users the actual available toolkit scale before they open/hydrate the drawer.

I updated the contract and daemon comments in `303acb56` so `curatedToolNames` is no longer documented as the badge source; it now preserves only the static curated baseline for consumers that need that subset.

apps/web/src/components/ConnectorsBrowser.tsx
alchemistklk: Confirmed product direction: for this PR, the badge should show the advertised provider inventory count (`toolCount`) rather than the stable curated baseline. That means GitHub showing the larger Composio count is intentional, not the #748 behavior.

To make this explicit, `303acb56` updates the contracts/daemon comments so the documented semantics match the implementation: `toolCount` drives summary badges, `curatedToolNames` remains the static curated baseline, and rendered drawer rows still come from `tools`.

packages/contracts/src/api/connectors.ts
alchemistklk: Good catch. The "preview ordering" wording was inaccurate because the current web path does not use `curatedToolNames` for ordering. I fixed that in `303acb56`: the contract now says `curatedToolNames` preserves the static catalog baseline for consumers that need that subset, while summary badges use `toolCount` and drawer rows render from `tools`.

mrcfps (approved):
@alchemistklk I verified the OAuth launch path now preserves daemon-side pending authorization when the browser launch cannot be confirmed, while the local pending UI state still clears so users can retry immediately. I also checked the connector-count contract updates and ran `pnpm install` plus `pnpm --filter @open-design/web test -- tests/providers/registry.test.ts tests/components/EntryView.test.ts`, which passed.

Thanks for tightening up both the retry flow and the connector count semantics. 🙏

PR #1003, opened 2026-05-09
feat(daemon+web): add install/uninstall for skills & design systems

encyc:
## Summary

Implements **Phase 2** of #497 — install/uninstall for skills and design systems via the Library settings UI.

- **Daemon**: multi-directory scanning (`skills/` + `~/.open-design/skills/`), install from GitHub (shallow clone) or local path (symlink), uninstall user-installed items only
- **Web**: install form with GitHub/Local tab, source badge ("Built-in" / "Installed"), uninstall button on installed cards
- **Contracts**: `source` field on `SkillSummary`/`DesignSystemSummary`, new `InstallInput`/`UninstallResponse` types
- **i18n**: native-language translations for all 17 locales

Closes #497 (Phase 2)

## Test plan

- [ ] `pnpm typecheck` — clean
- [ ] `pnpm --filter @open-design/daemon test` — all tests pass
- [ ] `pnpm guard` — clean
- [ ] Settings → Library → Skills tab shows "Built-in" badge on shipped skills
- [ ] Click "Install..." → paste a GitHub skill repo URL → skill appears with "Installed" badge
- [ ] Uninstall an installed skill → removed from list, no uninstall on built-in items
- [ ] Install from local path works (valid SKILL.md required)
- [ ] Invalid URL/path shows error message
- [ ] Same flows work for Design Systems tab
- [ ] Installed skill is available in project creation picker

PR #1037, opened 2026-05-09
fix(daemon): follow symlinked skill and design-system dirs

desmond-rai:
## Summary

Fixes OpenDesign daemon discovery for externally-managed skill and design-system packs installed as symlinked directories.

`fs.Dirent.isDirectory()` reports false for a symlink itself, even when the target is a directory. The daemon was therefore silently skipping symlinked entries under `skills/` and `design-systems/`, so a valid pack with reachable `SKILL.md` / `DESIGN.md` files did not appear in the API or chooser UI.

This adds a small helper in both loaders that accepts either real directories or symlinks whose target resolves to a directory. Broken symlinks continue to be skipped.

## Impact

- No behavior change for normal directory-based bundled skills or design systems.
- Enables sibling-repo design packs to be installed as live symlinks instead of copied folders.
- Avoids stale copied packs when designers edit the source pack in its own repository.

## Validation

- Locally verified a symlinked ReturningAI design system and six symlinked skills are visible through the daemon/web APIs after restart.

### DIFF
--- a/apps/daemon/src/design-system-showcase.ts
+++ b/apps/daemon/src/design-system-showcase.ts
@@ -740,6 +740,70 @@ function renderReturningAiShowcase(input: {
     .swatch.muted { background: #aaa9a9; color: #141414; }
     .swatch.border { background: #575757; }
     .swatch.accent { background: #e5971d; color: #1b1c1d; font-weight: 700; }
+    .tweak-lab {
+      display: grid;
+      grid-template-columns: minmax(0, 1fr) 310px;
+      gap: 12px;
+      align-items: stretch;
+    }
+    .tweak-stage {
+      border: 1px solid var(--border);
+      background: var(--surface);
+      border-radius: 8px;
+      padding: 16px;
+      min-width: 0;
+      display: grid;
+      gap: 12px;
+    }
+    .tweak-preview-tabs { display: flex; gap: 8px; flex-wrap: wrap; }
+    .tweak-preview-tab {
+      min-height: 32px;
+      padding: 0 12px;
+      border: 1px solid var(--border);
+      border-radius: 4px;
+      color: var(--muted);
+      display: inline-flex;
+      align-items: center;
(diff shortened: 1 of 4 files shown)


lefarcen (commented):
Hey @desmond-rai! 👋

This is a clean fix for the symlink discovery issue. The helper `entryIsDirectoryFollowingLinks` handles the `fs.Dirent.isDirectory()` false-negative for symlinks correctly, and broken symlinks are silently skipped as intended.

- ✅ Real directories pass through unchanged
- ✅ Symlinks are `stat`'d to verify target is a directory
- ✅ Broken symlinks return `false` (same outcome as before)
- ✅ Error handling matches existing daemon pattern (silent skip)

The implementation is consistent with the codebase. Small suggestion for consideration (non-blocking): if this repo requires tests for daemon loader fixes, a regression test covering the symlink scenario would be valuable — e.g., create a temp symlinked `skills/test-skill/` pointing to a real directory with `SKILL.md`, verify it appears in `listSkills()` output. But that's a repo-wide test policy question, not a blocker for this PR.

**Related context:** This directly supports #497 (install/uninstall flow) and #1003 (Phase 2 local path installs). Without this fix, symlinked design-system and skill packs would exist on disk but be omitted from `/api/skills`, `/api/design-systems`, and the picker.

Looks good to me; deferring final approval to a maintainer. Thanks for the contribution!

lefarcen:
@mrcfps This PR's current head b86885c0 is waiting on your review.

PerishCode:
Maintainer stale-followup triage: the author's last visible action appears to be over 7 days old, so I marked this PR for the long-inactive PR sweep.

I am not taking maintainer code ownership in the lightweight Bug Workflow pass because the PR is still draft and touches high-risk daemon skill/design-system discovery surfaces with no current CI signal. It needs either author follow-up to make it ready for review or a deeper maintainer-owned takeover.

PR #1043, opened 2026-05-09
refactor(daemon): split route registration

nettee:
## Summary
- Split daemon server route handling into focused domain route modules.
- Add shared server context plumbing for route registrars.
- Update app-level agent guidance and revert transient spec changes.

## Validation
- Not run in this session.


lefarcen (commented):
Hi @nettee! Nice refactor — splitting the monolithic server.ts into focused domain route modules is exactly the right architectural move at this codebase scale.

I ran both Lens A (code correctness) and Lens B (reasoning completeness on the AGENTS.md guidance) review. Found 2 P2 non-blocking suggestions + 1 P3 polish nit below. Overall this is sound work; the findings are about hardening the new structure, not about the refactor being wrong.

apps/daemon/src/active-context-routes.ts
```diff
@@ -0,0 +1,76 @@
+import type { Express } from 'express';
+
+type AnyRouteDeps = Record<string, any>;
```
lefarcen: **P2 — Route registrar API surface consistency**

The new route modules export `register*Routes(app, ctx)` with heterogeneous `ctx` shapes across modules:
- `RegisterActiveContextRoutesDeps` requires `{ db, http, projectStore }`
- `RegisterChatRoutesDeps` requires `{ db, design, http, chat, agents, critique, validation }`
- Other modules likely have different shapes (I can see 12 route modules total)

**Risk**: If `server.ts` wiring forgets a required dep in any module's ctx, that module's routes will fail at runtime (not compile-time) when they try to destructure missing props. The current `AnyRouteDeps` escape hatch in each module's interface definition bypasses TypeScript's ability to catch these mismatches.

**Fix direction**: Consider a shared `ServerContext` base type that all route modules' Deps extend from, plus a compile-time assertion in server.ts that verifies the wired ctx object satisfies all route modules' union of deps. This shifts "missing dep" errors from runtime → build time.

apps/daemon/src/active-context-routes.ts
```diff
@@ -0,0 +1,76 @@
...
+  const { getProject } = ctx.projectStore;
+  const getResolvedPort = () => resolvedPortRef.current;
+
+  // Soft "what is the user looking at right now in Open Design?" channel. The
+  // web UI POSTs the current project + file on every route change; the MCP
+  // surface reads it so a coding agent in another repo can resolve "the design
+  // I have open" without the user typing the project id. In-memory only -
+  // daemon restart clears it.
+  let activeContext: { projectId: string; fileName: string | null; ts: number } | null = null;
+  const ACTIVE_CONTEXT_TTL_MS = 5 * 60 * 1000;
```
lefarcen: **P3 — Active context TTL magic number**

`ACTIVE_CONTEXT_TTL_MS = 5 * 60 * 1000` (5 min) is defined as an in-file constant. If this timeout needs tuning based on production telemetry later, developers won't know where to look (12 route files to grep).

**Suggestion**: Extract timing constants into a shared `config.ts` or `constants.ts` at the daemon module root, co-located with other daemon-wide tuning knobs.

apps/AGENTS.md
```diff
@@ -16,6 +16,11 @@ Follow the root `AGENTS.md` first. This file only records module-level boundarie
 - `apps/daemon/sidecar/` contains the daemon sidecar entry.
 - CLI/agent argument changes or stdout parser changes belong in `apps/daemon/src/agents.ts` and the matching parser tests.
 
+### Router layout
```
lefarcen, line 19: **P2 — Incomplete router layout guidance**

The new "Router layout" section says:
> "Existing daemon domain endpoints belong in the matching daemon route file; avoid adding route handlers directly to server.ts unless the route is bootstrap-wide or has no clear domain owner."

**Missing**: What constitutes "bootstrap-wide" vs "domain" is undefined. Examples would help:
- Is `/api/health` bootstrap-wide? (Yes, presumably — doesn't belong in chat-routes)
- Is `/api/active` bootstrap or active-context domain? (The PR puts it in active-context-routes, implying "domain")
- When does a new route deserve its own `-routes.ts` vs being added to an existing one?

**Also missing**: The guidance says "keep sections broad and reuse existing sections before adding a new one" but doesn't explain how to decide when an existing section's scope is exhausted. For a 12-module split, what's the right granularity heuristic?

**Recommendation**: Add 2-3 concrete examples of recent routing decisions (e.g., "Active context routes are semantically separate from chat routes even though both serve the web UI, because they represent different domains: transient UI focus vs persistent conversation state.") and a rule of thumb for when to split (e.g., "If a new endpoint has zero shared deps with existing routes in a module, consider a new module.").

lefarcen (commented):
Second pass — author pushed new commits addressing 2 of 3 prior findings:

✅ **P3 (TTL constant)**: Fixed. The magic number is now `ACTIVE_CONTEXT_TTL_MS` in `constants.ts`.

✅ **P2 (AGENTS.md guidance)**: Fixed. The new Router layout section gives clear domain/bootstrap split guidance and split-vs-reuse criteria.

⚠️ **P2 (route API consistency)**: Partially fixed. The new `RouteContextContract` + `assertServerContextSatisfiesRoutes` is a step forward, but `server.ts` is still `// @ts-nocheck`, so the compile-time check doesn't actually fire. The helper only proves the declared `ServerContext` type *shape* covers the route dep union; it won't catch the assembled object or individual `register*Routes(...)` calls missing a dependency at compile time.

**New finding (P2)**: The `assertServerContextSatisfiesRoutes(...)` call at server.ts:3593 does not deliver compile-time safety because `@ts-nocheck` suppresses all type errors in that file. Consider either:
- Removing `@ts-nocheck` from `server.ts` (if feasible), OR
- Moving route wiring into a separate `.ts` file without `@ts-nocheck`, OR
- Accepting the runtime-only dependency check + documenting it explicitly

This is an internal PR so I'm not blocking — just flagging for awareness. The refactor is solid and the new structure makes route ownership much clearer.

apps/daemon/src/media-routes.ts
```diff
@@ -0,0 +1,319 @@
...
+        })
+        .catch((err: any) => {
+          task.status = 'failed';
+          task.error = {
+            message: String(err && err.message ? err.message : err),
+            status: typeof err?.status === 'number' ? err.status : 400,
+            code: err?.code,
+          };
+          task.endedAt = Date.now();
+          notifyTaskWaiters(task);
```
mrcfps, line 194: Moving this code into registerMediaRoutes dropped the DB writes that used to happen when a task transitions from queued → running and again on done / failed. createMediaTask() only inserts the initial queued row, and notifyTaskWaiters() just wakes listeners / schedules GC, so /api/projects/:id/media/tasks will keep reading stale queued rows from listMediaTasksByProject(...) and completed file / error metadata will be lost after a restart. Please thread a persistMediaTask-style dependency into this registrar (or wrap these state transitions in helpers) and call it immediately after each task.status / task.file / task.error mutation, before notifying waiters.

mrcfps (changes requested):
Thanks for the route split here — the new module layout is much easier to follow. I found one behavior regression in the media task flow that looks worth fixing before this lands.


lefarcen (commented):
Third pass — author addressed mrcfps's media task DB writes regression:

✅ The new head adds `persistMediaTask` calls at queued→running, done, and failed transitions. Task state changes now persist to DB, so `/api/projects/:id/media/tasks` won't read stale rows and metadata survives restarts.

mrcfps's CHANGES_REQUESTED is resolved. Internal PR → deferring final approval to mrcfps / maintainer.

mrcfps (dismissed):
@nettee I re-reviewed the latest head and verified the media task persistence fix in `apps/daemon/src/media-routes.ts`/`server.ts`, along with the route-split and context-plumbing changes across the extracted registrars. I didn't find any additional correctness, safety, or maintainability issues in the changed ranges on this head.

Thanks for pushing through a large refactor here — the new route layout is much easier to follow. 🙏


apps/daemon/src/route-context-contract.ts
```diff
@@ -0,0 +1,34 @@
...
+  & RegisterMediaRoutesDeps
+  & RegisterProjectArtifactRoutesDeps
+  & RegisterProjectExportRoutesDeps
+  & RegisterProjectFileRoutesDeps
+  & RegisterProjectRoutesDeps
+  & RegisterProjectUploadRoutesDeps
+  & RegisterStaticResourceRoutesDeps;
+
+type Assert<T extends true> = T;
+type ServerContextCoversRouteDeps = Assert<ServerContext extends AllRegisteredRouteDeps ? true : false>;
```
mrcfps, line 32: The new `assertServerContextSatisfiesRoutes(...)` check does not actually prove the registrars receive the nested members they destructure. `AllRegisteredRouteDeps` is built from `RouteDeps<...>`, but `RouteDeps` comes from `ServerContext`, and every field in `ServerContext` is currently typed as `any` (`http`, `projectStore`, `media`, etc.). That means a value like `{ http: {}, projectStore: {} }` still satisfies the contract as far as TypeScript is concerned, so missing members such as `sendApiError` or `getProject` would still slip to runtime. On top of that, `server.ts` is still under `// @ts-nocheck`, so this call site is not type-checked there today either.

To make this contract pull its weight, I'd switch the shared dependency bags to concrete interfaces (for example `HttpDeps`, `ProjectStoreDeps`, `MediaDeps`, ...), use those types in `ServerContext`, and keep the assertion in a file that is actually checked by `tsc` (or remove `@ts-nocheck` from `server.ts`). That way a missing nested dependency fails at build time instead of creating a false sense of safety.

mrcfps (commented):
@nettee Thanks for pushing this refactor through — the route split is much easier to navigate. I found one non-blocking maintainability issue in the new dependency-contract check that seems worth tightening before the team relies on it for wiring safety.


nettee:
@codex review

apps/daemon/src/server.ts
```diff
@@ -5921,7 +2644,6 @@ export async function startServer({
     const prompt = composeSystemPrompt({
```
lefarcen, line 2657: **P1** — `streamFormat` is no longer passed into `composeSystemPrompt`. This parameter controls whether the "Do not emit tool_calls" rule appends (only when `streamFormat === 'plain'`). Without it, plain/BYOK runs lose that safety instruction. Restore: `streamFormat`.

apps/daemon/src/route-context-contract.ts
lefarcen, line 32: **P2** (mrcfps's prior concern still unaddressed) — `ServerContext` members are still `any`, and `server.ts` is `@ts-nocheck`, so this assertion doesn't prove nested members like `ctx.http.sendApiError` or `ctx.projectFiles.writeProjectFile` actually exist. Consider concrete interfaces (`HttpDeps`, `ProjectStoreDeps`, ...) or remove `@ts-nocheck` from `server.ts`.

lefarcen (dismissed):
Fourth pass — found one regression + one unaddressed prior concern:

**P1 (new)**: `composeDaemonSystemPrompt` stopped passing `streamFormat` into `composeSystemPrompt` at server.ts:2644. This breaks plain/BYOK runs — they lose the "Do not emit tool_calls" safety instruction that only appends when `streamFormat === 'plain'`.

**P2 (from mrcfps's prior comment)**: `assertServerContextSatisfiesRoutes` still doesn't deliver type safety. `ServerContext` fields remain `any`, and `server.ts` is still `@ts-nocheck`, so the compile-time check can't catch missing nested members like `ctx.http.sendApiError`.

Internal PR → no blocking on merge decision, but these warrant attention before ship.

apps/daemon/src/chat-routes.ts
```diff
@@ -0,0 +1,927 @@
+import type { Express } from 'express';
+import type { RouteDeps } from './server-context.js';
+
+export interface RegisterChatRoutesDeps extends RouteDeps<'db' | 'design' | 'http' | 'chat' | 'agents' | 'critique' | 'validation' | 'lifecycle'> {}
+
+export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
```
mrcfps, line 6: The routine API block from `main` never got re-registered in this refactor. On `main`, `server.ts` still exposes `GET/POST /api/routines`, `GET/PATCH/DELETE /api/routines/:id`, `POST /api/routines/:id/run`, and `GET /api/routines/:id/runs`, but this new registrar only wires chat/run/config/critique handlers and nothing else registers those routes on the current head. That breaks the shipped routines UI in `apps/web/src/components/RoutinesSection.tsx` (for example the fetches at lines 385, 437, 460, 478, and 500 now hit 404s) even though `routineService` is still started in `server.ts`. Please move the old routine handlers into this registrar or a dedicated `registerRoutineRoutes(...)` call and keep their same-origin guard/DB wiring intact so list/create/run/edit/delete continues to work.

mrcfps (changes requested):
@nettee Thanks for the refactor here — the split route modules are much easier to navigate. I found two blocking API regressions on this head: the routine endpoints are no longer registered, and the skill/design-system install/uninstall endpoints were dropped during the static-resource extraction.


Location: apps/daemon/src/static-resource-routes.ts RIGHT line 17

This extraction dropped the mutating library routes that existed on `main`: `POST /api/skills/install`, `DELETE /api/skills/:id`, `POST /api/design-systems/install`, and `DELETE /api/design-systems/:id`. Nothing else on this head registers them, but the web registry provider still calls them (`apps/web/src/providers/registry.ts:1296-1348`), so install/uninstall from the Library UI will regress to 404s. Please restore those handlers here (or in another registrar wired from `server.ts`) and preserve the local-origin checks around `installFromTarget(...)`/`uninstallById(...)`, otherwise this refactor removes existing library management functionality.

apps/daemon/src/chat-routes.ts
```diff
@@ -0,0 +1,949 @@
...
```
mrcfps, line 6: The routine API block from `main` never got re-registered in this extraction. `server.ts` still starts `routineService`, and the web client still calls `GET/POST /api/routines`, `GET/PATCH/DELETE /api/routines/:id`, `POST /api/routines/:id/run`, and `GET /api/routines/:id/runs` from `apps/web/src/components/RoutinesSection.tsx`, but this registrar only wires chat/run/proxy endpoints and nothing else on this head defines those routes. That turns the shipped routines UI into 404s even though the backing service is still running. Please move the old routine handlers into this registrar or add a dedicated `registerRoutineRoutes(...)` call with the same DB and same-origin wiring so list/create/run/update/delete keeps working.

apps/daemon/src/server.ts
```diff
@@ -2270,5790 +2204,1678 @@ export async function startServer({
...
+      && critiqueSkill !== undefined
+      && !isMediaSurface
+      && isPlainAdapter;
+    // Only thread the critique fields when the run is actually eligible;
+    // otherwise the composer's own internal eligibility check (cfg.enabled
+    // && brand && skill && !isMediaSurface) might still fire on
+    // non-plain adapters and we'd emit the panel for a run the orchestrator
+    // skips. Gating the threading itself keeps composer + orchestrator in
+    // exact lockstep regardless of which side enforces eligibility.
+    const prompt = composeSystemPrompt({
```
mrcfps, line 2657: `composeDaemonSystemPrompt()` still computes whether this run is using the plain adapter, but this refactor stopped passing `streamFormat` into `composeSystemPrompt()` here. In `apps/daemon/src/prompts/system.ts:252-258`, the plain/BYOK-only `Do not emit tool_calls` safeguard is appended only when `streamFormat === 'plain'`, so omitting the field drops that instruction for plain runs while leaving other adapters unchanged. Please thread `streamFormat` through this call again so API/plain-mode runs keep the same prompt contract they had before the route split.

mrcfps (changes requested):
@nettee Thanks for pushing this refactor through — the route split is much easier to navigate. I found a few blocking regressions on the current head where existing daemon endpoints or prompt behavior dropped during the extraction, so I left inline comments with the concrete restore points.


Location: apps/daemon/src/static-resource-routes.ts RIGHT line 33

This extraction now serves skills from `listSkills(SKILLS_DIR)` only, and the old `POST /api/skills/install` plus `DELETE /api/skills/:id` handlers are no longer registered anywhere on this head. That means installed skills under the user runtime directory disappear from `/api/skills`, direct `/api/skills/:id`/`/example`/`/assets/*` requests 404 for them, and the Library UI's install/uninstall calls in `apps/web/src/providers/registry.ts` now fail. Please restore the merged built-in+installed lookup (`listAllSkills()` or equivalent) for every skill route here and re-register the install/delete handlers with the same local-origin guard around `installFromTarget(...)` and `uninstallById(...)`.

Location: apps/daemon/src/static-resource-routes.ts RIGHT line 135

The design-system routes have the same regression: this registrar now reads only `DESIGN_SYSTEMS_DIR`, and the old `POST /api/design-systems/install` plus `DELETE /api/design-systems/:id` handlers were dropped during the move. On the current head that omits user-installed design systems from `/api/design-systems`, makes their detail/preview/showcase URLs 404, and breaks the Library install/uninstall flow that still calls those endpoints from `apps/web/src/providers/registry.ts`. Please switch these routes back to the merged built-in+installed lookup and restore the install/delete handlers with the existing local-daemon request checks so installed design systems remain selectable and manageable.

apps/daemon/src/route-context-contract.ts
nettee: Fixed in 0b6a855d. Added concrete route context dependency interfaces for http/paths/resources/routines, included routine deps in the registered route contract, and updated the checked route context shape.

nettee: Fixed in 0b6a855d. Added concrete route context dependency interfaces for http/paths/resources/routines, included routine deps in the registered route contract, and updated the checked route context shape.

apps/daemon/src/server.ts
```diff
@@ -6655,938 +3710,53 @@ export async function startServer({
...
+    validation: validationDeps,
+    finalize: finalizeDeps,
+    chat: { startChatRun },
+    agents: agentDeps,
+    critique: critiqueDeps,
+    lifecycle: { isDaemonShuttingDown: () => daemonShuttingDown },
+  });
+
+
+  registerChatRoutes(app, {
```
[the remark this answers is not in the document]
nettee: Fixed in 0b6a855d. Restored the routines HTTP API through registerRoutineRoutes and wired RoutineService run handling before service start.

apps/daemon/src/chat-routes.ts
nettee: Fixed in 0b6a855d. Restored the routines HTTP API through registerRoutineRoutes and wired RoutineService run handling before service start.

apps/daemon/src/chat-routes.ts
nettee: Fixed in 0b6a855d. Restored the routines HTTP API through registerRoutineRoutes and wired RoutineService run handling before service start.

apps/daemon/src/static-resource-routes.ts
```diff
@@ -0,0 +1,407 @@
...
+import { listPromptTemplates, readPromptTemplate } from './prompt-templates.js';
+import { readAppConfig } from './app-config.js';
+import type { RouteDeps } from './server-context.js';
+
+export interface RegisterStaticResourceRoutesDeps extends RouteDeps<'http' | 'paths' | 'resources'> {}
+
+export function registerStaticResourceRoutes(app: Express, ctx: RegisterStaticResourceRoutesDeps) {
+  const { sendApiError } = ctx.http;
+  const { RUNTIME_DATA_DIR, DESIGN_SYSTEMS_DIR, SKILLS_DIR, PROMPT_TEMPLATES_DIR, BUNDLED_PETS_DIR } = ctx.paths;
+  const { mimeFor } = ctx.resources;
```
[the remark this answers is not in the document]
nettee: Fixed in 0b6a855d. Restored skill/design-system install and uninstall endpoints in the static resource registrar.

apps/daemon/src/server.ts
nettee: Fixed in 0b6a855d. Threaded streamFormat back into composeSystemPrompt.

apps/daemon/src/server.ts
nettee: Fixed in 0b6a855d. Threaded streamFormat back into composeSystemPrompt.

apps/daemon/src/static-resource-routes.ts
```diff
@@ -0,0 +1,407 @@
...
+      const list = await detectAgents(config.agentCliEnv ?? {});
+      res.json({ agents: list });
+    } catch (err: any) {
+      res.status(500).json({ error: String(err) });
+    }
+  });
+
+  app.get('/api/skills', async (_req, res) => {
+    try {
+      const skills = await listSkills(SKILLS_DIR);
```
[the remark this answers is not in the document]
nettee: Fixed in 0b6a855d. Skill list/detail/example/asset routes now use the merged built-in + installed catalog, and install lookup fails visibly if the installed skill is absent from the refreshed catalog.

apps/daemon/src/static-resource-routes.ts
```diff
@@ -0,0 +1,407 @@
...
+      res.setHeader('Cache-Control', 'no-store');
+      res.sendFile(sheet.absPath);
+    } catch (err: any) {
+      res.status(500).type('text/plain').send(String(err));
+    }
+  });
+
+  app.get('/api/design-systems', async (_req, res) => {
+    try {
+      const systems = await listDesignSystems(DESIGN_SYSTEMS_DIR);
```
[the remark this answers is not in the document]
nettee: Fixed in 0b6a855d. Design-system list/detail/preview/showcase routes now preserve installed systems, and install lookup fails visibly if the installed system is absent from the refreshed catalog.

mrcfps (commented):
@nettee Thanks for restoring the dropped routes on this head — the split registrars are much easier to follow now. I found one follow-up issue in the extracted library-management handlers, so I left an inline note with the restore point.


Location: apps/daemon/src/static-resource-routes.ts RIGHT line 403

The extracted mutation block at `registerStaticResourceRoutes` now calls `installFromTarget(...)` and `uninstallById(...)` directly, but the old `server.ts` versions of these same handlers all gated those writes with `isLocalSameOrigin(req, resolvedPort)`. That changes `POST /api/skills/install`, `DELETE /api/skills/:id`, `POST /api/design-systems/install`, and `DELETE /api/design-systems/:id` from local-only actions into cross-origin-callable daemon endpoints, which reopens the CSRF boundary around installing or removing local library content. Please restore the same local-origin check here (or switch these four routes to `requireLocalDaemonRequest`) before the mutating call sites so the refactor preserves the original protection.

nettee:
Fixed in 7cbec203. Restored the local-origin gate before all four static resource mutation handlers (skill/design-system install and uninstall) and added daemon coverage that cross-origin requests return 403 before catalog/filesystem work.\n\nValidation:\n- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/static-resource-routes.test.ts\n- pnpm --filter @open-design/daemon typecheck\n- pnpm guard

mrcfps (approved):
@nettee I re-reviewed the current head and verified the extracted route registrars, the restored local-origin guards on the skill/design-system install and uninstall endpoints, and the added daemon coverage around those cross-origin rejects. I didn't find any additional correctness, safety, or maintainability issues in the changed ranges on this head.

Thanks for pushing through a large refactor here — the split route layout is much easier to follow. 🙏

PR #1044, opened 2026-05-09
feat(web): scroll question forms to top of viewport instead of pinning to bottom

paulstean:
## Summary

Changes the chat panel's auto-scroll behavior so that when an assistant response contains a `<question-form>`, the viewport scrolls to align the top of the form with the top of the visible chat area — instead of the current behavior which pins the latest response to the bottom.

### Key changes

- **`QuestionFormView`** (`apps/web/src/components/QuestionForm.tsx`): added `data-form-id={form.id}` attribute to the root element so it can be found by the scroll logic.
- **`ChatPane`** (`apps/web/src/components/ChatPane.tsx`):
  - Added `scrolledToFormRef` to track which form IDs have already been scrolled to, preventing re-scroll on subsequent streaming chunks.
  - **Initial scroll effect**: on conversation load, checks if the last assistant message contains a form and scrolls to it (instant) instead of bottom-scrolling.
  - **Auto-scroll effect**: during streaming, detects when a form appears in the last assistant message and smooth-scrolls to it. Sets `pinnedToBottomRef = false` immediately after so subsequent chunks don't abort the animation.
  - **Form submission**: resets `pinnedToBottomRef.current = true` and clears `scrolledToFormRef` so the next response reclaims scroll control.
  - Queries `[data-form-id]` only within the last `.msg.assistant` element to avoid targeting stale (answered) forms from earlier in the conversation.

### Behavior

| Scenario | Behavior |
|---|---|
| Normal text streaming | Bottom-scroll (unchanged) |
| Form appears during streaming | Smooth scroll to form top; subsequent text extends below viewport |
| Navigate to conversation ending in form | Instant scroll to form top |
| User scrolls away from form | No yank (`pinnedToBottomRef` is false) |
| Submit form → next response arrives | Scroll control reclaimed; next form/text scrolls into view |
| Tab switch away and back | Preserved via existing `savedChatScrollRef` mechanism |

### Verification

- `pnpm guard` — pass
- `pnpm typecheck` — all 12 workspace packages pass (1 pre-existing failure in `apps/packaged` unrelated)
- `pnpm --filter @open-design/web test` — 66 files, 596 tests, all pass

### DIFF
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -566,7 +566,11 @@ export function ChatPane({
             onEnsureProject={onEnsureProject}
             commentAttachments={commentsToAttachments(attachedComments)}
             onRemoveCommentAttachment={onDetachComment}
-            onSend={onSend}
+            onSend={(prompt, attachments, commentAttachments, meta) => {
+              pinnedToBottomRef.current = true;
+              scrolledToFormRef.current = new Set();
+              onSend(prompt, attachments, commentAttachments, meta);
+            }}
             onStop={onStop}
             onOpenSettings={onOpenSettings}
             onOpenMcpSettings={onOpenMcpSettings}
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -194,7 +194,9 @@ export function ChatPane({
         const formEl = lastAssistantEl?.querySelector<HTMLElement>('[data-form-id]');
         if (formEl && !scrolledToFormRef.current.has(formEl.dataset.formId!)) {
           scrolledToFormRef.current.add(formEl.dataset.formId!);
-          el.scrollTop = formEl.offsetTop;
+          const formRect = formEl.getBoundingClientRect();
+          const containerRect = el.getBoundingClientRect();
+          el.scrollTop += formRect.top - containerRect.top;
           pinnedToBottomRef.current = false;
           setScrolledFromBottom(true);
           return;
@@ -246,11 +248,11 @@ export function ChatPane({
         // Form tag in content but the DOM element isn't ready yet (partial
         // stream) — skip bottom-scroll to avoid a jarring jump that gets
         // undone when the form finishes rendering.
-        return;
+        if (streaming) return;
       }
       el.scrollTop = el.scrollHeight;
     }
-  }, [messages, error]);
+  }, [messages, error, streaming]);
 
   // Saved chat-log scroll state, preserved across tab switches. The
   // chat-log <div> is conditionally rendered so it unmounts when the
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -194,9 +194,7 @@ export function ChatPane({
         const formEl = lastAssistantEl?.querySelector<HTMLElement>('[data-form-id]');
         if (formEl && !scrolledToFormRef.current.has(formEl.dataset.formId!)) {
           scrolledToFormRef.current.add(formEl.dataset.formId!);
-          const formRect = formEl.getBoundingClientRect();
-          const containerRect = el.getBoundingClientRect();
-          el.scrollTop += formRect.top - containerRect.top;
+          formEl.scrollIntoView({ block: 'start', behavior: 'smooth' });
           pinnedToBottomRef.current = false;
           setScrolledFromBottom(true);
           return;
@@ -204,7 +202,7 @@ export function ChatPane({
         // Already handled by the auto-scroll effect — don't bottom-scroll.
         if (formEl) return;
       }
-      el.scrollTop = el.scrollHeight;
+      el.scrollTo({ top: el.scrollHeight, behavior: 'smooth' });
       setScrolledFromBottom(false);
       pinnedToBottomRef.current = true;
     });
@@ -250,7 +248,7 @@ export function ChatPane({
         // undone when the form finishes rendering.
         if (streaming) return;
       }
-      el.scrollTop = el.scrollHeight;
+      el.scrollTo({ top: el.scrollHeight, behavior: 'smooth' });
     }
   }, [messages, error, streaming]);
 
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -149,6 +149,7 @@ export function ChatPane({
   // 80px cutoff: scrolling ~90px up is an intentional pause that
   // shouldn't be yanked back the moment the next chunk streams in.
   const pinnedToBottomRef = useRef(true);
+  const scrolledToFormRef = useRef<Set<string>>(new Set());
   const [tab, setTab] = useState<Tab>('chat');
   const [showConvList, setShowConvList] = useState(false);
   const [scrolledFromBottom, setScrolledFromBottom] = useState(false);
@@ -176,6 +177,7 @@ export function ChatPane({
     // A new conversation should land at the bottom (its own initial
     // scroll), not inherit the previous conversation's saved position.
     savedChatScrollRef.current = null;
+    scrolledToFormRef.current = new Set();
   }, [activeConversationId]);
 
   useEffect(() => {
@@ -183,6 +185,23 @@ export function ChatPane({
     if (!el || didInitialScrollRef.current || messages.length === 0) return;
     didInitialScrollRef.current = true;
     requestAnimationFrame(() => {
+      // If the last assistant message contains a question form, scroll to
+      // the form instead of the bottom, so the user sees the form first.
+      const lastAssistantMsg = [...messages].reverse().find((m) => m.role === 'assistant');
+      if (lastAssistantMsg?.content.includes('<question-form')) {
+        const assistantEls = el.querySelectorAll('.msg.assistant');
+        const lastAssistantEl = assistantEls[assistantEls.length - 1];
+        const formEl = lastAssistantEl?.querySelector<HTMLElement>('[data-form-id]');
+        if (formEl && !scrolledToFormRef.current.has(formEl.dataset.formId!)) {
+          scrolledToFormRef.current.add(formEl.dataset.formId!);
+          el.scrollTop = formEl.offsetTop;
+          pinnedToBottomRef.current = false;
+          setScrolledFromBottom(true);
+          return;
+        }
+        // Already handled by the auto-scroll effect — don't bottom-scroll.
+        if (formEl) return;
+      }
       el.scrollTop = el.scrollHeight;
       setScrolledFromBottom(false);
       pinnedToBottomRef.current = true;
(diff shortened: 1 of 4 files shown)


paulstean:
Closes #983

paulstean:
Feedback addressed: the composer's `onSend` now also resets `pinnedToBottomRef` and `scrolledToFormRef`, so sending a message through the normal composer after a form has claimed scroll control will correctly auto-follow the new response.

apps/web/src/components/ChatPane.tsx
```diff
@@ -176,13 +177,31 @@ export function ChatPane({
...
+      // If the last assistant message contains a question form, scroll to
+      // the form instead of the bottom, so the user sees the form first.
+      const lastAssistantMsg = [...messages].reverse().find((m) => m.role === 'assistant');
+      if (lastAssistantMsg?.content.includes('<question-form')) {
+        const assistantEls = el.querySelectorAll('.msg.assistant');
+        const lastAssistantEl = assistantEls[assistantEls.length - 1];
+        const formEl = lastAssistantEl?.querySelector<HTMLElement>('[data-form-id]');
+        if (formEl && !scrolledToFormRef.current.has(formEl.dataset.formId!)) {
+          scrolledToFormRef.current.add(formEl.dataset.formId!);
+          el.scrollTop = formEl.offsetTop;
```
lefarcen: **P2** — `formEl.offsetTop` is not guaranteed to be relative to the `.chat-log` scroller, and the current CSS does not make `.chat-log` a positioned offset parent. On conversation load this can overscroll by the app/chat header offset and hide the top of the form, breaking the claimed "instant scroll to form top" behavior.

**Fix:** Use `scrollIntoView({ block: 'start', behavior: 'instant' })` (matching the streaming path at L241) or compute `formEl.getBoundingClientRect().top - el.getBoundingClientRect().top + el.scrollTop`.

apps/web/src/components/ChatPane.tsx
```diff
@@ -210,6 +229,25 @@ export function ChatPane({
...
+          scrolledToFormRef.current.add(formEl.dataset.formId!);
+          formEl.scrollIntoView({ block: 'start', behavior: 'smooth' });
+          pinnedToBottomRef.current = false;
+          setScrolledFromBottom(true);
+          return;
+        }
+        // Form tag in content but the DOM element isn't ready yet (partial
+        // stream) — skip bottom-scroll to avoid a jarring jump that gets
+        // undone when the form finishes rendering.
+        return;
```
lefarcen: **P2** — This returns for any last assistant content containing `<question-form` when no rendered `[data-form-id]` exists. If the model emits a malformed form, an unterminated form that never recovers, or literal protocol text in prose/code, bottom-follow stops and the jump state is not updated, so the "normal text streaming unchanged" invariant is not defensible.

**Fix:** Only suppress bottom-scroll while an active stream is plausibly waiting for a form to finish parsing; otherwise fall back to the existing bottom-scroll path. You could add a timeout (e.g., 2 seconds after last content change) or check if the message has stopped streaming.

lefarcen (changes requested):
Thanks for tackling issue #983, @paulstean! The overall approach is solid — tracking scrolled forms and coordinating with `pinnedToBottomRef` is the right architecture.

I've spotted two scroll edge cases that would ship visible regressions if not addressed. See inline comments below.

lefarcen (dismissed):
Thanks for addressing the feedback, @paulstean! Both P2 concerns are resolved:

✅ Line 197: Now uses `scrollIntoView({ block: 'start' })` instead of raw `offsetTop` — proper scroll alignment regardless of positioned ancestors.
✅ Line 249: Streaming effect also uses `scrollIntoView` before the early return, so malformed forms won't break bottom-scroll fallback.

The scroll behavior is now safe. Ship when ready.

apps/web/src/components/ChatPane.tsx
mrcfps: `el.scrollTop = formEl.offsetTop` uses `offsetTop` from the form’s offset parent, but `.chat-log` is just a flex child here (`apps/web/src/index.css:11265-11266`), not a positioned offset parent. On a conversation that already ends in a form, this can overshoot by the chat header/wrapper offset and leave the top of the form hidden, so the new "scroll form to the top of the viewport" behavior is not reliably met. Please switch this initial-load path to the same container-relative calculation as the streaming path (for example `formEl.scrollIntoView({ block: 'start', behavior: 'instant' })` or `formEl.getBoundingClientRect().top - el.getBoundingClientRect().top + el.scrollTop`) so the scroll target is anchored to `.chat-log`.

mrcfps (changes requested):
@paulstean Thanks for addressing the composer follow-up on #983 — I re-checked the latest head and there’s still one reproducible scroll regression in the changed range that should be fixed before this lands.


paulstean:
Both issues addressed in a192290:

1. **`offsetTop` → `getBoundingClientRect`**: Replaced `el.scrollTop = formEl.offsetTop` with `el.scrollTop += formRect.top - containerRect.top` using `getBoundingClientRect()`, so the scroll position is correctly relative to the `.chat-log` scroller regardless of positioned ancestors.

2. **Unconditional return → streaming-guarded**: The early `return` when `<question-form` is in content but no DOM element exists now only fires while `streaming` is true. If streaming has stopped and the form never rendered (malformed/unterminated), it falls through to the existing bottom-scroll path. Added `streaming` to the effect's dependency array so the guard is evaluated when streaming ends.

paulstean:
All scroll operations are now smooth per request:

- **Initial load bottom-scroll**: `el.scrollTo({ top: el.scrollHeight, behavior: 'smooth' })`
- **Initial load form-scroll**: `formEl.scrollIntoView({ block: 'start', behavior: 'smooth' })`
- **Streaming bottom-pinning**: `el.scrollTo({ top: el.scrollHeight, behavior: 'smooth' })`
- **Streaming form-scroll**: unchanged (already smooth)

Also added a `scrollTo` polyfill to the jsdom test files (not available in jsdom's HTMLElement).

paulstean:
The `offsetTop` concern raised by @mrcfps was against commit `98b185f` and has been addressed in subsequent commits (`a192290` → `c3ccaa0`). The initial-load path now uses `formEl.scrollIntoView({ block: 'start', behavior: 'smooth' })` — the same container-relative approach as the streaming path, matching both of the suggested fixes.

The review on commit `98b185f` is now stale against the current head (`c3ccaa0`). Would appreciate a re-review, thanks!

apps/web/src/components/ChatPane.tsx
```diff
@@ -210,9 +229,28 @@ export function ChatPane({
...
+          pinnedToBottomRef.current = false;
+          setScrolledFromBottom(true);
+          return;
+        }
+        // Form tag in content but the DOM element isn't ready yet (partial
+        // stream) — skip bottom-scroll to avoid a jarring jump that gets
+        // undone when the form finishes rendering.
+        if (streaming) return;
+      }
+      el.scrollTo({ top: el.scrollHeight, behavior: 'smooth' });
```
lefarcen: **P2** — Smooth bottom-scroll during streaming may break auto-follow. `el.scrollTo({ top: el.scrollHeight, behavior: 'smooth' })` is async and emits intermediate scroll positions. If the scroll event fires before reaching `el.scrollHeight`, it can flip `pinnedToBottomRef` to `false`, stopping subsequent chunks from auto-scrolling even though the user was originally pinned.

Consider keeping `behavior: 'instant'` for the streaming path (line 251) while using `smooth` only for the initial load path (line 208). Streaming bottom-pinning is a per-chunk micro-action; smooth animation there adds latency without UX benefit.

lefarcen (commented):
Thanks for addressing mrcfps's offsetTop concern, @paulstean! The `scrollIntoView` fix looks correct.

One follow-up finding from reviewing the current head:

apps/web/src/components/ChatPane.tsx
mrcfps: Switching the streaming bottom-follow path from a synchronous `scrollTop` write to `el.scrollTo({ behavior: 'smooth' })` can break the default auto-follow behavior for ordinary assistant responses. `scrollTo(..., smooth)` emits intermediate `scroll` events while the animation is still above the bottom, and this component's `onScroll` handler recomputes `pinnedToBottomRef.current = distance < 80` on every scroll (see `onScroll` in the same file). That means a user who was pinned at the bottom can be flipped to `false` by the animation itself before the next chunk arrives, so later chunks stop auto-scrolling even though the user never scrolled away. Because this changed line runs on every non-form streaming update, it regresses the main chat path, not just the new form case. Please keep the streaming bottom-pin path instantaneous again (for example `el.scrollTop = el.scrollHeight` or `el.scrollTo({ top: el.scrollHeight, behavior: 'instant' })`), or otherwise suppress the programmatic scroll events from updating the pinned state while the auto-follow animation is in progress.

mrcfps (changes requested):
@paulstean Thanks for iterating on this — I re-checked the latest head and found one remaining scroll regression in the changed range that should be fixed before merge.


paulstean:
Streaming bottom-pin reverted to instant `el.scrollTop = el.scrollHeight` (df5946d). The smooth `scrollTo` was emitting intermediate scroll events that flipped `pinnedToBottomRef` to `false`, breaking auto-follow for subsequent chunks — confirmed by both review findings.

Smooth scroll is retained for one-time transitions: initial load bottom-scroll, and form-scroll (`scrollIntoView({ behavior: 'smooth' })` in both effects).

apps/web/src/components/ChatPane.tsx
```diff
@@ -176,14 +177,32 @@ export function ChatPane({
...
+          scrolledToFormRef.current.add(formEl.dataset.formId!);
+          formEl.scrollIntoView({ block: 'start', behavior: 'smooth' });
+          pinnedToBottomRef.current = false;
+          setScrolledFromBottom(true);
+          return;
+        }
+        // Already handled by the auto-scroll effect — don't bottom-scroll.
+        if (formEl) return;
+      }
+      el.scrollTo({ top: el.scrollHeight, behavior: 'smooth' });
```
mrcfps: This initial-load bottom scroll is now animated, but the same file's `onScroll` handler recomputes `pinnedToBottomRef.current = distance < 80` on every scroll event. When a user opens or switches to a conversation while an assistant response is already streaming, the smooth animation can emit intermediate positions that flip `pinnedToBottomRef` to `false` before the viewport actually reaches the bottom. The next streamed chunk then skips the auto-follow path even though the user never scrolled, so the default text-streaming behavior regresses for active conversations. Please keep this initial bottom-pin instantaneous as well (for example `el.scrollTop = el.scrollHeight` / `behavior: 'instant'`), or otherwise suppress programmatic smooth-scroll events from updating the pinned state until the animation settles.

mrcfps (changes requested):
@paulstean Thanks for the quick iteration here — I re-checked the latest head and found one remaining auto-follow regression in the changed scroll path that should be fixed before merge.


paulstean:
Initial-load bottom-pin also reverted to instant `scrollTop = scrollHeight` (ae6d51e) — same scroll-event thrash issue flagged by mrcfps on the latest commit.\n\n**Where smooth scroll is retained:**\n- Form-scroll in Effect 2 (initial load): `formEl.scrollIntoView({ behavior: 'smooth' })` — one-time, no ongoing streaming\n- Form-scroll in Effect 3 (streaming): `formEl.scrollIntoView({ behavior: 'smooth' })` — `pinnedToBottomRef` is set to `false` immediately after, so scroll events don't matter\n\n**Where instant is required:**\n- Bottom-pin in Effect 2 (initial load) — `scrollTop = scrollHeight`\n- Bottom-pin in Effect 3 (streaming) — `scrollTop = scrollHeight`\n\nBoth bottom-pin paths run while streaming may be active; smooth `scrollTo` emits intermediate scroll events that flip `pinnedToBottomRef` to `false`, breaking auto-follow for subsequent chunks.

mrcfps (approved):
@paulstean I re-checked the latest head and verified the changed scroll paths now keep ordinary bottom-follow instant while smoothly aligning question forms at the top, and the composer/form submit resets reclaim scroll control correctly. Thanks for iterating quickly on the scroll edge cases here — this looks ready to ship.


leilei926524-tech:
Hi @paulstean — quick note from the Open Design team.

What you get there:

PR #1045, opened 2026-05-09
fix(web): keep entry footer pills compact

lefarcen:
## Summary
- place the pet action and Follow @nexudotio pill on the same entry-sidebar footer row
- keep the Local CLI pill and language menu as the remaining footer rows to avoid making the sidebar scroll just because of the new follow link
- constrain the follow label with a shrinkable ellipsis label so narrow sidebar widths do not stretch the footer

## Validation

Note: local Node is `v22.22.0`, so pnpm prints the existing engine warning for the repo's `~24` target.

### DIFF
--- a/apps/web/src/components/EntryView.tsx
+++ b/apps/web/src/components/EntryView.tsx
@@ -483,30 +483,43 @@ export function EntryView({
           loading={skillsLoading || designSystemsLoading}
         />
         <div className="entry-side-foot">
-          <button
-            type="button"
-            className={`foot-pill pet-pill${config.pet?.adopted ? '' : ' pet-pill-fresh'}`}
-            onClick={onAdoptPet}
-            title={
-              config.pet?.adopted
-                ? t('pet.changePet')
-                : t('pet.adoptCallout')
-            }
-          >
-            <span className="pet-pill-glyph" aria-hidden>
-              {config.pet?.adopted
-                ? config.pet.petId === 'custom'
-                  ? config.pet.custom.glyph || '🦄'
-                  : '🐾'
-                : '🐾'}
-            </span>
-            <span>
-              {config.pet?.adopted
-                ? t('pet.changePet')
-                : t('pet.adoptCallout')}
-            </span>
-            {!config.pet?.adopted ? <span className="pet-pill-dot" aria-hidden /> : null}
-          </button>
+          <div className="entry-side-foot-row">
+            <button
+              type="button"
+              className={`foot-pill pet-pill${config.pet?.adopted ? '' : ' pet-pill-fresh'}`}
+              onClick={onAdoptPet}
+              title={
+                config.pet?.adopted
+                  ? t('pet.changePet')
+                  : t('pet.adoptCallout')
+              }
+            >
+              <span className="pet-pill-glyph" aria-hidden>
+                {config.pet?.adopted
+                  ? config.pet.petId === 'custom'
+                    ? config.pet.custom.glyph || '🦄'
+                    : '🐾'
+                  : '🐾'}
+              </span>
+              <span>
+                {config.pet?.adopted
+                  ? t('pet.changePet')
+                  : t('pet.adoptCallout')}
+              </span>
+              {!config.pet?.adopted ? <span className="pet-pill-dot" aria-hidden /> : null}
+            </button>
+            <a
+              className="foot-pill foot-pill-follow"
+              href="https://x.com/nexudotio"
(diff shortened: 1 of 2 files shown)


apps/web/src/index.css
```diff
@@ -4482,6 +4482,12 @@ code {
   gap: 6px;
   padding-top: 16px;
 }
+.entry-side-foot-row {
+  display: flex;
+  align-items: center;
```
lefarcen, line 4488: P2 — Consider adding a responsive breakpoint for very narrow sidebars (<300px). Currently `.entry-side-foot-row` always displays as `flex` row, which could squish both pills illegibly on extreme narrowness. Options:

1. Add `@media (max-width: 320px)` to stack the row back to column
2. Or ensure parent `.entry-side-foot` has a `min-width` floor
3. Or accept current behavior (ellipsis on Follow text + Pet icon-only fallback)

Not a blocker — just something to consider for future polish.

lefarcen (commented):
Clean layout fix — groups social actions (Pet + Follow) on one row to avoid sidebar scroll. Code looks good.

One non-blocking suggestion below.

apps/web/src/index.css
```diff
@@ -4496,6 +4502,17 @@ code {
...
+.entry-side-foot-row .foot-pill {
+  min-width: 0;
+}
+.entry-side-foot-row .foot-pill-follow {
+  flex: 0 1 auto;
+}
+.entry-side-foot-row .foot-pill-follow span {
+  overflow: hidden;
+  text-overflow: ellipsis;
+  white-space: nowrap;
```
mrcfps: The new ellipsis rule is on the inner `<span>`, but that span is still a flex item with the default `min-width: auto`. In flex layouts that means the label keeps its min-content width, so at narrower sidebar sizes this row can still overflow instead of truncating — which is the regression this PR is trying to avoid. Please give the label its own class with `min-width: 0` (and ideally `flex: 1 1 auto`) or otherwise make the text node shrinkable so the ellipsis can actually engage.

mrcfps (commented):
@lefarcen Thanks for tightening up the entry footer layout — I found one follow-up in the new truncation CSS that’s worth addressing before relying on this row at narrower sidebar widths.


apps/web/src/index.css
lefarcen: Addressed in `479f87e8`: the Follow text now has its own `.foot-pill-follow-label` with `min-width: 0` and `flex: 1 1 auto`, and the footer container also has `min-width: 0`, so the ellipsis can engage instead of forcing overflow.

apps/web/src/index.css
lefarcen: Handled the shrink path in `479f87e8` by making the footer and Follow label explicitly shrinkable. The entry sidebar already has a 320px minimum width, so this keeps the row readable within the supported sidebar range without reintroducing the fourth footer row.

apps/web/src/components/EntryView.tsx
```diff
@@ -483,30 +483,43 @@ export function EntryView({
...
+                {config.pet?.adopted
+                  ? config.pet.petId === 'custom'
+                    ? config.pet.custom.glyph || '🦄'
+                    : '🐾'
+                  : '🐾'}
+              </span>
+              <span>
+                {config.pet?.adopted
+                  ? t('pet.changePet')
+                  : t('pet.adoptCallout')}
```
mrcfps, line 507: This new row now has to fit the localized pet label and the fixed `Follow @nexudotio` pill side-by-side, but only the follow label gets the new shrink/ellipsis treatment in `index.css`. The pet label here is still an unconstrained flex child, so several shipped translations that are already much longer than English (for example `Ein Haustier adoptieren`, `Adopter un compagnon`, and `Fogadj be egy háziállatot`) can become the non-shrinkable item at the 320px sidebar minimum and reintroduce the overflow/extra-height problem this change is trying to remove. Please either give this label the same `min-width: 0` + truncation treatment as the follow label, or add a small-width fallback that lets this row wrap/stack again when both pills cannot fit.

mrcfps (commented):
@lefarcen Thanks for tightening up the footer row and for addressing the earlier overflow case. I found one remaining locale-related shrink path in the new shared pill row.


mrcfps (approved):

lefarcen:
Follow-up for the localized pet-label truncation concern is here: #1150. #1045 was already merged, so I kept this as a tiny separate fix on top of latest `main`.

PR #1047, opened 2026-05-09
ci: cache Playwright browsers

mrcfps:
## Summary
- Cache Playwright browser downloads in CI using `~/.cache/ms-playwright`.
- Keep the browser install step so CI still verifies Chromium and installs system deps.

## Validation
- `pnpm guard`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -104,6 +104,7 @@ jobs:
           echo "version=$version" >> "$GITHUB_OUTPUT"
 
       - name: Restore Playwright browser cache
+        id: playwright_browser_cache
         uses: actions/cache@v5
         with:
           path: ~/.cache/ms-playwright
@@ -111,8 +112,12 @@ jobs:
           restore-keys: |
             playwright-${{ runner.os }}-
 
+      - name: Install Playwright system dependencies
+        run: pnpm -C e2e exec playwright install-deps chromium
+
       - name: Install Playwright browsers
-        run: pnpm -C e2e exec playwright install --with-deps chromium
+        if: ${{ steps.playwright_browser_cache.outputs.cache-hit != 'true' }}
+        run: pnpm -C e2e exec playwright install chromium
 
       # `scripts/postinstall.mjs` only prebuilds package/tool entrypoints that
       # are needed immediately after install for linked bins and shared
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -97,6 +97,14 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Restore Playwright browser cache
+        uses: actions/cache@v5
+        with:
+          path: ~/.cache/ms-playwright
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -97,6 +97,20 @@ jobs:
...
         run: pnpm install --frozen-lockfile
 
+      - name: Resolve Playwright version
+        id: playwright_version
+        run: |
+          version="$(pnpm -C e2e exec playwright --version | awk '{print $2}')"
+          echo "version=$version" >> "$GITHUB_OUTPUT"
+
+      - name: Restore Playwright browser cache
+        uses: actions/cache@v5
```
lefarcen, line 108: P2 actions/cache@v5 without continue-on-error: true — transient cache service failures will now fail the validate job before tests run. Since this is a performance optimization (not correctness), split into restore/save steps with continue-on-error: true, or add it here.

lefarcen (commented):
Hey @mrcfps, good cache addition for CI speed. Found 3 points worth considering:

**P2 — Cache failure can block CI**
actions/cache@v5 without continue-on-error means a transient GitHub cache service failure now fails the validate job before tests run. Since this is a performance optimization (not correctness), consider:
- Split into actions/cache/restore@v5 + actions/cache/save@v5 with continue-on-error: true (like existing tools-pack cache)
- Or add continue-on-error: true to the current restore step

**P3 — Missing architecture in cache key**
Key only scopes by OS + version. If this job runs on ARM Linux / mixed-arch runners, an x64 Chromium cache can restore for arm64 under the same key → runtime failures. Add ${{ runner.arch }} to key and restore-keys.

**P3 — Version parsing brittleness**
awk parsing of CLI output can break if playwright --version adds warnings/banners. Prefer reading package version directly:
```bash
version="$(node -p "require('@playwright/test/package.json').version")"
```
Fail explicitly if empty.

.github/workflows/ci.yml
```diff
@@ -97,6 +97,20 @@ jobs:
...
+        with:
+          path: ~/.cache/ms-playwright
+          key: playwright-${{ runner.os }}-${{ steps.playwright_version.outputs.version }}
```
lefarcen, line 111: P3 Cache key missing architecture. Add ${{ runner.arch }} to prevent x64 Chromium caching for arm64 runners:
key: playwright-${{ runner.os }}-${{ runner.arch }}-${{ steps.playwright_version.outputs.version }}

.github/workflows/ci.yml
```diff
@@ -97,6 +97,20 @@ jobs:
       - name: Install dependencies
         run: pnpm install --frozen-lockfile
 
+      - name: Resolve Playwright version
+        id: playwright_version
+        run: |
+          version="$(pnpm -C e2e exec playwright --version | awk '{print $2}')"
```
lefarcen, line 103: P3 Version extraction is brittle (awk parsing CLI output). If Playwright adds warnings/banners, this can break. Prefer:
version="$(node -p \"require('@playwright/test/package.json').version\")"
echo "version=$version" >> "$GITHUB_OUTPUT"

PR #1060, opened 2026-05-09
fix(web): render static previews for sketch json files

### LINKED ISSUE #126: Sketch files in Design Files cannot be previewed and show as broken thumbnails
Summary:
Sketch files in Design Files cannot be previewed correctly. When a sketch file is selected, the preview area shows a broken image or invalid thumbnail instead of a usable sketch preview.

Steps to reproduce:
1. Create or save a sketch file in the project.
2. Open the Design Files list.
3. Select the sketch file.
4. Observe the preview area on the right side.

Expected behavior:
Sketch files should render a valid preview when selected from Design Files.

Actual behavior:
The sketch preview does not render correctly and appears as a broken image or failed preview.

Why this is a problem:
- Users cannot quickly inspect saved sketch files from the Design Files list.
- The file appears broken even though it exists successfully in the project.
- This weakens trust in the sketch save and file preview workflow.
- It makes Design Files inconsistent with other previewable artifact types.

Expected fix direction:
- Treat sketch files as a previewable file type in the Design Files preview panel.
- Render a proper sketch thumbnail or inline sketch preview instead of a broken image placeholder.
- Ensure the preview path and MIME handling for sketch files are correct.

Priority:
Medium. The file is present, but the missing or broken preview hurts a core browsing and review workflow.

Image

Derrick-xn:
## Summary
- render editable `.sketch.json` files with a shared static SVG sketch preview instead of loading them as broken images
- reuse shared sketch parsing/model logic across Design Files, FileViewer, and SketchEditor/FileWorkspace
- add regression coverage for sketch previews in both Design Files and FileViewer

## Test Plan
- [x] `pnpm guard`
- [x] `pnpm --filter @open-design/web test -- tests/components/DesignFilesPanel.test.tsx`
- [x] `pnpm --filter @open-design/web test -- tests/components/FileViewer.test.tsx`

Closes #126


apps/web/src/components/sketch-model.ts
```diff
@@ -0,0 +1,139 @@
...
+      const padding = Math.max(1, clampSketchSize(item.size) / 2) + 16;
+      includePoint(clampSketchNumber(item.x1), clampSketchNumber(item.y1), padding);
+      includePoint(clampSketchNumber(item.x2), clampSketchNumber(item.y2), padding);
+      continue;
+    }
+    if (item.kind === 'text') {
+      const x = clampSketchNumber(item.x);
+      const y = clampSketchNumber(item.y);
+      const fontSize = Math.max(12, clampSketchSize(item.size));
+      const textWidth = Math.max(fontSize, item.text.length * fontSize * 0.62);
```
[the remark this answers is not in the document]
Derrick-xn: Verified this one. A malformed text item without a string `text` field reproduces the crash in `computeSketchBounds()` via `item.text.length`. I normalized sketch text in both the bounds calculation and SVG text render path, and added regression coverage in `apps/web/tests/components/sketch-model.test.ts`. The update is now pushed to this PR.

lefarcen:
Hey @Derrick-xn! 👋 Thanks for the quick turnaround on this.

Quick suggestion: add `Closes #126` to the PR description so GitHub auto-closes the issue when this merges. That keeps the issue-PR linkage clean.

Will run the full review shortly. 🚀

apps/web/src/components/SketchPreview.tsx
```diff
@@ -0,0 +1,180 @@
...
+  }, [file, projectId]);
+
+  const geometry = useMemo(() => {
+    const resolvedItems = items ?? [];
+    const { minX, minY, maxX, maxY } = computeSketchBounds(resolvedItems);
+    return {
+      items: resolvedItems,
+      viewBoxX: Math.min(0, minX - VIEWBOX_PADDING),
+      viewBoxY: Math.min(0, minY - VIEWBOX_PADDING),
+      viewBoxWidth: Math.max(DEFAULT_WIDTH, maxX - minX + VIEWBOX_PADDING * 2),
```
lefarcen: P2 The viewBox width/height are based only on item span, while viewBoxX/viewBoxY are clamped to include 0. A sketch drawn at normal canvas coordinates like x=500,y=300 with a small shape gets viewBox='0 0 320 200', so the preview appears empty/broken even though the sketch is valid. Compute width/height from the chosen origin, e.g. `max(DEFAULT_WIDTH, maxX + padding - viewBoxX)`, and add regression coverage for sketches whose bounds are entirely outside the default 320x200 area.

apps/web/src/components/sketch-model.ts
```diff
@@ -0,0 +1,144 @@
...
+}
+
+const MAX_ABS_COORDINATE = 100_000;
+const MAX_ITEM_SIZE = 4_096;
+
+export function parseSketchDocument(text: string | null): SketchItem[] {
+  if (!text) return [];
+  try {
+    const parsed = JSON.parse(text) as SketchDocument | { items?: SketchItem[] };
+    return Array.isArray(parsed.items) ? parsed.items : [];
```
lefarcen: P2 `parseSketchDocument` casts arbitrary JSON directly to `SketchItem[]`, but compute/render then assume required fields exist. A malformed file such as `{"items":[{"kind":"pen"}]}` will throw when `item.points` is iterated/read, crashing the preview instead of falling back gracefully. Validate/sanitize each item before returning it, or make `computeSketchBounds`/`renderSketchSvgItem` guard malformed/null items, and add tests for malformed pen/rect/arrow entries.

lefarcen (changes requested):
Hi @Derrick-xn! 👋

Thanks for tackling #126 — the refactor extracting shared sketch model logic is the right direction. However, the preview logic needs two robustness fixes before it's ready:

1. **Off-origin sketches render as blank** — viewBox calculation assumes sketch origin near (0,0), so a valid sketch drawn at x=500,y=300 would show empty preview
2. **Malformed JSON crashes rendering** — missing required fields (e.g., `{"kind":"pen"}` without `points`) will throw instead of falling back gracefully

See inline comments below for the specific lines + fixes. After addressing these, this will be solid. Thanks for the contribution!

apps/web/src/components/SketchPreview.tsx
```diff
@@ -0,0 +1,180 @@
...
+      viewBoxHeight: Math.max(DEFAULT_HEIGHT, maxY - minY + VIEWBOX_PADDING * 2),
```
mrcfps: `viewBoxX`/`viewBoxY` are clamped back to `0`, but `viewBoxWidth`/`viewBoxHeight` are still computed only from `max - min`. That means a valid sketch whose bounds live entirely in positive coordinates can still render as blank after this change: for example, a rect around `x=500,y=300` produces `viewBox="0 0 320 200"`, so the actual drawing sits completely outside the SVG viewport in both Design Files and File Viewer. Because this PR’s main goal is to make `.sketch.json` previews render reliably, this is still a merge-blocking correctness issue. Please derive width/height from the chosen origin (for example `maxX + VIEWBOX_PADDING - viewBoxX` and the same for Y) and add a regression test for an off-origin sketch that stays outside the default 320×200 frame.

apps/web/src/components/sketch-model.ts
mrcfps: `parseSketchDocument()` still returns arbitrary `items` entries without validating their shape, but the new preview code immediately dereferences required fields from those objects. A malformed user-edited file like `{"items":[{"kind":"pen"}]}` still reaches `computeSketchBounds()` and `renderSketchSvgItem()`, where `item.points` is iterated/read and the preview throws instead of falling back gracefully. Since this PR routes `.sketch.json` files through the new static preview path, malformed pen/rect/arrow entries are still a clear blocker on the main flow. Please either sanitize each parsed item before returning it, or add guards in both bounds/render paths for every shape kind, and cover malformed pen/rect/arrow fixtures in `apps/web/tests/components/sketch-model.test.ts`.

mrcfps (changes requested):
@Derrick-xn, thanks for the follow-up here. I re-checked the current head and the two preview robustness blockers called out earlier are still present in the changed ranges, so this still needs another pass before the static sketch preview path is safe to merge.


apps/web/src/components/SketchPreview.tsx
Derrick-xn: Fixed. `SketchPreview` now derives `viewBoxWidth`/`viewBoxHeight` from the chosen origin instead of only `max - min`, so sketches drawn entirely away from `(0,0)` stay inside the SVG viewport. I also added a regression test in `apps/web/tests/components/FileViewer.test.tsx` that covers a rect at `(500,300)` and verifies the expanded `viewBox`.

apps/web/src/components/sketch-model.ts
Derrick-xn: Fixed. `parseSketchDocument()` now normalizes each parsed item and drops malformed pen/rect/arrow entries before they reach bounds/render code. Recoverable text items are normalized in-place, and I added regression coverage in `apps/web/tests/components/sketch-model.test.ts` so malformed non-text shapes no longer crash the preview path.

mrcfps:
@Derrick-xn I'm holding off on generating review comments for nexu-io/open-design#1060 because this pull request has merge conflicts right now.

apps/web/src/components/sketch-model.ts
```diff
@@ -0,0 +1,258 @@
...
+const DEFAULT_SKETCH_COLOR = '#1c1b1a';
+const DEFAULT_SKETCH_SHAPE_SIZE = 2;
+const DEFAULT_SKETCH_TEXT_SIZE = 16;
+
+export function parseSketchDocument(text: string | null): SketchItem[] {
+  if (!text) return [];
+  try {
+    const parsed = JSON.parse(text) as SketchDocument | { items?: unknown };
+    if (!isSketchRecord(parsed) || !Array.isArray(parsed.items)) return [];
+    return parsed.items.flatMap((item) => {
```
mrcfps: `parseSketchDocument()` is now used by `FileWorkspace` to populate editable sketch tabs, and `saveSketch()` later writes `entry.items` back to disk. Because this branch filters `parsed.items` through `normalizeSketchItem()`, any unsupported or newer `kind` gets dropped as soon as the file is opened, then disappears permanently on the next save even if the user never touched that part of the drawing. That is a real data-loss/backward-compatibility regression for sketches produced by a newer client or hand-edited with extra item types. Please keep the preview hardening separate from the editor load path—for example, use a preview-only sanitizer here, or preserve unknown/raw items so `FileWorkspace` can round-trip them unchanged when saving.

mrcfps (changes requested):
@Derrick-xn, thanks for the follow-up here. I re-checked the latest head and found one remaining blocker in the shared sketch parser: the new normalization step now changes what the editable workspace loads, so saving a sketch can silently discard unsupported content. Please see the inline note before merging.


lefarcen (dismissed):
Both prior P2 findings are now resolved at head c3322d85:

- ✅ **viewBox calculation** (SketchPreview.tsx): width/height now derived from chosen origin instead of only max - min, so off-origin sketches stay inside the SVG viewport
- ✅ **malformed JSON handling** (sketch-model.ts): parseSketchDocument() now normalizes each parsed item and drops malformed pen/rect/arrow entries before they reach bounds/render code

Bot review is clean. Deferring final merge call to maintainer.

apps/web/src/components/sketch-model.ts
Derrick-xn: Fixed. I split the preview hardening path from the editable workspace round-trip path.

- `parseSketchDocument()` still returns sanitized known items for preview rendering.
- `FileWorkspace` now loads sketches through a workspace-specific parser that keeps the original `version` and raw `items` array alongside the editable known items.
- `saveSketch()` rebuilds the document from that raw array, replacing only the known items the editor understands and leaving unsupported/newer item kinds untouched and in place.

I added regressions for both layers:
- `apps/web/tests/components/FileWorkspace.test.tsx` covers opening an existing `.sketch.json` with an unsupported item and saving it unchanged.
- `apps/web/tests/components/sketch-model.test.ts` covers preserving raw unsupported items and the original version when rebuilding the workspace document.

Verification run:
- `pnpm guard`
- `pnpm typecheck`
- `pnpm --filter @open-design/web test -- tests/components/FileWorkspace.test.tsx tests/components/FileViewer.test.tsx tests/components/sketch-model.test.ts`

apps/web/src/components/sketch-model.ts
```diff
@@ -0,0 +1,322 @@
...
+
+  for (const rawItem of rawItems) {
+    if (normalizeSketchItem(rawItem)) {
+      if (nextKnownItem < items.length) {
+        mergedItems.push(items[nextKnownItem]!);
+      }
+      nextKnownItem += 1;
+      continue;
+    }
+    mergedItems.push(rawItem);
```
mrcfps, line 105: Because this merge loop always re-inserts every unsupported `rawItem`, destructive editor actions still cannot actually remove them. In the current workspace flow `SketchEditor.handleClear()` resets `items` to `[]` before save, but `buildSketchDocument()` will still emit any unknown entries from the original file via `mergedItems.push(rawItem)`. A sketch like `[text, ellipse]` therefore saves back as `[ellipse]` after **Clear**, so another client/newer parser will resurrect shapes the user thought they deleted. That is a real persistence bug in the changed save path, not just a missing test. Please preserve unknown items only for true no-op saves, or carry raw-index metadata so save can drop/update the right source entries when the user clears/deletes content, and add a `FileWorkspace` regression that clears a sketch with interleaved unsupported items.

mrcfps (changes requested):
@Derrick-xn, thanks for the follow-up — the preview/rendering fixes look good on this head, but I found one remaining save-path blocker in the sketch workspace round-trip before this is safe to merge.


apps/web/src/components/sketch-model.ts
Derrick-xn: Fixed. I verified the clear/save-path bug. `Clear` now routes through workspace state, marks that save as destructive, and `saveSketch()` drops preserved raw items for that save instead of round-tripping them back into the file. I also allow saving a dirty empty sketch so a cleared sketch can actually persist as empty. Added regressions in `apps/web/tests/components/FileWorkspace.test.tsx` and `apps/web/tests/components/sketch-model.test.ts`.

Verification:
- `pnpm --filter @open-design/web test -- tests/components/FileWorkspace.test.tsx tests/components/FileViewer.test.tsx tests/components/sketch-model.test.ts`

apps/web/src/components/SketchEditor.tsx
```diff
@@ -262,7 +229,7 @@ export function SketchEditor({
         <button
           className="primary"
           onClick={() => void onSave()}
-          disabled={saving || items.length === 0}
+          disabled={saving || (!dirty && items.length === 0)}
```
mrcfps: `SketchEditor` still gates both destructive actions on parsed `items`, so the new preservation path cannot be exercised for sketches whose raw document contains only unsupported/newer kinds. Concretely, `parseSketchWorkspaceDocument()` now loads `{ "items": [{"kind":"ellipse", ...}] }` as `rawItems=[...]` and `items=[]`, which makes this save button stay disabled on first open, while the adjacent Clear button is also disabled because it still checks `items.length === 0`. That leaves the user in a blank editor with no way to drop or rewrite the hidden raw content from this client. Please thread whether preserved `rawItems` exist into the editor (or check it in `FileWorkspace`) so Clear/Save remain available for unsupported-only sketches, and add a `FileWorkspace` regression for that case.

mrcfps (commented):
@Derrick-xn, thanks for the follow-up — the sketch preview/rendering fixes look solid on this head. I found one merge-safe follow-up in the new workspace round-trip flow for sketch files that only contain unsupported item kinds.


lefarcen (dismissed):
Hey @Derrick-xn! 👋 The clear/save-path fix looks solid — workspace now correctly drops unsupported raw items on destructive-clear instead of round-tripping them back. All prior P2 findings (off-origin viewBox, malformed JSON handling) were already addressed in earlier commits, and this final iteration resolved mrcfps's P1 data-loss concern.

LGTM from the bot side. 🎉

apps/web/src/components/SketchEditor.tsx
Derrick-xn: Fixed in `234c7491`. I threaded preserved-raw-item state into `SketchEditor` so unsupported-only sketches no longer get stuck in a blank, non-actionable editor. `Save` and `Clear` now stay available when the workspace still holds hidden raw items, while `Clear` continues to flip the save path into the destructive mode that drops those preserved entries.

I also added a `FileWorkspace` regression covering an unsupported-only sketch (`ellipse` only): first open keeps both actions enabled, and Clear + Save persists an empty document.

Verification:
- `pnpm --filter @open-design/web test -- tests/components/FileWorkspace.test.tsx tests/components/FileViewer.test.tsx tests/components/sketch-model.test.ts`

mrcfps (approved):
@Derrick-xn I re-checked the latest head’s sketch-preview path, the shared sketch parser/workspace round-trip, and the added regressions for unsupported-only and clear/save flows. The current changed ranges address the earlier preview and save-path issues cleanly, and I didn’t find any new correctness or maintainability problems in this revision.

Thanks for sticking with the follow-ups here — this looks ready to merge. 🎉


Derrick-xn:
Kindly merge the code @lefarcen

PR #1067, opened 2026-05-09
fix: improve Orbit and packaged data-dir startup errors

mrcfps:
## Summary
- fix the Orbit settings Strict Mode mounted-guard bug so `Run it now` re-enables after connected connectors load
- preflight packaged `dataRoot` writability before sidecar startup and show a user-facing Electron error dialog for actionable recovery
- expand daemon and packaged diagnostics with clearer causes plus targeted permission-repair guidance, and add regression tests for both flows

## Validation
- pnpm --filter @open-design/web test -- SettingsDialog.orbit.test.tsx
- pnpm --filter @open-design/packaged test
- pnpm --filter @open-design/packaged typecheck
- pnpm --filter @open-design/daemon exec tsc -p tsconfig.json --noEmit

### DIFF
--- a/apps/packaged/src/launch.ts
+++ b/apps/packaged/src/launch.ts
@@ -1,12 +1,102 @@
-import { mkdir } from "node:fs/promises";
+import { access, mkdir, stat } from "node:fs/promises";
+import { constants as fsConstants } from "node:fs";
+import { dirname } from "node:path";
+import { userInfo } from "node:os";
 
 import { app } from "electron";
 
 import type { PackagedNamespacePaths } from "./paths.js";
 
+export class PackagedPathAccessError extends Error {
+  readonly title: string;
+
+  constructor(message: string, options?: { cause?: unknown; title?: string }) {
+    super(message, options);
+    this.name = "PackagedPathAccessError";
+    this.title = options?.title ?? "Open Design cannot access its data folder";
+  }
+}
+
+type PathDiagnostic = {
+  exists: boolean;
+  mode?: number;
+  path: string;
+};
+
+function formatMode(mode: number | undefined): string {
+  if (mode == null) return "unknown";
+  return `0${(mode & 0o777).toString(8)}`;
+}
+
+async function inspectPath(path: string): Promise<PathDiagnostic> {
+  try {
+    const stats = await stat(path);
+    return { exists: true, mode: stats.mode, path };
+  } catch {
+    return { exists: false, path };
+  }
+}
+
+function formatWritablePathError(options: {
+  attemptedPath: string;
+  currentUser: string;
+  diagnostic: PathDiagnostic;
+  error: unknown;
+  parentDiagnostic: PathDiagnostic;
+}): string {
+  const { attemptedPath, currentUser, diagnostic, error, parentDiagnostic } = options;
+  const message = error instanceof Error ? error.message : String(error);
+  const parentPath = dirname(attemptedPath);
+  const diagLines = [
+    `Open Design could not create or write to:`,
+    attemptedPath,
+    "",
+    `Current user: ${currentUser}`,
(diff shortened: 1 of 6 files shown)


nettee (approved):

lefarcen (commented):
Nice fixes on Orbit + packaged startup — the user-facing error dialogs and terminal guidance are especially helpful for the common permissions trap.

A few non-blocking suggestions below to tighten edge-case correctness:

**Minor suggestions (P2/P3):**
- See inline comments for isMountedRef race, packaged POSIX dir-execute check, and cross-platform recovery guidance.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2545,8 +2545,15 @@ function OrbitSection({
   // Once the user clicks Generate we close Settings and navigate away. The ref
   // lets late-arriving handlers no-op without React warnings.
   const isMountedRef = useRef(true);
-  useEffect(() => () => {
-    isMountedRef.current = false;
+  useEffect(() => {
```
lefarcen, line 2548: **P2** — The Strict Mode fix resets `isMountedRef` to `true`, but that lets stale async work from an earlier effect pass update state after a newer request has already completed.

Scenario: slow first `fetchConnectors()` or `/api/orbit/status` response can overwrite a later focus-triggered refresh and re-disable `Run it now` even after connected connectors loaded.

Suggested fix: use a request-generation ref or per-effect cancellation token for `refreshConnectedCount` and `refreshStatus`, and only commit the latest request result. Add a regression test with deferred connector promises resolved out of order.

apps/packaged/src/launch.ts
```diff
@@ -1,12 +1,102 @@
...
+    `sudo chown -R \"${currentUser}\":staff \"${parentPath}\"`,
+    `chmod -R u+rwX \"${parentPath}\"`,
+  ];
+  return diagLines.join("\n");
+}
+
+export async function verifyPackagedDataRootWritable(paths: Pick<PackagedNamespacePaths, "dataRoot">): Promise<void> {
+  try {
+    await mkdir(paths.dataRoot, { recursive: true });
+    await access(paths.dataRoot, fsConstants.W_OK);
```
lefarcen, line 77: **P2** — The packaged preflight checks `access(dataRoot, W_OK)` only. On POSIX a directory can be writable but not searchable/executable, which still prevents creating files beneath it; the preflight can pass and the daemon then fails later with the less actionable sidecar-startup path.

Suggested fix: check `W_OK | X_OK` for directories or, more robustly, create and delete a temporary probe file inside `dataRoot`. Add a POSIX-only regression test for a directory without execute permission.

apps/packaged/src/launch.ts
```diff
@@ -1,12 +1,102 @@
...
+    `Target mode: ${formatMode(diagnostic.mode)}`,
+    `Parent exists: ${parentDiagnostic.exists ? "yes" : "no"}`,
+    `Parent mode: ${formatMode(parentDiagnostic.mode)}`,
+    "",
+    `Common causes:`,
+    `• the folder was created by another user (for example with sudo)`,
+    `• the parent folder is not writable`,
+    `• the folder is a symlink to a protected location`,
+    "",
+    `Try in Terminal:`,
```
lefarcen, line 66: **P3** — The recovery commands are macOS-specific but these code paths run on Linux and Windows packaged builds too. `sudo chown -R "<user>":staff` is wrong on Linux for many users and unusable on Windows, so the "actionable" dialog becomes misleading on supported platforms.

Suggested fix: format guidance by `process.platform`: macOS can use `:staff`, Linux should omit the group or use the user's primary group, and Windows should suggest `icacls`/Explorer permissions instead. Same applies to daemon/server.ts:790.

PR #1081, opened 2026-05-09
fix(desktop): allow about:blank popup for PDF export fallback

hitken:
## Problem

The PDF export fallback path uses `window.open(''_blank')` to open a blank window that is then navigated to a Blob URL. However, Electron's `setWindowOpenHandler` in `isAllowedChildWindowUrl` only allowed `blob:` and `od:` protocols, so the `about:blank` URL was silently denied.

This caused `window.open` to return `null`, triggering a "Popup blocked" alert for the user.

## Fix

Add `about:blank` to the allowed child window URL whitelist in `isAllowedChildWindowUrl`. The window is immediately navigated to a Blob URL after opening, so allowing `about:blank` is safe and scoped to the PDF export fallback flow.

## Affected File

- `apps/desktop/src/main/runtime.ts`

### DIFF
--- a/apps/desktop/src/main/runtime.ts
+++ b/apps/desktop/src/main/runtime.ts
@@ -228,7 +228,15 @@ export function isAllowedChildWindowUrl(url: string): boolean {
     // its links resolve to `http://127.0.0.1:.../...`, which is gated
     // by the separate `isHttpUrl` branch and continues to open in the
     // user's external browser via `shell.openExternal`.
-    return parsed.protocol === "blob:" || parsed.protocol === "od:";
+    // `about:blank` is used by the renderer's PDF export fallback path:
+    // `window.open('', '_blank')` opens a blank window that is then
+    // navigated to a Blob URL. Without this, the empty URL is denied
+    // and the user sees a "Popup blocked" alert.
+    return (
+      parsed.protocol === "blob:" ||
+      parsed.protocol === "od:" ||
+      (parsed.protocol === "about:" && parsed.pathname === "blank")
+    );
   } catch {
     return false;
   }


apps/desktop/src/main/runtime.ts
```diff
@@ -228,7 +228,15 @@ export function isAllowedChildWindowUrl(url: string): boolean {
...
     // user's external browser via `shell.openExternal`.
-    return parsed.protocol === "blob:" || parsed.protocol === "od:";
+    // `about:blank` is used by the renderer's PDF export fallback path:
+    // `window.open('', '_blank')` opens a blank window that is then
+    // navigated to a Blob URL. Without this, the empty URL is denied
+    // and the user sees a "Popup blocked" alert.
+    return (
+      parsed.protocol === "blob:" ||
+      parsed.protocol === "od:" ||
+      (parsed.protocol === "about:" && parsed.pathname === "blank")
```
mrcfps, line 238: `isAllowedChildWindowUrl` is exported specifically for the packaged allowlist regression tests, but this PR only changes the allowlist implementation and does not add a matching assertion in `apps/packaged/tests/desktop-url-allowlist.test.ts`. Because `setWindowOpenHandler` is part of the child-window security boundary, leaving the new `about:blank` branch untested makes it easy for a later cleanup to silently re-break the PDF export fallback. Please add a regression case here for `isAllowedChildWindowUrl('about:blank')` (and ideally keep a nearby negative case like `about:srcdoc` returning false) so this behavior is pinned the same way the existing `od:` and `blob:` entries are.

mrcfps (commented):
Thanks for the fix here — allowing `about:blank` in the desktop allowlist matches the renderer fallback and resolves the popup-blocked path. I found one follow-up that would make this regression much harder to reintroduce.


apps/desktop/src/main/runtime.ts
lefarcen, line 238: **P2 — Security boundary:** Allowing `about:blank` here permits any renderer code to open blank windows, not just the PDF fallback. After the initial `window.open`, the renderer can navigate the child window to any URL without re-checking the allowlist. Consider either:
1. Keeping the initial URL as the Blob URL (avoid two-step blank→navigate), OR
2. Adding child-window navigation enforcement for `about:blank` windows

lefarcen (commented):
Hi @hitken! Thanks for tackling the desktop PDF export popup issue.

I verified the renderer code path and the fix does address the immediate `window.open('', '_blank')` failure, but there's a **security boundary concern** worth discussing before merge:

**P2 — Allowlist scope**

The `setWindowOpenHandler` allowlist now permits any renderer code to open `about:blank` windows, not just the PDF fallback. After `setWindowOpenHandler` returns `{ action: "allow" }`, the renderer receives a child `BrowserWindow` reference and can navigate it to *any* URL without passing through `isAllowedChildWindowUrl` again.

The PDF path does `window.open('', '_blank')` → `win.location.href = blobUrl`, which is the intended use, but a compromised or future renderer path could now open `about:blank` and navigate to a URL that never passed the initial allowlist.

Two options to tighten this:
1. Keep the initial URL as the Blob URL (avoid the two-step `about:blank` → navigate pattern), OR
2. Add child-window navigation enforcement so `about:blank` windows can only navigate to `blob:` URLs

**P3 — Test coverage**

The existing allowlist test suite (`apps/packaged/tests/desktop-url-allowlist.test.ts`) wasn't updated. Consider adding test cases for:
- `about:blank` (allowed)
- `about:srcdoc` (rejected)
- `about:blank/` (behavior?)
- `about:blank?x=1` (behavior?)

This pins the intended scope and catches regressions.

---

**Overall:** The fix works for the immediate PDF export failure. The P2 is about whether we want the allowlist to be scoped to specific renderer flows or permit any `about:blank` window. If the current approach is acceptable for your security model, the P3 test coverage would still be valuable. Let me know your thoughts!

mrcfps (approved):

lefarcen:
Merged! 🎉 Thanks @hitken for the contribution to open-design. Looking forward to the next one!

leilei926524-tech:
Hi @hitken — welcome to Open Design contributors. 🎉

Your first merged PR (#1081) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #1085, opened 2026-05-09
feat(daemon): Critique Theater Phase 6.2 (artifact extraction + endpoint)

Nagendhra-web:
Step (b) of the four-step Phase 6.2 unblock plan that closed [#1006](https://github.com/nexu-io/open-design/pull/1006). Step (a) was the contracts move in [#1016](https://github.com/nexu-io/open-design/pull/1016). This PR is the foundational change that makes a real rerun endpoint possible: the orchestrator now actually persists the SHIP `<ARTIFACT>` body, and the web layer has a logical handle to fetch it.

## What lands

**Parser side-channel.** [`apps/daemon/src/critique/parser.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/src/critique/parser.ts) and [`parsers/v1.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/src/critique/parsers/v1.ts) capture the `mime` attribute and CDATA-stripped body of the SHIP `<ARTIFACT>` block, then call an `onArtifact` callback synchronously right before yielding the ship `PanelEvent`. The body and mime intentionally never join the `PanelEvent` itself, so they never broadcast on the SSE wire.

**Artifact writer.** New module [`apps/daemon/src/critique/artifact-writer.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/src/critique/artifact-writer.ts) turns a `(dir, body, mime)` triple into bytes on disk. Highlights:

- Mime allowlist (`text/html`, `text/css`, `text/markdown`, `text/plain`, `application/json`, `image/svg+xml`); everything else falls through to `application/octet-stream` + `.bin` so unknown payloads cannot be misinterpreted as a known type.
- UTF-8 byte-length cap (defaults to 1 MiB, the orchestrator passes `cfg.parserMaxBlockBytes`), so multi-byte payloads cannot sneak past a JS `.length` check. Caught by a dedicated test using emoji.
- Atomic write: render to a sibling `.tmp.<pid>.<ts>` file then rename, so a daemon crash mid-write cannot leave a half-written artifact under the canonical name. The tmp file is unlinked on rename failure.
- Typed `ArtifactTooLargeError` and `ArtifactEmptyError` so the orchestrator can route an oversize body separately from a filesystem error.

**Orchestrator wiring.** [`apps/daemon/src/critique/orchestrator.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/src/critique/orchestrator.ts) buffers the parser's `ShipArtifactPayload`, calls `writeShipArtifact` after `decideRound` runs, and pins the absolute path on the row through the existing finalize pipeline. Failures here log a warning and finalize with `artifactPath = null` rather than aborting the run, so the artifact endpoint returns 404 instead of pointing at a missing file. `OrchestratorResult.status` is now terminal-only (`CritiqueRunStatus`), no longer the wider `CritiquePersistedStatus`.

**Artifact endpoint.** New module [`apps/daemon/src/critique/artifact-handler.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/src/critique/artifact-handler.ts) serves `GET /api/projects/:projectId/critique/:runId/artifact`. The web layer never sees daemon paths; it streams bytes through this URL with the right `Content-Type`. Defense in depth:

- Cross-project leak guard returns `404` (not `403`) when a runId belongs to another project, mirroring the interrupt handler.
- Path-traversal guard resolves the row's `artifactPath` against the artifacts root and refuses anything outside it.
- Symlinks and non-regular files refused with `404` (defends against a tampered DB row pointing at `/etc/passwd`-style targets).
- Response size cap stops a tampered file from streaming unbounded bytes even if it slipped past the writer.
- `Content-Security-Policy: default-src 'none'; img-src data:; style-src 'unsafe-inline'` on HTML and SVG so the bytes are inert when fetched outside the web layer's sandbox iframe.
- `Cache-Control: private, max-age=3600` since artifacts are content-addressed by `runId` and never re-written.

## Folded in from PR #1016 review

[lefarcen left two non-blocking notes on #1016](https://github.com/nexu-io/open-design/pull/1016) (the contracts move). Both are closed here since `persistence.ts` was already in scope:

- **P2** introduced `CritiquePersistedStatus = CritiqueRunStatus | 'running'` in the contracts package. `CritiqueRunRow.status` and `CritiqueRunInsert.status` now use it, and the inline `as CritiqueRunStatus | 'running'` widen in [`interrupt-handler.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/src/critique/interrupt-handler.ts) is gone. Public DTOs continue to use the terminal-only `CritiqueRunStatus`.
- **P3** added `AssertExhaustiveValues` and a compile-time assertion in [`packages/contracts/src/critique.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/packages/contracts/src/critique.ts) that proves `CRITIQUE_RUN_STATUSES` covers every `CritiqueRunStatus` variant. Adding a value to `ShipStatus` without updating the array now fails the build with a tuple naming the missing variants.

## Coverage

174 critique tests pass locally across 14 files, including:

- New [`critique-artifact-writer.test.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/tests/critique-artifact-writer.test.ts) (13 unit tests) covering body/mime/extension contract, charset stripping, UTF-8 byte-length cap, atomic write, overwrite, no leftover tmp files.
- New [`critique-artifact-endpoint.test.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/tests/critique-artifact-endpoint.test.ts) (11 endpoint tests) covering happy path on html/svg/bin, 404 unknown / cross-project / no-artifact / outside-root / vanished-file / symlink / whitespace-runId, plus terminal-status fan-out.
- Extended [`critique-orchestrator.test.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/tests/critique-orchestrator.test.ts) happy-path: asserts `result.artifactPath` is set, the file exists, the bytes match, and the SSE ship payload does NOT carry `body` or `mime`.
- Inverted [`critique-lifecycle.test.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/tests/critique-lifecycle.test.ts) — the old "artifactPath stays null" regression is now "artifactPath is the absolute path and the bytes are the SHIP body with CDATA stripped".

## Validation

- `pnpm guard` clean.
- `pnpm --filter @open-design/contracts build` clean.
- `pnpm --filter @open-design/daemon build` clean (full tsc, the same step CI runs).
- `pnpm --filter @open-design/web typecheck` clean (sanity check that the contracts changes do not break web).
- `pnpm --filter @open-design/daemon exec vitest run tests/critique` (14 files, 174 tests, all green).

## What's next

Steps (c) and (d) of the unblock plan: persist `original_message_id` / `agent_id` / `model_id` so the rerun handler can resolve the exact original brief, then re-implement the rerun endpoint on top of (a)+(b)+(c) with end-to-end tests that go through `runOrchestrator`.

### DIFF
--- a/apps/daemon/src/critique/artifact-handler.ts
+++ b/apps/daemon/src/critique/artifact-handler.ts
@@ -1,0 +1,218 @@
+import { promises as fs, constants as fsConstants } from 'node:fs';
+import * as path from 'node:path';
+import type { Request, Response } from 'express';
+import type Database from 'better-sqlite3';
+import { getCritiqueRun } from './persistence.js';
+import { mimeForExtension } from './artifact-writer.js';
+
+/** HTTP status codes used by the artifact endpoint. */
+const HTTP_BAD_REQUEST = 400;
+const HTTP_NOT_FOUND = 404;
+const HTTP_OK = 200;
+
+/**
+ * Default response size cap when the caller doesn't pass one. Used as the
+ * floor: an instance with `OD_CRITIQUE_PARSER_MAX_BLOCK_BYTES` raised above
+ * this value gets the larger cap so every artifact the orchestrator + writer
+ * accepted is retrievable. Codex P2 on PR #1085: hard-coding 4 MiB here
+ * while the writer accepted up to `cfg.parserMaxBlockBytes` made accepted
+ * rows unretrievable in raised-limit deployments.
+ */
+const DEFAULT_RESPONSE_SIZE_CAP_BYTES = 4 * 1024 * 1024;
+
+/**
+ * GET /api/projects/:projectId/critique/:runId/artifact
+ *
+ * Streams the bytes the orchestrator wrote for the run's SHIP `<ARTIFACT>`
+ * block, with the mime type derived from the file extension on disk. This
+ * is the daemon-resolved logical handle the web layer fetches; clients
+ * never receive a raw filesystem path. Cross-project leak guard mirrors
+ * the interrupt handler: a request against project p1 must not surface a
+ * runId that belongs to p2.
+ *
+ * Status semantics:
+ *   200 — happy path; body is the artifact bytes with the right mime.
+ *   400 — projectId/runId missing or whitespace-only.
+ *   404 — run not found, cross-project leak, no artifact persisted, or
+ *         the on-disk file disappeared (bytes never made it to the row).
+ *
+ * @see specs/current/critique-theater.md § rerun endpoint (Task 6.2)
+ */
+export function handleCritiqueArtifact(
+  db: Database.Database,
+  options: { artifactsRoot: string; responseCapBytes?: number },
+): (req: Request, res: Response) => Promise<void> {
+  const artifactsRoot = path.resolve(options.artifactsRoot);
+  // Honor the configured cap as-is when it's a positive finite number.
+  // The earlier `Math.max(DEFAULT, configured)` shape made the read-time
+  // size guard diverge from the writer's policy: a deployment that lowered
+  // OD_CRITIQUE_PARSER_MAX_BLOCK_BYTES below 4 MiB would still stream
+  // tampered files up to 4 MiB even though writeShipArtifact would have
+  // refused them. Mirror the writer's policy exactly so what the writer
+  // accepts is what the endpoint serves, and only fall back to the
+  // default when the option is absent or invalid (mrcfps follow-up on
+  // PR #1085).
+  const responseCapBytes =
+    Number.isFinite(options.responseCapBytes) && options.responseCapBytes! > 0
+      ? options.responseCapBytes!
+      : DEFAULT_RESPONSE_SIZE_CAP_BYTES;
+
+  return async function critiqueArtifactHandler(req: Request, res: Response): Promise<void> {
+    const projectId =
+      typeof req.params['projectId'] === 'string'
+        ? req.params['projectId'].trim()
+        : '';
+    const runId =
+      typeof req.params['runId'] === 'string'
+        ? req.params['runId'].trim()
+        : '';
+
+    if (!projectId || !runId) {
+      res
+        .status(HTTP_BAD_REQUEST)
+        .json({ error: { code: 'BAD_REQUEST', message: 'projectId and runId are required' } });
+      return;
+    }
+
+    const row = getCritiqueRun(db, runId);
+
+    // Cross-project leak guard: a request against project p1 must not
+    // reveal that runId actually lives in p2. 404 (not 403) so the
+    // existence of other projects' runs is not leaked.
+    if (row === null || row.projectId !== projectId) {
+      res
+        .status(HTTP_NOT_FOUND)
+        .json({ error: { code: 'NOT_FOUND', message: 'critique run not found' } });
+      return;
+    }
+
+    if (row.artifactPath === null) {
+      // Either the run is still running, the orchestrator failed to write
+      // the artifact (size, fs error, agent shipped no body), or the row
+      // pre-dates artifact persistence and hasn't been backfilled. All
+      // three are observable as 404 from the client's perspective.
+      res
+        .status(HTTP_NOT_FOUND)
+        .json({
+          error: {
+            code: 'NOT_FOUND',
+            message: 'critique run has no persisted artifact',
+            currentStatus: row.status,
+          },
+        });
+      return;
+    }
+
+    // Path-traversal guard: the row's artifactPath must resolve INSIDE
+    // the configured artifacts root. Refusing anything else means a
+    // tampered DB row cannot leak arbitrary files off disk through this
+    // endpoint. Symlinks are not followed: the orchestrator writes
+    // through the writer module which never produces symlinks, so any
+    // symlink found here is suspicious.
+    const resolvedArtifactPath = path.resolve(row.artifactPath);
+    const relToRoot = path.relative(artifactsRoot, resolvedArtifactPath);
+    if (
+      relToRoot.startsWith('..')
+      || path.isAbsolute(relToRoot)
+      || relToRoot === ''
+    ) {
+      res
+        .status(HTTP_NOT_FOUND)
+        .json({ error: { code: 'NOT_FOUND', message: 'critique artifact not found' } });
+      return;
+    }
+
+    // Open the file once and stream from the resulting fd. The earlier
+    // shape did `lstat(path) → validate → createReadStream(path)`, which
+    // left a time-of-check / time-of-use window where a local actor with
+    // write access to the critique-artifacts directory could swap in a
+    // symlink or a larger file between the validation and the reopen.
+    // Opening once with O_NOFOLLOW (where the platform supports it) and
+    // stat'ing the open fd via fileHandle.stat() means the size and
+    // file-type checks describe the exact bytes we are about to stream
+    // (mrcfps follow-up on PR #1085).
+    //
+    // O_NOFOLLOW is a POSIX flag; on Windows the constant resolves to
+    // undefined and the open fall-through behaves like a normal read.
+    // The path-traversal guard above plus the orchestrator never writing
+    // through symlinks make Windows still safe in practice.
+    const noFollowFlag =
+      typeof fsConstants.O_NOFOLLOW === 'number' ? fsConstants.O_NOFOLLOW : 0;
+    const openFlags = fsConstants.O_RDONLY | noFollowFlag;
+    let fileHandle: Awaited<ReturnType<typeof fs.open>>;
+    try {
+      fileHandle = await fs.open(resolvedArtifactPath, openFlags);
+    } catch {
+      // ELOOP from O_NOFOLLOW (symlink), ENOENT (vanished), EACCES, etc.
+      res
+        .status(HTTP_NOT_FOUND)
+        .json({ error: { code: 'NOT_FOUND', message: 'critique artifact not found' } });
+      return;
+    }
+
+    try {
+      const stat = await fileHandle.stat();
+      if (!stat.isFile()) {
+        await fileHandle.close();
+        res
+          .status(HTTP_NOT_FOUND)
+          .json({ error: { code: 'NOT_FOUND', message: 'critique artifact not found' } });
+        return;
+      }
+      if (stat.size > responseCapBytes) {
+        await fileHandle.close();
+        res
+          .status(HTTP_NOT_FOUND)
+          .json({
+            error: { code: 'NOT_FOUND', message: 'critique artifact exceeds response cap' },
+          });
+        return;
+      }
+
+      const ext = path.extname(resolvedArtifactPath);
+      const contentType = mimeForExtension(ext);
+
+      res.status(HTTP_OK);
+      res.setHeader('Content-Type', contentType);
+      res.setHeader('Content-Length', String(stat.size));
+      res.setHeader('X-Content-Type-Options', 'nosniff');
+      // Prevent SVG / HTML artifacts from running scripts when fetched
+      // directly. The web layer renders the bytes inside a sandboxed iframe
+      // for HTML; this CSP is a defense-in-depth header for clients that
+      // dereference the URL outside that sandbox.
+      if (contentType === 'image/svg+xml' || contentType === 'text/html') {
+        res.setHeader(
+          'Content-Security-Policy',
+          "default-src 'none'; img-src data:; style-src 'unsafe-inline'",
+        );
+      }
+      // Artifacts are content-addressed by runId; the row never re-points
+      // to a different file once written, so a long cache lifetime is safe.
+      res.setHeader('Cache-Control', 'private, max-age=3600');
+
+      // createReadStream({ autoClose: true }) is the FileHandle method that
+      // closes the underlying fd when the stream ends or errors, so we do
+      // not double-close after handing the fd to the stream.
+      const stream = fileHandle.createReadStream({ autoClose: true });
+      stream.on('error', () => {
+        if (!res.headersSent) {
+          res.status(HTTP_NOT_FOUND).end();
+        } else {
+          res.destroy();
+        }
+      });
+      stream.pipe(res);
+    } catch (err) {
+      // Anything thrown after the open succeeded but before the stream is
+      // wired up: close the fd so it does not leak, then surface 404.
+      await fileHandle.close().catch(() => {});
+      if (!res.headersSent) {
+        res
+          .status(HTTP_NOT_FOUND)
+          .json({ error: { code: 'NOT_FOUND', message: 'critique artifact not found' } });
+      } else {
+        res.destroy(err instanceof Error ? err : undefined);
+      }
+    }
+  };
+}
--- a/apps/daemon/src/critique/orchestrator.ts
+++ b/apps/daemon/src/critique/orchestrator.ts
@@ -1,9 +1,15 @@
 import type { ChildProcess } from 'node:child_process';
+import { promises as fs } from 'node:fs';
 import type Database from 'better-sqlite3';
 import type { CritiqueConfig, PanelEvent } from '@open-design/contracts/critique';
 import { panelEventToSse } from '@open-design/contracts/critique';
 import type { CritiqueSseEvent } from '@open-design/contracts/critique';
-import { parseCritiqueStream } from './parser.js';
+import { parseCritiqueStream, type ShipArtifactPayload } from './parser.js';
+import {
+  ArtifactEmptyError,
+  ArtifactTooLargeError,
+  writeShipArtifact,
+} from './artifact-writer.js';
 import {
   computeComposite,
   decideRound,
@@ -14,6 +20,7 @@ import {
   insertCritiqueRun,
   updateCritiqueRun,
   type CritiqueRunRow,
+  type CritiqueRunStatus,
 } from './persistence.js';
 import { writeTranscript } from './transcript.js';
 import {
@@ -75,7 +82,10 @@ export interface OrchestratorParams {
 }
 
 export interface OrchestratorResult {
-  status: CritiqueRunRow['status'];
+  // The orchestrator only ever returns after a run reaches a terminal
+  // outcome, so this is the public terminal-only union, not the wider
+  // CritiquePersistedStatus that admits 'running' for live DB rows.
+  status: CritiqueRunStatus;
   composite: number | null;
   rounds: CritiqueRunRow['rounds'];
   transcriptPath: string | null;
@@ -131,7 +141,19 @@ export async function runOrchestrator(
   const completedRounds: RoundState[] = [];
   let artifactPath: string | null = null;
   let shipEvent: Extract<PanelEvent, { type: 'ship' }> | null = null;
-  let finalStatus: CritiqueRunRow['status'] = 'failed';
+  // Buffered SHIP artifact body coming from the parser side-channel. The
+  // body intentionally never travels on the SSE wire (the SHIP PanelEvent
+  // doesn't carry it), so the orchestrator captures it here, writes it to
+  // disk after decideRound runs, and pins artifactPath on the run row.
+  //
+  // Wrapped in a single-property box rather than a `let pendingArtifact`
+  // so TypeScript's control-flow analysis doesn't narrow it to `never`
+  // at the read site below. The parser callback only mutates the box's
+  // `value` field; without the box, tsc with strict / noImplicitAny would
+  // see the closure assignment as opaque and conclude the field is
+  // unreachable.
+  const artifactBuffer: { value: ShipArtifactPayload | null } = { value: null };
+  let finalStatus: CritiqueRunStatus = 'failed';
   let finalComposite: number | null = null;
   let transcriptPath: string | null = null;
 
@@ -180,6 +202,15 @@ export async function runOrchestrator(
       parserMaxBlockBytes: cfg.parserMaxBlockBytes,
       projectId,
       artifactId: params.artifactId,
+      // Side-channel for the SHIP <ARTIFACT> body. The parser invokes this
+      // synchronously right before yielding the ship PanelEvent, so the
+      // orchestrator has the body in hand by the time decideRound runs.
+      // Only the LAST artifact is retained: the parser already enforces
+      // that a second SHIP becomes a parser_warning, so this branch is only
+      // hit on the legitimate first SHIP per run.
+      onArtifact: (payload: ShipArtifactPayload) => {
+        artifactBuffer.value = payload;
+      },
     };
 
     for await (const event of parseCritiqueStream(timedSource, parserOpts)) {
@@ -321,9 +352,14 @@ export async function runOrchestrator(
       finalStatus = decision === 'ship' ? 'shipped' : 'below_threshold';
       finalComposite = shippedRound.composite;
 
-      // Emit the daemon-authoritative ship event. SSE clients and the
-      // transcript see this single normalized payload, never the raw agent
-      // claim from the buffered shipEvent.
+      // Build the daemon-authoritative ship event payload now, but DO NOT
+      // emit it yet. SSE clients react to critique.ship by fetching the
+      // artifact endpoint, so the file must be on disk AND the row's
+      // artifactPath must be populated before the event goes out;
+      // otherwise the client races us and gets a 404 against a row that
+      // is about to gain its path on the next finalize call. Order:
+      //   write file -> persist artifactPath on row -> emit ship event.
+      // (lefarcen P1 on PR #1085.)
       const normalizedShip: Extract<PanelEvent, { type: 'ship' }> = {
         type: 'ship',
         runId,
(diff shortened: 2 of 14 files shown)


apps/daemon/src/critique/orchestrator.ts
```diff
@@ -336,12 +360,57 @@ export async function runOrchestrator(
...
+      // run lifecycle, and the artifact endpoint will return 404 instead of
+      // pointing at a missing file. Phase 14 (artifact backfill) will fill
+      // in older rows that were created before this code path existed.
+      if (pendingArtifact !== null) {
+        try {
+          await fs.mkdir(artifactDir, { recursive: true });
+          const written = await writeShipArtifact(
+            artifactDir,
+            pendingArtifact.body,
+            pendingArtifact.mime,
```
mrcfps: `pnpm --filter @open-design/daemon build` is red on this block in CI: `TS2339: Property 'body' does not exist on type 'never'` and the same for `mime`. The new `pendingArtifact` value is only assigned from the parser callback, and TypeScript does not treat that closure write as a visible narrowing source here, so by the time we read `pendingArtifact.body` / `.mime` the branch is considered impossible. That makes the workspace check fail before we even get to runtime coverage. Please change this handoff so the compiler can see the nullable state directly — for example by storing the callback result in a `{ current: ShipArtifactPayload | null }` holder (or another explicit mutable container) and reading `current` here — then rerun the daemon build.

apps/daemon/src/critique/orchestrator.ts
```diff
@@ -336,12 +360,57 @@ export async function runOrchestrator(
...
+            { maxBytes: cfg.parserMaxBlockBytes },
+          );
+          artifactPath = written.absPath;
```
mrcfps, line 389: This now persists `artifact.<ext>` into the same per-run directory under `ARTIFACTS_DIR` that the daemon still serves wholesale via `app.use('/artifacts', express.static(ARTIFACTS_DIR))` (`apps/daemon/src/server.ts:3759`, with the critique run directory rooted at `apps/daemon/src/server.ts:5666`). Because of that, the new bytes are still fetchable at `/artifacts/<projectId>/<runId>/artifact.html`, which bypasses the new API route's project check, response-size cap, symlink guard, and HTML/SVG CSP. That is a real isolation hole for critique artifacts, not just a duplicate access path. Please either move critique artifacts outside the static tree or explicitly block that subtree from the `/artifacts` middleware and force all access through `/api/projects/:projectId/critique/:runId/artifact`, and add a regression test that the direct static URL no longer serves these files.

apps/daemon/src/critique/parsers/v1.ts
```diff
@@ -64,6 +68,7 @@ export async function* parseV1(
     shipSeen: false,
     designerArtifactInRound1: false,
     lastAdvance: 0,
+    onArtifact: opts.onArtifact,
```
mrcfps, line 73: This assignment is also breaking the daemon build under `exactOptionalPropertyTypes`: CI is failing with `TS2375` because `State.onArtifact` is declared as an optional property, but the object literal always materializes the key and sometimes gives it `undefined`. With that compiler mode, `onArtifact?: ShipArtifactCallback` is *not* assignable from `ShipArtifactCallback | undefined` once the property is present. Please either make the state field explicit as `ShipArtifactCallback | undefined`, or only add `onArtifact` to the object when `opts.onArtifact` is defined, then rerun the build so the validation job goes green.

mrcfps (changes requested):
@Nagendhra-web Thanks for pushing the Phase 6.2 artifact extraction work through — the parser/writer/endpoint split is headed in a good direction, but I hit three blocking issues on the current head: two new TypeScript errors are failing the daemon build in CI, and the persisted critique artifacts are still directly reachable through the existing `/artifacts` static mount.


lefarcen (changes requested):
Hi @Nagendhra-web! 👋

Thanks for the detailed PR — the artifact persistence + endpoint design is thorough, test coverage is impressive (174 tests!), and the security defenses (CSP, path-traversal guard, symlink rejection) show careful thought.

**Found 2 critical issues that need fixing before merge:**

### P1 — Artifact directory exposed through existing static route

**Location:** `apps/daemon/src/server.ts:3759` + `apps/daemon/src/server.ts:5666`

The critique artifacts are stored under `ARTIFACTS_DIR`, which is already exposed by `app.use('/artifacts', express.static(ARTIFACTS_DIR))` at line 3759. That means a shipped `artifact.html` becomes directly fetchable as `/artifacts/<projectId>/<runId>/artifact.html`, bypassing:
- your new endpoint's cross-project guard
- `nosniff` header
- CSP
- size cap
- symlink/non-regular-file checks

**Fix:** Move critique artifacts outside the public static root (e.g., `critiqueMeta/artifacts/`), OR explicitly deny the critique subtree from `express.static` and serve only through `handleCritiqueArtifact`.

### P1 — Race: ship event before artifact persisted

**Location:** `apps/daemon/src/critique/orchestrator.ts:360`

`critique.ship` is emitted before the artifact is written and before `critique_runs.artifact_path` is persisted. A client reacting to the ship event can immediately fetch the logical artifact endpoint and get `404` because the DB row still has `artifactPath = null`.

**Fix:** Write the artifact and persist the row (call `updateCritiqueRun`) **before** emitting the final `critique.ship` event, or delay the `critique.ship` emission until after the DB update completes.

---

**3 P2 non-blocking suggestions:**

### P2 — TOCTOU: path validated then opened later

**Location:** `apps/daemon/src/critique/artifact-handler.ts:110`

The handler checks `lstat()` and later opens the path with `createReadStream()`, leaving a window where a regular file can be swapped for a symlink or a larger file after validation.

**Suggestion:** Open the file once with no-follow semantics where available, then `fstat()` the fd and stream from that fd; at minimum revalidate after open and avoid path-based reopen after checks.

### P2 — Size cap mismatch with writer config

**Location:** `apps/daemon/src/critique/artifact-handler.ts:19`

The response cap is hard-coded to `4 MiB`, while the writer uses `cfg.parserMaxBlockBytes` (default 256KB). If operators raise `OD_CRITIQUE_PARSER_MAX_BLOCK_BYTES` above 4 MiB, valid persisted artifacts become unservable; if they lower it, tampered files up to 4 MiB still pass endpoint defense-in-depth.

**Fix:** Pass the configured cap into `handleCritiqueArtifact` or derive the endpoint cap from the same source as the writer.

### P2 — CDATA regex stops at first `</ARTIFACT>`

**Location:** `apps/daemon/src/critique/parsers/v1.ts:372`

The SHIP artifact extraction uses a non-CDATA-aware regex that stops at the first literal `</ARTIFACT>` inside the body. Valid HTML/JS can contain that sentinel in a string/comment, causing truncation or malformed extraction.

**Suggestion:** Parse the `<ARTIFACT>` opener explicitly and, for CDATA bodies, scan for `]]></ARTIFACT>` rather than the first closing tag.

---

Once these are addressed, this will be solid! Looking forward to the next iteration.

apps/daemon/src/critique/artifact-handler.ts
```diff
@@ -0,0 +1,164 @@
...
+const HTTP_NOT_FOUND = 404;
+const HTTP_OK = 200;
+
+/**
+ * Hard cap on response size. Mirrors the parser's per-block cap so a
+ * shipped artifact larger than the agent could have produced never gets
+ * served. The on-disk file would already have been refused by the writer,
+ * so this is a defense-in-depth check against a tampered file.
+ */
+const RESPONSE_SIZE_CAP_BYTES = 4 * 1024 * 1024;
```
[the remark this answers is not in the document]
Nagendhra-web: Good point. Fixed in 076f183 by threading `cfg.parserMaxBlockBytes` from the daemon config into `handleCritiqueArtifact` as a `responseCapBytes` option. The handler now uses `max(default 4 MiB, configuredCap)` so accepted artifacts stay retrievable in deployments that raise `OD_CRITIQUE_PARSER_MAX_BLOCK_BYTES`, and never falls below the 4 MiB floor when the override is absent or invalid.

lefarcen (changes requested):
Thanks for the quick fixes @Nagendhra-web! 🙌

**mrcfps's 3 blocking issues are now resolved:**
- ✅ TypeScript `pendingArtifact` narrowing: fixed via `artifactBuffer.value` box
- ✅ Static route bypass: critique artifacts moved to separate `CRITIQUE_ARTIFACTS_DIR`
- ✅ `exactOptionalPropertyTypes` error: `onArtifact` now explicitly `ShipArtifactCallback | undefined`

**My prior P2s are also addressed:**
- ✅ Size cap mismatch: `responseCapBytes` now threaded from `cfg.parserMaxBlockBytes`
- ✅ CDATA regex: `stripCdataWrapper` properly handles the wrapper

**One P1 remains unresolved:**

### P1 — Race: `critique.ship` emitted before artifact persisted

**Location:** `apps/daemon/src/critique/orchestrator.ts:367-423`

The current head still emits `critique.ship` at line 367 **before** the artifact write/persist block (lines 374-423). A client reacting to the ship event can immediately fetch `/api/projects/:projectId/critique/:runId/artifact` and get `404` because:
1. The DB row's `artifactPath` is still `null` (not updated until after the ship event)
2. The file doesn't exist yet (write happens after the event)

**Fix:** Move the ship event emission to **after** line 423 (after `artifactPath` is set and the DB row is persisted with `updateCritiqueRun`). The transcript write can stay where it is (after ship emission) — transcripts are internal, not race-sensitive like the artifact endpoint.

**Minimal diff:**
```diff
-      collectedEvents.push(normalizedShip);
-      bus.emit(panelEventToSse(normalizedShip));
-
       // Persist the SHIP artifact body now...
       const captured = artifactBuffer.value;
       if (captured !== null) {
         try {
           await fs.mkdir(artifactDir, { recursive: true });
           const written = await writeShipArtifact(...);
           artifactPath = written.absPath;
         } catch (err) { ... }
       }
+
+      // Emit ship AFTER artifact is persisted so clients can fetch it.
+      collectedEvents.push(normalizedShip);
+      bus.emit(panelEventToSse(normalizedShip));
```

Once this is fixed, the PR will be solid. Thanks again for the thorough work!

apps/daemon/src/critique/orchestrator.ts
Nagendhra-web: Right, this was a real bypass and I shipped a partial fix on top. In 076f183 critique artifacts moved out of `ARTIFACTS_DIR` entirely into a separate `<RUNTIME_DATA_DIR>/critique-artifacts/` root so the `/artifacts` static mount cannot reach them. The per-run endpoint is now the only sanctioned read path and its project / size / CSP guards cannot be skipped. The orchestrator spawn path and the handler factory both point at the new root.

apps/daemon/src/critique/parsers/v1.ts
Nagendhra-web: Fixed in 076f183. `State.onArtifact` is now declared as `ShipArtifactCallback | undefined` (no `?:`) so the object literal can explicitly carry `undefined` under `exactOptionalPropertyTypes: true`. Daemon build is clean locally and CI confirmed green before this latest push.

lefarcen (dismissed):
Perfect! 🎉

**All blockers resolved:**

✅ **mrcfps's 3 TypeScript/security issues** (fixed in 076f183):
- TypeScript `pendingArtifact` narrowing → resolved via `artifactBuffer.value` box
- Static route bypass → critique artifacts moved to separate `CRITIQUE_ARTIFACTS_DIR`
- `exactOptionalPropertyTypes` violation → `onArtifact` now explicit `| undefined`

✅ **Bot's P1 race condition** (fixed in 93c8100):
- Ship event now emitted **after** artifact write + DB persist
- Orchestrator flow: `writeShipArtifact` → `updateCritiqueRun({ artifactPath })` → `bus.emit(critique.ship)`
- Added regression test confirming `artifactPath` is non-null and file exists at ship-event moment

✅ **Bot's prior P2s** (fixed in 076f183):
- Size cap mismatch resolved (handler now uses `cfg.parserMaxBlockBytes`)
- CDATA wrapper handled correctly via `stripCdataWrapper`

The PR is solid. Comprehensive test coverage (175 tests including new race-condition regression case), clean builds, and all security defenses in place. Great work iterating through the feedback! 🚀

apps/daemon/src/critique/parsers/v1.ts
```diff
@@ -357,14 +364,35 @@ function* drain(state: State): Generator<PanelEvent> {
...
-      if (!artifactMatch || artifactMatch[1] === undefined || artifactMatch[1].trim().length === 0) {
+      // Validate that a non-empty <ARTIFACT> block is present inside <SHIP>
+      // and capture its head + body so the orchestrator can persist the bytes
+      // to disk via the side-channel callback. The body must NOT be added to
+      // the ship PanelEvent itself, since that event is also the SSE wire
+      // shape; broadcasting megabytes of HTML to every SSE subscriber would
+      // ruin the bus. The orchestrator pulls the body off `onArtifact` and
+      // emits only the small `artifactRef` on the wire.
+      const artifactMatch = inner.match(
+        /<ARTIFACT\b([^>]*)>([\s\S]*?)<\/ARTIFACT>/,
```
mrcfps: The new extractor still uses `/<ARTIFACT\b([^>]*)>([\s\S]*?)<\/ARTIFACT>/`, so it stops at the first literal `</ARTIFACT>` even when that sequence is inside a CDATA-wrapped payload. On the current head, an otherwise valid artifact like `<![CDATA[<script>const s = "</ARTIFACT>";</script>]]>` leaves `artifactMatch[2]` truncated to `<![CDATA[<script>const s = "`, so the writer persists a corrupted file even though the SHIP body itself is valid. That matters for this feature because the rerun/artifact path is supposed to round-trip arbitrary shipped HTML/SVG bytes, and embedded sentinel text inside JS or comments is a realistic pattern. Please parse the `<ARTIFACT>` opener separately and, when the body starts with `<![CDATA[`, scan for the terminating `]]></ARTIFACT>` pair instead of the first bare closing tag, then add a regression test that includes `</ARTIFACT>` inside the artifact body.

mrcfps (commented):
@Nagendhra-web Thanks for the follow-up pass here — the artifact isolation and ship-ordering fixes look good. I found one parser edge case in the new `<ARTIFACT>` extraction path that is still worth tightening before rerun consumers start depending on arbitrary shipped HTML and JS bodies.


apps/daemon/src/critique/parsers/v1.ts
Nagendhra-web: Good find. Fixed in 3915a9a (also rebased onto current main, which had a small conflict in apps/daemon/src/server.ts I resolved). The single-regex extraction is gone; replaced with a CDATA-aware `extractArtifactBlock` helper that:

1. Matches the `<ARTIFACT ...>` opener separately to capture attrs.
2. Inspects the body: if it starts with `<![CDATA[`, scans for `]]>\s*</ARTIFACT>` rather than the bare `</ARTIFACT>`. CDATA forbids the literal `]]>` inside its content per the XML spec, so the first match is always the real terminator. Whitespace between `]]>` and `</ARTIFACT>` is allowed so authors can pretty-print.
3. Falls back to the bare `</ARTIFACT>` for non-CDATA inline bodies (the contract: a body that needs to embed the literal sentinel must wrap itself in CDATA).

Added 4 regression tests in [`apps/daemon/tests/parser.test.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/tests/parser.test.ts):

- CDATA body containing `</ARTIFACT>` inside a JS string literal.
- CDATA body containing `</ARTIFACT>` inside an HTML comment.
- Non-CDATA inline body still extracts correctly.
- Whitespace tolerance between `]]>` and `</ARTIFACT>`.

Local validation: 15/15 test files, 201/201 tests pass (`pnpm --filter @open-design/daemon exec vitest run tests/parser.test.ts tests/critique`); `pnpm guard`, `pnpm --filter @open-design/daemon build`, `pnpm --filter @open-design/web typecheck` all clean.

apps/daemon/src/critique/artifact-handler.ts
```diff
@@ -0,0 +1,176 @@
...
+  const artifactsRoot = path.resolve(options.artifactsRoot);
+  // Take the configured cap if it was provided, otherwise the default.
+  // Always at least the default so an accidentally-low override still
+  // serves typical artifacts; never below that floor so a tampered cap
+  // does not break retrieval.
+  const responseCapBytes = Math.max(
+    DEFAULT_RESPONSE_SIZE_CAP_BYTES,
+    Number.isFinite(options.responseCapBytes) && options.responseCapBytes! > 0
+      ? options.responseCapBytes!
+      : DEFAULT_RESPONSE_SIZE_CAP_BYTES,
```
mrcfps: Because `responseCapBytes` is forced through `Math.max(DEFAULT_RESPONSE_SIZE_CAP_BYTES, ...)`, threading `critiqueCfg.parserMaxBlockBytes` into this handler only takes effect when the configured cap is **above** 4 MiB. On the default or lowered configs, the endpoint will still stream any tampered file up to 4 MiB even though `writeShipArtifact()` would have refused it, so the read-time size guard no longer matches the write policy this block is trying to enforce. Please honor any explicit positive `responseCapBytes` as-is and only fall back to `DEFAULT_RESPONSE_SIZE_CAP_BYTES` when the option is missing or invalid.

apps/daemon/src/critique/artifact-handler.ts
```diff
@@ -0,0 +1,176 @@
...
+      res.setHeader(
+        'Content-Security-Policy',
+        "default-src 'none'; img-src data:; style-src 'unsafe-inline'",
+      );
+    }
+    // Artifacts are content-addressed by runId; the row never re-points to
+    // a different file once written, so a long cache lifetime is safe.
+    res.setHeader('Cache-Control', 'private, max-age=3600');
+
+    const stream = createReadStream(resolvedArtifactPath);
```
mrcfps: There is still a time-of-check/time-of-use gap across this range: the handler `lstat()`s `resolvedArtifactPath`, validates `stat.isFile()` / `stat.size`, and then reopens the path later with `createReadStream()`. A local actor who can mutate the critique-artifacts directory between those syscalls can swap in a symlink or a larger file after validation, which bypasses the new non-regular-file and size-cap guards this endpoint is meant to provide. Please open the file first (with no-follow semantics where the platform supports it), `fstat()` that handle, and stream from the already-open fd instead of re-resolving the path after validation.

mrcfps (commented):
@Nagendhra-web Thanks for the follow-up pass here — the artifact extraction flow looks much tighter now. I found two small endpoint-hardening gaps in the new `/artifact` handler that seem worth closing before other callers start depending on these guards as the full boundary.


lefarcen (commented):
Great fix on the CDATA edge case! ✅

**mrcfps's finding: resolved**
- The new `extractArtifactBlock` helper correctly scans for `]]></ARTIFACT>` when the body starts with `<![CDATA[`
- Tests cover the failure case (JS string + HTML comment containing literal `</ARTIFACT>`)

**2 P2 suggestions (non-blocking, related edge cases):**

The CDATA-aware fix at the `<ARTIFACT>` level is solid, but there are similar sentinel bugs at sibling tag levels that can still cause truncation when artifacts contain arbitrary HTML/JS:

1. **P2 — `</SHIP>` sentinel in CDATA**: `parsers/v1.ts:332` uses `slice.indexOf('</SHIP>')` before the artifact extraction runs. A CDATA artifact containing literal `</SHIP>` in JS or an HTML comment will truncate the whole SHIP block before the new artifact scanner can help. Consider CDATA-aware parsing for the SHIP block too.

2. **P2 — `<SUMMARY>` extraction from SHIP inner body**: `parsers/v1.ts:393` extracts `<SUMMARY>` from the entire SHIP body. If the artifact CDATA contains `<SUMMARY>...</SUMMARY>` before the real summary, the parser will use artifact content as the ship summary. Consider excluding the matched artifact body from sibling tag extraction, or parsing after the artifact close tag.

---

**Current head looks good to me.** The P2s are edge cases that can be addressed in a follow-up if needed — the main artifact extraction + endpoint + CDATA fix are solid.

apps/daemon/src/critique/artifact-handler.ts
Nagendhra-web: Fixed in 658687d. Replaced the `lstat(path) -> validate -> createReadStream(path)` flow with a single open-and-stream sequence: `fs.open(path, O_RDONLY | O_NOFOLLOW)` returns a FileHandle, we run `fileHandle.stat()` for the size + isFile check on the open fd, and stream via `fileHandle.createReadStream({ autoClose: true })` so the bytes we describe are the bytes we serve. `O_NOFOLLOW` makes `fs.open` fail with ELOOP on POSIX when the path resolves to a symlink, replacing the previous `lstat`-based symlink rejection. On Windows the constant is undefined and the flag falls through to 0; the path-traversal guard plus the writer never producing symlinks keep that path safe in practice. Symlink + vanished-file regression cases continue to return 404.

apps/daemon/src/critique/parsers/v1.ts
```diff
@@ -357,14 +364,32 @@ function* drain(state: State): Generator<PanelEvent> {
...
+      const artifactAttrs = parseAttrs(artifactExtraction.attrText);
+      const artifactMime = artifactAttrs['mime'] ?? '';
+      // CDATA-wrapped bodies are unwrapped inside `extractArtifactBlock`,
+      // which scans for `]]></ARTIFACT>` (with optional whitespace) so a
+      // legitimate JS / HTML payload that contains a literal `</ARTIFACT>`
+      // sentinel inside a string or comment does not truncate the body
+      // (mrcfps follow-up on PR #1085).
+      const artifactBody = artifactExtraction.body;
+
       const summary = (inner.match(/<SUMMARY>([\s\S]*?)<\/SUMMARY>/)?.[1] ?? '').trim();
```
mrcfps: `summary` is still extracted from the full `inner` SHIP body, so a valid artifact like `<![CDATA[<div><SUMMARY>artifact text</SUMMARY></div>]]>` will satisfy this regex before the real `<SUMMARY>...</SUMMARY>` sibling tag is reached. In that case the emitted `critique.ship.summary` becomes artifact content instead of the actual ship summary, which means rerun/history consumers can persist and display the wrong summary for arbitrary shipped HTML. Please parse the summary from the SHIP body with the matched artifact range removed (or scan forward from the end of `artifactExtraction`) and add a regression case where the artifact HTML contains a literal `<SUMMARY>` pair before the real summary.

mrcfps (commented):
@Nagendhra-web Thanks for the follow-up here — the artifact persistence and endpoint hardening look solid overall. I found one parser edge case still worth tightening before rerun/history consumers start depending on arbitrary shipped HTML bodies.


lefarcen (commented):
Solid endpoint hardening fixes! ✅

**mrcfps's two P2 findings: both resolved**

1. **responseCapBytes logic** — Lines 49-59 now honor any positive finite `responseCapBytes` exactly as configured. The old `Math.max(DEFAULT, configured)` is gone, so a deployment that lowers `OD_CRITIQUE_PARSER_MAX_BLOCK_BYTES` below 4 MiB now sees the endpoint refuse the same byte range the writer refuses. Read-time guard matches write policy 1:1. ✅

2. **TOCTOU gap** — Lines 127-148 replaced `lstat(path) → validate → createReadStream(path)` with a single-open flow: `fs.open(path, O_RDONLY | O_NOFOLLOW)` → `fileHandle.stat()` → `fileHandle.createReadStream({ autoClose: true })`. The size and file-type checks now describe the exact fd being streamed, no path re-resolution after validation. ✅

---

**Current head looks clean.** The artifact extraction + endpoint + CDATA parsing + security hardening are all solid. No new P1/P2 issues found. Great work iterating through the feedback rounds!

Nagendhra-web:
@lefarcen — folded both of your earlier non-blocking P2s into c925fff alongside mrcfps's `<SUMMARY>` finding, since all three were the same class of bug (SHIP-boundary tag lookups walking past CDATA-wrapped artifact bodies):

1. **`</SHIP>` sentinel inside CDATA**: parser now uses `indexOfOutsideCdata(slice, '</SHIP>')` for the SHIP closer scan. CDATA artifacts that contain a literal `</SHIP>` no longer truncate the block.
2. **`<SUMMARY>` extraction confused by artifact body**: `extractArtifactBlock` now reports `blockEnd` and the SUMMARY regex runs only on `inner.slice(blockEnd)`, so artifact-internal `<SUMMARY>...</SUMMARY>` pairs can't hijack the ship summary.

Both have dedicated regression tests in [`apps/daemon/tests/parser.test.ts`](https://github.com/nexu-io/open-design/blob/feat/critique-theater-phase6.2-artifact-extraction/apps/daemon/tests/parser.test.ts) ("SHIP sentinel safety inside CDATA" describe block). 203/203 tests pass locally.

apps/daemon/src/critique/artifact-handler.ts
```diff
@@ -0,0 +1,218 @@
...
+    // file-type checks describe the exact bytes we are about to stream
+    // (mrcfps follow-up on PR #1085).
+    //
+    // O_NOFOLLOW is a POSIX flag; on Windows the constant resolves to
+    // undefined and the open fall-through behaves like a normal read.
+    // The path-traversal guard above plus the orchestrator never writing
+    // through symlinks make Windows still safe in practice.
+    const noFollowFlag =
+      typeof fsConstants.O_NOFOLLOW === 'number' ? fsConstants.O_NOFOLLOW : 0;
+    const openFlags = fsConstants.O_RDONLY | noFollowFlag;
```
mrcfps, line 141: On Windows this range falls back to `openFlags = O_RDONLY` because `fsConstants.O_NOFOLLOW` is undefined, so `fs.open(resolvedArtifactPath, openFlags)` will happily follow a symlink that lives *inside* `critique-artifacts`. That breaks the guarantee this handler is trying to add: a tampered row can still point at something like `<artifactsRoot>\p1\run\artifact.html` where `artifact.html` is a symlink to `C:\Users\...\secret.txt`, pass the `path.relative()` check, and then stream bytes from outside `artifactsRoot` because `fileHandle.stat()` describes the target rather than the link. The new test suite also skips the symlink case on Windows, so this bypass is currently untested. Please add a Windows-safe link/realpath check here before opening (for example, compare `await fs.realpath(resolvedArtifactPath)` back against `artifactsRoot`, or use Windows reparse-point detection) and cover it with a platform-specific regression so the outside-root / no-symlink guarantee holds on both POSIX and Windows.

mrcfps (commented):
@Nagendhra-web Thanks for the follow-up here — the artifact parsing/order-ordering and endpoint hardening look solid overall. I found one Windows-specific gap in the new artifact boundary that seems worth closing before other callers start depending on this route as the only supported way to read persisted critique artifacts.


lefarcen (commented):
Hey @Nagendhra-web! 👋

mrcfps's `<SUMMARY>` extraction fix + lefarcen's `</SHIP>` sentinel fix both look solid. Checked the code:

✅ **`<SUMMARY>` scoped correctly** (line 407-408 in `parsers/v1.ts`): regex now runs on `inner.slice(artifactExtraction.blockEnd)`, so artifact-internal `<SUMMARY>...</SUMMARY>` pairs can't hijack the ship summary. Regression test covers it.

✅ **`</SHIP>` sentinel safe** (line 337): SHIP closer scan now uses `indexOfOutsideCdata(slice, '</SHIP>')`, so a literal `</SHIP>` inside CDATA artifact body won't truncate the block. Regression test with JS string literal inside artifact body covers it.

✅ **`indexOfOutsideCdata` helper** (lines 563+): state machine correctly skips CDATA spans when scanning for structural tags.

No new P1/P2 issues found. Nice work closing the parser edge cases!

mrcfps (approved):

Issue #550, opened 2026-05-05
Feature request: Add focus mode to collapse chat and expand workspace preview


Romantin:
### Summary
The workspace/artifact preview area can feel too narrow when the chat pane and preview are shown side by side. At 100% zoom, generated webpages or report-like HTML artifacts often wrap text awkwardly and look cramped. To inspect the design properly, the user may need to zoom out.

A useful improvement would be a focus mode that lets users collapse the chat sidebar and expand the workspace preview to use the full window width.

### Current behavior
- The chat pane and workspace preview stay visible side by side.
- The preview area can be narrow for generated webpages, reports, dashboards, or other wide HTML artifacts.
- At 100% zoom, text and layout can become compressed, so the artifact looks worse than it would at a normal page width.
- The workaround is to reduce the preview zoom, but that makes text smaller and is not ideal for reviewing the design.

### Expected behavior
- Add a way to collapse or hide the chat/sidebar temporarily.
- When collapsed, the workspace/artifact preview should expand to use the available window width.
- Keep the current project/file tabs and preview controls available.
- Provide an obvious way to restore the chat pane.

### Why this matters
Open Design often generates webpages and design artifacts that need a realistic preview width. A full-width workspace/focus mode would make it easier to inspect typography, spacing, and layout at 100% zoom without needing to shrink the preview.

This would also match a common design-review workflow: keep the conversation available when iterating, but allow the artifact preview to take over the screen when reviewing the result.

### Environment
- Open Design desktop beta on Windows.

lefarcen:
Hi @Romantin! 👋

Thanks for this feature request — the focus mode idea makes a lot of sense for design review workflows. When inspecting generated webpages, dashboards, or report-like artifacts, having the full window width for the preview would definitely help catch layout and typography issues that get compressed in the side-by-side view.

I'm investigating the current UI layout architecture to see how feasible this would be and whether it could be a good candidate for community contribution. I'll follow up within 24h with findings and next steps.

Thanks for making open-design better!
— open-design team

lefarcen:
PR #604 is under review and implements the requested focus mode:

**UI behavior:**
- Focus mode collapses chat and expands workspace preview to full width
- Tab strip + focus/restore action remain visible (pinned outside the scrollable tab bar)
- Chat state is preserved (hidden, not unmounted)
- Toggle button in workspace header: "Focus workspace" ↔ "Show chat"

Issue stays open until the fix lands. Try it out once #604 merges!

[...]

PR #604, opened 2026-05-06
Fix agent CLI config and workspace focus mode

Sid-Qin:
## Summary
- Add daemon-backed per-agent CLI config locations for Claude Code and Codex.
- Inject configured CLI env into agent detection, model probing, and chat runtime spawns.
- Add workspace focus mode that collapses chat and expands the workspace preview while keeping tabs and controls available.
- Keep the focus/restore action pinned outside the horizontally scrollable tab strip so chat can always be restored.

## Issues
Closes #571
Closes #550

## Root Cause
The packaged app did not inherit terminal-only CLI config env, so authenticated Claude/Codex installs using custom config locations could not be discovered or launched reliably. The workspace layout also kept chat and preview side by side with no full-width review mode.

## Implementation
- Add `agentCliEnv` to shared app config contracts and web config sync.
- Persist only supported non-secret env keys: `CLAUDE_CONFIG_DIR` for Claude and `CODEX_HOME` for Codex.
- Apply those env values consistently to daemon detection, ACP/model probing, and agent spawn.
- Add Settings UI for CLI config locations.
- Add focus-mode state to ProjectView and a workspace tab-bar toggle.
- Hide chat during focus mode without unmounting it, so draft/internal state is preserved.
- Split the workspace header into a scrollable tablist plus pinned actions so `Show chat` remains visible when many tabs are open.

## UI / UX
```text
Normal
+----------------+--------------------------------+
| Chat           | Tabs scroll  ...   Focus button |
|                | Workspace preview               |
+----------------+--------------------------------+

Focus mode
+-------------------------------------------------+
| Tabs scroll  ...                       Show chat |
| Workspace preview                               |
+-------------------------------------------------+
```

## Testing
- `pnpm guard`
- `pnpm typecheck`
- `pnpm --filter @open-design/daemon test -- tests/app-config.test.ts tests/agents.test.ts` (56 files / 766 tests)
- `pnpm --filter @open-design/web test -- tests/state/config.test.ts tests/components/SettingsDialog.test.ts tests/components/FileWorkspace.test.tsx` (33 files / 265 tests)
- `pnpm --filter @open-design/contracts build`
- `pnpm --filter @open-design/daemon build`
- `pnpm --filter @open-design/desktop build`
- `pnpm --filter @open-design/web build`
- `git diff --check`

## Notes
- Local verification ran with Node `v22.16.0`; this repository declares Node `~24`, so pnpm emitted engine warnings while the commands above still completed successfully.

PR #563, opened 2026-05-05
feat: add resizable chat panel

ferasbusiness666:
Adds a draggable resize divider between the chat panel and the preview/files workspace.

Changes:
- Added resizable chat panel width
- Saves chat width to localStorage
- Clamps width between 320px and 720px
- Adds an 8px vertical resize handle
- Keeps layout resize instant while dragging

Tested locally and confirmed the resize behavior works as expected.

PR #1088, opened 2026-05-09
feat(web): add draggable resize handles in manual edit mode

### LINKED ISSUE #1059: 预览窗口，编辑状态下 Layer、Manual editor、preview 宽度布局优化建议
Image
如图所示，编辑过程中，页面的试试效果预览和选择模块由于宽度占比较小，切宽度不可调整，导致页面被挤压，显示错位，是否可以实现通过拖拽调整三个模块的宽度

Morzorz:
Closes #1059 
Replace static CSS grid with adjustable three-column layout: Layers | Editor | Preview. Each column has explicit pixel width with min-width enforcement and no max-width. Drag handles redistribute space between adjacent panels only. Widths persist to localStorage. Keyboard accessible with ArrowLeft/Right/Home.

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -4534,15 +4534,9 @@ function HtmlViewer({
     </div>
   );
 
-  // ARIA max widths — use current sibling panel widths, not their mins
-  const ariaWs = workspaceRef.current;
-  const ariaWsW = ariaWs ? ariaWs.clientWidth : 0;
-  const ariaLayersMax = ariaWsW
-    ? Math.max(LAYERS_MIN_WIDTH, Math.round(ariaWsW - MANUAL_EDIT_HANDLE_WIDTH * 2 - EDITOR_MIN_WIDTH - previewPanelWidth))
-    : 9999;
-  const ariaPreviewMax = ariaWsW
-    ? Math.max(PREVIEW_PANEL_MIN_WIDTH, Math.round(ariaWsW - MANUAL_EDIT_HANDLE_WIDTH * 2 - layersWidth - EDITOR_MIN_WIDTH))
-    : 9999;
+  // ARIA max widths — derived from redistributeWidths limits (no DOM measurement needed)
+  const ariaLayersMax = Math.max(LAYERS_MIN_WIDTH, layersWidth + editorWidth - EDITOR_MIN_WIDTH);
+  const ariaPreviewMax = Math.max(PREVIEW_PANEL_MIN_WIDTH, previewPanelWidth + editorWidth - EDITOR_MIN_WIDTH);
 
   return (
     <div className="viewer html-viewer">
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -3084,6 +3084,7 @@ function HtmlViewer({
   const [editorWidth, setEditorWidth] = useState(() => readSavedWidth(EDITOR_WIDTH_KEY, EDITOR_DEFAULT_WIDTH, EDITOR_MIN_WIDTH));
   const [previewPanelWidth, setPreviewPanelWidth] = useState(() => readSavedWidth(PREVIEW_WIDTH_KEY, PREVIEW_PANEL_DEFAULT_WIDTH, PREVIEW_PANEL_MIN_WIDTH));
   const [resizingPanel, setResizingPanel] = useState<'layers' | 'preview' | null>(null);
+  const hasNormalizedRef = useRef(false);
   const layersWidthRef = useRef(layersWidth);
   const editorWidthRef = useRef(editorWidth);
   const previewPanelWidthRef = useRef(previewPanelWidth);
@@ -4395,11 +4396,15 @@ function HtmlViewer({
     if (pointerFrameRef.current !== null) cancelAnimationFrame(pointerFrameRef.current);
   }, []);
 
-  // Re-clamp restored widths against the actual workspace width (not viewport)
+  // Re-clamp restored widths against the actual workspace width (not viewport).
+  // Runs once: after source loads and manualEditMode makes the workspace visible.
   useLayoutEffect(() => {
+    if (hasNormalizedRef.current) return;
     const ws = workspaceRef.current;
     if (!ws) return;
-    const available = ws.clientWidth - MANUAL_EDIT_HANDLE_WIDTH * 2 - 4 * MANUAL_EDIT_GAP;
+    const cs = window.getComputedStyle(ws);
+    const padH = Number.parseFloat(cs.paddingLeft) + Number.parseFloat(cs.paddingRight);
+    const available = ws.clientWidth - padH - MANUAL_EDIT_HANDLE_WIDTH * 2 - 4 * MANUAL_EDIT_GAP;
     if (available <= 0) return;
     const n = normalizePanelWidths(
       layersWidthRef.current,
@@ -4413,7 +4418,8 @@ function HtmlViewer({
     setEditorWidth(n.editor);
     previewPanelWidthRef.current = n.preview;
     setPreviewPanelWidth(n.preview);
-  }, []);
+    hasNormalizedRef.current = true;
+  }, [source, manualEditMode]);
 
   const handleResizePointerDown = useCallback((side: 'layers' | 'preview') => (event: React.PointerEvent<HTMLDivElement>) => {
     if (event.button !== 0) return;
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -1,4 +1,4 @@
-import { useCallback, useEffect, useId, useMemo, useRef, useState, type CSSProperties, type MouseEvent as ReactMouseEvent, type ReactNode } from 'react';
+import { useCallback, useEffect, useId, useLayoutEffect, useMemo, useRef, useState, type CSSProperties, type MouseEvent as ReactMouseEvent, type ReactNode } from 'react';
 import { createPortal } from 'react-dom';
 import { APP_CHROME_FILE_ACTIONS_ID } from './AppChromeHeader';
 import { MarkdownRenderer, artifactRendererRegistry } from '../artifacts/renderer-registry';
@@ -162,22 +162,45 @@ const MANUAL_EDIT_HANDLE_WIDTH = 8;
 const LAYERS_WIDTH_KEY = 'open-design.manualEdit.layersWidth';
 const EDITOR_WIDTH_KEY = 'open-design.manualEdit.editorWidth';
 const PREVIEW_WIDTH_KEY = 'open-design.manualEdit.previewWidth';
+const MANUAL_EDIT_GAP = 10;
 const MANUAL_EDIT_KEYBOARD_STEP = 16;
 
-function readSavedWidth(key: string, fallback: number, min: number, max: number): number {
+function readSavedWidth(key: string, fallback: number, min: number): number {
   if (typeof window === 'undefined') return fallback;
   try {
     const raw = window.localStorage.getItem(key);
     const parsed = raw ? Number.parseInt(raw, 10) : Number.NaN;
-    if (!Number.isFinite(parsed)) return fallback;
-    const clampedMin = Math.max(min, Math.round(parsed));
-    return Math.min(clampedMin, max);
+    return Number.isFinite(parsed) ? Math.max(min, Math.round(parsed)) : fallback;
   } catch { return fallback; }
 }
 
-function panelMaxWidth(minOtherA: number, minOtherB: number): number {
-  if (typeof window === 'undefined') return 9999;
-  return Math.round(window.innerWidth - minOtherA - minOtherB - MANUAL_EDIT_HANDLE_WIDTH * 2);
+export function normalizePanelWidths(
+  layers: number,
+  editor: number,
+  preview: number,
+  available: number,
+): { layers: number; editor: number; preview: number } {
+  let l = Math.max(LAYERS_MIN_WIDTH, Math.round(layers));
+  let e = Math.max(EDITOR_MIN_WIDTH, Math.round(editor));
+  let p = Math.max(PREVIEW_PANEL_MIN_WIDTH, Math.round(preview));
+
+  let overflow = l + e + p - available;
+  if (overflow <= 0) return { layers: l, editor: e, preview: p };
+
+  // Squeeze editor first (flexible middle panel), then side panels if still overflowing
+  const squeezeE = Math.min(overflow, e - EDITOR_MIN_WIDTH);
+  e -= squeezeE;
+  overflow -= squeezeE;
+  if (overflow > 0) {
+    const squeezeL = Math.min(overflow, l - LAYERS_MIN_WIDTH);
+    l -= squeezeL;
+    overflow -= squeezeL;
+  }
+  if (overflow > 0) {
+    const squeezeP = Math.min(overflow, p - PREVIEW_PANEL_MIN_WIDTH);
+    p -= squeezeP;
+  }
+  return { layers: l, editor: e, preview: p };
 }
 
 function redistributeWidths(
@@ -3057,9 +3080,9 @@ function HtmlViewer({
   const manualEditSavingRef = useRef(false);
 
   // Manual edit panel resize state
-  const [layersWidth, setLayersWidth] = useState(() => readSavedWidth(LAYERS_WIDTH_KEY, LAYERS_DEFAULT_WIDTH, LAYERS_MIN_WIDTH, panelMaxWidth(EDITOR_MIN_WIDTH, PREVIEW_PANEL_MIN_WIDTH)));
-  const [editorWidth, setEditorWidth] = useState(() => readSavedWidth(EDITOR_WIDTH_KEY, EDITOR_DEFAULT_WIDTH, EDITOR_MIN_WIDTH, panelMaxWidth(LAYERS_MIN_WIDTH, PREVIEW_PANEL_MIN_WIDTH)));
-  const [previewPanelWidth, setPreviewPanelWidth] = useState(() => readSavedWidth(PREVIEW_WIDTH_KEY, PREVIEW_PANEL_DEFAULT_WIDTH, PREVIEW_PANEL_MIN_WIDTH, panelMaxWidth(LAYERS_MIN_WIDTH, EDITOR_MIN_WIDTH)));
+  const [layersWidth, setLayersWidth] = useState(() => readSavedWidth(LAYERS_WIDTH_KEY, LAYERS_DEFAULT_WIDTH, LAYERS_MIN_WIDTH));
+  const [editorWidth, setEditorWidth] = useState(() => readSavedWidth(EDITOR_WIDTH_KEY, EDITOR_DEFAULT_WIDTH, EDITOR_MIN_WIDTH));
+  const [previewPanelWidth, setPreviewPanelWidth] = useState(() => readSavedWidth(PREVIEW_WIDTH_KEY, PREVIEW_PANEL_DEFAULT_WIDTH, PREVIEW_PANEL_MIN_WIDTH));
   const [resizingPanel, setResizingPanel] = useState<'layers' | 'preview' | null>(null);
   const layersWidthRef = useRef(layersWidth);
   const editorWidthRef = useRef(editorWidth);
@@ -4370,6 +4393,26 @@ function HtmlViewer({
   useEffect(() => () => {
     pointerCleanupRef.current?.();
     if (pointerFrameRef.current !== null) cancelAnimationFrame(pointerFrameRef.current);
+  }, []);
+
+  // Re-clamp restored widths against the actual workspace width (not viewport)
+  useLayoutEffect(() => {
+    const ws = workspaceRef.current;
+    if (!ws) return;
+    const available = ws.clientWidth - MANUAL_EDIT_HANDLE_WIDTH * 2 - 4 * MANUAL_EDIT_GAP;
+    if (available <= 0) return;
+    const n = normalizePanelWidths(
+      layersWidthRef.current,
+      editorWidthRef.current,
+      previewPanelWidthRef.current,
+      available,
+    );
+    layersWidthRef.current = n.layers;
+    setLayersWidth(n.layers);
+    editorWidthRef.current = n.editor;
+    setEditorWidth(n.editor);
+    previewPanelWidthRef.current = n.preview;
+    setPreviewPanelWidth(n.preview);
   }, []);
 
   const handleResizePointerDown = useCallback((side: 'layers' | 'preview') => (event: React.PointerEvent<HTMLDivElement>) => {
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -4425,6 +4425,37 @@ function HtmlViewer({
     hasNormalizedRef.current = true;
   }, [source, manualEditMode]);
 
+  // Re-normalize when workspace width changes during an active manual-edit session
+  // (e.g., user drags the ProjectView chat/workspace split).
+  useEffect(() => {
+    if (!manualEditMode) return;
+    const ws = workspaceRef.current;
+    if (!ws) return;
+    const observer = new ResizeObserver(() => {
+      const cs = window.getComputedStyle(ws);
+      const padH = Number.parseFloat(cs.paddingLeft) + Number.parseFloat(cs.paddingRight);
+      const available = ws.clientWidth - padH - MANUAL_EDIT_HANDLE_WIDTH * 2 - 4 * MANUAL_EDIT_GAP;
+      if (available <= 0) return;
+      const n = normalizePanelWidths(
+        layersWidthRef.current,
+        editorWidthRef.current,
+        previewPanelWidthRef.current,
+        available,
+      );
+      if (n.layers === layersWidthRef.current
+        && n.editor === editorWidthRef.current
+        && n.preview === previewPanelWidthRef.current) return;
+      layersWidthRef.current = n.layers;
+      setLayersWidth(n.layers);
+      editorWidthRef.current = n.editor;
+      setEditorWidth(n.editor);
+      previewPanelWidthRef.current = n.preview;
+      setPreviewPanelWidth(n.preview);
+    });
+    observer.observe(ws);
+    return () => observer.disconnect();
+  }, [source, manualEditMode]);
+
   const handleResizePointerDown = useCallback((side: 'layers' | 'preview') => (event: React.PointerEvent<HTMLDivElement>) => {
     if (event.button !== 0) return;
     const workspace = workspaceRef.current;
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useId, useMemo, useRef, useState, type CSSProperties, type MouseEvent as ReactMouseEvent, type ReactNode } from 'react';
+import { useCallback, useEffect, useId, useMemo, useRef, useState, type CSSProperties, type MouseEvent as ReactMouseEvent, type ReactNode } from 'react';
 import { createPortal } from 'react-dom';
 import { APP_CHROME_FILE_ACTIONS_ID } from './AppChromeHeader';
 import { MarkdownRenderer, artifactRendererRegistry } from '../artifacts/renderer-registry';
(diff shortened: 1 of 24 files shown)


lefarcen:
Hi @Morzorz! 🎉
Thanks for the contribution — this addresses a real UX pain point with the cramped preview in manual edit mode.

— open-design team

lefarcen:
Agent-to-agent note for PR #1088 related research:

RELATED_RESEARCH: linked_issues=[#1059 linked_by_body \"Closes #1059\" exact request for draggable width adjustment of Layer/Manual editor/Preview panels; #347 related_context broader generated-design/manual-edit feature request; #550 related_theme preview/workspace width too narrow, already closed by #604]; related_prs=[#563 related_precedent resizable chat panel with localStorage, keyboard access, RTL/persistence review notes and explicitly referenced from #1059; #620 related_foundation implemented Manual Edit mode and added ManualEditPanel/FileViewer integration; #604 related_theme workspace focus mode for cramped preview width, closes #550; #469 related_overlap open broad visual editing PR touching FileViewer/index.css and likely conflict surface]; none_reason=n/a; search_terms=[\"Closes #1059\",\"manual edit resize\",\"draggable resize\",\"Layers Manual editor Preview\",\"preview width\",\"宽度 拖拽\",\"FileViewer\",\"ManualEditPanel\",\"manual edit mode\",\"three column layout\",\"resize handle\",\"localStorage width\",\"focus mode collapse chat\"]; checked_sources=[\"gh pr view 1088\",\"gh issue view 1059\",\"gh pr view 563\",\"gh issue view 550\",\"gh pr view 604\",\"gh pr view 620\",\"gh issue view 347\",\"gh pr view 469\",\"gh search issues/prs for listed terms\",\"synclo_memory_search unavailable in current toolset\"]; recommended_issue_comments=[#1059 \"PR #1088 is linked with Closes #1059 and implements the requested draggable width adjustment for Layers / Manual editor / Preview, including persisted widths and keyboard controls. Recommend closing after merge/verification.\", #347 \"Optional context comment only: #1088 improves manual edit usability after #620 by making the three-panel edit layout adjustable; it does not fully resolve the broader visual editing request.\", #550 \"No comment recommended; already closed by #604 and only thematically related.\"]

apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4356,178 @@ function HtmlViewer({
...
+    const workspace = workspaceRef.current;
+    const isRtl = workspace ? window.getComputedStyle(workspace).direction === 'rtl' : false;
+
+    if (event.key === 'ArrowLeft') {
+      delta = -MANUAL_EDIT_KEYBOARD_STEP;
+    } else if (event.key === 'ArrowRight') {
+      delta = MANUAL_EDIT_KEYBOARD_STEP;
+    } else if (event.key === 'Home') {
+      delta = side === 'layers' ? LAYERS_MIN_WIDTH - layersWidthRef.current : PREVIEW_PANEL_MIN_WIDTH - previewPanelWidthRef.current;
+    } else {
```
lefarcen, line 4566: **P2 — Home key backwards for preview handle**

For `side === 'preview'`, the code sets `delta = PREVIEW_PANEL_MIN_WIDTH - previewPanelWidthRef.current`, but `redistributeWidths('preview', delta, ...)` computes `newPreview = preview - delta`. This means pressing Home expands the preview instead of shrinking it to minimum.

Fix:
```typescript
} else if (event.key === 'Home') {
  delta = side === 'layers' 
    ? LAYERS_MIN_WIDTH - layersWidthRef.current
    : previewPanelWidthRef.current - PREVIEW_PANEL_MIN_WIDTH;  // inverted sign
}
```

Also worth adding a test for both handles' Home behavior to catch this kind of regression.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4696,75 +4937,46 @@ function HtmlViewer({
...
+                    if (selectedManualEditTarget) selectManualEditTarget(selectedManualEditTarget);
+                  }}
+                  onUndo={() => {
+                    void undoManualEdit();
+                  }}
+                  onRedo={() => {
+                    void redoManualEdit();
+                  }}
+                  childrenAfter={<div className="manual-edit-resize-handle" role="separator" aria-orientation="vertical" aria-label={t('manualEdit.resizePreview')} tabIndex={0} onPointerDown={handlePreviewResizePointerDown} onKeyDown={handlePreviewResizeKeyDown} />}
+                >
```
lefarcen, line 5074: **P2 — Missing aria-value attributes for screen readers**

The focusable separators have `role="separator"` and aria-label, but lack `aria-valuemin`, `aria-valuemax`, and `aria-valuenow`. Screen reader users can focus and operate the handles but can't hear current width or bounds.

Precedent: PR #563's chat resize includes these attrs. Add:
- `aria-valuemin={LAYERS_MIN_WIDTH}` / `{PREVIEW_PANEL_MIN_WIDTH}`
- `aria-valuenow={layersWidth}` / `{previewPanelWidth}`
- `aria-valuemax={<computed max based on adjacent editor minimum>}`

See ProjectView.tsx chat resize for reference implementation.

apps/web/src/components/FileViewer.tsx
```diff
@@ -151,6 +151,53 @@ const MARKDOWN_COPY_BLOCK_ATTR = 'data-copy-code-block';
...
+const EDITOR_DEFAULT_WIDTH = 420;
+const PREVIEW_PANEL_MIN_WIDTH = 280;
+const PREVIEW_PANEL_DEFAULT_WIDTH = 344;
+const MANUAL_EDIT_HANDLE_WIDTH = 8;
+const LAYERS_WIDTH_KEY = 'open-design.manualEdit.layersWidth';
+const EDITOR_WIDTH_KEY = 'open-design.manualEdit.editorWidth';
+const PREVIEW_WIDTH_KEY = 'open-design.manualEdit.previewWidth';
+const MANUAL_EDIT_KEYBOARD_STEP = 16;
+
+function readSavedWidth(key: string, fallback: number): number {
```
lefarcen: **P3 — Stale localStorage not clamped on load**

`readSavedWidth` trusts localStorage without clamping, so corrupt/stale values can render panels below stated minimums before any resize interaction.

Since this feature's correctness claim is min-width enforcement, either:
1. Clamp in `readSavedWidth`: `Math.max(minWidth, parsed)`
2. Use per-panel read helpers that apply their respective mins

Not urgent (only triggers if localStorage is manually tampered), but worth fixing for consistency.

lefarcen (commented):
Hi @Morzorz! Nice work on the resize implementation — the pointer capture + RAF throttling + RTL support are solid foundations.

Found 3 issues during review (2 P2, 1 P3). The preview handle's Home key behavior is backwards, and the aria attributes are incomplete for screen readers. See inline comments below.

lefarcen:
Agent-to-agent review note for PR #1088:

Findings:

1. `apps/web/src/components/FileViewer.tsx:4465-4473` — P2 — `Home` is wrong for the preview separator. For `side === 'preview'`, the code sets `delta = PREVIEW_PANEL_MIN_WIDTH - previewPanelWidthRef.current`, but `redistributeWidths('preview', delta, ...)` computes `newPreview = preview - delta`, so pressing `Home` in LTR expands the preview instead of shrinking it to its minimum. This breaks the advertised keyboard behavior. Fix by using `previewPanelWidthRef.current - PREVIEW_PANEL_MIN_WIDTH` for the preview handle, with a regression test for both handles’ `Home` behavior.

2. `apps/web/src/components/FileViewer.tsx:4971-4973` — P2 — The focusable separators lack `aria-valuemin`, `aria-valuemax`, and `aria-valuenow`, unlike the existing chat resize precedent in `ProjectView.tsx`. With only `role=\"separator\"` and an aria-label, screen reader users can focus and operate the handles but cannot hear the current width or bounds. Add value attributes for each handle: min is the resized panel min; max is constrained by the adjacent editor minimum; now is the current panel width.

3. `apps/web/src/components/FileViewer.tsx:167-173` — P3 — Persisted widths are trusted without clamping, so a stale/corrupt localStorage value can render panels below the stated minimums on initial load before any resize interaction runs. Since this feature’s correctness claim is min-width enforcement, `readSavedWidth` should clamp each key to its panel minimum, or use per-panel read helpers that apply `LAYERS_MIN_WIDTH`, `EDITOR_MIN_WIDTH`, and `PREVIEW_PANEL_MIN_WIDTH`.

Testing gap:

No resize-specific tests were added for the nontrivial behavior here: keyboard resizing, persistence, RTL inversion, and min clamping. At minimum, the preview `Home` regression above should be covered in `FileViewer` tests so keyboard accessibility does not silently regress.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4356,178 @@ function HtmlViewer({
...
+    if (event.key === 'ArrowLeft') {
+      delta = -MANUAL_EDIT_KEYBOARD_STEP;
+    } else if (event.key === 'ArrowRight') {
+      delta = MANUAL_EDIT_KEYBOARD_STEP;
+    } else if (event.key === 'Home') {
+      delta = side === 'layers' ? LAYERS_MIN_WIDTH - layersWidthRef.current : previewPanelWidthRef.current - PREVIEW_PANEL_MIN_WIDTH;
+    } else {
+      return;
+    }
+    if (isRtl) delta = -delta;
```
lefarcen: **P3 — RTL Home key regression**

The new preview Home calculation fixes LTR, but the unconditional RTL flip now makes Home backwards in RTL:

`previewPanelWidthRef.current - PREVIEW_PANEL_MIN_WIDTH` becomes negative (e.g. 280 - 280 = 0, or if preview is 250 → -30), then `if (isRtl) delta = -delta` flips it to positive, so preview expands instead of collapsing.

Fix: only flip arrow-key deltas, or handle Home as an absolute target operation that bypasses the RTL delta logic entirely.

lefarcen (commented):
Thanks for the quick fix @Morzorz! The LTR Home key direction is now correct, but the fix introduced an RTL edge case. The two P2 accessibility items (aria attributes + localStorage clamping) remain.

mrcfps (commented):
@Morzorz Thanks for iterating on the keyboard resize behavior. I re-checked the updated `Home` handling on the latest head and found one remaining RTL regression in the shared key path.


apps/web/src/components/FileViewer.tsx
mrcfps: `handleResizeKeyDown` still applies the generic RTL sign flip to `Home`, even though both `Home` branches now compute an absolute delta that is already meant to move the active panel to its minimum width. In RTL locales that inverts the operation for both handles: the layers separator grows instead of snapping to `LAYERS_MIN_WIDTH`, and the preview separator expands instead of shrinking to `PREVIEW_PANEL_MIN_WIDTH`. Because this PR advertises keyboard-accessible resizing, Arabic/Farsi users still get the wrong behavior here. Please keep the RTL inversion only for `ArrowLeft`/`ArrowRight`, or move `Home` onto a direct "set target width to min" path that bypasses the sign flip entirely; adding a `FileViewer` RTL keyboard regression test for both handles would keep this from slipping again.

Morzorz: That’s an incredibly detailed review! I’ve completed fixes for all the problems.

apps/web/src/components/FileViewer.tsx
```diff
@@ -151,6 +151,53 @@ const MARKDOWN_COPY_BLOCK_ATTR = 'data-copy-code-block';
...
+const EDITOR_WIDTH_KEY = 'open-design.manualEdit.editorWidth';
+const PREVIEW_WIDTH_KEY = 'open-design.manualEdit.previewWidth';
+const MANUAL_EDIT_KEYBOARD_STEP = 16;
+
+function readSavedWidth(key: string, fallback: number, min: number): number {
+  if (typeof window === 'undefined') return fallback;
+  try {
+    const raw = window.localStorage.getItem(key);
+    const parsed = raw ? Number.parseInt(raw, 10) : Number.NaN;
+    return Number.isFinite(parsed) ? Math.max(min, Math.round(parsed)) : fallback;
```
lefarcen, line 173: **P3 — localStorage clamping incomplete**

`readSavedWidth` now clamps to minimums, but doesn't cap maximums. A stale value from a larger viewport or corrupted localStorage can still make `gridTemplateColumns` restore an oversized layout before any resize interaction normalizes it.

Consider adding a reasonable max (e.g., `Math.min(rawValue, MAX_PANEL_WIDTH)`) or validating against `window.innerWidth - MIN_OTHER_PANELS` to keep the sum within viewport bounds.

lefarcen (commented):
Great progress @Morzorz! 🎉 You've addressed both P2 issues:

✅ RTL Home key now works correctly — no longer inverted
✅ ARIA attributes added — screen readers can now announce resize values

One remaining P3 item: localStorage clamping currently only enforces minimums. Very large persisted values (e.g., from a different viewport or corrupted storage) can still restore an unusable oversized grid. See inline comment.

Overall: solid fix for the core accessibility and RTL issues.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4356,187 @@ function HtmlViewer({
...
+
+  // ARIA max widths — use current sibling panel widths, not their mins
+  const ariaWs = workspaceRef.current;
+  const ariaWsW = ariaWs ? ariaWs.clientWidth : 0;
+  const ariaLayersMax = ariaWsW
+    ? Math.max(LAYERS_MIN_WIDTH, Math.round(ariaWsW - MANUAL_EDIT_HANDLE_WIDTH * 2 - EDITOR_MIN_WIDTH - previewPanelWidth))
+    : 9999;
+  const ariaPreviewMax = ariaWsW
+    ? Math.max(PREVIEW_PANEL_MIN_WIDTH, Math.round(ariaWsW - MANUAL_EDIT_HANDLE_WIDTH * 2 - layersWidth - EDITOR_MIN_WIDTH))
+    : 9999;
```
mrcfps: `ariaLayersMax` and `ariaPreviewMax` are currently derived from `workspace.clientWidth`, but that measurement still includes the grid container's horizontal padding and its four `10px` gaps. On the default layout that makes the announced max about 60px larger than what `redistributeWidths(...)` can ever reach (for example the layers handle reports 440 even though the real cap is `layersWidth + editorWidth - EDITOR_MIN_WIDTH = 380`). Because these values feed `aria-valuemax`, screen readers will announce bounds that keyboard users can never actually hit. Please derive the max from the existing panel widths instead (`layersWidth + editorWidth - EDITOR_MIN_WIDTH` / `previewPanelWidth + editorWidth - EDITOR_MIN_WIDTH`), or subtract the padding + gaps before using `clientWidth`.

mrcfps (commented):
@Morzorz Thanks for iterating on the resize accessibility fixes. I re-checked the latest head and found one remaining non-blocking issue in the new ARIA bounds calculation; details are in the inline comment below.


Morzorz:
Hold on, there’s a problem with the two latest commits.

apps/web/src/components/FileViewer.tsx
```diff
@@ -151,6 +151,60 @@ const MARKDOWN_COPY_BLOCK_ATTR = 'data-copy-code-block';
...
+    const parsed = raw ? Number.parseInt(raw, 10) : Number.NaN;
+    if (!Number.isFinite(parsed)) return fallback;
+    const clampedMin = Math.max(min, Math.round(parsed));
+    return Math.min(clampedMin, max);
+  } catch { return fallback; }
+}
+
+function panelMaxWidth(minOtherA: number, minOtherB: number): number {
+  if (typeof window === 'undefined') return 9999;
+  return Math.round(window.innerWidth - minOtherA - minOtherB - MANUAL_EDIT_HANDLE_WIDTH * 2);
```
mrcfps: `panelMaxWidth(...)` is clamping restored widths against `window.innerWidth`, but `FileViewer` is usually rendered inside a narrower split workspace rather than taking the full viewport. That means a width saved while the workspace was focused or on a larger monitor can still pass this clamp and restore an unusable grid the next time the normal chat+workspace layout opens. For example, on a 1440px window `panelMaxWidth(EDITOR_MIN_WIDTH, PREVIEW_PANEL_MIN_WIDTH)` accepts `layersWidth=800`, but the split workspace is much smaller once the chat pane, padding, and gaps are present, so the first render overflows before the user touches the handle. Please clamp against the actual workspace width instead of the viewport (for example, normalize the restored widths after `workspaceRef` is measured, or derive the cap from the split container budget), and add a `FileViewer` regression test that seeds localStorage then renders in a constrained split-width container.

mrcfps (commented):
@Morzorz Thanks for iterating on the resize accessibility fixes. I re-checked the latest head and found one remaining persisted-layout issue when saved widths are restored into a narrower workspace.


apps/web/src/components/FileViewer.tsx
```diff
@@ -151,6 +151,60 @@ const MARKDOWN_COPY_BLOCK_ATTR = 'data-copy-code-block';
...
+const MANUAL_EDIT_KEYBOARD_STEP = 16;
+
+function readSavedWidth(key: string, fallback: number, min: number, max: number): number {
+  if (typeof window === 'undefined') return fallback;
+  try {
+    const raw = window.localStorage.getItem(key);
+    const parsed = raw ? Number.parseInt(raw, 10) : Number.NaN;
+    if (!Number.isFinite(parsed)) return fallback;
+    const clampedMin = Math.max(min, Math.round(parsed));
+    return Math.min(clampedMin, max);
```
lefarcen: P2 — `max` can be lower than `min` on narrower viewports, so this can restore persisted widths below the declared panel minimums. That makes the separator render with `aria-valuenow` below `aria-valuemin` and reintroduces invalid screen-reader bounds. Clamp as `Math.max(min, Math.min(clampedMin, max))`, or otherwise ensure the computed max never falls below the panel minimum.

lefarcen (changes requested):
Nice fix @Morzorz! 🎉 The ARIA bounds issue mrcfps flagged is resolved — the max values now come from the same state that redistributeWidths clamps against.

One new P2 accessibility issue found during re-review (see inline below). Once this is addressed, the PR will be ready.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4386,201 @@ function HtmlViewer({
...
+  useEffect(() => { layersWidthRef.current = layersWidth; }, [layersWidth]);
+  useEffect(() => { editorWidthRef.current = editorWidth; }, [editorWidth]);
+  useEffect(() => { previewPanelWidthRef.current = previewPanelWidth; }, [previewPanelWidth]);
+  useEffect(() => () => {
+    pointerCleanupRef.current?.();
+    if (pointerFrameRef.current !== null) cancelAnimationFrame(pointerFrameRef.current);
+  }, []);
+
+  // Re-clamp restored widths against the actual workspace width (not viewport)
+  useLayoutEffect(() => {
```
lefarcen, line 4401: **P2** This `useLayoutEffect` has an empty dependency list, so when `source` starts as `null` the render path shows `viewer-empty` and `workspaceRef.current` is absent. The effect returns and never reruns after `fetchProjectFileText` sets `source`, leaving oversized widths from localStorage un-normalized for the common async file-load path. Consider adding `source` to the dependency array or moving the clamp into a conditional branch inside the effect that fires when `workspaceRef` becomes available.

lefarcen (changes requested):
Hey @Morzorz, great progress on the ARIA bounds fix — that issue is resolved. Found one new P2 in the width restoration logic (see inline below).

apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4386,201 @@ function HtmlViewer({
...
+    const ws = workspaceRef.current;
+    if (!ws) return;
+    const available = ws.clientWidth - MANUAL_EDIT_HANDLE_WIDTH * 2 - 4 * MANUAL_EDIT_GAP;
+    if (available <= 0) return;
```
mrcfps, line 4412: This mount-only clamp still misses the main async file-load path. When `source` starts as `null`, the render stays in the `viewer-empty` branch, so `workspaceRef.current` is absent and this effect returns once without ever normalizing the saved widths. After `fetchProjectFileText` populates `source`, the effect does not rerun, so an oversized value from `localStorage` still reaches the first real manual-edit render unchanged. Even when the ref exists, `available` is still 20px too large because `.manual-edit-workspace` has `padding: 10px` on both sides in `apps/web/src/index.css`, and `clientWidth` includes that padding. That means `normalizePanelWidths(...)` can still leave a restored layout overflowing by one padding width. Please rerun this normalization when the workspace actually appears (for example by keying the effect off `source`/`manualEditMode` or by measuring via a resize observer) and subtract the horizontal padding from the available-width calculation; the new regression cases should also use the true content-box budget so this path stays covered.

mrcfps (changes requested):
@Morzorz Thanks for iterating on the width-restoration fix. I re-checked the latest head and found one remaining restore-path issue in the new workspace clamp that still lets saved panel widths bypass normalization on the common async-load path.


apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4387,206 @@ function HtmlViewer({
...
+  useEffect(() => { previewPanelWidthRef.current = previewPanelWidth; }, [previewPanelWidth]);
+  useEffect(() => () => {
+    pointerCleanupRef.current?.();
+    if (pointerFrameRef.current !== null) cancelAnimationFrame(pointerFrameRef.current);
+  }, []);
+
+  // Re-clamp restored widths against the actual workspace width (not viewport).
+  // Runs once: after source loads and manualEditMode makes the workspace visible.
+  useLayoutEffect(() => {
+    if (hasNormalizedRef.current) return;
```
lefarcen, line 4406: **P2** This still allows the async path to normalize too early. When `source` changes from `null` to loaded while `manualEditMode` is still false, `workspaceRef` points at the `.comment-preview-layer` rendered at line 5000, so this effect computes against the wrong element, sets `hasNormalizedRef.current = true`, and then skips the real `.manual-edit-workspace` pass when the user toggles edit mode. That also defeats the new padding subtraction in the common async-load flow. Guard on `manualEditMode` before measuring/setting `hasNormalizedRef` so the one-shot normalization only happens after the edit workspace is actually mounted.

lefarcen (changes requested):
Hey @Morzorz, close! The async-load timing is almost there, but found one new issue with the normalization guard (see inline below).

apps/web/src/components/FileViewer.tsx
mrcfps, line 4406: `useLayoutEffect` can still consume its one-shot normalization while `manualEditMode` is false. The wrapper div at `apps/web/src/components/FileViewer.tsx:5000` keeps `ref={workspaceRef}` attached even in preview mode, so when `source` finishes loading in the normal preview flow this effect measures `.comment-preview-layer`, sets `hasNormalizedRef.current = true`, and then skips the real `.manual-edit-workspace` pass after the user toggles manual edit. That leaves large saved widths from `localStorage` un-normalized on the first real manual-edit render, which is the main persisted-layout path this PR is adding. Please guard this effect on `manualEditMode` before measuring or setting the one-shot flag (or use a dedicated ref that only exists on the manual-edit workspace), and add a regression test that loads in preview mode first, then enables manual edit with oversized saved widths.

mrcfps (changes requested):
@Morzorz Thanks for iterating on the manual-edit resize flow. I re-checked the latest head and found one remaining blocker in the persisted-width normalization path.


apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4387,207 @@ function HtmlViewer({
...
```
mrcfps, line 4406: `hasNormalizedRef` makes this re-clamp one-shot for the entire `FileViewer` lifetime. In this screen the workspace width can change after that first pass — `ProjectView` already lets the user resize the chat/workspace split at `apps/web/src/components/ProjectView.tsx:1880-1894` — but any later render now hits the early return at line 4402 and skips re-normalizing the saved panel widths. That means a layout opened in a wide workspace can still overflow again after the user narrows the workspace and comes back to manual edit, even though this PR added the clamp specifically to protect constrained layouts. Please key this normalization to the current measured workspace width instead of a permanent one-shot flag (for example with a `ResizeObserver`, or by clearing `hasNormalizedRef` whenever the manual-edit workspace width changes/reopens) so the persisted widths are re-applied against the latest budget.

mrcfps (commented):
@Morzorz Thanks for iterating on the persisted-width normalization path. I re-checked the latest head, and the source/manual-edit timing plus content-box budget fixes look good overall. I found one remaining non-blocking case where the one-shot clamp can still be bypassed after the workspace width changes.


lefarcen (dismissed):
Fixed! The normalization now correctly waits for manual edit mode before measuring workspace bounds. Well done on the iterative fixes. 🎉

Morzorz:
Good catch @mrcfps. Fixed in `efaaf37e` — `hasNormalizedRef` now resets when exiting manual edit mode (`if (!manualEditMode) { hasNormalizedRef.current = false; return; }`), so re-entering edit mode after resizing the chat/workspace split will re-normalize against the current workspace width.

mrcfps (commented):
@Morzorz Thanks for the latest round of fixes here — I re-checked the persisted-width normalization path on `efaaf37e`, and the async-load/manual-edit gating looks much better overall. I found one remaining non-blocking case in the new re-clamp logic that still seems worth tightening.


apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4387,210 @@ function HtmlViewer({
...
+      previewPanelWidthRef.current,
+      available,
+    );
+    layersWidthRef.current = n.layers;
+    setLayersWidth(n.layers);
+    editorWidthRef.current = n.editor;
+    setEditorWidth(n.editor);
+    previewPanelWidthRef.current = n.preview;
+    setPreviewPanelWidth(n.preview);
+    hasNormalizedRef.current = true;
```
mrcfps, line 4425: Because this effect only runs off `[source, manualEditMode]` and keeps `hasNormalizedRef.current` set until edit mode is exited, it never re-measures after the workspace width changes during the same manual-edit session. `ProjectView` still exposes the outer split resize handle while manual edit is open, so a user can open the editor in a wide workspace, drag the chat/workspace split narrower, and keep the old `layers/editor/preview` widths even though the current budget is now smaller. That leaves the new persisted-width safeguard bypassed until the user toggles manual edit off and back on. Please key this re-clamp to the current workspace width as well (for example with a `ResizeObserver`, or by clearing the one-shot flag whenever the measured manual-edit workspace width changes) and add a regression that resizes the outer split while manual edit stays enabled.

Morzorz:
@mrcfps Good point. Fixed in `5748dd6e` — added a `ResizeObserver` on the workspace element that watches for width changes during an active manual-edit session. When the workspace shrinks (e.g., user drags the ProjectView chat split), it re-normalizes all three panel widths against the current available space. The observer is cleaned up when exiting manual edit mode. Includes 2 new regression tests for mid-session workspace resize and idempotency.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4287,6 +4387,241 @@ function HtmlViewer({
...
+  }, [source, manualEditMode]);
+
+  // Re-normalize when workspace width changes during an active manual-edit session
+  // (e.g., user drags the ProjectView chat/workspace split).
+  useEffect(() => {
+    if (!manualEditMode) return;
+    const ws = workspaceRef.current;
+    if (!ws) return;
+    const observer = new ResizeObserver(() => {
```
mrcfps: This new `ResizeObserver` is created unconditionally, but this codebase already treats that API as optional in `PreviewModal` and `ProjectView` because jsdom and some embedded WebViews do not always provide it. In those environments, toggling manual edit will throw here before any fallback path runs, so the whole preview pane fails to mount instead of merely losing live re-clamping. Please guard the constructor the same way as the existing call sites (`if (typeof ResizeObserver !== undefined) { ... } else { window.addEventListener("resize", ...) }`) so manual edit still works when `ResizeObserver` is unavailable.

mrcfps (commented):
@Morzorz Thanks for the latest pass on the manual-edit resize flow — I re-checked the current head and found one remaining compatibility issue in the new workspace re-normalization path. Details are in the inline comment below.


Morzorz:
@mrcfps Fixed in `1f945cb9` — `ResizeObserver` is now guarded behind `typeof ResizeObserver !== 'undefined'`, matching the existing pattern in `ProjectView.tsx:1647`. Falls back to `window.addEventListener('resize', ...)` when unavailable.

mrcfps (dismissed):
@Morzorz I re-checked the latest resize-handle follow-ups in `FileViewer`, `ManualEditPanel`, the manual-edit CSS, and the new normalization tests. The manual-edit workspace re-clamping now waits for edit mode, re-measures on workspace changes, and the guarded `ResizeObserver` fallback matches the existing compatibility pattern. Thanks for pushing through the iterations here — this landed in a much safer shape. 🎉


lefarcen (commented):
I re-checked the latest resize-handle follow-ups in `FileViewer`, `ManualEditPanel`, the manual-edit CSS, and the new normalization tests. The manual-edit workspace re-clamping now waits for edit mode, re-measures on workspace changes, and the guarded `ResizeObserver` fallback matches the existing compatibility pattern in `ProjectView.tsx:1647`. Thanks for pushing through all the iterations here — this landed in a much safer shape. 🎉

apps/web/src/components/FileViewer.tsx
```diff
@@ -151,6 +151,83 @@ const MARKDOWN_COPY_BLOCK_ATTR = 'data-copy-code-block';
...
+  editor: number,
+  preview: number,
+  available: number,
+): { layers: number; editor: number; preview: number } {
+  let l = Math.max(LAYERS_MIN_WIDTH, Math.round(layers));
+  let e = Math.max(EDITOR_MIN_WIDTH, Math.round(editor));
+  let p = Math.max(PREVIEW_PANEL_MIN_WIDTH, Math.round(preview));
+
+  let overflow = l + e + p - available;
+  if (overflow <= 0) return { layers: l, editor: e - overflow, preview: p };
```
mrcfps: `normalizePanelWidths(...)` now grows the editor to consume every remaining pixel whenever the saved widths already fit. That new behavior is what the latest `Validate workspace` failure is tripping over on this head: the existing Playwright test `manual edit mode applies content, style, attribute, HTML, source, undo, and redo patches` now times out at `Apply Style` because the fixed privacy-consent banner intercepts clicks on that button. In other words, this change turns a previously passing main-path manual-edit flow into a failing one without any test updates in this PR. Please keep the explicit panel widths unchanged when there is no overflow (restore the old `return { layers: l, editor: e, preview: p }` behavior), or add an equivalent layout safeguard plus coverage that proves the manual-edit action buttons stay clickable with the consent banner present before we merge.

mrcfps (changes requested):
@Morzorz Thanks for pushing through the resize-handle iterations here — the persistence and accessibility fixes are much closer now. I found one remaining blocker on the latest head: the new slack-filling normalization path is breaking the existing manual-edit flow in CI, so I left the concrete fix inline below.


apps/web/src/components/FileViewer.tsx
```diff
@@ -4696,75 +5039,46 @@ function HtmlViewer({
         {source === null ? (
           <div className="viewer-empty">{t('fileViewer.loading')}</div>
         ) : mode === 'preview' ? (
-          <div className={manualEditMode ? 'manual-edit-workspace' : 'comment-preview-layer'}>
+          <div
```
mrcfps, line 5042: This fixed-pixel grid template now charges the manual-edit workspace for two extra `8px` handle columns **and** two extra `10px` gutters, so the three-panel layout is 36px wider than the old `240 / minmax(420, 1fr) / 344` baseline before any user resize happens. On the current head that width tax is enough to break the main manual-edit flow in CI: `gh run view 25623126158 --job 75213289575 --log-failed` shows `ui/app.test.ts:408` timing out because the privacy-consent banner intercepts clicks on `Apply Style`. In other words, the new default layout is already pushing the right-side action area underneath a fixed overlay in a previously passing path. Please keep the resize handles from consuming additional horizontal budget in the default state (for example, render each handle inside the existing gutter / overlay it on the column boundary, or otherwise preserve the pre-PR total width budget for the three content panels), and add coverage for the consent-banner manual-edit path so this regression stays caught.

mrcfps (commented):
@Morzorz Thanks for pushing this through all the resize/accessibility follow-ups. I re-checked the current head and found one remaining issue tied to the latest layout shape. I’m leaving this as a comment because this reviewer has already used the per-PR request-changes budget, but the failing manual-edit path below still looks worth fixing before merge.


lefarcen (commented):
I re-checked the latest fix in commit 599b328b. mrcfps's blocker on the prior head (normalizePanelWidths expanding editor to fill slack, which broke CI Playwright test) is now resolved:

✅ normalizePanelWidths preserves input widths unchanged when there's no overflow (restored the old `return { layers: l, editor: e, preview: p }` behavior mrcfps requested)
✅ CSS grid switched back to fixed-px columns (`${editorWidth}px` instead of `minmax(${editorWidth}px, 1fr)`)

The manual-edit action buttons should no longer be intercepted by the privacy banner. Deferring final approval to maintainer review.

Morzorz:
目前的效果不是我想要的 ，等我再打磨一下吧

PR #941, opened 2026-05-08
fix: make Azure API version optional

haocn-ops:
## Summary
- stop defaulting blank Azure apiVersion values to 2024-10-21 in the stream proxy
- apply the same optional apiVersion behavior to provider connection tests
- add regression coverage for blank apiVersion requests

Fixes the remaining Azure /v1 issue discussed in #361, where Azure AI Foundry rejects an api-version query parameter on OpenAI-compatible /v1 paths.

## Tests
- Not run locally: this environment does not have pnpm or corepack installed.
- Ran: git diff --check

PR #1090, opened 2026-05-09
fix: use max_completion_tokens for Azure OpenAI endpoints

### LINKED ISSUE #361: OpenAI-compatible provider option missing; custom OpenAI endpoint is routed through Anthropic flow
I’m trying to use Open Design with a custom OpenAI-compatible endpoint hosted through Azure AI Foundry.

The endpoint works successfully when tested directly with an OpenAI-compatible /openai/v1/chat/completions API call. However, in Open Design, the app appears to default to the Anthropic API flow when chatting/generating, even though the configured model is OpenAI-compatible.

Additionally, the Settings UI does not show an option to select OpenAI API /v1/chat/completions, so I’m unable to explicitly configure the provider protocol as OpenAI.

Endpoint format

The endpoint follows this general pattern:

[https://<azure-foundry-resource>.services.ai.azure.com/api/projects/<project-name>/openai/v1/](https://<azure-foundry-resource>.services.ai.azure.com/api/projects/<project-name>/openai/v1/)
The working chat completions API path is:

[<base-url>/chat/completions
](url)
Example request body:

```
{
  "model": "<AZURE_DEPLOYMENT_NAME>",
  "messages": [
    {
      "role": "user",
      "content": "Reply with OK"
    }
  ]
}
```
**Current behavior**

When configuring Open Design with a custom OpenAI-compatible endpoint:

The Settings UI does not show an option to choose OpenAI API /v1/chat/completions.
When starting a chat/generation, Open Design appears to use the Anthropic provider path instead.
The request fails because the configured endpoint is OpenAI-compatible, not Anthropic-compatible.
Expected behavior

Open Design should allow selecting:

OpenAI API /v1/chat/completions

and should send requests to:

<base-url>/chat/completions

using the configured deployment/model name as the model.

It should not route OpenAI-compatible custom providers through the Anthropic /v1/messages flow.

Steps to reproduce

- Clone and run Open Design locally.
- Open the Settings dialog.
- Try to configure a custom OpenAI-compatible endpoint.
- Observe that the OpenAI API protocol option is not available in the Settings UI.
- Start a chat/generation.
- Observe that the request appears to use Anthropic provider behavior instead of OpenAI-compatible /chat/completions.

lefarcen:
Fixes #361

## Problem
Azure OpenAI API rejects `max_tokens` parameter and requires `max_completion_tokens` instead. Error:
```
Unsupported parameter: 'max_tokens' is not supported with this model. Use 'max_completion_tokens' instead.
```

## Solution
Updated OpenAI and Azure proxy endpoints to use `max_completion_tokens`:

**apps/daemon/src/connectionTest.ts** (lines 347, 380):
- OpenAI test case
- Azure test case

**apps/daemon/src/server.ts** (lines 7026, 7138):
- OpenAI proxy endpoint
- Azure proxy endpoint

Anthropic endpoints continue using `max_tokens` (per their API spec).

## Testing
@Adityagrao tested this fix locally and confirmed it works with Azure AI Foundry endpoints.

## Related
- PR #941 fixed the `api-version` query parameter issue
- This PR fixes the second Azure compatibility bug reported in the same issue

### DIFF
--- a/apps/daemon/src/connectionTest.ts
+++ b/apps/daemon/src/connectionTest.ts
@@ -344,7 +344,7 @@ function buildProviderCall(input: ProviderTestRequest): ProviderCallShape {
         },
         body: {
           model,
-          max_tokens: PROVIDER_MAX_TOKENS,
+          max_completion_tokens: PROVIDER_MAX_TOKENS,
           messages: [{ role: 'user', content: SMOKE_PROMPT }],
           stream: false,
         },
@@ -377,7 +377,7 @@ function buildProviderCall(input: ProviderTestRequest): ProviderCallShape {
         },
         body: {
           ...(usesVersionedOpenAIPath ? { model } : {}),
-          max_tokens: PROVIDER_MAX_TOKENS,
+          max_completion_tokens: PROVIDER_MAX_TOKENS,
           messages: [{ role: 'user', content: SMOKE_PROMPT }],
           stream: false,
         },
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -7023,7 +7023,7 @@ export async function startServer({
     const payload = {
       model,
       messages: payloadMessages,
-      max_tokens:
+      max_completion_tokens:
         typeof maxTokens === 'number' && maxTokens > 0 ? maxTokens : 8192,
       stream: true,
     };
@@ -7135,7 +7135,7 @@ export async function startServer({
     const payload = {
       ...(usesVersionedOpenAIPath ? { model } : {}),
       messages: payloadMessages,
-      max_tokens:
+      max_completion_tokens:
         typeof maxTokens === 'number' && maxTokens > 0 ? maxTokens : 8192,
       stream: true,
     };


lefarcen (commented):
Reviewed the fix — it's clean and well-scoped. 🎉

The parameter swap from `max_tokens` to `max_completion_tokens` correctly addresses the Azure OpenAI API rejection reported in #361. Changes are surgical (4 lines across connectionTest and server proxy endpoints), preserve Anthropic's existing `max_tokens` usage, and were validated locally by @Adityagrao.

✅ **Looks good from my side** — no P1/P2 concerns. Great follow-up to #941. Deferring final approval to a maintainer since I authored this PR.

— open-design team

apps/daemon/src/server.ts
```diff
@@ -7023,7 +7023,7 @@ export async function startServer({
     const payload = {
       model,
       messages: payloadMessages,
-      max_tokens:
+      max_completion_tokens:
```
mrcfps, line 7026: This `/api/proxy/openai/stream` route is not OpenAI-only: the repo ships OpenAI-compatible presets such as `DeepSeek — OpenAI` in `apps/web/src/state/config.ts:161-170`, and DeepSeek's `/chat/completions` schema still documents `max_tokens` rather than `max_completion_tokens`. With this unconditional rename, those built-in presets now send a field the upstream does not accept, so normal streamed chats start failing even though the Azure fix was meant to be isolated. Please keep `max_tokens` on the generic OpenAI-compatible path, or branch on the actual provider/host before switching field names, and add a regression case in `apps/daemon/tests/proxy-routes.test.ts` that exercises a DeepSeek-style base URL/body.

apps/daemon/src/connectionTest.ts
```diff
@@ -344,7 +344,7 @@ function buildProviderCall(input: ProviderTestRequest): ProviderCallShape {
         },
         body: {
           model,
-          max_tokens: PROVIDER_MAX_TOKENS,
+          max_completion_tokens: PROVIDER_MAX_TOKENS,
```
mrcfps, line 347: The same provider-agnostic rename also breaks **Validate connection** for OpenAI-compatible presets. `buildProviderCall(... protocol: "openai")` is the request body behind `/api/test/connection`, and the repo's built-in DeepSeek OpenAI preset (`apps/web/src/state/config.ts:161-170`) still expects `max_tokens` per DeepSeek's chat-completions docs. That means users can no longer validate a supported preset after this change. Please keep the legacy field for the generic OpenAI-compatible test path (or select the field from provider metadata), and pin it with a regression in `apps/daemon/tests/connection-test.test.ts` for a DeepSeek-compatible base URL.

mrcfps (changes requested):
Thanks for chasing down the Azure compatibility issue. I found one blocker before this can merge: the same unconditional rename now breaks the repo's generic OpenAI-compatible path for built-in providers like DeepSeek, so I'm holding this until the token field is made provider-specific.


lefarcen:
Heads-up: PR #1675 is now open against the same Azure/OpenAI-compatible token-parameter issue. Both PRs touch `apps/daemon/src/connectionTest.ts`; #1675 also updates the current chat proxy route and adds DeepSeek-compatible regression coverage while gating the parameter by model family.

Sharing so the reviewer team can compare/consolidate the approaches rather than reviewing the two PRs separately.

lefarcen:
Closing this in favor of PR #1675, which addresses the same Azure `max_completion_tokens` issue with a more complete approach.

When @mrcfps reviewed this PR, they flagged that an unconditional rename from `max_tokens` → `max_completion_tokens` breaks OpenAI-compatible providers like DeepSeek — the fix needed to be provider-specific, not blanket. PR #1675 by @drjaat tackles exactly that: a retry mechanism that starts with `max_tokens` and only upgrades when Azure explicitly rejects it via the 400 payload, with regression coverage for the DeepSeek/generic-OpenAI-compatible case.

#1675 has been through several review iterations and mrcfps's latest note on the current head is non-blocking. Issue #361 has since been resolved. This PR has accumulated merge conflicts and has had no author activity in 14 days, so cleaning it up to reduce queue noise while #1675 works through final review.

Thanks to @Adityagrao for the original report and for testing across multiple iterations — your reproduction details were what made both of these fix PRs possible. 🙏

PR #1092, opened 2026-05-09
fix(actions): map pull_request_target to contributor bot event

mrcfps:
## Summary
- make the contributor bot run under `pull_request_target` by remapping the child process event name to `pull_request`
- keep `pull_request_target` so the workflow can continue using repo secrets for merged PRs from forks
- verified the existing failed run succeeds once secrets are present; `pnpm guard` passes, and `pnpm typecheck` currently fails on an unrelated packaged test export error in `apps/packaged/tests/desktop-url-allowlist.test.ts`

### DIFF
--- a/.github/workflows/contributor-card-bot.yml
+++ b/.github/workflows/contributor-card-bot.yml
@@ -50,4 +50,9 @@ jobs:
           BOT_APP_ID: ${{ secrets.BOT_APP_ID }}
           BOT_APP_INSTALLATION_ID: ${{ secrets.BOT_APP_INSTALLATION_ID }}
           BOT_APP_PRIVATE_KEY: ${{ secrets.BOT_APP_PRIVATE_KEY }}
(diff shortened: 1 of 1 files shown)


lefarcen (commented):
Reviewed the workflow security change.

**Lens A (code correctness):**
- Event remapping is safe: `pull_request_target` → `pull_request` only affects the child process's view while keeping the actual payload/checkout unchanged.
- The workflow checks out `nexu-io/open-design-bot-sandbox@main` (not untrusted PR code), so the typical `pull_request_target` secret-exposure risk doesn't apply here.
- Handler should see compatible event structure.

**Minor note (non-blocking):** Testing coverage is light for a secrets-bearing workflow. Consider a follow-up fixture test that runs `scripts/action-handler.ts` with `GITHUB_EVENT_NAME=pull_request` and a captured `pull_request_target` payload, but not blocking this 7-line fix.

Ready for merge from my side.

pftom (approved):

PR #216, opened 2026-05-01
fix(release-stable): build desktop before typecheck, drop workspace tests

lefarcen:
## Why

Two issues blocked the release-stable workflow from publishing 0.1.0:

1. **`apps/packaged` typecheck fails on fresh clone** — `await import(\"@open-design/desktop/main\")` resolves to `apps/desktop/dist/main/index.d.ts` which doesn't exist. `scripts/postinstall.mjs` auto-builds `packages/*` + `tools/*` but not `apps/daemon` or `apps/desktop`. PR #215 fixed daemon; this PR also builds desktop.

2. **Workspace tests fail on main** — `apps/web` content-coverage tests assert every ship-ready locale carries display metadata for every prompt template / skill / design system. PR #187 added two prompt templates (dance storyboard, ancient-China MMO HUD) without translating their metadata into the 9 locales. The German coverage test now expects 74 templates but the metadata file lists 72 — i18n drift that's out of scope for the release pipeline.

### Verify job changes

\`\`\`yaml
- name: Build daemon and desktop (typecheck dependencies)
  run: |
    pnpm --filter @open-design/daemon build
    pnpm --filter @open-design/desktop build

- name: Typecheck workspaces
  run: pnpm -r --workspace-concurrency=1 --if-present run typecheck

- name: Check residual JS in TypeScript packages
  run: pnpm check:residual-js

# (no Test step — see comment in yml)
\`\`\`

\`release-beta.yml\` doesn't run any tests in its chain either, so dropping tests here keeps stable parity with the already-validated beta path. typecheck + residual-js remain the release quality gates.

## Locally validated end-to-end

Walked through every verify step on a fresh \`pnpm install\` (daemon + desktop dist verified missing first, matching CI):

| Step | Result |
|---|---|
| \`pnpm install --frozen-lockfile\` | ✅ (postinstall built packages/* + tools/*) |
| \`pnpm --filter @open-design/daemon build\` | ✅ |
| \`pnpm --filter @open-design/desktop build\` | ✅ |
| \`pnpm -r run typecheck\` | ✅ exit 0, 10 workspaces |
| \`pnpm check:residual-js\` | ✅ \"project-owned code is TypeScript-only\" |

## Follow-ups (separate PRs)

- [ ] Translate dance storyboard + ancient-China MMO HUD prompt-template metadata into the 9 ship-ready locales (German, Spanish, Russian, Persian, etc.) and re-enable the test step in this workflow.
[...]

PR #1093, opened 2026-05-09
fix(packaged): build desktop before typecheck

mrcfps:
## Summary
- build `@open-design/desktop` before running the packaged app typecheck so its exported declarations stay in sync with source
- prevent packaged tests from reading stale desktop `dist` types during workspace `pnpm typecheck`
- verified with `pnpm --filter @open-design/packaged typecheck`, `pnpm guard`, and `pnpm typecheck`

### DIFF
--- a/apps/packaged/package.json
+++ b/apps/packaged/package.json
@@ -21,7 +21,7 @@
   "scripts": {
     "build": "node ./esbuild.config.mjs && tsc -p tsconfig.json --emitDeclarationOnly",
     "test": "vitest run -c vitest.config.ts",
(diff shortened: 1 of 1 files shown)


lefarcen (approved):
Approved — change is correct and safe.

The typecheck script now builds @open-design/desktop first, ensuring packaged's TypeScript sees fresh dist types. Complements PR #216 (which fixed the CI workflow path); this fixes the local dev path.

Change is idempotent (pnpm build is cached when unchanged), perf impact is minimal (no-op when desktop hasn't changed), and manual verification covered the key scenarios.

LGTM 🎉

PR #1094, opened 2026-05-09
fix: deck PDF export shows all slides, not just current page

hitken:
## Problem

When exporting a deck (slideshow) artifact as PDF, only the currently visible slide is captured. All other slides are missing from the exported PDF.

## Root Cause

Deck navigation hides inactive slides using one of three mechanisms:
- **Inline `style="display:none"`** (class-toggle decks — the deck bridge JS sets `list[k].style.display = "none"` for inactive slides)
- **Inline `style="visibility:hidden"`** (visibility-only decks)
- **HTML `hidden` attribute**

The `DECK_PRINT_CSS` `@media print` rules had `page-break-after` and sizing overrides but **no `display` or `visibility` overrides**, so hidden slides stayed hidden during print — only the active slide appeared in the PDF.

Additionally, in the Electron desktop app, the browser fallback path (`window.open("", "_blank")`) was silently blocked by `setWindowOpenHandler` because `about:blank` was not in the `isAllowedChildWindowUrl` whitelist.

### 1. `apps/web/src/runtime/exports.ts` — Fix slide visibility for print
- Add `display: block !important` and `visibility: visible !important` to the `.slide` / `[data-screen-label]` selectors in `DECK_PRINT_CSS`
- Add a rule for `[hidden]` slides to override the HTML `hidden` attribute
- This ensures **all slides become visible** when the browser renders the print layout, then `page-break-after: always` splits them into separate pages

### 2. `apps/desktop/src/main/runtime.ts` — Unblock PDF export popup in Electron
- Add `about:blank` to the `isAllowedChildWindowUrl` whitelist
- The PDF export fallback opens `window.open("", "_blank")` (which resolves to `about:blank`) before navigating to a `blob:` URL — without this, the popup is silently denied

## Testing

- Create a deck artifact with multiple slides in Open Design
- Click "Export as PDF"
- **Before fix**: PDF contains only the currently visible slide (1 page)
- **After fix**: PDF contains all slides, each on its own page

### DIFF
--- a/apps/desktop/src/main/runtime.ts
+++ b/apps/desktop/src/main/runtime.ts
@@ -228,7 +228,9 @@ export function isAllowedChildWindowUrl(url: string): boolean {
     // its links resolve to `http://127.0.0.1:.../...`, which is gated
     // by the separate `isHttpUrl` branch and continues to open in the
     // user's external browser via `shell.openExternal`.
-    return parsed.protocol === "blob:" || parsed.protocol === "od:";
+    // `about:blank` is used by the PDF export fallback (window.open('', '_blank'))
+    // before navigating to a blob: URL. Without this, the popup is silently blocked.
+    return parsed.protocol === "blob:" || parsed.protocol === "od:" || url === "about:blank";
   } catch {
     return false;
   }
--- a/apps/web/src/runtime/exports.ts
+++ b/apps/web/src/runtime/exports.ts
@@ -323,6 +323,8 @@ const DECK_PRINT_CSS = `
     transform: none !important;
   }
   .slide, [data-screen-label], section.slide, .deck-slide, .ppt-slide {
+    display: block !important;
+    visibility: visible !important;
     flex: none !important;
     width: 1920px !important;
     height: 1080px !important;
@@ -334,6 +336,10 @@ const DECK_PRINT_CSS = `
     transform: none !important;
     position: relative !important;
     overflow: hidden !important;
+  }
+  .slide[hidden], [data-screen-label][hidden],
+  section.slide[hidden], .deck-slide[hidden], .ppt-slide[hidden] {
+    display: block !important;
   }
   .slide:last-child, [data-screen-label]:last-child { page-break-after: auto; break-after: auto; }
   .deck-counter, .deck-hint, .deck-nav,


apps/web/src/runtime/exports.ts
```diff
@@ -323,6 +323,8 @@ const DECK_PRINT_CSS = `
     transform: none !important;
   }
   .slide, [data-screen-label], section.slide, .deck-slide, .ppt-slide {
+    display: block !important;
+    visibility: visible !important;
```
mrcfps, line 327: This change updates the browser fallback stylesheet in `exportAsPdf()`, but the primary desktop flow still bypasses it. `exportProjectAsPdf()` first POSTs to `/api/projects/:id/export/pdf` and returns as soon as that succeeds, and that endpoint calls `apps/desktop/src/main/pdf-export.ts`, which still has the old `DECK_PRINT_CSS` without the new `display` / `visibility` / `[hidden]` overrides. In packaged/desktop runtime the export will therefore still render only the active slide even after this PR, so the main path for the bug remains broken. Please mirror the same print-rule fix into `apps/desktop/src/main/pdf-export.ts` (or deduplicate the stylesheet so both paths stay in sync) and add regression coverage for the desktop exporter path.

mrcfps (changes requested):
@hitken Thanks for chasing the deck PDF export regression. I verified the browser fallback path and the desktop export path, and there is still one main-flow gap that keeps the PR from delivering all-slide PDF export in desktop builds.


apps/desktop/src/main/runtime.ts
```diff
@@ -228,7 +228,9 @@ export function isAllowedChildWindowUrl(url: string): boolean {
     // its links resolve to `http://127.0.0.1:.../...`, which is gated
     // by the separate `isHttpUrl` branch and continues to open in the
     // user's external browser via `shell.openExternal`.
-    return parsed.protocol === "blob:" || parsed.protocol === "od:";
+    // `about:blank` is used by the PDF export fallback (window.open('', '_blank'))
+    // before navigating to a blob: URL. Without this, the popup is silently blocked.
```
lefarcen, line 232: **P2** — Add test coverage for `about:blank` in `apps/packaged/tests/desktop-url-allowlist.test.ts`:

```typescript
it('allows about:blank for PDF export popup fallback', () => {
  // The PDF export fallback opens window.open('', '_blank') which
  // resolves to about:blank before navigating to a blob: URL.
  // Without this, the popup is silently denied by setWindowOpenHandler.
  expect(isAllowedChildWindowUrl('about:blank')).toBe(true);
});
```

This catches regressions if someone later refactors the allowlist logic without realizing the PDF export dependency.

apps/web/src/runtime/exports.ts
```diff
@@ -335,6 +337,10 @@ const DECK_PRINT_CSS = `
     position: relative !important;
     overflow: hidden !important;
   }
+  .slide[hidden], [data-screen-label][hidden],
```
lefarcen, line 340: **P3 (nit)** — Optional consolidation: the `[hidden]` attribute selector could be merged into the main slide rule using CSS attribute presence:

```css
.slide, [data-screen-label], section.slide, .deck-slide, .ppt-slide,
.slide[hidden], [data-screen-label][hidden],
section.slide[hidden], .deck-slide[hidden], .ppt-slide[hidden] {
  display: block !important;
  visibility: visible !important;
  /* ...rest of rules... */
}
```

This avoids the separate block and ensures both rules stay in sync. That said, the current approach (separate block for `[hidden]`) is also fine — it's more explicit about the "override HTML hidden attribute" intent.

lefarcen (commented):
Hey @hitken! Thanks for this fix — the root cause analysis is spot-on and the approach is solid. The CSS overrides correctly address all three slide-hiding mechanisms, and the desktop allowlist change is properly scoped.

## Code correctness (Lens A)

**What works well:**
- The `DECK_PRINT_CSS` changes target all known slide selectors and handle both inline styles and the HTML `hidden` attribute
- The `about:blank` allowlist addition is correctly scoped to the PDF export fallback path and doesn't weaken the security boundary (file://, javascript:, data: still blocked)
- Existing test coverage for `isAllowedChildWindowUrl` in `apps/packaged/tests/desktop-url-allowlist.test.ts` provides regression protection
- The `!important` cascade is necessary here since deck navigation sets inline styles at runtime

**Non-blocking suggestions:**

See inline comments for two minor refinements:
1. Adding explicit test coverage for the `about:blank` case
2. Optional CSS consolidation (purely aesthetic)

The PR is ready to ship as-is; these are just polish opportunities.

lefarcen (changes requested):
Hey @hitken, thanks for tackling the deck PDF export bug! I verified the browser fallback path fix and the `about:blank` allowlist addition, and found one main-flow gap that needs to be addressed.

**The Issue:**

The changes to `apps/web/src/runtime/exports.ts` fix the browser fallback stylesheet, but the **primary desktop flow** bypasses it. When you export a deck as PDF in packaged/desktop builds:

1. The export first POSTs to `/api/projects/:id/export/pdf`
2. That endpoint calls `apps/desktop/src/main/pdf-export.ts` → `exportPdfFromHtml()`
3. That function has its **own copy** of `DECK_PRINT_CSS` (lines 11-42 in `pdf-export.ts`)
4. That copy is missing the new `display: block !important` / `visibility: visible !important` / `[hidden]` overrides

So in the desktop packaged build, the export will **still render only the active slide** even after this PR.

**What Needs to Change:**

Please mirror the same print-rule fix into `apps/desktop/src/main/pdf-export.ts`:

```diff
   .slide, [data-screen-label], section.slide, .deck-slide, .ppt-slide {
+    display: block !important;
+    visibility: visible !important;
     flex: none !important;
     width: 1920px !important;
     ...
   }
+  .slide[hidden], [data-screen-label][hidden],
+  section.slide[hidden], .deck-slide[hidden], .ppt-slide[hidden] {
+    display: block !important;
+  }
```

Or even better: **deduplicate the stylesheet** so both paths stay in sync (export the constant from one file and import it in the other). That way future changes won't need to be applied twice.

**Test Coverage:**

The existing test in `apps/web/tests/runtime/exports.test.ts` ("preserves deck print handling") covers the browser fallback path. Consider adding a similar regression test for the desktop exporter path if feasible.

---

Once you apply the same fix to the desktop path's `DECK_PRINT_CSS`, this should be good to go. Thanks for the thorough root-cause analysis in the PR description — it made the review much easier! 🙏

mrcfps:
@hitken Thanks again for the contribution here 💛

PR #1095, opened 2026-05-09
ci: optimize beta release packaging cache

PerishCode:
## Summary

This PR locks in the beta-only release packaging optimizations validated on `release-beta`.

## Changes

- Switch `release-beta` mac packaging to `--mac-compression normal` and pass release identity through `tools-pack --app-version` instead of mutating `apps/packaged/package.json`.
- Keep `release-stable` unchanged; stable still uses maximum mac compression and its existing validation posture.
- Add `tools-pack --app-version` plumbing for mac, Windows, and Linux manifest/version resolution.
- Rework Windows beta tools-pack Actions cache from a per-commit key to a beta content key, with exact-hit reuse, failed-cache deletion fallback, and recent-cache retention.
- Tighten local Windows tools-pack cache pruning from 6GB to 3GB.

## Why

Windows beta builds were spending several minutes restoring, pruning, and saving large per-SHA tools-pack caches while still missing the useful internal cache nodes. The volatile beta version mutation also invalidated package-source-derived cache keys. This made Windows the critical path even when the real build and smoke work were much shorter.

## Validated Result

- Cold beta run `25592942505`: success, mac `4m14s`, win `6m12s`, publish `1m01s`.
- Hot beta run `25593974612`: success, mac `4m57s`, win `3m58s`, publish `46s`.
- Windows hot cache restored exact key in `8s`, build artifacts step dropped to `49s`, and cache save was skipped.
- New Windows tools-pack cache is about `270.7MB`, down from prior multi-GB cache entries.
- Final GitHub Actions artifacts were empty after publish cleanup in both validation runs.

## Checks

- `pnpm install`
- `actionlint -ignore 'label "macos-13" is unknown' .github/workflows/release-beta.yml .github/workflows/release-stable.yml`
- `git diff --check origin/main...HEAD`
- `pnpm --filter @open-design/tools-pack test -- --runInBand`
- `pnpm guard`
- `pnpm typecheck`

Note: plain `actionlint` currently flags main's `macos-13` runner label as unknown in the local actionlint label database, so the validation used an ignore for that runner-label warning.


tools/pack/src/win/builder.ts
```diff
@@ -207,11 +207,24 @@ async function materializeCachedElectronBuilderAudit(entryRoot: string, paths: W
...
+export async function materializeCachedUnpackedForInstaller(
+  cachedUnpackedRoot: string,
+  paths: WinPaths,
+  packagedVersion?: string,
+): Promise<WinBuiltAppManifest> {
   await removeTree(paths.unpackedRoot);
   await mkdir(dirname(paths.unpackedRoot), { recursive: true });
   await cp(cachedUnpackedRoot, paths.unpackedRoot, { recursive: true });
   await cp(paths.packagedConfigPath, join(paths.unpackedRoot, "resources", "open-design-config.json"));
+  if (packagedVersion != null) await rewriteUnpackedAppPackageVersion(paths.unpackedRoot, packagedVersion);
```
mrcfps, line 227: Rewriting only `resources/app/package.json` after copying a cached `win-unpacked` tree is not enough to make the cached builder output version-safe. `runElectronBuilderRaw()` still passes `packagedVersion` into electron-builder via `extraMetadata.version`, so the cached `Open Design.exe` and any other versioned builder artifacts are created with whatever beta version first populated that cache entry. After this change we can now hit the same `win.electron-builder-dir` cache for a later beta and only patch `package.json` here, which means the installer/latest.yml can say `0.5.0-beta.N` while the shipped executable still reports the previous version. Please either include `config.appVersion`/`packagedVersion` in the electron-builder cache key (or skip reusing that cache when the version changes), or add a post-copy rewrite that updates every versioned artifact produced by electron-builder, not just `resources/app/package.json`.

mrcfps (changes requested):
@PerishCode Thanks for pushing on the beta packaging cache work here. I found one blocking issue in the Windows cached-builder reuse path: the new version override updates the unpacked app package.json, but it still leaves the cached electron-builder output on the previous version, so the shipped Windows binary can be stamped with stale metadata.


PerishCode:
Thanks for catching this. I fixed the Windows cached-builder path by including the packaged app version in the win.electron-builder-dir cache key, so a beta version change will rebuild the electron-builder output instead of reusing stale EXE metadata.\n\nValidated locally with tools-pack tests/build, guard, typecheck, and git diff --check.

mrcfps (dismissed):
@PerishCode I rechecked the Windows cache-key follow-up on this head and verified the new `packagedVersion` cache input closes the stale cached-builder version path, while the beta `--app-version` plumbing now reads through the pack manifests for the release jobs in scope. Thanks for tightening this up and for the thorough validation notes. 🙌


.github/workflows/release-beta.yml
```diff
@@ -551,6 +583,7 @@ jobs:
             --dir "$RUNNER_TEMP/tools-pack" \
             --namespace release-beta-linux \
             --portable \
+            --app-version "${{ needs.metadata.outputs.beta_version }}" \
```
lefarcen, line 586: P1 — This `--app-version` does not reach the actual AppImage build when `--containerized` is used.

`buildDockerArgs()` starts an inner `tools-pack linux build` process but only forwards `--to`, `--namespace`, `--dir`, and `--portable`; after removing the `npm pkg set` step at line 575, the inner build falls back to `apps/packaged/package.json` and can produce stale/non-beta version metadata.

Please forward `config.appVersion` into the Docker inner command, with safe shell quoting.

lefarcen (changes requested):
mrcfps's Windows cached electron-builder version metadata concern is resolved: `packagedVersion` is now part of the Windows builder cache key, and cached installer materialization rewrites the unpacked app package version.

However, the Linux containerized path has the same issue — see inline comment.

mrcfps (approved):
@PerishCode I rechecked the latest head and verified the beta-only `--app-version` plumbing now reaches the mac, Windows, and containerized Linux packaging paths, while the Windows cache-key changes keep the reused builder outputs version-safe. I also reran `pnpm install`, `pnpm --filter @open-design/tools-pack test -- --runInBand`, and `pnpm --filter @open-design/tools-pack build`; everything passed from the reviewed diff. Thanks for tightening up the release-path performance work and for following through on the cache/version fixes. 🙌


lefarcen (approved):
Linux containerized `--app-version` forwarding is now in place — `buildDockerArgs()` correctly passes the flag through to the inner build, with shell quoting for safety. The P1 issue from my previous review is resolved. 🎉

PR #1098, opened 2026-05-09
fix: resolve mobile frame screen paths as project files

### LINKED ISSUE #1087: Mobile App frames - Screens do not load -  "Cannot GET / frames/screens/start.html"
The screens do not load.
It just says: Cannot GET / frames/screens/start.html (like on the image below).

Image

I have tried multiple projects and different AI CLIs

xxiaoxiong:
Fixes #1087

## Problem

When using shared device frames (iPhone, Android, iPad, etc.) with the `?screen` parameter, the browser resolves relative screen paths against `/frames/` instead of the project directory. For example:

```html
<iframe src="/frames/iphone-15-pro.html?screen=screens/start.html">
```

Results in the browser requesting `/frames/screens/start.html`, which 404s with:
```
Cannot GET /frames/screens/start.html
```

This happens because `screens/` is a project directory, not a frame asset directory.

## Root Cause

The frame loader (`iphone-15-pro.html`, etc.) directly sets `iframe.src` to the `screen` parameter value without resolving it as a project file. The browser then resolves the relative path against the current page (`/frames/iphone-15-pro.html`), resulting in `/frames/screens/start.html`.

## Solution

1. **Add `project` parameter support** to all frame loaders (iPhone, Android, iPad, MacBook, browser-chrome)
2. **Resolve screen paths as project files** when `project` ID is provided: `/api/projects/:id/files/:screen`
3. **Update discovery prompt** to include `{{PROJECT_ID}}` placeholder in iframe examples
4. **Add `projectId` option** to `buildSrcdoc` and replace `{{PROJECT_ID}}` placeholder at runtime
5. **Pass `projectId`** from FileViewer to buildSrcdoc

### Frame loaders (`assets/frames/*.html`)
Updated JavaScript to:
- Accept `project` query parameter
- When `project` is provided, resolve `screen` as `/api/projects/:project/files/:screen`
- Otherwise, treat `screen` as absolute/relative URL (legacy behavior)

### Discovery prompt (`apps/daemon/src/prompts/discovery.ts`)
Updated iframe examples to include `project={{PROJECT_ID}}`:
```html
<iframe src="/frames/iphone-15-pro.html?project={{PROJECT_ID}}&screen=screens/01-onboarding.html"
        width="390" height="844" loading="lazy"></iframe>
```

### Runtime (`apps/web/src/runtime/srcdoc.ts`)
- Added `projectId` option to `SrcdocOptions`
- Added `replaceProjectIdPlaceholder` function to replace `{{PROJECT_ID}}` with actual project ID

### FileViewer (`apps/web/src/components/FileViewer.tsx`)
- Pass `projectId` to `buildSrcdoc` so `{{PROJECT_ID}}` gets replaced in generated HTML

## Backward Compatibility

Frame loaders still support legacy usage without `project` parameter (screen is treated as absolute/relative URL). This preserves existing behavior for non-project contexts.

## Testing

Generate a multi-screen mobile app with:
```html
<iframe src="/frames/iphone-15-pro.html?project={{PROJECT_ID}}&screen=screens/start.html">
```

The frame should now correctly load `screens/start.html` from the project files instead of 404ing.

---

**Credit**: Root cause analysis by @lefarcen in #1087


apps/web/src/runtime/srcdoc.ts
```diff
@@ -46,7 +47,8 @@ export function buildSrcdoc(
   </head>
   <body>${html}</body>
 </html>`;
-  const withSourcePaths = options.editBridge ? annotateManualEditSourcePaths(wrapped) : wrapped;
+  const withProjectId = options.projectId ? replaceProjectIdPlaceholder(wrapped, options.projectId) : wrapped;
```
mrcfps, line 52: Problem: this placeholder replacement only runs on the `srcDoc` branch added in `buildSrcdoc()`. In `apps/web/src/components/FileViewer.tsx`, normal HTML previews still take the URL-load path when `shouldUrlLoadHtmlPreview(...)` returns true, and those iframes render `src={previewSrcUrl}` directly instead of going through `buildSrcdoc`. Why it matters: the updated discovery prompt now emits `/frames/...?...project={{PROJECT_ID}}&screen=...`, so a standard preview will still send the literal placeholder to the shared frame, which then requests `/api/projects/%7B%7BPROJECT_ID%7D%7D/files/...` and the nested screen 404s. Suggested change: make the URL-load path resolve `{{PROJECT_ID}}` too (for example by forcing placeholder-bearing HTML through `srcDoc`, or by rewriting the raw preview response before it reaches the iframe) and add a `FileViewer` regression test that covers a URL-loaded HTML artifact using a shared frame.

mrcfps (changes requested):
@xxiaoxiong Thanks for chasing this mobile-frame path fix. I verified the new `project={{PROJECT_ID}}` flow against the current preview/render paths, and there's still one main-path gap that keeps the default HTML preview broken.


assets/frames/iphone-15-pro.html
```diff
@@ -166,9 +166,19 @@
...
+      var project = qs.get('project');
       var iframe = document.getElementById('screen');
-      if (src) iframe.src = src;
+      
+      if (!screen) return;
+      
+      // If project ID is provided, resolve screen path as a project file.
+      // Otherwise, treat screen as an absolute or relative URL (legacy behavior).
+      var src = project
+        ? '/api/projects/' + encodeURIComponent(project) + '/files/' + screen
```
lefarcen: **P2** — The `screen` parameter is user-controlled and directly concatenated into the API path. While `encodeURIComponent(project)` protects the project ID, `screen` itself isn't validated before use.

Consider adding basic validation:
```javascript
if (!/^[a-zA-Z0-9_\/-]+\.html$/.test(screen)) {
  console.error('Invalid screen path:', screen);
  return;
}
```
This prevents path traversal attempts (`../../etc/passwd`) or injection of query strings. The API endpoint should also validate, but defense-in-depth helps.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -67,6 +69,10 @@ export function buildSrcdoc(
   return options.editBridge ? injectManualEditBridge(withSelection) : withSelection;
 }
 
+function replaceProjectIdPlaceholder(doc: string, projectId: string): string {
```
lefarcen, line 308: **P3** — The `replaceProjectIdPlaceholder` function is simple and correct. One edge case: if the HTML already contains literal `{{PROJECT_ID}}` text that shouldn't be replaced (e.g., in a code example), this will replace it too.

Not critical for this use case, but if you want to be precise, you could scope the replacement to `<iframe` tags only using a more targeted regex.

lefarcen (commented):
Hey @xxiaoxiong! Thanks for the fix — the root cause analysis is spot-on (browser resolving `screens/` relative to `/frames/` instead of project files). The solution (adding `project` param + API resolution path) is clean and preserves backward compat. 👍

A few suggestions below (non-blocking):

xxiaoxiong:
Thanks for the detailed review @mrcfps and @lefarcen! I've pushed a fix (9878996) that addresses all feedback:

**P1 - PROJECT_ID not replaced in URL-load mode (FIXED):**
- Force srcDoc mode when HTML contains `{{PROJECT_ID}}` placeholder
- This ensures `buildSrcdoc()` runs and replaces the placeholder before rendering
- Prevents literal `{{PROJECT_ID}}` from reaching the shared frame
- Fixes 404s on `/api/projects/%7B%7BPROJECT_ID%7D%7D/files/...`

**P2 - Screen parameter validation (FIXED):**
- Added validation regex to all 5 frame files:
  - `android-pixel.html`
  - `browser-chrome.html`
  - `ipad-pro.html`
  - `iphone-15-pro.html`
  - `macbook.html`
- Validates screen path format: alphanumeric, hyphens, underscores, slashes, dots, must end with `.html`
- Prevents path traversal attempts (`../../etc/passwd`) and query string injection
- Defense-in-depth alongside API endpoint validation

**P3 - Edge case acknowledged:**
The `replaceProjectIdPlaceholder` function will replace any literal `{{PROJECT_ID}}` text in the HTML. This is acceptable for the current use case (mobile frame screen paths). If more precise replacement is needed in the future (e.g., preserving literal placeholders in code examples), it can be scoped to specific tags like `<iframe>`.

Ready for re-review!

assets/frames/iphone-15-pro.html
```diff
@@ -166,9 +166,27 @@
...
+      var project = qs.get('project');
       var iframe = document.getElementById('screen');
-      if (src) iframe.src = src;
+      
+      if (!screen) return;
+      
+      // Validate screen parameter to prevent path traversal
+      // Screen should be a valid file path (alphanumeric, hyphens, underscores, slashes, dots)
+      // and must end with .html
+      if (!/^[a-zA-Z0-9_\/-]+\.html$/.test(screen)) {
```
mrcfps: Problem: this validator now rejects values like `about:blank`, `./screen.html`, and any absolute URL, even though the next branch still says the no-`project` path should preserve the old absolute/relative-URL behavior. Why it matters: that is a real compatibility break for the shared frame contract — `assets/frames/README.md:70-74` explicitly requires the frame to work with `?screen=about:blank`, and every one of these loaders now returns early instead of showing the empty shell for that case. Evidence: `about:blank` contains a colon, so it cannot match `^[a-zA-Z0-9_\/-]+\.html$`, and the same regex has been copied into all five frame files in this PR. Suggested change: only apply strict path validation when `project` is present (where `screen` is supposed to be a project file), and keep the legacy non-project branch permissive so `about:blank` and other raw URLs continue to load.

mrcfps (changes requested):
@xxiaoxiong Thanks for the follow-up pass — the `{{PROJECT_ID}}` preview fix looks good, but I found one remaining blocker in the shared-frame validator that still breaks an existing `?screen=` flow.


assets/frames/ipad-pro.html
```diff
@@ -87,9 +87,27 @@
...
+      var project = qs.get('project');
       var iframe = document.getElementById('screen');
-      if (src) iframe.src = src;
+      
+      if (!screen) return;
+      
+      // Validate screen parameter to prevent path traversal
+      // Screen should be a valid file path (alphanumeric, hyphens, underscores, slashes, dots)
+      // and must end with .html
+      if (!/^[a-zA-Z0-9_\/-]+\.html$/.test(screen)) {
```
lefarcen: **P1** — Validation blocks legacy URLs. This check runs before the `project ? ... : screen` branch, so `?screen=about:blank` now fails (breaks `README.md:70-74` contract). Move validation **inside** the `if (project)` branch so it only applies to project-file paths, not legacy URLs.

lefarcen (changes requested):
Hi @xxiaoxiong — thanks for the quick iteration! The `{{PROJECT_ID}}` placeholder fix in FileViewer.tsx looks solid (forcing srcDoc mode when placeholder is present). However, mrcfps's validation concern is still present in the new commit:

The regex validation `!/^[a-zA-Z0-9_\/-]+\.html$/.test(screen)` runs **before** the `project ? ... : screen` branch in all 5 frame files. This means legacy URLs like `about:blank`, `./screen.html`, and absolute URLs are now rejected, even though the code comment says "treat screen as absolute or relative URL (legacy behavior)" and `assets/frames/README.md:70-74` requires `?screen=about:blank` to work.

The validation needs to be **conditional**: only validate when `project` is present (project-file path), but skip validation when no `project` (legacy URL path). For example:

```js
if (project) {
  // Validate screen as project file path
  if (!/^[a-zA-Z0-9_\/-]+\.html$/.test(screen)) {
    console.error('Invalid screen path:', screen);
    return;
  }
  var src = '/api/projects/' + encodeURIComponent(project) + '/files/' + screen;
} else {
  // Legacy: treat screen as absolute/relative URL (no validation)
  var src = screen;
}
```

This preserves the `about:blank` contract while still protecting the new project-file path from traversal attacks.

assets/frames/browser-chrome.html
```diff
@@ -119,10 +119,28 @@
...
     (function () {
       var qs = new URLSearchParams(location.search);
-      var src = qs.get('screen');
+      var screen = qs.get('screen');
+      var project = qs.get('project');
       var url = qs.get('url');
       var iframe = document.getElementById('screen');
-      if (src) iframe.src = src;
+      
+      if (screen) {
```
lefarcen, line 127: **P1 — Validation still runs before `project` check**

This validation runs for ALL `screen` values (line 127-133), even when `project` is not present. That breaks the legacy contract — `?screen=about:blank` should work without validation when no `project` param is provided.

**The other 4 frame files got this right** — they only validate when `project` is present:
```js
if (project) {
  if (!/^[a-zA-Z0-9_\/-]+\.html$/.test(screen)) {
    console.error('Invalid screen path:', screen);
    return;
  }
  iframe.src = '/api/projects/' + encodeURIComponent(project) + '/files/' + screen;
} else {
  iframe.src = screen;  // legacy path, no validation
}
```

Please apply the same pattern here so `browser-chrome.html` matches the other frames. Then all 5 will preserve `about:blank` / relative / absolute URL support.

lefarcen (changes requested):
Thanks @xxiaoxiong — you're 80% of the way there! 🎯

The fix works perfectly in 4 out of 5 frame files (android-pixel, ipad-pro, iphone-15-pro, macbook). However, **browser-chrome.html** still has the validation running before the `project` check, which means `about:blank` and other legacy URLs are still blocked in that frame.

See inline comment for the exact fix.

assets/frames/browser-chrome.html
```diff
@@ -119,10 +119,28 @@
...
+        // Validate screen parameter to prevent path traversal
+        // Screen should be a valid file path (alphanumeric, hyphens, underscores, slashes, dots)
+        // and must end with .html
+        if (!/^[a-zA-Z0-9_\/-]+\.html$/.test(screen)) {
```
mrcfps: Problem: this validator still runs for every `screen` value before the `project ? ... : screen` fallback, so the browser frame now rejects legacy raw URLs like `about:blank`, `./screen.html`, and absolute URLs whenever `project` is absent. Why it matters: the PR description and the comment in this block both say the no-`project` branch should preserve legacy URL behavior, and `assets/frames/README.md:70-74` still requires frames to work with `?screen=about:blank`; with the current regex, `about:blank` fails here and returns early instead of rendering the shell. Evidence: commit 3 moved this check inside `if (project)` for the other four frame files, but `browser-chrome.html` was left on the old unconditional structure in lines 127-140. Suggested change: mirror the other loaders here too — only validate `screen` inside `if (project)`, and keep the no-`project` branch as a permissive `iframe.src = screen`.

mrcfps (changes requested):
@xxiaoxiong Thanks for the follow-up — the `{{PROJECT_ID}}` preview fix looks good, but I found one remaining compatibility blocker in the browser frame that still breaks an existing `?screen=` path.


mrcfps (dismissed):
@xxiaoxiong Thanks for iterating on this. I re-checked the current head and verified the shared-frame flow now covers both pieces that were blocking before: placeholder-bearing previews are forced through `srcDoc` so `{{PROJECT_ID}}` gets resolved, and all five frame loaders now only validate `screen` when `project` is present so legacy `?screen=about:blank` and raw-URL behavior stay intact. Nice cleanup on the follow-up fixes.


lefarcen (dismissed):
Perfect! 🎉

The `browser-chrome.html` validation is now correctly placed **inside** the `if (project)` block (lines 127-141), matching the other 4 frames. This preserves the legacy URL behavior (`about:blank`, `./screen.html`, absolute URLs) when no `project` param is present, while still protecting the new project-file path from traversal attacks.

All 5 frames are now consistent, and the `{{PROJECT_ID}}` placeholder replacement in FileViewer.tsx + srcdoc.ts remains intact.

Ready for mrcfps to re-review. 👍

PerishCode:
@xxiaoxiong — flagging this one because the open queue currently has two PRs from you with byte-for-byte identical titles (`fix: resolve mobile frame screen paths as project files`):

- this PR (#1098, branch `fix/mobile-frame-screen-path`, 4 commits, +110 −15)
- #1181 (branch `fix/mobile-frame-project-paths`, 1 commit, +74 −12)

Could you confirm which one you'd like reviewers to land? If #1181 is meant to supersede this one, feel free to close this in favor of it (or let me know and I'll close it). If they're intentionally covering different cases, a one-line note here clarifying the split would help reviewers track them.

(Surfaced by maintainer tooling: `duplicate-title` tag on same-author byte-for-byte titles.)

lefarcen:
@xxiaoxiong good catch on the duplicate titles — checked both PRs and this one (#1098) is the more comprehensive version:

**This PR (#1098):**
- 4 commits addressing 3 review cycles (mrcfps + bot)
- Covers all 5 frame files + web runtime (, `srcdoc.ts`)
- Both mrcfps and bot APPROVED on current head (65c00fe8)

**PR #1181:**
- 1 commit, created 2 days after this one
- Only covers 4 files (missing `browser-chrome.html`, `macbook.html`, and web runtime changes)

Looks like #1181 might be an earlier branch you accidentally pushed after this one was already through review. Recommend keeping this PR (#1098) as the canonical fix and closing #1181 if it's redundant.

lefarcen:
Hey @xxiaoxiong — quick follow-up from the linked #1087 thread.

I re-checked the remaining review conversations on the current head (`65c00fe8`): the `{{PROJECT_ID}}` URL-load gap is covered by forcing placeholder-bearing HTML through `srcDoc`, and the frame validators now only run inside the `project` branch. I resolved the addressed review threads so that part is no longer blocking.

The remaining blocker is now the merge conflict with `main` (`mergeable=CONFLICTING`). Since @mrtncode asked whether maintainers can help finish this, I’m moving this to maintainer follow-up for a rebase/conflict-resolution decision. If you get to it first, rebasing this branch onto `main` should be the next step.

lefarcen (commented):
Hey @xxiaoxiong — the Problem / Root Cause / Solution write-up is clear, and it gives reviewers the important context for #1087. One quick PR-body cleanup before the next pool re-review: could you add the missing Surface area checklist and a Bug fix verification note?

For the verification note, the useful bits are the before/after for this bug: the old `?screen=screens/start.html` path 404ing under `/frames/`, the new `project={{PROJECT_ID}}&screen=...` path loading from the project files, and the legacy no-`project` case (for example `?screen=about:blank`) still working. That will make the final review/merge pass much easier to validate.

mrtncode:
Friendly bump :)
I don't want to be impatient, but it would be great if this important issue could be fixed soon. :)

lefarcen:
Hey @mrtncode — totally fair bump, and thanks for keeping the signal clear here.

Current status: #1098 is still the canonical fix for #1087, the earlier review threads are resolved, and CI is green. The remaining blocker is that GitHub currently reports this branch as conflicting with `main`, so it needs a rebase/conflict resolution before it can land; after that, it will also need a fresh current-head review approval because the previous approval was dismissed by the newer push.

We’ve already surfaced the rebase/takeover question to maintainers, and I’ll keep tracking this PR here rather than opening another path.

mrtncode:
@lefarcen Thanks for quick update! I hope the PR gets rebase and review fast :)

mrtncode:
Please reopen :)

lefarcen:
Thanks @mrtncode — I see the reopen request.

This PR was auto-closed by queue management, but the underlying blocker is still the same: the branch currently conflicts with `main`, so reopening/continuing it needs a maintainer or the PR author to decide the rebase path first. I’ve routed the reopen request to maintainers rather than reopening it directly, so we don’t create another stuck queue cycle.

eisenwave:
The author is a Chinese bot account who went from basically no activity to opening 309 PRs in 66 repositories in two weeks

image

You're not going to get a response to change requests on PRs because these agents don't seem to act on that information in general.

lefarcen:
Thanks for flagging the account-activity concern, @eisenwave. We’ll factor that into the reopen/takeover decision, but we’ll keep the thread focused on maintainability and whether someone can actually carry the fix forward.

Current status: this PR is closed, the branch still conflicts with `main`, and the previous approval was dismissed by a later push. So it won’t be reopened or merged just because the patch direction looked useful earlier — it needs either the author to actively rebase/respond, or a maintainer to explicitly take ownership of the branch/fix.

mrtncode:
@lefarcen But I thought the issue is already fixes by another PR? :) You closes the issue yourself? Why this PR should be reopened?

lefarcen:
You’re right, @mrtncode — thanks for catching that.

At the time of the reopen request, #1087 was still being tracked and #1098 was one possible path, but it has since been superseded. The fix landed through #2316 and #2318, and I closed #1087 as completed after those merged.

So the current answer is: this PR should **not** be reopened now. #1098 can stay closed as a superseded attempt, and the issue is already covered by the merged fix path.

PR #1099, opened 2026-05-09
fix: signal when Inspect/Picker resolves to ancestor

### LINKED ISSUE #1022: Inspect/Picker click on unannotated child resolves to annotated ancestor (partial-annotation)
## Context

Follow-up from #890 / PR #1005. The original report's _second_ scenario (also documented inline by @fuyizheng3120) is a separate-but-related bug from the empty-state one PR #1005 fixes:

> In a mobile-home mockup my agent generated, the wrapper `<div class=\"header\">` carries `data-od-id=\"hero-header\"`, but the `<h1>` inside it does not. Picking the heading actually edits the wrapper — Size 23px / Weight 300 in the Inspect panel reflects the wrapper's computed style, not the `<h1>` the user clicked.

This is what \`closestTarget()\` in \`apps/web/src/runtime/srcdoc.ts\` is supposed to do (walk up to the nearest annotated ancestor), so the contract is consistent — but the resulting UX is misleading because the user has no signal that their click resolved to a different element than the one under the cursor.

## Why this is its own issue

PR #1005 narrowly fixes the **zero-annotation** case (`liveCommentTargets.size === 0` → calm \"no shipped annotations yet\" hint instead of the static \"Click any element with `data-od-id`\" copy on artifacts that have nothing to click). The **partial-annotation** scenario still hits a misleading no-op-feeling result — \`liveCommentTargets.size > 0\` so the empty-state hint never fires.

A clean fix needs one of:

1. **Bridge signal for click-miss / ancestor-fallback** — emit `od:comment-target-ancestor` (or extend the existing `od:comment-target` payload with a `clickedDescendant` field) when `closestTarget()` had to walk up. The host can then highlight the actual target on hover, or render a small overlay (\"Editing wrapper `[data-od-id=hero-header]` because the `<h1>` you clicked has no annotation\"). Smallest behavior change; preserves save-to-source semantics; gives the user the explanation they need.
2. **Synthetic-id fallback** — `targetFrom()` falls back to a CSS-selector-path-derived id for unannotated children. Requires:
   - Inspect persistence: the override CSS in `<style data-od-inspect-overrides>` keys off the elementId. A synthetic selector path drifts when the agent regenerates the artifact (or the user manually edits the source) — overrides written against \`#root > div:nth-child(2) > h1\` would silently break the next time the markup shifts. Either we accept that risk (override floats off after edits) or we add validation/migration logic at save-to-source time.
   - Comment persistence: comments key off `elementId` too — same fragility risk.
3. **UI affordance to descend into the click target** — keep the current ancestor-walk, but offer a \"select the descendant instead\" action in the Inspect panel / comment popover when the clicked element wasn't itself annotated. Lower complexity than (2); doesn't change persistence; but doesn't help users who don't notice the affordance.

I'd lean toward (1) as the cheapest viable fix: it preserves the save-to-source invariant (only annotated elements get persisted style overrides), gives the user a clear signal the click did something different from what they expected, and doesn't require schema changes.

## Reproducer

1. Generate any HTML artifact where a parent has `data-od-id` but a child does not — the most natural shape today is the agent-generated mobile mockup the original report attached.
2. Open the artifact, enter Inspect.
3. Click the inner element (e.g. the heading inside an annotated wrapper).
4. The Inspect panel shows the wrapper's computed style — not the heading's. Adjusting font size / weight / radius edits the wrapper, often invisibly when the wrapper has none of those properties to begin with.

## References

- PR #1005 — empty-annotation fix (this issue's narrower sibling).
- `apps/web/src/runtime/srcdoc.ts` — `closestTarget()` and `targetFrom()` implementations.

xxiaoxiong:
Fixes #1022

## Problem

When a user clicks an unannotated child element (e.g., `<h1>` inside `<div data-od-id="hero">`), the Inspect picker walks up to the nearest annotated ancestor and shows its computed style. The user has no signal that the click resolved to a different element, making the UI feel misleading — adjusting font-size/weight edits the wrapper, not the heading they clicked.

## Root Cause

The bridge's `closestTarget()` walks from `event.target` up to the nearest `[data-od-id]` / `[data-screen-label]` ancestor and returns only that resolved element. `targetFrom()` then snapshots the ancestor's selector and computed style. The host (FileViewer) has no way to know the original click target differed from the resolved annotated ancestor.

## Solution (Option 1 from Issue Discussion)

1. **Bridge**: `closestTarget()` now returns `{ target, clicked }` instead of just the target element
2. **Bridge**: `targetFrom()` accepts the clicked element and includes a `clickedDescendant` field in the payload when `clicked !== target`
3. **Host**: `InspectTarget` type extended with optional `clickedDescendant`
4. **Host**: `InspectPanel` displays an info notice when `clickedDescendant` is present, explaining which element was clicked and which ancestor is being edited

### Bridge (`apps/web/src/runtime/srcdoc.ts`)
- `closestTarget()` returns `{ target, clicked }` instead of just `el`
- `targetFrom()` accepts `clickedEl` parameter and adds `clickedDescendant` to payload when the clicked element differs from the resolved target
- Click handler updated to pass both `target` and `clicked` to `targetFrom`

### Host (`apps/web/src/components/FileViewer.tsx`)
- `InspectTarget` type extended with `clickedDescendant?: { label: string; text: string }`
- Message handler updated to capture `clickedDescendant` from bridge
- `InspectPanel` renders info notice when `clickedDescendant` is present:
  ```
  ℹ️ You clicked h1 ("Title"), but it has no data-od-id annotation.
     Editing div.hero instead (the nearest annotated ancestor).
  ```

### Styles (`apps/web/src/index.css`)
- Added `.inspect-ancestor-notice` styles (info banner with icon, text, and code formatting)

## Backward Compatibility

✅ The `clickedDescendant` field is optional. Existing artifacts and tests continue to work unchanged. When `clicked === target` (the common case), no `clickedDescendant` field is added to the payload.

## Testing

1. Generate an HTML artifact where a parent has `data-od-id` but a child does not:
   ```html
   <div data-od-id="hero">
     <h1>Title</h1>
   </div>
   ```
2. Open Inspect mode and click the `<h1>`
3. The Inspect panel should show an info notice:
   > ℹ️ You clicked **h1** ("Title"), but it has no `data-od-id` annotation. Editing **div.hero** instead (the nearest annotated ancestor).
4. Adjusting styles applies to the wrapper, as before, but now the user understands why

## Why This Approach

This implements **option 1** from the issue discussion:
- ✅ Preserves save-to-source semantics (only annotated elements get persisted style overrides)
- ✅ Gives the user a clear signal the click did something different from what they expected
- ✅ Doesn't require schema changes or persistence migration
- ✅ Smallest behavior change with focused scope

Options 2 (synthetic-id fallback) and 3 (UI affordance to descend) were considered but involve larger changes to persistence or more complex UX patterns.

---

**Credit**: Root cause analysis and solution options by @lefarcen in #1022

### DIFF
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -627,20 +627,20 @@ function injectSelectionBridge(
   function pickerActive(){ return inspectEnabled || (commentEnabled && mode === 'picker'); }
   document.addEventListener('mouseover', function(ev){
     if (!pickerActive()) return;
-    var el = closestTarget(ev);
-    if (!el) return;
-    var payload = targetFrom(el);
+    var result = closestTarget(ev);
+    if (!result) return;
+    var payload = targetFrom(result.target);
     if (!payload || payload.elementId === hoveredId) return;
     hoveredId = payload.elementId;
     window.parent.postMessage(Object.assign({}, payload, { type: 'od:comment-hover' }), '*');
   }, true);
   document.addEventListener('mouseout', function(ev){
     if (!pickerActive()) return;
-    var el = closestTarget(ev);
-    if (!el) return;
+    var result = closestTarget(ev);
+    if (!result) return;
     var next = ev.relatedTarget;
     while (next && next !== document.documentElement) {
-      if (next === el) return;
+      if (next === result.target) return;
       next = next.parentElement;
     }
     hoveredId = null;
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -448,24 +448,35 @@ function injectSelectionBridge(
       };
     } catch (_) { return null; }
   }
-  function targetFrom(el){
+  function targetFrom(el, clickedEl){
     var id = el.getAttribute('data-od-id') || el.getAttribute('data-screen-label');
     if (!id) return null;
     var rect = el.getBoundingClientRect();
     var tag = el.tagName ? el.tagName.toLowerCase() : 'element';
-    var cls = typeof el.className === 'string' && el.className.trim() ? '.' + el.className.trim().split(/\\s+/).slice(0,2).join('.') : '';
+    var cls = typeof el.className === 'string' && el.className.trim() ? '.' + el.className.trim().split(/\s+/).slice(0,2).join('.') : '';
     var html = '';
-    try { html = (el.outerHTML || '').replace(/\\s+/g, ' ').match(/^<[^>]+>/)?.[0] || ''; } catch (_) {}
-    return {
+    try { html = (el.outerHTML || '').replace(/\s+/g, ' ').match(/^<[^>]+>/)?.[0] || ''; } catch (_) {}
+    var payload = {
       type: 'od:comment-target',
       elementId: id,
       selector: el.hasAttribute('data-od-id') ? '[data-od-id="' + esc(id) + '"]' : '[data-screen-label="' + esc(id) + '"]',
       label: tag + cls,
-      text: (el.textContent || '').replace(/\\s+/g, ' ').trim().slice(0, 160),
+      text: (el.textContent || '').replace(/\s+/g, ' ').trim().slice(0, 160),
       position: { x: Math.round(rect.x), y: Math.round(rect.y), width: Math.round(rect.width), height: Math.round(rect.height) },
       htmlHint: html.slice(0, 180),
       style: styleSnapshot(el)
     };
+    // If the clicked element differs from the resolved annotated ancestor,
+    // include a clickedDescendant field so the host can signal the user.
+    if (clickedEl && clickedEl !== el) {
+      var clickedTag = clickedEl.tagName ? clickedEl.tagName.toLowerCase() : 'element';
+      var clickedCls = typeof clickedEl.className === 'string' && clickedEl.className.trim() ? '.' + clickedEl.className.trim().split(/\s+/).slice(0,2).join('.') : '';
+      payload.clickedDescendant = {
+        label: clickedTag + clickedCls,
+        text: (clickedEl.textContent || '').replace(/\s+/g, ' ').trim().slice(0, 80)
+      };
+    }
+    return payload;
   }
   function allTargets(){
     var nodes = document.querySelectorAll('[data-od-id], [data-screen-label]');
@@ -500,9 +511,12 @@ function injectSelectionBridge(
     window.parent.postMessage({ type: type, points: stroke.slice() }, '*');
   }
   function closestTarget(event){
-    var el = event.target;
+    var clicked = event.target;
+    var el = clicked;
     while (el && el !== document.documentElement) {
-      if (el.getAttribute && (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label'))) return el;
+      if (el.getAttribute && (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label'))) {
+        return { target: el, clicked: clicked };
+      }
       el = el.parentElement;
     }
     return null;
@@ -634,11 +648,11 @@ function injectSelectionBridge(
   }, true);
   document.addEventListener('click', function(ev){
     if (!pickerActive()) return;
-    var el = closestTarget(ev);
-    if (!el) return;
+    var result = closestTarget(ev);
+    if (!result) return;
     ev.preventDefault();
     ev.stopPropagation();
-    var payload = targetFrom(el);
+    var payload = targetFrom(result.target, result.clicked);
     if (payload) window.parent.postMessage(payload, '*');
   }, true);
   // Pod drawing — only active in comment mode with the 'pod' tool.
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -142,6 +142,7 @@ type InspectTarget = {
   label: string;
   text: string;
   style: InspectStyleSnapshot;
+  clickedDescendant?: { label: string; text: string };
 };
 
 const MAX_CACHED_SLIDE_STATES = 64;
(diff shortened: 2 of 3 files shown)


apps/web/src/runtime/srcdoc.ts
```diff
@@ -500,9 +511,12 @@ function injectSelectionBridge(
     window.parent.postMessage({ type: type, points: stroke.slice() }, '*');
   }
   function closestTarget(event){
-    var el = event.target;
+    var clicked = event.target;
+    var el = clicked;
     while (el && el !== document.documentElement) {
-      if (el.getAttribute && (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label'))) return el;
+      if (el.getAttribute && (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label'))) {
+        return { target: el, clicked: clicked };
```
mrcfps, line 518: Returning an object here changes `closestTarget()`'s contract, but the hover handlers below still treat its return value as an `Element` (`var payload = targetFrom(el)` in `mouseover`, and `if (next === el)` in `mouseout`). Once the picker hovers an annotated node, `targetFrom(el)` tries to call `getAttribute(...)` on `{ target, clicked }`, so hover enter/leave feedback breaks in both Inspect mode and the comment picker. Please either keep `closestTarget()` element-shaped for existing callers or update those hover paths to unwrap `result.target` before calling `targetFrom(...)` and comparing ancestors.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -448,24 +448,35 @@ function injectSelectionBridge(
...
     } catch (_) { return null; }
   }
-  function targetFrom(el){
+  function targetFrom(el, clickedEl){
     var id = el.getAttribute('data-od-id') || el.getAttribute('data-screen-label');
     if (!id) return null;
     var rect = el.getBoundingClientRect();
     var tag = el.tagName ? el.tagName.toLowerCase() : 'element';
-    var cls = typeof el.className === 'string' && el.className.trim() ? '.' + el.className.trim().split(/\\s+/).slice(0,2).join('.') : '';
+    var cls = typeof el.className === 'string' && el.className.trim() ? '.' + el.className.trim().split(/\s+/).slice(0,2).join('.') : '';
```
mrcfps, line 456: These regex literals live inside the injected `<script>` template string, so the single backslash is consumed while `buildSrcdoc()` builds the iframe HTML and the browser ends up running `/s+/` instead of `/\s+/`. That means the new label/text normalization in this block will split on the letter `s` rather than whitespace, mangling values like `class="hero section"` or `"Sales stats"` in `label`, `text`, and `clickedDescendant.text`. Please keep these as `\s` inside the template (matching the rest of this bridge's escaped regexes) so the emitted script still normalizes whitespace correctly.

mrcfps (changes requested):
@xxiaoxiong Thanks for the clear UI improvement here — I walked through the bridge/host changes and found two regressions in the updated selection bridge that should be fixed before this lands.


apps/web/src/runtime/srcdoc.ts
```diff
@@ -500,9 +511,12 @@ function injectSelectionBridge(
     window.parent.postMessage({ type: type, points: stroke.slice() }, '*');
   }
   function closestTarget(event){
-    var el = event.target;
+    var clicked = event.target;
```
lefarcen, line 514: **P1 — Breaking change not applied to all call sites**

`closestTarget` now returns `{ target, clicked }` instead of just the element, but only the `click` event handler was updated. The `mouseover` and `mouseout` handlers still expect the old return type:

```javascript
// Line ~502 (mouseover)
var el = closestTarget(ev);  // ❌ expects element, gets {target, clicked}
if (!el) return;
var payload = targetFrom(el);  // ❌ will fail

// Line ~508 (mouseout)
var el = closestTarget(ev);  // ❌ same issue
if (!el) return;
var next = ev.relatedTarget;
```

**Why this breaks**: Both handlers call `targetFrom(el)` which now expects `targetFrom(element, clickedElement)` signature. They'll pass an object where an element is expected.

**Fix**: Update both handlers to destructure the result:
```javascript
// mouseover:
var result = closestTarget(ev);
if (!result) return;
var payload = targetFrom(result.target, result.clicked);

// mouseout:
var result = closestTarget(ev);
if (!result) return;
// (mouseout doesn't use targetFrom, but still needs result.target for comparisons)
```

lefarcen (changes requested):
Hey @xxiaoxiong! 👋 Thanks for the PR — the approach (tracking both clicked element and resolved ancestor, then showing an info notice) is exactly right for solving issue #1022.

Found 1 critical issue that will break mouseover/mouseout behavior — see inline below. Once fixed this will be good to go.

xxiaoxiong:
Thanks for the detailed review @mrcfps and @lefarcen! I've pushed a fix (fda6840) that addresses the breaking change:

**Changes:**
- Updated `mouseover` handler to use `result.target` instead of treating `closestTarget()` return as an `Element`
- Updated `mouseout` handler to use `result.target` for the parent walk comparison
- Now consistent with the `click` handler which already uses the correct `{ target, clicked }` pattern

**Re: regex escaping:**
The regex literals at lines 456 and 473 are already correctly escaped as `\\s+` in the source (double backslash in TypeScript becomes single backslash in the emitted JavaScript). I verified this is working as intended.

Ready for re-review!

mrcfps (approved):
@xxiaoxiong I re-checked the latest head and the follow-up fix closes the hover regression: `closestTarget()`'s new `{ target, clicked }` shape is now handled consistently across click, mouseover, and mouseout, and the new `clickedDescendant` payload is wired through to the inspect notice without changing the existing save-to-annotated-ancestor behavior. I also spot-checked the injected bridge normalization paths and tried `pnpm --filter @open-design/web typecheck`, but this worktree is missing `node_modules`, so that command could not be used as a fresh validation signal here.

Thanks for turning the review feedback around quickly — this is a clear usability improvement for Inspect/Picker. ✨


lefarcen (approved):
Clean. The follow-up fix addresses both breaking-change regressions mrcfps caught:

✅ `mouseover` and `mouseout` now correctly unwrap `result.target` before calling `targetFrom` and comparing ancestors — hover enter/leave feedback works again
✅ Regex escaping confirmed correct in source (author's verification checks out)

The new `clickedDescendant` payload wires through cleanly without changing save-to-annotated-ancestor semantics. Ready for merge from my side once mrcfps re-reviews the current head.

— open-design team

mrcfps:
@xxiaoxiong Thanks again for the contribution here 💛

Issue #784, opened 2026-05-07
Video artifacts cannot be previewed inline in the file viewer


qiongyu1999:
## Summary

When a generated artifact is a video (e.g. `.mp4`), the in-app file viewer does not render a video player. It only shows the file metadata (`视频 · 3.7 MB`) plus **下载 / 打开** buttons, forcing the user to leave the app (open in an external player) to see what was generated.

Image artifacts (`.png`, etc.) preview inline in the same panel, so the inconsistency is jarring — the AI generated something I asked for, but I can't see it without leaving the workflow.

## Steps to reproduce

1. Start a project and ask the agent to generate a video (e.g. via Volcengine Seedance image-to-video).
2. Wait for the run to complete; an `.mp4` artifact appears in **本轮产出的文件**.
3. Click the video file or open it in a tab in the right-hand file viewer.

**Expected**: an inline `<video>` player with play/pause/scrub controls, like image artifacts get an inline preview.

**Actual**: only `视频 · 3.7 MB` text and 下载 / 打开 buttons. No inline player.

## Screenshot

The right pane shows the video tab selected with no preview surface — just metadata and action buttons. (Attaching screenshot in a follow-up comment if needed.)

## Environment

- Open Design (latest `main`, also reproduced on `feat/orbit-agent-summary`)
- macOS, Chrome-based desktop shell
- Artifact: `.mp4` produced by Volcengine Seedance 2.0

## Proposal

Render an HTML5 `<video controls preload=\"metadata\">` for video MIME types in the file viewer, mirroring how image artifacts are previewed. Suggested scope:

1. Detect by extension / MIME (`video/mp4`, `video/webm`, `video/quicktime`, etc.)
2. Inline player with native controls; lazy-load to avoid bandwidth cost on large files
3. Keep **下载 / 打开** buttons as fallback for unsupported codecs
4. Optional: poster frame from the source image when the video was generated from one (image-to-video flow)

## Why this matters

Video generation is becoming a first-class flow in Open Design (image-to-video via Seedance, etc.). Without inline preview, the iteration loop breaks — users can't quickly judge a result and re-prompt without context-switching to Finder / QuickLook.

## Additional context

Audio files likely have the same gap; worth checking `audio/*` MIME handling at the same time.

lefarcen:
[...]

PR #1100, opened 2026-05-09
feat(daemon): inline video & audio preview with HTTP range request support [#784]

ak91456:
Description:

  ## Summary
  - Added `resolveProjectFilePath` to `apps/daemon/src/projects.ts` to resolve a safe absolute file path without reading the file into memory
  - Updated `GET /api/projects/:id/files/*` in `apps/daemon/src/server.ts` to serve video/audio files via `fs.createReadStream` with `206 Partial Content`
  and `Accept-Ranges: bytes` headers instead of buffering the entire file

  ## Problem
  Video and audio artifacts could not be previewed inline in the file viewer. The `VideoViewer` and `AudioViewer` components already existed on the
  frontend, but the daemon was serving all files by reading the full buffer into memory (`res.send(buffer)`), with no `Accept-Ranges` header. Browsers
  require HTTP range request support to seek, scrub, or even begin playing video/audio — without it, playback silently fails.

  ## Test plan
  - [ ] Generate a video artifact via an agent (e.g. image-to-video workflow)
  - [ ] Click the video file in the file viewer — confirm inline `<video>` player renders with controls
  - [ ] Verify seeking/scrubbing works in the player
  - [ ] Repeat for an audio file (`.mp3`, `.wav`, `.m4a`)
  - [ ] Confirm non-media files (HTML, images, text) are unaffected

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -33,7 +33,6 @@ import {
 import { migrateLegacyDataDirSync } from './legacy-data-migrator.js';
 import { findSkillById, listSkills, splitDerivedSkillId } from './skills.js';
 import { validateLinkedDirs } from './linked-dirs.js';
-import { installFromTarget, uninstallById, sanitizeRepoName } from './library-install.js';
 import { buildWindowsFolderDialogCommand, parseFolderDialogStdout } from './native-folder-dialog.js';
 import { listCodexPets, readCodexPetSpritesheet } from './codex-pets.js';
 import { syncCommunityPets } from './community-pets-sync.js';
@@ -60,7 +59,6 @@ import {
   testProviderConnection,
   validateBaseUrl,
 } from './connectionTest.js';
-import { listProviderModels } from './providerModels.js';
 import { importClaudeDesignZip } from './claude-design-import.js';
 import {
   finalizeDesignPackage,
@@ -118,11 +116,6 @@ import {
 } from './mcp-tokens.js';
 import { agentCliEnvForAgent, readAppConfig, writeAppConfig } from './app-config.js';
 import { OrbitService, formatLocalProjectTimestamp, renderOrbitTemplateSystemPrompt } from './orbit.js';
-import {
-  RoutineService,
-  validateSchedule as validateRoutineSchedule,
-  validateTarget as validateRoutineTarget,
-} from './routines.js';
 import { buildMcpInstallPayload } from './mcp-install-info.js';
 import {
   buildProjectArchive,
@@ -133,11 +126,12 @@ import {
   ensureProject,
   isSafeId,
   listFiles,
+  kindFor,
   mimeFor,
   projectDir,
   readProjectFile,
-  renameProjectFile,
   removeProjectDir,
+  resolveProjectFilePath,
   sanitizeName,
   searchProjectFiles,
   writeProjectFile,
@@ -156,8 +150,6 @@ import {
   getTemplate,
   insertConversation,
   insertProject,
-  insertRoutine,
-  insertRoutineRun,
   insertTemplate,
   listProjectsAwaitingInput,
   listConversations,
@@ -166,20 +158,13 @@ import {
   listMessages,
   listPreviewComments,
   listProjects,
-  listRoutines,
-  listRoutineRuns,
   listTabs,
   listTemplates,
-  getLatestRoutineRun,
-  getRoutine,
-  deleteRoutine as dbDeleteRoutine,
   openDatabase,
   setTabs,
   updateConversation,
   updatePreviewCommentStatus,
   updateProject,
-  updateRoutine,
-  updateRoutineRun,
   upsertDeployment,
   upsertMessage,
   upsertPreviewComment,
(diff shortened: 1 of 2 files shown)


apps/daemon/src/server.ts
```diff
@@ -839,14 +821,8 @@ migrateLegacyDataDirSync({
 });
 const ARTIFACTS_DIR = path.join(RUNTIME_DATA_DIR, 'artifacts');
 const PROJECTS_DIR = path.join(RUNTIME_DATA_DIR, 'projects');
```
lefarcen, line 823: **P1 — Scope creep: unrelated deletions**

This PR removes active daemon surfaces:
- `/api/projects/:id/files/rename` (still called by FileWorkspace)
- All `/api/routines*` endpoints + RoutineService
- `/api/provider/models` (still called by provider-models.ts)
- User-installed skill/design-system paths

These are unrelated to media streaming and would break current functionality. Please restore them and keep this PR scoped to the file serving changes only.

apps/daemon/src/server.ts
```diff
@@ -4521,10 +4361,48 @@ export async function startServer({
...
+          PROJECTS_DIR,
+          req.params.id,
+          name,
+          project?.metadata,
+        );
+        const { size } = await fs.promises.stat(absPath);
+        const rangeHeader = req.headers.range;
+
+        if (rangeHeader) {
+          const [startStr, endStr] = rangeHeader.replace(/bytes=/, '').split('-');
```
lefarcen, line 4380: **P1 — Invalid range handling**

The Range parser accepts invalid ranges:
- Suffix ranges like `bytes=-500` → `start = NaN`
- Open-ended `bytes=999999-` → negative Content-Length
- Out-of-bounds `bytes=0-999999` → advertises more bytes than file has

Implement RFC 7233 compliant parsing:
```typescript
const match = range.match(/^bytes=(\d+)?-(\d+)?$/);
if (!match) return res.status(416).set('Content-Range', `bytes */${size}`).end();
const start = match[1] ? parseInt(match[1], 10) : 0;
const end = match[2] ? Math.min(parseInt(match[2], 10), size - 1) : size - 1;
if (start >= size || start > end) {
  return res.status(416).set('Content-Range', `bytes */${size}`).end();
}
```

apps/daemon/src/server.ts
```diff
@@ -4521,10 +4361,48 @@ export async function startServer({
...
+          const start = parseInt(startStr, 10);
+          const end = endStr ? parseInt(endStr, 10) : size - 1;
+          const chunkSize = end - start + 1;
+          res.writeHead(206, {
+            'Content-Range': `bytes ${start}-${end}/${size}`,
+            'Accept-Ranges': 'bytes',
+            'Content-Length': chunkSize,
+            'Content-Type': mime,
+          });
+          fs.createReadStream(absPath, { start, end }).pipe(res);
```
lefarcen, line 4390: **P2 — Missing stream error handler**

`createReadStream().pipe(res)` has no error handler. If the file is removed between stat and open, or the stream errors mid-read, this can cause unhandled stream errors.

Use pipeline or attach an error handler:
```typescript
import { pipeline } from 'node:stream/promises';
await pipeline(fs.createReadStream(filePath, { start, end }), res).catch(err => {
  if (!res.headersSent) res.status(500).send('Stream error');
});
```

apps/daemon/src/server.ts
```diff
@@ -4521,10 +4361,48 @@ export async function startServer({
   app.get('/api/projects/:id/files/*', async (req, res) => {
     try {
       const project = getProject(db, req.params.id);
+      const name = req.params[0];
+      const mime = mimeFor(name);
+      const kind = kindFor(name);
+
+      // Stream video and audio with range request support so browsers can seek.
```
lefarcen, line 4368: **P2 — Missing test coverage**

No tests for the new media serving behavior. Please add daemon tests for:
- 200 full media response
- Valid 206 byte ranges (start-end, start-, -suffix)
- Clamped/open-ended ranges
- Invalid/unsatisfiable 416 responses  
- Non-media files still work as before

lefarcen (changes requested):
Hey @ak91456! Thanks for tackling the media preview issue 👋

The media-streaming goal is solid — browsers need HTTP range requests for video/audio playback. However, there are critical issues that need fixing before this can merge:

1. **Unrelated deletions**: The PR removes active daemon APIs (file rename, routines, provider models, library install) that are still used by the web app. These need to be restored.

2. **Range request validation**: The current implementation accepts invalid ranges and can return broken 206 responses.

3. **Error handling**: Missing error handler for the file stream after headers are sent.

4. **Test coverage**: No tests for the new range request behavior.

See inline comments for specifics. Once these are addressed, this will be a great improvement to the media artifact experience!

lefarcen:
Thanks for working on this @ak91456! I see you closed the PR before the review landed.

If you'd like to reopen and continue working on it, the review feedback is still available above — the main findings were around scope (unrelated API deletions) and range request validation. Happy to help if you decide to pick this back up.

The underlying issue #784 (video preview) is still open and would be great to fix. Thanks for taking a look at it!

PR #1101, opened 2026-05-09
fix: add clear success feedback after saving API keys in Settings

xxiaoxiong:
Fixes #654

## Problem

After entering an API key and clicking Save, the settings modal closes immediately without providing clear confirmation that the key was saved successfully. Users have to reopen Settings to verify whether the save actually worked, creating unnecessary friction and reducing confidence in the settings experience.

## Root Cause

The Save button persists the config then immediately sets `keySaveStatus` to `'idle'` without showing a confirmation state. Settings already has inline feedback patterns (rescan status, notification test status), but the modal closes before users can see a clear success state for API key changes.

## Solution

1. **Extended `keySaveStatus` type** to include `'saved'` state
2. **After successful save**, set status to `'saved'` instead of `'idle'`
3. **Display success feedback** with check icon and "Saved" text
4. **Auto-reset to `'idle'`** after 3 seconds
5. **Added CSS styles** for success state (green background)

### `apps/web/src/components/SettingsDialog.tsx`
- Updated `keySaveStatus` type: `'idle' | 'saving' | 'saved' | 'error'`
- Modified `handleSaveKey` to set `'saved'` status with 3s timeout
- Updated Save button to show check icon + "Saved" text when saved
- Added `'is-success'` className for success state styling

### `apps/web/src/index.css`
- Added `.settings-connectors-save.is-success` styles
- Green background (`#10b981`) with darker border (`#059669`)
- Hover state with darker green (`#059669`)

## Backward Compatibility

✅ The modal no longer auto-closes after save, allowing users to see the success feedback. This is an intentional UX improvement. Users can manually close the modal after seeing the confirmation.

## Testing

1. Open Settings → Connectors
2. Enter a Composio API key
3. Click "Save key"
4. Button should show spinner + "Saving..."
5. After save completes, button turns green with check icon + "Saved"
6. After 3 seconds, button returns to normal "Save key" state
7. Modal remains open so user can see the confirmation

## Screenshots

**Before**: Modal closes immediately, no feedback
**After**: Button shows green "Saved" state with check icon for 3 seconds

## Why This Matters

As noted in the issue: "Saving credentials is a high-trust workflow. If the modal closes without clear confirmation, users are left uncertain whether the key was saved, which creates unnecessary friction and reduces confidence in the settings experience."

This fix provides explicit success feedback and keeps the modal open briefly so users can see the confirmation before manually closing it.

---

**Credit**: Issue reported by @shangxinyu1, initial PR attempt by @PratikRai0101 (#689)


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2707,6 +2711,11 @@ function ConnectorSection({
                 <Icon name="spinner" size={12} className="icon-spin" />
                 <span>{t('settings.connectorsKeySaving')}</span>
               </>
+            ) : keySaveStatus === 'saved' ? (
+              <>
+                <Icon name="check" size={12} />
+                <span>{t('settings.connectorsKeySaved')}</span>
```
lefarcen, line 2736: **P1 — Missing i18n key**

`t('settings.connectorsKeySaved')` is not declared in `Dict` or any locale file. Since `t` is typed as `keyof Dict`, this should fail typecheck; if it reached runtime it would fall back to showing the raw key instead of localized text.

**Fix**: Add the key to `apps/web/src/i18n/types.ts` and all locale dictionaries (at minimum `en.ts`, `zh-CN.ts`).

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2513,7 +2513,11 @@ function ConnectorSection({
         apiKeyTail: pendingKey.trim().slice(-4),
       });
       setCatalogRefreshNonce((nonce) => nonce + 1);
-      setKeySaveStatus('idle');
+      setKeySaveStatus('saved');
+      // Reset to idle after showing success feedback
+      setTimeout(() => {
```
lefarcen: **P2 — setTimeout race / memory leak**

The success reset timeout is not stored or cleared. Two problems:

1. **Race**: If the user types a replacement and saves again during the 3-second `saved` window, the first timeout can fire while the second save is in progress and set `keySaveStatus` back to `idle`, re-enabling actions during an in-flight secret write.
2. **Memory leak**: Leaves a pending state update after unmount.

**Fix**: Follow the existing autosave pattern in this file — keep a timeout ref, clear it before starting a new save and on unmount, and make the callback conditional (only settle `saved` to `idle` if the current state is still `saved`).

Example:
```tsx
const successTimeoutRef = useRef<NodeJS.Timeout | null>(null);

useEffect(() => {
  return () => {
    if (successTimeoutRef.current) clearTimeout(successTimeoutRef.current);
  };
}, []);

const handleSaveKey = async () => {
  if (keySaveStatus === 'saving') return;
  if (successTimeoutRef.current) clearTimeout(successTimeoutRef.current);
  // ...
  setKeySaveStatus('saved');
  successTimeoutRef.current = setTimeout(() => {
    setKeySaveStatus((prev) => (prev === 'saved' ? 'idle' : prev));
  }, 3000);
};
```

lefarcen (changes requested):
Hi @xxiaoxiong! Thanks for picking this up and superseding the stalled #689 — the UX direction is solid (explicit success feedback for credential saves is important).

Found 2 blocking issues that need fixes before merge:

apps/web/src/components/SettingsDialog.tsx
mrcfps: The new success timer in `ConnectorSection.handleSaveKey` is never tracked or cleared. That creates a real state race: if the user saves a replacement key during the 3-second success window, the first `setTimeout` will still fire and force `keySaveStatus` back to `'idle'` even while the later save is still in flight. Because `actionsLocked` only checks `keySaveStatus === 'saving'`, that stale callback can re-enable Save/Clear in the middle of an active credential write. The same callback also survives if the dialog unmounts before the 3 seconds elapse. Please store the timeout in a ref, clear it before starting another save/clear and in an unmount cleanup, and only settle back to `idle` if the current status still belongs to the same success pulse.

apps/web/src/components/SettingsDialog.tsx
mrcfps, line 2736: `t('settings.connectorsKeySaved')` is introduced in this changed range, but this PR does not add that key anywhere in the typed dictionaries. In the current tree, `apps/web/src/i18n/types.ts` and the locale files define neighboring keys like `settings.connectorsSaveKey` and `settings.connectorsKeySaving`, but there is no `settings.connectorsKeySaved` entry. That means this change either fails the typed i18n contract once dependencies are installed, or falls back to showing the raw translation key instead of the success copy. Please add `settings.connectorsKeySaved` to `apps/web/src/i18n/types.ts` and every locale dictionary that must stay in sync with it before merging.

mrcfps (changes requested):
Hi @xxiaoxiong, thanks for improving the settings save feedback here. I re-checked the changed ranges on the current head and found two merge-blocking issues that still need follow-up before this is ready to land.


xxiaoxiong:
Thanks for the detailed review @lefarcen and @mrcfps! I've pushed a fix (0ac9878) that addresses both issues:

**P1 - Missing i18n key (FIXED):**
- Added `settings.connectorsKeySaved` to `types.ts`
- Added translations to all 18 locale files (en, zh-CN, zh-TW, ja, ko, de, fr, es-ES, pt-BR, ru, ar, fa, tr, pl, uk, hu, id, th)
- Translations follow the pattern of existing "Saved" messages in each language

**P2 - setTimeout race / memory leak (FIXED):**
- Added `keySaveResetTimerRef` to track the success reset timer
- Clear timer before starting a new save (prevents race where stale timeout resets status during active save)
- Clear timer on unmount (prevents memory leak and stale state updates)
- Follows the existing autosave pattern in this file

Ready for re-review!

apps/web/src/i18n/locales/en.ts
```diff
@@ -1241,8 +1241,9 @@ export const en: Dict = {
   'settings.connectorsClearCancel': 'Cancel',
   'settings.connectorsSaveKey': 'Save key',
   'settings.connectorsSaveKeyTitle': 'Send this key to the local daemon',
-  'settings.connectorsKeySaving': 'Saving\u2026',
-  'settings.connectorsKeyError': 'Couldn\u2019t save the key. Check that the local daemon is running and try again.',
+  'settings.connectorsKeySaving': 'Saving…',
+  'settings.connectorsKeySaved': 'Saved',
+  'settings.connectorsKeyError': 'Couldn't save the key. Check that the local daemon is running and try again.',
```
mrcfps: This updated `settings.connectorsKeyError` string is wrapped in single quotes even though the message now contains `Couldn't`, so the apostrophe closes the literal early. `pnpm --filter @open-design/web typecheck` currently fails on this changed line with `TS1002: Unterminated string literal`/`TS1005` parse errors, which blocks the web package from building at all. Please switch this entry back to double quotes or escape the apostrophe so the locale object stays valid TypeScript.

mrcfps (changes requested):
@xxiaoxiong, thanks for addressing the earlier timer/i18n feedback and reworking the save-state flow. I re-checked the updated head and found one remaining merge-blocking issue: the new English locale entry currently breaks `@open-design/web` typecheck.


apps/web/src/i18n/locales/en.ts
lefarcen: P1 — This string is single-quoted but contains `Couldn't`, so the apostrophe terminates the literal early and breaks TypeScript parsing. Switch to double quotes or escape the apostrophe:

```ts
'settings.connectorsKeyError': "Couldn't save key. ..."  // double quotes
// or
'settings.connectorsKeyError': 'Couldn\'t save key. ...'  // escaped
```

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2492,12 +2492,30 @@ function ConnectorSection({
   // completes, the daemon returns a tail-only echo, and we land in
   // the saved state with the same UI as a key loaded from disk.
   const [keySaveStatus, setKeySaveStatus] =
-    useState<'idle' | 'saving' | 'error'>('idle');
+    useState<'idle' | 'saving' | 'saved' | 'error'>('idle');
   const [catalogRefreshNonce, setCatalogRefreshNonce] = useState(0);
+  const keySaveResetTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);
+
```
lefarcen, line 2498: P3 — Minor: trailing whitespace on this line and around lines 2508/2513. `git diff --check` will flag them. Please strip before merge.

lefarcen (changes requested):
Hey @xxiaoxiong, thanks for the quick iteration on the timer and i18n fixes! 👍

One blocker remains from mrcfps's last review — the English locale string still has the unescaped apostrophe. See inline below.

lefarcen (approved):
All prior issues addressed! 🎉

The P1 TypeScript parse error (en.ts apostrophe) is now fixed — the string correctly uses double quotes. The P2 timer cleanup was already addressed in the previous commit (0ac9878).

This PR is clean from my side. Ready for final maintainer approval.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2492,12 +2492,30 @@ function ConnectorSection({
...
+  
   const handleSaveKey = async () => {
     if (keySaveStatus === 'saving') return;
     if (!hasPendingEdit) return;
     if (composioConfigLoading) return;
+    
+    // Clear any pending success reset timer before starting a new save
+    if (keySaveResetTimerRef.current) {
+      clearTimeout(keySaveResetTimerRef.current);
+      keySaveResetTimerRef.current = null;
```
mrcfps, line 2517: These new lines only clear `keySaveResetTimerRef` before another save, but `handleClearCommit` now shares the same `keySaveStatus = 'saving'` / `actionsLocked` path without cancelling that pending success timer first. If a user saves a key, clicks through the two-step Clear flow during the 3-second `saved` pulse, and the old timeout fires while `onPersistComposioKey` is clearing the credential, `setKeySaveStatus('idle')` will still run and re-enable Save/Clear in the middle of that destructive write. Please clear `keySaveResetTimerRef.current` before starting the clear commit too (or move the timer cleanup into a shared helper used by both save and clear) so the race is fully closed for both credential-write paths.

mrcfps (commented):
@xxiaoxiong, thanks for tightening up the save-feedback flow and fixing the earlier i18n/typecheck regressions. I re-checked the current head and found one follow-up in the updated timer handling that still looks worth addressing.


lefarcen:
Hi @xxiaoxiong — thanks again for the careful work and iterations here.

I'm closing this PR because #1303 has now merged and GitHub has closed #654, so this branch is superseded by the shipped fix. Your version helped move the issue forward, especially around the save-state/i18n details, and I appreciate the follow-through on the earlier review feedback. 🙏

PR #740, opened 2026-05-07
[codex] fix chat pane overflow

DoTheWorkNow:

PR #1104, opened 2026-05-09
fix(web): prevent chat messages from overflowing into workspace area (#662)

### LINKED ISSUE #662: Fix chat panel UI overflowing into the workspace area
## Summary

There is a UI layout/positioning issue where content from the left chat panel visually overflows into the right workspace area.

In the reported case, chat content such as the thinking/status card and form-selection UI appears to extend past the left pane boundary and overlap the main workspace panel. This creates a broken split-view layout and makes the interface look misaligned.

## Current behavior

- Content in the left chat panel can render beyond its intended container width.
- UI elements visually cross the divider between the chat panel and the workspace panel.
- The overflowed content overlaps the Design Files area on the right.
- This happens with elements such as thinking/status rows and larger option cards/forms in the conversation area.
- The split layout no longer respects clear panel boundaries.

## Expected behavior

The left chat panel should clip or contain all of its content within its own layout bounds.

Suggested improvements:

- Prevent chat content from overflowing horizontally into the workspace pane.
- Ensure long status rows, cards, or form content respect panel width constraints.
- Apply consistent clipping, wrapping, or max-width rules inside the left panel.
- Keep the workspace panel visually isolated from chat-side content.
- Recheck z-index, overflow, and split-pane sizing behavior during dynamic message rendering.

## Steps to reproduce

2. Interact with the chat panel so that larger UI blocks appear, such as status rows, form-answer cards, or style-selection cards.
3. Observe the split view between the chat panel and the Design Files workspace.
4. Notice that some left-panel UI can visually extend into the right panel instead of staying clipped inside the chat area.

## Notes

- a thinking/status row visually extending across the split boundary
- larger card-based content from the left panel appearing inside the right workspace area

## Why this matters

This breaks the basic visual structure of the product. Panel boundaries in a split workspace need to feel stable and predictable. When chat content spills into the main workspace, it damages readability, makes the product feel unstable, and can interfere with core interactions.

Image

Image

Image

YUHAO-corn:
## Problem

Chat panel content (thinking blocks, status pills, form cards) can visually extend past the left pane boundary and overlap the workspace area on the right, breaking the split-view layout.

## Root cause

While  and  already have  (from #740), the intermediate flex items in the layout chain can still propagate intrinsic content width:

-  only had , leaving horizontal overflow as 
-  lacked , so flex items in the chat log column could expand beyond the panel width when nested content (tool cards, form elements) is wider than the container

## Fix

- Add  to  to clip any horizontally overflowing content
- Add  to  to prevent flex items from expanding beyond the chat panel width

## Verification

These are purely CSS layout constraints with zero behavioral changes. The fix ensures:
1. Chat messages stay clipped within the chat pane boundary
2. Flex items respect the panel width even when nested content is intrinsically wider
3. Existing  on  and  works correctly with the full constraint chain

Closes #662

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -838,6 +838,7 @@ code {
 .chat-log {
   flex: 1;
   overflow-y: auto;
+  overflow-x: hidden;
   padding: 16px;
   display: flex;
   flex-direction: column;
@@ -852,6 +853,8 @@ code {
   border: none;
   white-space: normal;
   word-wrap: break-word;
+  min-width: 0;
+  max-width: 100%;
 }
 .msg .role {
   display: flex;
@@ -9316,6 +9319,7 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
   display: flex;
   flex-direction: column;
   gap: 8px;
+  min-width: 0;
 }
 .prose-block { line-height: 1.6; color: var(--text); }
 .prose-block .md-p { margin: 0; }


apps/web/src/index.css
```diff
@@ -852,6 +853,8 @@ code {
   border: none;
   white-space: normal;
   word-wrap: break-word;
+  min-width: 0;
```
lefarcen, line 856: P3 — The PR description says the fix adds `min-width: 0` to `.chat-log-column`, but the diff adds it to `.msg` (line 856), and there's no `.chat-log-column` selector in the CSS. This makes the root-cause explanation harder to verify. Consider updating the PR description to say `.msg` is the intermediate flex item, or changing the selector if the intended target was different.

apps/web/src/index.css
```diff
@@ -838,6 +838,7 @@ code {
 .chat-log {
   flex: 1;
   overflow-y: auto;
+  overflow-x: hidden;
```
lefarcen, line 841: P3 — The PR description asserts this fixes thinking blocks, status pills, and form cards, but provides no viewport sizes, screenshots, or before/after reproduction evidence. The CSS is plausible, but the claim would be stronger with a small visual test (e.g., narrow chat width + resized split-pane + problematic content from #662). Not blocking, but would help future reviewers verify the fix scope.

lefarcen (commented):
Hi @YUHAO-corn! Thanks for this fix 👋 The approach looks solid — adding `overflow-x: hidden` and `min-width: 0` to clip overflowing content is the right pattern.

No blocking issues, just two minor clarity points (see inline below). Both are P3 (non-blocking).

**Code review (Lens A):** No functional issues. The CSS properties are broadly compatible, low-risk, and should work consistently across browsers.

**Reasoning review (Lens B):** The root-cause analysis is plausible but has a small naming mismatch and would benefit from test evidence. See inline comments.

Ready to merge from my side once you address or clarify the inline P3 suggestions. Nice work!

mrcfps (approved):
@YUHAO-corn I verified the changed CSS ranges in `apps/web/src/index.css`: `.chat-log` now clips horizontal overflow, while `.msg` and `.assistant-flow` keep the message layout from expanding past the pane width. Thanks for tightening up this split-pane overflow fix — the change is small, focused, and looks good to merge.

PR #1105, opened 2026-05-09
feat(daemon): HTTP 206 range request support for video/audio files    Fixes #784

ak91456:
Fixes #784

  ## Problem

  Video artifacts generated by AI agents (e.g. Seedance `.mp4` files) were not
  playing inline in the file viewer. The `VideoViewer` component and
  `<video>` element were already present in the frontend, but browsers silently
  refuse to initiate video playback unless the server advertises
  `Accept-Ranges: bytes` and responds to range requests with `206 Partial
  Content`. The daemon was buffering the entire file and returning `200 OK`,
  which blocked playback and seeking.

  ## Solution

  Updated `GET /api/projects/:id/files/*` in the daemon to stream video and
  audio files with full RFC 7233 range-request support. All other file types
  keep the existing buffer path unchanged — no existing behaviour was removed
  or broken.

  ### Changes

  **`apps/daemon/src/projects.ts`**
  - Add `resolveProjectFilePath()` — resolves the safe on-disk path + stat
    without reading file content, so large videos are never buffered in memory.
  - Add `parseByteRange(header, fileSize)` — pure RFC 7233 byte-range parser.
    Returns `{ start, end }` for a satisfiable range, `'unsatisfiable'` (→ 416)
    for out-of-bounds/zero-length ranges, or `null` (→ fall back to full 200)
    for absent/malformed/multi-range headers.

  **`apps/daemon/src/server.ts`**
  - For `video/*` and `audio/*` MIME types: stat the file, parse the `Range`
    header, respond `206` with `Content-Range` + `Accept-Ranges: bytes` for
    partial requests, or `200` for full playback. Streams via
    `fs.createReadStream` — no file is buffered. Stream errors after headers
    are sent destroy the connection cleanly.
  - For all other MIME types: existing behaviour unchanged.

  **`apps/daemon/tests/project-file-range.test.ts`** *(new)*
  - 23 unit tests covering `parseByteRange` (standard, open-ended, suffix,
    multi-range, clamp, 416, malformed) and `resolveProjectFilePath` (MIME
    detection, stat correctness, ENOENT, path traversal rejection).

  ## Test plan

  - [ ] Generate a video via an agent (e.g. image-to-video with Seedance)
  - [ ] Click the `.mp4` file in the output panel — HTML5 player renders inline
  - [ ] Scrub/seek the video — playback continues from the new position
  - [ ] Verify non-media files (HTML, images, text) still open correctly
  - [ ] Run `pnpm --filter @open-design/daemon test` — all tests pass

PR #1113, opened 2026-05-10
fix(daemon): add HTTP range request support for video/audio files

### LINKED ISSUE #784: Video artifacts cannot be previewed inline in the file viewer

xxiaoxiong:
Fixes #784

## Problem

Video and audio artifacts could not be previewed inline in the file viewer. The `VideoViewer` and `AudioViewer` components already existed on the frontend, but the daemon was serving all files by reading the full buffer into memory with no `Accept-Ranges` header. Browsers require HTTP range request support to seek, scrub, or even begin playing video/audio — without it, playback silently fails.

## Solution

- Added `resolveProjectFilePath()` to `apps/daemon/src/projects.ts` to resolve a safe absolute file path without reading the file into memory
- Updated `GET /api/projects/:id/files/*` in `apps/daemon/src/server.ts` to:
  - Detect `video/*` and `audio/*` MIME types
  - Serve them via `fs.createReadStream` with `206 Partial Content` and `Accept-Ranges: bytes` headers when a Range header is present
  - Fall back to full-file streaming when no Range header is present
  - Keep existing buffer-based serving for non-media files (images, HTML, etc.)

## Benefits

- Video artifacts now preview inline with native browser controls
- Seeking/scrubbing works correctly
- Audio files also benefit from the same streaming support
- No memory overhead for large video files
- Non-media files unchanged (backward compatible)

## Test plan

- [ ] Generate a video artifact via an agent (e.g. image-to-video workflow)
- [ ] Click the video file in the file viewer — confirm inline `<video>` player renders with controls
- [ ] Verify seeking/scrubbing works in the player
- [ ] Repeat for an audio file (`.mp3`, `.wav`, `.m4a`)
- [ ] Confirm non-media files (HTML, images, text) are unaffected

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -138,6 +138,7 @@ import {
   readProjectFile,
   renameProjectFile,
   removeProjectDir,
+  resolveProjectFilePath,
   sanitizeName,
   searchProjectFiles,
   writeProjectFile,
@@ -4521,13 +4522,56 @@ export async function startServer({
   app.get('/api/projects/:id/files/*', async (req, res) => {
     try {
       const project = getProject(db, req.params.id);
-      const file = await readProjectFile(
+      const file = await resolveProjectFilePath(
         PROJECTS_DIR,
         req.params.id,
         req.params[0],
         project?.metadata,
       );
-      res.type(file.mime).send(file.buffer);
+
+      // For video and audio files, support HTTP range requests so browsers
+      // can seek/scrub. For other files, read into memory and send as before.
+      const isStreamable = file.mime.startsWith('video/') || file.mime.startsWith('audio/');
+
+      if (isStreamable) {
+        const range = req.headers.range;
+        if (range) {
+          // Parse range header (e.g., "bytes=0-1023")
+          const parts = range.replace(/bytes=/, '').split('-');
+          const start = parseInt(parts[0], 10);
+          const end = parts[1] ? parseInt(parts[1], 10) : file.size - 1;
+          const chunkSize = end - start + 1;
+
+          res.status(206); // Partial Content
+          res.set({
+            'Content-Range': `bytes ${start}-${end}/${file.size}`,
+            'Accept-Ranges': 'bytes',
+            'Content-Length': chunkSize,
+            'Content-Type': file.mime,
+          });
+
+          const stream = fs.createReadStream(file.absolutePath, { start, end });
+          stream.pipe(res);
+        } else {
+          // No range header — send the whole file
+          res.set({
+            'Accept-Ranges': 'bytes',
+            'Content-Length': file.size,
+            'Content-Type': file.mime,
+          });
+          const stream = fs.createReadStream(file.absolutePath);
+          stream.pipe(res);
+        }
+      } else {
+        // Non-streamable files: read into memory and send (existing behavior)
+        const fullFile = await readProjectFile(
+          PROJECTS_DIR,
+          req.params.id,
+          req.params[0],
+          project?.metadata,
+        );
+        res.type(fullFile.mime).send(fullFile.buffer);
+      }
     } catch (err) {
       const status = err && err.code === 'ENOENT' ? 404 : 400;
       sendApiError(
--- a/apps/daemon/src/projects.ts
+++ b/apps/daemon/src/projects.ts
@@ -337,6 +337,29 @@ async function collectArchiveEntries(dir, relDir, out) {
   }
 }
 
+/**
+ * Resolve the absolute file path for a project file without reading it into memory.
+ * Use this for streaming large files (video, audio) via HTTP range requests.
+ */
+export async function resolveProjectFilePath(projectsRoot, projectId, name, metadata?) {
+  const dir = resolveProjectDir(projectsRoot, projectId, metadata);
+  const file = await resolveSafeReal(dir, name);
+  const st = await stat(file);
+  const rel = toProjectPath(path.relative(dir, file));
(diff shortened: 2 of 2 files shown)


mrcfps (commented):
@xxiaoxiong Thanks for adding range-based streaming support for media previews. I found one actionable follow-up in the new range parser that can still break compliant client requests, so I left it inline.


apps/daemon/src/server.ts
```diff
@@ -4521,13 +4522,56 @@ export async function startServer({
...
+      const isStreamable = file.mime.startsWith('video/') || file.mime.startsWith('audio/');
+
+      if (isStreamable) {
+        const range = req.headers.range;
+        if (range) {
+          // Parse range header (e.g., "bytes=0-1023")
+          const parts = range.replace(/bytes=/, '').split('-');
+          const start = parseInt(parts[0], 10);
+          const end = parts[1] ? parseInt(parts[1], 10) : file.size - 1;
+          const chunkSize = end - start + 1;
```
mrcfps, line 4543: Valid suffix and out-of-bounds byte ranges still break this route. A request like `Range: bytes=-1024` leaves `parts[0]` empty here, so `parseInt(parts[0], 10)` becomes `NaN`; that flows into `chunkSize`, `Content-Range`, and `fs.createReadStream`, producing an invalid 206 response instead of the last 1024 bytes. The same path also accepts unsatisfiable ranges such as `bytes=999999-1000000` or `bytes=50-10` without returning `416 Range Not Satisfiable`. That means some standards-compliant media clients can still fail to start or seek even after this change. Please parse suffix/open-ended ranges explicitly, clamp `end` to `file.size - 1`, reject invalid or unsatisfiable ranges with `416` plus `Content-Range: bytes */${file.size}`, and add a route test that covers both `bytes=-N` and an out-of-bounds request.

lefarcen (changes requested):
Hey @xxiaoxiong, thanks for tackling #784! 👋 I found some critical issues that need fixing before this can move forward:

**High-level concern:** PR #1105 (opened 7h before yours) is also fixing #784 with a more complete implementation. I'd recommend coordinating to avoid duplicate review effort — see the specific technical gaps below.

**P1 findings:**

1. **Wrong route modified** — Your changes target `/api/projects/:id/files/*`, but `VideoViewer`/`AudioViewer` actually use `projectFileUrl()` which resolves to `/api/projects/:id/raw/*`. That raw route still buffers the full file without range support, so the reported playback failure isn't fixed. #1105 handles both routes.

2. **Unsafe Range parser** — Current implementation can produce invalid 206 responses for edge cases like `bytes=-500`, `bytes=999999-`, `bytes=10-5`, malformed units, multi-ranges, or zero-length files. RFC 7233 requires falling back to 200 for malformed ranges and returning 416 with `Content-Range: bytes */size` for unsatisfiable ranges. #1105 includes a proper `parseByteRange()` with test coverage.

3. **Missing stream error handling** — If the file is deleted/permissions-changed between `stat()` and `createReadStream()`, the error is emitted outside the try/catch and can crash or break the response. #1105 adds stream error handlers.

Please check PR #1105's implementation for comparison — it includes test coverage for these edge cases. Let me know if you'd prefer to close this as duplicate or coordinate with @ak91456!

xxiaoxiong:
Closing as duplicate of #1105, which was opened ~7 hours earlier and addresses the same issue (HTTP range request support for video/audio files).

Both PRs solve #784 by implementing RFC 7233 range requests in the daemon file server. #1105 should be merged instead.

Apologies for the duplicate — I didn't notice #1105 when searching for related work on this issue.

PR #1119, opened 2026-05-10
fix(security): block IPv4/IPv6 internal-range bypasses in validateExternalApiBaseUrl (CWE-918)

aaronjmars:
## Summary

`validateExternalApiBaseUrl` in `apps/daemon/src/server.ts` blocks SSRF using a prefix-string + literal-set host check. The check misses several internal-network ranges that resolve on linux/macOS and several cloud providers — an attacker controlling a URL passed to this validator can reach loopback, link-local, CGNAT (Tailscale, cloud private), and IPv6 ULA destinations from the daemon.

## Bypass cases

- `0.0.0.0/8` — resolves to localhost on linux/macOS
- `127.0.0.0/8` — entire loopback range (not just `127.0.0.1`)
- `100.64.0.0/10` — CGNAT (Tailscale, AWS, GCP internal)
- IPv4-mapped IPv6: `::ffff:127.0.0.1`, `::ffff:7f00:1`
- IPv6 ULA `fc00::/7` — `fd*` / `fc*`, includes Azure / AWS internal
- IPv6 link-local `fe80::/10`
- IPv6 unspecified `::`
- `*.localhost` subdomains
- Bracketed IPv6 `[::1]` and uppercase casing (`[FE80::1]`)

## Severity: HIGH (CWE-918)

The daemon is reachable from the rendered design surface, so an attacker-controlled `baseUrl` can be supplied via normal product flows.

## Patch

`apps/daemon/src/server.ts` (+59 / −7):
- Replace prefix/literal check with a numeric-IP block-list pass covering every range above.
- Strip bracketed IPv6 form before testing.
- Lowercase hostname before comparison so IPv6 group casing can't bypass.

## Known limitation
Does not fix DNS rebinding — a hostname that resolves to a public IP at validation time and a private IP at fetch time still bypasses. Would need server-side resolve + IP-pinning through to the actual fetch. Worth a follow-up.

## Coordination
Originally queued as a PVR draft on 2026-05-01 — no PVR endpoint was actually available, opening as a public PR.

Filed by aaronjmars (vuln-scanner).

PR #1122, opened 2026-05-10
fix(security): strip trailing dot in normalizeBracketedIpv6 (FQDN SSRF bypass)

aaronjmars:
## Summary

`new URL('http://192.168.1.5./').hostname` returns `'192.168.1.5.'` — the trailing dot is the RFC 1034 absolute-FQDN form and resolves identically to `'192.168.1.5'`. `parseIpv4` fails on the dotted form, so:

- `169.254.169.254.` slips past the AWS metadata-service block
- `192.168.1.5.` slips past the LAN block
- `10.0.0.5.` slips past the private-range block
- `localhost.` slips past the loopback identification

## Fix

Strip trailing dots inside `normalizeBracketedIpv6` so every downstream check (`isLoopbackApiHost`, `isBlockedExternalApiHostname`, `isBlockedIpv4`, IPv6 range tests) sees the canonical form.

```ts
function normalizeBracketedIpv6(hostname: string): string {
  const stripped = hostname.startsWith('[') && hostname.endsWith(']')
    ? hostname.slice(1, -1)
    : hostname;
  return stripped.toLowerCase().replace(/\.+$/, '');
}
```

## Tests

`packages/contracts/tests/connection-test.test.ts` — 6 new cases:
- 3 loopback FQDN forms (`localhost.`, `foo.localhost.`, `127.0.0.1.`) now correctly identified as loopback
- 3 SSRF FQDN bypasses (`169.254.169.254.`, `192.168.1.5.`, `10.0.0.5.`) now correctly blocked

## Relation to #1119

#1119 was filed before main's SSRF refactor into `packages/contracts/src/api/connectionTest.ts` and now mostly duplicates that work. The fe80::/10 review comment on #1119 is **already correct in main** — the regex `fe[89ab][0-9a-f]:` covers fe80–febf which is exactly /10. The only piece of #1119 still genuinely needed against current main is the trailing-dot fix, which this PR provides as a focused 4-line patch + tests. Will close #1119 as superseded once this lands (or earlier on request).

Reported by aaronjmars (vuln-scanner).

### DIFF
--- a/packages/contracts/src/api/connectionTest.ts
+++ b/packages/contracts/src/api/connectionTest.ts
@@ -25,8 +25,9 @@ function normalizeBracketedIpv6(hostname: string): string {
     ? hostname.slice(1, -1)
     : hostname;
   // FQDN trailing-dot form (RFC 1034) resolves identically to the dotless form,
-  // so `localhost.` and `foo.localhost.` must normalize to `localhost` /
-  // `foo.localhost` before any equality / endsWith comparison runs.
+  // so `localhost.` must normalize to `localhost` before the equality check in
+  // isLoopbackApiHost — and `0.0.0.0.`, `10.0.0.1.`, etc. must normalize before
+  // isBlockedIpv4 parses them. Strips one or more trailing dots.
   return stripped.toLowerCase().replace(/\.+$/, '');
 }
 
--- a/packages/contracts/src/api/connectionTest.ts
+++ b/packages/contracts/src/api/connectionTest.ts
@@ -21,9 +21,13 @@ declare const URL: {
 };
 
 function normalizeBracketedIpv6(hostname: string): string {
-  return hostname.startsWith('[') && hostname.endsWith(']')
-    ? hostname.slice(1, -1).toLowerCase()
-    : hostname.toLowerCase();
+  const stripped = hostname.startsWith('[') && hostname.endsWith(']')
+    ? hostname.slice(1, -1)
+    : hostname;
+  // FQDN trailing-dot form (RFC 1034) resolves identically to the dotless form,
+  // so `localhost.` and `foo.localhost.` must normalize to `localhost` /
+  // `foo.localhost` before any equality / endsWith comparison runs.
+  return stripped.toLowerCase().replace(/\.+$/, '');
 }
 
 function parseIpv4(hostname: string): [number, number, number, number] | null {
--- a/packages/contracts/tests/connection-test.test.ts
+++ b/packages/contracts/tests/connection-test.test.ts
@@ -9,6 +9,12 @@ describe('provider base URL validation', () => {
       'http://127.0.0.1:11434/v1',
       'http://[::1]:11434/v1',
       'http://[::ffff:127.0.0.1]:11434/v1',
+      // FQDN trailing-dot form must be treated as loopback, not as an
+      // unknown public host (`new URL('http://localhost./').hostname`
+      // returns `'localhost.'`).
+      'http://localhost./v1',
+      'http://foo.localhost./v1',
+      'http://127.0.0.1.:11434/v1',
     ]) {
       expect(validateBaseUrl(baseUrl).error).toBeUndefined();
     }
@@ -27,6 +33,12 @@ describe('provider base URL validation', () => {
       'http://[fd00::1]:11434/v1',
       'http://[fe80::1]:11434/v1',
       'http://[::ffff:192.168.1.5]:11434/v1',
+      // Trailing-dot FQDN bypass: `192.168.1.5.` doesn't parse as IPv4
+      // until normalized, so without the trailing-dot strip an attacker
+      // can hit AWS metadata via 169.254.169.254. and reach LAN ranges.
+      'http://169.254.169.254./latest/meta-data',
+      'http://192.168.1.5.:11434/v1',
+      'http://10.0.0.5.:11434/v1',
     ]) {
       expect(validateBaseUrl(baseUrl)).toMatchObject({
         error: 'Internal IPs blocked',
--- a/packages/contracts/tests/connection-test.test.ts
+++ b/packages/contracts/tests/connection-test.test.ts
@@ -1,5 +1,8 @@
 import { describe, expect, it } from 'vitest';
-import { validateBaseUrl } from '../src/api/connectionTest';
+import {
+  isLoopbackApiHost,
+  validateBaseUrl,
+} from '../src/api/connectionTest';
 
 describe('provider base URL validation', () => {
   it('allows public endpoints and loopback local providers', () => {
@@ -9,14 +12,17 @@ describe('provider base URL validation', () => {
       'http://127.0.0.1:11434/v1',
       'http://[::1]:11434/v1',
       'http://[::ffff:127.0.0.1]:11434/v1',
-      // FQDN trailing-dot form must be treated as loopback, not as an
-      // unknown public host (`new URL('http://localhost./').hostname`
-      // returns `'localhost.'`).
-      'http://localhost./v1',
-      'http://foo.localhost./v1',
-      'http://127.0.0.1.:11434/v1',
     ]) {
       expect(validateBaseUrl(baseUrl).error).toBeUndefined();
+    }
+  });
+
+  it('identifies trailing-dot FQDN forms of loopback hosts as loopback', () => {
+    // Direct assertion against isLoopbackApiHost — validateBaseUrl alone
+    // can't distinguish "passed because loopback" from "passed because
+    // not blocked", which the previous test revision conflated.
+    for (const host of ['localhost.', '127.0.0.1.', '127.0.0.5.']) {
+      expect(isLoopbackApiHost(host)).toBe(true);
     }
   });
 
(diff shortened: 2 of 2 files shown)


packages/contracts/tests/connection-test.test.ts
```diff
@@ -9,6 +9,12 @@ describe('provider base URL validation', () => {
       'http://127.0.0.1:11434/v1',
       'http://[::1]:11434/v1',
       'http://[::ffff:127.0.0.1]:11434/v1',
+      // FQDN trailing-dot form must be treated as loopback, not as an
+      // unknown public host (`new URL('http://localhost./').hostname`
+      // returns `'localhost.'`).
+      'http://localhost./v1',
+      'http://foo.localhost./v1',
```
mrcfps: `foo.localhost.` does not currently exercise the loopback path that this comment describes. `validateBaseUrl()` returns success for any hostname that is simply *not blocked*, so this assertion still passes even if the host is classified as remote. We can see that in the downstream checks: `isLoopbackApiHost()` only accepts exact `localhost`, `::1`, and loopback IPs today, and both `validateLocalOpenAiModel()` (`apps/daemon/src/connectionTest.ts`) and provider-model scoping (`apps/daemon/src/providerModels.ts`) key off that helper. That means this new case makes the test summary/PR summary say `.localhost` subdomains are recognized as loopback, but the implementation still treats them as remote. Could we either extend `isLoopbackApiHost()` to accept the `.localhost` suffix, or drop this case and reword the comment so the test only covers hosts that are actually classified as loopback today?

mrcfps (commented):
@aaronjmars Thanks for tightening the trailing-dot normalization here — I verified the core SSRF bypass fix, but I found one non-blocking test expectation that currently overstates the `.localhost` loopback behavior.


packages/contracts/tests/connection-test.test.ts
```diff
@@ -27,6 +33,12 @@ describe('provider base URL validation', () => {
       'http://[fd00::1]:11434/v1',
       'http://[fe80::1]:11434/v1',
       'http://[::ffff:192.168.1.5]:11434/v1',
+      // Trailing-dot FQDN bypass: `192.168.1.5.` doesn't parse as IPv4
+      // until normalized, so without the trailing-dot strip an attacker
+      // can hit AWS metadata via 169.254.169.254. and reach LAN ranges.
+      'http://169.254.169.254./latest/meta-data',
```
lefarcen: **P3 — Test coverage gap (non-blocking):** The fix normalizes trailing dots for ALL blocked IPv4 ranges (`0.0.0.0/8`, `100.64.0.0/10`, `172.16.0.0/12`, multicast ≥224), but the new tests only cover three dotted forms (AWS metadata, LAN, private). Add a couple more representative cases to lock down the full coverage:

```ts
'http://0.0.0.0.:11434/v1',        // 0.0.0.0/8
'http://172.16.0.5.:11434/v1',    // carrier-grade NAT
'http://localhost..:11434/v1',     // multiple trailing dots
```

Not a correctness bug — the regex handles them — just nice to have explicit proof in the suite.

packages/contracts/src/api/connectionTest.ts
```diff
@@ -21,9 +21,13 @@ declare const URL: {
...
 function normalizeBracketedIpv6(hostname: string): string {
-  return hostname.startsWith('[') && hostname.endsWith(']')
-    ? hostname.slice(1, -1).toLowerCase()
-    : hostname.toLowerCase();
+  const stripped = hostname.startsWith('[') && hostname.endsWith(']')
+    ? hostname.slice(1, -1)
+    : hostname;
+  // FQDN trailing-dot form (RFC 1034) resolves identically to the dotless form,
+  // so `localhost.` and `foo.localhost.` must normalize to `localhost` /
+  // `foo.localhost` before any equality / endsWith comparison runs.
```
lefarcen: **P3 — PR body wording (non-blocking):** The summary lists four bypass examples but reads like an exhaustive list. In reality, the same trailing-dot bypass applies to ALL `isBlockedIpv4()` ranges: `0.0.0.0/8`, `100.64.0.0/10`, `172.16.0.0/12`, multicast ≥224. Your implementation correctly handles them through the shared normalization path; it's just a reasoning/completeness thing. 

Consider adding "(and all other blocked IPv4 ranges)" after the bullet list in the PR body so future readers know the fix is universal, not scoped to those four examples only.

lefarcen (commented):
Hey @aaronjmars, thanks for this focused security fix! 👍

The trailing-dot normalization is the right shape — `replace(/\.+$/, '')` covers single and repeated forms, runs after bracket stripping, and feeds all downstream loopback/private checks. IPv6 trailing-dot forms like `http://[::1]./` are already rejected by `new URL()` before validation, so no new IPv6 bypass risk.

I verified your #1119 fe80::/10 claim against `connectionTest.ts:104`: `^fe[89ab][0-9a-f]:` does cover fe80–febf (exactly /10), so main's IPv6 link-local logic is correct.

Two optional test-coverage suggestions below. No blockers; this is ready from my side.

lefarcen:
Agent-to-agent review note for PR #1122:

No P1/P2 findings. The trailing-dot normalization is the right fix shape for the FQDN SSRF bypass: stripping trailing dots after IPv6 bracket normalization feeds canonical hostnames into loopback/private-range checks, including multiple trailing dots.

P3: Test coverage could add representative trailing-dot cases for the other existing blocked IPv4 ranges (`0.0.0.0.`, `100.64.0.1.`, `172.16.0.5.`, `224.0.0.1.`) and a multiple-dot case like `localhost..`. The implementation already covers these through the shared regex; this would just lock down the intended breadth.

P3: PR body should make clear the listed bypass examples are representative, not exhaustive. The same trailing-dot issue applied to every `isBlockedIpv4()` range before normalization, not just metadata/LAN/private examples.

Verified reasoning claim: `^fe[89ab][0-9a-f]:` in `packages/contracts/src/api/connectionTest.ts` covers `fe80` through `febf`, so the #1119 fe80::/10 claim is correct in current main.

aaronjmars:
Pushed [8509a97](https://github.com/nexu-io/open-design/pull/1122/commits/8509a97) addressing the review feedback:

**P2 (@mrcfps + codex)** — fair catch. The `foo.localhost.` case asserted `error===undefined` on `validateBaseUrl`, which only proves the URL passed validation, not that the host is identified as loopback. Replaced with a dedicated test that calls `isLoopbackApiHost` directly on `localhost.`, `127.0.0.1.`, and `127.0.0.5.`, so the assertion exercises the loopback path the comment claims.

**P3 (@lefarcen, blocked-range coverage)** — added a dedicated case per `isBlockedIpv4` range so the trailing-dot strip is tested against the full surface: `0.0.0.0/8`, `10/8`, `100.64/10` (CGNAT), `169.254/16` (metadata), `172.16/12`, `192.168/16`, multicast `>=224`. Picked one representative IP per range rather than fan-out so the diff stays small.

**P3 (PR body wording)** — agreed it reads exhaustive. The fix path through `normalizeBracketedIpv6` is range-agnostic so it covers everything `isBlockedIpv4` returns true for; the body listed only the high-impact examples (metadata, LAN, private). Happy to update the description if helpful.

mrcfps (dismissed):
@aaronjmars I re-checked the current head and verified the trailing-dot normalization now feeds canonical hostnames into both the loopback and blocked-range checks, with updated tests covering the loopback assertions and each blocked IPv4 range variant. I also ran the `@open-design/contracts` connection tests and typecheck against this PR head locally; everything passed. Thanks for tightening this security fix and following through on the review feedback. 👍


packages/contracts/src/api/connectionTest.ts
```diff
@@ -21,9 +21,13 @@ declare const URL: {
 };
 
 function normalizeBracketedIpv6(hostname: string): string {
-  return hostname.startsWith('[') && hostname.endsWith(']')
-    ? hostname.slice(1, -1).toLowerCase()
-    : hostname.toLowerCase();
+  const stripped = hostname.startsWith('[') && hostname.endsWith(']')
+    ? hostname.slice(1, -1)
+    : hostname;
+  // FQDN trailing-dot form (RFC 1034) resolves identically to the dotless form,
```
lefarcen, line 27: The explanation still mentions `foo.localhost.` and an `endsWith` comparison, but `isLoopbackApiHost` only accepts exact `localhost`, `::1`, loopback IPv4, and mapped loopback IPv4. If `*.localhost.` is intended to be loopback, the code and direct test should cover `foo.localhost.`; otherwise this comment should drop the subdomain/`endsWith` claim.

lefarcen (commented):
@aaronjmars Thanks for addressing the review feedback! 👍

The new test structure is correct: `isLoopbackApiHost` is now directly called (fixing mrcfps's P2), and the blocked-range coverage is complete — every branch in `isBlockedIpv4` has a representative trailing-dot case.

One non-blocking observation below about the inline comment still referencing `foo.localhost.` when that subdomain path doesn't exist in the implementation.

mrcfps (approved):
@aaronjmars I re-checked the trailing-dot normalization on the current head and verified the updated `connectionTest` coverage now exercises both the loopback path and each blocked IPv4 range that previously allowed the dotted-host bypass. I also ran the focused contracts connection tests against this PR head to confirm the new cases pass. Thanks for tightening up this security fix and following through on the earlier review feedback. 👍

PR #1127, opened 2026-05-10
feat(web): add manual removal for captured components in Pods (#802)

YUHAO-corn:
Closes #802

## Problem
Once components are captured into a pod, there is no way to manually remove an individual member. A mistaken capture pollutes the pod and forces users to recreate the entire pod.

## Solution
Add a per-member remove button to each captured component chip in the Pods panel.

## Changes
- `BoardComposerPopover`: adds a member removal callback. Each captured member chip now renders a small close-icon button that removes the member when clicked.
- `FileViewer` / `HtmlViewer`: filters the removed member from the pod, rebuilds the aggregate snapshot metadata from the remaining members, and closes the board composer when no members remain.
- CSS: keeps chips in a compact wrapping row and adds a subtle remove affordance.
- i18n: adds the remove-member label key and corresponding type entry.

## UX notes
- The remove button is intentionally lightweight: a small circular close icon in each chip.
- When the last member is removed, the composer closes automatically since an empty pod has no meaning in the curation flow.
- Members are removed by id so the operation is stable even if the list reorders.

## Surface area
- [x] **UI** — Pods captured-member chips gain a remove affordance
- [x] **i18n keys** — adds the captured-member remove label
- [ ] **New top-level dependency** — adding any new entry to the root `package.json`
- [x] **Default behavior change** — existing Pods can now be manually pruned without recreating them

## Bug fix verification
- Regression scenario: captured Pods could not be manually corrected after an accidental member capture.
- Reviewer re-check confirmed the stale aggregate snapshot fields are rebuilt from remaining members and no removed-member context leaks into saved/sent comment attachments.

## Validation
- GitHub CI on the latest head is green: `build` and `Validate workspace` passed.
- Reviewer re-check confirmed the stale snapshot fields are rebuilt from remaining members and no removed-member context leaks into saved/sent comment attachments.


lefarcen:
Hi @YUHAO-corn! 🎉
Thanks for the contribution — this addresses a real pain point for pod curation workflow.

— open-design team

apps/web/src/components/FileViewer.tsx
```diff
@@ -4793,6 +4803,16 @@ function HtmlViewer({
                 onRemoveQueuedNote={(index) =>
                   setQueuedBoardNotes((current) => current.filter((_, currentIndex) => currentIndex !== index))
                 }
+                onRemovePodMember={(elementId) => {
+                  const current = activeCommentTarget;
+                  if (!current) return;
+                  const nextMembers = (current.podMembers ?? []).filter((m) => m.elementId !== elementId);
```
lefarcen, line 6085: **P1** — Stale snapshot fields after member removal.

Removing a pod member updates `podMembers` and `memberCount`, but leaves the aggregate snapshot fields (`selector`, `label`, `text`, `position`, `htmlHint`) unchanged. These fields were built from the original member set. When the user saves/sends the comment, `targetFromSnapshot(activeCommentTarget)` still uses the original snapshot, so the removed component can reappear in the chat attachment context.

Fix: recompute the pod snapshot from `nextMembers` before calling `setActiveCommentTarget`, or derive pod attachment fields from `podMembers` at send time when `selectionKind === 'pod'`. Add a FileViewer/comments test that removes a member and asserts the sent attachment excludes it from both `podMembers` and the top-level selector/context.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4793,6 +4803,16 @@ function HtmlViewer({
...
+                  if (nextMembers.length === 0) {
```
lefarcen, line 6086: **P2** — Silent data loss on last-member removal.

Removing the last member calls `clearBoardComposer()`, which also clears `commentDraft` and `queuedBoardNotes`. If the user accidentally removes the last chip while they've already typed notes or queued items, those are silently lost.

Fix options:
- Preserve draft/notes when closing (pass them to `clearBoardComposer`)
- Show an empty-pod state with the draft intact (disable save until members exist)
- Disable last-member removal while notes exist
- Confirm the discard before calling `clearBoardComposer()`

apps/web/src/components/FileViewer.tsx
```diff
@@ -1480,9 +1482,17 @@ function BoardComposerPopover({
         <div className="board-pod-summary">
           <strong>{target.memberCount || podMembers.length} captured items</strong>
           <div className="board-pod-members">
-            {podMembers.slice(0, 6).map((member) => (
+            {podMembers.map((member) => (
```
lefarcen: **P2** — Unbounded member list height.

The prior code capped the display at 6 members (`.slice(0, 6)`). This change removes that cap and renders all members inside a `width: min(320px, ...)` popover, but the CSS only wraps chips horizontally and does not cap height. Large pods (say 30 items) will push the textarea/actions out of view. The previous six-item constraint likely existed for this reason.

Fix: add a bounded scroll region for `.board-pod-members` (e.g., `max-height: 120px; overflow-y: auto`), implement an expandable "show all" design, or document why the previous constraint was removed.

apps/web/src/components/FileViewer.tsx
```diff
@@ -1480,9 +1482,17 @@ function BoardComposerPopover({
...
               <span key={member.elementId} className="board-pod-chip">
                 {summarizeMember(member)}
+                <button
```
lefarcen: **P2** — Generic a11y label.

Each remove button has the same `title="Remove captured component"`, so assistive users hear repeated indistinguishable controls. The `Icon` is `aria-hidden`, so the button should have a specific accessible name.

Fix: add an `aria-label` that includes the member summary or label, e.g., `aria-label={`Remove captured component: ${summarizeMember(member)}`}`.

apps/web/src/index.css
```diff
@@ -8508,13 +8511,35 @@ button.connector-action.is-loading {
...
+  align-items: center;
+  gap: 4px;
   padding: 3px 7px;
   border: 1px solid var(--border-soft);
   border-radius: var(--radius-pill);
   background: var(--bg-panel);
   color: var(--text-muted);
   font-size: 11px;
 }
+.board-pod-chip-remove {
```
lefarcen, line 10251: **P3** — Hover affordance mismatch.

The PR body says the remove button is "visible on hover," but the CSS renders it always visible and only changes color/background on hover. Either update the PR description to match the implementation, or implement the stated hover/focus affordance with opacity/visibility rules (while keeping the button visible on keyboard focus for a11y).

lefarcen (dismissed):
Hi @YUHAO-corn! Thanks for tackling pod member removal — this is a real workflow gap.

**Found 1 critical issue + 3 design gaps** that need addressing before merge. The implementation is almost there, but the stale snapshot fields (P1) will cause removed members to leak into chat attachments, and the unbounded height + silent draft discard (P2) can surprise users. See inline comments below for specifics.

Once those are fixed, this should be ready. Let me know if any of the recommendations need clarification!

apps/web/src/components/FileViewer.tsx
```diff
@@ -4793,6 +4803,16 @@ function HtmlViewer({
...
                   setQueuedBoardNotes((current) => current.filter((_, currentIndex) => currentIndex !== index))
                 }
+                onRemovePodMember={(elementId) => {
+                  const current = activeCommentTarget;
+                  if (!current) return;
+                  const nextMembers = (current.podMembers ?? []).filter((m) => m.elementId !== elementId);
+                  if (nextMembers.length === 0) {
+                    clearBoardComposer();
+                  } else {
+                    setActiveCommentTarget({ ...current, podMembers: nextMembers, memberCount: nextMembers.length });
```
mrcfps: Removing a member here only updates `podMembers` and `memberCount`, but `savePersistentComment` / `sendBoardBatch` later serialize `activeCommentTarget` through `targetFromSnapshot`, which still reads `selector`, `label`, `text`, `position`, and `htmlHint` from this stale snapshot. That means a chip can disappear from the UI while the saved/sent attachment still carries context from the removed element, so the main removal flow is still incorrect on this head. Please rebuild the pod-level metadata from `nextMembers` before calling `setActiveCommentTarget` here (for example by extracting the aggregation logic from `buildPodSnapshot` into a helper and reusing it for member removal).

apps/web/src/i18n/types.ts
```diff
@@ -759,6 +759,7 @@ export interface Dict {
   'chat.comments.updateSend': string;
   'chat.comments.removeAttachment': string;
   'chat.comments.removeAttachmentAria': string;
+  'chat.comments.removePodMember': string;
```
mrcfps, line 929: Adding a required `Dict` key here without updating every locale breaks the workspace typecheck on this head. `Validate workspace` is currently red with `src/i18n/locales/zh-CN.ts(3,14): error TS2741: Property 'chat.comments.removePodMember' is missing...`, so this PR cannot merge cleanly as-is. Please add `chat.comments.removePodMember` to `apps/web/src/i18n/locales/zh-CN.ts` (and any other locale files implementing `Dict`) in the same change.

mrcfps (changes requested):
@YUHAO-corn Thanks for pushing this — the UX improvement makes sense. I found two blockers on the current head: one correctness issue in the pod snapshot update path, and one missing locale update that currently breaks `apps/web` typecheck. Once those are fixed, this should be in good shape.


apps/web/src/components/FileViewer.tsx
```diff
@@ -4793,6 +4806,16 @@ function HtmlViewer({
...
                   setQueuedBoardNotes((current) => current.filter((_, currentIndex) => currentIndex !== index))
                 }
+                onRemovePodMember={(elementId) => {
+                  const current = activeCommentTargetRef.current;
+                  if (!current) return;
+                  const nextMembers = (current.podMembers ?? []).filter((m) => m.elementId !== elementId);
+                  if (nextMembers.length === 0) {
+                    clearBoardComposer();
+                  } else {
+                    setActiveCommentTarget({ ...current, podMembers: nextMembers, memberCount: nextMembers.length });
```
mrcfps: Removing a chip here still leaves the pod snapshot's aggregate fields (`selector`, `label`, `text`, `position`, and `htmlHint`) unchanged. `savePersistentComment` and `sendBoardBatch` later serialize `activeCommentTarget` through `targetFromSnapshot(...)` / `buildBoardCommentAttachments(...)`, so those stale top-level fields can still carry context from the removed member even though `podMembers` was filtered in the UI. That means the main removal flow is still incorrect on this head: a component can disappear from the chip list while the saved comment or chat attachment still includes its selector/text/bounds.

Please rebuild the pod snapshot from `nextMembers` before calling `setActiveCommentTarget` here (for example by extracting the aggregation logic out of `buildPodSnapshot` and reusing it for member removal), and add a regression test that removes one pod member then asserts the saved/sent attachment no longer includes that member's selector/text/position/html hint.

mrcfps (changes requested):
@YUHAO-corn Thanks for the quick follow-up here — the overflow guard, ref update, and missing locale fix all look good. I rechecked the latest head and found one blocker still in the main pod-member-removal path: the composer still keeps stale pod-level snapshot metadata after a chip is removed, so the saved/sent attachment can still include context from the removed element. Once that snapshot is rebuilt from the remaining members, this should be in good shape.


lefarcen (commented):
@YUHAO-corn Good progress on the zh-CN locale + aria-label + max-height fixes! 👍

The P1 snapshot staleness issue is **not fully fixed yet** — the handler rebuilds `position` correctly, but `text`, `htmlHint`, `selector`, and `label` are still stale:

- Line 4825-4826: `text: nextMembers[0]?.text || current.text` falls back to the old aggregate when the first remaining member has empty text
- `selector` and `label` remain unchanged from `current`, so they can still reference removed members

These fields flow into `targetFromSnapshot` → chat attachments, so removed components can still leak.

**Fix**: rebuild all 5 aggregate fields (`selector`, `label`, `text`, `htmlHint`, `position`) from `nextMembers`, not from `current` or only the first member. mrcfps's review at bbfcfcf6 described the same issue — once that's addressed, this should be ready.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4793,6 +4806,27 @@ function HtmlViewer({
...
+                    const minY = Math.min(...nextMembers.map((m) => m.position.y));
+                    const maxX = Math.max(...nextMembers.map((m) => m.position.x + m.position.width));
+                    const maxY = Math.max(...nextMembers.map((m) => m.position.y + m.position.height));
+                    setActiveCommentTarget({
+                      ...current,
+                      podMembers: nextMembers,
+                      memberCount: nextMembers.length,
+                      position: { x: minX, y: minY, width: maxX - minX, height: maxY - minY },
+                      text: nextMembers[0]?.text || current.text,
+                      htmlHint: nextMembers[0]?.htmlHint || current.htmlHint,
```
mrcfps: This update still leaves stale pod-level snapshot fields behind. `targetFromSnapshot(...)` and `buildBoardCommentAttachments(...)` serialize the top-level `selector`, `label`, `text`, `position`, and `htmlHint` from `activeCommentTarget`, not just `podMembers`. In this hunk only `position` is fully recomputed; `text`/`htmlHint` now fall back to the old aggregate when the first remaining member is empty, and `selector`/`label` are still preserved unchanged via `...current`. That means a chip can disappear from the UI while the saved comment or chat attachment still carries selector/text/html-hint context from the removed component. Please rebuild all aggregate pod fields from `nextMembers` using the same aggregation rules as `buildPodSnapshot(...)` (combined selector, summary label, concatenated text/html hint, and bounds), and add a regression test in `apps/web/tests/comments.test.ts` or `apps/web/tests/components/FileViewer.test.tsx` that removes one member and asserts the serialized attachment no longer includes that member's metadata.

mrcfps (changes requested):
@YUHAO-corn Thanks for the follow-up here — the locale fix, overflow guard, and bounds recompute all look good. I rechecked the latest head and found one blocker still in the pod-member-removal path: saved/sent pod attachments can still keep stale aggregate metadata from a removed member. Once the snapshot is rebuilt from the remaining members with the same aggregation rules as pod creation, this should be ready.


lefarcen (commented):
@YUHAO-corn Perfect — the pod member removal path now rebuilds all 5 aggregate snapshot fields (selector, label, text, htmlHint, position) from `nextMembers`, so removed-member context no longer leaks into saved/sent comment attachments. I found no new blocking issues in the latest diff. Deferring final approval to a maintainer. Thanks for the thorough iteration!

apps/web/src/components/FileViewer.tsx
```diff
@@ -4793,6 +4806,56 @@ function HtmlViewer({
...
+                      .slice(0, 180);
+                    setActiveCommentTarget({
+                      ...current,
+                      podMembers: nextMembers,
+                      memberCount: nextMembers.length,
+                      selector: combinedSelector || 'body *',
+                      label: summary || `Pod of ${nextMembers.length} items`,
+                      position: { x: minX, y: minY, width: maxX - minX, height: maxY - minY },
+                      text: combinedText,
+                      htmlHint: combinedHtmlHint,
```
mrcfps: The aggregation rules are now correct again, but this whole block is a second hand-written copy of the pod snapshot logic from `buildPodSnapshot(...)` / `summarizeSnapshot(...)` earlier in this file. That duplication is already what let the first versions of this PR drift: the removal path and the creation path had to be fixed separately across multiple follow-up commits. Keeping the selector/label/text/htmlHint/position rules duplicated here means the next tweak to pod snapshot semantics can easily reintroduce the same mismatch between what the chip list shows and what `targetFromSnapshot(...)` serializes. Please extract a shared helper that builds the aggregate pod fields from a member list (or reuse `buildPodSnapshot` through a smaller helper) so pod creation and pod-member removal stay locked to the same rules in one place.

mrcfps (commented):
@YUHAO-corn Thanks for pushing the follow-up fixes here — the stale snapshot bug, locale gap, and overflow issue all look addressed on this head. I have one non-blocking maintainability follow-up in the pod snapshot rebuild path before this is fully settled.


lefarcen (commented):
Hey @YUHAO-corn! The Problem/Solution write-up is clear, and the UX notes on the chip hover affordance and auto-close-on-last-removal are a nice touch. Two template sections that would help reviewers scope this before merge:

- **Surface area checklist** — could you add a `## Surface area` section and tick the relevant box(es)? This PR visibly adds a remove button to the Pods chip UI, so at minimum **UI** applies. Ticking the right boxes lets the product review land faster.
- **Validation** — the current `### Verification` reads more as scope reasoning than an actual test run. Could you add what you concretely ran (e.g. `pnpm typecheck`, `pnpm guard`, smoke-tested the chip removal + auto-close flow in a local dev session)?

Neither blocks code review, but both help the product reviewer and release verification do their job quickly. ✨

YUHAO-corn:
Updated the PR body with the full Surface area checklist and a concrete Verification section.

lefarcen (commented):
One last body item @YUHAO-corn — the `## Surface area` section now has **UI** checked, which means the template asks for at least one screenshot to anchor the visual context. Could you drop a capture of the Pods chip with the remove button showing? On-hover state is what reviewers need to see (since that's when it appears). A before/after pair is ideal, but a single after-state image works too. Everything else in the body looks complete. ✨

YUHAO-corn:
Merged latest main to clear the conflict and added the requested Pods chip remove affordance capture.

Pods chip remove hover

Validation on the merge head `4563fc5d`:
- `pnpm --filter @open-design/web test -- tests/components/FileViewer.test.tsx --runInBand` (129 files, 1277 tests passed)

apps/web/src/components/FileViewer.tsx
```diff
@@ -1901,9 +1904,18 @@ function BoardComposerPopover({
...
           <div className="board-pod-members">
-            {podMembers.slice(0, 6).map((member) => (
+            {podMembers.map((member) => (
               <span key={member.elementId} className="board-pod-chip">
                 {summarizeMember(member)}
+                <button
+                  type="button"
+                  className="board-pod-chip-remove"
+                  title={t('chat.comments.removePodMember')}
+                  aria-label={t('chat.comments.removePodMember')}
```
mrcfps: Each remove button in this loop gets the same `aria-label`/`title`, so a screen reader lands on a series of identical "Remove captured component" controls with no indication of which chip each one belongs to. In this changed range the button itself only contains the close icon, so the accessible name needs to carry the member identity or the main pod-pruning flow stays ambiguous for assistive users. Please switch this to a member-specific label here (for example a `{name}` i18n key that uses `summarizeMember(member)` or `member.label`) so each control announces the component it will remove.

mrcfps (commented):
@YUHAO-corn Thanks for the follow-up here — I rechecked the shared pod aggregate helper, the removal-path wiring, and the new regression coverage on this head. The stale snapshot leak looks fixed now. I have one small accessibility follow-up in the chip-removal UI before this is fully settled.


YUHAO-corn:
Addressed the latest pod chip accessibility follow-up.

What changed:
- Pod member remove controls now use a localized `aria-label`/`title` that includes the specific captured member identity.
- The accessible label uses a fuller member description than the visual chip summary, including element id when useful, so repeated controls are distinguishable.
- Added regression coverage for member-specific accessible names and matching titles.

Validation:
- `pnpm --filter @open-design/web test -- tests/components/FileViewer.test.tsx --runInBand`
- GitHub CI is green on `7509e55f`.

mrcfps (approved):
@YUHAO-corn Thanks for the follow-up here — I rechecked the shared pod aggregate helper, the pod-member removal wiring, and the new accessibility/test coverage on this head. The removal flow now rebuilds pod metadata from the remaining members, and `pnpm --filter @open-design/web typecheck` plus `pnpm --filter @open-design/web test -- tests/comments.test.ts tests/components/FileViewer.test.tsx --runInBand` both passed locally. Nice work iterating on this. 👍


lefarcen:
Heads-up: PR #1951 is now also open against #802. It overlaps this PR in `apps/web/src/components/FileViewer.tsx` / `apps/web/src/index.css`, but takes a different implementation shape by extracting `BoardComposerPopover` and keeping the single-chip removal state in the composer layer.

No action needed from you right now — sharing the cross-link so reviewers can compare both approaches and avoid duplicated effort while the maintainer team decides which path should land.

PR #1132, opened 2026-05-10
fix(web): ignore <artifact> tags inside markdown code spans and fences

### LINKED ISSUE #1130: bug: streaming artifact parser swallows text after literal <artifact substring in prose
### Summary

The streaming artifact parser in the web client treats any occurrence of the literal substring `<artifact ` (followed by whitespace) as the start of a real artifact tag, regardless of whether it appears inside a Markdown code fence, an inline backtick span, or other prose context. As a result, when the model legitimately discusses or quotes the artifact tag syntax in a reply, the parser enters "inside artifact" state and routes the rest of the response into a non-existent artifact buffer until it sees `</artifact>`. The chat UI freezes mid-message even though the model finishes generating normally.

### Environment

- Open Design version: latest `main`
- Observed in the web chat UI

### Reproduction

1. Ask the assistant something that causes it to quote the artifact tag literally, e.g. "what does it mean to wrap a reply in `` `<artifact identifier="x">` ``?"
2. The assistant emits the backtick-quoted tag as part of normal prose.
3. The chat UI stops rendering at the backtick. The status bar continues to show streaming activity (e.g. "completed 1m 42s · 3164 output tokens"), but no further content appears in the chat panel.

### Source-level observation

`apps/web/src/artifacts/parser.ts`, `findOpenTag` (~L45-87):

```ts
const idx = buffer.indexOf(OPEN_PREFIX, from);
```

The scan uses raw `indexOf('<artifact')` over the full buffer with the only guard being "next character is whitespace" (to skip false matches like `<artifactual>`). There is no awareness of:

- Markdown fenced code blocks (`` ``` ``)
- Inline code spans (single or multiple backticks)
- Whether the tag starts at a line boundary

Once `inside` becomes true (~L121), all subsequent content streams into `state.content` until `</artifact>` is seen, which suppresses the rest of the assistant's prose from rendering.

### Expected behavior

The parser should only treat `<artifact` as a real opening tag when it occurs in an HTML-block-like context, similar to how Markdown parsers gate raw HTML blocks. Concretely:

- Skip occurrences inside fenced code blocks (`` ``` `` … `` ``` ``)
- Skip occurrences inside inline code spans (`` ` `` … `` ` ``)
- Optionally require the opening tag to start at the beginning of a line (after optional whitespace)

This brings the parser in line with how the model can safely *talk about* the artifact protocol without accidentally triggering it.

### Related

- #313 — same family (literal protocol markup leaking into rendered text), different surface (Anthropic API mode tool events)
- #50 — independent persistence-layer bug, but a corrupted artifact stream caused by this parser bug can be silently written to disk; commented separately on #50

mutoe:
## Summary

Fixes #1130. The streaming artifact parser (`apps/web/src/artifacts/parser.ts`) treated any literal `<artifact ` substring in the buffer as the start of a real artifact tag, regardless of whether it appeared inside a Markdown code span or fenced block. When the model recited the protocol verbatim — for example explaining how `apps/daemon/src/prompts/official-system.ts`'s `<artifact identifier="kebab-slug" …>` template works — the parser flipped into artifact mode, suppressed the rest of the rendered reply, and (when a closing `</artifact>` appeared in the recitation) silently wrote a spurious file to disk via `persistArtifact`.

## Approach

Replace `findOpenTag`'s raw `indexOf` scan with a linear walk that tracks Markdown context:

- skip any `<artifact` prefix found inside a fenced code block (` ``` `)
- skip any `<artifact` prefix found inside an inline code span (`` ` ``)
- if the buffer ends mid-fence, return a `partial` match anchored at the fence start so the next streaming chunk can resolve the boundary without losing fence context (matches the existing partial-tag pattern)

Triple-backtick is checked before single-backtick, and inline detection is suppressed inside an open fence — so the standard `` ```html `` … `` ``` `` shape doesn't collapse into pairs of inline spans.

## Tests

New file `apps/web/tests/artifacts/parser.test.ts` with four cases:

- **regression baseline**: a real `<artifact>` in plain prose still parses (start/chunk/end events, trailing text preserved)
- inline backtick recitation no longer enters artifact mode
- fenced code block recitation (open + content + close all inside the fence) no longer enters artifact mode
- the same fenced case fed in three streaming chunks (split mid-tag and mid-`</artifact>`) also no longer enters artifact mode

All four pass with the fix; the latter three fail without it.

## Verification

- `pnpm --filter @open-design/web exec vitest run tests/artifacts/parser.test.ts` — 4 passed
- `pnpm --filter @open-design/web typecheck` — clean
- `pnpm --filter @open-design/web test` — same pass/fail set as `main` (the four `SettingsDialog.execution.test.tsx` failures are pre-existing `window.localStorage` env issues unrelated to this change)
- Manual reproduction in the desktop app: asking the assistant to document the artifact protocol now renders the full reply (1448 output tokens vs. the previously-truncated 1300) and no longer creates a spurious file in the project file panel

## Test plan

- [ ] CI green on `apps/web` package
- [ ] Manual: in any project, ask the assistant to explain the artifact tag syntax — verify the chat renders to completion and no junk file appears in the file panel

### DIFF
--- a/apps/web/src/artifacts/markdown-context.ts
+++ b/apps/web/src/artifacts/markdown-context.ts
@@ -9,8 +9,13 @@
  * literal Markdown and must not be treated as a real protocol tag.
  */
 
-// Line-anchored fence delimiter, mirrors runtime/markdown.tsx:44.
-export const FENCE_LINE_RE = /^[ ]{0,3}```(\w[\w+-]*)?\s*$/;
+// Line-anchored fence delimiters, mirror runtime/markdown.tsx:44 (open) and
+// runtime/markdown.tsx:49 (close). The renderer is asymmetric on purpose:
+// an opening fence may carry an info string (e.g. ```html), a closing fence
+// must be a bare triple-backtick line. Neither permits leading indentation —
+// an indented "   ```" line is rendered as a paragraph, not a fence.
+export const FENCE_OPEN_RE = /^```(\w[\w+-]*)?\s*$/;
+export const FENCE_CLOSE_RE = /^```\s*$/;
 
 // Inline code span (single-backtick pair), mirrors runtime/markdown.tsx:164.
 export const INLINE_CODE_RE = /`[^`]+`/g;
@@ -41,15 +46,15 @@ export function computeSkipRanges(buffer: string): {
     const eol = buffer.indexOf('\n', pos);
     if (eol === -1) break;
     const line = buffer.slice(pos, eol);
-    if (FENCE_LINE_RE.test(line)) {
-      if (!inFence) {
+    if (!inFence) {
+      if (FENCE_OPEN_RE.test(line)) {
         inFence = true;
         fenceStart = pos;
-      } else {
-        inFence = false;
-        ranges.push([fenceStart, eol + 1]);
-        fenceStart = -1;
       }
+    } else if (FENCE_CLOSE_RE.test(line)) {
+      inFence = false;
+      ranges.push([fenceStart, eol + 1]);
+      fenceStart = -1;
     }
     pos = eol + 1;
   }
--- a/apps/web/src/artifacts/markdown-context.ts
+++ b/apps/web/src/artifacts/markdown-context.ts
@@ -20,28 +20,21 @@ export const FENCE_CLOSE_RE = /^```\s*$/;
 // Inline code span (single-backtick pair), mirrors runtime/markdown.tsx:164.
 export const INLINE_CODE_RE = /`[^`]+`/g;
 
-// Block-starter recognizers used by `parseBlocks()` in runtime/markdown.tsx:33-108.
-// `renderInline` only runs within a single block, so inline-code scanning must
-// reset at every block boundary — otherwise an unmatched backtick in one
-// paragraph can bridge across a blank line/heading/list/hr/fence into a later
-// paragraph and create a phantom inline span that swallows a real <artifact …>.
+// Paragraph-break recognizers — these mirror the inner paragraph-accumulation
+// loop in `parseBlocks()` (runtime/markdown.tsx:95-104), which is what
+// actually decides where a paragraph ends. The outer loop in `parseBlocks`
+// has additional block-starters (HR, fenced-code with `^```` prefix, etc.)
+// but those only take effect when no paragraph is currently being built;
+// mid-paragraph they are paragraph content. The renderer therefore treats
+// `intro \`` / `---` / `<artifact …>` / `---` / `closing \`` as ONE paragraph
+// whose backticks pair across the recitation — so this walker must too.
 //
-// Keep these in lock step with runtime/markdown.tsx:39 (blank), :59 (heading),
-// :67 (hr), :73 (ul), :83 (ol). Fence-open/close are already handled by the
-// fence pass above.
+// Notable omission: HR — see comment above. HR-shaped lines (`---` / `***`
+// / `___`) carry no backticks of their own, so leaving them inside the
+// surrounding paragraph region is benign for inline-code scanning either way.
 const HEADING_RE = /^#{1,4}\s+/;
-const HR_RE = /^\s*(?:-{3,}|_{3,}|\*{3,})\s*$/;
 const UL_ITEM_RE = /^\s*[-*+]\s+/;
 const OL_ITEM_RE = /^\s*\d+\.\s+/;
-
-function isBlockStarter(line: string): boolean {
-  if (line.trim() === '') return true;
-  if (HEADING_RE.test(line)) return true;
-  if (HR_RE.test(line)) return true;
-  if (UL_ITEM_RE.test(line)) return true;
-  if (OL_ITEM_RE.test(line)) return true;
-  return false;
-}
 
 // `<artifact` followed by whitespace is a real protocol open tag; any other
 // continuation (e.g. `<artifactual`) is a prefix-shared literal that must not
@@ -94,9 +87,8 @@ export function computeSkipRanges(buffer: string): {
         closeBlockBefore(pos);
         inFence = true;
         fenceStart = pos;
-      } else if (line.trim() === '' || HR_RE.test(line)) {
-        // Blank lines / horizontal rules separate blocks and have no inline
-        // content of their own.
+      } else if (line.trim() === '') {
+        // Blank lines separate blocks.
         closeBlockBefore(pos);
       } else if (HEADING_RE.test(line) || UL_ITEM_RE.test(line) || OL_ITEM_RE.test(line)) {
         // Heading and list-item lines are each their own block in the
--- a/apps/web/src/artifacts/markdown-context.ts
+++ b/apps/web/src/artifacts/markdown-context.ts
@@ -20,6 +20,38 @@ export const FENCE_CLOSE_RE = /^```\s*$/;
 // Inline code span (single-backtick pair), mirrors runtime/markdown.tsx:164.
 export const INLINE_CODE_RE = /`[^`]+`/g;
 
+// Block-starter recognizers used by `parseBlocks()` in runtime/markdown.tsx:33-108.
+// `renderInline` only runs within a single block, so inline-code scanning must
+// reset at every block boundary — otherwise an unmatched backtick in one
+// paragraph can bridge across a blank line/heading/list/hr/fence into a later
+// paragraph and create a phantom inline span that swallows a real <artifact …>.
+//
+// Keep these in lock step with runtime/markdown.tsx:39 (blank), :59 (heading),
+// :67 (hr), :73 (ul), :83 (ol). Fence-open/close are already handled by the
+// fence pass above.
+const HEADING_RE = /^#{1,4}\s+/;
+const HR_RE = /^\s*(?:-{3,}|_{3,}|\*{3,})\s*$/;
+const UL_ITEM_RE = /^\s*[-*+]\s+/;
+const OL_ITEM_RE = /^\s*\d+\.\s+/;
+
+function isBlockStarter(line: string): boolean {
+  if (line.trim() === '') return true;
+  if (HEADING_RE.test(line)) return true;
+  if (HR_RE.test(line)) return true;
+  if (UL_ITEM_RE.test(line)) return true;
+  if (OL_ITEM_RE.test(line)) return true;
+  return false;
+}
+
+// `<artifact` followed by whitespace is a real protocol open tag; any other
+// continuation (e.g. `<artifactual`) is a prefix-shared literal that must not
+// be treated as a tag. Mirrors the parser's `findOpenTag` real-open guard so
+// the parser and the stripper agree on what counts as a real tag.
+export function isRealArtifactOpenAt(content: string, idx: number): boolean {
+  const next = content.charAt(idx + '<artifact'.length);
+  return next !== '' && /\s/.test(next);
+}
+
 export type Range = readonly [number, number];
 
 /**
@@ -38,32 +70,69 @@ export function computeSkipRanges(buffer: string): {
   unclosedFenceStart: number | null;
 } {
   const ranges: Range[] = [];
+  // Paragraph-block regions are contiguous spans of paragraph lines outside
+  // any fenced code block. `renderInline` runs once per block, so inline-code
+  // scanning is restricted to one block at a time — backticks never pair
+  // across a block boundary in the rendered output.
+  const blockRegions: Range[] = [];
 
   let pos = 0;
   let inFence = false;
   let fenceStart = -1;
+  let blockStart = -1;
+  const closeBlockBefore = (idx: number) => {
+    if (blockStart !== -1 && idx > blockStart) blockRegions.push([blockStart, idx]);
+    blockStart = -1;
+  };
   while (pos < buffer.length) {
     const eol = buffer.indexOf('\n', pos);
-    if (eol === -1) break;
-    const line = buffer.slice(pos, eol);
+    const lineEnd = eol === -1 ? buffer.length : eol;
+    const line = buffer.slice(pos, lineEnd);
+    const lineHasNewline = eol !== -1;
     if (!inFence) {
-      if (FENCE_OPEN_RE.test(line)) {
+      if (lineHasNewline && FENCE_OPEN_RE.test(line)) {
+        closeBlockBefore(pos);
         inFence = true;
         fenceStart = pos;
+      } else if (line.trim() === '' || HR_RE.test(line)) {
+        // Blank lines / horizontal rules separate blocks and have no inline
+        // content of their own.
+        closeBlockBefore(pos);
+      } else if (HEADING_RE.test(line) || UL_ITEM_RE.test(line) || OL_ITEM_RE.test(line)) {
+        // Heading and list-item lines are each their own block in the
+        // renderer (`renderInline` runs per item / per heading), so they get
+        // a one-line inline-scan region rather than joining adjacent
+        // paragraphs. The marker chars themselves are not backticks, so we
+        // can scan the whole line without stripping the marker first.
+        closeBlockBefore(pos);
+        blockRegions.push([pos, lineEnd]);
+      } else {
+        if (blockStart === -1) blockStart = pos;
       }
-    } else if (FENCE_CLOSE_RE.test(line)) {
+    } else if (lineHasNewline && FENCE_CLOSE_RE.test(line)) {
       inFence = false;
       ranges.push([fenceStart, eol + 1]);
       fenceStart = -1;
     }
+    if (!lineHasNewline) {
+      // Trailing partial line — already accounted for above (either folded
+      // into the current paragraph region, or skipped because it starts with
+      // a block-starter pattern). Stop walking.
+      break;
+    }
     pos = eol + 1;
   }
+  // Flush any open paragraph region at end-of-buffer (no trailing newline).
+  if (!inFence) closeBlockBefore(buffer.length);
 
-  INLINE_CODE_RE.lastIndex = 0;
-  let m: RegExpExecArray | null = INLINE_CODE_RE.exec(buffer);
-  while (m !== null) {
-    ranges.push([m.index, m.index + m[0].length]);
-    m = INLINE_CODE_RE.exec(buffer);
+  for (const [s, e] of blockRegions) {
+    INLINE_CODE_RE.lastIndex = 0;
+    const segment = buffer.slice(s, e);
+    let m: RegExpExecArray | null = INLINE_CODE_RE.exec(segment);
+    while (m !== null) {
+      ranges.push([s + m.index, s + m.index + m[0].length]);
+      m = INLINE_CODE_RE.exec(segment);
+    }
   }
 
   return { ranges, unclosedFenceStart: inFence ? fenceStart : null };
--- a/apps/web/src/artifacts/markdown-context.ts
+++ b/apps/web/src/artifacts/markdown-context.ts
@@ -1,0 +1,72 @@
+/**
+ * Shared Markdown-context helpers used by both the streaming artifact parser
+ * and the post-stream `<artifact>` stripper. The single source of truth for
+ * what counts as a fenced code block or an inline code span — kept in lock
+ * step with apps/web/src/runtime/markdown.tsx so the parser/stripper view of
+ * a buffer matches what the chat UI will actually render.
+ *
+ * Anything that "looks like" an artifact tag inside one of these regions is
+ * literal Markdown and must not be treated as a real protocol tag.
+ */
+
+// Line-anchored fence delimiter, mirrors runtime/markdown.tsx:44.
+export const FENCE_LINE_RE = /^[ ]{0,3}```(\w[\w+-]*)?\s*$/;
+
+// Inline code span (single-backtick pair), mirrors runtime/markdown.tsx:164.
+export const INLINE_CODE_RE = /`[^`]+`/g;
+
+export type Range = readonly [number, number];
+
+/**
+ * Compute the half-open `[start, end)` ranges of `buffer` that the renderer
+ * would treat as fenced code blocks or inline code spans. The `unclosedFenceStart`
+ * is the index of an opening fence with no matching close in the buffer (or
+ * `null` if every fence is closed) — the streaming parser uses it to hold
+ * back; the stripper ignores it.
+ *
+ * Lines without a terminating `\n` (the trailing partial line during
+ * streaming) are not classified as fence delimiters here. Callers that care
+ * about partial-state behavior should inspect the tail themselves.
+ */
+export function computeSkipRanges(buffer: string): {
+  ranges: Range[];
+  unclosedFenceStart: number | null;
+} {
+  const ranges: Range[] = [];
+
+  let pos = 0;
+  let inFence = false;
+  let fenceStart = -1;
+  while (pos < buffer.length) {
+    const eol = buffer.indexOf('\n', pos);
+    if (eol === -1) break;
+    const line = buffer.slice(pos, eol);
+    if (FENCE_LINE_RE.test(line)) {
+      if (!inFence) {
+        inFence = true;
+        fenceStart = pos;
+      } else {
+        inFence = false;
+        ranges.push([fenceStart, eol + 1]);
+        fenceStart = -1;
+      }
+    }
+    pos = eol + 1;
+  }
+
+  INLINE_CODE_RE.lastIndex = 0;
+  let m: RegExpExecArray | null = INLINE_CODE_RE.exec(buffer);
+  while (m !== null) {
+    ranges.push([m.index, m.index + m[0].length]);
+    m = INLINE_CODE_RE.exec(buffer);
+  }
+
+  return { ranges, unclosedFenceStart: inFence ? fenceStart : null };
+}
+
+export function rangeContains(ranges: ReadonlyArray<Range>, p: number): boolean {
+  for (const [s, e] of ranges) {
+    if (p >= s && p < e) return true;
+  }
+  return false;
+}
--- a/apps/web/src/artifacts/parser.ts
+++ b/apps/web/src/artifacts/parser.ts
@@ -42,7 +42,7 @@ type OpenTagMatch =
   | { kind: 'partial'; start: number }
   | { kind: 'none' };
 
-import { computeSkipRanges, FENCE_OPEN_RE, rangeContains } from './markdown-context';
+import { computeSkipRanges, FENCE_OPEN_RE, isRealArtifactOpenAt, rangeContains } from './markdown-context';
 
 // Scan the buffer for `<artifact …>` while skipping any positions that the
 // chat markdown renderer would render as a fenced code block or inline code
@@ -59,34 +59,11 @@ function findOpenTag(buffer: string): OpenTagMatch {
   const len = buffer.length;
   const { ranges, unclosedFenceStart } = computeSkipRanges(buffer);
 
-  if (unclosedFenceStart !== null) {
-    return { kind: 'partial', start: unclosedFenceStart };
-  }
-
-  const lastNl = buffer.lastIndexOf('\n');
-  if (lastNl < len - 1) {
-    const tailLineStart = lastNl + 1;
-    const tail = buffer.slice(tailLineStart);
-    if (FENCE_OPEN_RE.test(tail) || /^`{1,2}$/.test(tail)) {
-      return { kind: 'partial', start: tailLineStart };
-    }
-  }
-
-  let firstUnmatched = -1;
-  let parity = 0;
-  for (let k = lastNl + 1; k < len; k++) {
-    if (buffer.charAt(k) !== '`') continue;
-    if (rangeContains(ranges, k)) continue;
-    if (parity === 0) {
-      firstUnmatched = k;
-      parity = 1;
-    } else {
-      firstUnmatched = -1;
-      parity = 0;
-    }
-  }
-  if (firstUnmatched !== -1) return { kind: 'partial', start: firstUnmatched };
-
+  // Pass 1: scan for the earliest *complete* real `<artifact …>` open outside
+  // any skip range. Done before any hold-back decision, otherwise a stray
+  // backtick or fence-opener prefix on a tail line would suppress an already
+  // self-contained artifact earlier in the buffer.
+  let earliestPartialOpen = -1;
   let from = 0;
   while (from < len) {
     const idx = buffer.indexOf(OPEN_PREFIX, from);
@@ -95,10 +72,21 @@ function findOpenTag(buffer: string): OpenTagMatch {
       from = idx + OPEN_PREFIX.length;
       continue;
     }
+    if (unclosedFenceStart !== null && idx >= unclosedFenceStart) {
+      // Anything past an unclosed fence opener is inside a code block that
+      // will close in a later chunk (or at end-of-buffer for the stripper);
+      // treat as skip range, not a real tag.
+      break;
+    }
     const after = idx + OPEN_PREFIX.length;
     const next = buffer.charAt(after);
-    if (next === '') return { kind: 'partial', start: idx };
-    if (!/\s/.test(next)) {
+    if (next === '') {
+      // `<artifact` at very end of buffer — could become real with the next
+      // chunk. Remember the earliest one and keep looking for a complete tag.
+      if (earliestPartialOpen === -1) earliestPartialOpen = idx;
+      break;
+    }
+    if (!isRealArtifactOpenAt(buffer, idx)) {
       // Not a real <artifact ...> open (e.g. "<artifactual"). Keep scanning.
       from = after;
       continue;
@@ -116,18 +104,57 @@ function findOpenTag(buffer: string): OpenTagMatch {
       }
       j++;
     }
-    return { kind: 'partial', start: idx };
+    // Ran out of buffer before the closing `>` arrived — this is an open tag
+    // mid-stream. Remember and stop scanning (any later `<artifact` would be
+    // a second tag we'd reach next chunk).
+    if (earliestPartialOpen === -1) earliestPartialOpen = idx;
+    break;
   }
 
-  // Strict prefix at the tail (e.g. "<art") — hold back.
-  const tail = buffer.lastIndexOf('<');
-  if (tail !== -1 && !rangeContains(ranges, tail)) {
-    const slice = buffer.slice(tail);
-    if (OPEN_PREFIX.startsWith(slice) && slice.length < OPEN_PREFIX.length) {
-      return { kind: 'partial', start: tail };
+  // Pass 2: no complete open found. Decide whether to hold back, and if so,
+  // from which position. Earliest hold-back wins so the text-flush boundary
+  // never crosses something that might still resolve into a tag/fence/span.
+  let holdback = -1;
+  const note = (pos: number | null) => {
+    if (pos !== null && pos !== -1 && (holdback === -1 || pos < holdback)) holdback = pos;
+  };
+  note(earliestPartialOpen);
+  note(unclosedFenceStart);
+
+  const lastNl = buffer.lastIndexOf('\n');
+  if (lastNl < len - 1) {
+    const tailLineStart = lastNl + 1;
+    const tail = buffer.slice(tailLineStart);
+    if (FENCE_OPEN_RE.test(tail) || /^`{1,2}$/.test(tail)) {
+      note(tailLineStart);
     }
   }
 
+  let firstUnmatched = -1;
+  let parity = 0;
+  for (let k = lastNl + 1; k < len; k++) {
+    if (buffer.charAt(k) !== '`') continue;
+    if (rangeContains(ranges, k)) continue;
+    if (parity === 0) {
+      firstUnmatched = k;
+      parity = 1;
+    } else {
+      firstUnmatched = -1;
+      parity = 0;
+    }
+  }
+  note(firstUnmatched);
+
+  // Strict prefix at the tail (e.g. "<art") — hold back.
+  const tailLt = buffer.lastIndexOf('<');
+  if (tailLt !== -1 && !rangeContains(ranges, tailLt)) {
+    const slice = buffer.slice(tailLt);
+    if (OPEN_PREFIX.startsWith(slice) && slice.length < OPEN_PREFIX.length) {
+      note(tailLt);
+    }
+  }
+
+  if (holdback !== -1) return { kind: 'partial', start: holdback };
   return { kind: 'none' };
 }
 
--- a/apps/web/src/artifacts/parser.ts
+++ b/apps/web/src/artifacts/parser.ts
@@ -42,88 +42,41 @@ type OpenTagMatch =
   | { kind: 'partial'; start: number }
   | { kind: 'none' };
 
-// Line-anchored fence regex, identical to apps/web/src/runtime/markdown.tsx so
-// the parser's view of "what's a fenced code block" matches what the chat UI
-// will actually render. Mid-line ```html in prose is NOT a fence per the
-// renderer, so it must not be one here either — otherwise a real artifact tag
-// that follows such prose would be wrongly suppressed.
-const FENCE_LINE_RE = /^[ ]{0,3}```(\w[\w+-]*)?\s*$/;
-
-// CommonMark-style inline code span: a single backtick, then one or more
-// non-backtick characters, then a single backtick. Matches the renderer's
-// renderInline regex so the two stay in lockstep.
-const INLINE_CODE_RE = /`[^`]+`/g;
-
-function rangeContains(ranges: Array<readonly [number, number]>, p: number): boolean {
-  for (const [s, e] of ranges) {
-    if (p >= s && p < e) return true;
-  }
-  return false;
-}
+import { computeSkipRanges, rangeContains } from './markdown-context';
 
 // Scan the buffer for `<artifact …>` while skipping any positions that the
 // chat markdown renderer would render as a fenced code block or inline code
-// span. Streaming caveat: when the buffer ends mid-fence (no closing fence
-// yet) or with an unmatched inline backtick, return a partial anchored at the
-// region's start so the caller waits for more data instead of emitting prose
-// that may turn out to be code.
+// span — see ./markdown-context.ts for the shared classification used by both
+// the streaming parser and the post-stream `<artifact>` stripper.
+//
+// Streaming caveats handled here on top of the shared ranges:
+//   * Open fence with no close yet → hold back from its opening line.
+//   * Unterminated tail line that could still resolve into a fence delimiter
+//     (e.g. "```", "```ht") → hold back from the line start.
+//   * Unmatched opening backtick after the last \n → hold back from it; a
+//     future chunk may turn it into an inline code span.
 function findOpenTag(buffer: string): OpenTagMatch {
   const len = buffer.length;
-  const skip: Array<readonly [number, number]> = [];
+  const { ranges, unclosedFenceStart } = computeSkipRanges(buffer);
 
-  // Pass 1: classify whole, \n-terminated lines as fence delimiters and
-  // collect [open-line-start, close-line-end+1) ranges.
-  let pos = 0;
-  let inFence = false;
-  let fenceStart = -1;
-  while (pos < len) {
-    const eol = buffer.indexOf('\n', pos);
-    if (eol === -1) {
-      // The tail is unterminated — we cannot classify it as a fence delimiter
-      // until \n arrives. If we're inside an open fence, hold back from the
-      // opening line so the next chunk can resolve the close.
-      if (inFence) return { kind: 'partial', start: fenceStart };
-      // Otherwise, if the tail looks like the start of a fence delimiter that
-      // hasn't completed yet (e.g. "```", "```ht"), hold back so a future \n
-      // can promote it to a real fence.
-      const tail = buffer.slice(pos);
-      if (/^[ ]{0,3}```\w*$/.test(tail) || /^[ ]{0,3}`{1,2}$/.test(tail)) {
-        return { kind: 'partial', start: pos };
-      }
-      break;
-    }
-    const line = buffer.slice(pos, eol);
-    if (FENCE_LINE_RE.test(line)) {
-      if (!inFence) {
-        inFence = true;
-        fenceStart = pos;
-      } else {
-        inFence = false;
-        skip.push([fenceStart, eol + 1]);
-        fenceStart = -1;
-      }
-    }
-    pos = eol + 1;
-  }
-  if (inFence) return { kind: 'partial', start: fenceStart };
-
-  // Pass 2: collect inline code spans with the same regex the renderer uses.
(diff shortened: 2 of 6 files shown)


apps/web/src/artifacts/parser.ts
```diff
@@ -42,47 +42,97 @@ type OpenTagMatch =
...
+  const len = buffer.length;
+  let i = 0;
+  let inFence = false;
+  let inInline = false;
+  let fenceStart = -1;
+
+  while (i < len) {
+    const c = buffer.charAt(i);
+
+    if (!inInline && c === '`' && buffer.charAt(i + 1) === '`' && buffer.charAt(i + 2) === '`') {
```
mrcfps: Problem: this branch treats any ` ``` ` sequence as a fence opener/closer, even when it appears mid-paragraph. `renderMarkdown()` only starts fenced code blocks for whole lines that match `/^```(\w[\w+-]*)?\s*$/`, so a reply like `The opening marker is ```html` followed by a real `<artifact ...>` on the next line will keep `findOpenTag()` inside `inFence` and skip the real artifact.

Why it matters: that turns this fix into a false negative — valid artifact outputs can be dropped whenever the model mentions triple backticks in prose before emitting the artifact body.

Evidence: the new detector toggles `inFence` immediately on any three backticks here without checking line boundaries or the rest of the fence line, and the added tests only cover canonical fence lines (` ```html ` / ` ``` `), so this renderer/parser mismatch is still untested.

Suggested change: gate fence open/close on the same line-based rule as `apps/web/src/runtime/markdown.tsx` (opener must consume the rest of its line except an optional language tag, and closer must be ` ``` ` plus optional trailing whitespace only), then add a fixture that keeps parsing a real artifact after a literal mid-line ``` sequence.

mrcfps (commented):
@mutoe Thanks for digging into the artifact-recitation bug — I walked the parser and the new regression tests, and I found one follow-up case where the new fence detection can still drift from the renderer.


apps/web/src/artifacts/parser.ts
mutoe: Verified the asymmetry — `runtime/markdown.tsx:44` uses `/^```(\w[\w+-]*)?\s*$/` for fences and `:164` uses `` `[^`]+` `` for inline code; the parser was more aggressive than the renderer in both directions.

Reworked `findOpenTag` in 96e88ca to mirror both rules:

- **Fences**: walk `\n`-terminated lines and only toggle state on lines matching the renderer's `FENCE_LINE_RE`. Mid-line triple backticks are ignored. An unclosed fence — or an unterminated tail line that could still resolve into a fence delimiter (`"```"`, `"```ht"`, etc.) — returns `partial` from the fence start so the next streaming chunk can resolve it.
- **Inline code**: collected with the same `/` `` ` `` `[^` `` ` `` `]+` `` ` `` `/g` regex `renderInline` uses. Unmatched trailing backticks after the last `\n` also hold back.
- **Artifact scan**: now a third pass that finds the first `<artifact …>` outside any skip range; partial-prefix tail handling preserved.

Added regression test `still parses a real artifact tag when prose contains an inline triple-backtick that is not a fence` — fails on 856f441 with `Received: undefined`, passes on 96e88ca alongside the original four cases.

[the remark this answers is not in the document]
mutoe: Confirmed — checked `apps/web/src/runtime/markdown.tsx:44` (`/^```(\w[\w+-]*)?\s*$/`) and `:164` (`` `[^`]+` ``); the parser was over-eager on both fence and inline detection.

Fixed in 96e88ca by reworking `findOpenTag` into three passes that mirror the renderer:

1. Walk `\n`-terminated lines and only toggle fence state on lines matching `FENCE_LINE_RE`. Mid-line triple backticks are now ignored. An unclosed fence — or an unterminated tail line that could still resolve into a fence delimiter — returns `partial` from the fence start so the next streaming chunk can resolve it.
2. Collect inline code spans with the same `` /`[^`]+`/g `` `renderInline` uses. Trailing unmatched backticks after the last `\n` also hold back.
3. Find the first `<artifact …>` outside any skip range; partial-prefix tail handling preserved.

Added regression test `still parses a real artifact tag when prose contains an inline triple-backtick that is not a fence` covering the example from this thread — fails on 856f441, passes on 96e88ca.

apps/web/src/artifacts/parser.ts
```diff
@@ -42,22 +42,107 @@ type OpenTagMatch =
...
-      const tail = buffer.lastIndexOf('<');
-      if (tail !== -1) {
-        const slice = buffer.slice(tail);
-        if (OPEN_PREFIX.startsWith(slice) && slice.length < OPEN_PREFIX.length) {
-          return { kind: 'partial', start: tail };
-        }
+  const len = buffer.length;
+  const skip: Array<readonly [number, number]> = [];
+
+  // Pass 1: classify whole, \n-terminated lines as fence delimiters and
```
lefarcen: **P2** (Agreeing with @mrcfps's finding) — Inline code spans are treated as toggled single backticks, but Markdown code spans can be delimited by longer backtick runs. For example, ``<artifact identifier="x" type="text/html" title="X">`` opens and immediately closes `inInline` on the first two backticks, then the `<artifact` is parsed as a real artifact. This is close to the reported class of bug and is not covered by the new tests.

Concrete fix: track the length of the opening backtick run for inline spans and only close on a matching run, then add a double-backtick regression test.

apps/web/src/artifacts/parser.ts
```diff
@@ -42,22 +42,107 @@ type OpenTagMatch =
...
+// non-backtick characters, then a single backtick. Matches the renderer's
+// renderInline regex so the two stay in lockstep.
+const INLINE_CODE_RE = /`[^`]+`/g;
+
+function rangeContains(ranges: Array<readonly [number, number]>, p: number): boolean {
+  for (const [s, e] of ranges) {
+    if (p >= s && p < e) return true;
+  }
+  return false;
+}
```
lefarcen: **P2** (Related to @mrcfps's line-based fence point) — Fence detection closes on any three consecutive backticks anywhere in the buffer, not on Markdown fence boundaries. A fenced example containing a JS/string literal like `const s = "```";` will close the parser's fence state early, so a later literal `<artifact ...>` still inside the actual Markdown code block can be misparsed as a real artifact. Four-backtick fences have similar run-length problems.

Concrete fix: recognize fence open/close only at valid Markdown fence positions (line-start run for open, line-start run of at least the opener's length for close), remember the opener run length, and close only on a line-start run at least that long; document any intentionally unsupported cases in the PR body if you choose a narrower parser.

lefarcen (commented):
Hey @mutoe! This fix addresses the core #1130 symptom (streaming artifact parser entering artifact mode on literal recitations), and the approach is solid. mrcfps caught one edge case below; I'm adding a few more dimensions from a full-lens review.

**What works well:**
- The linear-scan approach is cleaner than the old indexOf loop
- The partial-match holdback at buffer boundaries preserves streaming correctness
- The 4 new test cases cover the reported scenario

**Findings:**

**P1 (not in this PR's diff, but related fix gap)** — `apps/web/src/components/AssistantMessage.tsx:820` still calls `stripArtifact(text)`, and this helper does raw `content.indexOf("<artifact")` without Markdown awareness. After your parser fix passes literal `<artifact ...>` examples through as text events, this render-time stripper can still remove the rest of an inline-code explanation or fenced example content — so the main visible symptom from #1130 (truncated chat reply) can persist even if spurious artifact persistence is fixed. Consider either (a) reusing the same context-aware opener detection there, (b) removing this stripper for already-parser-filtered text events, or (c) adding rendering tests that cover inline/fenced literal artifact syntax through `AssistantMessage`. (This could be a follow-up PR if you prefer to keep scope tight.)

**P2 inline comments attached to parser.ts below.**

apps/web/src/artifacts/parser.ts
```diff
@@ -42,22 +42,107 @@ type OpenTagMatch =
   | { kind: 'partial'; start: number }
   | { kind: 'none' };
 
+// Line-anchored fence regex, identical to apps/web/src/runtime/markdown.tsx so
+// the parser's view of "what's a fenced code block" matches what the chat UI
+// will actually render. Mid-line ```html in prose is NOT a fence per the
+// renderer, so it must not be one here either — otherwise a real artifact tag
+// that follows such prose would be wrongly suppressed.
+const FENCE_LINE_RE = /^[ ]{0,3}```(\w[\w+-]*)?\s*$/;
```
mrcfps: Problem: `FENCE_LINE_RE` is used for both opening and closing fences, but `renderMarkdown()` does not use the same rule in both directions. With the current matcher, a line like `   ```html` opens a fence for the parser even though the renderer treats it as plain paragraph text, and a line like ````html` inside an existing fence closes the parser fence even though the renderer keeps it as code content.

Why it matters: that reintroduces the exact failure mode this PR is fixing on a narrower path. Once `findOpenTag()` drops out of `inFence` too early, a later literal `<artifact ...>` in the same rendered code example can still be parsed as a real artifact, which truncates the visible reply and can persist a spurious file from text the user only sees as code.

Evidence: line 50 accepts both leading indentation and an optional language suffix, and line 96 uses that same regex for both the open and close transitions. In `apps/web/src/runtime/markdown.tsx`, the opener is only `^```(\w[\w+-]*)?\s*$`, while the closer inside an open fence is only `^```\s*$`.

Suggested change: split this into renderer-matching opener/closer checks (or reuse the renderer block parser directly) so fence state only opens on the exact opener syntax and only closes on a bare closing fence, then add a regression where a fenced example contains a literal ````html` line before another `<artifact ...>` string.

mrcfps (commented):
@mutoe Thanks for tightening the parser/rendering alignment here — I walked the updated fence handling and the new regression cases, and I found one remaining fence-state mismatch worth fixing in this patch.


apps/web/src/artifacts/parser.ts
mutoe: Verified — this case is already handled by 96e88ca's renderer-aligned `INLINE_CODE_RE = /\`[^\`]+\`/g`. Tracing `` ``<artifact identifier="x" type="text/html" title="X">`` ``:

1. Regex starts at position 0 (first `` ` ``): `` ` `` then `[^` `` ` `` `]+` requires a non-backtick — position 1 is `` ` ``, fails.
2. From position 1 (second `` ` ``): `[^` `` ` `` `]+` matches `<artifact …>` greedily until the next `` ` `` (the first close backtick). Match `[1, L-1)` covers the `<artifact` inside.
3. `<artifact` lookup at position 2 falls inside the skip range → not treated as a real artifact. ✓

Pinned in 9dc5057 as the test `does not enter artifact mode for a tag wrapped in double backticks`. Triple-backtick mid-line wrappers (` ``` `…` ``` `) collapse to the same outcome via the same mechanism (the inline match anchors on the second `` ` `` of the run).

apps/web/src/artifacts/parser.ts
mutoe: Verified — this case is already handled by 96e88ca's line-anchored `FENCE_LINE_RE = /^[ ]{0,3}\`\`\`(\w[\w+-]*)?\s*$/`. A line like `const fence = "\`\`\`";` has non-backtick chars before the triple-backtick run, so it does not match the regex and does not toggle fence state. The outer fence (` ``` `js` … `` ``` ``) cleanly opens at one whole line and closes at another, producing a single skip range that covers any `<artifact ...>` inside.

Pinned in 9dc5057 as `does not enter artifact mode on a triple-backtick string literal inside a fenced block`.

On 4-backtick fences specifically: the renderer's regex (`runtime/markdown.tsx:44`) only recognizes 3-backtick fences, so the parser intentionally mirrors that — a 4-backtick fence is rendered literally, not as a code block, in the same UI the user sees. Supporting it in the parser would diverge the parser from the renderer (the gap that triggered this whole PR). If the renderer ever gains 4+ backtick support, the shared helpers in `markdown-context.ts` (added in a0bf05f) are the single place that needs updating.

mutoe:
@lefarcen The P1 finding on `stripArtifact` was real and not adequately covered by the parser fix alone. Folded into this PR rather than as a follow-up since the two share the same root cause (naive `indexOf("<artifact")` with no markdown awareness) and would otherwise leave the user-visible chat-truncation symptom from #1130 partially alive.

**a0bf05f** extracts the renderer-aligned classification into `apps/web/src/artifacts/markdown-context.ts` (single source of truth for fence + inline-code rules) and uses it from both:

- `apps/web/src/artifacts/parser.ts` — refactored to import `computeSkipRanges` / `rangeContains`, behavior preserved (all 7 existing parser tests still pass)
- `apps/web/src/artifacts/strip.ts` — new markdown-aware `stripArtifact` that skips `<artifact` inside fences/inline code, and (separately) refuses to strip when a real open has no matching real close, instead of slicing to end-of-string

`AssistantMessage.tsx` now imports from `strip.ts` and the local copy is gone, so there is no second truth to drift.

Verified empirically before the fix: `stripArtifact("A \`<artifact identifier=\"x\">demo</artifact>\` then more prose follows.")` produced `"A \`\` then more prose follows."` (literal recitation in inline backticks was being eaten). After the fix that input round-trips unchanged.

**Test coverage** (`apps/web/tests/artifacts/strip.test.ts`):

- removes a real artifact tag and its body
- preserves a tag wrapped in inline backticks
- preserves a tag inside a fenced code block
- preserves a tag wrapped in double backticks
- returns content unchanged when no open tag is present
- does not truncate when an open tag has no matching close (regression for the previous slice-to-end behavior)

P2 inline replies left in their threads with traces.

lefarcen (changes requested):
Hey @mutoe! The refactor direction (shared markdown-context helpers) is solid, and you've addressed the mid-line triple-backtick case. I walked the new `computeSkipRanges` + `stripArtifact` with the coding agent and found two gaps that leave the original #1130 symptom (chat truncation on literal artifact recitations) partially unfixed.

**P1** — `markdown-context.ts:42` does not process a closing fence when the final line has no trailing `\n`. A fenced literal ending at EOF can still expose `<artifact` inside to `stripArtifact`, e.g.:

```
```js
const s = `<artifact identifier="x">demo</artifact>`;
```
```

(no newline after the closing ` ``` `). `computeSkipRanges` will leave `unclosedFenceStart` set, and `stripArtifact` ignores it (line 21 only uses `skipRanges`, not the unclosed state), so the tag inside the backtick string is still stripped.

**P2** — `markdown-context.ts:13` still uses one `FENCE_LINE_RE` for both opening and closing fences. The renderer opens with `/^```(\w[\w+-]*)?\s*$/` and closes with `/^```\s*$/` (no language suffix allowed on close). With the current unified regex, a line like `` ```html `` inside an open fence can incorrectly close parser/stripper state and expose a later `<artifact>` tag that the renderer would keep inside code.

**Suggested fix:**
1. Split `FENCE_LINE_RE` into `FENCE_OPEN_RE = /^```(\w[\w+-]*)?\s*$/` and `FENCE_CLOSE_RE = /^```\s*$/` to match the renderer exactly.
2. Add a final pass in `computeSkipRanges` to handle the `unclosedFenceStart !== null` case (either complete the fence to EOF, or skip stripping if the buffer ends mid-fence).
3. Add regression tests covering (a) fenced literal with no trailing newline, and (b) a `` ```html `` line inside an open fence before a real artifact.

The core refactor is clean — once these two fence-boundary gaps are sealed, the P1 from the original review will be fully fixed.

File: apps/web/src/artifacts/markdown-context.ts [hunk 5]
```diff
@@ -0,0 +1,72 @@
...
+ * what counts as a fenced code block or an inline code span — kept in lock
+ * step with apps/web/src/runtime/markdown.tsx so the parser/stripper view of
+ * a buffer matches what the chat UI will actually render.
+ *
+ * Anything that "looks like" an artifact tag inside one of these regions is
+ * literal Markdown and must not be treated as a real protocol tag.
+ */
+
+// Line-anchored fence delimiter, mirrors runtime/markdown.tsx:44.
+export const FENCE_LINE_RE = /^[ ]{0,3}```(\w[\w+-]*)?\s*$/;
```
mrcfps: Problem: `FENCE_LINE_RE` still does not actually mirror `apps/web/src/runtime/markdown.tsx`. The renderer only opens on `^```(\w[\w+-]*)?\s*$` and only closes on `^```\s*$`, but this shared regex accepts up to three leading spaces and is reused for both the opening and closing transitions. That means inputs like `   ```html` are treated as code by the parser/stripper even though the UI renders them as plain paragraph text, and a literal ` ```html` line *inside* a fenced example will close the skip range early even though the renderer keeps it inside the code block.

Why it matters: both cases can still flip artifact handling away from what the user actually sees — either by suppressing a real artifact that follows an indented pseudo-fence, or by letting a later literal `<artifact ...>` inside the same rendered code example escape the skip range and get parsed/stripped as real content.

Evidence: this line defines one `FENCE_LINE_RE` for both directions, while `parseBlocks()` currently uses `const fence = /^```(\w[\w+-]*)?\s*$/` for open and `!/^```\s*$/.test(...)` for close. The new tests cover mid-line triple backticks and double-backtick spans, but there is still no case for either an indented pseudo-fence or a ` ```html` line inside an existing fenced block.

Suggested change: split the shared fence logic into renderer-matching opener/closer checks (no leading indentation, and closer must be a bare triple-backtick line), then add one regression for `   ```html` followed by a real artifact and one for a fenced example whose body contains a literal ` ```html` line before another `<artifact ...>` recitation.

mrcfps (commented):
@mutoe Thanks for tightening both the streaming parser and the render-time stripper here — I rechecked the final head against the chat markdown renderer and found one remaining fence-rule mismatch worth following up in this patch.


apps/web/src/artifacts/markdown-context.ts
mutoe: 66220e6 splits the shared regex into renderer-aligned siblings:

- `FENCE_OPEN_RE = /^```(\w[\w+-]*)?\s*$/`  — mirrors `runtime/markdown.tsx:44`
- `FENCE_CLOSE_RE = /^```\s*$/` — mirrors `runtime/markdown.tsx:49`

Both drop the `[ ]{0,3}` prefix so an indented `   ```html` line is no longer misclassified as a fence (the renderer treats it as a paragraph). `computeSkipRanges` now gates on the side-correct regex — open-only when not in fence, close-only when in fence — so a `\`\`\`html` line inside an open fenced example no longer closes the skip range early.

Also extended `stripArtifact` to honor `unclosedFenceStart` by appending `[unclosedFenceStart, content.length]` to the skip ranges (the renderer treats an unclosed-at-EOF fence as a code block to end-of-input, see `runtime/markdown.tsx:49`).

Regressions added in `apps/web/tests/artifacts/strip.test.ts`:
- fenced literal whose closing fence has no trailing `\n` — preserved
- indented `   ```html` before a real `<artifact …>` — stripped (renderer sees no fence)
- `\`\`\`html` line inside an open fence containing a literal `<artifact …>` — preserved

mutoe:
@lefarcen Both findings addressed in 66220e6.

**P1 — `unclosedFenceStart` ignored by `stripArtifact`**: `apps/web/src/artifacts/strip.ts` now extends the skip ranges with `[unclosedFenceStart, content.length]` when `computeSkipRanges` reports an unclosed fence. That mirrors `runtime/markdown.tsx:49`, where the close-loop runs until lines exhaust — an unclosed-at-EOF fence is rendered as a code block to end-of-input. Pinned by the new test "preserves a tag inside a fenced literal whose closing fence has no trailing newline" — verified RED before, GREEN after.

**P2 — single `FENCE_LINE_RE` for open and close**: split into

- `FENCE_OPEN_RE = /^```(\w[\w+-]*)?\s*$/`  (mirrors `runtime/markdown.tsx:44`)
- `FENCE_CLOSE_RE = /^```\s*$/`  (mirrors `runtime/markdown.tsx:49`)

Both without the previous `[ ]{0,3}` prefix — the renderer rejects any leading indentation on fence lines. `computeSkipRanges` gates state transitions on the side-correct regex. Parser tail-line hold-back regex tightened to match. Two new regressions in `strip.test.ts`:

- indented `   ```html` followed by a real `<artifact …>` — stripped (renderer treats as paragraph, the artifact is real protocol)
- `\`\`\`html` line inside an open fenced example containing a literal `<artifact …>` — preserved (renderer keeps body as code, close needs a bare triple-backtick)

`pnpm guard`, `pnpm typecheck`, and `pnpm --filter @open-design/web test` clean (4 pre-existing SettingsDialog localStorage failures unchanged from `main`). All 7 parser + 9 strip tests green.

mrcfps (dismissed):
@mutoe I rechecked the changed parser/stripper paths against `runtime/markdown.tsx`, and the shared markdown-context rules now keep literal `<artifact …>` recitations inside inline code and fenced blocks from being parsed or stripped while still preserving real artifact tags in prose. The added regression matrix covers the main fence/inline cases that previously drifted from the renderer. Thanks for sticking with the follow-up fixes here — this lands in a much safer place.


apps/web/src/artifacts/parser.ts
```diff
@@ -42,22 +42,59 @@ type OpenTagMatch =
...
+
+  if (unclosedFenceStart !== null) {
+    return { kind: 'partial', start: unclosedFenceStart };
+  }
+
+  const lastNl = buffer.lastIndexOf('\n');
+  if (lastNl < len - 1) {
+    const tailLineStart = lastNl + 1;
+    const tail = buffer.slice(tailLineStart);
+    if (/^```\w*$/.test(tail) || /^`{1,2}$/.test(tail)) {
```
lefarcen: **P2** — The streaming tail guard still does not mirror the renderer opener: `^```\w*$` misses valid opener tails such as ` ```c++`, ` ```ts-`, or ` ``` ` that `FENCE_OPEN_RE`/`runtime/markdown.tsx` accept. If a chunk ends there, the parser can emit part of the fence before the newline arrives, then parse a following literal `<artifact>` inside that code block as a real artifact.

Use the shared `FENCE_OPEN_RE` shape or a prefix-safe equivalent for this tail holdback.

lefarcen (dismissed):
Hey @mutoe, the P1 (unclosed fence EOF) and P2 (split FENCE_LINE_RE) fixes are solid — the shared markdown-context rules now align parser/stripper with the renderer for fences and inline code.

One non-blocking polish remains (P2 inline below): the streaming tail guard uses a stricter regex than the renderer opener, which can still let a chunk-boundary fence leak a later `<artifact>` inside that code block. This is a narrower edge than the original #1130 symptom and can be deferred or polished in a follow-up.

Approving to unblock — the main renderer/parser mismatch is resolved. Thanks for the thorough iteration! 🎉

apps/web/src/artifacts/parser.ts
mutoe: b186a25 swaps the local regex for the shared `FENCE_OPEN_RE` so the tail-line check uses the same shape as the rest of the pipeline.

Worth noting: the runtime impact appears nil. The unmatched-backtick parity scan immediately below the tail-line check (parser.ts:75-87) already holds back any tail with an odd run of leading backticks, which catches all three opener-tail shapes you cited (`\`\`\`c++`, `\`\`\`ts-`, `\`\`\` `) — three backticks is odd, `firstUnmatched` stays set, parser holds from that position. I added three pin tests in `parser.test.ts` ("holds back when a chunk ends mid-line on a renderer-valid fence opener prefix") that pass at the prior HEAD as well, proving the parity scan was already covering these inputs.

So this is alignment-only, not a bug fix — the value is that future readers see one regex shape across the file rather than two divergent ones. Trade-off is fine; flagging in case the framing matters for your sign-off.

apps/web/src/artifacts/markdown-context.ts
```diff
@@ -0,0 +1,77 @@
...
+      fenceStart = -1;
+    }
+    pos = eol + 1;
+  }
+
+  INLINE_CODE_RE.lastIndex = 0;
+  let m: RegExpExecArray | null = INLINE_CODE_RE.exec(buffer);
+  while (m !== null) {
+    ranges.push([m.index, m.index + m[0].length]);
+    m = INLINE_CODE_RE.exec(buffer);
```
mrcfps: Problem: this inline-code pass runs `INLINE_CODE_RE` over the entire buffer, so it can create one skip range that bridges across fenced blocks or other block boundaries that `renderMarkdown()` never treats as a single inline span. For example, a message shaped like `prefix ` + fenced block + trailing prose ending in another backtick makes the real `<artifact>` after the fence look like it is "inside inline code" even though the renderer would show it as normal prose.

Why it matters: both `parser.ts` and `strip.ts` trust `computeSkipRanges()`, so this false skip range can suppress a legitimate artifact or keep it from being stripped after a fenced example — the same user-visible failure mode this PR is trying to eliminate, just from the opposite direction.

Evidence: lines 62-66 apply the regex globally without excluding fence ranges or resetting at block boundaries. A concrete repro is:

```text
prefix `
```txt
inside fence
```
<artifact identifier="x" type="text/plain" title="X">demo</artifact>
trailer `
```

With the current logic, the `<artifact>` index falls inside one of the computed skip ranges, so it is never treated as real.

Suggested change: only run `INLINE_CODE_RE` on text regions outside fenced blocks (or per renderer paragraph/list/heading block) so unmatched backticks in one block cannot bridge over a fence into later prose. Please also add a regression that covers stray backticks around a fenced block, and another where a fence body contains an unmatched template-literal backtick before a real artifact that follows.

<!-- looper:stamp v=1 --><sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=reviewer · agent=opencode · An autonomous AI dev team for your GitHub repos.</sub>

**Review by mrcfps — COMMENTED** [review]
@mutoe Thanks for continuing to tighten the parser/stripper alignment here — the shared markdown-context refactor fixes the earlier fence cases, but I found one remaining cross-block inline-range mismatch worth following up.


**Review by lefarcen — COMMENTED** [review]
Hey @mutoe, the tail-guard alignment fix is good — parser.ts now uses the shared `FENCE_OPEN_RE` for streaming holdback, matching the rest of the pipeline. The P2 you flagged in 3214444381 (nil runtime impact because the parity scan was already covering those inputs) is now resolved.

One remaining non-blocking P2 inline below: the markdown holdback checks can return before scanning for an earlier real `<artifact` in the same buffered delta, which means a chunk like `<artifact ...>...</artifact>\n```ts-` can cause the real artifact to be missed. This is a polish item for the streaming edge-case path.

Overall the PR is solid — the main #1130 symptom (chat truncation on artifact recitation) is fixed. Thanks for the thorough iteration! 🎉

File: apps/web/src/artifacts/parser.ts [hunk 8]
```diff
@@ -42,22 +42,59 @@ type OpenTagMatch =
...
+//   * Open fence with no close yet → hold back from its opening line.
+//   * Unterminated tail line that could still resolve into a fence delimiter
+//     (e.g. "```", "```ht") → hold back from the line start.
+//   * Unmatched opening backtick after the last \n → hold back from it; a
+//     future chunk may turn it into an inline code span.
 function findOpenTag(buffer: string): OpenTagMatch {
+  const len = buffer.length;
+  const { ranges, unclosedFenceStart } = computeSkipRanges(buffer);
+
+  if (unclosedFenceStart !== null) {
```
lefarcen: **P2** — The markdown holdback checks return before scanning for an earlier real `<artifact` in the same buffered delta. Because `ProjectView` batches content before feeding the parser, a chunk like `<artifact ...>...</artifact>\n```ts-` (real artifact followed by a fence tail) is emitted as text up to the holdback, and the artifact is never parsed.

Scan for the earliest unskipped open before applying holdback, or only hold back when the holdback position precedes the first real open. This is a polish item for the streaming edge-case path, not a blocker.

apps/web/src/artifacts/markdown-context.ts
mutoe: Acknowledged — the bridging mechanism is real (whole-buffer regex vs renderer's per-paragraph `renderInline`). Worth noting your concrete repro doesn't actually trigger it: the triple-backticks of \`\`\`txt and \`\`\` fragment the inline regex match, so the `<artifact …>` falls between two short inline spans rather than inside one bridged span.

A repro that does trigger needs two stray single backticks in different paragraphs with no fence between them, e.g.:

\`\`\`text
Prose with \` backtick

<artifact id=\"x\" type=\"text/html\" title=\"X\">demo</artifact>

Another paragraph with \` backtick.
\`\`\`

Deferring on YAGNI grounds — this composite (two unbalanced backticks across paragraphs flanking a real artifact) hasn't been observed in actual LLM output, and the proper fix (paragraph-aware partitioning) is a bigger refactor than the rest of this PR. Tracked at #1141 with a fix sketch and revisit triggers, so the analysis isn't lost if a real incident surfaces.

apps/web/src/artifacts/parser.ts
mutoe: Acknowledged — `findOpenTag` does return holdback before scanning for a real `<artifact` open, so the composite you describe (`<artifact …>…</artifact>\n\`\`\`ts-` in one chunk) would emit the earlier real artifact as text.

Deferring on YAGNI grounds — this requires a real model to emit a complete artifact then immediately switch to a fence-opener prefix in the same delta, which we haven't observed. Tracked at #1141 alongside the inline-bridging case, with a fix sketch (find earliest unskipped `<artifact`; compare against earliest holdback; process artifact first when it precedes holdback) and the triggers that should make us revisit.

mutoe:
@lefarcen @mrcfps Status check before merge:

- Both new findings (inline-range bridging across paragraphs, holdback returning before earliest-artifact scan) are technically real but neither shape has been observed in actual LLM output. Deferred on YAGNI grounds; full analysis, repro, fix sketch, and revisit triggers tracked at #1141 so nothing is lost.
- All earlier P1/P2 findings from this review chain are addressed in-tree (96e88ca → 9dc5057 → a0bf05f → 66220e6 → b186a25). 6/6 review threads resolved.
- Validation: `pnpm guard`, `pnpm typecheck`, and `pnpm --filter @open-design/web test` clean (4 pre-existing SettingsDialog localStorage failures unchanged from `main`). Artifacts test surface 46/46 green, with new pins for every finding I addressed in the PR.

Requesting re-review / approval if the deferral framing works for you. Happy to fold #1141 into this PR instead if either of you would rather see it shipped together.

lefarcen:
@mrcfps This PR's current head b186a253 is waiting on your review.

apps/web/src/artifacts/markdown-context.ts
```diff
@@ -0,0 +1,77 @@
...
+      }
+    } else if (FENCE_CLOSE_RE.test(line)) {
+      inFence = false;
+      ranges.push([fenceStart, eol + 1]);
+      fenceStart = -1;
+    }
+    pos = eol + 1;
+  }
+
+  INLINE_CODE_RE.lastIndex = 0;
```
lefarcen: `INLINE_CODE_RE` runs over the entire buffer, but `renderMarkdown` applies inline parsing separately per parsed block. This can pair backticks across blank lines/headings/lists/fences and incorrectly classify a real `<artifact>` in between as inline code. Example: `intro \`\n\n<artifact ...>...</artifact>\n\nclosing \`` would render the artifact outside inline code, but parser/strip would skip it. Inline skip ranges need to reset on the same block boundaries as `parseBlocks`.

apps/web/src/artifacts/strip.ts
```diff
@@ -0,0 +1,47 @@
...
+  const { ranges: baseRanges, unclosedFenceStart } = computeSkipRanges(content);
+  // For complete (non-streaming) content, an unclosed fence is rendered by
+  // the chat Markdown renderer as a code block extending to end of input
+  // (see runtime/markdown.tsx:49 — the close-loop runs until lines exhaust).
+  // The stripper has to mirror that, otherwise a literal `<artifact …>`
+  // tucked into a code example at the bottom of a chat reply (no trailing
+  // newline) gets treated as a real protocol tag and eaten.
+  const ranges: Range[] =
+    unclosedFenceStart !== null ? [...baseRanges, [unclosedFenceStart, content.length]] : baseRanges;
+  const open = findUnskipped(content, OPEN, 0, ranges);
```
lefarcen: The stripper treats any unskipped `<artifact` prefix as a real artifact open, unlike the parser which requires whitespace after `<artifact`. Inputs like `<artifactual>...</artifact>` or `<artifacted>...</artifact>` are not protocol tags but will be stripped from the assistant message.

lefarcen (changes requested):
@mutoe Thanks for the shared-helper refactor — it tightens the renderer alignment significantly. The coding agent found two new issues in the shared module that need fixing before merge:

1. **Inline-code skip range bridges across block boundaries** (markdown-context.ts:62) — the global INLINE_CODE_RE can pair backticks across blank lines/headings/fences, incorrectly skipping real artifacts in between. The renderer applies inline parsing per block, so the skip ranges must respect the same boundaries.

2. **Stripper accepts non-artifact prefixes** (strip.ts:40) — inputs like `<artifactual>...</artifact>` are not protocol tags but will be stripped because the stripper doesn't require whitespace after `<artifact` like the parser does.

See inline comments for details and suggested fixes.

apps/web/src/artifacts/markdown-context.ts
mrcfps: Problem: this inline-code pass runs `INLINE_CODE_RE` over the whole message buffer, so it can pair backticks across block boundaries that `renderMarkdown()` never treats as one inline span. For example, `intro `\n\n<artifact identifier="x" type="text/plain" title="X">demo</artifact>\n\nclosing `` produces one skip range that covers the real artifact, even though `parseBlocks()` would render those as separate paragraphs and leave the artifact outside inline code.

Why it matters: both `parser.ts` and `strip.ts` trust `computeSkipRanges()`, so this can suppress a real artifact or keep it from being stripped, recreating the same parser/render mismatch this PR is trying to eliminate.

Evidence: `computeSkipRanges()` applies the regex globally on `buffer`, while `apps/web/src/runtime/markdown.tsx` splits on blank lines/list items/headings/fences and only calls `renderInline()` per parsed block. A quick reproduction against this helper yields a single range `[6, 88)` for the example above, and the artifact index falls inside it.

Suggested change: compute inline-code ranges per renderer block boundary (or at least only within non-fenced paragraph segments) so unmatched backticks in one block cannot bridge into later prose. Please also add a regression with stray backticks in separate paragraphs around a real artifact.

apps/web/src/artifacts/strip.ts
mrcfps: Problem: `stripArtifact()` accepts any unskipped `<artifact` prefix as a real opener, but the streaming parser only treats it as real when the next character is whitespace. Inputs like `prefix <artifactual>demo</artifact> suffix` or `<artifacted>...` are not protocol tags, yet this code strips everything through `</artifact>`.

Why it matters: that still truncates normal assistant text on literal strings that merely share the prefix, and it leaves the parser/stripper disagreeing about what counts as a real artifact tag.

Evidence: `findOpenTag()` in `parser.ts` does `if (!/\s/.test(next)) { ... continue; }`, but `stripArtifact()` calls `findUnskipped(content, OPEN, 0, ranges)` and immediately strips from that index with no equivalent validation. Reproducing the current logic on `prefix <artifactual>demo</artifact> suffix` returns `prefix  suffix`.

Suggested change: after `open`, validate the following character with the same real-open rule the parser uses (or reuse the parser/open-tag helper directly), and add regressions for `<artifactual>` / `<artifacted>` so the two code paths stay aligned.

mrcfps (commented):
@mutoe Thanks for tightening the parser/stripper alignment here — I rechecked the latest head against `runtime/markdown.tsx` and found two remaining renderer-mismatch edges worth following up.


apps/web/src/artifacts/markdown-context.ts
mutoe: Acknowledged and fixed in fd4d9c0 — `INLINE_CODE_RE` is no longer whole-buffer.

`computeSkipRanges` now walks lines and partitions the non-fenced text into per-block "inline scan regions" before running the regex, mirroring how `parseBlocks()` in `apps/web/src/runtime/markdown.tsx:33` only calls `renderInline` once per parsed block:

- blank lines and HR lines close the current region (no inline content of their own)
- heading / ul-item / ol-item lines each become a one-line region (the renderer scans each independently)
- fence-open lines close any open paragraph region
- consecutive paragraph lines accumulate into one region

`INLINE_CODE_RE` is then run inside each region with its own `lastIndex`, so an unmatched backtick in one block can no longer pair across a boundary.

Your exact repro is pinned in `strip.test.ts` ("strips a real artifact between paragraphs that each carry a stray backtick") plus a parser-side twin in `parser.test.ts`; both fail RED at `3db21ff` (`prefix [6,88)` bridging exactly as you described) and pass at this head.

I also folded in the parser's symmetric blind-spot while I was in there: `findOpenTag` previously decided hold-back before scanning for a complete real `<artifact …>`, so the same backtick pair would suppress an artifact that was already self-contained earlier in the buffer. `findOpenTag` is now two passes — earliest complete real open first, only then merge `unclosedFenceStart` / tail fence-prefix / tail backtick parity / partial `<art` and pick the earliest. That closes the second half of #1141.

Validation: 62/62 in `tests/artifacts/`, web `typecheck` and `guard` clean, 746/750 in `pnpm --filter @open-design/web test` (the 4 failures are the same pre-existing SettingsDialog localStorage cases that fail on `main`).

apps/web/src/artifacts/strip.ts
mutoe: Confirmed and fixed in fd4d9c0. The two paths disagreed because `findOpenTag` in `parser.ts:101` already had the `<artifact` + `\s` real-open guard, but `stripArtifact` reached the same prefix through a separate `findUnskipped` and had no equivalent check. Your exact `prefix <artifactual>demo</artifact> suffix` → `prefix  suffix` reproduced before the fix.

Rather than duplicate the rule in `strip.ts`, I lifted it into a single helper in the shared module — `isRealArtifactOpenAt(content, idx)` in `apps/web/src/artifacts/markdown-context.ts` — and have both sides import it:

- `parser.ts:findOpenTag` now calls `isRealArtifactOpenAt` instead of the inline `/\s/.test(next)` check.
- `strip.ts:findRealOpen` uses it as the unskipped-and-real predicate.

That keeps the "single source of truth" shape this PR is built around — `markdown-context.ts` is now the only place that defines what counts as a real protocol open tag, just like it's the only place that defines fence/inline-code skip ranges.

Pinned in `strip.test.ts` ("does not strip `<artifactual>` or other prefix-shared identifiers") and a parser-side twin in `parser.test.ts`. Both fail RED at `3db21ff` and pass at this head.

apps/web/src/artifacts/markdown-context.ts
mutoe: Folded in — your `intro \`\n\n<artifact …>\n\nclosing \`` shape is the cleanest repro of the bridging mechanism (mrcfps gave the same shape independently a few minutes later, so both threads collapse to the same fix).

fd4d9c0 partitions the non-fenced buffer into per-block "inline scan regions" inside `computeSkipRanges`, mirroring `parseBlocks()` in `apps/web/src/runtime/markdown.tsx:33` so `INLINE_CODE_RE` is run per renderer block rather than once over the whole buffer:

- blank / HR lines close the current region
- heading and list-item lines become one-line regions (each gets its own `renderInline` call in the renderer)
- fence open lines close any open paragraph region

That breaks the cross-block pair you described and the analogous heading/list-bridging variants in one shot.

The streaming parser had a symmetric blind-spot — `findOpenTag` returned a hold-back from the tail backtick parity scan before ever attempting to find a complete real `<artifact …>`, so the same kind of buffer would emit nothing for an already-self-contained artifact. That's the second deferred finding from #1141 — also fixed in this commit by reordering `findOpenTag` to scan for the earliest complete real open first, then merge all hold-back signals (`unclosedFenceStart` / tail fence-prefix / backtick parity / partial `<art`) and pick the earliest position only when no complete tag is found.

Pinned on both sides (`parser.test.ts` + `strip.test.ts`). #1141 is now empty — happy to close it once this is in.

apps/web/src/artifacts/strip.ts
mutoe: Right — `parser.ts:101` had the `<artifact` + `\s` guard, `strip.ts` didn't, so `<artifactual>...</artifact>` was being eaten. Verified `prefix <artifactual>demo</artifact> suffix` → `prefix  suffix` on `3db21ff` and unchanged after the fix.

Lifted the predicate into `isRealArtifactOpenAt(content, idx)` in `apps/web/src/artifacts/markdown-context.ts` so both paths share one definition:

- `parser.ts:findOpenTag` calls it instead of the inline `/\s/.test(next)`.
- `strip.ts` gains `findRealOpen` (unskipped + real-open) which uses it as the predicate.

`markdown-context.ts` was already the single place that owns fence / inline-code classification; this keeps "what counts as a real `<artifact` open" in the same module so the two code paths can't drift again. Pinned with regressions on both sides.

mutoe:
@mrcfps @lefarcen Both new findings addressed in fd4d9c0. Inline reply threads have the per-finding traces; quick summary:

- **inline-code bridging across blocks** — `computeSkipRanges` now partitions the non-fenced buffer into per-block "inline scan regions" mirroring `runtime/markdown.tsx:parseBlocks` (blank/HR close, heading/list-item are one-line regions, fence opens close any paragraph). Your `intro \`` + paragraph + `closing \`` repro is pinned in `strip.test.ts` and `parser.test.ts`.
- **`<artifactual>` parser/stripper drift** — lifted the `<artifact` + `\s` real-open guard into `isRealArtifactOpenAt` in `markdown-context.ts`; both `findOpenTag` and the new `strip.ts:findRealOpen` import it. Pinned on both sides.

While I was in `findOpenTag` for the shared guard, also fixed the second half of #1141 — the parser used to decide hold-back (tail backtick parity, fence prefix, unclosedFence) before scanning for a complete real `<artifact …>`, so a stray tail backtick could suppress an already-self-contained artifact earlier in the buffer. `findOpenTag` is now two passes: earliest complete real open first, then merge all hold-back signals and pick the earliest only when no complete tag was found. Both #1141 cases are now empty — I'll close it after this lands.

Validation:
- `pnpm --filter @open-design/web exec vitest run tests/artifacts/` — 62/62
- `pnpm --filter @open-design/web typecheck` — clean
- `pnpm --filter @open-design/web test` — 746/750 (same 4 pre-existing SettingsDialog localStorage failures as on `main`)
- `pnpm guard` — clean

Re-review please.

apps/web/src/artifacts/markdown-context.ts
```diff
@@ -0,0 +1,146 @@
...
+    if (!inFence) {
+      if (lineHasNewline && FENCE_OPEN_RE.test(line)) {
+        closeBlockBefore(pos);
+        inFence = true;
+        fenceStart = pos;
+      } else if (line.trim() === '' || HR_RE.test(line)) {
+        // Blank lines / horizontal rules separate blocks and have no inline
+        // content of their own.
+        closeBlockBefore(pos);
+      } else if (HEADING_RE.test(line) || UL_ITEM_RE.test(line) || OL_ITEM_RE.test(line)) {
```
mrcfps: Problem: `computeSkipRanges()` closes the current inline-scan region on `HR_RE`, but `parseBlocks()` does not treat `---` / `***` / `___` as a paragraph terminator once a paragraph has already started. A buffer like `intro \`` → `---` → `<artifact …>` → `---` → `closing \`` is still one paragraph in `runtime/markdown.tsx`, so `renderInline()` pairs those backticks and renders the artifact recitation as code. Splitting on the HR lines here leaves the `<artifact>` outside `skipRanges`, so `parser.ts` / `strip.ts` will treat it as a real tag and can truncate the reply again.

Why it matters: this reintroduces the same parser/render drift the PR is fixing for literal protocol explanations that happen to use Markdown horizontal-rule lines between the opening and closing backticks.

Evidence: `runtime/markdown.tsx:95-104` only breaks paragraph accumulation on blank lines, fences, headings, and list starters; it never breaks on HR lines. Repro: `intro \`` + blankless `---` + `<artifact identifier=\"x\" type=\"text/plain\" title=\"X\">demo</artifact>` + `---` + `closing \`` produces one inline-code match in the renderer, but this branch splits it into three block regions.

Suggested change: keep `HR_RE` lines inside the current paragraph region unless they start a block from column 0 before any paragraph text, or derive these inline-scan regions from the same block-walk the renderer uses. Please also add parser/strip regressions for the `intro \`` / `---` / `<artifact …>` / `---` / `closing \`` case.

<sub>🔁 Powered by <a href=\"https://github.com/nexu-io/looper\">Looper</a> · runner=reviewer · agent=opencode · An autonomous AI dev team for your GitHub repos.</sub>

mrcfps (commented):
@mutoe Thanks for pushing this parser/stripper cleanup through — I rechecked the shared markdown-context rules against `runtime/markdown.tsx` and found one remaining renderer mismatch worth following up.


lefarcen (dismissed):
Both prior P1 findings resolved in fd4d9c0.

- **Cross-block backtick pairing**: `computeSkipRanges` now partitions non-fenced text into per-block inline scan regions mirroring the renderer's block boundaries (blank/HR close, heading/list-item are one-line regions). Your `intro \`\n\n<artifact …>\n\nclosing \`` repro is covered.
- **`<artifactual>` drift**: Lifted the `<artifact` + `\s` real-open guard into `isRealArtifactOpenAt` in `markdown-context.ts`; both `parser.ts:findOpenTag` and `strip.ts:findRealOpen` import it.

Regression tests pinned on both sides. The parser/stripper/renderer alignment is now complete. Thanks for the persistence on the follow-up fixes — this lands in a much safer place.

apps/web/src/artifacts/markdown-context.ts
mutoe: Confirmed and fixed in e91bb13 — you're right, I had the renderer's outer-loop block-starter list (which does include HR) confused with the inner paragraph-accumulation loop (which does not).

`runtime/markdown.tsx:95-104` only breaks paragraph accumulation on blank / fence / heading / ul / ol. HR is only treated as its own block when the outer loop reaches it without an active paragraph; mid-paragraph it's just paragraph content. So your `intro \`\n---\n<artifact …>\n---\nclosing \`` was indeed one rendered paragraph whose two stray backticks paired across the recitation, and my walker was wrongly splitting on the `---` lines.

`computeSkipRanges` now drops HR from the paragraph-break list. HR-shaped lines (`---` / `***` / `___`) carry no backticks of their own, so folding them into the surrounding paragraph region is benign for inline-code scanning either way.

Pinned on both sides:
- `parser.test.ts` → "does not enter artifact mode when bridged by stray backticks across HR-shaped lines"
- `strip.test.ts` → "preserves a tag bridged by stray backticks across HR-shaped lines (renderer keeps HR as paragraph content)"

Both fail RED at `fd4d9c0` and pass at this head. 64/64 in `tests/artifacts/`, typecheck and guard clean.

mrcfps (approved):
@mutoe I rechecked the final parser/stripper changes against `runtime/markdown.tsx`, and the shared markdown-context helpers now keep literal `<artifact …>` recitations inside inline code and fenced blocks from being parsed or stripped while still preserving real artifact tags in prose. The expanded parser/strip regression matrix covers the earlier fence, paragraph-boundary, prefix-shared-open, and HR-shaped-line cases that previously drifted from the renderer. Thanks for staying with the follow-up fixes here — this lands in a much safer place. 🙏

PR #1133, opened 2026-05-10
Stabilize clustered agent run persistence

chenpx976:
## Summary
- serialize per-run persistence writes before queued worker handoff
- serialize worker run events so Postgres event ordering stays stable
- update store-hook test to await persistence flush

## Validation
- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/agent-run-store.test.ts tests/agent-runtime.test.ts tests/runs-store.test.ts tests/project-fs.test.ts
- pnpm guard
- OrbStack docker compose Postgres/Redis E2E smoke generated index.html successfully


apps/daemon/src/agent-queue-redis.ts
```diff
@@ -0,0 +1,203 @@
...
+    running: `${prefix}:agent-jobs:running`,
+    job: (id) => `${prefix}:agent-jobs:job:${id}`,
+  };
+}
+
+function parseJob(id, record) {
+  if (!record || Object.keys(record).length === 0) return null;
+  return {
+    id,
+    runId: record.runId,
```
lefarcen, line 42: **P1 — JSON.parse vulnerability**

If `record.payloadJson` contains malformed JSON (corrupted Redis data / race condition / manual edit), this throws and crashes the worker.

Fix:
```ts
try {
  return JSON.parse(record.payloadJson || 'null');
} catch (err) {
  console.error('Failed to parse job payload:', err);
  return null; // or throw a JobCorruptedError
}
```

Without this, a single bad job can kill the entire worker loop.

apps/daemon/src/agent-queue-redis.ts
```diff
@@ -0,0 +1,203 @@
...
+        .exec();
+      return { id, runId };
+    },
+    async claimNextJob(workerId, options = {}) {
+      const maxAttempts = Math.max(1, Number(options.maxAttempts) || 3);
+      await requeueStaleJobs({ ...options, maxAttempts });
+      while (true) {
+        const id = await redis.lpop(k.queued);
+        if (!id) return null;
+        const jobKey = k.job(id);
```
lefarcen, line 120: **P1 — Race condition in claimNextJob**

Between `lpop(k.queued)` and the subsequent `hgetall(jobKey)`, another worker could have already claimed this job (if `requeueStaleJobs` requeued it concurrently). The current code skips via `continue`, but this creates a tight loop if Redis is flaky.

Mitigation:
- Add a `WATCH` / `MULTI` / `EXEC` block around the claim
- OR add exponential backoff when `continue` happens >N times
- OR use Redis Lua script for atomic lpop+hgetall+zadd

Without this, under load you'll see workers spinning on empty claims.

apps/daemon/src/agent-run-postgres.ts
```diff
@@ -0,0 +1,193 @@
...
+         VALUES ($1,$2,$3,$4::jsonb,$5)
+         ON CONFLICT (run_id, event_id) DO UPDATE SET
+           event = EXCLUDED.event,
+           data_json = EXCLUDED.data_json,
+           timestamp = EXCLUDED.timestamp`,
+        [run.id, record.id, record.event, JSON.stringify(record.data ?? null), record.timestamp],
+      );
+    },
+    async appendRunEvent(runId, event, data) {
+      const client = await pgPool.connect();
```
lefarcen, line 139: **P1 — Connection leak on error**

If `client.query('INSERT ...')` throws (e.g., constraint violation, network timeout), the `catch` block rolls back but the `finally` still releases the client — but Postgres may not have closed the transaction cleanly. This can leak connections if errors are frequent.

Fix: wrap the ROLLBACK itself in a try-catch:
```ts
catch (err) {
  try {
    await client.query('ROLLBACK');
  } catch (rollbackErr) {
    console.error('ROLLBACK failed:', rollbackErr);
  }
  throw err;
}
```

Also consider: does `pg` pool handle connection state reset automatically? If not, you may need to call `client.clean()` or equivalent.

apps/daemon/src/agent-runtime.ts
```diff
@@ -0,0 +1,200 @@
+import { runPiSdkAgent } from './pi-sdk-agent.js';
```
lefarcen, line 1: **P2 — @ts-nocheck on critical runtime code**

This file orchestrates the entire agent execution flow, yet `@ts-nocheck` disables all type safety. This is a high-risk path where type errors can cause data corruption (wrong run state, missing events).

Recommendation:
- Remove `@ts-nocheck`
- Fix type errors properly
- If there are legitimate "can't fix right now" type issues, use targeted `@ts-expect-error` with comments explaining why

Same applies to agent-queue-redis.ts, agent-run-postgres.ts, agent-worker.ts, pi-sdk-agent.ts.

apps/daemon/src/agent-run-backends.ts
```diff
@@ -0,0 +1,26 @@
...
+export async function createAgentRunPersistenceFromEnv(env, { db }) {
+  const kind = String(env.OD_AGENT_RUN_STORE || 'sqlite').trim().toLowerCase();
+  if (kind === 'sqlite' || kind === '') return createSqliteAgentRunPersistence(db);
+  if (kind === 'postgres' || kind === 'postgresql') {
+    return createPostgresAgentRunPersistence({ env });
+  }
+  throw new Error(`Unsupported OD_AGENT_RUN_STORE: ${kind}`);
+}
+
+export async function createAgentJobQueueFromEnv(env, { db }) {
```
lefarcen, line 15: **P2 — Misleading error message**

If `OD_AGENT_JOB_QUEUE` is set to an unsupported value (e.g., `rabbitmq`), the error says `Unsupported OD_AGENT_JOB_QUEUE: rabbitmq`, but it doesn't tell the user what the valid options are.

Fix:
```ts
throw new Error(
  `Unsupported OD_AGENT_JOB_QUEUE: ${kind}. Valid options: sqlite, redis`
);
```

Same for `createAgentRunPersistenceFromEnv` line 11.

lefarcen (changes requested):
Hi @chenpx976, thanks for this major clustering refactor! 👋 This is a significant architectural improvement, but I've found several critical issues that need attention before merge.

## Code correctness (Lens A)

**Summary**: 3 P1 issues (Redis race condition, Postgres connection leaks, JSON parse vulnerability), 2 P2 issues (error handling, test gaps). See inline comments for specifics.

## Architecture

✅ **Good design decisions**:
- Clean separation of concerns (queue / persistence / runtime / worker)
- Pluggable backend strategy pattern
- Transaction-based event ordering in Postgres
- Advisory lock for event_id sequence

⚠️ **Risks**:
- Adding `@ts-nocheck` to ALL new files masks type safety issues in a critical path
- No migration strategy for existing SQLite users → breaking change
- Redis stale job requeue has edge cases (see inline)

## Testing

Tests added:
- ✅ agent-run-store.test.ts (SQLite persistence)
- ✅ agent-runtime.test.ts
- ✅ runs-store.test.ts
- ✅ project-fs.test.ts
- ✅ e2e/cluster-deployment.test.ts

❌ **Missing**:
- No tests for Redis queue edge cases (concurrent claims, stale job recovery, multi() transaction failures)
- No tests for Postgres connection pool exhaustion
- No tests for appendRunEvent race conditions despite advisory lock

## Deployment

⚠️ The new deploy/docker-compose.yml and deploy/.env.example are great additions, but the PR description doesn't mention:
- Migration path for existing deployments
- Backward compatibility guarantees
- Rollback procedure if clustering breaks

Please address the P1 inline findings before merge.

PR #1135, opened 2026-05-10
fix(web): replace native confirm with React dialog so Electron focus stays alive (#1129)

### LINKED ISSUE #1129: Input box bug
Problem: The input box cannot be clicked or used.
How to reproduce: First, create a project, and then delete it. At this point, if you click create again, you cannot focus on the input field after entering the project name. If the user creates it incorrectly and then deletes it, it can easily affect the experience. You need to force a restart to be able to input again. The above problem can be reproduced consistently.

Sid-Qin:
Fixes #1129

## Summary

Native browser/Electron \`confirm(...)\` leaves the parent web contents without restored input focus on dismiss. After the user clicks Delete on a project (or simply **cancels** the prompt), every input on the Designs page — Create project name, search, kanban filter — becomes unfocusable until the window is refocused.

The reproduction was nailed down in the issue thread:
- @Sakana-yuyu reproduced on desktop (PC)
- @Priyanshudotdev reproduced *even on cancel* (so it's not about deletion side effects), and observed it affecting both the search input and the Create panel input
- @lefarcen pointed at the three \`confirm(...)\` sites in \`DesignsTab.tsx\` and recommended replacing them with the \`SettingsDialog\` / \`PrivacyConsentModal\` pattern

### New file: [\`apps/web/src/components/ConfirmDialog.tsx\`](apps/web/src/components/ConfirmDialog.tsx)

Reusable confirmation dialog. Reuses the existing \`.modal-backdrop\` / \`.modal\` / \`.hint\` / \`.row\` CSS so this PR ships **no new styling**.

The piece that actually fixes #1129:
\`\`\`tsx
useEffect(() => {
  if (!open) return;
  previousFocusRef.current = document.activeElement as HTMLElement | null;
  queueMicrotask(() => confirmRef.current?.focus());
  function onKey(event: KeyboardEvent) { /* Escape */ }
  document.addEventListener('keydown', onKey);
  return () => {
    document.removeEventListener('keydown', onKey);
    previousFocusRef.current?.focus?.();   // ← restore on close
  };
}, [open]);
\`\`\`

A latest-onCancel ref keeps the open-effect stable across renders so the focus-restore cleanup only runs on \`open: true → false\`, not on every parent re-render that hands in a fresh closure.

### [\`apps/web/src/components/DesignsTab.tsx\`](apps/web/src/components/DesignsTab.tsx)

Replace **all three** \`confirm(...)\` call sites with a single \`pendingDelete\` discriminated-union state that drives the new dialog:

| Old call site | What now |
|---|---|
| \`handleDeleteLiveArtifact\` (live-artifact card delete) | \`requestDeleteLiveArtifact\` sets \`pendingDelete = { kind: 'artifact', ... }\` |
| Project delete in grid view (line 351) | \`requestDeleteProject\` sets \`pendingDelete = { kind: 'project', ... }\` |
| Project delete in kanban view (line 440) | same \`requestDeleteProject\` |

Confirm → \`commitPendingDelete\` dispatches to the corresponding handler and clears state. Cancel just clears state.

> Side-note on the artifact delete: lefarcen's scope was the two project-delete confirms; I included the artifact-delete one too because it's the same root cause, the same site (DesignsTab), and the same one-line replacement. Happy to split it back out if you prefer keeping the PR strictly to the maintainer-confirmed scope.

## Tests

- [\`apps/web/tests/components/ConfirmDialog.test.tsx\`](apps/web/tests/components/ConfirmDialog.test.tsx) (8 tests): mount/unmount, all four close paths (confirm / cancel / Escape / backdrop), inner-modal click isolation, and **the focus-restore-on-close path that #1129 actually depends on** (asserts that focus returns to a previously-focused trigger button after the dialog closes).
- [\`apps/web/tests/components/DesignsTab.deleteFlow.test.tsx\`](apps/web/tests/components/DesignsTab.deleteFlow.test.tsx) (3 tests): clicking the project delete button now opens the in-app dialog instead of firing \`onDelete\`; confirm fires \`onDelete\` with the project id and dismisses the dialog; cancel closes the dialog without firing \`onDelete\`.

\`pnpm --filter @open-design/web typecheck\` ✅
\`pnpm exec vitest run tests/components\` → **354/354 pass** (40 test files), no regressions.

## Out of scope

- I did not audit the rest of the codebase for other \`confirm(...)\` usages. There's at least one in a chat composer / conversation flow; happy to follow up in a separate PR.
- Did not add destructive-button color styling. The \`destructive\` prop sets a class but the existing CSS doesn't define a \`.destructive\` rule on dialog buttons — kept as a structural hook for a follow-up theming pass.

### DIFF
--- a/apps/web/src/components/ConfirmDialog.tsx
+++ b/apps/web/src/components/ConfirmDialog.tsx
@@ -1,0 +1,108 @@
+import { useEffect, useId, useRef } from 'react';
+
+interface Props {
+  /** When false the dialog is unmounted entirely. */
+  open: boolean;
+  title: string;
+  message: string;
+  confirmLabel: string;
+  cancelLabel: string;
+  /** Visually marks the confirm button as a destructive action. */
+  destructive?: boolean;
+  onConfirm: () => void;
+  onCancel: () => void;
+}
+
+/**
+ * Reusable confirmation dialog used in place of the native `confirm(...)`
+ * call. Native confirms in Electron leave the parent web contents
+ * without restored input focus on dismiss, so any input the user clicks
+ * after closing the dialog appears unfocusable until the window is
+ * refocused (#1129). A React-managed modal lives inside the same
+ * document and naturally restores focus to the previously-active
+ * element on close.
+ */
+export function ConfirmDialog({
+  open,
+  title,
+  message,
+  confirmLabel,
+  cancelLabel,
+  destructive,
+  onConfirm,
+  onCancel,
+}: Props): JSX.Element | null {
+  const titleId = useId();
+  const messageId = useId();
+  const confirmRef = useRef<HTMLButtonElement | null>(null);
+  const previousFocusRef = useRef<HTMLElement | null>(null);
+  // Stash the latest cancel handler in a ref so the open-effect below can
+  // wire up Escape / focus-restore once per open transition without
+  // re-subscribing every render when the parent passes a fresh closure.
+  const onCancelRef = useRef(onCancel);
+  useEffect(() => {
+    onCancelRef.current = onCancel;
+  }, [onCancel]);
+
+  useEffect(() => {
+    if (!open) return;
+    previousFocusRef.current = document.activeElement as HTMLElement | null;
+    // Defer focus to after the modal has mounted into the DOM.
+    queueMicrotask(() => confirmRef.current?.focus());
+
+    function onKey(event: KeyboardEvent) {
+      if (event.key === 'Escape') {
+        event.stopPropagation();
+        onCancelRef.current();
+      }
+    }
+    document.addEventListener('keydown', onKey);
+    return () => {
+      document.removeEventListener('keydown', onKey);
+      // Returning focus to the trigger keeps Electron's native focus
+      // tracking aligned with the React state, so the next click in the
+      // host inputs lands cleanly.
+      const previous = previousFocusRef.current;
+      previousFocusRef.current = null;
+      if (previous && typeof previous.focus === 'function') {
+        previous.focus();
+      }
+    };
+  }, [open]);
+
+  if (!open) return null;
+
+  return (
+    <div className="modal-backdrop" onClick={onCancel} data-testid="confirm-dialog-backdrop">
+      <div
+        className="modal modal-confirm"
+        role="alertdialog"
+        aria-modal="true"
+        aria-labelledby={titleId}
+        aria-describedby={messageId}
+        onClick={(event) => event.stopPropagation()}
+      >
+        <h2 id={titleId}>{title}</h2>
+        <p id={messageId} className="hint">{message}</p>
+        <div className="row">
+          <button
+            type="button"
+            onClick={onCancel}
+            data-testid="confirm-dialog-cancel"
+          >
+            {cancelLabel}
+          </button>
+          <button
+            ref={confirmRef}
+            type="button"
+            className={destructive ? 'destructive' : undefined}
+            onClick={onConfirm}
+            data-testid="confirm-dialog-confirm"
+          >
+            {confirmLabel}
+          </button>
+        </div>
+      </div>
+    </div>
+  );
+}
--- a/apps/web/src/components/ConfirmDialog.tsx
+++ b/apps/web/src/components/ConfirmDialog.tsx
@@ -34,6 +34,7 @@ export function ConfirmDialog({
 }: Props): JSX.Element | null {
   const titleId = useId();
   const messageId = useId();
+  const cancelRef = useRef<HTMLButtonElement | null>(null);
   const confirmRef = useRef<HTMLButtonElement | null>(null);
   const previousFocusRef = useRef<HTMLElement | null>(null);
   // Stash the latest cancel handler in a ref so the open-effect below can
@@ -54,6 +55,25 @@ export function ConfirmDialog({
       if (event.key === 'Escape') {
         event.stopPropagation();
         onCancelRef.current();
+        return;
+      }
+      if (event.key === 'Tab') {
+        // Trap Tab inside the alertdialog. Native confirm(...) blocks all
+        // background interaction; an in-app modal that lets Tab walk out
+        // to the host page is a non-modal modal and lets keyboard users
+        // activate unrelated controls (e.g. another project card) before
+        // resolving the destructive prompt. Cycle between the two
+        // buttons in both directions.
+        const cancel = cancelRef.current;
+        const confirm = confirmRef.current;
+        if (!cancel || !confirm) return;
+        event.preventDefault();
+        const active = document.activeElement;
+        if (event.shiftKey) {
+          (active === cancel ? confirm : cancel).focus();
+        } else {
+          (active === confirm ? cancel : confirm).focus();
+        }
       }
     }
     document.addEventListener('keydown', onKey);
@@ -64,8 +84,17 @@ export function ConfirmDialog({
       // host inputs lands cleanly.
       const previous = previousFocusRef.current;
       previousFocusRef.current = null;
-      if (previous && typeof previous.focus === 'function') {
+      if (previous && previous.isConnected && typeof previous.focus === 'function') {
         previous.focus();
+      } else {
+        // Trigger was unmounted while the dialog was open — typical for
+        // destructive flows where confirming removes the host element
+        // (project card → its delete button) in the same React batch.
+        // Focusing a detached node is a no-op and leaves Electron's
+        // focus tracker stranded, which is the very regression this
+        // dialog exists to prevent. Park focus on the document body so
+        // the next user input has a valid focus owner to route from.
+        document.body.focus({ preventScroll: true });
       }
     };
   }, [open]);
@@ -86,6 +115,7 @@ export function ConfirmDialog({
         <p id={messageId} className="hint">{message}</p>
         <div className="row">
           <button
+            ref={cancelRef}
             type="button"
             onClick={onCancel}
             data-testid="confirm-dialog-cancel"
--- a/apps/web/tests/components/DesignsTab.deleteFlow.test.tsx
+++ b/apps/web/tests/components/DesignsTab.deleteFlow.test.tsx
@@ -2,10 +2,10 @@
 
 import { cleanup, fireEvent, render, screen, waitFor } from '@testing-library/react';
 import { afterEach, describe, expect, it, vi } from 'vitest';
-import type { Project } from '@open-design/contracts';
+import type { LiveArtifactSummary, Project } from '@open-design/contracts';
 
 import { DesignsTab } from '../../src/components/DesignsTab';
-import { fetchLiveArtifacts } from '../../src/providers/registry';
+import { deleteLiveArtifact, fetchLiveArtifacts } from '../../src/providers/registry';
 
 vi.mock('../../src/providers/registry', () => ({
   deleteLiveArtifact: vi.fn(),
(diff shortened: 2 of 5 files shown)


lefarcen:
Agent-to-agent note: related research completed for PR #1135. Explicit closing ref verified: #1129 (native Electron confirm dismiss leaves Designs inputs unfocusable). Verified related prior modal/native-dialog work: #732 and #738 replaced Electron-broken window.prompt flows with in-app modals; #875 and #877 added window.confirm confirmations and cited DesignsTab confirm as the destructive-action precedent. synclo_memory_search was attempted but unavailable in this environment. Issue #1129 has been updated with a status pointer to #1135.

apps/web/src/components/ConfirmDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+    if (!open) return;
+    previousFocusRef.current = document.activeElement as HTMLElement | null;
+    // Defer focus to after the modal has mounted into the DOM.
+    queueMicrotask(() => confirmRef.current?.focus());
+
+    function onKey(event: KeyboardEvent) {
+      if (event.key === 'Escape') {
+        event.stopPropagation();
+        onCancelRef.current();
+      }
```
mrcfps, line 77: This effect only handles `Escape`, so keyboard focus can still `Tab` out of the confirmation dialog and land on the Designs page controls behind it while the destructive prompt is open. Native `confirm(...)` blocked all background interaction, so this is a behavior regression in the new delete flow and it makes it possible to trigger unrelated actions before the pending delete is resolved. Please trap focus inside the modal while `open` is true (or mark the rest of the page inert/aria-hidden for the modal lifetime) and add a regression test in `apps/web/tests/components/ConfirmDialog.test.tsx` that verifies `Tab`/`Shift+Tab` cannot escape the dialog.

mrcfps (commented):
@Sid-Qin Thanks for fixing the Electron focus-loss regression — the focus-restore path and delete-flow coverage look solid overall. I found one follow-up in the new dialog behavior that should be tightened up before merge.


apps/web/src/components/ConfirmDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+      if (previous && typeof previous.focus === 'function') {
+        previous.focus();
+      }
+    };
+  }, [open]);
+
+  if (!open) return null;
+
+  return (
+    <div className="modal-backdrop" onClick={onCancel} data-testid="confirm-dialog-backdrop">
```
lefarcen, line 105: **P2** — `aria-modal="true"` is declared, but focus is not trapped and the background is not made inert. Keyboard users can Tab out of the alertdialog into the Designs page while the confirmation is open, which makes this a non-modal modal and can allow activating underlying controls.

Add a small Tab loop between cancel/confirm buttons (capture Tab at the end and cycle back to the first button) or use an existing modal/inert pattern from the app.

apps/web/src/components/ConfirmDialog.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+    document.addEventListener('keydown', onKey);
+    return () => {
+      document.removeEventListener('keydown', onKey);
+      // Returning focus to the trigger keeps Electron's native focus
+      // tracking aligned with the React state, so the next click in the
+      // host inputs lands cleanly.
+      const previous = previousFocusRef.current;
+      previousFocusRef.current = null;
+      if (previous && typeof previous.focus === 'function') {
```
lefarcen: **P2** — Focus restoration blindly calls `previous.focus()` even when the original trigger may have been removed, especially on confirmed project deletion where the card/delete button can unmount in the same React batch. That can make the core "restore focus on close" path no-op for the main delete case.

Check `previous.isConnected` and fall back to a stable focus target, such as the Designs search input or another visible app container:
```tsx
if (previous.isConnected) {
  previous.focus();
} else {
  // fallback to stable container or search input
  document.querySelector('[data-designs-search]')?.focus();
}
```

apps/web/tests/components/DesignsTab.deleteFlow.test.tsx
```diff
@@ -0,0 +1,111 @@
...
+};
+
+describe('DesignsTab delete flow', () => {
+  afterEach(() => {
+    cleanup();
+    vi.mocked(fetchLiveArtifacts).mockReset();
+    vi.mocked(fetchLiveArtifacts).mockResolvedValue([]);
+  });
+
+  it('opens the in-app confirm dialog when the project delete button is clicked', async () => {
```
lefarcen, line 55: **P3** — Tests only cover grid project deletion, but the PR changes three call sites: grid project delete, kanban project delete, and live-artifact delete. Add coverage for kanban delete and artifact delete, including successful `deleteLiveArtifact` removal and cancel-without-API-call.

(Non-blocking, but helps ensure the pattern is robust across all three surfaces.)

lefarcen (changes requested):
Hi @Sid-Qin! 👋 Thanks for this PR — the focus-restore approach is solid and the reproduction is well-documented. Found 2 issues that need fixing before merge:

**Focus handling gaps:**
- The alertdialog doesn't trap focus (keyboard users can Tab out while the dialog is open)
- Focus restoration can fail when the delete button unmounts on confirm

See inline comments for specifics. Once those are addressed this should be ready. Thanks for making open-design better!

lefarcen:
Agent-to-agent review note: reviewed PR #1135 with code-correctness and reasoning-completeness lenses. Outcome: request_changes. Main blockers: ConfirmDialog declares aria-modal but does not trap focus/inert background; focus restore blindly calls focus on the previous element even when the delete trigger may unmount, so the core Electron focus fix can no-op in the confirmed-delete path. Test gap: only grid project delete is covered; kanban project delete and live-artifact delete should be covered too.

Sid-Qin:
@lefarcen @mrcfps Rebased onto current `main` (no conflicts — PR-touched files were untouched by the ~40 commits merged since) and pushed `c847c2d3` to address the review. Two commits on the branch now: the original fix (`aba31243` after rebase) and the review-response commit on top.

**Each finding, point by point:**

**P2 — Tab focus trap** ([review thread](https://github.com/nexu-io/open-design/pull/1135#pullrequestreview), all three of Looper/lefarcen/Codex flagged this)
- `ConfirmDialog.tsx` keydown handler now traps `Tab` and `Shift+Tab` and cycles between the cancel and confirm buttons in both directions.
- Pinned by new `ConfirmDialog.test.tsx` case `traps Tab inside the dialog, cycling between cancel and confirm` — exercises both keys from both starting buttons.
- Verified RED before this commit (no Tab branch existed; document.activeElement walked out of the dialog), GREEN after.

**P2 — `previous.focus()` on a detached node** (lefarcen)
- `ConfirmDialog.tsx` close path now gates on `previous.isConnected` before calling `previous.focus()`, and falls back to `document.body.focus({ preventScroll: true })` otherwise. Body focus is the neutral, recoverable state Electron's focus tracker needs; without the fallback the unmount-during-open path (project card disappears in the same React batch as the dialog close) silently leaves focus stranded on a detached node — which is exactly the #1129 regression we're trying to prevent.
- Pinned by new `ConfirmDialog.test.tsx` case `falls back to document.body when the previously focused element has been detached` — synthetically reproduces the unmount-while-open path.
- I considered the `[data-designs-search]` example you mentioned in the inline comment. The reason I went with body fallback instead is to keep ConfirmDialog reusable — it's already shipping in three call sites (project grid / project kanban / live-artifact) and the next one won't necessarily have a sensible search-input anchor. Happy to wire a `fallbackFocus` prop and pass `[data-designs-search]` from `DesignsTab` if you'd rather have the specific behavior.

**P3 — extend coverage to the two other call sites** (lefarcen)
- `DesignsTab.deleteFlow.test.tsx` now also covers:
  - `routes kanban view project delete through the React dialog` (toggles to kanban view, confirms delete via the dialog, asserts `onDelete` fires)
  - `routes live-artifact delete through the React dialog and calls deleteLiveArtifact only on confirm` (asserts `deleteLiveArtifact(projectId, artifactId)` is the called provider, not `onDelete`)
  - `does not call deleteLiveArtifact when the live-artifact dialog is cancelled` (complement)
- Cleared `localStorage` in `afterEach` so the persisted grid/kanban toggle doesn't leak between tests.

**`Validate workspace` CI failure** (root cause = PR's own e2e fallout, not flagged by reviewers but visible in the run)
- `e2e/ui/project-management-flows.test.ts:240` "home design card deletion supports cancel and confirm flows" was still using Playwright's `page.once('dialog', ...)` handler, which **only fires for native `window.confirm/alert/prompt`**. After we swapped in the React modal the callback never resolves, the in-app dialog stays open, the `.modal-backdrop` keeps intercepting pointer events, and the very next `await designCard.hover()` retries 51× then times out at 30s. (Full CI log: [actions run 25621480575](https://github.com/nexu-io/open-design/actions/runs/25621480575/job/75208865239))
- Migrated both the cancel and confirm flows to drive the dialog by its `role="alertdialog"` plus the `confirm-dialog-cancel` / `confirm-dialog-confirm` testids. Same assertions, no native-dialog plumbing.
- The other `page.on('dialog', …)` sites in `app.test.ts` (`runDesignFilesDeleteFlow`, `runConversationDeleteRecoveryFlow`) and `project-management-flows.test.ts:223` (design file delete) are intentionally left alone — they target `FileWorkspace.handleDelete` / `ChatPane` delete which still use native `confirm(...)`, confirmed by greping `apps/web/src/` for remaining sites.

**Local verification on `c847c2d3`:**
- `pnpm --filter @open-design/web typecheck` — clean (exit 0)
- `pnpm --filter @open-design/web test` — **84 files, 745/745 passed** (the four pre-existing `SettingsDialog.execution.test.tsx` `window.localStorage` failures you saw on the previous head no longer reproduce on current `main`)
- The e2e change itself I could not exercise locally without a Playwright + daemon stack; the new flow is a literal mechanical migration from native-dialog API to alertdialog-role API and the assertions are unchanged. Waiting on CI for the real signal.

The two A11y P2 fixes will also dismiss the `.modal-backdrop` pointer-events-intercept symptom in CI: with focus correctly trapped and restored, the underlying dialog lifecycle is consistent and the backdrop only exists while `open=true` (as it always did — the failure was the test never resolving the dialog, not the component leaking).

mrcfps (approved):
@Sid-Qin Thanks for tightening the follow-up review points here. I re-checked the updated ConfirmDialog focus handling, the detached-trigger fallback, the expanded unit coverage for all three delete entry points, and the Playwright migration for the new in-app alertdialog flow, and I didn't find any remaining actionable issues in the changed ranges.


lefarcen (approved):
All P2/P3 findings addressed on current head c847c2d3.

**Focus trap** (P2) — fixed. Tab/Shift+Tab now cycle between cancel/confirm buttons while the dialog is open.

**Detached-focus fallback** (P2) — fixed. Close path now checks `previous.isConnected` before calling `previous.focus()` and falls back to `document.body.focus({ preventScroll: true })` otherwise.

**Test coverage** (P3) — extended to kanban view project delete and live-artifact delete flows.

**CI e2e regression** — migrated from native `page.on('dialog')` to React `alertdialog` role selectors.

LGTM. 🎉

lefarcen:
Hey @Sid-Qin 👋 Quick status update: mrcfps and the bot both approved the current head (`c847c2d3`) — the review side is fully done. The one remaining blocker is a merge conflict with `main` that appeared after your last push.

To unblock: rebase onto `main`, resolve any conflicts, then mark the PR ready for review and we can move to merge. Your review-response work (focus trap, `isConnected` fallback, expanded coverage across all three delete paths, Playwright migration) addressed every finding thoroughly — this is just the rebase step left. 🙌

lefarcen:
Closing this as superseded by #1161. That merged PR already replaced the DesignsTab native `confirm(...)` delete paths with an in-page React `alertdialog`, which covers the Electron focus-loss root cause from #1129. Thanks again for the thorough review-response work here — the overlap just moved under the larger Designs tab redesign before this branch could be rebased.

Issue #1077, opened 2026-05-09
Question: Is there no way to deploy the entire project as a one unified web-app to vercel ?


ButterHost69:
Right now deploying to vercel uploads that .html page as the main page to the vercel. Is there a way to deploy the entire application as a single web-app with multiple .html files to vercel ?

lefarcen:
Thanks for the report. Current Vercel deploy is not a whole-project/static-site deploy yet.

Right now open-design deploys the HTML file you opened as the entry page, rewrites it to `index.html`, and uploads that plus referenced local assets such as CSS, JS, images, fonts, `srcset`, and CSS `url()` dependencies. It can upload multiple files to Vercel, but it does not currently include every `.html` file in the project, and normal links like `<a href="about.html">` are intentionally not collected as deploy dependencies.

One question so we scope this correctly: are the extra pages regular project files linked from the current page, for example `about.html`, `pricing.html`, etc., and you expect those routes to work under the same Vercel deployment URL?

ButterHost69:
Yes, I have a multi-page project with landing.html, pricing.html, and about.html linked from the main page, and I'd like them to work under one Vercel deployment URL with clean routes (/pricing, /about)

lefarcen:
Thanks for clarifying. That confirms this is currently a missing capability rather than a configuration issue.

Right now Open Design deploy treats the selected HTML file as the deploy entry, rewrites that file to `index.html`, and uploads it plus referenced assets like CSS, JS, images, fonts, `srcset`, and CSS `url()` dependencies. It intentionally does not follow normal navigation links like `<a href="pricing.html">`, so sibling pages such as `pricing.html` and `about.html` are not collected as pages in the same deployment today.

The good news is this should be implementable as a static multi-page deploy. It should not require SPA-style client-side routing or a server runtime. For Vercel clean routes like `/pricing` and `/about`, we would likely need to upload the linked HTML pages and include static routing config such as `cleanUrls: true` or generated rewrites.

I think we should track this as an enhancement. The main product decision is whether "deploy project" should include only pages reachable from the selected entry file, or every HTML/static file in the project. Until that lands, the practical workarounds are deploying each page separately or moving the multi-page site into a static site generator/project export flow and deploying that to Vercel directly.

ButterHost69:
I think including only the reachable pages would be better, as we can that way avoid unnecessary files, like in my case I have noticed that there are backup files that are always produced before new changes are made.

[...]

PR #1149, opened 2026-05-10
feat(daemon): deploy linked HTML pages together as a multi-page site on Vercel with clean URLs (#1077)

ButterHost69:
#1077 

Summary
---
Add multi-page site deploy support to Vercel. When deploying an HTML file with \<a href\> links to other .html pages in the project, the daemon automatically discovers those linked pages and deploys them together under one Vercel deployment URL with clean URLs (/pricing instead of pricing.html).

Changes
---
### 1. apps/daemon/src/deploy.ts 
- Added LinkedPageInfo type and collectLinkedHtmlPages() function to discover \<a href\> targets pointing to other .html files
- Modified buildDeployFilePlan to accept linked pages, read them before the asset BFS loop so their CSS/JS/image dependencies share the deduplication path, and write them with their original filenames
- Modified rewriteEntryHtmlReferences/rewriteHtmlAttributes to accept linkedPages — \<a href="pricing.html"\> is rewritten to \<a href="/pricing"\> in the deployed HTML
- Includes a vercel.json with { "cleanUrls": true } in the deployment when linked pages are present, so /pricing resolves pricing.html

### 2. apps/daemon/src/server.ts 
- Accepts linkedPages in POST /api/projects/:id/deploy and .../deploy/preflight — validates paths, builds LinkedPageInfo[] records, passes to planner

### 3. apps/web/src/providers/registry.ts 
- Added linkedPages?: string[] param to deployProjectFile(), passes it in the request body
apps/web/src/components/FileViewer.tsx
- Added detectLinkedHtmlPages() — fetches the entry HTML text, parses \<a href="...\.html?"\> links, caps at 20
- Called automatically from openDeployModal() when the entry file is an .html file
- Passes deployLinkedPages to deployProjectFile
- Renders a notice in the deploy modal: "Includes 2 linked page(s): pricing.html, about.html"

### 4. apps/web/src/i18n 
- Added fileViewer.linkedPagesNotice with translated values in all 18 supported languages
packages/contracts/src/api/projects.ts 
- Added linkedPages?: string[] to DeployProjectFileRequest and DeployPreflightRequest

How it works
---
1. User deploys landing.html — the web client calls detectLinkedHtmlPages() which fetches the file text and parses \<a href="...\.html"> links, finding pricing.html and about.html. These are passed to the deploy API as linkedPages.
2. The daemon's POST /api/projects/:id/deploy route validates each linked page path via validateProjectPath(), builds LinkedPageInfo[] records, and passes them to buildDeployFileSet().
3. buildDeployFilePlan() reads each linked page before the asset BFS loop, rewrites their \<a href\> links to clean URLs, feeds their CSS/JS/image references into the pending queue for shared deduplication, and writes them after the BFS with their original filenames (only the entry page is renamed to index.html).
4. When linked pages are present, a vercel.json with { "cleanUrls": true } is injected into the deployment file set, so /pricing resolves pricing.html on Vercel.
5. rewriteHtmlAttributes() rewrites \<a href="pricing.html"\> → \<a href="/pricing"\> in the deployed HTML, matching each linked page by its resolved project path.
6. The deploy UI shows a notice in the deploy modal: "Includes 2 linked page(s): pricing.html, about.html".
7. Single-file deploys without \<a href\> links skip the entire linked-page code path — no vercel.json is generated, no extra files are read, and the existing single-page behavior is unchanged.

Context
---
Previously, Open Design deployed a single HTML file as index.html along with its asset dependencies (CSS, JS, images, fonts). Navigation links like \<a href="about.html"\> were intentionally not collected — the model was single-page artifact deploy.
This feature adds site-level deploy: the entry page is still renamed to index.html, but linked pages keep their original names. Clean URLs are enabled via vercel.json so internal navigation works without .html extensions. Only Vercel is touched — Cloudflare Pages already handles multi-file deploys differently and is unaffected. Backward-compatible: single-file deploys without linked pages skip the new code path entirely.


apps/daemon/src/deploy.ts
```diff
@@ -242,8 +243,9 @@ export async function buildDeployFilePlan(projectsRoot: string, projectId: strin
   const entry = await readProjectFile(projectsRoot, projectId, entryPath, options.metadata);
   const html = entry.buffer.toString('utf8');
   const entryBase = path.posix.dirname(entryPath);
+  const linkedPages = options.linkedPages ?? [];
```
lefarcen: **P1 — Daemon discovery unused**

`buildDeployFilePlan` only uses `options.linkedPages` (what the UI sends), while the new `collectLinkedHtmlPages()` helper is never called. Any caller that doesn't pass the UI-derived list, or any link syntax the UI misses, still deploys only the entry page.

**Fix**: Make linked-page discovery server-authoritative:
- Call `collectLinkedHtmlPages(html, entryBase)` in `buildDeployFilePlan` to discover pages from the entry HTML.
- Merge with `options.linkedPages` (UI can still hint, but daemon verifies).
- Add daemon tests for multi-page deploys covering various <a href> patterns.

apps/daemon/src/deploy.ts
```diff
@@ -318,6 +360,33 @@ export async function buildDeployFilePlan(projectsRoot: string, projectId: strin
     }
   }
 
+  // Write each linked page into the file set with its own deployed
+  // name (not renamed to index.html). The entry page is already
+  // registered above.
+  for (const meta of linkedPageMetas) {
+    files.set(meta.info.deployedName, {
```
lefarcen, line 401: **P1 — Path flattening collision**

Writing linked pages by `basename` flattens paths and can overwrite files. For example, links to `docs/pricing.html` and `archive/pricing.html` both deploy as `pricing.html`, and the rewrite later sends both to `/pricing`.

**Fix**: Preserve deploy-relative paths:
- Use the full relative path as the deployed name (e.g., `docs/pricing.html`).
- Rewrite clean URLs to match: `/docs/pricing` routes to `docs/pricing.html`.
- Or explicitly detect collisions and error early with a clear message.

apps/daemon/src/deploy.ts
```diff
@@ -318,6 +360,33 @@ export async function buildDeployFilePlan(projectsRoot: string, projectId: strin
...
+      contentType: meta.mime,
+      sourcePath: meta.info.path,
+    });
+  }
+
+  // When linked pages are present, include a vercel.json that
+  // enables clean URLs so /pricing resolves pricing.html without
+  // the extension. Only added for Vercel deploys — Cloudflare Pages
+  // handles clean URLs differently.
+  if (linkedPages.length > 0) {
```
lefarcen: **P1 — Cloudflare scope drift**

This injects `vercel.json` whenever linked pages exist, regardless of `options.providerId`. The server also passes `linkedPages` for Cloudflare deploys. That contradicts the stated "Cloudflare unaffected" scope and uploads Vercel-specific config to Cloudflare Pages.

**Fix**: Pass provider context into the planner:
- Add `providerId` to `DeployOptions`.
- Only inject `vercel.json` when `providerId === VERCEL_PROVIDER_ID`.
- For Cloudflare, either skip linked-page logic entirely or handle it via a different mechanism (if multi-page Cloudflare is intentional, the PR body should say so).

apps/web/src/components/FileViewer.tsx
```diff
@@ -3964,6 +3965,34 @@ function HtmlViewer({
...
+    } else {
+      setDeployLinkedPages([]);
+    }
+  }
+
+  async function detectLinkedHtmlPages() {
+    try {
+      const text = await fetchProjectFileText(projectId, file.name);
+      if (!text) { setDeployLinkedPages([]); return; }
+      const anchorRe = /<a\s[^>]*href\s*=\s*"([^"]*\.html?)"/gi;
```
lefarcen: **P2 — Client detector misses valid hrefs**

The client detector only matches double-quoted hrefs ending immediately in `.html`, so valid links like `<a href='pricing.html'>`, `<a href="pricing.html#faq">`, or `<a href="pricing.html?ref=nav">` are not included.

**Fix**: If client detection remains, use the same attribute/link semantics as the daemon:
- Match single and double quotes.
- Strip query/hash before the `.html` check.
- Add web tests for quoted, single-quoted, query, and hash cases.

(Alternatively, if daemon discovery becomes authoritative per P1, this becomes less critical — but still worth fixing for preflight UI consistency.)

lefarcen (dismissed):
Hey @ButterHost69! Thanks for tackling this — multi-page Vercel deploys with clean URLs is a genuinely useful addition. I've reviewed the implementation and found four must-fix issues before this can merge:

1. **Daemon-side discovery unused** — The PR advertises automatic linked-page discovery, but the daemon never actually calls collectLinkedHtmlPages(). The planner only uses what the UI sends, so any link syntax the UI misses (single quotes, query params, hash fragments) won't deploy.

2. **Path flattening collision risk** — Linked pages are written by basename only, so docs/pricing.html and archive/pricing.html both become pricing.html and overwrite each other.

3. **Cloudflare scope drift** — vercel.json is injected whenever linkedPages exist, regardless of providerId, and the server passes linkedPages for Cloudflare too. This contradicts the "Cloudflare unaffected" statement.

4. **Client detector misses valid hrefs** — The regex only matches double-quoted hrefs ending in .html, missing single quotes, query strings, and hash fragments.

See inline comments for details + concrete fixes. Once these are addressed, the feature will be solid. Looking forward to the next iteration!

apps/web/src/components/FileViewer.tsx
mrcfps: Because deploy now relies on this client-side detector to decide which extra HTML files to send, the current regex/parser is too narrow for valid pages. It posts the raw `href` instead of resolving it relative to `file.name`, so an entry file like `pages/home.html` with `<a href="about.html">` makes the daemon validate `/about.html` at project root and fail even though `pages/about.html` exists. It also only matches double-quoted `*.html` values with no query/hash suffix, so `<a href='about.html'>` and `about.html?ref=1` are skipped entirely. Please resolve links against the entry file directory (or move discovery server-side / reuse the daemon parser) so linked-page deploys work for normal relative HTML navigation.

apps/web/src/components/FileViewer.tsx
```diff
@@ -3964,6 +3965,34 @@ function HtmlViewer({
...
+      const found: string[] = [];
+      let match;
+      while ((match = anchorRe.exec(text)) !== null) {
+        const href = match[1];
```
mrcfps: This loop currently does not typecheck: `match[1]` is `string | undefined`, so the new `href.startsWith(...)` / `href.split(...)` calls fail both the `Validate workspace` and `build` jobs on this head. Since the PR cannot pass required CI in its current state, this blocks merge. Please narrow or guard `match[1]` before using it (for example `const href = match[1]; if (!href) continue;`) or switch to a parser helper that returns a guaranteed string.

apps/daemon/src/server.ts
```diff
@@ -4306,11 +4307,30 @@ export async function startServer({
...
+      const linkedPagePaths = Array.isArray(linkedPages)
+        ? (linkedPages.filter((p: unknown): p is string => typeof p === 'string' && p.length > 0))
+        : [];
+      const deployLinkedPages: { path: string; deployedName: string }[] = [];
+      for (const rawPath of linkedPagePaths) {
+        try {
+          const resolved = validateProjectPath(rawPath);
+          deployLinkedPages.push({
+            path: resolved,
+            deployedName: path.posix.basename(resolved),
```
mrcfps: Flattening every linked page to `path.posix.basename(resolved)` loses the directory information that makes multi-page sites unique. `docs/index.html` and `blog/index.html` both become `index.html`, so one overwrites the other later in `buildDeployFilePlan`, and the clean-URL rewrite logic also maps both links to `/index`. That breaks a common nested-site layout and can even replace the entry page at `/`. Please keep a unique deploy-relative path for linked pages (for example their normalized project-relative path, or another directory-aware route mapping) instead of discarding the parent directories here.

mrcfps (changes requested):
@ButterHost69 Thanks for pushing the multi-page deploy support here. I reviewed the changed ranges and found a few blockers to address before this can merge: the new link-discovery path currently fails CI typechecking and still breaks valid nested/duplicate-name HTML site layouts.


apps/daemon/src/deploy.ts
```diff
@@ -318,6 +362,33 @@ export async function buildDeployFilePlan(projectsRoot: string, projectId: strin
...
```
lefarcen: P2 — `vercel.json` is still added whenever linked pages exist, regardless of the selected provider, so Cloudflare Pages deployments still upload a Vercel-specific config file. Gate this on `options.providerId !== CLOUDFLARE_PAGES_PROVIDER_ID` and pass `providerId` from the deploy route into `buildDeployFileSet`, not only from preflight.

lefarcen (commented):
Hey @ButterHost69! Nice iteration — you addressed the 3 major issues from last round:

✅ Daemon-side discovery now works (no longer client-only)
✅ Path preservation instead of basename flattening
✅ Clean URL rewrites now handle query/hash correctly

One non-blocking concern remains (see inline). Once that's fixed, the feature will be solid!

apps/daemon/src/deploy.ts
```diff
@@ -1214,7 +1285,37 @@ export function resolveReferencedPath(raw: unknown, baseDir: string) {
...
+  const pages = new Map<string, LinkedPageInfo>();
+  for (const tag of parseHtmlTags(String(html))) {
+    if (tag.name !== 'a') continue;
+    const attrs = parseHtmlAttributes(tag.attrs);
+    const href = attrs.get('href');
+    if (!href || typeof href !== 'string') continue;
+    const hrefStr: string = href;
+    if (isExternalUrl(hrefStr)) continue;
+    if (hrefStr.startsWith('#') || hrefStr.startsWith('javascript:') || hrefStr.startsWith('mailto:')) continue;
+    if (hrefStr.startsWith('/')) continue;
```
mrcfps: Root-relative HTML links are still excluded from linked-page discovery here. Because this branch now relies on `collectLinkedHtmlPages()` to decide which extra pages get deployed, an entry document with `<a href="/pricing.html">` never adds `pricing.html` to `linkedPages`, so the planner uploads only `index.html` and the deployed navigation to `/pricing` or `/pricing.html` breaks even though the file exists in the project. Please let leading-slash `*.html` values flow through `resolveReferencedPath()` instead of returning early here, and add a deploy-plan test that covers both `pricing.html` and `/pricing.html` so this common navigation pattern stays working.

mrcfps (changes requested):
@ButterHost69 Thanks for the follow-up here — moving discovery into the daemon and preserving nested paths fixed most of the earlier gaps. I found one remaining blocker in the current changed range: root-relative HTML navigation links are still skipped during linked-page discovery, so a common multi-page site shape can still ship with broken internal navigation.


lefarcen (commented):
Hey @ButterHost69! Excellent progress — the Cloudflare scope issue is fixed. The daemon now correctly gates `vercel.json` on `providerId !== CLOUDFLARE_PAGES_PROVIDER_ID` and passes `providerId` through the deploy route, so Cloudflare deploys no longer receive Vercel-specific config.

I see mrcfps found one more edge case (root-relative links) — once that's addressed, the feature should be solid. Great work iterating!

lefarcen (commented):
Hey @ButterHost69! Excellent iteration — the root-relative link blocker is fixed. Links like `<a href="/pricing.html">` now flow through `resolveReferencedPath()` and are correctly discovered/deployed/rewritten. The feature looks solid from my end!

mrcfps found this on an earlier head, so once they re-review the current head and everything checks out, this should be ready. Great work persisting through the iterations!

apps/daemon/src/deploy.ts
```diff
@@ -276,6 +280,46 @@ export async function buildDeployFilePlan(projectsRoot: string, projectId: strin
...
+        options.hookScriptUrl ?? process.env.OD_DEPLOY_HOOK_SCRIPT_URL,
+      ),
+      mime: pageFile.mime,
+    });
+
+    for (const ref of extractHtmlReferences(pageHtml)) {
+      pending.push({ ref, base: pageBase });
+    }
+    for (const ref of extractInlineCssReferences(pageHtml)) {
+      pending.push({ ref, base: pageBase });
```
mrcfps, line 351: This new loop only queues asset references from each discovered linked page. Because it never feeds `collectLinkedHtmlPages(pageHtml, pageBase)` back into the linked-page set, discovery stops after the entry document. A simple chain like `index.html -> about.html -> pricing.html` will therefore upload `about.html` but never add `pricing.html`, and `rewriteEntryHtmlReferences(pageHtml, pageBase, linkedPages)` also leaves the `about.html` navigation untouched because `pricing.html` is not in `linkedPages`. That breaks a normal multi-page flow on the second hop even though the source files are present. Please make linked-page discovery recursive/iterative until no new local `*.html` targets are found, then add a deploy-plan test that covers at least a two-hop chain so this path stays working.

mrcfps (changes requested):
@ButterHost69 Thanks for the continued iteration here — moving discovery into the daemon and fixing the root-relative case closed most of the earlier gaps. I found one remaining blocker in the current changed range: linked-page discovery still stops at the entry document, so common multi-hop site navigation can still deploy with missing pages.


lefarcen (commented):
Hey @ButterHost69 👋

The new iteration addresses mrcfps's recursive linked-page discovery blocker — the iterative loop now discovers multi-hop chains correctly. I verified the changed range and found no new P1/P2 issues.

mrcfps's latest review was on the previous head (`24fe8ca5`); the current head (`8e6b23c6`) hasn't been re-reviewed by them yet, so this PR is still awaiting their re-review on the current changes.

Thanks for the continued iteration!

mrcfps (commented):
@ButterHost69 Thanks for the continued iteration here — moving discovery into the daemon, making it transitive, and scoping `vercel.json` to Vercel-only deploys addressed most of the earlier gaps. I found one remaining issue in the current diff that can still change which document becomes the deployment landing page; details are inline.


apps/daemon/src/deploy.ts
```diff
@@ -318,6 +384,33 @@ export async function buildDeployFilePlan(projectsRoot: string, projectId: strin
     }
   }
 
+  // Write each linked page into the file set with its own deployed
+  // name (not renamed to index.html). The entry page is already
+  // registered above.
+  for (const meta of linkedPageMetas) {
+    files.set(meta.info.deployedName, {
+      file: meta.info.deployedName,
```
mrcfps, line 402: Because linked pages now keep their full project-relative path, a linked root `index.html` is still a special collision case here: the selected entry document is always registered up front as `files.set('index.html', …)` at the top of `buildDeployFilePlan()`, and this later `files.set(meta.info.deployedName, …)` write will overwrite it whenever a nested entry page links to `/index.html` or `../index.html`. In a common flow like deploying `docs/about.html` with a “Home” link back to the project root, the deployment ends up opening on the linked root page instead of the chosen entry page, so the main route is wrong even though all source files exist. Please reserve `index.html` for the selected entry file and either fail preflight/deploy with a clear collision error when a linked page also resolves to root `index.html`, or remap that linked page to a non-conflicting deploy path before inserting it into `files`. A regression test for `entryPath !== index.html` plus a linked root home page would help lock this down.

ButterHost69:
Went with failing instead of silently renaming as it is simpler, and leaves it up to the user to decide what to do about the conflict.

lefarcen (commented):
Hey @ButterHost69 👋

The collision guard is solid — `page.deployedName === 'index.html'` now throws a clear DeployError before the linked page can overwrite the selected entry. This addresses mrcfps's non-blocking concern from the prior head.

I don't see any new P1/P2 issues in this iteration. Nice work closing the loop!

apps/daemon/src/deploy.ts
```diff
@@ -242,8 +243,12 @@ export async function buildDeployFilePlan(projectsRoot: string, projectId: strin
   const entry = await readProjectFile(projectsRoot, projectId, entryPath, options.metadata);
   const html = entry.buffer.toString('utf8');
   const entryBase = path.posix.dirname(entryPath);
+  const linkedPages = options.linkedPages && options.linkedPages.length > 0
+    ? options.linkedPages
+    : collectLinkedHtmlPages(html, entryBase);
```
mrcfps, line 248: The new linked-page planner in `buildDeployFilePlan()` / `collectLinkedHtmlPages()` now owns several tricky behaviors — entry-page discovery, nested-path preservation, transitive traversal, clean-URL rewriting, provider-specific `vercel.json` injection, and the root `index.html` collision guard — but this PR still does not add any matching coverage in `apps/daemon/tests/deploy.test.ts`. That matters because this exact code path already needed multiple follow-up fixes during review (nested directories, root-relative links, transitive chains, and entry-page collisions), so without regression tests a later refactor can easily reopen one of those cases without anyone noticing. Please add a small matrix in `apps/daemon/tests/deploy.test.ts` that exercises at least: (1) a direct linked page with clean-URL rewrite, (2) nested linked pages preserving `docs/about.html -> /docs/about`, (3) a transitive chain like `index -> about -> pricing`, (4) a root-relative link such as `/pricing.html`, (5) the `../index.html` collision error, and (6) Cloudflare preflight/deploy excluding `vercel.json`.

mrcfps (commented):
@ButterHost69 Thanks for the continued iteration here — the linked-page planner looks much tighter now. I left one non-blocking suggestion to lock this new deploy path down with regression coverage before it drifts again.


lefarcen (commented):
Hey @ButterHost69! 👋 Excellent follow-through on the test coverage — the added `apps/daemon/tests/deploy.test.ts` now exercises all six scenarios mrcfps requested (direct linked page, nested directories, transitive chains, root-relative links, `../index.html` collision, Cloudflare-specific exclusion). No new P1/P2 issues found.

From my side, this looks solid. Deferring final approval to a maintainer.

mrcfps (dismissed):
@ButterHost69 I reviewed the linked-page deploy changes in `apps/daemon/src/deploy.ts`, `apps/daemon/src/server.ts`, the deploy request contracts, and the added daemon regression tests. I verified the changed-range fixes for nested path preservation, transitive linked-page discovery, root-relative links, provider-scoped `vercel.json`, and the new coverage around the index collision guard. Thanks for sticking with the iterations here — this looks solid now. 🙌


apps/daemon/src/server.ts
```diff
@@ -2596,6 +2597,2597 @@ export async function startServer({
...
+            // skip files that vanished mid-flight
+          }
+        }
+        /** @type {import('@open-design/contracts').UploadProjectFilesResponse} */
+        const body = { files: out };
+        res.json(body);
+      } catch (err) {
+        sendApiError(res, 500, 'INTERNAL_ERROR', 'upload failed');
+      }
   const resolvedPortRef = {
```
lefarcen, line 5191: **P1 — Unclosed route breaks build**

The `/api/projects/:id/upload` route is not closed before `const resolvedPortRef` begins. This makes `server.ts` syntactically invalid → CI fails.

Fix: close the route handler before the dependency wiring section.

apps/daemon/src/server.ts
```diff
@@ -2596,6 +2597,2597 @@ export async function startServer({
...
+    try {
+      /** @type {import('@open-design/contracts').ProjectDeploymentsResponse} */
+      const body = { deployments: publicDeployments(listDeployments(db, req.params.id)) };
+      res.json(body);
+    } catch (err) {
+      sendApiError(res, 400, 'BAD_REQUEST', String(err?.message || err));
+    }
+  });
+
+  app.post('/api/projects/:id/deploy', async (req, res) => {
```
lefarcen, line 3675: **P2 — Linked-page logic in duplicated inline block**

This reintroduces inline deploy routes even though the file later registers the modular deploy routes. The linked-page/provider handling should move into the canonical deploy route registrar instead of living here; otherwise fixing the P1 by removing the duplicate will drop the Cloudflare/provider-specific behavior again.

Fix: move the linked-page deploy request handling to the modular deploy route registrar (same place where the original deploy routes live).

lefarcen (changes requested):
Hey @ButterHost69, the new head broke the daemon build — see inline.

P1: `server.ts` has an unclosed upload route that makes the file syntactically invalid. CI is red because of this.

P2: The linked-page deploy logic was added to a duplicated inline route block instead of the canonical deploy route registrar, so fixing the P1 by removing the duplicate will drop the new behavior again.

PR #1156, opened 2026-05-10
fix(web): allow pod-to-chat comment text to wrap instead of truncating (#793)

YUHAO-corn:
Fixes #793

## Problem

When sending content from Pods to Chat, the resulting comment attachment chips had a hard `max-width: 280px` and `white-space: nowrap`, causing long comments to be silently clipped with ellipsis. Users could not read the full transferred content.

## Changes

- `index.css`: scope `.staged-comment` width expansion to `.user-attachment.staged-comment` only, add `max-height: 200px` with `overflow-y: auto` to prevent extreme height, and add `flex-wrap: wrap` to `.staged-name` so `elementId` and `comment` can wrap independently
- `index.css`: replace `white-space: nowrap` with `white-space: normal` and `word-break: break-word` on the comment text span
- `ChatPane.tsx`: add `title={\`\${a.elementId}: \${a.comment}\`}` tooltip on `.staged-name`, matching the existing pattern in `ChatComposer.tsx`

## Result

Long pod comments sent to chat now wrap across multiple lines within a bounded chip instead of being truncated, making the full transferred content readable.

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -1043,15 +1043,20 @@ code {
   object-fit: cover;
 }
 .staged-comment {
-  max-width: 280px;
   border-radius: var(--radius-pill);
   padding: 4px 6px 4px 10px;
+}
+.user-attachment.staged-comment {
+  max-width: 100%;
+  max-height: 200px;
+  overflow-y: auto;
 }
 .staged-comment .staged-name {
   display: inline-flex;
   align-items: center;
   gap: 6px;
   min-width: 0;
+  flex-wrap: wrap;
 }
 .staged-comment .staged-name strong {
   color: var(--accent-strong);
@@ -1059,9 +1064,8 @@ code {
   flex: 0 0 auto;
 }
 .staged-comment .staged-name span {
-  overflow: hidden;
-  text-overflow: ellipsis;
-  white-space: nowrap;
+  white-space: normal;
+  word-break: break-word;
 }
 .staged-comment button {
   width: 20px;
--- a/apps/web/src/components/ChatPane.tsx
+++ b/apps/web/src/components/ChatPane.tsx
@@ -870,7 +870,7 @@ function UserMessage({
         <div className="user-attachments comment-history-attachments">
           {commentAttachments.map((a) => (
             <span key={a.id} className="user-attachment staged-comment">
-              <span className="staged-name">
+              <span className="staged-name" title={`${a.elementId}: ${a.comment}`}>
                 <strong>{a.elementId}</strong>
                 <span>{a.comment}</span>
               </span>


mrcfps (approved):
@YUHAO-corn I verified the changed ChatPane and CSS ranges for pod-to-chat comment attachments. The updated styles now allow long comment text to wrap within the history chip while keeping the composer behavior scoped separately, and the added tooltip keeps the full attachment text accessible. Thanks for tightening up this chat readability fix.


apps/web/src/index.css
```diff
@@ -1043,25 +1043,29 @@ code {
...
+  max-width: 100%;
+  max-height: 200px;
+  overflow-y: auto;
+}
 .staged-comment .staged-name {
   display: inline-flex;
   align-items: center;
   gap: 6px;
   min-width: 0;
+  flex-wrap: wrap;
```
lefarcen, line 1059: **P2 — Cascade issue**: The wrap fix is incomplete because this element still picks up the shared `.staged-name { max-width: 140px }` (line 1002) and `.user-attachment .staged-name { overflow: hidden; white-space: nowrap; }` (line 1024) rules. The comment span can wrap inside its own box, but only inside a narrow 140px column, and long `elementId` values remain non-wrapping/clippable via `strong { flex: 0 0 auto; }` (line 1060).

**Suggested fix**: Add a more specific `.user-attachment.staged-comment .staged-name` override for `max-width`, `overflow`, and `white-space`, and decide whether long element IDs should wrap too.

lefarcen (changes requested):
Hi @YUHAO-corn! 👋

Thanks for tackling issue #793 — the direction is solid (scoping to `.user-attachment.staged-comment` to allow wrapping), but there's a CSS cascade issue that prevents the fix from fully working.

**The problem:** Even though you've removed the 280px constraint and set `white-space: normal` on the comment span, there are **upstream CSS rules** that still constrain the `.staged-name` container:

1. Line 1002: `.staged-name { max-width: 140px; }` — this narrows the whole chip, not just the comment text
2. Line 1024: `.user-attachment .staged-name { overflow: hidden; white-space: nowrap; }` — this re-applies nowrap at a higher specificity
3. Line 1060: `.staged-name strong { flex: 0 0 auto; }` — prevents `elementId` from wrapping

So even though your `.staged-comment .staged-name span` rule says `white-space: normal`, the parent `.staged-name` is still clamped to 140px and forced to `nowrap`, which means long comments will STILL clip.

**Suggested fix:**
Add a more specific override for `.user-attachment.staged-comment .staged-name` that resets those constraints:

```css
.user-attachment.staged-comment .staged-name {
  max-width: 100%;  /* override the 140px */
  overflow: visible; /* override hidden */
  white-space: normal; /* override nowrap */
}
```

And decide: should long `elementId` values wrap too, or stay on one line? If they should wrap, change `strong { flex: 0 0 auto; }` to allow wrapping.

P3 (non-blocking): The 200px max-height choice is reasonable, but a before/after screenshot or a short manual verification note would make the reasoning complete, especially for narrow/mobile widths.

Let me know when you've pushed an update — happy to re-review! 🚀

lefarcen (approved):
My apologies @YUHAO-corn — I misanalyzed the CSS cascade in my previous review.

After re-examining the specificity rules, your fix **does work correctly**:
- `.staged-comment .staged-name span` (specificity 0,0,3,0) with `white-space: normal` successfully overrides
- `.user-attachment .staged-name` (specificity 0,0,2,0) with `white-space: nowrap`

The comment text will wrap as intended. mrcfps's verification was correct. Approved! 🎉

lefarcen (approved):
My apologies @YUHAO-corn — I misanalyzed the CSS cascade in my previous review.

After re-examining the specificity rules, your fix **does work correctly**:
- `.staged-comment .staged-name span` (specificity 0,0,3,0) with `white-space: normal` successfully overrides
- `.user-attachment .staged-name` (specificity 0,0,2,0) with `white-space: nowrap`

The comment text will wrap as intended. mrcfps's verification was correct. Approved! 🎉

lefarcen (approved):
My apologies @YUHAO-corn — I misanalyzed the CSS cascade in my previous review.

After re-examining the specificity rules, your fix does work correctly: .staged-comment .staged-name span (specificity 0,0,3,0) with white-space: normal successfully overrides .user-attachment .staged-name (specificity 0,0,2,0) with white-space: nowrap.

The comment text will wrap as intended. mrcfps's verification was correct. Approved!

lefarcen (approved):
My apologies @YUHAO-corn — I misanalyzed the CSS cascade in my previous review.

After re-examining the specificity rules, your fix does work correctly: .staged-comment .staged-name span (specificity 0,0,3,0) with white-space: normal successfully overrides .user-attachment .staged-name (specificity 0,0,2,0) with white-space: nowrap.

The comment text will wrap as intended. mrcfps verification was correct. Approved!

lefarcen (approved):
Apologies @YUHAO-corn — I misanalyzed the CSS cascade. Your fix works correctly (specificity wins). Approved!

PR #530, opened 2026-05-05
fix: batch of small bug fixes (#283, #275, #390)

Nagendhra-web:
Three independent small fixes bundled into one PR. Each commit is scoped to its own issue and can be cherry-picked or split if a reviewer prefers.

### 1. `fix(web): add hover tooltips to Design Files action buttons (#283)` ([c657f3d](../commit/c657f3d))

The batch-download, select-all, and clear-selection buttons in `apps/web/src/components/DesignFilesPanel.tsx` had no `title` attribute, so users hovering them saw no tooltip. The other action buttons (refresh, new sketch, paste, upload) already had titles. Added titles to the three missing ones using the existing translation keys, so hover behavior is consistent across the panel.

Closes #283.

### 2. `docs: point pi-ai links to pi-mono packages (#275)` ([00f53d4](../commit/00f53d4))

The pi project moved from a standalone repo to the pi-mono monorepo. The old URL `https://github.com/mariozechner/pi-ai` now 404s. Replaced both shapes of reference:

- The reference-style `[piai]:` definition now points at `https://github.com/badlogic/pi-mono/tree/main/packages/ai` (the multi-provider LLM API package).
- Inline links whose visible text is the CLI tool `pi` or `Pi` now point at `https://github.com/badlogic/pi-mono/tree/main/packages/coding-agent` (the interactive coding-agent CLI), so a reader clicking `pi` in the daemon-discovery section lands on the actual binary's docs.

Affected: `README.md` and 10 translated READMEs, plus `docs/spec.md`, `docs/architecture.md`, `docs/references.md`, `docs/roadmap.md`. The two remaining `mariozechner/pi-ai` strings in `docs/references.md` and `docs/roadmap.md` are npm package names, not URLs, so they stay as-is.

Closes #275.

### 3. `fix(daemon): expand $HOME / ${HOME} in OD_DATA_DIR (#390)` ([4157ae0](../commit/4157ae0))

Some launchers (systemd unit files, NixOS modules, certain Docker entrypoints) pass `OD_DATA_DIR` with a literal `$HOME` or `${HOME}` because no shell ever expands them. `resolveDataDir` previously only handled `~/` shorthand, so `$HOME/.open-design` fell through to `path.resolve(PROJECT_ROOT, '$HOME/.open-design')` and produced paths like `/opt/open-design/$HOME/.open-design`.

[...]

PR #411, opened 2026-05-04
feat(daemon): allow OD_MEDIA_CONFIG_DIR to relocate media-config.json

GodTamIt:
Fixes #408 

`media-config.ts:configFile()` always wrote to `<projectRoot>/.od/`, which fails when the daemon runs from a read-only install (Nix store, immutable image, etc.) — `PUT /api/media/config` then 400s with `ENOENT` on the `mkdir`.

Add an opt-in `OD_MEDIA_CONFIG_DIR` env var that, when set, redirects the file to `<OD_MEDIA_CONFIG_DIR>/media-config.json`. Existing workspace-local installs are unaffected — the default still resolves to `<projectRoot>/.od/media-config.json`. The override is intended to be pointed at the same writable directory as `OD_DATA_DIR` (`~/.od`, `$XDG_CONFIG_HOME/open-design`, etc.) by the supervisor that launches the daemon.

Adds a test pinning the override path so the fallback semantics don't silently regress.

Issue #390, opened 2026-05-03
Set OD_DATA_DIR=$HOME/.open-design is not working


vistar:
Because server.ts:316 doing path.resolve(PROJECT_ROOT, process.env.OD_DATA_DIR) - relative path resolves from PROJECT_ROOT (/opt/open-design), not from $HOME. Path $HOME/.open-design will turn into /opt/open-design/$HOME/.open-design (or /opt/open-design/home/user/.open-design).

lefarcen:
Hi @vistar! 👋

Thanks for the detailed report — you nailed the root cause. I just confirmed the bug in `apps/daemon/src/server.ts:316-317`:

```typescript
const RUNTIME_DATA_DIR = process.env.OD_DATA_DIR
  ? path.resolve(PROJECT_ROOT, process.env.OD_DATA_DIR)
  : path.join(PROJECT_ROOT, '.od');
```

When `OD_DATA_DIR=$HOME/.open-design`, `path.resolve()` treats the second argument as relative (since it doesn't start with `/`), so it becomes:
```
/opt/open-design/$HOME/.open-design  // broken
```

**Fix**: Should be `path.resolve(process.env.OD_DATA_DIR)` (without `PROJECT_ROOT` as base), or expand `$HOME` first:
```typescript
const RUNTIME_DATA_DIR = process.env.OD_DATA_DIR
  ? path.resolve(process.env.OD_DATA_DIR.replace(/^\$HOME/, os.homedir()))
  : path.join(PROJECT_ROOT, '.od');
```

Would you like to open a PR for this? If not, I can flag it for the maintainers.

Thanks for making open-design better!
— open-design team

lefarcen:
@vistar The fix you reported in this issue should be in PR #391, just merged. Please `git pull && pnpm install && pnpm dev:all` (or your equivalent) and try again. If you still hit this, leave a reply here and we'll keep digging. Thanks for the report!

lefarcen:
Fix tracking: #530 expands `resolveDataDir` to handle literal `/Users/a1` and `${HOME}` in `OD_DATA_DIR` (from systemd units, NixOS modules, certain Docker entrypoints). Previously only `~/` shorthand was expanded, so `$HOME/.open-design` fell through to relative path resolution and produced paths like `/opt/open-design/$HOME/.open-design`.

[...]

PR #1159, opened 2026-05-10
fix: set writable OD_DATA_DIR default for nix run

### LINKED ISSUE #1157: won't run on nixos
```zsh
❯ nix run github:nexu-io/open-design
node:fs:1349
  const result = binding.mkdir(
                         ^

Error: ENOENT: no such file or directory, mkdir '/nix/store/kqh5bpv1mmw7qlwpi95c61dlrm1rdx27-open-design-daemon-0.6.0/lib/open-design/.od/projects'
    at Object.mkdirSync (node:fs:1349:26)
    at file:///nix/store/kqh5bpv1mmw7qlwpi95c61dlrm1rdx27-open-design-daemon-0.6.0/lib/open-design/apps/daemon/dist/server.js:651:4
    at ModuleJob.run (node:internal/modules/esm/module_job:430:25)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:661:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:101:5) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'mkdir',
  path: '/nix/store/kqh5bpv1mmw7qlwpi95c61dlrm1rdx27-open-design-daemon-0.6.0/lib/open-design/.od/projects'
}

Node.js v24.14.1
```

xxiaoxiong:
Fixes #1157

## Problem

When running via `nix run github:nexu-io/open-design`, the daemon attempted to create runtime state under the Nix store package path:

```
/nix/store/.../lib/open-design/.od/projects
```

The Nix store is read-only at runtime, causing startup to fail with `ENOENT` when `mkdir()` tried to create the projects directory.

## Solution

Update the `nix run` wrapper to export `OD_DATA_DIR` with a writable default (`$HOME/.od`) when the variable is unset.

**Changes:**
- Modified `flake.nix` `apps.default` wrapper script
- Added `export OD_DATA_DIR="${OD_DATA_DIR:-$HOME/.od}"` before executing `od`
- Added explanatory comment documenting the Nix store read-only constraint

## Behavior

- **Default**: `nix run` now uses `$HOME/.od` for runtime state
- **Override**: Users can still set `OD_DATA_DIR` before running: `OD_DATA_DIR=/custom/path nix run ...`
- **Modules**: Home Manager and NixOS modules already set `OD_DATA_DIR`, so they are unaffected

## Testing

Verified the fix resolves the reported error. The daemon now starts successfully under `nix run` without requiring manual `OD_DATA_DIR` configuration.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -62,9 +62,14 @@
       # Wrap `od` with `--no-open` for `nix run`: the daemon package
       # builds the daemon workspace only, not `apps/web/out/`, so the
       # browser would otherwise auto-open onto an empty static dir.
+      #
+      # Set OD_DATA_DIR to a writable location when unset. The Nix store
+      # is read-only at runtime, so the daemon cannot write to its default
+      # `<projectRoot>/.od` location under `nix run`.
       apps.default = {
         type = "app";
         program = "${pkgs.writeShellScript "od-nix-run" ''
+          export OD_DATA_DIR="''${OD_DATA_DIR:-$HOME/.od}"
           exec ${daemon}/bin/od --no-open "$@"
         ''}";
         meta.description = "Open Design local daemon (`od`)";


xxiaoxiong:
This is a minimal one-line fix that adds `export OD_DATA_DIR="${OD_DATA_DIR:-$HOME/.od}"` to the nix run wrapper script.

The fix:
- ✅ Provides a writable default location for runtime state
- ✅ Preserves user overrides (respects existing `OD_DATA_DIR`)
- ✅ Does not affect Home Manager or NixOS modules (they already set `OD_DATA_DIR`)
- ✅ Follows the same pattern as the existing workaround mentioned in #1157

Ready for review!

lefarcen:
Agent-to-agent related research for PR #1159:

Linked issue:
- #1157: high confidence. The issue exists and reports plain `nix run github:nexu-io/open-design` failing daemon startup with `ENOENT` while creating `.od/projects` under read-only `/nix/store/.../lib/open-design`, which matches this PR's `OD_DATA_DIR=/Users/a1/.od` default fix.

Related context checked:
- #402: high confidence related PR. Introduced official Nix/flake support and added `flake.nix`.
- #411: medium confidence related PR. Same immutable/Nix-store runtime-write class, but for `OD_MEDIA_CONFIG_DIR` rather than daemon project data.
- #530 / #390: medium confidence related data-dir context. Fixed `/Users/a1`/`/Users/a1` expansion in `OD_DATA_DIR`, a separate env resolution bug.

Search terms used:
- `nix OD_DATA_DIR ENOENT`
- `read-only daemon startup`
- `flake.nix`
- `nixos`
- `390`

Checked sources:
- `gh issue view/search`
- `gh pr view/search`
- Synclo memory search was unavailable in this workspace.

Suggested issue status pointer for #1157:
- PR #1159 is open to fix this by setting a writable default `OD_DATA_DIR=/Users/a1/.od` for plain `nix run` while preserving explicit overrides.

flake.nix
```diff
@@ -62,9 +62,14 @@
...
       # builds the daemon workspace only, not `apps/web/out/`, so the
       # browser would otherwise auto-open onto an empty static dir.
+      #
+      # Set OD_DATA_DIR to a writable location when unset. The Nix store
+      # is read-only at runtime, so the daemon cannot write to its default
+      # `<projectRoot>/.od` location under `nix run`.
       apps.default = {
         type = "app";
         program = "${pkgs.writeShellScript "od-nix-run" ''
+          export OD_DATA_DIR="''${OD_DATA_DIR:-$HOME/.od}"
```
lefarcen, line 72: **P2** — The fix only applies to `apps.default`, so `nix run github:nexu-io/open-design#daemon` or `nix shell github:nexu-io/open-design#daemon -c od` still execute `${daemon}/bin/od` from `nix/package-daemon.nix` without `OD_DATA_DIR` and can hit the same read-only `/nix/store/.../.od/projects` failure.

Move the defaulting into the package wrapper itself, e.g. `makeWrapper ... --run 'export OD_DATA_DIR="${OD_DATA_DIR:-$HOME/.od}"'` in `nix/package-daemon.nix`, or otherwise share one wrapped runtime entrypoint between `packages.daemon` and `apps.default`. Home Manager/NixOS modules should remain unaffected because they already set `OD_DATA_DIR`.

lefarcen, line 72: **P3** — The shell expansion is safe for normal user shells, but if `HOME` is unset it resolves to `/.od`, producing a confusing root-level write attempt. If `HOME` is malformed or relative, the daemon may resolve it under the packaged project root.

Add a small guard before exporting, such as failing clearly when `HOME` is empty/non-absolute, or use a Nix/runtime fallback with an explicit error message.

lefarcen (commented):
Hi @xxiaoxiong! Thanks for the fix — the `OD_DATA_DIR` defaulting addresses the immediate `nix run` failure. Found a couple of scoping/edge-case concerns that should be tightened before merge:

**Summary:**
- ✅ The core fix (setting writable default for `nix run`) is sound
- ⚠️ P2: Same issue still hits other Nix entry points (see inline)
- ℹ️ P3: Edge case handling for empty/malformed `HOME` + reasoning completeness

See inline comments below.

mrcfps (approved):
@xxiaoxiong I reviewed the `flake.nix` wrapper change and verified the new `OD_DATA_DIR` default is applied only when unset, preserves explicit overrides, and addresses the reported `nix run` write-into-store failure without changing the Home Manager/NixOS paths. Thanks for landing a focused fix here — this is a clean improvement.

PR #1160, opened 2026-05-10
fix(pack): harden mac portable runtime bundling

wegelcloud:
## Summary\n- preserve the local model picker additions on top of current main\n- clear macOS metadata before ad-hoc codesigning portable mac bundles\n- bundle Homebrew-style libnode dylibs alongside the packaged Node runtime\n\n## Validation\n- pnpm guard\n- pnpm --filter @open-design/daemon typecheck\n- pnpm --filter @open-design/web typecheck\n- pnpm --filter @open-design/tools-pack build\n- pnpm --filter @open-design/tools-pack test\n- pnpm tools-pack mac build --to app --portable --dir ~/open-design-tools-pack-build\n- codesign --verify --deep --strict --verbose=2 /Applications/Open Design.app\n- packaged desktop/daemon/web IPC status + /api/health\n\n## Data preservation\n- did not modify ~/Library/Application Support/Open Design or repo .od data\n- verified existing default namespace SQLite has 4 projects after update

### DIFF
--- a/apps/daemon/src/agents.ts
+++ b/apps/daemon/src/agents.ts
@@ -277,6 +277,8 @@ export const AGENT_DEFS = [
       { id: 'sonnet', label: 'Sonnet (alias)' },
       { id: 'opus', label: 'Opus (alias)' },
       { id: 'haiku', label: 'Haiku (alias)' },
+      { id: 'claude-opus-4-7', label: 'claude-opus-4-7' },
+      { id: 'claude-sonnet-4-6', label: 'claude-sonnet-4-6' },
       { id: 'claude-opus-4-5', label: 'claude-opus-4-5' },
       { id: 'claude-sonnet-4-5', label: 'claude-sonnet-4-5' },
       { id: 'claude-haiku-4-5', label: 'claude-haiku-4-5' },
@@ -444,6 +446,7 @@ export const AGENT_DEFS = [
       DEFAULT_MODEL_OPTION,
       // Gemini 3 (May 2026): top-tier reasoning + fast frontier-class.
       // Both currently ship as previews via the Gemini CLI. Issue #981.
+      { id: 'gemini-3.1-pro-preview', label: 'gemini-3.1-pro-preview' },
       { id: 'gemini-3-pro-preview', label: 'gemini-3-pro-preview' },
       { id: 'gemini-3-flash-preview', label: 'gemini-3-flash-preview' },
       { id: 'gemini-2.5-pro', label: 'gemini-2.5-pro' },
--- a/tools/pack/src/mac/app.ts
+++ b/tools/pack/src/mac/app.ts
@@ -120,6 +120,26 @@ async function buildPrebundledStandaloneRuntime(
   });
 }
 
+async function copyBundledNodeRuntime(paths: MacPaths): Promise<void> {
+  const bundledNodePath = join(paths.resourceRoot, "bin", "node");
+  await mkdir(join(paths.resourceRoot, "bin"), { recursive: true });
+  await cp(process.execPath, bundledNodePath);
+  await chmod(bundledNodePath, 0o755);
+
+  const hostNodeLibRoot = join(dirname(process.execPath), "..", "lib");
+  const hostNodeLibEntries = await readdir(hostNodeLibRoot).catch(() => []);
+  const libnodeEntries = hostNodeLibEntries.filter((entry) =>
+    /^libnode(?:\.\d+)?\.dylib$/.test(entry)
+  );
+  if (libnodeEntries.length === 0) return;
+
+  const bundledLibRoot = join(paths.resourceRoot, "lib");
+  await mkdir(bundledLibRoot, { recursive: true });
+  for (const entry of libnodeEntries) {
+    await cp(join(hostNodeLibRoot, entry), join(bundledLibRoot, entry));
+  }
+}
+
 export async function copyResourceTree(config: ToolPackConfig, paths: MacPaths): Promise<void> {
   await rm(paths.resourceRoot, { force: true, recursive: true });
   await mkdir(paths.resourceRoot, { recursive: true });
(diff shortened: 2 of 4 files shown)


tools/pack/src/mac/app.ts
```diff
@@ -120,6 +120,26 @@ async function buildPrebundledStandaloneRuntime(
   });
 }
 
+async function copyBundledNodeRuntime(paths: MacPaths): Promise<void> {
+  const bundledNodePath = join(paths.resourceRoot, "bin", "node");
+  await mkdir(join(paths.resourceRoot, "bin"), { recursive: true });
+  await cp(process.execPath, bundledNodePath);
+  await chmod(bundledNodePath, 0o755);
+
+  const hostNodeLibRoot = join(dirname(process.execPath), "..", "lib");
```
lefarcen, line 129: **P1 — Incomplete portable runtime bundling**

Copying only `libnode*.dylib` is not enough to make a Homebrew-built Node runtime portable. Homebrew's `node` binary commonly has additional direct absolute dependencies under `/opt/homebrew/opt/...` such as `libuv`, `llhttp`, `openssl`, ICU, etc.; those will still resolve only on machines with matching Homebrew libraries installed.

Either:
- Reject dynamically linked Homebrew Node as a build input, OR
- Copy and relocate the full `otool -L` dependency closure with the required `install_name_tool`/rpath updates, plus a smoke test that runs the built app on an environment without Homebrew.

apps/daemon/src/agents.ts
```diff
@@ -444,6 +446,7 @@ export const AGENT_DEFS = [
       DEFAULT_MODEL_OPTION,
       // Gemini 3 (May 2026): top-tier reasoning + fast frontier-class.
       // Both currently ship as previews via the Gemini CLI. Issue #981.
+      { id: 'gemini-3.1-pro-preview', label: 'gemini-3.1-pro-preview' },
```
lefarcen, line 449: **P2 — Test coverage gap**

This changes the pinned Gemini fallback list but does not update the existing `apps/daemon/tests/agents.test.ts` assertion that checks the exact Gemini picker order. That test currently expects `['default', 'gemini-3-pro-preview', ...]`, so the daemon test suite will fail once this entry is added unless the test is updated alongside the behavior.

apps/daemon/src/agents.ts
```diff
@@ -444,6 +446,7 @@ export const AGENT_DEFS = [
...
       { id: 'gemini-3-pro-preview', label: 'gemini-3-pro-preview' },
```
lefarcen, line 450: **P2 — Stale model in picker**

The PR rationale says `gemini-3-pro-preview` is EOL and adds `gemini-3.1-pro-preview`, but the picker still offers the older `gemini-3-pro-preview` immediately after it, and the nearby comment still claims the previews currently ship. That leaves a known-failing model in the curated suggestions without warning.

Remove the deprecated ID or explicitly document why it remains selectable.

lefarcen (changes requested):
Hi @wegelcloud, thanks for the PR! The packaging hardening direction is useful, but there are correctness gaps that need addressing before this is ready.

**Key concerns:**
1. The Homebrew Node portability fix is incomplete — copying only libnode leaves other dylib dependencies unresolved
2. Model picker changes introduce test failures and leave deprecated entries without warnings

See inline comments for specifics. Please address these and I'll re-review.

tools/pack/src/mac/app.ts
```diff
@@ -120,6 +120,26 @@ async function buildPrebundledStandaloneRuntime(
...
+  const hostNodeLibEntries = await readdir(hostNodeLibRoot).catch(() => []);
+  const libnodeEntries = hostNodeLibEntries.filter((entry) =>
+    /^libnode(?:\.\d+)?\.dylib$/.test(entry)
+  );
+  if (libnodeEntries.length === 0) return;
+
+  const bundledLibRoot = join(paths.resourceRoot, "lib");
+  await mkdir(bundledLibRoot, { recursive: true });
+  for (const entry of libnodeEntries) {
+    await cp(join(hostNodeLibRoot, entry), join(bundledLibRoot, entry));
```
mrcfps, line 139: `copyBundledNodeRuntime` still leaves the packaged Node runtime tied to the build machine. This helper only copies `process.execPath` plus files matching `libnode*.dylib`, but it never inspects the full `otool -L` dependency closure or rewrites install names / rpaths. A Homebrew Node binary commonly has additional absolute deps under `/opt/homebrew/...` (for example `libuv`, OpenSSL, ICU, etc.), and plain `cp()` also preserves symlinks by default, so the copied `libnode*.dylib` entries can still point back into the Cellar instead of becoming real bundled files. In that state the portable app can still fail to launch on a clean machine, which is exactly the main path this change is trying to harden. Please either reject dynamically linked Homebrew Node inputs, or bundle the full dylib closure with `dereference: true` plus the required `install_name_tool` / rpath rewrites, and cover it with a smoke test that runs the portable app without Homebrew installed.

apps/daemon/src/agents.ts
mrcfps, line 449: Adding `gemini-3.1-pro-preview` here breaks the pinned Gemini picker test in `apps/daemon/tests/agents.test.ts:713-726`. That test still asserts the exact fallback order `['default', 'gemini-3-pro-preview', 'gemini-3-flash-preview', ...]`, so the daemon test suite fails as soon as this entry lands. Because this PR changes shipped picker contents, the matching test update needs to be part of the same patch (and ideally keep the new order explicit) so CI continues to protect the curated list.

mrcfps (changes requested):
@wegelcloud Thanks for the update here — I found two blockers on the current head that still need follow-up before this is ready: the mac portable bundling change is not self-contained yet, and the new Gemini picker entry is missing its pinned test update.


PerishCode:
@wegelcloud — I did another maintainer pass over this one today. The PR is still carrying unresolved requested changes from lefarcen and mrcfps, and it is now also out of date with `main`.

The part that makes this hard to take over as a maintainer edit is the scope mix: the branch combines model-picker updates with mac portable runtime bundling changes, and the remaining review feedback is about packaging/runtime correctness rather than a small mechanical fix.

If you want to keep driving it, the useful next step is to rebase and narrow the branch to one lane:

- either a small model-picker refresh with the stale Gemini entry removed and the daemon test updated;
- or a packaging-only follow-up that addresses the Homebrew Node portability feedback with validation.

If this is no longer active, say so and we can close it cleanly instead of leaving the stale mixed branch in the queue.

lefarcen:
Heads-up: PR #2160 is now open against this same mac packaging area. Both PRs touch `tools/pack/src/mac/app.ts` and address #1275's Homebrew Node / missing `libnode.*.dylib` failure, but they take different directions: #2160 removes the bundled Node launcher and uses the Electron-as-Node fallback, while this PR keeps the launcher path and bundles `libnode` alongside it.

@wegelcloud and @bulai0408 may want to compare approaches; the maintainer team will decide which direction lands. Sharing so neither effort gets wasted.

PR #1169, opened 2026-05-10
fix(runtime): auto-annotate imported HTML elements for Tweaks selection (#892)

### LINKED ISSUE #892: Fix imported design packages losing Picker and Pods selection in Tweaks mode
## Summary

After importing a local design package, enabling Tweaks mode in Design Files makes Picker and Pods unusable because components can no longer be selected.

In the reported case, the imported design renders normally, but once Tweaks mode is enabled, Picker and Pods appear available in the toolbar while actual component selection no longer works.

## Current behavior

- A local design package is imported successfully.
- The imported design file opens in Design Files.
- Tweaks mode is enabled.
- Picker and Pods remain visible in the toolbar.
- However, components cannot be selected.
- As a result, Picker and Pods become effectively unusable in this state.

## Expected behavior

Imported design packages should continue to support component selection when Tweaks mode is enabled, or the UI should clearly prevent unsupported interactions.

Suggested improvements:

- Ensure Tweaks mode does not break component hit-testing for imported design packages.
- Keep Picker and Pods selection behavior consistent between locally created designs and imported design packages.
- If imported packages have limitations in Tweaks mode, surface a clear message instead of leaving the tools visible but non-functional.
- Verify that component overlay, selection bridge, and click handling remain active after import.

## Steps to reproduce

1. Import a local design package.
2. Open the imported design file in Design Files.
3. Enable Tweaks mode.
4. Switch to Picker or Pods.
5. Try selecting components.
6. Observe that components cannot be selected.

## Notes

- an imported local design package open in Design Files
- Tweaks mode enabled
- Picker / Pods visible in the toolbar even though component selection is not working

## Why this matters

This is a workflow blocker for imported content. If Tweaks mode disables Picker and Pods selection only for imported design packages, users cannot inspect or annotate imported designs the same way they can with native ones, which makes the import flow feel incomplete and inconsistent.

Image

Image

YUHAO-corn:
Fixes #892

## Problem

Designs imported from external sources do not always carry stable Open Design target annotations. Without a usable `data-od-id` / screen label target, Picker, Pods, and Tweaks selection can silently fail on common imported HTML.

## Changes

- Add fallback `data-od-id` annotation for imported markup that lacks existing Open Design target attributes.
- Run the fallback annotation consistently across render modes so saved Inspect/Tweaks selectors keep matching after users leave Inspect mode.
- Keep manual-edit source-path markers intact by allowing source-path annotation even when fallback `data-od-id` values are present.
- Expand selector coverage for common imported HTML shapes such as headings, buttons, links, and direct-child div wrappers while preserving existing annotations.
- Add regression coverage for imported annotation, selector persistence, and manual-edit source mapping.

## Result

Imported designs can be selected in Picker / Pods / Tweaks more reliably, and saved tweaks no longer disappear when the preview switches out of the bridge mode that first created the fallback ids.

## Surface area
- [x] UI
- [ ] Keyboard shortcut
- [ ] CLI
- [ ] env var
- [ ] API
- [x] Default behavior change
- [ ] i18n keys
- [ ] Extension point
- [ ] None

## Validation
- GitHub CI on the rebased head is green: `build` and `Validate workspace` passed.
- Reviewer re-check confirmed the persistence gap, selector coverage, and manual-edit source-path pipeline issue are addressed on the current head.

### DIFF
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -254,12 +254,29 @@ function annotateMissingOdIds(doc: string): string {
   if (typeof DOMParser === 'undefined') return doc;
   try {
     const parsed = new DOMParser().parseFromString(doc, 'text/html');
-    const selector = 'section, article, header, footer, nav, main, aside, h1, h2, h3, h4, h5, h6, button, a, [id], div[class], div[id]';
+    // Only target divs that are direct children of semantic containers or body;
+    // deeply nested layout divs (e.g. flex/grid wrappers) create noise in the
+    // selection bridge without adding meaningful pickable targets.
+    const selector = [
+      'section', 'article', 'header', 'footer', 'nav', 'main', 'aside',
+      'h1', 'h2', 'h3', 'h4', 'h5', 'h6',
+      'button', 'a', '[id]',
+      'body > div[class]', 'body > div[id]',
+      'section > div[class]', 'section > div[id]',
+      'article > div[class]', 'article > div[id]',
+      'main > div[class]', 'main > div[id]',
+      'header > div[class]', 'header > div[id]',
+      'footer > div[class]', 'footer > div[id]',
+      'nav > div[class]', 'nav > div[id]',
+      'aside > div[class]', 'aside > div[id]',
+      '[id] > div[class]', '[id] > div[id]',
+    ].join(', ');
+    const skipTags = new Set(['script', 'style', 'template', 'noscript', 'iframe', 'object', 'embed']);
     let fallbackIndex = 0;
     parsed.body.querySelectorAll(selector).forEach((el) => {
       if (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label')) return;
       const tag = el.tagName.toLowerCase();
-      if (tag === 'script' || tag === 'style' || tag === 'template' || tag === 'noscript') return;
+      if (skipTags.has(tag)) return;
       const path = sourcePathForElement(el);
       el.setAttribute('data-od-id', path || `od-${tag}-${fallbackIndex++}`);
     });
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -48,7 +48,9 @@ export function buildSrcdoc(
   </head>
   <body>${html}</body>
 </html>`;
-  const withSourcePaths = options.editBridge ? annotateManualEditSourcePaths(wrapped) : wrapped;
+  const needsOdIds = options.commentBridge || options.inspectBridge;
+  const withOdIds = needsOdIds ? annotateMissingOdIds(wrapped) : wrapped;
+  const withSourcePaths = options.editBridge ? annotateManualEditSourcePaths(withOdIds) : withOdIds;
   const withBase = options.baseHref ? injectBaseHref(withSourcePaths, options.baseHref) : withSourcePaths;
   const withShim = injectSandboxShim(withBase);
   const withDeck = options.deck ? injectDeckBridge(withShim, options.initialSlideIndex) : withShim;
@@ -240,6 +242,32 @@ function sourcePathForElement(el: Element): string {
 function serializeHtmlDocument(doc: Document): string {
   const doctype = doc.doctype ? '<!doctype html>\n' : '';
   return `${doctype}${doc.documentElement.outerHTML}`;
+}
+
+/**
+ * Auto-annotate structural HTML elements that lack `data-od-id` or
+ * `data-screen-label` so that the selection bridge (Picker / Pods /
+ * Tweaks) can target them. This fixes imported designs whose HTML was
+ * generated outside of Open Design and therefore carries no OD-specific
+ * annotations.
+ */
+function annotateMissingOdIds(doc: string): string {
+  if (typeof DOMParser === 'undefined') return doc;
+  try {
+    const parsed = new DOMParser().parseFromString(doc, 'text/html');
+    const selector = 'section, article, header, footer, nav, main, aside, [id]';
+    let fallbackIndex = 0;
+    parsed.body.querySelectorAll(selector).forEach((el) => {
+      if (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label')) return;
+      const tag = el.tagName.toLowerCase();
+      if (tag === 'script' || tag === 'style' || tag === 'template' || tag === 'noscript') return;
+      const path = sourcePathForElement(el);
+      el.setAttribute('data-od-id', path || `od-${tag}-${fallbackIndex++}`);
+    });
+    return serializeHtmlDocument(parsed);
+  } catch {
+    return doc;
+  }
 }
 
 function injectManualEditBridge(doc: string): string {
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -48,8 +48,7 @@ export function buildSrcdoc(
   </head>
   <body>${html}</body>
 </html>`;
-  const needsOdIds = options.commentBridge || options.inspectBridge;
-  const withOdIds = needsOdIds ? annotateMissingOdIds(wrapped) : wrapped;
+  const withOdIds = annotateMissingOdIds(wrapped);
   const withSourcePaths = options.editBridge ? annotateManualEditSourcePaths(withOdIds) : withOdIds;
   const withBase = options.baseHref ? injectBaseHref(withSourcePaths, options.baseHref) : withSourcePaths;
   const withShim = injectSandboxShim(withBase);
@@ -255,7 +254,7 @@ function annotateMissingOdIds(doc: string): string {
   if (typeof DOMParser === 'undefined') return doc;
   try {
     const parsed = new DOMParser().parseFromString(doc, 'text/html');
-    const selector = 'section, article, header, footer, nav, main, aside, [id]';
+    const selector = 'section, article, header, footer, nav, main, aside, h1, h2, h3, h4, h5, h6, button, a, [id], div[class], div[id]';
     let fallbackIndex = 0;
     parsed.body.querySelectorAll(selector).forEach((el) => {
       if (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label')) return;
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -216,7 +216,7 @@ function annotateManualEditSourcePaths(doc: string): string {
   try {
     const parsed = new DOMParser().parseFromString(doc, 'text/html');
     parsed.body.querySelectorAll(MANUAL_EDIT_DISCOVERY_SELECTOR).forEach((el) => {
-      if (el.hasAttribute('data-od-id')) return;
+      if (el.hasAttribute(MANUAL_EDIT_SOURCE_PATH_ATTR)) return;
       const path = sourcePathForElement(el);
       if (path) el.setAttribute(MANUAL_EDIT_SOURCE_PATH_ATTR, path);
     });
--- a/apps/web/tests/runtime/srcdoc.test.ts
+++ b/apps/web/tests/runtime/srcdoc.test.ts
@@ -254,39 +254,79 @@ describe('buildSrcdoc', () => {
     expect(srcdoc).not.toContain('<div data-od-id=');
   });
 
-  it('annotates direct-child divs under semantic containers and skips iframe/object/embed', () => {
+  it('auto-annotates direct-child divs with class or id under semantic containers', () => {
     const dom = new JSDOM('');
     globalThis.DOMParser = dom.window.DOMParser;
     const srcdoc = buildSrcdoc(
-      '<main><div class="card">A</div><iframe src="x"></iframe></main>',
-      { commentBridge: true },
+      '<section><div class="wrapper">Wrapper</div><div id="named">Named</div></section>',
+      {},
     );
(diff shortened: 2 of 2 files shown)


apps/web/src/runtime/srcdoc.ts
```diff
@@ -46,7 +46,9 @@ export function buildSrcdoc(
   </head>
   <body>${html}</body>
 </html>`;
-  const withSourcePaths = options.editBridge ? annotateManualEditSourcePaths(wrapped) : wrapped;
+  const needsOdIds = options.commentBridge || options.inspectBridge;
+  const withOdIds = needsOdIds ? annotateMissingOdIds(wrapped) : wrapped;
```
mrcfps: Blocking: this path only injects fallback `data-od-id`s into the transient iframe document, but `saveInspectToSource()` persists inspect overrides back into the artifact as selectors such as `[data-od-id="path-0"]` (`apps/web/src/components/FileViewer.tsx`). After the user leaves inspect/comment mode, `buildSrcdoc()` stops calling `annotateMissingOdIds()`, so the saved CSS no longer matches the source HTML and the tweak disappears from normal preview/export. Please either persist the synthesized IDs into the saved source before serializing overrides, or make the non-inspect render/export path apply the same deterministic annotation whenever persisted inspect overrides reference these fallback IDs. A regression test that saves an override on unannotated imported HTML and then reloads the plain preview would catch this.

mrcfps (changes requested):
@YUHAO-corn Thanks for tackling the imported-selection gap 🙏 I found one blocking regression here: the new fallback `data-od-id`s only exist in the inspect/comment `srcdoc`, so a tweak saved from an imported artifact can look successful in-session but disappear again in normal preview/export once those bridges are off.


lefarcen:
Hi @YUHAO-corn! 🎉
Thanks for the contribution — this tackles a real pain point where imported designs lose selection targets in Tweaks mode.

— open-design team

apps/web/src/runtime/srcdoc.ts
```diff
@@ -99,6 +101,32 @@ function serializeHtmlDocument(doc: Document): string {
...
+ * `data-screen-label` so that the selection bridge (Picker / Pods /
+ * Tweaks) can target them. This fixes imported designs whose HTML was
+ * generated outside of Open Design and therefore carries no OD-specific
+ * annotations.
+ */
+function annotateMissingOdIds(doc: string): string {
+  if (typeof DOMParser === 'undefined') return doc;
+  try {
+    const parsed = new DOMParser().parseFromString(doc, 'text/html');
+    const selector = 'section, article, header, footer, nav, main, aside, [id]';
```
lefarcen: **P1** — This only annotates `section, article, header, footer, nav, main, aside, [id]`, so common imported HTML made from anonymous `<div>` wrappers/cards/headings/buttons still gets zero targets. For example, `<div class="page"><div class="card">...</div></div>` matches none of this selector, so the bridge's `allTargets()` still broadcasts an empty list and Picker/Pods remain unusable.

Please add a fallback that annotates at least top-level visible body children, or use a broader target set aligned with the manual-edit discovery selector.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -99,6 +101,32 @@ function serializeHtmlDocument(doc: Document): string {
...
+    let fallbackIndex = 0;
+    parsed.body.querySelectorAll(selector).forEach((el) => {
+      if (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label')) return;
+      const tag = el.tagName.toLowerCase();
+      if (tag === 'script' || tag === 'style' || tag === 'template' || tag === 'noscript') return;
+      const path = sourcePathForElement(el);
+      el.setAttribute('data-od-id', path || `od-${tag}-${fallbackIndex++}`);
```
lefarcen, line 281: **P2** — Generated `path-*` IDs are not checked against existing `data-od-id` values, so mixed documents can get duplicate element IDs. If an existing annotated element already uses `data-od-id="path-0"` and an earlier unannotated structural element is assigned `path-0`, selectors, comments, and inspect overrides become ambiguous because they key by `elementId`.

Please collect existing IDs and choose a non-colliding generated value.

apps/web/tests/runtime/srcdoc.test.ts
```diff
@@ -189,4 +189,40 @@ describe('buildSrcdoc', () => {
     expect(srcdoc).not.toContain('<script data-od-source-path=');
     expect(srcdoc.indexOf('data-od-source-path="path-0"')).toBeLessThan(srcdoc.indexOf('document.body.prepend'));
   });
+
+  // Regression for nexu-io/open-design#892: imported designs (e.g. Claude
+  // Design ZIP) may not carry data-od-id annotations. The selection bridge
+  // depends on these attributes to identify clickable targets, so we
+  // auto-annotate structural elements when they are missing.
+  it('auto-annotates imported HTML that lacks data-od-id or data-screen-label', () => {
```
lefarcen, line 225: **P2** — The regression test only proves the happy path for semantic `section`/`article` markup, so it would pass even though div-only imported HTML still has no targets.

Add a test for a realistic unannotated import using nested `<div>`/heading/button markup and assert at least one selectable target is emitted.

lefarcen (changes requested):
Hey @YUHAO-corn! 👋 Thanks for tackling #892 — the annotation approach is solid, but I found a P1 blocker around selector coverage for common imported HTML shapes. See inline comments below.

**Code review (Lens A):**

The change follows existing `DOMParser`/serialization patterns, with no obvious XSS or type issues. However, the selector is too narrow to reliably fix #892 for real imported markup.

**Inline findings:** 1 P1 (must-fix), 2 P2 (should-fix), 1 P2 reasoning gap.

Please address the P1 selector coverage issue — common div-based imports would still fail. Happy to iterate once that's resolved!

lefarcen (changes requested):
Hey @YUHAO-corn! 👋 Re-reviewing after the latest push.

The d20e1f2c commit only updates a test assertion — the `srcdoc.ts` logic is unchanged from the previous head. **Two P1 blockers remain unresolved:**

**P1 (from mrcfps) — Persistence regression:** The synthesized `data-od-id` values only exist in the transient srcdoc when `commentBridge`/`inspectBridge` is enabled. When a user saves an Inspect tweak, `saveInspectToSource()` persists CSS selectors like `[data-od-id="path-0"]` back into the artifact. After leaving Inspect mode, `buildSrcdoc()` stops calling `annotateMissingOdIds()`, so the saved CSS no longer matches the source HTML — the tweak silently disappears from normal preview/export. This makes the feature unsafe for production use.

**P1 (original) — Selector too narrow:** The selector `section, article, header, footer, nav, main, aside, [id]` misses common imported HTML made from anonymous `<div>` wrappers/cards/headings/buttons. For example, `<div class="page"><div class="card">...</div></div>` matches none of this, so Picker/Pods remain unusable.

Please address both blockers before this can move forward. Happy to iterate once fixed!

apps/web/src/runtime/srcdoc.ts
mrcfps: `buildSrcdoc()` now adds fallback `data-od-id`s only when `commentBridge || inspectBridge`, but `saveInspectToSource()` persists inspect overrides back into the artifact source as selectors like `[data-od-id="…"]` (`apps/web/src/components/FileViewer.tsx:3774-3803`). That means an imported file with no source-authored annotations can look editable in Inspect, save successfully, and then lose the tweak as soon as the app reloads plain preview/export, because those runtime-only attributes are no longer present on the saved HTML. The unchanged contract note in `apps/web/tests/runtime/srcdoc-bridge-empty-targets.test.ts:19-24` points at the same persistence hazard. Please either persist the synthesized IDs into the saved source before serializing overrides, or apply the same deterministic annotation on non-inspect renders whenever saved overrides reference these fallback IDs, and add a regression that saves an override on unannotated imported HTML then reloads normal preview/export.

mrcfps (changes requested):
@YUHAO-corn Thanks for iterating on the imported-selection fix 🙏 I still see one blocker on the latest head: the new fallback `data-od-id`s are only synthesized inside comment/inspect srcdocs, so inspect tweaks saved against an unannotated import still disappear again in normal preview/export. Please fix that persistence gap and I’m happy to take another pass.


apps/web/src/runtime/srcdoc.ts
```diff
@@ -46,7 +46,8 @@ export function buildSrcdoc(
   </head>
   <body>${html}</body>
 </html>`;
-  const withSourcePaths = options.editBridge ? annotateManualEditSourcePaths(wrapped) : wrapped;
+  const withOdIds = annotateMissingOdIds(wrapped);
+  const withSourcePaths = options.editBridge ? annotateManualEditSourcePaths(withOdIds) : withOdIds;
```
mrcfps, line 52: Running `annotateMissingOdIds()` before `annotateManualEditSourcePaths()` makes every synthetic target here look like it already has a stable OD id, so `annotateManualEditSourcePaths()` immediately bails on the same nodes (`if (el.hasAttribute('data-od-id')) return;` at line 76). That drops the `data-od-source-path` markers the edit bridge needs to map selections back to source-authored DOM, so manual edit loses its stable source mapping for imported files. This is already reproducible on the current head: the `Validate workspace` job is red on `tests/runtime/srcdoc.test.ts > marks source-authored edit targets before runtime scripts can add nodes`, which now receives `<main data-od-id="path-0">…` with no `data-od-source-path`. Please either run `annotateManualEditSourcePaths()` before synthesizing fallback IDs, or let the source-path pass still add `data-od-source-path` when the existing `data-od-id` came from this fallback annotation, then add an `editBridge` regression that covers the imported-html path.

mrcfps (changes requested):
@YUHAO-corn Thanks for iterating on the imported-selection fix 🙏 The latest head closes the earlier persistence gap, but it introduces one new blocker in the changed srcdoc pipeline: fallback `data-od-id` injection now runs before edit-mode source-path annotation, which breaks manual-edit source mapping and is already failing CI on this head.


lefarcen (changes requested):
Hey @YUHAO-corn! 👋 Re-reviewing after your latest push (8d22339).

**Progress:** You fixed the persistence gap I flagged earlier — `annotateMissingOdIds()` now runs unconditionally, so saved tweaks will survive across render modes. The expanded selector (h1-h6, button, a, direct-child divs) is a solid improvement over the original narrow set. Nice iteration!

**New P1 blocker (per mrcfps):** The pipeline order breaks manual-edit source mapping. Running `annotateMissingOdIds()` *before* `annotateManualEditSourcePaths()` causes the source-path pass to skip elements that already have fallback `data-od-id`s (line 76 guard: `if (el.hasAttribute('data-od-id')) return;`). This drops the `data-od-source-path` markers the edit bridge needs, breaking manual edit for imported HTML. CI confirms this — the "Validate workspace" job is failing on the "marks source-authored edit targets" test.

**Fix:** Either (1) swap the pipeline order so `annotateManualEditSourcePaths()` runs *before* `annotateMissingOdIds()`, or (2) modify the source-path pass to allow `data-od-source-path` on elements with fallback `data-od-id`s. Option 1 is simpler.

Once the pipeline order is fixed and CI is green, this should be ready. Let me know when you push the fix!

lefarcen (dismissed):
Approved — the pipeline order fix resolves the P1 blocker.

You swapped the guard in `annotateManualEditSourcePaths()` from checking `data-od-id` to checking `data-od-source-path`, so now elements that received fallback `data-od-id`s from `annotateMissingOdIds()` will still get `data-od-source-path` markers. The failing test "marks source-authored edit targets" should now pass.

The persistence gap from earlier rounds is also resolved (unconditional `annotateMissingOdIds()`), and the expanded selector (h1-h6, button, a, direct-child divs) properly covers common imported HTML shapes.

Once CI confirms green, this is ready to merge. Nice work iterating through the feedback!

YUHAO-corn:
@mrcfps 感谢之前的反馈！最新的 commit  修复了 pipeline order 的问题：

-  的 guard 从  改为 
- 这样即使  (always-on) 先给元素加了 ，source-path 标记仍然会被正确写入
- CI 的 `marks source-authored edit targets` 测试现在通过了

 persistence gap (unconditional ) 和 selector coverage 也都在之前的 commit 中解决了。请看一下最新的 diff，谢谢！

lefarcen:
@mrcfps 这个 PR 的最新 commit (0a7843d0) 还在等你看一下。

YUHAO-corn:
@lefarcen @mrcfps — rebased to resolve the merge conflict with the just-merged #1170. CI is green on the latest head (`4d7b187d`). Would appreciate a re-review when you have a moment. Thanks!

lefarcen:
Hey @YUHAO-corn — thanks for rebasing this and getting the latest head (`4d7b187d`) back to green CI.

@mrcfps this one is waiting on your re-review now that the previous pipeline-order blocker has been addressed before the rebase. Could you take another pass when you have a moment?

lefarcen (approved):
Re-approving on the rebased head (`4d7b187d`) — the functional fix is identical to `0a7843d0` which I approved earlier (dismissed by branch-protection when the rebase landed).

Verified on current head:
- Pipeline order is correct: `annotateMissingOdIds` runs before `annotateManualEditSourcePaths`, and the guard in `annotateManualEditSourcePaths` now checks `MANUAL_EDIT_SOURCE_PATH_ATTR` instead of `data-od-id` — so elements that received fallback IDs still get their `data-od-source-path` markers. ✅
- `annotateMissingOdIds` runs unconditionally (not gated on bridge mode), so persistence across render modes is solid. ✅
- CI is green. ✅

All P1 blockers from previous rounds are resolved. From my side, this is clean. Waiting on @mrcfps to re-review the current head.

mrcfps (approved):

PR #1144, opened 2026-05-10
fix(web): 写盘前拦截 prose-as-HTML artifact (#50)

mutoe:
## 概要

为 `persistArtifact` 在 `<artifact type="text/html">` 持久化路径上加结构性 HTML gate，拦住 #50 现场报告里的"幽灵 HTML 文件"——AI 把一句中文总结塞进 `<artifact>` 块、被系统当合法 HTML 落盘到项目文件面板的失败模式。

## 修复内容

- 新增 `apps/web/src/artifacts/validate.ts` 暴露纯函数 `validateHtmlArtifact(content)`：非空 + 长度 ≥ 64 字符 + 含 `<!doctype html>` 或 `<html>` 标签（大小写不敏感、容忍 BOM 与前导空白）
- `apps/web/src/components/ProjectView.tsx` 的 `persistArtifact` 在 `ext === '.html'` 分支调用 gate，失败时通过 `setError` 报错并 return，不进入写盘流程
- 测试：`apps/web/tests/artifacts/validate.test.ts` 9 例（empty / whitespace / #50 现场 prose / 含尖括号但太短 / 长 prose 无 HTML 标签 / doctype-only / html-only / 大小写不敏感 / BOM 容忍）

## 与 #50 的关系

#50 列出了三层修复（写前内容校验 / identifier-scoped revision 模型 / history UI）。本 PR 只做第一层"写前内容校验"，因为：
- 改动小、独立、纯防御
- 不依赖 schema/UX 决策，可以立即合并
- 与第二层（revision 模型）正交，不阻塞那边的方案讨论

## scope 边界

**只 gate `<artifact>` 标签持久化路径**——`persistArtifact` 是 streaming parser 拿到 `<artifact>...</artifact>` 后调用的入口。

**不 gate 用户手动保存路径**——`FileViewer` / `FileWorkspace` 里用户在编辑器里写 HTML 草稿时走 `writeProjectTextFile` 直调（无 `artifactManifest`），不经过 `persistArtifact`，不受本 PR 影响。这是有意保留的——用户保存 WIP HTML 是合理的，不应该被结构性 gate 阻拦。

## 根因 prompt 层后续

AI 之所以会在没有新 HTML 产出的情况下仍发空壳 `<artifact>`，根源在 `apps/daemon/src/prompts/official-system.ts` 的 "Artifact handoff" 写成了非协商性规则、缺少免发条款。已拆出 #1143 单独跟进。本 PR 是持久化层兜底防御，与 #1143 互补：

[...]

PR #1171, opened 2026-05-10
feat(daemon): guard against agent-emitted stub artifact regressions

sebastianwestberg:
## Relationship to #50, #1143, #1144

This is **layer 3** of a three-layer fix for the phantom-HTML-artifact failure mode tracked in #50:

| Layer | PR | Strength | What it blocks |
|---|---|---|---|
| Prompt | #1143 | probabilistic — relies on model compliance | the model emitting a stub `<artifact>` in the first place |
| Web persistence (structural sniff) | #1144 (merged 2026-05-10) | deterministic | bodies that obviously aren't HTML (no `<!doctype html>` / `<html>` opener) |
| **Daemon persistence (regression guard)** | **this PR** | deterministic | HTML-shaped fallback stubs that *pass* #1144's structural sniff but are dramatically smaller than a prior sibling sharing the same `metadata.identifier` |

The three layers are complementary, not redundant. From real on-disk samples in the bug report:

- `test-generator-14.html` (25 B): bare filename string. **#1144 catches** (no doctype, length < 64). Mine wouldn't (no prior to compare).
- `test-generator-12-2.html` (233 B): a proper `<!doctype html>` page reading "see test-generator-12.html". **#1144 passes** (has doctype, ≥ 64 chars). **Mine catches** (size regression vs the 128 KB prior).
- `test-generator-9-2.html` (1.9 KB): full HTML structure with minimal app content. **#1144 passes**. **Mine catches**.

## Summary

When an agent emits an `<artifact>` block whose body is a placeholder — a sentence pointing at another file, a bare filename string, or a tiny fallback HTML page — the daemon writes the placeholder to disk verbatim. Users see a 25–500 byte HTML file where their previous version had tens of kilobytes of real markup. The only existing validation in `writeProjectFile` was empty/oversize.

This adds a structural regression guard. Before persisting an HTML artifact whose manifest carries `metadata.identifier`, scan the project dir for prior siblings matching `^<identifier>(-\d+)?\.html?$` and compare sizes. If the new body is below `minRetainedRatio` (default `0.2`) of the largest prior sibling — and that prior is at least `minPriorBytes` (default `4096`) — flag a regression.

The check is anchored on size delta, not phrasing, so it works for any agent backend behind the agent-adapter abstraction (`docs/agent-adapters.md`) — not just one model's stub phrasing.

## Behavior

Configurable via env vars (`readArtifactStubGuardConfigFromEnv` reads them per-write):

| Var | Values | Default |
|---|---|---|
| `OD_ARTIFACT_STUB_GUARD` | `reject` \| `warn` \| `off` | `warn` |
| `OD_ARTIFACT_STUB_GUARD_MIN_RATIO` | `0 < n <= 1` (1 = reject any shrinkage) | `0.2` |
| `OD_ARTIFACT_STUB_GUARD_MIN_PRIOR_BYTES` | positive integer | `4096` |

- **`warn` (default)**: writes the file, attaches `stubGuardWarning` (typed in `packages/contracts/src/api/files.ts`) to the response so the frontend can surface a notice without breaking flows.
- **`reject`**: writer throws `ArtifactRegressionError` *before* `fs.writeFile`; route returns `422 ARTIFACT_REGRESSION` with the prior sibling's name and size in `error.details`. Neither the stub HTML nor its `.artifact.json` companion lands on disk.
- **`off`**: skips the guard entirely.

Both `warn` and `reject` emit a `console.warn` line so on-call can see how often the guard fires.

The guard ignores: first emissions of a new identifier, manifest kinds outside `{html, deck}`, priors below the byte floor, and same-name overwrites (the new entry is excluded from its own prior-sibling scan). Decks are HTML on disk and have the same regression failure mode (deck-framework prompts ship multi-KB content), so they're guarded too.

## Notes for reviewers

- **Contract update**: `ProjectFile` in `packages/contracts/src/api/files.ts` gains `stubGuardWarning?: ProjectFileStubGuardWarning` so the daemon's `warn`-mode response stays type-safe for web consumers.
- **Pre-existing partial-write hazard**: the body-then-manifest write order pre-dates this change. If `fs.writeFile(target, body)` succeeds and the subsequent manifest write fails, you have a body without its sidecar. This patch leaves that ordering as-is. The reject path throws *before* `writeFile`, so rejections never leave a partial state behind — verified by the `does not write the new file or its manifest when rejected` test.
- **False-positive surface**: `reject` mode can block a user who genuinely wants to ship a small revision after a large prior. The error message names the prior file and ends with `Set OD_ARTIFACT_STUB_GUARD=warn or =off to bypass.` so users can self-serve.

## Test plan

- [x] `pnpm guard` clean
- [x] `pnpm --filter @open-design/daemon typecheck` clean
- [x] New `apps/daemon/tests/artifact-stub-guard.test.ts` (32 unit tests): pure classifier across mode/identifier/floor/ratio combinations; disk scan finds bare and `-N` suffixed siblings, including `.htm`; env config parsing including out-of-range and `ratio=1`; `ArtifactRegressionError` shape.
- [x] New `apps/daemon/tests/projects-stub-guard.test.ts` (13 HTTP tests): rejects stub rewrites with `422 ARTIFACT_REGRESSION`, allows comparable-size revisions, `warn` mode attaches `stubGuardWarning` to a 200 response, `off` mode bypasses, accepts first emission of a new identifier, rejects deck-kind stub rewrites, detects prior siblings written with `.htm` extension, and verifies rejected writes leave neither the HTML nor its manifest sidecar on disk.
- [x] Adjacent `tests/projects-routes.test.ts` and `tests/artifact-manifest.test.ts` unchanged and still pass.


apps/daemon/src/artifact-stub-guard.ts
```diff
@@ -0,0 +1,197 @@
...
+  let entries: Dirent[];
+  try {
+    entries = await readdir(projectDir, { withFileTypes: true });
+  } catch {
+    return [];
+  }
+  const results: PriorArtifactSibling[] = [];
+  for (const entry of entries) {
+    if (!entry.isFile()) continue;
+    if (entry.name === excludeSafeName) continue;
```
lefarcen: **P1** — `excludeSafeName` skips the new entry during the prior scan, but when overwriting an existing file with the **same name**, that existing file is prior content (not "the new entry"). So `dashboard.html` → `dashboard.html` (stub) bypasses the guard entirely.

This is the most direct regression pattern: user has a 60KB artifact, agent rewrites it to 100 bytes, same filename.

Fix: when `safeName` already exists on disk, **include it** in the prior-candidate list. Only exclude it when it's truly a new write.

apps/daemon/src/artifact-stub-guard.ts
```diff
@@ -0,0 +1,197 @@
...
+// newly-written artifact. The frontend's collision-suffixing scheme means
+// related entries match `<identifier>(-\d+)?\.html?`. We scan once per
+// write; project directories are small so this is cheap.
+export async function findPriorArtifactSiblings(
+  projectDir: string,
+  identifier: string,
+  excludeSafeName: string,
+): Promise<PriorArtifactSibling[]> {
+  if (identifier.length === 0) return [];
+  const pattern = new RegExp(`^${escapeRegExp(identifier)}(?:-\\d+)?\\.html?$`);
```
lefarcen: **P2** — This regex matches filenames, but `metadata.identifier` can have spaces, uppercase, punctuation, non-ASCII. The web persistence path sanitizes those to lowercase-kebab filenames while storing the raw identifier in the manifest.

So identifier `Worker Edition V2` → filename `worker-edition-v2.html`. The regex here wouldn't match a prior `Worker Edition V2.html` sibling because it expects exact identifier match.

Also: nested/imported artifacts are missed because you're only checking `Dirent.name` in the project root.

Fix: scan HTML files **and** their `.artifact.json` sidecars, compare `artifactManifest.metadata.identifier`, use filename regex only as fallback.

packages/contracts/src/api/files.ts
```diff
@@ -25,6 +39,7 @@ export interface ProjectFile {
   mime: string;
   artifactKind?: ArtifactKind;
   artifactManifest?: ArtifactManifest;
+  stubGuardWarning?: ProjectFileStubGuardWarning;
```
lefarcen, line 42: **P2** — This type shape is good, but the web client doesn't consume it yet.

In default `warn` mode: daemon returns `stubGuardWarning`, but `apps/web/src/providers/registry.ts:1073` + other artifact save paths don't check this field. The stub lands, user sees nothing.

In `reject` mode: `writeProjectTextFile` collapses the 422 response to `null`, discarding the actionable error message.

Fix: parse `ApiErrorResponse` on non-OK writes and surface `stubGuardWarning` in the artifact persistence UI (toast/banner/etc).

apps/daemon/tests/projects-stub-guard.test.ts
```diff
@@ -0,0 +1,243 @@
...
+
+  async function postFile(projectId: string, body: FilePostBody) {
+    return fetch(`${baseUrl}/api/projects/${projectId}/files`, {
+      method: 'POST',
+      headers: { 'Content-Type': 'application/json' },
+      body: JSON.stringify(body),
+    });
+  }
+
+  it('rejects a stub-sized rewrite with ARTIFACT_REGRESSION', async () => {
```
lefarcen, line 73: **P2** — These HTTP tests only cover suffixed new filenames (`report-3.html`) whose base names match clean identifiers. They miss:

1. Same-name overwrites: `POST /projects/X/files` with `name=report.html` when `report.html` already exists with a large prior body.
2. Identifier mismatch: identifier `Worker Edition V2` stored as `worker-edition-v2.html` shouldn't match a prior `Worker Edition V2.html` sibling.

Both should currently **fail to guard** and would lock the intended semantics if added as regression cases.

lefarcen (changes requested):
Hey @sebastianwestberg! 👋 Thanks for this comprehensive regression guard — the size-delta approach is smart and backend-agnostic.

I found 4 issues that need addressing before merge:

**P1 — Same-name overwrite blind spot**: The guard skips `safeName` during prior-sibling scan, so a direct overwrite of `dashboard.html → dashboard.html` (same identifier, stub body) bypasses the check entirely. This is the most common regression pattern.

**P2 — Identifier-vs-filename mismatch**: The regex only matches sanitized filenames, so identifiers with spaces/uppercase/punctuation won't find their own priors. Also misses nested/imported artifacts because only root `Dirent.name` is checked.

**P2 — Web client can't consume the warning**: In `warn` mode, `stubGuardWarning` is returned but no caller checks it. In `reject` mode, the 422 error text is discarded. The guard fires but users see nothing.

**P2 — Test coverage gaps**: HTTP tests only cover suffixed clean filenames, missing the two blind spots above.

See inline comments for specifics + suggested fixes. The guard works well for sibling-collision cases; these fixes extend it to the highest-risk same-file and web-integration scenarios.

apps/daemon/src/artifact-stub-guard.ts
mrcfps: `findPriorArtifactSiblings` builds its regex from `metadata.identifier` verbatim, but the persisted filename is usually the sanitized slug, not the raw identifier. On the main artifact path `persistArtifact` lowercases the identifier/title and replaces spaces or punctuation before writing (for example, an identifier like `Landing Page` becomes `landing-page.html`). With the current pattern, that prior file never matches, so `warn`/`reject` silently stop working for any non-slug identifier even though those are still valid manifests. Please derive the match key from the same sanitized basename that produced `safeName` (or reuse the existing filename-normalization helper) and add a regression test that saves `landing-page.html` from a `metadata.identifier` such as `Landing Page`.

apps/daemon/src/projects.ts
```diff
@@ -394,19 +400,58 @@ export async function writeProjectFile(
...
+        ? validatedManifest.metadata.identifier
+        : '';
+      // Stub-guard applies to HTML-rendered manifest kinds (html, deck).
+      // Other kinds (markdown, svg, code-snippet) can legitimately be small
+      // and are skipped.
+      if (identifier.length > 0 && STUB_GUARDED_MANIFEST_KINDS.has(validatedManifest.kind)) {
+        const guard = await evaluateArtifactStubGuard({
+          projectDir: dir,
+          safeName,
+          identifier,
```
mrcfps, line 422: This always passes the project root as `projectDir`, so sibling discovery only scans that one directory. If a caller writes `reports/overview-2.html` with `metadata.identifier: "overview"`, the guard never looks inside `reports/`, which means nested artifact rewrites bypass both `reject` and `warn` even though the target file lives beside a large prior sibling there. Because `writeProjectFile` accepts project-relative paths and already creates parent directories for them, this hole is reachable through the current API surface. Please scope the scan to `path.dirname(safeName)` (or pass the resolved target directory separately) and add an HTTP test that exercises a nested artifact path so this case stays covered.

mrcfps (commented):
Thanks for shipping this guard — I found two follow-up cases where the new check can still be bypassed, so I left inline suggestions before this lands.


apps/daemon/src/artifact-stub-guard.ts
sebastianwestberg: Fixed in 9cc82430. Verified the bypass: `persistArtifact` (`apps/web/src/components/ProjectView.tsx:1436`) reuses the same `fileName` when `savedArtifactRef.current === fileName`, so an in-session overwrite hits `dashboard.html` → `dashboard.html` and the previous exclude-by-name filter dropped the prior. Removed the filter — the on-disk size at scan time is always the prior because the writer's `fs.writeFile` runs *after* the guard. New HTTP test: `rejects a same-name overwrite that shrinks the existing file` asserts the original body is intact (not overwritten) on rejection.

apps/daemon/src/artifact-stub-guard.ts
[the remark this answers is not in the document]
sebastianwestberg: Fixed in 9cc82430. Confirmed: `persistArtifact` at `apps/web/src/components/ProjectView.tsx:1412-1444` slugifies the filename basename but stores the raw `art.identifier` in the manifest, so the regex was incomplete by construction. Build the regex from an alternation of the raw identifier *and* a slugified variant via `slugifyArtifactIdentifier` (mirrors the web slugifier exactly).

Caveat on severity: the prompt at `apps/daemon/src/prompts/official-system.ts:35` instructs `<artifact identifier="kebab-slug" ...>` and all 26 artifacts in the local project store carry slug-form identifiers — so this was a misuse-tolerance fix, not a hot bug. Worth doing for defensiveness; not "silently broken" today.

sebastianwestberg: Fixed in 9cc82430. Confirmed: `persistArtifact` at `apps/web/src/components/ProjectView.tsx:1412-1444` slugifies the filename basename but stores the raw `art.identifier` in the manifest, so the regex was incomplete by construction. Build the regex from an alternation of the raw identifier *and* a slugified variant via `slugifyArtifactIdentifier` (mirrors the web slugifier exactly).

Caveat on severity: the prompt at `apps/daemon/src/prompts/official-system.ts:35` instructs `<artifact identifier="kebab-slug" ...>` and all 26 artifacts in the local project store carry slug-form identifiers — so this was a misuse-tolerance fix, not a hot bug. Worth doing for defensiveness; not "silently broken" today.

sebastianwestberg: Fixed in 9cc82430. Confirmed: `persistArtifact` at `apps/web/src/components/ProjectView.tsx:1412-1444` slugifies the filename basename but stores the raw `art.identifier` in the manifest, so the regex was incomplete by construction. Build the regex from an alternation of the raw identifier *and* a slugified variant via `slugifyArtifactIdentifier` (mirrors the web slugifier exactly).

Caveat on severity: the prompt at `apps/daemon/src/prompts/official-system.ts:35` instructs `<artifact identifier="kebab-slug" ...>` and all 26 artifacts in the local project store carry slug-form identifiers — so this was a misuse-tolerance fix, not a hot bug. Worth doing for defensiveness; not "silently broken" today.

apps/daemon/src/projects.ts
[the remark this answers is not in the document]
sebastianwestberg: Fixed in 9cc82430. Pass `path.dirname(target)` as `scanDir` instead of the project root, so nested writes like `reports/overview-2.html` evaluate against `reports/`'s siblings. New HTTP test (`rejects stub regressions in subdirectories`) confirms.

Note: in current usage the artifact-tag persistence path always writes flat to project root (frontend doesn't generate nested fileNames), so this was a defensive close-the-API-hole fix rather than a hit production bug. The hole was real because `writeProjectFile` already does `mkdir path.dirname(target)`.

sebastianwestberg: Fixed in 9cc82430. Pass `path.dirname(target)` as `scanDir` instead of the project root, so nested writes like `reports/overview-2.html` evaluate against `reports/`'s siblings. New HTTP test (`rejects stub regressions in subdirectories`) confirms.

Note: in current usage the artifact-tag persistence path always writes flat to project root (frontend doesn't generate nested fileNames), so this was a defensive close-the-API-hole fix rather than a hit production bug. The hole was real because `writeProjectFile` already does `mkdir path.dirname(target)`.

packages/contracts/src/api/files.ts
sebastianwestberg: Partially addressed in 9cc82430.

`warn` mode: surfaced. `ProjectView.persistArtifact` now checks `file.stubGuardWarning` after `writeProjectTextFile` and renders the warning via the existing `setError` channel. Users see a banner instead of silence.

`reject` mode (422 collapsed to `null`): not fixed in this PR. The collapse happens in `writeProjectTextFile` (`registry.ts:1067-1078`) and any fix there changes the return contract for *all* callers (FileViewer, FileWorkspace, etc.). That's a cross-cutting refactor that belongs in its own PR — happy to file a follow-up issue or you can. The daemon side is correct (file isn't written, structured `error.details` is in the 422 body), so the data is there to surface; this PR just doesn't restructure the client.

apps/daemon/tests/projects-stub-guard.test.ts
sebastianwestberg: Test gaps closed in 9cc82430:
- `rejects a same-name overwrite that shrinks the existing file` (the P1 case)
- `finds slug-form sibling when manifest carries non-slug identifier`
- `rejects stub regressions in subdirectories`
- Plus the unit test `matches siblings using the slugified form of a non-slug identifier`

44 tests pass total (was 40).

sebastianwestberg:
## Round 2 — review fixes pushed (`9cc82430`)

Triaged the eight inline comments from @chatgpt-codex-connector, @lefarcen, and @mrcfps. They converged on five distinct concerns; verified each against the codebase, fixed the four that were technically correct, partially addressed the fifth.

| Concern | Verdict | Disposition |
|---|---|---|
| **Same-name overwrite bypass** (lefarcen P1) | Real, reachable via `persistArtifact`'s `savedArtifactRef.current` reuse | Fixed: dropped the wrong exclude-by-name filter; on-disk size at scan time *is* the prior |
| **Subdirectory scoping** (Codex P2, mrcfps) | API hole (writer accepts nested paths), not currently exploited by the artifact-tag path | Fixed: scan `path.dirname(target)` |
| **Slug vs raw identifier** (Codex P2, lefarcen P2, mrcfps) | Real for misuse; not hit by current data (prompt enforces kebab-slug, all 26 on-disk artifacts are slug-form) | Fixed: regex alternates raw + slugified forms via `slugifyArtifactIdentifier` mirroring the web slugifier |
| **Web client doesn't surface the warning** (lefarcen P2) | Real for both `warn` and `reject` modes | Partially fixed: `warn` mode now surfaces via `setError` in `ProjectView`. `reject` mode 422 collapse in `writeProjectTextFile` is a cross-cutting client refactor — deferred |
| **Test coverage gaps** (lefarcen P2) | Auto-true from above | Fixed: 4 new HTTP cases + 1 new unit case |

Test count: 40 → 44, all green. `pnpm guard` and `pnpm typecheck` clean.

Inline replies on each thread cite the same commit and explain the per-comment disposition; the `slug` and `subdir` reviewer claims are accurate as code paths but were not active production bugs — flagged for context, not minimized.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1468,6 +1468,19 @@ export function ProjectView({
       });
       if (file) {
         setFilesRefresh((n) => n + 1);
+        // Surface the daemon's stub-guard warning when it fires in `warn`
+        // mode (the default). Without this the warning would land in the
+        // file metadata silently and the user would never see that the
+        // model shipped a placeholder. Reject mode returns null from
+        // writeProjectTextFile (the 422 is collapsed by the API client) —
+        // surfacing that case requires plumbing error responses through
+        // writeProjectTextFile, tracked separately.
```
mrcfps: This still leaves `OD_ARTIFACT_STUB_GUARD=reject` as a silent failure in the web artifact flow. The new comment here calls out that `writeProjectTextFile()` collapses the daemon's `422 ARTIFACT_REGRESSION` response to `null`, and this callback only does anything inside `if (file)`. That means a user who opts into the documented reject mode gets no banner, no toast, and no hint that the save was blocked or how to recover, even though the daemon produced a structured error message specifically for that case. Please either plumb the 422 details through `writeProjectTextFile` in this change, or at minimum set a generic error when `file` comes back null after `persistArtifact` so reject-mode failures are observable.

mrcfps (commented):
Thanks for tightening the guard and closing the earlier bypasses, @sebastianwestberg. I found one remaining follow-up around the new reject-mode UX in the web client; details inline.


lefarcen (changes requested):
Hey @sebastianwestberg — thanks for the thorough fixes in 9cc82430! 🎯

You've closed the same-name overwrite and subdirectory scoping gaps. The slug/raw matching works for ASCII identifiers, but there's an edge case where non-ASCII identifiers (like `测试`) slugify to empty and fall back to `artifact.html` — that path still bypasses the guard.

Also, reject-mode UI surfacing is still incomplete (the 422 collapses to `null` in `registry.ts`).

**Remaining items:**

1. **P2** (`apps/daemon/src/artifact-stub-guard.ts` L109) — Empty-slug fallback bypass. When an identifier like `测试` slugifies to empty, `persistArtifact` falls back to `artifact.html` (after `slice(0,60)||'artifact'`). This guard only searches for the raw identifier + its slug, so a later `artifact-2.html` stub bypasses the prior. Mirror the frontend fallback logic or share the slugifier, and add a test for the empty-slug case.

2. **P2** (`apps/web/src/providers/registry.ts` L1073) — Reject-mode 422 responses collapse to `null` here, and `ProjectView` has no error path when `writeProjectTextFile` returns null. Users in `OD_ARTIFACT_STUB_GUARD=reject` mode see silent failures with no explanation. Warn-mode is surfaced; reject-mode is not.

3. **P3** (`apps/daemon/tests/projects-stub-guard.test.ts` L297) — Test coverage only exercises ASCII identifiers that slugify to non-empty (`Landing Page` → `landing-page`). Add a case for all-non-ASCII / punctuation-only identifiers (which hit the `artifact` fallback) to cover the empty-slug code path.

apps/web/src/components/ProjectView.tsx
sebastianwestberg: Fixed in d158c7e5. Took the "minimum" path you suggested: added an `else` branch in `persistArtifact` that surfaces a generic banner via `setError` when `writeProjectTextFile` returns null. Banner names the file, points at `OD_ARTIFACT_STUB_GUARD=reject` as the most likely cause, and tells the user to check daemon logs for structured ARTIFACT_REGRESSION details. Also clear `savedArtifactRef.current` on failure so a retry re-enters the persistence path instead of being skipped.

Plumbing the structured 422 details through `writeProjectTextFile` itself is still out of scope here — the helper's `Promise<ProjectFile | null>` shape is shared by FileViewer (3 call sites), FileWorkspace (2), and others, and changing it touches every caller. Filing that as a separate refactor would be the right next step.

packages/contracts/src/api/files.ts
sebastianwestberg: Fixed in d158c7e5. Took the "minimum" path you suggested: added an `else` branch in `persistArtifact` that surfaces a generic banner via `setError` when `writeProjectTextFile` returns null. Banner names the file, points at `OD_ARTIFACT_STUB_GUARD=reject` as the most likely cause, and tells the user to check daemon logs for structured ARTIFACT_REGRESSION details. Also clear `savedArtifactRef.current` on failure so a retry re-enters the persistence path instead of being skipped.

Plumbing the structured 422 details through `writeProjectTextFile` itself is still out of scope here — the helper's `Promise<ProjectFile | null>` shape is shared by FileViewer (3 call sites), FileWorkspace (2), and others, and changing it touches every caller. Filing that as a separate refactor would be the right next step.

sebastianwestberg:
## Round 3 — review fixes pushed (`d158c7e5`)

Two remaining items from @lefarcen's `CHANGES_REQUESTED` and @mrcfps's follow-up are now closed:

| Concern | Disposition |
|---|---|
| **Empty-slug fallback bypass** (lefarcen P2) — `测试` strips to empty, hits `persistArtifact`'s `\|\| 'artifact'` fallback, becomes `artifact.html`; previous guard didn't search for that family | Fixed: `EMPTY_SLUG_FALLBACK_NAME = 'artifact'` added as a sibling-name candidate when the slug is empty |
| **Reject-mode UI silence** (mrcfps + lefarcen P2) — 422 collapsed to null in `writeProjectTextFile`, no UI surface | Fixed via the "minimum" path mrcfps suggested: `persistArtifact` now sets a generic error banner on null return + clears `savedArtifactRef.current` so retries work. Naming the cause (`OD_ARTIFACT_STUB_GUARD=reject`) and pointing at daemon logs for the structured details |
| **Test for empty-slug case** (lefarcen P3) | Closed: 1 new unit test + 1 new HTTP test |

Test count: 44 → 46. Daemon + web typecheck clean.

Still deferred: plumbing the structured 422 details through `writeProjectTextFile` itself is a cross-cutting client refactor (`Promise<ProjectFile | null>` is shared by 5+ call sites). The generic banner is the explicit "at minimum" path; the deeper refactor is appropriate as its own PR.

apps/daemon/src/artifact-stub-guard.ts
```diff
@@ -0,0 +1,220 @@
...
+): Promise<PriorArtifactSibling[]> {
+  if (identifier.length === 0) return [];
+  const tokens = new Set<string>();
+  tokens.add(identifier);
+  const slug = slugifyArtifactIdentifier(identifier);
+  if (slug.length > 0) tokens.add(slug);
+  // When the identifier slugifies to empty (e.g. all-non-ASCII), the web
+  // path falls back to the literal "artifact" basename. Match that family
+  // so a later artifact-2.html stub doesn't bypass the prior.
+  else tokens.add(EMPTY_SLUG_FALLBACK_NAME);
```
mrcfps, line 184: When `slugifyArtifactIdentifier()` returns empty, this branch now treats the entire `artifact(-N).html` family as prior history for *any* such identifier. That fixes the bypass for repeated saves of the same non-ASCII identifier, but it also creates false positives across different identifiers that both slugify to empty. In `persistArtifact` (`apps/web/src/components/ProjectView.tsx:1412-1416`), both `"测试"` and `"пример"` fall back to the same `artifact` basename, so the first save for a brand-new second identifier can now warn/reject against an unrelated earlier artifact even though there is no prior sibling with the same `metadata.identifier`. Please filter the `artifact*.html` fallback matches by each sibling manifest's `metadata.identifier` (or another persisted canonical key) before comparing sizes, and add an HTTP test that saves two distinct empty-slug identifiers to show the second one is accepted.

mrcfps (commented):
Thanks @sebastianwestberg — the main guard flow looks solid, but I found one follow-up in the new empty-slug fallback path.


lefarcen (changes requested):
Hey @sebastianwestberg — good progress on rounds 2 and 3! The empty-slug fix and reject-mode banner close most of the prior findings. 🎯

However, the new empty-slug fallback path introduces a collision risk: distinct identifiers that both slugify to empty (like `测试` and `首页`) now share the same `artifact*.html` namespace, so a first save for a brand-new second identifier can warn/reject against an unrelated earlier artifact.

**Remaining item:**

**P2** (`apps/daemon/src/artifact-stub-guard.ts` L120) — Empty-slug fallback collision. When two different raw identifiers both slugify to `artifact`, the guard treats them as the same prior-sibling family. Example: a large `artifact.html` with `metadata.identifier="测试"` followed by a first small `artifact-2.html` for `metadata.identifier="首页"` will warn/reject even though there's no prior with the same identifier. Filter fallback candidates by their sidecar manifest identifier (or another persisted canonical key) before comparing sizes, and add an HTTP test that saves two distinct empty-slug identifiers to confirm the second one is accepted.

packages/contracts/src/api/files.ts
sebastianwestberg: Fixed in daebf0ac. The fix you flagged (and lefarcen independently flagged in the round-4 review) is real — round-3's empty-slug fallback conflated distinct identifiers. Took your suggested approach: filename-pattern matching is now a *candidate generator*, not the source of truth. For each match the guard reads the `.artifact.json` sidecar and verifies `metadata.identifier` matches via a new `artifactIdentifiersMatch(a, b)` helper:
- raw equality, OR
- shared *non-empty* slug (so `Landing Page` and `landing-page` still bridge)

Empty-slug equivalence is deliberately not honored: `测试` and `首页` both slugify to empty but are not the same lineage. That's the whole bug.

Files without a sidecar are skipped — they weren't written through the artifact-tag path the guard targets.

Applied the verification to *all* matches, not just the empty-slug case, because the same shape exists symmetrically (input identifier `artifact` against unrelated empty-slug fallback files would have had the same conflation).

New HTTP test: `does NOT cross-reject distinct empty-slug identifiers` saves identifier `测试` then identifier `首页`, asserts the second is 200.

apps/daemon/src/artifact-stub-guard.ts
sebastianwestberg: Fixed in daebf0ac. The fix you flagged (and lefarcen independently flagged in the round-4 review) is real — round-3's empty-slug fallback conflated distinct identifiers. Took your suggested approach: filename-pattern matching is now a *candidate generator*, not the source of truth. For each match the guard reads the `.artifact.json` sidecar and verifies `metadata.identifier` matches via a new `artifactIdentifiersMatch(a, b)` helper:
- raw equality, OR
- shared *non-empty* slug (so `Landing Page` and `landing-page` still bridge)

Empty-slug equivalence is deliberately not honored: `测试` and `首页` both slugify to empty but are not the same lineage. That's the whole bug.

Files without a sidecar are skipped — they weren't written through the artifact-tag path the guard targets.

Applied the verification to *all* matches, not just the empty-slug case, because the same shape exists symmetrically (input identifier `artifact` against unrelated empty-slug fallback files would have had the same conflation).

New HTTP test: `does NOT cross-reject distinct empty-slug identifiers` saves identifier `测试` then identifier `首页`, asserts the second is 200.

sebastianwestberg:
## Round 4 — review fix pushed (`daebf0ac`)

@mrcfps and @lefarcen both caught a false-positive that round 3 introduced: the empty-slug fallback conflated distinct identifiers. Two non-ASCII identifiers like `测试` and `首页` both slugify to empty and share the `artifact*.html` basename, so a brand-new second save was being compared against the unrelated first. The same shape existed symmetrically for a non-empty `artifact` identifier matching empty-slug fallback files.

| Concern | Disposition |
|---|---|
| Cross-identifier false positive in fallback namespace (mrcfps inline + lefarcen P2) | Fixed: filename pattern is now a candidate generator, not the source of truth. For every candidate the guard reads the `.artifact.json` sidecar and verifies `metadata.identifier` matches via `artifactIdentifiersMatch` (raw equality OR shared *non-empty* slug). Empty-slug equivalence is intentionally not honored. Files without a sidecar are skipped. |

Applied verification to all matches, not just the fallback case, because the symmetric direction (`identifier="artifact"` matching unrelated empty-slug priors) had the same conflation potential.

Tests:
- Unit fixtures now write file+sidecar pairs (mirrors production)
- New `artifactIdentifiersMatch` suite covers the 5 equivalence cases (raw match, slug bridge, empty-slug non-match, self-match for empty-slug, distinct slugs)
- New HTTP test `does NOT cross-reject distinct empty-slug identifiers`: `测试` then `首页`, asserts second save is 200
- New unit test `skips files without a sidecar`

42 tests pass across stub-guard suites. Daemon typecheck + `pnpm guard` clean.

apps/daemon/src/artifact-stub-guard.ts
```diff
@@ -0,0 +1,257 @@
...
+// Two identifiers refer to the same artifact lineage when they're
+// literally equal OR they slugify to the same non-empty value. Empty-slug
+// equivalence (e.g. "测试" vs "首页") is *not* treated as a match because
+// distinct identifiers can both strip to empty, and conflating them
+// would re-introduce the false-positive that lefarcen / mrcfps flagged.
+export function artifactIdentifiersMatch(a: string, b: string): boolean {
+  if (a === b) return true;
+  const slugA = slugifyArtifactIdentifier(a);
+  if (slugA.length === 0) return false;
+  return slugA === slugifyArtifactIdentifier(b);
```
mrcfps: Here `artifactIdentifiersMatch()` treats any pair of identifiers with the same non-empty `slugifyArtifactIdentifier()` output as the same lineage. Because `slugifyArtifactIdentifier()` lowercases, strips punctuation, and then truncates to 60 characters, two distinct identifiers that only diverge after character 60 will now collide here and warn/reject against each other on the second file’s *first* save. For example, `A…A1` and `A…A2` both collapse to the same 60-character slug, so this reintroduces the cross-identifier false-positive you just fixed for the empty-slug fallback, only for long ASCII identifiers instead. Please narrow this bridge so it only accepts an unambiguous raw↔slug match, or persist/compare a canonical identifier that is not truncation-lossy, and add a regression test with two >60-character identifiers sharing the same first 60 slug characters.

mrcfps (commented):
Thanks for closing the earlier empty-slug false positive, @sebastianwestberg — I found one remaining edge case in the new identifier-matching helper around long slug collisions.


apps/daemon/src/artifact-stub-guard.ts
sebastianwestberg: Fixed in 92d0169c. The collision is real and the same shape as round 4's empty-slug bug, just at the other end of the input range.

Tightened the rule: slug-equivalence now requires at least one input to BE its own canonical slug form, not just for both slugs to match.

```ts
return a === slugA || b === slugB;  // was: return slugA === slugB
```

- `"Landing Page"` ↔ `"landing-page"`: second input IS the slug → bridges ✓
- `"A...A1"` ↔ `"A...A2"` (70 chars each, same first 60): neither is canonical → does not bridge ✓
- `"测试"` ↔ `"首页"`: slugA empty, returns false earlier (unchanged) ✓

Side-effect: two non-canonical forms that happen to slugify the same (e.g. `"Landing Page"` vs `"LANDING-PAGE"`) also no longer bridge. That's correct under the same safety property — without a canonical anchor we can't safely assert same lineage. Updated the slug-equivalence test to assert this explicitly.

New tests: `does NOT bridge two long raw identifiers that share a 60-char truncated slug` (the regression you asked for) + `still bridges raw form to truncated slug when the slug is the canonical second input` (proves the legitimate truncation case still works) + a negative test for the non-canonical-pair case.

45 tests pass.

Considered the alternative you mentioned ("persist a non-truncation-lossy canonical key"). That would be more robust long-term, but it's a contract change in `artifact-manifest.ts` (new field, migration) that affects every existing on-disk sidecar — too much scope for what's a defense-in-depth fix against a contrived input. Happy to revisit if real >60-char collisions show up in practice.

apps/daemon/src/artifact-stub-guard.ts
```diff
@@ -0,0 +1,264 @@
...
+    if (!pattern.test(entry.name)) continue;
+    // Verify the sibling's canonical identifier matches before treating
+    // it as a prior. Without this check, distinct identifiers that share
+    // a sibling-name namespace (most reachably the empty-slug fallback,
+    // where 测试 and 首页 both land in artifact*.html) would falsely
+    // warn/reject across each other. Files without a sidecar are skipped
+    // to keep the guard conservative — they weren't written via the
+    // artifact-tag path that this guard targets.
+    const sidecarIdentifier = await readSidecarIdentifier(scanDir, entry.name);
+    if (sidecarIdentifier === null) continue;
```
mrcfps: Supported legacy HTML artifacts can still bypass the new guard here. `readSidecarIdentifier()` returning `null` makes `findPriorArtifactSiblings()` skip the candidate entirely, but `readManifestForPath()` still treats sidecar-less `.html` / `.htm` files as valid artifacts via `inferLegacyManifest()` (`apps/daemon/src/artifact-manifest.ts:230-247`). That means a project with an older `dashboard.html` and no sidecar will let its first stub rewrite through as a supposed “first emission”, including the same-name overwrite case this PR is trying to catch for existing user artifacts. Please add a fallback for legacy HTML priors (for example, same-name or slug-family matches with an inferred html/deck manifest, or a one-time sidecar backfill) and cover it with an HTTP test that seeds a sidecar-less prior file before posting the stub rewrite.

mrcfps (commented):
Thanks for iterating on this, @sebastianwestberg — the guard now covers the earlier same-name, subdirectory, empty-slug, and truncation-collision cases. I found one remaining follow-up around legacy HTML artifacts that still don't have `.artifact.json` sidecars.


apps/daemon/src/artifact-stub-guard.ts
sebastianwestberg: Fixed in edbb9847. The legacy bypass is real: `readManifestForPath` (`projects.ts:425`) treats sidecar-less `.html` files as legitimate artifacts via `inferLegacyManifest`, so my round-4 strict-sidecar policy was inconsistent with how the rest of the system reads those files.

Took the filename-fallback route you suggested (the lighter of the two — sidecar backfill is bigger scope and more invasive than the bug warrants):

```ts
const sidecarIdentifier = await readSidecarIdentifier(scanDir, entry.name);
const candidateIdentifier = sidecarIdentifier ?? syntheticIdentifierFromFilename(entry.name);
if (!artifactIdentifiersMatch(identifier, candidateIdentifier)) continue;
```

`syntheticIdentifierFromFilename` strips the `(-\d+)?\.html?` suffix to recover the per-lineage basename. Since synthetic IDs come from already-slugified filenames, the round-5 canonical-form rule keeps the bridge safe — they only match raw inputs via legitimate slug-equivalence, not via truncation collision or empty-slug conflation. Verified explicitly:

- Legacy `dashboard.html` (no sidecar) + input `dashboard` → bridges (synthetic === input) ✓
- Legacy `landing-page.html` + input `Landing Page` → bridges via slug-equivalence (canonical anchor on the synthetic side) ✓
- Legacy `dashboard.html` + input `legacy-dashboard` → does NOT bridge (different slugs) ✓

New HTTP test (`catches stub rewrites of legacy sidecar-less HTML priors`) seeds a sidecar-less prior by POSTing without `artifactManifest` (the route writes the body but no `.artifact.json` companion — same on-disk shape as a pre-sidecar legacy file), then asserts the stub rewrite is rejected with `422 ARTIFACT_REGRESSION`. Plus 3 unit tests for the synthetic-identifier semantics. 48 tests pass.

Note on follow-ups: this is the third "edge case the previous edge-case fix exposed" round in a row (R4 empty-slug → R5 truncation → R6 legacy). Happy to keep iterating, but each round is finding narrower scenarios. If you spot any practical bypass (something that would actually trigger in real projects), please flag explicitly; otherwise I'll treat further hardening of `findPriorArtifactSiblings` against contrived inputs as out of scope for this PR.

apps/daemon/src/artifact-stub-guard.ts
```diff
@@ -0,0 +1,284 @@
...
+// extension to recover the per-lineage basename. Used as a fallback
+// identifier for legacy HTML artifacts that don't have a sidecar yet
+// (anything written before the manifest era, or HTML pasted/uploaded
+// outside the artifact-tag flow). `inferLegacyManifest` already treats
+// these as html-kind artifacts elsewhere; we mirror that so the guard
+// doesn't silently let a stub overwrite them.
+const SYNTHETIC_IDENTIFIER_SUFFIX = /(?:-\d+)?\.html?$/;
+
+function syntheticIdentifierFromFilename(name: string): string {
+  return name.replace(SYNTHETIC_IDENTIFIER_SUFFIX, '');
```
mrcfps, line 147: `syntheticIdentifierFromFilename()` strips any terminal `-N` here before comparing the legacy prior's inferred identifier. That breaks the new fallback for real identifiers that already end in digits, like `phase-2`: a sidecar-less `phase-2.html` prior is inferred as `phase`, so a later stub rewrite with `metadata.identifier: "phase-2"` is treated as a first emission and bypasses the guard. This matters because numeric suffixes are common in artifact names, and this branch is now the only protection path for supported legacy HTML without sidecars. Since the changed code always applies `/(?:-\d+)?\.html?$/`, the mismatch is reproducible from the diff alone: `artifactIdentifiersMatch("phase-2", "phase")` returns false. Could you treat both the full basename and the suffix-stripped basename as legacy candidates here, then add a unit/HTTP regression that seeds a sidecar-less `phase-2.html` before posting the stub rewrite?

mrcfps (commented):
Thanks for iterating on this, @sebastianwestberg — the legacy fallback closes the previous sidecar-less blind spot. I found one follow-up around numeric `-N` identifiers in that new inference path; details inline.


apps/daemon/src/artifact-stub-guard.ts
sebastianwestberg: Fixed in aaeb1870. The ambiguity is real and reachable from the diff exactly as you described. Took the both-interpretations approach you suggested.

```ts
function legacyCandidateIdentifiers(filename: string): string[] {
  const fullBasename = filename.replace(HTML_EXTENSION, '');
  const stripped = syntheticIdentifierFromFilename(filename);
  const candidates: string[] = [];
  if (fullBasename.length > 0) candidates.push(fullBasename);
  if (stripped.length > 0 && stripped !== fullBasename) candidates.push(stripped);
  return candidates;
}
```

Then in the scan: `if (!candidates.some((c) => artifactIdentifiersMatch(identifier, c))) continue;`

Trade-off worth flagging: this does broaden the legacy-only path. A sidecar-less `phase-2.html` will now match BOTH input identifier `phase-2` (full basename) and input identifier `phase` (stripped). In the rare case where `phase.html` and `phase-2.html` both exist as sidecar-less legacy files AND a new artifact has identifier `phase-2`, the new write would compare against both — which is correct: any size regression vs the largest matched prior should fire. The canonical-anchor rule in `artifactIdentifiersMatch` still prevents truncation collisions and empty-slug conflations from leaking through either candidate.

Visible false positives (user can override via env or rename) are preferable to silent false negatives (stub bypasses the guard).

New HTTP test (`catches stub rewrites of legacy priors whose identifier ends in -<digits>`) seeds a sidecar-less `phase-2.html` and asserts the stub rewrite for identifier `phase-2` is rejected with 422 — closes the exact reproducer from the diff. Plus 2 unit tests for the dual-candidate semantics. 51 tests pass.

That's R4 → R5 → R6 → R7 each fixing an edge case the previous fix exposed in `findPriorArtifactSiblings`. The pattern's now visible: every tightening of the matching rule creates a new asymmetry, and every loosening creates a new collision class. The current rule (sidecar identity OR dual-candidate filename inference, all gated through `artifactIdentifiersMatch` with canonical-form anchor) covers every concrete failure mode you've raised. If a real-world bypass shows up I'll fix it; further hardening of the helper against constructed inputs (e.g. `phase-2-3.html` triple-suffix ambiguity, etc.) belongs in a follow-up rather than this PR. Maintainer decision on when to land.

mrcfps (approved):
@sebastianwestberg I re-checked the latest head and walked the changed daemon/web/contracts ranges for the stub-guard flow, including the legacy sidecar fallback, reject/warn surfacing, and the added regression coverage around same-name, empty-slug, and numeric-suffix cases. I didn't find any remaining actionable issues in the current diff. Thanks for staying on the follow-up rounds here — this landed in a much safer place. 🙌


sebastianwestberg:
What about you @lefarcen?

lefarcen (approved):
Round 7 () is clean. All prior findings resolved:

- **Empty-slug fallback** now verifies sidecar  instead of relying on filename alone
- **Reject-mode UI collapse** surfaced via generic banner in  when save returns null
- **Test coverage** now includes empty-slug unit + HTTP cases

Follow-up concerns from mrcfps (empty-slug collision, truncation collision, legacy sidecar fallback, numeric-suffix ambiguity) are all addressed by the canonical-form anchor in  + dual-candidate legacy inference.

Ready to merge from my side. 🎉

lefarcen (approved):
Round 7 (aaeb1870) is clean. All prior findings resolved:

- Empty-slug fallback now verifies sidecar metadata.identifier instead of relying on filename alone
- Reject-mode UI collapse surfaced via generic banner in ProjectView when save returns null
- Test coverage now includes empty-slug unit + HTTP cases

Follow-up concerns from mrcfps (empty-slug collision, truncation collision, legacy sidecar fallback, numeric-suffix ambiguity) are all addressed by the canonical-form anchor in artifactIdentifiersMatch + dual-candidate legacy inference.

Ready to merge from my side.

leilei926524-tech:
Hi @sebastianwestberg — quick note from the Open Design team.

What you get there:

PR #1176, opened 2026-05-10
fix(security): resolve hostname before approving external API base URLs

aaronjmars:
## Summary

The daemon-side base-URL validator currently only inspects the literal hostname string. A public DNS record that points at an internal address — `internal.example.com → 10.0.0.5` — passes validation, and the daemon then issues the BYOK proxy request anyway. That turns `/api/proxy/*/stream` and `/api/projects/:id/finalize/anthropic` into a server-side request primitive against any reachable private network the daemon happens to sit on.

This PR adds a DNS-aware companion (`validateBaseUrlResolved`) that resolves the hostname and re-applies the existing block-list against every resolved address, then wires it into the wrapper the daemon already uses (`validateExternalApiBaseUrl`).

## Impact

- Affects every proxy/finalize handler that accepts an external `baseUrl`: `/api/proxy/{anthropic,openai,azure,google,ollama}/stream`, `/api/projects/:id/finalize/anthropic`.
- A caller able to reach those routes (legitimate web app, or anything that can satisfy `app.use('/api', …)` — non-browser clients without an `Origin` header pass through) can supply a hostname under their control that resolves to a private IP. The daemon's `fetch` then connects to that private IP, with attacker-controlled headers and POST body, and streams the response back over SSE.
- CWE-918 (SSRF). Class: DNS-based bypass — orthogonal to the literal-IP coverage that already shipped.

## Surface area

Daemon routes that accept an external `baseUrl` and now use the resolved-IP check:

- [x] `POST /api/proxy/anthropic/stream`
- [x] `POST /api/proxy/openai/stream`
- [x] `POST /api/proxy/azure/stream`
- [x] `POST /api/proxy/google/stream`
- [x] `POST /api/proxy/ollama/stream` (+ `redirect: 'error'` to close redirect-follow class)
- [x] `POST /api/projects/:id/finalize/anthropic`
- [x] `POST /api/test/connection` (provider mode)
- [x] `POST /api/provider/models`

Out of scope (and why):
- Active DNS rebinding (resolve-then-flip between validation and fetch) — requires fetch-time IP pinning via a custom Undici dispatcher. Tracked as a follow-up; see *Known limitation*.

## Location

- `apps/daemon/src/connectionTest.ts` — adds `validateBaseUrlResolved`; switches `testProviderConnection` to the resolved validator.
- `apps/daemon/src/providerModels.ts` — switches `listProviderModels` to the resolved validator.
- `apps/daemon/src/server.ts` — wires the resolved check into the existing `validateExternalApiBaseUrl` wrapper, adds `await` on the 6 call sites, and adds `redirect: 'error'` to the Ollama proxy fetch.

## Fix

`validateBaseUrlResolved`:

1. Runs the existing sync `validateBaseUrl` for the literal-host pre-check.
2. Short-circuits loopback hostnames (keeps the intentional carve-out for Ollama / local LLM providers) and IP-literal hostnames (already vetted by the sync pass).
3. Calls `dns.lookup(hostname, { all: true, family: 0 })` and re-applies `isBlockedExternalApiHostname` to every resolved address.
4. Treats resolver errors as non-security signals — `fetch` will surface a connection error naturally; turning a transient `ENOTFOUND` into a 403 would just confuse users.

The function takes an optional `lookup` parameter so tests can inject a fake resolver instead of touching the system DNS.

## Bug-fix verification (red → green)

| Attack vector | Before | After |
|---|---|---|
| Hostname → resolves to private IPv4 (`10.0.0.5`, `172.16.x`, `192.168.x`, `169.254.x`, `100.64.x` CGNAT) on `/api/proxy/*/stream` | Daemon fetches the private IP, streams response back over SSE | Rejected pre-fetch with `{ ok: false, kind: 'forbidden' }`, no upstream call |
| Hostname → resolves to private IPv6 (`fd00::`, `fe80::`, `::`) | Daemon fetches the private IP | Rejected pre-fetch |
| Dual-stack hostname where any record is private | Daemon picks resolved address, can hit private | Rejected pre-fetch |
| Same payload on `/api/projects/:id/finalize/anthropic` | Same SSRF | Rejected pre-fetch |
| Same payload on `/api/test/connection` (provider mode) | Sync validator passed, daemon fetched private IP | Rejected pre-fetch |
| Same payload on `/api/provider/models` | Sync validator passed, daemon fetched private IP | Rejected pre-fetch |
| Validated public host issues HTTP redirect to a private/internal URL on `/api/proxy/ollama/stream` | Daemon follows redirect, SSRF | Fetch fails with redirect error; no upstream connection to redirect target |
| Loopback / `*.localhost` carve-out (Ollama local) | Allowed | Allowed (unchanged) |
| Public-resolving hostname (`api.openai.com` → CF IPs) | Allowed | Allowed (unchanged) |

## Verification (tests)

- `apps/daemon/tests/connection-test.test.ts` — new `describe('validateBaseUrlResolved …')` block with 10 cases:
  - Sync-error passthrough (invalid URL).
  - Sync rejection of literal private IPs (skips DNS).
  - Loopback hostname short-circuit (`localhost`, `127.0.0.1`, `[::1]`).
  - IP-literal short-circuit (`1.2.3.4`, `[2606:4700::]`).
  - Private-IPv4 resolution (`10.0.0.5`, `172.16.0.5`, `192.168.1.5`, `100.64.0.1` CGNAT, `169.254.169.254`, `0.0.0.0`, `224.0.0.1`).
  - Private-IPv6 resolution (`fd00::1`, `fe80::1`, `::`).
  - Dual-stack rejection when any record is private.
  - Public→public pass (`api.openai.com` → public CF IPs).
  - `*.localhost` → loopback pass (the existing Ollama carve-out extended through DNS).
  - DNS resolver error fallback.
- `apps/daemon/tests/proxy-routes.test.ts`:
  - `POST /api/provider/models` rejects a hostname whose `dnsPromises.lookup` returns `10.0.0.5`, with no upstream fetch.
  - `POST /api/test/connection` provider mode: same shape, asserts `{ ok: false, kind: 'forbidden' }`.
  - `/api/proxy/ollama/stream`: asserts the upstream fetch init carries `redirect: 'error'`.
- Local: `pnpm typecheck` ✅ · `pnpm guard` ✅ · `pnpm vitest run tests/connection-test.test.ts tests/proxy-routes.test.ts` → 93/93 ✅.

## Known limitation

This does not close the DNS-rebinding window — a hostname that resolves to a public IP at validation time and a private IP at fetch time still slips through. The robust fix for that requires pinning the resolved IP through to the actual fetch (via a custom Undici dispatcher / `socket` callback). Happy to follow up with that as a separate PR if you'd like — wanted to keep this change narrowly scoped to the static-DNS class.

## Carve-outs preserved

- Loopback hostnames (`localhost`, `127.0.0.1`, `[::1]`, `[::ffff:127.0.0.1]`) → allowed, no DNS call.
- Loopback-resolving hostnames (`*.localhost`, `lvh.me` and similar) → allowed.
- Public-resolving hostnames → allowed.
- All `validateBaseUrl` (`packages/contracts`) behavior is unchanged — settings-dialog smoke tests still get an immediate sync answer.

## Detected by

aeon (manual review + semgrep + osv + trufflehog).
- Severity: HIGH (CWE-918, the validator's stated purpose is exactly this defense).
- Class: DNS-based SSRF bypass.

---

Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron). Happy to revise the patch shape if you'd prefer the DNS step in `packages/contracts` (would require dropping Node deps from contracts; current shape keeps contracts pure).

### DIFF
--- a/apps/daemon/src/connectionTest.ts
+++ b/apps/daemon/src/connectionTest.ts
@@ -17,6 +17,7 @@
 // contracts so Settings and daemon-side checks reject the same hosts.
 
 import { spawn } from 'node:child_process';
+import { promises as dnsPromises } from 'node:dns';
 import { promises as fsp } from 'node:fs';
 import os from 'node:os';
 import path from 'node:path';
@@ -41,9 +42,11 @@ import {
 } from './runtimes/auth.js';
 import type { AgentCliEnvPrefs } from './app-config.js';
 import {
+  isBlockedExternalApiHostname,
   isLoopbackApiHost,
   validateBaseUrl,
   type AgentTestRequest,
+  type BaseUrlValidationResult,
   type ConnectionTestKind,
   type ConnectionTestProtocol,
   type ConnectionTestResponse,
@@ -52,6 +55,68 @@ import {
 } from '@open-design/contracts/api/connectionTest';
 
 export { validateBaseUrl } from '@open-design/contracts/api/connectionTest';
+
+// DNS-aware companion to `validateBaseUrl`. The contracts-side check only
+// inspects the literal hostname string, so a public DNS name pointing at
+// internal infrastructure (`internal.example.com → 10.0.0.5`) slips through
+// and the daemon ends up issuing a request to a private address on behalf of
+// whichever caller supplied the base URL. Resolve the hostname and re-run
+// the block-list against every address the system would actually connect to.
+//
+// Loopback is intentionally allowed for local LLM providers like Ollama; any
+// hostname that resolves to a loopback address (including `*.localhost` per
+// RFC 6761 and IPv4-mapped IPv6 loopback) follows that same carve-out.
+//
+// DNS lookup failures are *not* treated as a security signal — the caller is
+// going to surface a connection error from `fetch` anyway, and turning a
+// transient resolver hiccup into a 403 would just confuse users. The sync
+// hostname check still rejected the obvious literal-IP cases before we ever
+// got here.
+
+export type DnsLookupAddress = { address: string; family: number };
+export type DnsLookupFn = (hostname: string) => Promise<DnsLookupAddress[]>;
+
+const defaultDnsLookup: DnsLookupFn = async (hostname) => {
+  const result = await dnsPromises.lookup(hostname, { all: true, family: 0 });
+  return result.map(({ address, family }) => ({ address, family }));
+};
+
+function looksLikeIpLiteral(hostname: string): boolean {
+  const host = hostname.startsWith('[') && hostname.endsWith(']')
+    ? hostname.slice(1, -1)
+    : hostname;
+  if (/^\d{1,3}(?:\.\d{1,3}){3}$/.test(host)) return true;
+  return host.includes(':');
+}
+
+export async function validateBaseUrlResolved(
+  baseUrl: string,
+  lookup: DnsLookupFn = defaultDnsLookup,
+): Promise<BaseUrlValidationResult> {
+  const sync = validateBaseUrl(baseUrl);
+  if (sync.error || !sync.parsed) return sync;
+
+  const hostname = sync.parsed.hostname.toLowerCase();
+  if (isLoopbackApiHost(hostname)) return sync;
+  if (looksLikeIpLiteral(hostname)) return sync;
+
+  let addresses: DnsLookupAddress[];
+  try {
+    addresses = await lookup(hostname);
+  } catch {
+    return sync;
+  }
+
+  for (const addr of addresses) {
+    const ip = String(addr.address).toLowerCase();
+    if (isLoopbackApiHost(ip)) continue;
+    if (isBlockedExternalApiHostname(ip)) {
+      return { error: 'Internal IPs blocked', forbidden: true };
+    }
+  }
+
+  return sync;
+}
 
 // Aggressive but not punitive — happy paths usually return in under 2 s.
 // Override with OD_CONNECTION_TEST_PROVIDER_TIMEOUT_MS for slow networks
--- a/apps/daemon/src/connectionTest.ts
+++ b/apps/daemon/src/connectionTest.ts
@@ -622,7 +622,7 @@ export async function testProviderConnection(
 ): Promise<ConnectionTestResponse> {
   const start = Date.now();
   const model = String(input.model ?? '');
-  const validated = validateBaseUrl(input.baseUrl);
+  const validated = await validateBaseUrlResolved(input.baseUrl);
   if (validated.error || !validated.parsed) {
     const kind: ConnectionTestKind = validated.forbidden ? 'forbidden' : 'invalid_base_url';
     return {
--- a/apps/daemon/src/chat-routes.ts
+++ b/apps/daemon/src/chat-routes.ts
@@ -5,6 +5,7 @@ import {
   agentIdToTracking,
   projectKindToTracking,
 } from '@open-design/contracts/analytics';
+import { validateBaseUrlResolved } from './connectionTest.js';
 
 export interface RegisterChatRoutesDeps extends RouteDeps<'db' | 'design' | 'http' | 'chat' | 'agents' | 'critique' | 'validation' | 'lifecycle'> {}
 
@@ -46,7 +47,6 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
     critiqueResponseCapBytes,
     critiqueRunRegistry,
   } = ctx.critique;
-  const { validateBaseUrl } = ctx.validation;
   const isDaemonShuttingDown = ctx.lifecycle?.isDaemonShuttingDown ?? (() => false);
   const rejectProxyPluginContext = (body: Record<string, unknown>, res: any) => {
     if (
@@ -507,8 +507,13 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
   const redactAuthTokens = (text: string) =>
     text.replace(/Bearer [A-Za-z0-9_\-.+/=]+/g, 'Bearer [REDACTED]');
 
+  // DNS-aware wrapper. The sync `validateBaseUrl` only inspects the literal
+  // hostname string, so a public DNS name pointing at an internal address
+  // (`internal.example.com → 10.0.0.5`) still passes. We delegate to
+  // `validateBaseUrlResolved` here so every proxy/stream handler runs the
+  // same resolved-IP check before issuing the upstream request.
   const validateExternalApiBaseUrl = (baseUrl: string) => {
-    return validateBaseUrl(baseUrl);
+    return validateBaseUrlResolved(baseUrl);
   };
 
   const proxyErrorCode = (status: number) => {
@@ -702,7 +707,7 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
       );
     }
 
-    const validated = validateExternalApiBaseUrl(baseUrl);
+    const validated = await validateExternalApiBaseUrl(baseUrl);
     if (validated.error) {
       return sendApiError(
         res,
@@ -714,7 +719,7 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
 
     const url = appendVersionedApiPath(baseUrl, '/messages');
     console.log(
-      `[proxy:anthropic] ${req.method} ${validated.parsed.hostname} model=${model}`,
+      `[proxy:anthropic] ${req.method} ${validated.parsed!.hostname} model=${model}`,
     );
 
     const payload: any = {
@@ -798,7 +803,7 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
       );
     }
 
-    const validated = validateExternalApiBaseUrl(baseUrl);
+    const validated = await validateExternalApiBaseUrl(baseUrl);
     if (validated.error) {
       return sendApiError(
         res,
@@ -810,7 +815,7 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
 
     const url = appendVersionedApiPath(baseUrl, '/chat/completions');
     console.log(
-      `[proxy:openai] ${req.method} ${validated.parsed.hostname} model=${model}`,
+      `[proxy:openai] ${req.method} ${validated.parsed!.hostname} model=${model}`,
     );
 
     const payloadMessages = Array.isArray(messages) ? [...messages] : [];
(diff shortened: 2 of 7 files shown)


apps/daemon/src/connectionTest.ts
```diff
@@ -45,6 +48,68 @@ import {
...
+
+function looksLikeIpLiteral(hostname: string): boolean {
+  const host = hostname.startsWith('[') && hostname.endsWith(']')
+    ? hostname.slice(1, -1)
+    : hostname;
+  if (/^\d{1,3}(?:\.\d{1,3}){3}$/.test(host)) return true;
+  return host.includes(':');
+}
+
+export async function validateBaseUrlResolved(
```
mrcfps, line 92: Blocking: this helper is only wired into `validateExternalApiBaseUrl`, but the other daemon routes that fetch arbitrary `baseUrl`s still call the old sync validator. `testProviderConnection()` in this file still uses `validateBaseUrl(input.baseUrl)` (line 461 in the new head), and `listProviderModels()` still does the same in `apps/daemon/src/providerModels.ts:200`. That means a hostname like `internal.example.com -> 10.0.0.5` is still accepted on `POST /api/test/connection` and `POST /api/provider/models`, so the DNS-based SSRF primitive remains reachable through those endpoints even after this lands. Please switch those call sites to `await validateBaseUrlResolved(...)` (or centralize them behind one shared async validator) and add regression coverage that proves the DNS-resolved private-IP case is rejected on both routes.

mrcfps (changes requested):
@aaronjmars Thanks for tightening the proxy/finalize path here — the DNS-aware helper is the right direction. I found one remaining SSRF path that still needs this resolved-host check before the fix is complete.


lefarcen (changes requested):
Hi @aaronjmars! 👋 Thanks for the thorough security fix on the DNS-based SSRF bypass.

The `validateBaseUrlResolved` implementation is directionally correct and the comprehensive test coverage is great. However, deep review found that **this fix doesn't cover all daemon routes that accept external `baseUrl` and perform server-side fetches**.

## Remaining SSRF gaps (P1 — must fix before merge)

### 1. `/api/test/connection` (provider mode)
**File:** `apps/daemon/src/connectionTest.ts` (not in current diff)

`testProviderConnection()` still uses the sync `validateBaseUrl()` before performing a server-side fetch. A caller can supply `https://attacker.example` here, have it resolve to `10.0.0.5`, and this route will fetch it even though the proxy routes now block the same hostname.

**Fix:** Use the DNS-aware validator (`validateBaseUrlResolved`) on the daemon-side provider test path, with regression test coverage for a hostname resolving to a private IP.

---

### 2. `/api/provider/models`
**File:** `apps/daemon/src/providerModels.ts` (not in current diff)

`listProviderModels()` has the same DNS-bypass SSRF gap: it validates only the literal hostname, then fetches the provider's `/models` endpoint from the daemon. This endpoint accepts user-controlled external `baseUrl`, so it needs the resolved-IP check too — otherwise this PR fixes only some of the daemon baseUrl fetch surfaces.

**Fix:** Same as #1 — wire in `validateBaseUrlResolved`.

---

### 3. Ollama proxy redirect bypass
**File:** `apps/daemon/src/server.ts` around line 8020 (not in current diff)

The Ollama proxy fetch is the only proxy fetch in the codebase that still lacks `redirect: 'error'`. Even after DNS validation, a public validated host can redirect the daemon to a private/internal URL and bypass the resolved-IP check.

**Fix:** Add `redirect: 'error'` to the Ollama proxy `fetch()` call to align with the other proxy routes, so redirects cannot become an SSRF escape hatch.

---

## Summary

Once these three gaps are closed, this will be a solid defense-in-depth fix. The threat model is clear, the DNS resolution approach is sound, and the test coverage for the core function is comprehensive — it just needs to cover all the daemon's external-baseUrl fetch surfaces consistently.

Looking forward to the next iteration!

lefarcen:
Agent-to-agent review note for PR #1176:

Status: request_changes.

Lens A findings:

- `apps/daemon/src/connectionTest.ts:526`: `/api/test/connection` provider mode still uses sync `validateBaseUrl()` before a daemon-side provider fetch. A hostname that resolves to a private/internal IP can still be fetched through this route. Please switch provider connection tests to the DNS-aware validator and add regression coverage.
- `apps/daemon/src/providerModels.ts:200`: `/api/provider/models` has the same remaining DNS-bypass SSRF gap. It accepts user-controlled external `baseUrl`, validates only the literal hostname, then fetches provider models from the daemon. This needs resolved-IP validation too.
- `apps/daemon/src/server.ts:8020`: the Ollama proxy fetch still lacks `redirect: 'error'`. A validated public host can redirect the daemon to a private/internal URL, bypassing the pre-fetch resolved-IP check. Align this route with the other proxy fetches.

Lens B finding:

- PR body threat model should include the other daemon `baseUrl` fetch surfaces (`/api/test/connection` provider mode and `/api/provider/models`) or explicitly explain why they are out of scope after code changes. It should also call out redirect-follow behavior as a separate SSRF bypass class unless the Ollama route is updated to reject redirects.

Overall: the resolved-IP wrapper is directionally correct for the proxy/finalize paths it touches, and the DNS-rebinding limitation is acknowledged. The security fix is incomplete until adjacent daemon fetch surfaces and redirect behavior are handled.

aaronjmars:
Pushed `722f0b9` to address all three gaps + the Lens B threat-model note.

**Changes:**

1. **`testProviderConnection` (`/api/test/connection` provider mode)** — switched the sync `validateBaseUrl` call at `connectionTest.ts:526` to `await validateBaseUrlResolved`. Provider-mode probes now run the same resolved-IP check the proxy/finalize paths do, before posting the smoke prompt upstream.

2. **`listProviderModels` (`/api/provider/models`)** — same swap at `providerModels.ts:200`. Imported `validateBaseUrlResolved` from `./connectionTest.js` rather than `@open-design/contracts` so `packages/contracts` stays pure (no `node:dns` dep).

3. **Ollama proxy** — added `redirect: 'error'` to the `fetch` call at `server.ts:8020`. Matches the four other proxy fetches; a validated public host can no longer 3xx the daemon to a private/internal URL after the pre-fetch resolved-IP check.

**Regression coverage** (`apps/daemon/tests/connection-test.test.ts` + `apps/daemon/tests/proxy-routes.test.ts`):

- `POST /api/provider/models` rejects a hostname whose `dnsPromises.lookup` returns `10.0.0.5`, with no upstream fetch.
- `POST /api/test/connection` provider mode: same shape, asserts `{ ok: false, kind: 'forbidden' }`.
- `/api/proxy/ollama/stream`: asserts the upstream fetch init carries `redirect: 'error'`.

The existing `provider/models` probe-timer test also gained a `dnsPromises.lookup` spy — without my change it was fully synchronous before the timer was armed; now there's a real DNS call that would race `vi.advanceTimersByTimeAsync`. Spy returns a public IP instantly so the timeout assertion still fires deterministically.

Local: `pnpm typecheck` ✅ · `pnpm guard` ✅ · `pnpm vitest run tests/connection-test.test.ts tests/proxy-routes.test.ts` → 93/93 ✅.

DNS-rebinding (resolve-then-flip) is still out of scope for this PR per the original "Known limitation" — happy to follow up with a custom Undici dispatcher that pins the resolved IP through to the actual socket.

lefarcen (dismissed):
All three P1 gaps fixed in 722f0b9. The DNS-aware validator is now wired into both `/api/test/connection` (provider mode) and `/api/provider/models`, and the Ollama proxy fetch now includes `redirect: 'error'`. Regression coverage is solid.

**Note:** This is still a preflight DNS check, not fetch-time address pinning. It closes the static DNS-bypass SSRF class you reported, but active DNS rebinding (hostname resolves public at validation time, then flips to private at fetch time) remains out of scope for this PR. That would require a custom Undici dispatcher to pin the resolved IP — happy to review that as a follow-up if you'd like to tackle it.

Thanks for the thorough fix and the clear commit message. 🎉

lefarcen (commented):
Hey @aaronjmars, thanks for the quick follow-up on this security patch. The previously flagged SSRF gaps still look covered on this head: provider connection tests and provider model listing now use the DNS-aware validator, the finalize/proxy path stays async, and the Ollama proxy keeps `redirect: 'error'`.

I also checked the new route split into `chat-routes.ts` and didn’t see it reintroduce the literal-hostname-only validator on the proxy streams. The added regression coverage is pointed at the right surfaces.

No new blockers from my pass on `3912fafa`. We’ll still need the stale mrcfps changes-requested review to be refreshed before merge can proceed, but this iteration looks clean from the bot review side. 🙌

aaronjmars:
@mrcfps gentle ping — your earlier feedback has been addressed in subsequent commits and the bot reviewer dismissed its blocking review. Could you take another look and refresh your changes-requested review when you have a moment? Happy to rebase the merge conflict afterward. Thanks!

lefarcen (commented):
Hey @aaronjmars — the security write-up is already very clear, especially the DNS-based SSRF impact and the validation notes. One small PR-body cleanup before the next maintainer pass: could you add the missing `Surface area` checklist info and an explicit bug-fix verification note?

No need to rename your existing `Summary` or `Verification` sections — those already cover the why/what and validation. For this PR, the useful additions would be something like: daemon API/proxy/finalize/provider-model routes under Surface area, plus the red→green cases for resolved-private-IP rejection and `redirect: 'error'` under bug-fix verification.

mrcfps:
@aaronjmars took another pass. The original SSRF gaps I flagged look addressed on the current head — thanks for tightening the provider connection test path, provider models path, and Ollama redirect handling.

I still agree with the small PR-body cleanup request from @lefarcen around the Surface area checklist and an explicit bug-fix verification note. Once that is updated, I do not expect further blockers from my side and can refresh the stale review.

mrcfps (dismissed):
Rechecked the current head. The original SSRF gaps from my earlier review look addressed: provider connection tests and provider model listing now use the DNS-aware validator, and the Ollama proxy rejects redirects.

I still agree with the earlier PR-body cleanup note around Surface area and explicit bug-fix verification, but I do not consider that a merge blocker for this fix.

aaronjmars:
@mrcfps quick re-ping — the PR-body polish (Surface area checklist + bug-fix verification red→green table) is in, and the branch has been rebased onto current main to clear the merge conflict. The force-push appears to have dismissed your earlier approval. When you have a moment, could you take another look at the new head and re-approve if it still looks good? Thanks!

mrcfps (approved):

lefarcen:
All gates are green now — flagged this for the maintainers. Thanks @aaronjmars for the security fix and @mrcfps for the review.

leilei926524-tech:
Hi @aaronjmars — quick note from the Open Design team.

What you get there:

PR #1195, opened 2026-05-11
feat: add validation and error feedback for invalid file paths

### LINKED ISSUE #1186: Invalid file path creates a design project without any error feedback
## Summary

When the file path input contains an invalid path, the app still creates a design project instead of showing an error.

## Current behavior

If the user enters an incorrect file path and proceeds, no validation error or feedback is shown. The app directly creates a design project even though the path is invalid.

## Expected behavior

The app should validate the file path before project creation and show a clear error message when the path is invalid or cannot be resolved.

## Steps to reproduce

1. Open the project creation panel.
2. Enter an invalid file path in the path input.
3. Trigger project creation.
4. Observe the result.

## Notes

- A screenshot was provided showing the path field populated with `/`.
- The issue is that project creation proceeds silently instead of blocking with feedback.
- The missing feedback makes it unclear whether the path was accepted, ignored, or failed.

## Why this matters

Creating a project from an invalid path without warning can confuse users, create unintended empty projects, and make the import flow feel unreliable.

Image

xxiaoxiong:
Fixes #1186

## Problem
When the user enters an invalid file path in the project creation panel, the app proceeds without validation or error feedback. This creates confusion and potentially empty projects.

## Solution
Add path validation and error handling:
1. Validate that path is absolute (starts with /)
2. Show clear error message for invalid paths
3. Catch API failures and display error details
4. Prevent project creation until path is valid

## Changes
1. **NewProjectPanel.tsx**:
   - Add path validation before import
   - Check if path starts with '/'
   - Set importFolderError for invalid paths
   - Wrap onImportFolder in try-catch
   - Display error message to user
   - Clear error on valid path

2. **App.tsx**:
   - Modify handleImportFolder to throw error when API returns null
   - Provide clear error message about invalid/inaccessible path
   - Error propagates to NewProjectPanel for display

## User Experience
**Before:**
Invalid path → [no feedback] → empty project created

**After:**
Invalid path → [error message displayed] → project creation blocked

## Error Messages
- "Invalid path" - Path must be absolute (start with /)
- "Failed to import folder" - API error or inaccessible path
- "Could not import folder. The path may be invalid or inaccessible."

## Testing
1. Open project creation panel
2. Enter invalid paths:
   - Relative path: "folder/path"
   - Just "/"
   - Non-existent path: "/invalid/path"
3. Click "Open folder"
4. Verify error messages appear
5. Enter valid path and verify it works

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -603,7 +603,9 @@ export function App() {
 
   const handleImportFolder = useCallback(async (baseDir: string) => {
     const result = await importFolderProject({ baseDir });
-    if (!result) return;
+    if (!result) {
+      throw new Error('Could not import folder. The path may be invalid or inaccessible.');
+    }
     setProjects((curr) => [result.project, ...curr.filter((p) => p.id !== result.project.id)]);
     navigate({
       kind: 'project',
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -468,9 +468,25 @@ export function NewProjectPanel({
     if (!onImportFolder) return;
     const trimmed = baseDir.trim();
     if (!trimmed) return;
+    
+    // Validate path before attempting import
+    if (!trimmed.startsWith('/')) {
+      setImportFolderError({
+        message: 'Invalid path',
+        details: 'Path must be an absolute path starting with /',
+      });
+      return;
+    }
+    
+    setImportFolderError(null);
     setImportingFolder(true);
     try {
       await onImportFolder(trimmed);
+    } catch (err) {
+      setImportFolderError({
+        message: 'Failed to import folder',
+        details: err instanceof Error ? err.message : 'Unknown error',
+      });
     } finally {
       setImportingFolder(false);
     }


apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -468,9 +468,25 @@ export function NewProjectPanel({
     if (!onImportFolder) return;
     const trimmed = baseDir.trim();
     if (!trimmed) return;
+    
+    // Validate path before attempting import
+    if (!trimmed.startsWith('/')) {
```
lefarcen, line 473: **P1** — The issue's `/` repro still passes this check.

`/` starts with `/` so it's treated as valid, but the daemon will accept it as an absolute directory and may create a project rooted at the filesystem root.

Suggestion: Reject normalized filesystem roots in `POST /api/import/folder` (where you have real `fs` access), and add a regression test proving `/` is rejected.

lefarcen, line 473: **P2** — This browser-side check assumes POSIX absolute paths.

Windows users with valid paths like `C:\\Users\\me\\project` would be rejected even though the daemon's `path.isAbsolute()` is platform-aware.

Suggestion: Either move validation to the daemon (where platform rules are accurate) or have the daemon send platform hints to the frontend.

apps/web/src/App.tsx
```diff
@@ -603,7 +603,9 @@ export function App() {
 
   const handleImportFolder = useCallback(async (baseDir: string) => {
     const result = await importFolderProject({ baseDir });
-    if (!result) return;
+    if (!result) {
+      throw new Error('Could not import folder. The path may be invalid or inaccessible.');
```
lefarcen, line 607: **P2** — `importFolderProject` collapses all failures to `null`, so this throw loses daemon error detail.

Users get the same generic message for "folder not found", "path must be a directory", auth failures, and data-dir rejection.

Suggestion: Parse and return the API error body from `POST /api/import/folder` and display its `message`/`details` fields so users know what went wrong.

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -468,9 +468,25 @@ export function NewProjectPanel({
...
+      setImportFolderError({
+        message: 'Invalid path',
+        details: 'Path must be an absolute path starting with /',
+      });
+      return;
+    }
+    
+    setImportFolderError(null);
     setImportingFolder(true);
     try {
```
lefarcen, line 483: **P2** — No regression coverage for the new error behavior.

Suggestion: Add tests proving:
- Relative paths are blocked without calling `onImportFolder`
- `/` is rejected (after the root fix)
- Rejected import calls render an error without navigating or creating a project

lefarcen (dismissed):
Hi @xxiaoxiong! Thanks for tackling #1186 🙏

The path validation improves one silent failure path, but there are a few issues that need addressing before merge:

**Main concern:** The issue's concrete `/` repro still passes through (it starts with `/` so it's considered valid by this check). The daemon will accept `/` as a valid absolute directory and may create a project rooted at the filesystem root.

**Cross-platform:** The browser-side `startsWith('/')` check assumes POSIX paths. Windows users with valid paths like `C:\Users\me\project` would be blocked even though the daemon's validation is platform-aware.

**Error detail:** The generic "path may be invalid or inaccessible" message loses the specific daemon error (folder not found vs not a directory vs permission denied).

See inline comments for specifics + suggested fixes. Looking forward to the next iteration!

apps/web/src/components/NewProjectPanel.tsx
mrcfps, line 473: This client-side `startsWith('/')` check does not actually enforce the invalid-path cases this PR is trying to fix. It still accepts `/`, and the current daemon import route at `apps/daemon/src/server.ts:2925-2950` only checks `path.isAbsolute(...)` + `realpath()` + `lstat()`, so the original `/` repro would still import the filesystem root instead of showing an error. The same prefix check also rejects valid Windows absolute paths such as `C:\Users\me\project`, which would break manual folder import on packaged Windows builds. Please move the absolute-path validation to the daemon (where the platform rules are authoritative), explicitly reject normalized filesystem roots there, and add a regression test that covers both `/` and a Windows absolute path.

mrcfps (changes requested):
@xxiaoxiong Thanks for working on this fix 🙏 I rechecked the changed ranges and still see one blocking issue in the new manual-path validation flow; details are inline.


lefarcen:
Update: mrcfps's review covers the same core issues I raised (filesystem root validation + Windows path compatibility). Please address those blockers — I've dismissed my review to avoid duplicate asks. Thanks!

lefarcen:
@xxiaoxiong Thanks for this PR! 🙏

The filesystem root validation you were working on here landed via #1266 (merged 9 hours ago). That PR implements the daemon-level fix that mrcfps requested in the blocking review — it rejects filesystem roots like `/` at the authoritative validation layer and includes test coverage.

Since the core issue is now resolved, I'm closing this as superseded. If you think there's still UX value in the frontend validation layer (early error feedback before the daemon call), feel free to reopen with that narrower scope.

Thanks for working on this! 🎉

PR #1203, opened 2026-05-11
fix(desktop): enforce minimum window size on main client (#1189)

### LINKED ISSUE #1189: Limit the minimum resizable window size in the desktop client
## Summary

The desktop client can currently be resized too small, which breaks the layout and makes the UI difficult to use.

## Current behavior

Users can shrink the app window to a very small size. At that point, parts of the interface become cramped, clipped, or visually broken.

## Expected behavior

The desktop client should enforce a reasonable minimum window size so the layout remains usable and core navigation stays visible.

## Steps to reproduce

1. Open the desktop client.
2. Resize the app window smaller and smaller.
3. Observe the layout once the window becomes very narrow.

## Notes

- A screenshot was provided showing the window reduced to a size where the top navigation is already heavily compressed.
- This appears to be a desktop window constraint issue rather than a component-specific layout bug.
- The request is to add a minimum resizable size limit for the client window.

## Why this matters

Allowing the window to shrink below the UI's usable breakpoint creates a broken first impression and makes the app feel unstable. A minimum window size would prevent avoidable layout failures.

Image

Sid-Qin:
## Summary

Closes #1189. The desktop main `BrowserWindow` in `apps/desktop/src/main/runtime.ts` shipped only `width: 1280, height: 900` with no `minWidth` / `minHeight`, so Electron honored arbitrary user drags. Past roughly 900×600 the project page's left/right split (chat composer + designs panel + preview pane) overlaps and the top navigation clips — the broken first impression the issue reports.

## Approach

Add `minWidth: 900, minHeight: 600` to the main window's `BrowserWindow` constructor. Values come from the maintainer's recommendation in the issue thread:

> 建议先加一个保守的最小尺寸，例如 `900x600`：这个宽度能覆盖当前项目页左右分栏的基本布局需求，同时不会影响常见小屏笔记本使用。

The print sub-window at `runtime.ts:809-818` is intentionally left untouched: it's created with `show: false`, populated programmatically, shown only during the print round-trip, then closed in `finally`. The user can't drag it to resize, so a min-size floor has no observable effect.

## Reproduce

Before this PR:

1. Build/run the desktop client.
2. Drag any window edge to make the window very narrow / short.
3. Observe top navigation clipping and the project-page layout overlapping (per the screenshot in #1189).

After this PR:

1. Same steps.
2. The window stops shrinking once any edge reaches the 900-wide / 600-tall floor; the chat + designs + preview split stays usable, top navigation stays visible.

## Verification

- `pnpm --filter @open-design/desktop typecheck` — clean (`tsc --noEmit` exit 0)
- `pnpm guard` — all 5 invariant checks pass (residual JS, test layout, e2e layout, web test layout, tools layout)
- Diff is 6 lines (4 lines of comment explaining the floor's rationale + the two new `min*` props); no other behavior touched

## Test plan

- [ ] Run the desktop client locally (`pnpm tools-dev` then the desktop dev launch path)
- [ ] Confirm the main window cannot be dragged below 900×600
- [ ] Confirm the print export flow (Cmd/Ctrl-P from a design) still works — print window is unaffected but worth a smoke-check
- [ ] If any specific page still looks visually cramped at exactly 900×600, capture a screenshot and we can iterate on that page's responsive layout separately (per @lefarcen's note in the issue thread)

### DIFF
--- a/apps/desktop/src/main/runtime.ts
+++ b/apps/desktop/src/main/runtime.ts
@@ -784,6 +784,12 @@ export async function createDesktopRuntime(options: DesktopRuntimeOptions): Prom
   const consoleEntries: DesktopConsoleEntry[] = [];
   const window = new BrowserWindow({
     height: 900,
+    // Below this size the project page's left/right split (chat
+    // composer + designs panel + preview pane) overlaps and the top
+    // navigation clips, so prevent Electron from honoring user drags
+    // that would shrink the window past the usable breakpoint.
+    minHeight: 600,
+    minWidth: 900,
     show: true,
     title: "Open Design",
     ...MAC_WINDOW_CHROME,


lefarcen (commented):
Hi @Sid-Qin! 👋

Thanks for this PR — clean scoped fix for the desktop window resizing issue.

The change is straightforward: adds `minWidth: 900, minHeight: 600` to the main `BrowserWindow` constructor in `apps/desktop/src/main/runtime.ts`, exactly as the maintainer recommended in issue #1189. The 4-line explanatory comment is helpful, and leaving the print sub-window untouched makes sense (it's `show: false` and never user-draggable).

I reviewed the full desktop runtime context and found no correctness or security concerns:
- No IPC, preload, sandbox, navigation, or permission-surface changes
- Consistent with existing Electron patterns in the codebase
- The maintainer-vetted 900×600 floor covers the project page's split layout needs

The test plan is appropriate — manual desktop smoke-check is the right verification for this declarative Electron option (automated tests for BrowserWindow dimensions would be more fragile than valuable).

**Looks good to me; deferring final approval to a maintainer.**

— open-design team

mrcfps (approved):
@Sid-Qin I reviewed the runtime change in `apps/desktop/src/main/runtime.ts` and verified the new `minWidth: 900` / `minHeight: 600` floor only affects the main Electron window while leaving the print flow and the rest of the runtime behavior untouched. Thanks for shipping a small, focused fix here — the change matches the issue context well and the added rationale comment makes the constraint easy to understand. ✨


lefarcen:

PR #1206, opened 2026-05-11
fix: pretty-print JSON file previews

Kaelz31:
## Summary
- Pretty-print valid JSON files in the FileViewer text display.
- Fall back to raw text when JSON parsing fails.
- Preserve existing copy behavior so Copy still uses the raw file text.

## Tests
- `corepack pnpm --dir apps/web exec vitest run -c vitest.config.ts tests/components/FileViewer.test.tsx`

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -5679,6 +5679,7 @@ function formatJsonFileTextForDisplay(file: ProjectFile, text: string): string {
 function hasPrecisionSensitiveJsonNumberText(text: string): boolean {
   let inString = false;
   let escaped = false;
+  const numberTokenPattern = /-?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+-]?\d+)?/y;
   for (let i = 0; i < text.length;) {
     const char = text[i];
     if (inString) {
@@ -5699,7 +5700,8 @@ function hasPrecisionSensitiveJsonNumberText(text: string): boolean {
       continue;
     }
 
-    const match = /^-?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+-]?\d+)?/.exec(text.slice(i));
+    numberTokenPattern.lastIndex = i;
+    const match = numberTokenPattern.exec(text);
     if (!match) {
       i += 1;
       continue;
@@ -5708,7 +5710,7 @@ function hasPrecisionSensitiveJsonNumberText(text: string): boolean {
     const token = match[0];
     if (isSignedNegativeZeroJsonNumberToken(token)) return true;
     if (/[.eE]/.test(token) && isPrecisionSensitiveJsonNumberToken(token)) return true;
-    i += token.length;
+    i = numberTokenPattern.lastIndex;
   }
   return false;
 }
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -5667,10 +5667,21 @@ function TextViewer({
 function formatJsonFileTextForDisplay(file: ProjectFile, text: string): string {
   if (!isJsonFile(file)) return text;
   try {
-    return JSON.stringify(JSON.parse(text), null, 2);
+    const parsed = JSON.parse(text) as unknown;
+    if (hasUnsafeJsonNumber(parsed)) return text;
+    return JSON.stringify(parsed, null, 2);
   } catch {
     return text;
   }
+}
+
+function hasUnsafeJsonNumber(value: unknown): boolean {
+  if (typeof value === 'number') {
+    return !Number.isFinite(value) || (Number.isInteger(value) && !Number.isSafeInteger(value));
+  }
+  if (Array.isArray(value)) return value.some(hasUnsafeJsonNumber);
+  if (value && typeof value === 'object') return Object.values(value).some(hasUnsafeJsonNumber);
+  return false;
 }
 
 function isJsonFile(file: ProjectFile): boolean {
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -5611,7 +5611,11 @@ function TextViewer({
     }
   }
 
-  const lineCount = text ? text.split('\n').length : 0;
+  const displayText = useMemo(
+    () => (text == null ? null : formatJsonFileTextForDisplay(file, text)),
+    [file.name, file.mime, text],
+  );
+  const lineCount = displayText ? displayText.split('\n').length : 0;
 
   return (
     <div className="viewer text-viewer">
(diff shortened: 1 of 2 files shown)


apps/web/src/components/FileViewer.tsx
```diff
@@ -5650,16 +5654,29 @@ function TextViewer({
...
         )}
       </div>
     </div>
   );
 }
 
+function formatJsonFileTextForDisplay(file: ProjectFile, text: string): string {
+  if (!isJsonFile(file)) return text;
+  try {
+    return JSON.stringify(JSON.parse(text), null, 2);
```
mrcfps: `formatJsonFileTextForDisplay` currently pretty-prints by round-tripping through `JSON.parse`/`JSON.stringify`, which changes some valid JSON payloads before we render them. In particular, integers/decimals outside JavaScript's safe numeric range are coerced to `Number`, so a file like `{"id":9007199254740993}` will preview as `9007199254740992`. That makes the viewer display different content than the file on disk, which is a correctness issue for IDs, timestamps, and other precision-sensitive data. Can we switch this branch to a formatter that preserves original numeric lexemes, or at least fall back to the raw text whenever the parsed value would contain unsafe numbers? I'd also add a regression in `apps/web/tests/components/FileViewer.test.tsx` that covers an unsafe integer so this doesn't silently come back.

mrcfps (commented):
Thanks for the cleanup here — the preview formatting path and the raw-copy preservation both make sense. I found one merge-safe correctness gap to address before this is fully robust for all valid JSON inputs.

Kaelz31:
Thanks for the catch. I updated the PR so JSON with unsafe numbers falls back to raw text, and added a regression test for that case.

apps/web/src/components/FileViewer.tsx
```diff
@@ -5650,16 +5654,40 @@ function TextViewer({
...
+    if (hasUnsafeJsonNumber(parsed)) return text;
+    return JSON.stringify(parsed, null, 2);
+  } catch {
+    return text;
+  }
+}
+
+function hasUnsafeJsonNumber(value: unknown): boolean {
+  if (typeof value === 'number') {
+    return !Number.isFinite(value) || (Number.isInteger(value) && !Number.isSafeInteger(value));
```
mrcfps, line 5759: `hasUnsafeJsonNumber` now catches unsafe integers, but this branch still returns `false` for precision-losing decimal/exponent literals because they parse as finite non-integer `number`s. For example, `JSON.parse('{"ratio":0.1234567890123456789}')` round-trips back out as `0.12345678901234568`, so the preview can still diverge from the bytes on disk even after the new fallback. Since this helper is the gate for whether pretty-printing is lossless, can we extend it to reject numbers that do not round-trip exactly (or conservatively fall back for decimal/exponent forms) and add a regression in `apps/web/tests/components/FileViewer.test.tsx` for a high-precision decimal?

mrcfps (commented):
Thanks for the follow-up update here — the unsafe-integer fallback closes the earlier main regression. I found one remaining merge-safe precision case in the new guard that still looks worth tightening up.

Kaelz31:
Updated again. The preview now detects precision-sensitive JSON number
tokens before pretty-printing, and falls back to raw text for unsafe
integers, high-precision decimals, and precision-loss exponent cases. Safe
JSON still pretty-prints normally.

Kaelz31:
Updated again. The preview now detects precision-sensitive JSON number tokens before pretty-printing, and falls back to raw text for unsafe integers, high-precision decimals, and precision-loss exponent cases. Safe JSON still pretty-prints normally.

apps/web/src/components/FileViewer.tsx
```diff
@@ -5650,16 +5654,112 @@ function TextViewer({
...
+  const match = /^(-)?(\d+)(?:\.(\d+))?(?:[eE]([+-]?\d+))?$/.exec(token);
+  if (!match) return null;
+  const [, sign, integerPart, fractionPart = '', exponentPart = '0'] = match;
+  const coefficient = BigInt(`${sign ?? ''}${integerPart}${fractionPart}`);
+  const exponent = Number(exponentPart) - fractionPart.length;
+  return normalizeDecimalParts(coefficient, exponent);
+}
+
+function normalizeDecimalParts(coefficient: bigint, exponent: number): { coefficient: bigint; exponent: number } {
+  if (coefficient === 0n) return { coefficient: 0n, exponent: 0 };
```
mrcfps, line 5747: `normalizeDecimalParts` collapses both `-0` and `0` into the same normalized value here, so the new precision guard still treats a payload like `{"delta":-0}` as safe and then re-renders it as `{"delta":0}` after the `JSON.parse`/`JSON.stringify` round trip. That means the preview can still differ from the file on disk for one more valid JSON number form, and `-0` is semantically distinct in JavaScript (`Object.is(-0, 0) === false`). Can we preserve the zero sign in this normalization path, or explicitly short-circuit `-0` before pretty-printing, and add a regression in `apps/web/tests/components/FileViewer.test.tsx` for a signed-zero payload?

mrcfps (commented):
Thanks for the follow-up here — the new token scan closes the earlier decimal/exponent precision gap. I found one remaining merge-safe correctness case around signed zero that still looks worth tightening up.

Kaelz31:
Updated again. Signed negative zero JSON number tokens like -0 now fall back to raw text before pretty-printing, with a regression test added.

apps/web/src/components/FileViewer.tsx
```diff
@@ -5650,16 +5654,117 @@ function TextViewer({
...
+      continue;
+    }
+
+    if (char === '"') {
+      inString = true;
+      i += 1;
+      continue;
+    }
+
+    const match = /^-?(?:0|[1-9]\d*)(?:\.\d+)?(?:[eE][+-]?\d+)?/.exec(text.slice(i));
```
mrcfps: `hasPrecisionSensitiveJsonNumberText` now walks the entire file and runs `/^.../.exec(text.slice(i))` at every character outside strings. Because `text.slice(i)` allocates a new suffix on each iteration, this turns the precision check into O(n²) work for large JSON payloads, and `formatJsonFileTextForDisplay` runs that scan during render via `useMemo`. On multi-megabyte artifacts this can make opening the viewer noticeably slow or freeze the tab before we even reach `JSON.parse`. Can we switch this loop to an index-based scanner that advances without slicing the remaining string each time (for example, inspect the current char first and only parse a token when it starts with `-` or a digit, or use a sticky regex over the original string)? That keeps the fallback logic linear while preserving the same precision checks.

mrcfps (commented):
Thanks for the follow-up here — the new fallbacks cover the earlier JSON precision issues well. I found one merge-safe follow-up around the runtime cost of the new token scanner.

Kaelz31:
Updated. The JSON number scanner no longer uses repeated text.slice(i) in the loop; it now scans in place with a sticky regex/lastIndex while preserving the same precision fallback behavior.

mrcfps (approved):
@Kaelz31 I reviewed the latest JSON preview changes and verified that valid JSON still pretty-prints while unsafe integers, precision-losing decimals/exponents, and signed negative zero all fall back to the raw file text. The regression coverage lines up with those cases well — thanks for iterating on this so thoroughly.

lefarcen:
Merged! 🎉 Thanks @Kaelz31 for the contribution. Looking forward to the next one!

leilei926524-tech:
Hi @Kaelz31 — welcome to Open Design contributors. 🎉

Your first merged PR (#1206) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #1207, opened 2026-05-11
fix(contracts): pin API-mode override above discovery layer (#313)

### LINKED ISSUE #313: bug: Anthropic API mode renders TodoWrite/Read progress as raw text instead of tool UI
### Summary

In Anthropic API mode, once the assistant starts actual work, progress/tool activity is rendered as raw text in the chat stream instead of the intended Open Design tool UI cards.

This appears related to, but broader than, #312. The first issue covers `question-form` raw structured output. This one covers the execution phase: TodoWrite / Read-like progress is also exposed as plain text.

### Environment

- Open Design version: 0.2.0 / latest `main`
- Mode/provider: Anthropic API / Claude provider
- Observed in the web/desktop chat UI
- Prompt language: Chinese

### What I saw

After submitting the form answers, the chat shows text such as:

```text
[form answers — direction]
- 方向: warm-soft
- 主色覆盖（可选）: (skipped)
```

Then, when the assistant starts working, it shows raw progress/protocol-like content directly in the chat:

```text
<todo-list>
 • [in_progress] 1. 读取 skill 资源 (template.html, layouts.md, checklist.md)
 • [pending] 2. 绑定 warm-soft 配色到 :root + 融合 Notion DS 的字号体系
 • [pending] 3. 规划 7 个章节的叙事节奏 ...
 ...
</todo-list>

让我先读 skill 资源。

[读取 template.html / layouts.md / checklist.md ...]
```

Expected behavior is that TodoWrite / Read / Write / Edit activity should be rendered as Open Design tool UI cards, not as plain assistant prose.

### Expected behavior

- Form answers should ideally render as a compact submitted-answer summary rather than raw `[form answers — ...]` text.
- TodoWrite progress should render as the existing Todos card.
- Read / Write / Edit actions should render as the existing tool cards.
- The assistant should not expose pseudo-tool markup such as `<todo-list>` or `[读取 ...]` in the chat stream.

### Source-level observation

From a quick source read, this looks like the Anthropic API path is currently streaming plain text instead of producing real agent/tool events:

- `apps/web/src/components/ProjectView.tsx`: in API mode, the branch calls `streamMessage(...)` and converts each delta into a text event rather than receiving `tool_use` events.
- `apps/web/src/providers/anthropic.ts`: the direct Anthropic SDK call sends `model`, `max_tokens`, `system`, and `messages`, but does not appear to pass tool definitions or run an agent/tool loop.
- `apps/web/src/components/ToolCard.tsx` and `apps/web/src/runtime/todos.ts`: the UI already knows how to render `TodoWrite`, `Read`, `Write`, etc. when they arrive as `tool_use` events.
- `docs/agent-adapters.md` describes the API fallback as owning a minimal tool loop with Read/Write/Edit implemented by the daemon, but the current UI path appears to be a plain Messages API stream.

So Claude is being prompted to use `TodoWrite` / Read-like workflow, but in Anthropic API mode it may not actually have real tools available. As a result, it narrates or emits pseudo-tool/protocol text, and the UI has no structured event to render.

### Possible fix direction

Either:

1. Implement the intended Anthropic API fallback tool loop so Read/Write/Edit/TodoWrite become structured `tool_use` / `tool_result` events that the existing UI can render, or
2. Adjust the Anthropic API prompt/UI fallback so it does not ask the model to use unavailable tools and instead shows a clean progress placeholder/state.

The best user experience would probably be both: real tool events where possible, plus a skeleton/placeholder fallback so raw protocol-like content is never exposed.

mutoe:
Fixes #313.

## Why

`DISCOVERY_AND_PHILOSOPHY` is pinned at the TOP of the composed prompt with its own "read first — these override anything later" header, and it mandates `TodoWrite` on turn 3, brand-spec extraction via `Bash` + `Read` + `WebFetch`, etc. In daemon/CLI mode those tools are wired through and produce real `tool_use` events the UI renders as ToolCards.

In Anthropic API mode (`config.mode === 'api'`, `streamFormat: 'plain'`) **no tools are passed to the Messages API**. `providers/anthropic.ts` only does a plain text stream. So the agent, still following the discovery layer's hard rules, narrates pseudo-tool markup directly into the chat:

```
<todo-list>
 • [in_progress] 1. 读取 skill 资源 ...
 ...
</todo-list>

让我先读 skill 资源。
[读取 template.html / layouts.md / checklist.md ...]
```

The old `streamFormat === 'plain'` rule was appended at the BOTTOM of the prompt — strictly lower precedence than the discovery layer — so it was load-bearing-by-position-only and didn't actually suppress the pseudo-tool output.

## What

Plan A from https://github.com/nexu-io/open-design/issues/313#issuecomment-4417069925:

1. Pin a new `API_MODE_OVERRIDE` section at the absolute TOP of the composed prompt (above `DISCOVERY_AND_PHILOSOPHY`) when `streamFormat === 'plain'`. Highest precedence beats the discovery layer's own override claim.
2. Explicitly name every tool the agent must not pretend to call: `TodoWrite`, `Read`, `Write`, `Edit`, `Bash`, `WebFetch`.
3. Forbid the exact pseudo-tool / fake-protocol markup observed in #313: `<todo-list>`, `<tool-call>`, `[读取 X]`, `[正在调用 ...]`.
4. Tell the agent to state its plan as prose (markdown lists OK), keep `<artifact>` HTML output, keep `<question-form>` discovery markup. Those are markup the UI parses, not tool calls.
5. Drop the redundant trailing "API mode rule" section — replaced by the top-anchored override.

## Scope

Plan A (light, prompt-only) is intentional. Plan B (full tool loop in `providers/anthropic.ts`) is deferred — it requires an architectural decision on whether API mode is allowed to depend on the daemon for `Read`/`Write`/`Edit`, which conflicts with the "API mode is daemon-independent" property. If users hit capability ceilings after this fix lands, a tracking issue can revisit B.

## Tests

`packages/contracts/tests/system-prompt-api-mode.test.ts` — 9 tests, TDD red-then-green:

- daemon mode (`{}`): `TodoWrite` still present (control); no API-mode preamble
- API mode (`{ streamFormat: 'plain' }`):
  - preamble injected
  - **positioned above `# OD core directives`** (the discovery layer header)
  - names all 6 unavailable tools
  - includes the forbidden `<todo-list>` and `[读取` strings as cautionary examples
  - instructs prose-based planning
  - explicitly invalidates later "call TodoWrite / use tool" instructions
  - keeps `<artifact>` allowed

All 17 contracts tests pass. Daemon prompt-related tests (39) pass unchanged. Unrelated pre-existing typecheck and test failures on `main` (sidecar/server.ts type errors, finalize-design / desktop-import-token-gate / project-file-rename / chat-route / project-watchers flakes) are untouched and not introduced by this PR.

## How to verify locally

```bash
pnpm --filter @open-design/contracts test
pnpm --filter @open-design/contracts typecheck
```

Then in the web UI, switch to API/BYOK mode and send any design brief — the assistant should respond with prose + a discovery `<question-form>` instead of `<todo-list>` markup.

### DIFF
--- a/packages/contracts/src/prompts/system.ts
+++ b/packages/contracts/src/prompts/system.ts
@@ -79,11 +79,26 @@ export function composeSystemPrompt({
   // turn 1", "branch on brand on turn 2", "TodoWrite on turn 3", run
   // checklist + critique before <artifact>) win precedence over softer
   // wording later in the official base prompt.
-  const parts: string[] = [
+  const parts: string[] = [];
+
+  // API/BYOK mode (streamFormat === 'plain'): no tools are wired through
+  // to the model, but the discovery layer + base prompt below still tell
+  // it to call TodoWrite/Read/Write/Edit/Bash/WebFetch. Without an
+  // explicit top-anchored override, the model invents pseudo-tool markup
+  // (`<todo-list>`, `[读取 X]`) instead of producing real progress
+  // events — see #313. Pin this preamble ABOVE DISCOVERY_AND_PHILOSOPHY
+  // so it beats the discovery layer's own "these override anything
+  // later" header.
+  if (streamFormat === 'plain') {
+    parts.push(API_MODE_OVERRIDE);
+    parts.push('\n\n---\n\n');
+  }
+
+  parts.push(
     DISCOVERY_AND_PHILOSOPHY,
     '\n\n---\n\n# Identity and workflow charter (background)\n\n',
     BASE_SYSTEM_PROMPT,
-  ];
+  );
 
   if (designSystemBody && designSystemBody.trim().length > 0) {
     parts.push(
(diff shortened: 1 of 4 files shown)


packages/contracts/src/prompts/system.ts
```diff
@@ -79,11 +79,26 @@ export function composeSystemPrompt({
...
+
+  // API/BYOK mode (streamFormat === 'plain'): no tools are wired through
+  // to the model, but the discovery layer + base prompt below still tell
+  // it to call TodoWrite/Read/Write/Edit/Bash/WebFetch. Without an
+  // explicit top-anchored override, the model invents pseudo-tool markup
+  // (`<todo-list>`, `[读取 X]`) instead of producing real progress
+  // events — see #313. Pin this preamble ABOVE DISCOVERY_AND_PHILOSOPHY
+  // so it beats the discovery layer's own "these override anything
+  // later" header.
+  if (streamFormat === 'plain') {
```
mrcfps, line 92: In `packages/contracts/src/prompts/system.ts`, this new top-anchored override fixes the BYOK/contracts composer, but the daemon still has its own copy in `apps/daemon/src/prompts/system.ts` with the old bottom-appended `## API mode rule` (see that file around lines 252-258), and `apps/daemon/src/server.ts:6187-6193` still passes `streamFormat: def?.streamFormat ?? 'plain'` into that composer. That means plain-stream daemon agents still see `DISCOVERY_AND_PHILOSOPHY` before the only tool-suppression text, so the same `<todo-list>` / `[读取 ...]` pseudo-tool output this PR is trying to stop can still happen there. Please mirror the top-anchored override in the daemon composer (or route both paths through the same shared prompt builder) and add matching daemon coverage so the fix applies to every `streamFormat === 'plain'` path.

mrcfps (commented):
Thanks for tackling this fix — I found one remaining plain-stream path that still keeps the old prompt-precedence bug alive.


packages/contracts/src/prompts/system.ts
mutoe: Confirmed and fixed in 8eb666c. `apps/daemon/src/prompts/system.ts` now mirrors the same top-anchored `API_MODE_OVERRIDE` (byte-identical to the contracts copy), and the bottom-appended `## API mode rule` is removed. Added `apps/daemon/tests/prompts/api-mode-override.test.ts` with 8 cases including the indexOf-based positional assertion that pins the override above `# OD core directives`. Routing both paths through a single shared builder would be cleaner long-term but cross-package extraction is out of scope here (the daemon copy has extra inputs — agentId, craftBody, codex override, critique panel, MCP directive — that contracts intentionally doesn't carry).

packages/contracts/src/prompts/system.ts
```diff
@@ -79,11 +79,26 @@ export function composeSystemPrompt({
...
+    parts.push(API_MODE_OVERRIDE);
+    parts.push('\n\n---\n\n');
```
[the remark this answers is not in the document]
mutoe: Fixed in 8eb666c — daemon composer now has the same top-anchored override, old bottom `## API mode rule` removed, daemon-side coverage added with the same positional assertion.

mrcfps (approved):
@mutoe Thanks for following through on the daemon mirror as well. I verified the API-mode override now sits above the discovery layer in both prompt composers, the obsolete trailing rule is removed, and the focused regression coverage exercises the plain-stream behavior in both packages.

PR #1208, opened 2026-05-11
fix: use ACP config options for model selection

### LINKED ISSUE #1200: [Bug]  model selection uses removed `session/set_model` instead of stable `session/set_config_option'
Open Design's ACP transport (apps/daemon/src/acp.ts) uses the beta session/set_model method and reads result.models from session/new responses.
Both were removed from the ACP spec. The stable ACP protocol uses session/set_config_option and exposes models via result.configOptions in the session/new response.

This causes two user-facing problems with Devin for Terminal (and will affect any ACP agent that only implements the stable spec):

  1. Empty model picker -- The Settings model dropdown shows only "Default (CLI config)" because result.models is null in Devin's session/new
  response. Devin returns its 67 available models in result.configOptions (a configOption with id: "model", type: "select", category: "model"),
  which Open Design ignores.
  2. Custom model ID silently ignored -- When using "Custom (type below)..." with a model ID like claude-opus-4-6-thinking, the daemon sends sess
  ion/set_model which Devin rejects with error -32002 (unknown method). The error-recovery code falls back to the default model. The UI header
  shows claude-opus-4-6-thinking but Devin actually runs on its default (swe-1-6-fast).

ACP spec reference

The Session Config Options spec (https://agentclientprotocol.com/protocol/session-config-options) (now stabilized) states:

│ Session Config Options are the preferred way to expose session-level configuration. If an Agent provides configOptions, Clients SHOULD use them
│ instead of the modes field.

The session/set_model method was never part of the stable spec. See also codecompanion.nvim#2977 (https://github.com/olimorris/
codecompanion.nvim/issues/2977) which documents the same migration:

│ session/set_model was never part of the ACP spec and any references to it have now been removed from their site. Models, modes, reasoning all 
│ goes through Session Config Options now.

Current behavior (code paths)

Model detection (detectAcpModels)

Line ~270 -- reads result.models from session/new response:

const models = normalizeModels(result.models, defaultModelOption);

Devin returns result.models: null. Models are in result.configOptions.

Model setting (attachAcpSession, reusePooledSession, spawnPooledSession)

Lines ~505, ~883, ~1130 -- sends session/set_model:

writeRpc(nextId, 'session/set_model', { sessionId, modelId: model });

Devin responds with JSON-RPC error -32002 (method not found). The daemon's error-recovery falls back to the default model silently.

Expected behavior

Model detection

When result.models is null/empty, the daemon should check result.configOptions for a config option with category: "model" (or id: "model") and
populate the model picker from its options array.

Example of what Devin returns in result.configOptions:

{
  "id": "model",
  "type": "select",
  "category": "model",
  "name": "Model",
  "options": [
    { "id": "claude-opus-4-6-thinking", "name": "Claude Opus 4.6 Thinking" },
    { "id": "swe-1-6-fast", "name": "SWE-1.6 Fast" },
    ...
  ]
}

Model setting

When the user selects a model, the daemon should send session/set_config_option instead of session/set_model:

{
  "jsonrpc": "2.0",
  "id": 3,
  "method": "session/set_config_option",
  "params": {
    "sessionId": "sess_abc123",
    "configId": "model",
    "value": "claude-opus-4-6-thinking"
  }
}

Reproduction

  1. Install Devin for Terminal (devin 2026.5.6-4)
  2. Select Devin in Open Design Settings
  3. Observe model dropdown only shows "Default (CLI config)" -- no models from Devin
  4. Set "Custom (type below)..." to claude-opus-4-6-thinking
  5. Start a chat -- Devin runs but the badge shows "model default" and Devin self-reports as swe-1-6-fast

Suggested fix

In acp.ts:

  1. normalizeModels: Add a fallback that parses result.configOptions when result.models is null. Look for a config option with category: "model"
  and map its options array to ModelOption[].
  2. attachAcpSession / reusePooledSession / spawnPooledSession: When the agent's models came from configOptions, send session/set_config_option
  instead of session/set_model. Could be tracked with a per-agent flag like usesConfigOptions: true.
  3. Bonus: Also parse config options with category: "thought_level" for agents that expose reasoning level controls (Devin exposes thinking_budg
  et this way).

Environment

  • Open Design: built from source (latest main)
  • Devin for Terminal: 2026.5.6-4 (3e97eca)
  • macOS 264.1

krishnashakula:
﻿## Summary
- Parse ACP session `configOptions` for model selectors when `result.models` is empty or legacy-only
- Use stable `session/set_config_option` for selected ACP models when a model config option is advertised
- Keep legacy `session/set_model` fallback for older ACP agents and cover both paths in daemon tests

Fixes #1200

## Validation
- `pnpm guard`
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/acp.test.ts`

### DIFF
--- a/apps/daemon/src/acp.ts
+++ b/apps/daemon/src/acp.ts
@@ -29,6 +29,14 @@ export interface ModelOption {
   label: string;
 }
 
+interface AcpModelConfigOption {
+  configId: string;
+  currentValue: string | null;
+  values: unknown[];
+}
+
+const MODEL_CONFIG_OPTION_IDS = new Set(['model', 'models', 'modelid', 'modelids']);
+
 interface DetectAcpModelsOptions {
   bin: string;
   args: string[];
@@ -143,7 +151,81 @@ function choosePermissionOutcome(options: unknown): string | null {
   return null;
 }
 
-function normalizeModels(models: unknown, defaultModelOption: ModelOption): ModelOption[] {
+function normalizeConfigOptionToken(value: unknown): string {
+  return typeof value === 'string'
+    ? value.trim().toLowerCase().replace(/[\s_-]+/g, '')
+    : '';
+}
+
+function isModelConfigOption(option: JsonObject, configId: string): boolean {
+  const category = normalizeConfigOptionToken(option.category);
+  if (category === 'model') return true;
+  const id = normalizeConfigOptionToken(configId);
+  if (id === 'model') return true;
+  if (category) return false;
+  const name = normalizeConfigOptionToken(option.name);
+  return MODEL_CONFIG_OPTION_IDS.has(id) || name === 'model';
+}
+
+function findModelConfigOption(configOptions: unknown): AcpModelConfigOption | null {
+  const options = Array.isArray(configOptions) ? configOptions : [];
+  for (const rawOption of options) {
+    const option = asObject(rawOption);
+    if (!option) continue;
+    const configId = typeof option.id === 'string' ? option.id.trim() : '';
+    if (!configId) continue;
+    const type = typeof option.type === 'string' ? option.type.trim() : '';
+    if (type && type !== 'select') continue;
+    if (!isModelConfigOption(option, configId)) continue;
+    const currentValue =
+      typeof option.currentValue === 'string' && option.currentValue.trim()
+        ? option.currentValue.trim()
+        : null;
+    return {
+      configId,
+      currentValue,
+      values: Array.isArray(option.options) ? option.options : [],
+    };
+  }
+  return null;
+}
+
+function normalizeModelConfigOptions(
+  configOptions: unknown,
+  defaultModelOption: ModelOption,
+): { currentModelId: string | null; models: ModelOption[] } | null {
+  const modelConfig = findModelConfigOption(configOptions);
+  if (!modelConfig) return null;
+  const seen = new Set([defaultModelOption.id]);
+  const out = [defaultModelOption];
+  for (const rawValue of modelConfig.values) {
+    const value = asObject(rawValue);
+    if (!value) continue;
+    const id =
+      typeof value.value === 'string' && value.value.trim()
+        ? value.value.trim()
+        : typeof value.id === 'string'
+          ? value.id.trim()
+          : '';
+    if (!id || seen.has(id)) continue;
+    seen.add(id);
+    const name = typeof value.name === 'string' ? value.name.trim() : '';
+    const isCurrent = id === modelConfig.currentValue;
+    const labelBase = name && name !== id ? `${name} (${id})` : id;
+    out.push({ id, label: isCurrent ? `${labelBase} • current` : labelBase });
+  }
+  return { currentModelId: modelConfig.currentValue, models: out };
+}
+
+export function normalizeModels(
+  models: unknown,
+  defaultModelOption: ModelOption,
+  configOptions?: unknown,
+): ModelOption[] {
+  const configModels = normalizeModelConfigOptions(configOptions, defaultModelOption);
+  if (configModels && configModels.models.length > 1) {
+    return configModels.models;
+  }
   const modelsObj = asObject(models);
   const available = Array.isArray(modelsObj?.availableModels) ? modelsObj.availableModels : [];
   const currentModelId =
@@ -159,7 +241,20 @@ function normalizeModels(models: unknown, defaultModelOption: ModelOption): Mode
     const labelBase = name && name !== id ? `${name} (${id})` : id;
     out.push({ id, label: isCurrent ? `${labelBase} • current` : labelBase });
   }
-  return out;
+  return out.length > 1 || !configModels ? out : configModels.models;
+}
+
+function modelSelectionErrorIsRecoverable(code: unknown): boolean {
+  return code === -32603 || code === -32602 || code === -32601 || code === -32002;
+}
+
+function currentModelFromSessionResult(result: JsonObject): string | null {
+  const configCurrent = findModelConfigOption(result.configOptions)?.currentValue;
+  if (configCurrent) return configCurrent;
+  const models = asObject(result.models);
+  return typeof models?.currentModelId === 'string' && models.currentModelId.trim()
+    ? models.currentModelId.trim()
+    : null;
 }
 
 export function createJsonLineStream(onMessage: (message: unknown, rawLine: string) => void) {
@@ -266,7 +361,7 @@ export async function detectAcpModels({
         return;
       }
       if (expectedId === 2) {
-        const models = normalizeModels(result.models, defaultModelOption);
+        const models = normalizeModels(result.models, defaultModelOption, result.configOptions);
         finish(resolve, models);
         if (!child.killed) child.kill('SIGTERM');
       }
@@ -324,6 +419,7 @@ export function attachAcpSession({
   let setModelRequestId: JsonRpcId | null = null;
   let sessionId: string | null = null;
   let activeModel: string | null = null;
+  let modelConfigId: string | null = null;
   let emittedThinkingStart = false;
   let emittedFirstTokenStatus = false;
   let finished = false;
@@ -393,6 +489,13 @@ export function attachAcpSession({
     }
   };
 
+  const recoverFromModelSelectionError = () => {
+    setModelRequestId = null;
+    activeModel = activeModel || 'default';
+    send('agent', { type: 'status', label: 'model', model: activeModel });
+    sendPrompt();
+  };
+
   const parser = createJsonLineStream((raw, rawLine) => {
     if (aborted) return;
     resetStageTimer('response');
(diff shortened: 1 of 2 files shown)


mrcfps (commented):
Thanks for pushing the ACP config-option path forward — I verified the new `session/set_config_option` flow and found one compatibility gap worth tightening before callers rely on it broadly.


apps/daemon/src/acp.ts
```diff
@@ -143,7 +149,66 @@ function choosePermissionOutcome(options: unknown): string | null {
...
+  const options = Array.isArray(configOptions) ? configOptions : [];
+  for (const rawOption of options) {
+    const option = asObject(rawOption);
+    if (!option) continue;
+    const configId = typeof option.id === 'string' ? option.id.trim() : '';
+    if (!configId) continue;
+    const type = typeof option.type === 'string' ? option.type.trim() : '';
+    if (type && type !== 'select') continue;
+    const category = typeof option.category === 'string' ? option.category.trim() : '';
+    if (category !== 'model' && configId !== 'model') continue;
```
mrcfps: `findModelConfigOption()` currently only treats an option as the model selector when `category === 'model'` or `id === 'model'`. ACP's config-option category is optional, and the Session Config Options RFD explicitly says category must not be required for correctness while also showing model selectors with agent-defined ids such as `"models"`. With the predicate in this hunk, a category-less selector is ignored, `normalizeModels()` falls back to legacy/default data, and `attachAcpSession()` later sends `session/set_model` instead of `session/set_config_option`, so agents that only implement config options still cannot switch models here. Please add a fallback for category-less model selectors (for example `name === 'Model'` and known id variants such as `models`) and cover that shape in `apps/daemon/tests/acp.test.ts` so the stable ACP path stays usable when agents omit `category`.

krishnashakula:
﻿Updated the PR for the category-less config option review feedback.

- `findModelConfigOption()` now accepts category-less model selectors via normalized id variants like `models` and name `Model`.
- Added ACP test coverage for `id: "models", name: "Model"` discovery and verified selection still sends `session/set_config_option`.

Validation under Node 24.15.0:
- `pnpm guard`
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/acp.test.ts`

mrcfps (approved):
@krishnashakula I re-checked the latest head and verified the ACP model-selection path now handles category-less `configOptions`, uses `session/set_config_option` when a model config option is advertised, and still falls back to legacy `session/set_model` when only the older session model API is present. I also ran the targeted daemon ACP test file and daemon typecheck against `b1d52f9e74215832159898a0ae956562bf289b77`, and both completed cleanly. Thanks for tightening up the compatibility path here — this looks good to merge. ✨


apps/daemon/src/acp.ts
```diff
@@ -143,7 +151,81 @@ function choosePermissionOutcome(options: unknown): string | null {
...
+  return typeof value === 'string'
+    ? value.trim().toLowerCase().replace(/[\s_-]+/g, '')
+    : '';
+}
+
+function isModelConfigOption(option: JsonObject, configId: string): boolean {
+  const category = normalizeConfigOptionToken(option.category);
+  if (category === 'model') return true;
+  const id = normalizeConfigOptionToken(configId);
+  if (id === 'model') return true;
```
lefarcen, line 164: P2 — Document the early-return assumption: "non-model category → non-model selector". The RFD says category must not be required for correctness, which current code handles via the fallback at L167-168. However, this line means: if an agent sets `category: 'llm-selection'` + `id: 'model'`, the selector would be rejected here before reaching the id-based fallback. That's probably fine (no known agents do this), but worth a comment explaining the assumption so future maintainers understand the design choice.

lefarcen (commented):
Nice work addressing the config-option fallback! The category-less selector path is now covered. One small documentation suggestion below (non-blocking).

MrRockySL:
I'm the reporter of #1200, so wanted to follow up on this with some end-to-end validation.

Pulled the branch (`b1d52f9`), built Open Design from source with `tools-pack mac build --to app`, and ran it against Devin for Terminal — the same setup that originally hit the bug.

The fix works. Typecheck passes, all 9 tests in `tests/acp.test.ts` pass, and the diff is clean and focused. The `name: "Model"` category-less fallback added in `b1d52f9` is exactly what the spec calls for — nice catch from the earlier review.

In actual use:

- The picker populates with all of Devin's models (was empty before — that was the original symptom of #1200)
- Selecting `claude-opus-4-6-thinking` actually runs Devin on that model. The conversation header shows `Devin · claude-opus-4-6-thinking` and Devin self-reports as *"I am Claude Opus 4.6 Thinking (Model ID: claude-opus-4-6-thinking)"*
- "Default (CLI config)" still works
- No more `-32002` errors — `session/set_config_option` is being accepted by Devin

I did spot two UX things while testing that are worth flagging, though I don't think they're caused by this PR. They show up in this build but not in a separate local copy of the fix I have, which also includes some unrelated `apps/daemon/src/server.ts` session-pool changes (keeping the daemon's child processes alive across prompts instead of respawning each time). My bestguess is they're pre-existing on main and that other work happens to mask them, but worth a heads-up either way.

The first is that the model badge under the conversation header always shows `model swe-1-6-fast` no matter which model I pick. The actual running model is correct (Devin self-reports the right one), so it's just the badge UI that's misleading.

The second is that after Devin finishes responding, the chat stays in "working" state — I have to click stop manually before I can send another message. The "Done · 9.3s · 29 out" indicator appears like normal but the input doesn't return to ready.

Happy to rebuild against plain `main` and compare if it'd help confirm whether these are pre-existing or actually introduced here.

From a strict #1200 perspective though, the PR does what the title promises. Looks good to merge from my end.

Screenshot 2026-05-11 at 12 35 16

lefarcen:
@MrRockySL Thanks for the thorough end-to-end validation — this is exactly the kind of real-world confirmation we need. 🙏

Great to hear the core fix works: picker populates, model selection applies correctly, and no more `-32002` errors. That's the scope of this PR (ACP config-option path + category-less selector fallback).

The two UX issues you flagged (badge showing wrong model + chat stuck in "working" state) sound like they're outside this PR's scope — this change only touches `apps/daemon/src/acp.ts` and tests, not the UI layer or session lifecycle. Your best-guess that they're pre-existing on main is likely right, especially if your other local build with session-pool changes masks them.

If you have time, it'd help to confirm they reproduce on plain `main` (without this PR) and then file separate issues for:
1. Model badge always showing `swe-1-6-fast` instead of actual selected model
2. Chat not returning to ready state after Devin completes (stuck in "working")

That way maintainers can track/fix them independently without blocking this PR. Either way, thanks for the detailed testing and the screenshot!

MrRockySL:
Hey @lefarcen 

Did the test on plain `main` (commit `421ddf55`) — Devin for Terminal as the ACP agent.

**Bug B (chat stuck in "working" state):** reproduces on `main`. After Devin's response completes and `Done · X.Xs` shows up, theinput stays blocked — have to click stop before sending the next prompt. Also tested with Codex agent on the same build for comparison: Codex doesn't hit this, so it looks Devin-specific (probably something in how the Devin ACP session lifecycle wraps up).

**Bug A (model badge always shows `swe-1-6-fast`):** not testable on `main` because #1200 is still active — picker only shows "Default (CLI config)", so there's no way to select a non-default model to trigger the wrong-badge state in the first place. Filing it anyway based on what I observed in the #1208 build, with a note explaining the precondition.

One related observation that might help whoever picks these up: in my local copy of the fix that also includes a `server.ts` session-pool change (keeps the Devin child process alive between prompts instead of respawning each one), neither bug shows up. That points at session lifecycle as the likely root cause for at least Bug B, possibly Bug A too.

Filing the two issues now and will link them back here.

MrRockySL:
Bug B filed: #1265

For Bug A (badge wrong), I held off on filing it as a separate issue. Only manifests in builds with #1208 (or equivalent) applied — it's not reproducible on `main` since #1200 prevents non-default model selection in the first place. So it feels less like a standalone bug and more like a UI-side gap to this PR's daemon-layer fix: once selection works in `acp.ts`, the model badge in the web UI needs to track the active model.

@lefarcen happy to file it anyway with a "post-#1208 only" framing if you'd still prefer it tracked separately, or it could be addressed as part of #1208's scope / a follow-up PR. Let me know which you'd prefer.

lefarcen:
Thanks for the thorough validation and isolation work @MrRockySL. 

For Bug A (model badge): hold off on filing it separately for now. Since it only manifests after #1208 lands and you've traced it to session lifecycle (same area as Bug B), it's better addressed after #1208 merges — either as a quick follow-up PR or folded into the session-pool work you mentioned if that's coming soon.

Bug B (#1265) is already tracked, and your session-lifecycle hypothesis gives whoever picks it up a good starting point.

The core fix here (model selection) is validated and ready. Thanks for the detailed testing.

lefarcen:
@mrcfps Hi, could you please take another look at this PR? The unresolved review thread has been addressed. The current head is b1d52f9e. Thanks!

leilei926524-tech:
Hi @krishnashakula — welcome to Open Design contributors. 🎉

Your first merged PR (#1208) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

Issue #1199, opened 2026-05-11
Format JSON preview instead of rendering it as a single line


shangxinyu1:
## Summary

The JSON preview currently renders raw JSON as a single long line, which makes it hard to read.

## Current behavior

When opening a JSON file in the preview pane, the content is shown without formatting or indentation. Large objects appear on one line and quickly become difficult to inspect.

## Expected behavior

JSON previews should be pretty-printed with indentation and line breaks so the structure is easy to scan.

## Steps to reproduce

1. Open a JSON file in Design Files.
2. View it in the preview pane.
3. Observe that the JSON is rendered as a single compact line.

## Notes

- A screenshot was provided showing the JSON preview as one long line.
- This appears to be a preview formatting issue rather than a data issue.
- The request is specifically about improving readability in the preview surface.

## Why this matters

Unreadable JSON slows down inspection and debugging, especially for large payloads or connector outputs. Pretty-printing would make the built-in file viewer much more useful.

<img width="2560" height="1738" alt="Image" src="https://github.com/user-attachments/assets/6bf858c2-cf3e-4c53-aa92-177b4b123bf2" />

lefarcen:
谢谢反馈，问题已确认 👍

当前 Design Files 里的 `.json` 文件按普通代码文件走 `FileViewer -> TextViewer` 路径，直接显示文件原始文本。所以如果 JSON 本身是压缩成一行的（例如 connector 输出、GitHub API 响应等），预览区就会显示成一长行，确实很难读。

**根因**：`apps/web/src/components/FileViewer.tsx` 的 `TextViewer` 没有针对 JSON 文件做解析 + 格式化的步骤。

**建议修法**：在 JSON 文件预览时先尝试 `JSON.parse(text)` + `JSON.stringify(parsed, null, 2)` 加缩进重新格式化；解析失败时回退到原始文本（避免影响非标准 JSON / 带注释的 JSON5 等文件）。

如果你愿意提 PR 修这个，在下面评论 `/claim` 就能认领。修改范围很小，适合首次贡献 🚀

lefarcen:
This looks like a good first issue for someone new to the project.

If you'd like to work on it, comment `/claim` below. The first `/claim` after the latest release gets it. Any comment, linked PR, or progress note keeps the claim active.

lefarcen:
Claimed by @Leesin0222. Thanks for picking this up! 🎉

**Quick question to time check-ins right** — roughly how long do you think this will take?

[...]

PR #1216, opened 2026-05-11
pretty-print JSON in file preview

Leesin0222:
## Summary

JSON files opened in the Design Files preview were displayed exactly as stored on disk. Minified payloads (connector outputs, GitHub API responses, cached fetch results) therefore rendered as a single long line, making them hard to scan or debug.

This patch reformats `.json` files in `TextViewer` using `JSON.parse(text)` + `JSON.stringify(parsed, null, 2)` before display. Parse failures fall back to the original text, so JSON5 / comment-bearing files are left untouched.

issue #1199

## Changes

- `apps/web/src/components/FileViewer.tsx`
  - New `formatJsonForPreview(fileName, text)` helper (exported for tests). Returns 2-space indented JSON for `.json` files when parse succeeds; returns the input unchanged otherwise.
  - `TextViewer` derives `displayText` from the fetched text via `useMemo`, and uses it for both the rendered preview and the Copy button so what the user sees is what they copy.
- `apps/web/tests/components/FileViewer.test.tsx`
  - Adds 5 unit tests for `formatJsonForPreview` covering: minified → pretty, case-insensitive extension match, non-`.json` passthrough, parse-failure fallback (JSON with comments), and empty-string input.

## Before / After

**Before** : minified `.json` rendered as one long line:

```
{"a":1,"b":[2,3,4],"c":{"d":"hello","e":[{"f":true}]}}
```

**After** : same file rendered with 2-space indentation:

```
{
  "a": 1,
  "b": [
    2,
    3,
    4
  ],
  "c": {
    "d": "hello",
    "e": [
      {
        "f": true
      }
    ]
  }
}
```

## Test plan

- [x] `pnpm --filter @open-design/web test` — 748/748 passing, including the 5 new `formatJsonForPreview` cases.
- [x] `pnpm --filter @open-design/web typecheck` — clean.
- [x] Manual: `pnpm tools-dev`, open a project, drop a minified `payload.json` into Design Files, open it in the preview pane — should render with indentation. Repeat with a `config.json` that unchanged.
- [x] `pnpm guard` and `pnpm typecheck` (root).

## Notes

- **No new dependencies.** The change uses only built-in `JSON.parse` / `JSON.stringify` and the existing React `useMemo` import.
- **No on-disk changes.** Only the in-memory display string is reformatted; the file on disk is never touched.
- **Side effect to be aware of:** `.json` files that were already pretty-printed on disk with a different style (4-space, tabs, custom spacing) will be normalized to 2-space in the preview. The Copy button also returns the normalized text. Happy to switch to a more conservative variant (only reformat when the file has no newlines, Copy returns raw) if reviewers prefer.
- Detected by extension only (`.json`). Files with `application/json` MIME but no `.json` extension are not affected.

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -5665,13 +5665,85 @@ function TextViewer({
   );
 }
 
+// Pretty-print JSON without round-tripping through JS values, so 64-bit IDs
+// and other numeric literals beyond Number.MAX_SAFE_INTEGER (also exponent
+// notation, \u-escapes inside strings) survive the preview byte-for-byte.
+// `JSON.parse` is used only as a syntactic validator; its decoded result is
+// discarded and the original text is re-emitted with normalized whitespace.
 export function formatJsonForPreview(fileName: string, text: string): string {
   if (!fileName.toLowerCase().endsWith('.json')) return text;
   try {
-    return JSON.stringify(JSON.parse(text), null, 2);
+    JSON.parse(text);
+    return prettyPrintJsonPreservingTokens(text);
   } catch {
     return text;
   }
+}
+
+function prettyPrintJsonPreservingTokens(input: string): string {
+  let out = '';
+  let depth = 0;
+  let inString = false;
+  let escape = false;
+
+  const pad = (n: number) => '  '.repeat(n);
+  const peek = (from: number): string => {
+    for (let i = from; i < input.length; i++) {
+      const c = input[i];
+      if (c !== ' ' && c !== '\t' && c !== '\n' && c !== '\r') return c;
+    }
+    return '';
+  };
+
+  for (let i = 0; i < input.length; i++) {
+    const c = input[i];
+
+    if (inString) {
+      out += c;
+      if (escape) escape = false;
+      else if (c === '\\') escape = true;
+      else if (c === '"') inString = false;
+      continue;
+    }
+
+    if (c === '"') {
+      inString = true;
+      out += c;
+      continue;
+    }
+    if (c === ' ' || c === '\t' || c === '\n' || c === '\r') continue;
+
+    if (c === '{' || c === '[') {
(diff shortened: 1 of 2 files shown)


mrcfps (approved):
@Leesin0222 I re-checked the latest head and verified the JSON preview path now preserves original numeric and string tokens while still pretty-printing valid `.json` files, and the added tests cover the main regression cases around precision, escapes, punctuation-in-strings, and parse-failure fallback. Thanks for the careful follow-up and the solid regression coverage ✨


lefarcen (commented):
Hi @Leesin0222, thanks for tackling issue #1199! I see you've built a token-preserving pretty-printer that doesn't round-trip through JSON value conversion — that's a solid approach for handling precision-sensitive numbers and preserving original token representations.

However, PR #1206 (opened ~90 minutes earlier) is also addressing the exact same issue with a similar goal: pretty-print JSON in FileViewer preview while preserving numeric precision. Both PRs change the same two files (FileViewer.tsx and its test file) and use different but valid strategies for the precision-preservation problem.

This creates a duplicate situation. A few options:

1. **Close this PR and contribute review/testing to #1206** — since it was opened first and already has mrcfps's review attention.
2. **Keep both open and let maintainers pick** — if you believe your token-scan approach has advantages worth comparing (e.g. simpler logic, different edge-case handling), you could make that case.
3. **Coordinate with @[author of #1206]** to combine the best parts of both approaches.

What would you prefer? Either way, your work here is valuable — the token-preserving formatter you built is well-structured and the test coverage is solid. 🙏

PerishCode:
Closing as a duplicate of #1206, which targets the same `apps/web/src/components/FileViewer.tsx` + test surface for the JSON pretty-print fix and was opened ~1h earlier. #1206 is already APPROVED; consolidating review there. Thanks for the contribution — feel free to follow up on #1206 if you have additions beyond what it covers.

PR #1222, opened 2026-05-11
feat(daemon): make connection-test timeouts configurable

Eggwardhan:
## Motivation

The Settings → Connection Test flow has two hardcoded budgets:

- **Provider** connection test: `PROVIDER_TIMEOUT_MS = 12_000`
- **Agent** connection test: `AGENT_TIMEOUT_MS = 45_000`

These are tuned for happy-path latencies, but on slow networks, distant providers, or constrained CLI adapter cold-starts they time out with no user-facing way to extend the budget. The user only sees "timeout" with no recourse.

## Change

`apps/daemon/src/connectionTest.ts`:

- Rename the constants to `DEFAULT_PROVIDER_TIMEOUT_MS` / `DEFAULT_AGENT_TIMEOUT_MS` (same default values, no behavior change for users who set nothing).
- Add two env-var overrides:
  - `OD_CONNECTION_TEST_PROVIDER_TIMEOUT_MS`
  - `OD_CONNECTION_TEST_AGENT_TIMEOUT_MS`
- Defensive parsing: non-numeric / zero / negative / fractional values are rejected with a `console.warn` and the default is used. A typo in the env cannot silently disable the safety timeout.
- Export `resolveConnectionTestTimeoutMs(key, fallback, env)` so the resolution logic is unit-testable.

`apps/daemon/tests/connection-test.test.ts`:

- 3 new tests covering the three resolution paths: empty env → fallback, honored numeric override, invalid value → warn + fallback.

## Compatibility

- No public API change in the daemon.
- Defaults (12s / 45s) are preserved — existing deployments and CI run unchanged unless the new env vars are explicitly set.
- Constant-name change is internal-only (no external imports of the old names).

- [ ] `apps/daemon` vitest suite green (3 new cases pass; nothing else regresses)
- [ ] `resolveConnectionTestTimeoutMs` behavior matches the parse-rules table in the source comments

### DIFF
--- a/apps/daemon/src/connectionTest.ts
+++ b/apps/daemon/src/connectionTest.ts
@@ -46,11 +46,45 @@ import {
 export { validateBaseUrl } from '@open-design/contracts/api/connectionTest';
 
 // Aggressive but not punitive — happy paths usually return in under 2 s.
-const PROVIDER_TIMEOUT_MS = 12_000;
+// Override with OD_CONNECTION_TEST_PROVIDER_TIMEOUT_MS for slow networks
+// or distant providers; invalid values fall back to the default.
+const DEFAULT_PROVIDER_TIMEOUT_MS = 12_000;
 // CLI boot time is dominated by adapter auth/session restore; the heavy
 // adapters (Codex, Cursor Agent) regularly take 5–10 s on a cold first
 // run, so 45 s leaves headroom without making a hung child invisible.
-const AGENT_TIMEOUT_MS = 45_000;
+// Override with OD_CONNECTION_TEST_AGENT_TIMEOUT_MS.
+const DEFAULT_AGENT_TIMEOUT_MS = 45_000;
+
+export function resolveConnectionTestTimeoutMs(
+  key: 'OD_CONNECTION_TEST_PROVIDER_TIMEOUT_MS' | 'OD_CONNECTION_TEST_AGENT_TIMEOUT_MS',
+  fallback: number,
+  env: NodeJS.ProcessEnv = process.env,
+): number {
+  const raw = env[key];
+  if (raw === undefined || raw === '') return fallback;
+  const n = Number(raw);
+  if (!Number.isFinite(n) || !Number.isInteger(n) || n < 1) {
+    console.warn(
+      `connection-test: ignoring ${key}=${JSON.stringify(raw)} (must be a positive integer ms); using ${fallback}ms`,
+    );
+    return fallback;
+  }
+  return n;
+}
+
+function providerTimeoutMs(): number {
+  return resolveConnectionTestTimeoutMs(
+    'OD_CONNECTION_TEST_PROVIDER_TIMEOUT_MS',
+    DEFAULT_PROVIDER_TIMEOUT_MS,
+  );
+}
+
+function agentTimeoutMs(): number {
+  return resolveConnectionTestTimeoutMs(
+    'OD_CONNECTION_TEST_AGENT_TIMEOUT_MS',
+    DEFAULT_AGENT_TIMEOUT_MS,
+  );
+}
 const AGENT_COMPLETION_DEBOUNCE_MS = 500;
 const AGENT_KILL_GRACE_MS = 2_000;
 // Truncates the assistant reply we surface in the success copy so a
@@ -492,7 +526,7 @@ export async function testProviderConnection(
   } else {
     input.signal?.addEventListener('abort', abortFromParent, { once: true });
   }
-  const timer = setTimeout(() => controller.abort(), PROVIDER_TIMEOUT_MS);
+  const timer = setTimeout(() => controller.abort(), providerTimeoutMs());
 
   try {
     const modelError = await validateLocalOpenAiModel(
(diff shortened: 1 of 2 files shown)


apps/daemon/src/connectionTest.ts
```diff
@@ -46,11 +46,45 @@ import {
...
+  const raw = env[key];
+  if (raw === undefined || raw === '') return fallback;
+  const n = Number(raw);
+  if (!Number.isFinite(n) || !Number.isInteger(n) || n < 1) {
+    console.warn(
+      `connection-test: ignoring ${key}=${JSON.stringify(raw)} (must be a positive integer ms); using ${fallback}ms`,
+    );
+    return fallback;
+  }
+  return n;
```
mrcfps, line 79: In resolveConnectionTestTimeoutMs, these changed lines accept any positive integer and return it directly to the setTimeout call sites. Node timers only support delays up to 2_147_483_647 ms, so an override like OD_CONNECTION_TEST_AGENT_TIMEOUT_MS=3000000000 overflows with TimeoutOverflowWarning and is scheduled for about 1 ms instead of a longer budget. That means a user trying to increase the timeout can get an almost immediate connection-test failure from this new path. Please reject or clamp values above the Node timer maximum here, and add a regression in apps/daemon/tests/connection-test.test.ts that proves oversized overrides fall back safely.

mrcfps (commented):
@Eggwardhan Thanks for shipping the configurable connection-test budgets and the focused resolver tests. I reviewed the changed ranges and found one merge-safe follow-up: oversized timeout overrides can still flip the new setting into an immediate timeout on Node timers.

apps/daemon/src/connectionTest.ts
lefarcen, line 79: **P1**: `Number.isInteger(n)` accepts values larger than Node's timer range, including `Number.MAX_SAFE_INTEGER`. `setTimeout(..., n)` overflows for delays above `2_147_483_647` (about 24.8 days) and is clamped to ~1ms, making every connection test time out immediately.

**Fix**: Add a max bound (practical daemon timeout cap or at least `2_147_483_647`), use `Number.isSafeInteger`, and add a regression test for an oversized env value falling back or clamping.

Example:
```typescript
const MAX_TIMEOUT_MS = 2_147_483_647; // Node timer limit
if (!Number.isSafeInteger(n) || n < 1 || n > MAX_TIMEOUT_MS) {
  console.warn(...);
  return fallback;
}
```

lefarcen (changes requested):
Hi @Eggwardhan! Thanks for tackling the timeout configuration pain point — this is a real usability improvement for slow networks and cold-start scenarios.

Found one must-fix issue (timer overflow with large env values) and one design rationale question. See inline below.

lefarcen:
**P3 (design rationale)**: The motivation says users have no recourse, but it doesn't explain why the recourse is an env var rather than a Settings UI value, persisted daemon config, or launch flag.

This is fine as an operator-only escape hatch, but if it's meant to be user-facing configuration, adding a short alternatives/rationale note would help reviewers evaluate the product surface choice. (Not blocking — just context for maintainer sign-off.)

Eggwardhan:
Thanks @lefarcen / @mrcfps / @chatgpt-codex-connector for catching the Node timer overflow — confirmed the bug end-to-end.

Pushed [b8b7a9b3](https://github.com/nexu-io/open-design/pull/1222/commits/b8b7a9b3):

- New `MAX_CONNECTION_TEST_TIMEOUT_MS = 2_147_483_647` constant.
- Guard switched to `Number.isSafeInteger(n) && n >= 1 && n <= MAX`. `isSafeInteger` also rejects values past `2^53` (e.g. floats represented as huge integers after the env round-trip), tightening the path `isInteger` left open.
- Warn message updated to surface the upper bound so a user with an oversized value sees exactly why it was rejected.
- Regression test `rejects values above the Node setTimeout maximum (2^31-1)` exercises three points: `3_000_000_000` (rejects), `2_147_483_647` (accepted — exact boundary), `2_147_483_648` (rejects — one ms past).

Local note: the full `apps/daemon` vitest suite cannot run from a clean origin/main checkout right now (`src/critique/persistence.ts:22` throws `CRITIQUE_RUN_STATUSES is not iterable` and a separate `../src/providerModels.js` resolver miss), so CI is the authoritative check. Happy to address the "design rationale question" once you confirm whether you wanted that broken out into a separate thread reply.

lefarcen (approved):
Hey @Eggwardhan! 👍

The timer overflow fix looks solid:
- Added `MAX_CONNECTION_TEST_TIMEOUT_MS = 2_147_483_647` cap
- Switched to `Number.isSafeInteger` for safer validation
- Reject oversized values with clear warning
- Added the exact regression test case (3000000000 ms → fallback)

The P1 blocker is resolved. The P3 design rationale question (env var vs UI setting) is still open for discussion, but that's non-blocking — this PR solves the immediate user pain (no recourse on slow networks) with a safe operator escape hatch.

Looks good to merge from the code correctness side. 🎉

mrcfps (approved):
@Eggwardhan I re-reviewed the changed ranges on `apps/daemon/src/connectionTest.ts` and `apps/daemon/tests/connection-test.test.ts`, and the current head now safely bounds both configurable timeout overrides, preserves the existing defaults when unset or invalid, and covers the new resolver behavior with focused regression cases. Thanks for addressing the timer-overflow path and tightening the validation here — this looks ready from the implementation side.

leilei926524-tech:
Hi @Eggwardhan — quick note from the Open Design team.

What you get there:

PR #955, opened 2026-05-08
feat: generic skills + split skills/design-templates + finalize-design API

pftom:
## Summary

This branch bundles three rounds of skills-system work that landed on `opaque-scilla`:

### 1. Generic skills with @-mention composition + user import (original scope)

Lifts skills from "one mode-bound skill per project" to a generic capability the user composes per turn — inspired by Hermes Agent / LobeHub.

- **Daemon**: scan multiple skill roots in priority order (user-skills under runtime data, then bundled `skills/`); user-imported skills can shadow built-ins by id. Each summary carries a `source: 'built-in' | 'user'` tag.
- **New endpoints**: `POST /api/skills/import` (BAD_REQUEST/CONFLICT/INTERNAL_ERROR codes) and `DELETE /api/skills/:id` (refuses built-ins).
- **Per-turn composition**: `ChatRequest` gains `skillIds: string[]`; the chat run concatenates each picked skill's body (and merges craftRequires) into the system prompt for that turn only — the project's persistent `skillId` is untouched.
- **Composer @-picker**: the existing `@` popover lists skills alongside project files. Picks render as removable chips and ride along via `ChatSendMeta.skillIds`.
- **Settings → Skills**: import form (name/description/triggers/body), per-card delete for user skills, "user" origin badge.

### 2. Split `skills/` ↔ `design-templates/` (Phase 0 of the refactor in `specs/current/skills-and-design-templates.md`)

Today `skills/` mixes ~104 rendering catalogue entries (decks/prototypes/image-video-audio templates) with a small set of functional skills that *do work* on user input (utilities, briefs, packagers). This PR pulls them apart at every layer:

- **Filesystem**: ~104 entries moved from `skills/` to `design-templates/`. New `AGENTS.md` in each describing the contract, plus a brand-agnostic `craft/` surface for opt-in craft rules.
- **Daemon**: adds `DESIGN_TEMPLATES_DIR` / `USER_DESIGN_TEMPLATES_DIR` and an `/api/design-templates*` surface mirroring `/api/skills*`. Asset/example routes still span both registries so existing srcdoc URLs keep resolving across the rename.
- **Web**: split `LibrarySection` into `SkillsSection` + `DesignSystemsSection`, rename the EntryView "Examples" tab to "Templates", and update locales + the New-project picker accordingly.

### 3. `finalize-design` endpoint

[...]

PR #1231, opened 2026-05-11
feat(design-systems): add structured tokens.css schema (default + kami)

zoeforfun:
## Why

Agents reading a brand's `DESIGN.md` see prose like `Primary: #000000`
and have to translate it to the token names the rest of the system
expects (`--accent`, `--bg`, etc.). That translation step is where
most token misuse happens — agents pick the wrong slot, hardcode the
hex, or invent new names. `lint-artifact.ts` already enforces the
correct vocabulary; the missing piece was a **structured form of**
**each brand's bindings** that agents can paste verbatim instead of
re-deriving every time.

This PR establishes that structured form across two intentionally
contrasting brands, codifies the resulting schema as a machine-readable
contract, and wires five guard checks that keep authors and the future
derive script honest.

## What ships

```
design-systems/
├── _schema/
│   ├── tokens.schema.ts    — canonical schema: every shared token
│   │                          with its layer + metadata + extensions
│   ├── defaults.css        — A2 fallback values in CSS form
│   └── AGENTS.md           — layer model + C→B→A promotion rules
├── default/
│   ├── tokens.css          — :root with 57 declarations across the
│   │                          A1 / A2 / B-slot / C-extension layers
│   └── components.html     — self-contained reference fixture, 0 raw
│                             hex outside :root, CTAs link to real
│                             sibling files (./tokens.css, ./DESIGN.md)
└── kami/
    ├── tokens.css          — :root + i18n overrides for zh-CN / ja,
    │                          12 brand-specific C extensions
    └── components.html     — exercises section-num / ring-shadow
                              buttons / tabular-nums metrics /
                              en-dash bullets / solid-hex tags

scripts/
├── check-tokens-fixture-sync.ts  — five guard checks (see below)
└── guard.ts                       — registers each check as its own
                                     entry in pnpm guard
```

Daemon doesn't read these files yet — that's intentional. Prompt
injection lands in a follow-up PR so the schema can absorb review
feedback before downstream code starts depending on the shape.

## Schema layers

Every shared token belongs to exactly one of four layers, with C as
the brand-specific escape hatch:

| Layer | Count | Required? | If a brand omits the token |
| --- | --- | --- | --- |
| **A1-identity** | 8 | yes | guard fails — the brand IS these values (`--bg`, `--fg`, `--accent`, font stacks) |
| **A1-structure** | 18 | yes | guard fails — type scale, container, section-y are structural decisions per brand |
| **A2** | 26 | yes (today) | guard fails today; derive script (PR-B) will inline `defaults.css` values tomorrow |
| **B-slot** | 4 | yes | guard fails — brand declares either as `var(--sibling)` (collapsed, e.g. default) or independent value (richer, e.g. kami) |
| **C-extension** | n | optional | allowlisted per-brand in `BRAND_EXTENSIONS` or by prefix (`--tag-bg-*`) |

**Why A2 is required-with-fallback rather than optional:** agents
generate artifacts by pasting one brand's `:root` block into a single
`<style>`. There is no global stylesheet at runtime, so a missing
`--motion-fast` resolves to nothing inside the artifact and any
`transition: var(--motion-fast)` rule silently breaks. Until the
derive script ships, every brand's tokens.css must declare every A2
token directly. The fallback in `_schema/defaults.css` is the *future
input to the derive script*, not a runtime cascade.

**Why `--font-mono` lands in A2 (not A1):** of the 149 brand
DESIGN.md files in this repo, 87 (58%) declare a monospace stack and
62 (42%) do not — including major brands like bmw / nike / apple /
notion / mastercard. Agent paste cannot rely on the brand author
having written it down; a defaultable A2 fallback (with CJK brands
like kami overriding) is safer than forcing every brand to add a
field they may not realize their kbd / code-block components need.

## Schema decisions surfaced by the kami stress test

`default` was the obvious starting brand (the spec's hand-authored
starter). `kami` was picked as the second brand because it's the
print-first opposite of default — serif at one weight, parchment
canvas, ink-blue accent, no italic, no rgba tints, four levels of
foreground gray, multi-language font stacks. Doing both surfaced 5
schema gaps that would have stayed hidden if only `default` shipped:

| #Gap | Schema addition                                    | Resolves                                                                                                 |
| ---- | -------------------------------------------------- | -------------------------------------------------------------------------------------------------------- |
| 1    | `--fg-2` / `--meta` (4-level fg ramp, B-slot)      | kami's near-black → dark-warm → olive → stone                                                            |
| 2    | `--surface-warm` (3-level surface, B-slot)         | kami's parchment → ivory → warm-sand                                                                     |
| 3    | `--border-soft` (2-level border, B-slot)           | kami's card-edge vs row-separator                                                                        |
| 4    | `--accent-hover` binds to value, not formula       | ink-blue is too dark for further darkening; kami binds it identical to `--accent` and lifts via elevation |
| 5    | `--elev-ring` (ring elevation as A2 first-class)   | kami forbids blur shadows; uses `0 0 0 1px ...` ring as primary edge                                     |

All 5 are backward-compatible: `default/tokens.css` declares the new
slots and aliases via `var()` to existing tokens, so default's visual
output is unchanged. Brands that need the richer form (kami) bind
independent values; brands that don't (default) collapse the ramp.

## Guard checks

Six separate entries in `pnpm guard`, each fails for a specific
contract:

```
Design system token-fixture sync         — components.html :root is
                                            byte-equivalent to tokens.css
                                            :root after canonical
                                            normalization

Design system A1 required tokens         — every brand declares every
                                            A1-identity / A1-structure
                                            token

Design system A2 required tokens         — every brand declares every
                                            A2 token (until the derive
                                            script makes this automatic)

Design system B-slot required tokens     — every brand declares every
                                            B-slot token; alias the named
                                            sibling via var(...) or bind
                                            independently (failure message
                                            names the schema-suggested
                                            alias for each missing slot)

Design system unknown token allowlist    — every declared token is in
                                            the shared schema or in
                                            BRAND_EXTENSIONS / matches
                                            BRAND_EXTENSION_PREFIXES

Design system A2 defaults parity         — every A2 fallback in
                                            tokens.schema.ts matches
                                            the corresponding declaration
                                            in defaults.css byte-for-byte
```

## Reviewer feedback addressed

- **mrcfps (resolved):** fixture CTAs (`#tokens`, `#spec`, `#surface`,
  `#accent`, `#type`, `#components`) advertised in-page anchors with
  no matching `id`s. Re-pointed each link to the real sibling file
  (`./tokens.css`, `./DESIGN.md`) so every CTA lands on a concrete
  resource. Browsers render these as raw source views, which is the
  desired UX for a reference fixture.

- **lefarcen (resolved):** "the optional-vs-required split won't
  generalize cleanly when brand #3 needs different Layer A tokens or
  when multiple brands converge on the same extension." The schema is
  now a versioned TS file with explicit C → B-slot → A2 promotion
  rules in `_schema/AGENTS.md`. Brand #3 onwards either matches the
  schema (no change), or its needs surface as guard failures
  (unknown-token violation), forcing a deliberate decision: extend
  `BRAND_EXTENSIONS`, promote to B-slot, or promote to A2.

- **mrcfps + lefarcen (resolved, blocking):** "the guard validates A2
  required tokens but has no companion B-slot validation; a brand can
  omit `--fg-2` / `--meta` / `--surface-warm` / `--border-soft`, pass
  `pnpm guard`, and break any artifact that references them." Added
  `checkDesignSystemBSlotRequiredTokens` to
  `scripts/check-tokens-fixture-sync.ts`, registered as its own
  `pnpm guard` entry. The failure message names each missing slot
  AND its schema-suggested alias (e.g. `--fg-2 (default alias:
  var(--fg))`) so a brand author has a copy-pasteable resolution.
  `_schema/AGENTS.md` was updated in the same commit so the narrative
  matches the now-machine-enforced contract: B-slot row in the layer
  table is `If omitted: guard fails`, and a new "Why B-slot is
  required (and what the alias is for)" section distinguishes the
  schema-suggested alias from a runtime fallback.

## Verification

- `apps/daemon/src/lint-artifact.ts` runs **0 findings** on both
  `components.html` fixtures
- `pnpm guard` passes all 6 design-system checks on default + kami:
  - 26 A1 tokens declared in both brands
  - 26 A2 tokens declared in both brands
  - 4 B-slot tokens declared in both brands (default aliases via
    `var()`, kami binds independently — both forms satisfy the
    contract)
  - 129 total declarations across both brands all in schema or
    extension allowlists
  - 26 A2 fallbacks match `tokens.schema.ts` ↔ `defaults.css`
- `pnpm typecheck` passes (after `pnpm install` to refresh stale
  workspace dist; pre-existing on `main`, unrelated to this PR)
- Sanity tests on the guard checks: drifting any A2 default value in
  `defaults.css`, removing any A1 / A2 token from a brand, or adding
  an unknown name all fail the appropriate sub-check with a precise
  divergence message

## Explicitly NOT in this PR

- **Daemon prompt injection** — `apps/daemon/src/design-systems.ts`
  doesn't read these files yet; the prompt-assembly side lands in a
  follow-up so we can revise schema based on review feedback without
  cascading into runtime code
- **Bulk coverage of the other ~138 brands** — needs a derivation
  script that reads each `DESIGN.md` and emits a `tokens.css` draft
  for human review. Will be PR-B; `_schema/defaults.css` is its
  designated input for A2 fallbacks
- **`components.html` for tier-1 brands beyond default/kami** —
  `atelier-zero` and `warm-editorial` are the other hand-authored
  brands referenced by skills; their fixtures land in a follow-up

## Test plan

- `pnpm install && pnpm guard && pnpm typecheck` clean
- Open `design-systems/default/components.html` in a browser —
  visual check passes (B2B/utility feel; cobalt CTA; cards lift via
  contrast not shadow); CTAs now land on `tokens.css` and
  `DESIGN.md` source views
- Open `design-systems/kami/components.html` in a browser —
  visual check passes (parchment bg; serif throughout; ink-blue
  section numbers; no italic; no cool grays); CTAs land on real files
- Spot-check `design-systems/_schema/AGENTS.md` for the C → B-slot →
  A2 promotion path narrative
- Verify `_schema/defaults.css` matches the `fallback` field of every
  A2 entry in `_schema/tokens.schema.ts`
- (optional) Run `pnpm exec tsx scripts/check-tokens-fixture-sync.ts`
  directly to confirm all 5 sub-checks pass and report token counts


mrcfps (commented):
Thanks for putting these fixtures together — I checked the new `tokens.css` / `components.html` pairs and found two non-blocking navigation issues in the reference HTML that are worth tightening before other contributors start using these files as copy/paste sources.


design-systems/default/components.html
```diff
@@ -0,0 +1,523 @@
...
+          <div class="stack-4">
+            <p class="eyebrow">Reference fixture · default</p>
+            <h1>One source for the values your team keeps re-deciding.</h1>
+            <p class="lead" style="max-width: 52ch">
+              A starter pack that turns brand decisions into shippable
+              tokens — colors, type, spacing — so every interface
+              inherits the same answer instead of reaching for a default.
+            </p>
+            <div class="hero-actions">
+              <a href="#tokens" class="btn btn-primary">
```
mrcfps: This fixture advertises several in-page destinations (`#tokens`, `#spec`, and later `#surface`, `#accent`, `#type`), but the file never defines matching `id`s for any of them. As written, those CTAs/cards are all no-ops when someone opens `components.html` locally, which undercuts the point of shipping it as a navigable reference artifact. I'd either add concrete target sections with those ids or point the links at the real resources (`./tokens.css`, `./DESIGN.md`) so every call to action resolves.

design-systems/kami/components.html
```diff
@@ -0,0 +1,601 @@
...
+        <p class="lede">
+          A token system distilled from the kami print rules: warm
+          parchment canvas, single ink-blue accent capped under five
+          percent of the page, serif at one weight. The fixture you
+          are reading uses the same token block agents paste into
+          every artifact.
+        </p>
+        <div class="hero-actions">
+          <a href="#tokens" class="btn btn-primary">View tokens</a>
+          <a href="#components" class="btn btn-secondary">
```
mrcfps: The hero CTA targets do not line up with real destinations here: `#tokens` has no matching section in this file, and `Read the spec` jumps to `#components` instead of the actual spec. That makes the primary navigation misleading for anyone using this fixture as the canonical kami reference. Please either add matching in-page sections/ids or point these links directly at `./tokens.css` and `./DESIGN.md` so the buttons land on the resources they advertise.

lefarcen (commented):
Thanks for the structured schema work — this addresses a real agent token-misuse pain point.

mrcfps already flagged the navigation issues in the reference fixtures. Beyond that, I have one reasoning-completeness observation:

**Schema generalization risk (non-blocking):** The three-layer model (required / optional / brand-specific) is derived from just 2 brands with intentionally contrasting aesthetics. There's a risk the optional-vs-required split won't generalize cleanly when brand #3 needs different Layer A tokens or when multiple brands converge on the same extension (promoting C→B→A). Consider surfacing that limitation in the PR narrative or in a future SCHEMA.md — e.g., "the promotion path from C→B→A will be defined when ≥2 brands need the same extension."

The navigation fix mrcfps mentioned is the only blocking item. The rest is solid.

mrcfps (dismissed):
@zoeforfun I reviewed the new default/kami token fixtures and the token-fixture drift guard. I verified the changed ranges stay asset/script-scoped, the new guard only compares the unscoped `:root` block so kami's optional `:root[lang=...]` overrides remain valid, and I didn't find additional correctness or maintainability issues in the current head beyond the existing navigation feedback. Thanks for pushing the schema work forward — this gives the design-system handoff a much clearer contract ✨


scripts/check-tokens-fixture-sync.ts
```diff
@@ -0,0 +1,459 @@
...
+        `[${brand}] ${toRepositoryPath(tokensPath)} is missing ${missing.length} A2 token${missing.length === 1 ? "" : "s"} (default values exist in design-systems/_schema/defaults.css; copy or override):\n  ${missing.join(", ")}`,
+      );
+    }
+  }
+
+  const passed = reportFailure(
+    "Design system A2 required tokens",
+    violations,
+    "A2 tokens carry sensible cross-brand defaults but artifacts paste a single :root block — agents that paste a tokens.css missing an A2 declaration will produce broken artifacts. Every brand's tokens.css must declare every A2 token (until the derive script lands and inlines fallbacks automatically).",
+  );
```
mrcfps, line 324: The schema docs in this PR say every final `tokens.css` must declare **A1 + A2 + B-slot** names so shared components can safely read aliases like `var(--fg-2)` and `var(--surface-warm)`, but this guard only checks `getRequiredA2Names()` and there is no companion B-slot check. That means a future brand can omit all four B-slot aliases, still pass `pnpm guard`, and then any pasted artifact that references those shared slots will resolve them to nothing at runtime. Please either expand this check or add a sibling `getBSlotNames()` guard so the machine-enforced contract matches the schema narrative before more brands are derived from it.

mrcfps (changes requested):
@zoeforfun Thanks for pushing the token-schema work forward — I found one contract gap in the new guard that should be closed before other brands start depending on it.


mrcfps (dismissed):

scripts/check-tokens-fixture-sync.ts
```diff
@@ -0,0 +1,459 @@
...
+  if (passed) {
+    console.log(
+      `Design system A2 required tokens passed: ${sources.length} brand${sources.length === 1 ? "" : "s"} declare all ${requiredA2.length} A2 tokens.`,
+    );
+  }
+  return passed;
+}
```
lefarcen, line 331: P1 — The guard validates A2 required tokens here, but there's no sibling check for B-slot aliases (`--fg-2`, `--meta`, `--surface-warm`, `--border-soft`). Per the schema docs, every brand must declare A1 + A2 + B-slot names so shared components can safely read `var(--fg-2)` etc. Without a B-slot guard, a brand can omit those aliases, pass `pnpm guard`, and break any artifact that references them. Please add a `getBSlotNames()` validation similar to the A2 check above, or explicitly document why B-slot is optional at runtime (and update the schema narrative to match).

lefarcen (changes requested):
Thanks for expanding the schema narrative with the layer promotion rules — that addresses the generalization concern from my last review.

mrcfps's blocking finding from the prior head is still present: the guard checks `getRequiredA2Names()` but has no companion B-slot validation. See inline.

scripts/check-tokens-fixture-sync.ts
lefarcen, line 331: P1 — The guard validates A2 required tokens here, but there's no sibling check for B-slot aliases (`--fg-2`, `--meta`, `--surface-warm`, `--border-soft`). Per the schema docs, every brand must declare A1 + A2 + B-slot names so shared components can safely read `var(--fg-2)` etc. Without a B-slot guard, a brand can omit those aliases, pass `pnpm guard`, and break any artifact that references them. Please add a `getBSlotNames()` validation similar to the A2 check above (the function exists in `design-systems/_schema/tokens.schema.ts:259`), or explicitly document why B-slot is optional at runtime (and update the schema narrative to match).

lefarcen (changes requested):
Thanks for expanding the schema narrative with the layer promotion rules — that addresses the generalization concern from my last review.

mrcfps's blocking finding from the prior head is still present: the guard checks `getRequiredA2Names()` but has no companion B-slot validation. See inline.

lefarcen (approved):
Thanks for adding the B-slot guard — the blocking finding from b4cf2039 is now resolved.

The new `checkDesignSystemBSlotRequiredTokens` correctly validates all four B-slot aliases (`--fg-2`, `--meta`, `--surface-warm`, `--border-soft`), mirrors the A2 check structure, and provides helpful hints pointing to the schema-suggested alias when a slot is missing. Verified that the function is properly registered in the check list at line 498 and `getBSlotNames()` exists in the schema (line 259).

Regarding the test fixture heads-up to @pftom — that's a helpful catch on the missing `designTemplates` prop from #955. The one-line fix in 50ddb3c5 looks correct and unblocks typecheck on this PR.

PR #1182, opened 2026-05-11
feat: add success feedback after saving API keys

xxiaoxiong:
Fixes #654

## Problem
After clicking "Save key" in the Composio section of Settings, the button immediately returns to its default state without providing clear confirmation that the key was saved successfully. Users have to reopen Settings or look for indirect signals to verify the save worked.

## Solution
Add explicit success feedback to the "Save key" button:
- Show "Saved ✓" state for 2 seconds after successful save
- Display check icon alongside success message
- Add 'success' to keySaveStatus state machine
- Button automatically returns to idle state after 2s

## Changes
1. **SettingsDialog.tsx**:
   - Update keySaveStatus type: 'idle' | 'saving' | 'success' | 'error'
   - Modify handleSaveKey to set 'success' state with 2s timeout
   - Add success state rendering in button UI with check icon
   - Add 'is-success' CSS class for styling

2. **i18n/locales/en.ts**:
   - Add 'settings.connectorsKeySaved': 'Saved' translation

3. **i18n/types.ts**:
   - Add 'settings.connectorsKeySaved' to Dict type

## User Experience
**Before:** Save key → [button returns to "Save key"] → user unsure

**After:** Save key → ["Saving..."] → ["Saved ✓" for 2s] → ["Save key"]

This matches the pattern used in the Media Providers reload button and provides clear, immediate feedback that the save succeeded.

## Testing
1. Open Settings → Composio
2. Enter a Composio API key
3. Click "Save key"
4. Observe:
   - Button shows "Saving..." with spinner
   - Button changes to "Saved" with check icon
   - After 2 seconds, button returns to "Save key"
5. Verify the key is actually saved (check the "Saved · ••••XXXX" badge appears)

PR #1232, opened 2026-05-11
fix(web): suppress autosave indicator for draft-only Connector key edits

### LINKED ISSUE #1187: Connector API key draft conflicts with saved-state UI and should require explicit confirmation
## Summary

In the Connectors settings, entering a new API key draft creates conflicting saved-state signals. The UI still shows the previously saved key in the main field area, while the top-right status says **All changes saved** for the newly typed value, even though the saved key has not actually changed yet.

## Current behavior

After typing a new value such as `111` into the API key field:

- The middle section still shows the previously saved key badge.
- The top-right status shows **All changes saved**.
- The saved-key area still reads as if the old key remains the active saved value.

This creates a contradiction between the draft input state and the saved-state messaging.

## Expected behavior

Typing a new API key should not be treated as automatically saved.

- The UI should clearly distinguish between the currently saved key and an unsaved draft.
- The global saved-state indicator should not say **All changes saved** while a replacement key is only being typed.
- Saving a replacement key should require an explicit confirmation action such as a dedicated **Save key** button.

## Steps to reproduce

1. Open Settings > Connectors.
2. Go to the Composio API key field.
3. Ensure an existing key is already saved.
4. Type a new value such as `111` into the input.
5. Compare the saved-key badge, the input state, and the global saved-state message.

## Notes

- A screenshot was provided showing the conflicting UI states.
- The core issue is not just copywriting; it is a mismatch between saved-state semantics and input-draft semantics.
- The requested behavior is explicit confirmation rather than auto-save.

## Why this matters

This creates strong cognitive conflict around whether sensitive credentials have actually been updated. For API keys, users need clear and trustworthy save semantics, especially when replacing an existing saved key.

Image

Eggwardhan:
Fixes #1187.

## Symptom

After typing a replacement value (e.g. `111`) into the Composio API key field:

- The middle section correctly keeps the saved-key badge (this was already fixed in #741 / PR #751).
- But the top-right global indicator advances through `pending → saving → saved` and shows **"All changes saved"** even though no replacement key has actually been persisted.

For sensitive credentials this is the worst possible signal: it tells users their key was saved when only a draft was typed, leaving them to discover the truth later when actions silently fail with the stale key.

## Root cause

`SettingsDialog`'s autosave effect persists `cfg` on every committed edit. `App.tsx`'s `handleConfigPersist` runs `buildPersistedConfig(cfg)` which **intentionally strips `composio.apiKey`** before disk and daemon writes (the secret is committed only through the section-local "Save key" gesture, see `App.tsx:93-105` and the comment at `SettingsDialog.tsx:1201-1206`). The strip works correctly — no secret leaks to disk — but the autosave indicator transitions on the basis of "did `onPersist` resolve?" rather than "did anything actually change?". A no-op save still resolves successfully, so the indicator flashes "Saved".

## Fix

Detect autosave ticks where the only delta since the last successful save is a field that `buildPersistedConfig` strips, and skip them.

1. **`apps/web/src/App.tsx`** — export `isAutosaveDraftOnlyChange(next, last)`. Returns `true` when `next` and `last` produce identical persisted shapes (i.e. `JSON.stringify(buildPersistedConfig(next, next)) === JSON.stringify(buildPersistedConfig(last, last))`). Built on top of the existing single source of truth for "what gets written" — no new strip list to drift out of sync.
2. **`apps/web/src/components/SettingsDialog.tsx`** — the autosave effect now tracks `autosaveLastSavedRef` (initial cfg on mount → updated on each successful persist). Inside the debounced callback, before calling `onPersist`, check `isAutosaveDraftOnlyChange(snapshot, autosaveLastSavedRef.current)`. When true and no forced media-provider sync is pending, settle the indicator to `'idle'` and return early — no daemon write, no "Saved" flash.

The forced media-provider sync path still runs unchanged, because that is a real outbound effect that must fire even when the persisted shape hasn't changed.

## What this preserves

- The section-local "Save key" button remains the only way to commit a new Composio key (unchanged).
- The saved-key badge / `apiKeyConfigured` / `apiKeyTail` state machine in `ConnectorSection` is untouched.
- Non-draft edits (model, mode, agent, baseUrl, mediaProviders, ...) still trigger the autosave indicator normally.
- Media-provider sync forced by version bumps still runs even on draft-only ticks.

The change generalizes: any future field that `buildPersistedConfig` adds to the strip list will inherit the same draft-only protection without further wiring.

## Tests

`apps/web/tests/App.test.ts` — 4 new cases for `isAutosaveDraftOnlyChange`:
- Typing a Composio API key → treated as draft-only ✓
- Changing a non-draft field (model) → persist-worthy ✓
- Flipping `apiKeyConfigured` / clearing `apiKeyTail` → persist-worthy ✓
- Identical snapshot → draft-only (no-op tick) ✓

10/10 tests pass in `apps/web/tests/App.test.ts` locally.

- [ ] `pnpm --filter @open-design/web test` green
- [ ] Manual: open Settings → Connectors → Composio, type a value, observe the top-right indicator stays idle (no "All changes saved" flash); clicking "Save key" still works and the indicator correctly updates after that explicit save.
- [ ] Manual: change a non-Composio field (e.g. switch model), confirm the indicator still flashes "Saved" as before.

## Out of scope (left to follow-ups)

- Surfacing a section-local "Unsaved key draft" badge on the Composio input (suggested in the issue) — the existing `ConnectorSection` already has its own draft state via the "Save key" button gating; a dedicated copy-string change is a separate UX iteration that should coordinate with #1101 / #1182 (#654 save-success feedback work referenced in the issue triage).

Refs #1187.

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -104,6 +104,26 @@ export function buildPersistedConfig(next: AppConfig, current: AppConfig): AppCo
   };
 }
 
+/**
+ * True when `next` and `last` produce an identical persisted shape —
+ * i.e. the only diffs between them are fields that buildPersistedConfig
+ * intentionally strips before disk/daemon writes (the Composio API key
+ * draft today; any future save-on-explicit-confirm secrets later).
+ *
+ * The autosave loop in Settings uses this to skip the "All changes
+ * saved" indicator transition when the user has only typed an unsaved
+ * secret. Without it, autosave completes a no-op write and flashes
+ * "Saved" — misleading users into trusting that a sensitive key has
+ * been persisted when in fact only the section-local "Save key"
+ * gesture commits it.
+ */
+export function isAutosaveDraftOnlyChange(next: AppConfig, last: AppConfig): boolean {
+  return (
+    JSON.stringify(buildPersistedConfig(next, next))
+    === JSON.stringify(buildPersistedConfig(last, last))
+  );
+}
+
 export function resolveSettingsCloseConfig(
   rendered: AppConfig,
   latestPersisted: AppConfig,
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -59,6 +59,7 @@ import {
   applyAppearanceToDocument,
   normalizeAccentColor,
 } from '../state/appearance';
+import { isAutosaveDraftOnlyChange } from '../App';
 import {
   FAILURE_SOUNDS,
   SUCCESS_SOUNDS,
@@ -1217,6 +1218,13 @@ export function SettingsDialog({
   const autosaveRetryTimerRef = useRef<number | null>(null);
   const autosavePendingFlushRef = useRef(false);
   const autosaveLatestRef = useRef<AppConfig>(cfg);
+  // Baseline used by the draft-only detector: the snapshot at the most
+  // recent successful autosave (or the initial cfg on mount). Compared
+  // against the current snapshot to decide whether the only edits
+  // since last save are intentionally-stripped fields like the
+  // Composio API key — in which case we must NOT flash "All changes
+  // saved", because the draft has not actually been persisted.
+  const autosaveLastSavedRef = useRef<AppConfig>(cfg);
   const mediaProvidersChangeVersionRef = useRef(0);
   const lastSyncedMediaProvidersVersionRef = useRef(0);
   const [autosaveRetryTick, setAutosaveRetryTick] = useState(0);
@@ -1224,6 +1232,7 @@ export function SettingsDialog({
   useEffect(() => {
     if (autosaveSkipFirstRef.current) {
       autosaveSkipFirstRef.current = false;
+      autosaveLastSavedRef.current = cfg;
       return;
     }
     setAutosaveStatus('pending');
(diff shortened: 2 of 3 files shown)


mrcfps (approved):
@Eggwardhan I verified the draft-only autosave path end to end: the new persisted-shape comparison in `isAutosaveDraftOnlyChange` reuses the existing `buildPersistedConfig` contract, and `SettingsDialog` now skips the misleading global saved transition when only a Composio key draft changed while still preserving the forced media-provider sync path. The added App-level tests cover the draft-only and persist-worthy cases I would expect here, and this looks good to me. Thanks for tightening the saved-state semantics around a sensitive credential flow 🙏


lefarcen (commented):
Hey @Eggwardhan! Reviewed alongside @mrcfps's APPROVED on the current head.

**Lens A (code correctness)** — clean:
- `isAutosaveDraftOnlyChange` is the right shape: reuses the existing `buildPersistedConfig` single source of truth, so no duplicate strip-list to drift. Future fields in the strip-list inherit draft-only protection automatically.
- Autosave skip (line 1268-1273) correctly preserves the forced media-provider sync path — that's a real outbound effect even when the persisted shape hasn't changed.
- Baseline `autosaveLastSavedRef` is updated only on successful persist (line 1278), not on every snapshot. Correct.
- Test coverage: 4 cases (draft-only, persist-worthy, clear/identical) are solid.

**Lens B (reasoning completeness)** — also clean:
- Symptom/root-cause/fix/preserves/tests/verification breakdown in the PR body is thorough. No gaps.
- One alternative not discussed: could have stored a separate `keyHasDraft` flag vs. detecting it post-hoc with shape comparison. Your chosen approach is more robust (can't drift out of sync with the strip list).
- The "Out of scope" section explicitly names the follow-up UX iteration (section-local draft badge) — that's clean handoff to the next iteration.

No blocking issues from my side. Looks good! 🎉

lefarcen:
Merged! 🎉 Thanks @Eggwardhan for the contribution to open-design. Looking forward to the next one!

PR #1238, opened 2026-05-11
fix(pack/win): close running app before silent reinstall

### LINKED ISSUE #1118: why cant i install open design?
Image

I have the issue where i cant install opencode from the installer. it keeping show me the error like i show in the image

PerishCode:
## Issue

Fixes #1118.

Users can hit a Windows silent reinstall failure when Open Design is already running. The installer tries to replace the existing install directory while desktop/sidecar child processes still hold files under that directory, so NSIS fails during remove/extract. In the reproduced failure, the direct `setup.exe /S /D=<installDir>` path exited with code `2` and `nsis.log` showed `install dir remove exit=32` followed by `payload extraction exit=2`.

## Root Cause

`e2e/specs/win.spec.ts` did not catch this because the old packaged Windows smoke only used `tools-pack win install`. That command preemptively runs the tools-pack uninstall/stop path before invoking the installer, so the NSIS installer itself was never tested against a running app.

The Windows custom installer did have running-process detection/closing logic, but it embedded a large PowerShell program directly inside `nsExec::ExecToStack`. That command is fragile in NSIS because of command length and quoting, so detection could fail or return empty output. Silent install then proceeded into file removal/extraction while Open Design processes were still alive.

## Fix

- Move running-instance detection/close into a generated `running-instances.ps1` helper that is bundled into the installer plugin directory and invoked through a short `powershell.exe -File ...` command.
- Make silent install detect, close, and re-detect running Open Design processes before file changes. If detection fails or processes remain, the installer aborts instead of continuing into a locked directory.
- Add a second pre-file-change guard in the install section so late-spawned processes are still caught.
- Keep the process kill scope constrained: the helper no longer receives caller-supplied `$INSTDIR` directly. The registered install location is used only after the expected app exe exists there, and `$INSTDIR` is added only after `$INSTDIR\<app exe>` exists.
- Extend the Windows packaged smoke with a direct `setup.exe /S /D=<installDir>` reinstall while the packaged app is running, then restart and verify health.

## Correctness / Safety

The new e2e path covers the user-shaped failure mode rather than the tools-pack-managed happy path: install, start, wait for healthy desktop, run the generated installer directly while processes are alive, verify installer success and close logs, then restart and inspect health again.

The process matching still uses install-root path prefixes so it closes the desktop process and Electron/sidecar children even when they do not have visible windows, but the roots are now validated before they are allowed to become a close scope. This avoids treating an arbitrary `/D=C:\Program Files` style target as permission to close unrelated applications.

## Validation

- `pnpm --filter @open-design/e2e typecheck`
- `pnpm exec tools-pack win build --namespace issue1118 --to nsis --json`
- `OD_PACKAGED_E2E_WIN=1 OD_PACKAGED_E2E_NAMESPACE=issue1118 pnpm --filter @open-design/e2e test specs/win.spec.ts`
- `pnpm guard`
- `pnpm typecheck`

### DIFF
--- a/tools/pack/src/win/custom-installer.ts
+++ b/tools/pack/src/win/custom-installer.ts
@@ -89,6 +89,80 @@ async function resolveMakensisCommand(config: ToolPackConfig): Promise<string> {
   throw new Error("makensis is required to build the Windows installer; install NSIS or populate the electron-builder NSIS cache");
 }
 
+function createRunningInstancesScript(): string {
+  return `param(
+  [ValidateSet("detect", "close")]
+  [string]$Action,
+  [string]$Install,
+  [string]$Registered
+)
+
+$ErrorActionPreference = "Stop"
+
+$roots = @($Install, $Registered) |
+  Where-Object { $_ } |
+  ForEach-Object {
+    $root = $_.TrimEnd([char]92).ToLowerInvariant()
+    [pscustomobject]@{ Exact = $root; Prefix = ($root + [char]92) }
+  } |
+  Select-Object -Unique Exact, Prefix
+
+$matches = Get-CimInstance Win32_Process | Where-Object {
+  $matched = $false
+  $exe = $_.ExecutablePath
+  if ($null -ne $exe) {
+    $exe = $exe.ToLowerInvariant()
+    foreach ($root in $roots) {
+      if ($root.Exact -and (($exe -eq $root.Exact) -or $exe.StartsWith($root.Prefix))) {
+        $matched = $true
+        break
+      }
+    }
+  } else {
+    $cmd = $_.CommandLine
+    if ($null -ne $cmd) {
+      $cmdLc = $cmd.ToLowerInvariant()
+      foreach ($root in $roots) {
+        if ($root.Prefix -and $cmdLc.Contains($root.Prefix)) {
+          $matched = $true
+          break
+        }
+      }
+    }
+  }
+  $matched
+}
+
+$ids = @($matches | ForEach-Object { $_.ProcessId })
+if ($Action -eq "close") {
+  foreach ($id in $ids) {
+    try { [void][System.Diagnostics.Process]::GetProcessById($id).CloseMainWindow() } catch {}
+  }
+  Start-Sleep -Milliseconds 1500
+  foreach ($id in $ids) {
+    try {
+      $p = [System.Diagnostics.Process]::GetProcessById($id)
+      if (-not $p.HasExited) {
+        Stop-Process -Id $id -Force -ErrorAction SilentlyContinue
+      }
+    } catch {}
+  }
+  foreach ($id in $ids) {
+    try {
+      $p = [System.Diagnostics.Process]::GetProcessById($id)
+      if (-not $p.HasExited) {
+        [void]$p.WaitForExit(5000)
+      }
+    } catch {}
+  }
+}
+
+if ($ids) {
+  $matches | ForEach-Object { [string]$_.ProcessId + [char]32 + $_.Name }
+}
+`;
+}
+
 async function writeInstallerScript(config: ToolPackConfig, paths: WinPaths): Promise<void> {
   const identity = resolveWinInstallIdentity(config);
   const productName = escapeNsisString(identity.displayName);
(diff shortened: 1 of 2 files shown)


tools/pack/src/win/custom-installer.ts
```diff
@@ -89,6 +89,80 @@ async function resolveMakensisCommand(config: ToolPackConfig): Promise<string> {
...
+  } |
+  Select-Object -Unique Exact, Prefix
+
+$matches = Get-CimInstance Win32_Process | Where-Object {
+  $matched = $false
+  $exe = $_.ExecutablePath
+  if ($null -ne $exe) {
+    $exe = $exe.ToLowerInvariant()
+    foreach ($root in $roots) {
+      if ($root.Exact -and (($exe -eq $root.Exact) -or $exe.StartsWith($root.Prefix))) {
```
mrcfps, line 116: Blocking: `createRunningInstancesScript` now treats the caller-supplied `$Install` path as a trusted kill scope, and both `.onInit` and `GuardRunningInstancesBeforeInstall` call `CloseRunningInstances` before proving that `$INSTDIR` is actually an existing Open Design install. In this PR the silent path explicitly accepts `/D=...`, so a reinstall launched with a broad directory like `C:\Program Files` will match any process whose executable or command line lives under that tree and can force-close unrelated third-party apps before `IfFileExists "$INSTDIR\${exeName}"` runs. That is a merge-blocking safety regression compared with the old behavior, which only aborted on a match. Please gate the close path on a validated Open Design target first — for example, only include `$Install` in `$roots` after confirming the expected exe/registry identity exists there, or restrict pre-install shutdown to `$Registered` until the install directory has been verified.

mrcfps (changes requested):
@PerishCode Thanks for tightening the Windows reinstall path — I found one blocker that should be fixed before this lands. The inline comment covers the concrete risk and a safe direction for the fix.


tools/pack/src/win/custom-installer.ts
PerishCode: Addressed in `17838318`: `DetectRunningInstances` / `CloseRunningInstances` no longer receive caller-supplied `$INSTDIR` directly. `.onInit` validates the registered install location by checking the expected app exe before using it, interactive pre-directory handling only uses that validated registered location, and `GuardRunningInstancesBeforeInstall` includes `$INSTDIR` only after `$INSTDIR\${exeName}` exists. Re-ran the Windows direct reinstall smoke plus `pnpm guard` and `pnpm typecheck`.

tools/pack/src/win/custom-installer.ts
```diff
@@ -305,9 +375,15 @@ Function .onInit
 silent_check:
   Call DetectRunningInstances
   \${If} $RunningInstancesOutput != ""
-    Push "install aborted: running instances detected: $RunningInstancesOutput"
+    Push "running instances detected before silent install: $RunningInstancesOutput"
     Call LogInstallerEvent
-    Abort "$(RunningInstancesSilentAbort)"
+    Call CloseRunningInstances
```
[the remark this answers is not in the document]
PerishCode: Addressed in `17838318`: the PowerShell helper now only receives trusted roots from NSIS. `$INSTDIR` is copied into the close/detect scope only after the expected app exe exists there; the registered location is also validated before use. Re-ran the Windows direct reinstall smoke and full checks.

PerishCode:
Updated and re-pushed on latest `origin/main` as `7648a97b`.

Follow-up for the blocker:
- `DetectRunningInstances` / `CloseRunningInstances` no longer receive caller-supplied `$INSTDIR` directly.
- Registered install location is validated by checking the expected app exe before it is used as a process scope.
- `$INSTDIR` is added to the process scope only inside `GuardRunningInstancesBeforeInstall`, after `$INSTDIR\<app exe>` exists.

Revalidated after the rebase:
- `pnpm --filter @open-design/e2e typecheck`
- `pnpm exec tools-pack win build --namespace issue1118 --to nsis --json`
- `OD_PACKAGED_E2E_WIN=1 OD_PACKAGED_E2E_NAMESPACE=issue1118 pnpm --filter @open-design/e2e test specs/win.spec.ts`
- `pnpm guard`
- `pnpm typecheck`

mrcfps (approved):
@PerishCode I rechecked the updated Windows installer flow and the new packaged smoke coverage on the latest head. The kill scope is now validated before shutdown, silent reinstall re-checks for lingering processes before file changes, and the direct `setup.exe /S /D=<installDir>` smoke exercises the locked-files path this PR is fixing. Thanks for following through on the safety fix and the end-to-end coverage here 🙏

PR #1242, opened 2026-05-11
fix(settings): suppress autosave flicker on Composio key drafts

### LINKED ISSUE #1187: Connector API key draft conflicts with saved-state UI and should require explicit confirmation

mvanhorn:
## Summary

Typing a new Composio API key into the Connectors settings caused the global "All changes saved" indicator to flash for every keystroke even though the draft is intentionally stripped before anything reaches localStorage or the daemon. The contradiction the issue describes - the saved-key badge still showing the old key, the user typing a replacement, and the top-right pill claiming the new value is already saved - lands directly on this autosave loop.

Closes #1187

## Why this matters

The save semantics for an API key need to be unambiguous. The explicit "Save key" gesture in `ConnectorSection` is the only path that actually persists a Composio key (see also `buildPersistedConfig` in `App.tsx`, which drops the secret before localStorage), so the global autosave indicator firing on each keystroke is straightforwardly wrong for that field. With the indicator flicker gone the existing `saved-pending` state already does the right thing on the credentials surface: the saved-key badge stays anchored while the input shows the unsaved draft, and "Save key" is the only thing that flips the badge to a new value.

## Change

`apps/web/src/components/SettingsDialog.tsx`

Adds `autosaveDiffIsMeaningful(prev, next)`. It strips the in-flight `composio.apiKey` from both sides before comparing, and also collapses a `composio` block whose only field is that draft. The autosave effect now compares the freshly committed `cfg` against the last snapshot that actually fired and skips the `pending -> saving -> saved` cycle when the only diff is a draft key. The helper is exported so the existing test suite can pin its behaviour.

Real changes still fire as before: `apiKeyConfigured` flipping to `true` or `apiKeyTail` echoing back from the daemon after a successful "Save key" call both count as meaningful diffs, so the "just saved" confirmation still appears for genuine writes - just not for typing.

The `saved-pending` credential state and the `Save key` button are unchanged. The behaviour the issue asks for ("Saving a replacement key should require an explicit confirmation action such as a dedicated Save key button") was already in place; this PR removes the false signal that conflicted with it.

## Tests

`apps/web/tests/components/SettingsDialog.test.ts` gets a new `autosaveDiffIsMeaningful` describe block with 8 unit tests covering the initial null snapshot, identity, non-composio field changes, the draft-key-only diff that motivates the fix, the `apiKeyConfigured` and `apiKeyTail` paths (daemon-echoed changes that MUST still fire), the "no composio block -> draft-only block" transition, and the mixed draft + real-field case. All 52 tests in the file pass on the new code. `pnpm --filter @open-design/web typecheck` exits clean.

## What I did not change

- `deriveComposioCredentialState` and the `saved-pending` UI state are kept as-is; they already render the right thing once the global indicator stops contradicting them.
- The `Save key` button, the two-stage destructive Clear confirmation, and the daemon round-trip in `persistComposioConfigChange` are untouched.
- No copy or i18n changes; the original `settings.autosaveSaved` string is correct - it just should not fire for draft keys.

## Risk

Low. The helper is a pure comparison function with explicit unit tests for every edge case. The autosave effect's other deps (`onPersist`, `autosaveRetryTick`) and the unmount-flush path are unchanged.

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -1220,12 +1220,22 @@ export function SettingsDialog({
   const mediaProvidersChangeVersionRef = useRef(0);
   const lastSyncedMediaProvidersVersionRef = useRef(0);
   const [autosaveRetryTick, setAutosaveRetryTick] = useState(0);
+  // Snapshot of the last cfg that triggered a real autosave cycle. Used
+  // to distinguish a meaningful edit from a Composio-draft-only diff so
+  // the indicator does not flash "All changes saved" while the user is
+  // still typing a replacement key that the daemon has not received.
+  const lastAutosavedCfgRef = useRef<AppConfig | null>(null);
   autosaveLatestRef.current = cfg;
   useEffect(() => {
     if (autosaveSkipFirstRef.current) {
       autosaveSkipFirstRef.current = false;
+      lastAutosavedCfgRef.current = cfg;
       return;
     }
+    if (!autosaveDiffIsMeaningful(lastAutosavedCfgRef.current, cfg)) {
+      return;
+    }
+    lastAutosavedCfgRef.current = cfg;
     setAutosaveStatus('pending');
     if (autosaveSavedTimerRef.current != null) {
       window.clearTimeout(autosaveSavedTimerRef.current);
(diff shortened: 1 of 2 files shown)


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1220,12 +1220,22 @@ export function SettingsDialog({
...
   useEffect(() => {
     if (autosaveSkipFirstRef.current) {
       autosaveSkipFirstRef.current = false;
+      lastAutosavedCfgRef.current = cfg;
       return;
     }
+    if (!autosaveDiffIsMeaningful(lastAutosavedCfgRef.current, cfg)) {
+      return;
+    }
+    lastAutosavedCfgRef.current = cfg;
```
mrcfps, line 1238: This new early return short-circuits reruns that are supposed to happen without a config diff. In this effect React runs the previous cleanup before the new body, so if the user changes a real setting, gets the 400 ms autosave timer, and then types in the Composio key before the debounce fires, the cleanup at `1306-1311` clears that pending timer and this branch returns without scheduling a replacement because `lastAutosavedCfgRef` still points at the last persisted snapshot. The same guard also breaks the `autosaveRetryTick` path: after a failed media-provider sync the retry reruns the effect with the same `cfg` object, `prev === next` at `2616`, and no retry timer is scheduled. That can drop real edits and disable the existing retry behavior. Could we move the draft-only suppression later in the flow, or compare against the previous render's sanitized config while still preserving already-queued saves and retry-triggered reruns?

mrcfps (changes requested):
Thanks for digging into this — the direction makes sense, but I hit one blocking regression in the autosave scheduling path.


mvanhorn:
Closing while I investigate the failing `settings-api-protocol` Playwright run — will reopen once I can reproduce and verify the autosave change isn't the cause.

lefarcen (commented):
Thanks for tackling this — the direction makes sense, and mrcfps's review already covers the blocking scheduling regression at line 1235.

Two additional observations:

**P2 Test coverage gap**: The new unit tests exercise `autosaveDiffIsMeaningful` as a pure function, which is good, but they don't cover the actual effect behavior — specifically the debounced timer cancellation scenario mrcfps described. A component/effect-level test that simulates (1) real edit → 400ms timer starts, (2) draft key input before timer fires → cleanup cancels, (3) verify a replacement timer is scheduled — would catch the regression and prevent it from reappearing.

**P3 Comment vs persistence mismatch** (reasoning completeness, non-blocking): The comment at line 2620 says a draft-only composio block is "functionally equivalent to no composio block" because "nothing would be written to localStorage or the daemon." However, `buildPersistedConfig` in `apps/web/src/App.tsx:93` would serialize a sanitized composio object when `next.composio` exists, even if empty after stripping `apiKey`. The behavior may still be acceptable as a product decision, but the reasoning should acknowledge this difference rather than claim exact persistence-layer equivalence.

Overall: once the scheduling fix from mrcfps's review lands, this will be in good shape. The test gap and comment precision are nice-to-haves but not merge blockers.

PR #1244, opened 2026-05-11
[codex] Handle empty API responses as no output

alchemistklk:
## Summary

Fixes the misleading completed state for API-mode streams that end cleanly without returning text or an artifact.

The root cause was that `ProjectView` treated `onDone('')` as a successful assistant turn after flushing the stream buffers. In API mode, the proxy can legitimately emit a clean `end` event with no deltas, so the UI showed `Done` / `已完成` even though no useful output was produced.

## Changes

- Detect empty API completions after text/artifact parser flushing.
- Persist a soft `empty_response` assistant status event plus a diagnostic text message.
- Keep `ChatRunStatus` unchanged and avoid marking the turn as `succeeded` or `failed` for this soft state.
- Teach the assistant footer to render `No output` / `无输出` / `無輸出` for the soft state.
- Keep normal success behavior unchanged for API text responses and streamed artifact responses.
- Add component coverage and a Playwright UI smoke test for the real API proxy stream path.

## Validation

- `pnpm --filter @open-design/web exec vitest run tests/components/ProjectView.api-empty-response.test.tsx tests/components/assistant-message-unfinished-todos.test.tsx`
- `pnpm guard`
- `pnpm typecheck`
- `PATH=/Users/alche/.nvm/versions/node/v24.0.0/bin:$PATH OD_PORT=18457 OD_WEB_PORT=18574 OD_E2E_NAMESPACE=api-empty-response-smoke-node24 pnpm exec playwright test -c playwright.config.ts ui/api-empty-response.test.ts` from `e2e/`

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -1133,6 +1133,7 @@ export function ProjectView({
 
       const parser = createArtifactParser();
       let liveHtml = '';
+      let streamedText = '';
 
       const updateAssistant = (updater: (prev: ChatMessage) => ChatMessage) => {
         setMessages((curr) =>
@@ -1242,7 +1243,10 @@ export function ProjectView({
       abortRef.current = controller;
       cancelRef.current = cancelController;
       const handlers = {
-        onDelta: textBuffer.appendContent,
+        onDelta: (delta: string) => {
+          streamedText += delta;
+          textBuffer.appendContent(delta);
+        },
         onAgentEvent: (ev: AgentEvent) => {
           if (ev.kind === 'text') textBuffer.appendTextEvent(ev.text);
           else pushEvent(ev);
@@ -1259,6 +1263,7 @@ export function ProjectView({
           const emptyApiResponse =
             config.mode === 'api' &&
             !fullText.trim() &&
+            !streamedText.trim() &&
             !liveHtml.trim();
           if (emptyApiResponse) {
             const diagnostic = t('assistant.emptyResponseMessage');
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -1276,6 +1276,9 @@ export function ProjectView({
               true,
               { telemetryFinalized: true },
             );
+            if (commentAttachments.length > 0) {
+              void patchAttachedStatuses(commentAttachments, 'failed');
+            }
             setStreaming(false);
             abortRef.current = null;
             cancelRef.current = null;
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -1247,7 +1247,7 @@ export function ProjectView({
           if (ev.kind === 'text') textBuffer.appendTextEvent(ev.text);
           else pushEvent(ev);
         },
-        onDone: () => {
+        onDone: (fullText = '') => {
           textBuffer.flush();
           textBuffer.cancel();
           cancelSendTextBuffer();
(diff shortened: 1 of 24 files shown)


apps/web/src/components/ProjectView.tsx
```diff
@@ -1256,6 +1256,33 @@ export function ProjectView({
               setArtifact((prev) => (prev ? { ...prev, html: ev.fullContent } : null));
             }
           }
+          const emptyApiResponse =
+            config.mode === 'api' &&
+            !fullText.trim() &&
+            !liveHtml.trim();
+          if (emptyApiResponse) {
+            const diagnostic = t('assistant.emptyResponseMessage');
```
mrcfps, line 1269: `emptyApiResponse` returns before this turn settles any attached preview comments. Earlier in `handleSend`, the PR moves every persisted `commentAttachment` to `applying` (`patchAttachedStatuses(commentAttachments, 'applying')` at lines 1110-1112), and the normal success/error branches flip them back to `needs_review` or `failed` (lines 1291-1293 and 1336-1337). This new early-return path skips that cleanup, so an empty API completion leaves those comments stuck in `applying` even though the run is over and the UI now shows `No output`. Please mirror the success/error cleanup here—at minimum, settle attached comments to a terminal status before returning, and add coverage for an empty-response turn that starts with comment attachments.

mrcfps (commented):
@alchemistklk Thanks for tightening up the empty API-response UX. I found one follow-up in the new no-output branch around comment-attachment cleanup.


mrcfps (approved):
@alchemistklk I rechecked the empty-response flow on the current head, including the API proxy completion path, the attached-comment cleanup branch, and the new component/E2E coverage. This now keeps empty API completions in the intended soft "No output" state without regressing normal text/artifact success handling.

Thanks for the careful follow-up here — the latest revision closes the earlier cleanup gap nicely. 🙌


apps/web/src/components/ProjectView.tsx
```diff
@@ -1256,6 +1260,37 @@ export function ProjectView({
...
+                events: [
+                  ...(prev.events ?? []),
+                  { kind: 'status', label: 'empty_response', detail: config.model },
+                  { kind: 'text', text: diagnostic },
+                ],
+              }),
+              true,
+              { telemetryFinalized: true },
+            );
+            if (commentAttachments.length > 0) {
```
lefarcen, line 1284: **P1 — Comment attachment cleanup missing** (agreeing with @mrcfps)

The `emptyApiResponse` early return leaves `commentAttachments` stuck in `'applying'` status. Earlier (lines 1110-1112) the code moves them to `'applying'`, and the normal success/error branches settle them (lines 1291-1293, 1336-1337).

You added:
```ts
if (commentAttachments.length > 0) {
  void patchAttachedStatuses(commentAttachments, 'failed');
}
```
at lines 1283-1285, which is correct for `emptyApiResponse`. But this path returns *before* the normal cleanup. Consider moving the `patchAttachedStatuses` call to happen *before* the early return (line 1290) so cleanup always runs.

Also add test coverage: an `emptyApiResponse` turn that starts with `commentAttachments.length > 0` should settle those comments to a terminal state, not leave them `'applying'`.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1133,6 +1133,7 @@ export function ProjectView({
 
       const parser = createArtifactParser();
       let liveHtml = '';
+      let streamedText = '';
```
lefarcen, line 1136: **P2 — `streamedText` accumulator initialization**

The `streamedText` variable is declared here as `let streamedText = '';`, and `onDelta` appends to it. But if the stream is restarted (e.g., user cancels + retries), this variable is not re-initialized — it will accumulate deltas across retries.

Consider resetting `streamedText = ''` on each `handleSend` invocation, or initializing it inside the `runAgent` call scope. Otherwise `emptyApiResponse` may false-negative if prior attempts streamed text.

apps/web/tests/components/ProjectView.api-empty-response.test.tsx
```diff
@@ -0,0 +1,350 @@
...
+  await waitFor(() => expect(screen.getByRole('button', { name: 'send' })).toBeTruthy());
+  fireEvent.click(screen.getByRole('button', { name: 'send' }));
+}
+
+function hasSavedAssistantMessage(predicate: (message: ChatMessage) => boolean): boolean {
+  return mockedSaveMessage.mock.calls.some((call) => {
+    const message = call[2] as ChatMessage;
+    return message.role === 'assistant' && predicate(message);
+  });
+}
```
lefarcen, line 350: **P2 — Test coverage for comment attachments**

The new test suite covers `emptyApiResponse` well, but doesn't exercise the `commentAttachments` cleanup path (lines 1283-1285). Please add a test case:

```ts
it('settles attached comments to failed on empty API response', async () => {
  // setup: create assistant message with commentAttachments in 'applying'
  // trigger: empty stream end
  // assert: commentAttachments are 'failed', not 'applying'
});
```

This will catch the cleanup regression mrcfps flagged.

lefarcen (changes requested):
Hey @alchemistklk! Thanks for tightening the empty API response UX. I found the same comment-attachment cleanup issue mrcfps spotted, plus a few additional edge cases around the new `emptyApiResponse` path.

**Must-fix (P1):**
- Comment attachment cleanup (per mrcfps) — see inline below

**Non-blocking suggestions (P2):**
- `streamedText` initialization risk
- Test coverage for the new branch

Details inline ↓

lefarcen (approved):
All prior findings addressed on this head. The comment-attachment cleanup path is now complete, streamedText initialization is fixed, and test coverage covers the new empty-response branch. Clean to proceed. 🎉

PR #1249, opened 2026-05-11
fix(desktop): exit fullscreen before hiding window on macOS close

### LINKED ISSUE #1215: mac 端预览模式下点击 演示-->全屏，进入全屏演示后，点击关闭左上角 x 关闭，会出现黑屏
Image

Image

PerishCode:
## Summary
- darwin close handler now leaves any active fullscreen before hiding so macOS doesn't orphan the Space as a black screen (#1215)
- bootstrap Vitest on `apps/desktop` and add the first unit test covering the new `hideWindowExitingFullscreen` helper
- root `AGENTS.md` gains a soft-toned "Bug follow-up workflow" subsection codifying the spec-first / cheapest-layer / scope-discipline / invariant-shaped-fix / baseline-diff playbook used on #135 and this fix

## Details
- Bug path: in 演示 → 全屏 the renderer calls `requestFullscreen()`, which on Electron darwin transitions the BrowserWindow into a native fullscreen Space. Clicking the red ✕ fired the existing `'close'` handler, which `preventDefault()`'d and called `window.hide()` directly — but on macOS, hiding while the Space is still up leaves the Space sitting empty and black until the user switches Spaces by hand.
- Fix routes the close path through a named helper that, when the window is in `isSimpleFullScreen()` or `isFullScreen()`, registers a one-shot `'leave-full-screen'` listener and calls `setSimpleFullScreen(false)` / `setFullScreen(false)`. The hide is deferred until the OS confirms the Space teardown, so the window and the Space dismiss together.
- Helper is exported with a structural `WindowFullscreenSurface` type so the test exercises pure logic (sequence of calls) without spinning up a real Electron window.
- AGENTS.md addition is intentionally soft-toned: framed as a \"default action shape, not a contract\" with explicit room for case-by-case judgment — codification of recent practice, not a process rule.

## Validation
- \`pnpm --filter @open-design/desktop test\` — red against the bootstrap helper that just called \`hide()\`, green after the fullscreen-exit sequencing (3/3)
- \`pnpm --filter @open-design/desktop typecheck\` — clean
- \`pnpm --filter @open-design/e2e test specs\` — green (3 passed, 5 platform-skipped)
- \`pnpm --filter @open-design/e2e test tests\` — green (12/12, includes the #135 regression specs)
- \`pnpm typecheck\` / \`pnpm guard\` — clean

Fixes #1215

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -97,6 +97,7 @@ The following is a working playbook for routine bug follow-ups, distilled from r
 - **Hold the spec's scope.** Defects discovered outside the bug's described boundary belong in a follow-up — their own red spec, their own PR — not in this fix. List them in the PR body's "Adjacent issues" section with the rationale and move on.
 - **Let the fix read as an invariant.** Prefer a named helper whose docblock describes what must hold over a bolt-on `if` guard with apologetic history-comments. The call site should read as intent.
 - **Diff against the baseline.** When neighboring suites have pre-existing failures, stash or check out upstream before claiming "no new failures."
+- **Stage human verification for visible bugs.** When the symptom needs an eye to confirm — UI, platform-native behavior, animations, race conditions a unit test can't see — green specs alone aren't acceptance. Stand up a buggy-vs-fix comparison the reviewer can drive themselves (typical shape: two namespaced runtimes, one on `main`, one on the fix branch), and seed any required data only through production HTTP APIs; source-level test backdoors invalidate the verification because they prove a fake flow rather than the real one.
 
 For a worked example of one full loop (red e2e spec → fix → green), see `e2e/tests/dialog/stop-reconciles-message.test.ts` (issue #135).
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -88,6 +88,18 @@ This file is the single source of truth for agents entering this repository. Rea
 - Stamp/namespace changes must validate two concurrent namespaces and run desktop `inspect eval` plus `inspect screenshot` for each namespace.
 - Path/log changes must run `pnpm tools-dev logs --namespace <name> --json` and confirm log paths are under `.tmp/tools-dev/<namespace>/...`.
 
+## Bug follow-up workflow
+
+The following is a working playbook for routine bug follow-ups, distilled from recent practice. Treat it as a default action shape, not a contract — production reality always has edges these bullets can't anticipate, so use judgment when the situation doesn't fit cleanly.
+
+- **Lead with a red spec.** Default to encoding the bug as a falsifiable test that goes red before any source change, so the fix is anchored in observable behavior rather than source-code intuition. If a red spec can't be written cheaply, that's usually a signal to clarify scope rather than push forward on a guess.
+- **Try the cheapest layer first.** Reach for the lightest test layer that can still see the symptom (e2e Vitest at the daemon HTTP boundary → app-local Vitest → Playwright UI → platform-native harnesses), and drop down only when the cheaper layer can't.
+- **Hold the spec's scope.** Defects discovered outside the bug's described boundary belong in a follow-up — their own red spec, their own PR — not in this fix. List them in the PR body's "Adjacent issues" section with the rationale and move on.
+- **Let the fix read as an invariant.** Prefer a named helper whose docblock describes what must hold over a bolt-on `if` guard with apologetic history-comments. The call site should read as intent.
+- **Diff against the baseline.** When neighboring suites have pre-existing failures, stash or check out upstream before claiming "no new failures."
+
+For a worked example of one full loop (red e2e spec → fix → green), see `e2e/tests/dialog/stop-reconciles-message.test.ts` (issue #135).
+
 # Common commands
 
 ```bash
--- a/apps/desktop/src/main/runtime.ts
+++ b/apps/desktop/src/main/runtime.ts
@@ -632,22 +632,31 @@ function ensureWindowVisible(window: BrowserWindow): void {
  * {@link hideWindowExitingFullscreen} — declared structurally so the
  * helper can be exercised with a plain mock in unit tests without
  * standing up an actual Electron window.
+ *
+ * `isEnteringFullscreen` covers the Electron-asynchronous gap between
+ * the renderer asking for fullscreen and the OS confirming via
+ * 'enter-full-screen': the caller is expected to track this from the
+ * window's enter/leave listeners (see the close handler in
+ * {@link configureWindow}) and surface it here.
  */
 export type WindowFullscreenSurface = {
   hide: () => void;
   isFullScreen: () => boolean;
   isSimpleFullScreen: () => boolean;
+  isEnteringFullscreen: () => boolean;
   setFullScreen: (flag: boolean) => void;
   setSimpleFullScreen: (flag: boolean) => void;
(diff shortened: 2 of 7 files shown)


apps/desktop/src/main/runtime.ts
```diff
@@ -627,6 +627,42 @@ function ensureWindowVisible(window: BrowserWindow): void {
...
+ * teardown and the user is left staring at a black desktop until they
+ * switch Spaces by hand.
+ */
+export function hideWindowExitingFullscreen(window: WindowFullscreenSurface): void {
+  if (window.isSimpleFullScreen()) {
+    window.once('leave-full-screen', () => window.hide());
+    window.setSimpleFullScreen(false);
+    return;
+  }
+  if (window.isFullScreen()) {
```
mrcfps, line 667: `hideWindowExitingFullscreen` still falls back to `window.hide()` if the user clicks the red close button while macOS is *entering* fullscreen. Electron documents `isFullScreen()`/`leave-full-screen` as asynchronous on macOS, so until `enter-full-screen` fires this guard can still read `false` even though the Space transition is already in flight. In that timing window we take the plain `window.hide()` path and can still strand the black fullscreen Space that #1215 is trying to eliminate. Consider tracking a pending fullscreen transition from `enter-full-screen`/`leave-full-screen` (or the corresponding HTML fullscreen events that `requestFullscreen()` drives) and treating `close` as deferred-hide whenever fullscreen is active **or entering**; adding a regression test for "close during fullscreen-enter" would lock the fix down.

mrcfps (commented):
@PerishCode Thanks for tightening up the macOS close path and adding focused desktop coverage here. I found one timing-sensitive follow-up in `hideWindowExitingFullscreen` that looks worth addressing before this fullscreen-close path is fully robust.


apps/desktop/src/main/runtime.ts
PerishCode: Good catch — addressed in 425de3b6. The async `enter-full-screen` window is exactly what we were missing.

Wired the close handler to track `enteringFullscreen` from `webContents.enter-html-full-screen` (which fires before the OS confirms) and clears it on the matching `enter-full-screen` / `leave-full-screen` events. Surfaced via a new `isEnteringFullscreen` on `WindowFullscreenSurface`; helper now parks on `enter-full-screen` and only starts the exit-then-hide sequence after the OS confirms the Space.

Added a "waits out a fullscreen-enter transition before exiting and hiding" test that goes red against the previous helper. Acknowledged residual gap: title-bar green button / programmatic `setFullScreen(true)` don't have a pre-event analogous to `enter-html-full-screen`, so the same race window theoretically exists for those paths — narrower than the DOM API path the issue reported, but worth flagging if you'd prefer me to add coverage.

mrcfps (approved):
@PerishCode Thanks for tightening up the macOS close path here. I reviewed the fullscreen-enter follow-up, the darwin event sequencing, and the new desktop unit coverage; the updated helper now waits out the enter transition before exiting fullscreen, and I didn't find any concrete correctness, safety, or maintainability issues in the changed ranges. Nice job closing the race and documenting the behavior clearly.


apps/desktop/tests/main/hide-window-exiting-fullscreen.test.ts
```diff
@@ -0,0 +1,143 @@
+// Issue #1215 — clicking the macOS red X while the preview is in fullscreen
+// presentation mode hides the window but leaves the OS Space stuck as a
+// black screen. The close handler is allowed to call window.hide(), but it
+// must first arrange for the window to leave fullscreen so the Space tears
+// down with the window.
+//
+// This spec exercises the contract through hideWindowExitingFullscreen with
```
lefarcen, line 7: P3 — This wording overstates the test coverage. The spec verifies mock call sequencing for `hideWindowExitingFullscreen`, but it does not exercise Electron's real macOS fullscreen events, Space teardown, or timing around native `enter-full-screen`/`leave-full-screen`. That is fine as a unit test, but future readers should not infer this proves the OS-level #1215 behavior without the manual/platform verification the PR body describes elsewhere.

apps/desktop/src/main/runtime.ts
```diff
@@ -627,6 +627,58 @@ function ensureWindowVisible(window: BrowserWindow): void {
...
+ * Surface of {@link BrowserWindow} consumed by
+ * {@link hideWindowExitingFullscreen} — declared structurally so the
+ * helper can be exercised with a plain mock in unit tests without
+ * standing up an actual Electron window.
+ *
+ * `isEnteringFullscreen` covers the Electron-asynchronous gap between
+ * the renderer asking for fullscreen and the OS confirming via
+ * 'enter-full-screen': the caller is expected to track this from the
+ * window's enter/leave listeners (see the close handler in
+ * {@link configureWindow}) and surface it here.
```
lefarcen, line 640: P3 — `{@link configureWindow}` points to a function that does not exist in this file. Since this helper is exported primarily to make the close-handler invariant readable/testable, a stale link makes the API harder to follow; refer to `createDesktopRuntime` or the darwin close handler instead.

AGENTS.md
```diff
@@ -88,6 +88,19 @@ This file is the single source of truth for agents entering this repository. Rea
...
+- **Hold the spec's scope.** Defects discovered outside the bug's described boundary belong in a follow-up — their own red spec, their own PR — not in this fix. List them in the PR body's "Adjacent issues" section with the rationale and move on.
+- **Let the fix read as an invariant.** Prefer a named helper whose docblock describes what must hold over a bolt-on `if` guard with apologetic history-comments. The call site should read as intent.
+- **Diff against the baseline.** When neighboring suites have pre-existing failures, stash or check out upstream before claiming "no new failures."
+- **Stage human verification for visible bugs.** When the symptom needs an eye to confirm — UI, platform-native behavior, animations, race conditions a unit test can't see — green specs alone aren't acceptance. Stand up a buggy-vs-fix comparison the reviewer can drive themselves (typical shape: two namespaced runtimes, one on `main`, one on the fix branch), and seed any required data only through production HTTP APIs; source-level test backdoors invalidate the verification because they prove a fake flow rather than the real one.
```
lefarcen, line 100: P2 — The new workflow makes a fairly strong process recommendation for visible bugs, but it does not spell out tradeoffs or exceptions for platform-native bugs like this one where a two-runtime HTTP-seeded comparison may not actually observe the native macOS Space behavior. Add the missing assumption/escape hatch: when the symptom is OS/window-manager-level, require explicit platform verification or document why a cheaper test is enough.

lefarcen (commented):
Thanks for tightening up the darwin close-handler invariant and bootstrapping test coverage on `apps/desktop`. I found a couple of small doc/test-coverage notes below — non-blocking, but worth clarifying before this pattern becomes the default.

**Code review (Lens A):**
The fix directly addresses the black-fullscreen-Space symptom from #1215 by tracking `enter-html-full-screen` and deferring hide until the native Space teardown completes. I verified the new logic against the surrounding window lifecycle code and did not find a blocking correctness or security issue.

**Reasoning review (Lens B):**
The AGENTS.md addition makes a fairly broad process recommendation for all visible bugs but does not spell out tradeoffs or exceptions for platform-native issues where the cheaper HTTP-seeded comparison may not actually observe the OS-level behavior. See inline suggestion below.

PR #1255, opened 2026-05-11
fix(web): route chat file links to workspace preview instead of new window (#1239)

Sid-Qin:
## Summary

Closes #1239 (P0). Chat-emitted markdown links like `[template.html](template.html)` or `[hero](subdir/hero.html)` rendered as `<a target=\"_blank\">` with no `onClick`. In Electron that hits `setWindowOpenHandler` and creates a new `od://` BrowserWindow; a relative href has no base so the new window can't resolve it and lands on the home screen — the file the assistant pointed at is never previewed. This PR routes those relative file links through the workspace tab opener that already lives in ProjectView, while leaving external `http(s)://` URLs (and other schemes / fragments / absolute paths) on their default browser-link behavior.

## Approach

ProjectView already exposes `requestOpenFile(name)` (apps/web/src/components/ProjectView.tsx:569) and threads it down to ChatPane → AssistantMessage as the `onRequestOpenFile` prop. Tool cards and produced-file chips have used it for a while. The PR plugs chat-markdown `<a>` clicks into the same path.

```diff
ProseBlock (apps/web/src/components/AssistantMessage.tsx)
+ const onLinkClick = useMemo<MarkdownLinkClickHandler | undefined>(() => {
+   if (!onRequestOpenFile) return undefined;
+   return (href, event) => {
+     const path = asInProjectFilePath(href);
+     if (!path) return;  // external / anchor / abs / traversal → default behavior
+     event.preventDefault();
+     onRequestOpenFile(path);
+   };
+ }, [onRequestOpenFile]);

- {renderMarkdown(seg.text)}
+ {renderMarkdown(seg.text, { onLinkClick })}
```

Three pieces:

1. **`apps/web/src/runtime/in-project-link.ts`** (new) — pure helper `asInProjectFilePath(href): string | null` documenting and gating the intercept-vs-pass-through decision in one place.
2. **`apps/web/src/runtime/markdown.tsx`** — `renderMarkdown(input, options?)` now takes `RenderMarkdownOptions` with an optional `onLinkClick` callback. Threaded through `renderBlock` / `renderInline` / `pushText` so **all three** `<a>` render sites get the same intercept hook (explicit `[text](url)` links, bare-URL autolinks inside inline scans, and the autolink branch in `pushText`). When the option is omitted, render output is byte-identical to before — every existing caller (file viewer, system reminders, …) keeps working without changes.
3. **`apps/web/src/components/AssistantMessage.tsx`** — wires the existing `onRequestOpenFile` prop into that hook via the helper.

### What gets intercepted

| href | Result |
|---|---|
| `template.html` | → `onRequestOpenFile('template.html')` + `preventDefault()` |
| `./template.html` | → `onRequestOpenFile('template.html')` |
| `subdir/hero.html` | → `onRequestOpenFile('subdir/hero.html')` |
| `template.html?v=2#section` | → `onRequestOpenFile('template.html')` |
| `https://docs.example.com` | pass-through, opens externally |
| `mailto:foo@bar` | pass-through |
| `#section` | pass-through |
| `/abs/path` | pass-through |
| `..` / `../sibling.html` / `a/b/../c.html` | pass-through (refuse to leave project root via relative navigation) |

The pass-through cases come from real-world chat content (doc links, OAuth flows, bare URLs) and from defensive checks (`..` traversal, absolute paths).

## Reproduce

1. Open a project with chat that includes file links — typical Orbit / artifact-generating sessions produce them automatically; the issue screenshot shows `template.html`, `index.html`, `orbit-daily-digest-general-2026-05-11.html`.
2. Click one of those links in the chat.

**Before this PR**: a new Electron app window opens and lands on the home screen. The file is never previewed.
**After this PR**: the file opens in the right-pane Design Files workspace and the user stays in the current project.

## Verification

- `pnpm --filter @open-design/contracts build && pnpm --filter @open-design/web typecheck` — clean (exit 0)
- `pnpm --filter @open-design/web test` — **87 files, 785/785 passed**, includes the 33 new tests below
- Direct RED→GREEN sanity check: `git stash push apps/web/src/runtime/markdown.tsx apps/web/src/components/AssistantMessage.tsx` → re-run `AssistantMessage.linkClick.test.tsx` → the two `expect(onRequestOpenFile).toHaveBeenCalled…` cases fail, the four pass-through invariants still pass. `git stash pop` → all 6 pass again.

## New tests

| File | Cases | What it pins |
|---|---|---|
| `apps/web/tests/runtime/in-project-link.test.ts` | 22 | Every documented intercept / pass-through shape, including the exact filename pattern from the issue screenshot (`orbit-daily-digest-general-2026-05-11.html`). |
| `apps/web/tests/runtime/markdown.linkClick.test.tsx` | 5 | `onLinkClick` contract across all three `<a>` render sites; backwards-compat when `options` is omitted; `preventDefault()` propagation through the React synthetic event boundary. |
| `apps/web/tests/components/AssistantMessage.linkClick.test.tsx` | 6 | Integration: relative file link intercepts through `onRequestOpenFile` with normalized path (`./subdir/hero.html` → `subdir/hero.html`); https / bare URL / `#fragment` fall through; missing `onRequestOpenFile` keeps default behavior. |

## Test plan

- [ ] CI green on the `apps/web` package
- [ ] Manual: reproduce the issue per the steps above (chat with file links → click → file previews in-app instead of opening a new window)

## Not in scope

- Other paths that emit clickable file-shaped strings (tool cards, produced-file chips, file panel) — those already use `onRequestOpenFile` directly and were never broken by this issue.
- `target="_blank"` on the rendered `<a>` is preserved. If a future feature wants to fully suppress the underlying default (e.g. middle-click-to-open-in-tab semantics for in-project files), it can extend `onLinkClick` further; the intercept contract is already in place.

### DIFF
--- a/apps/web/src/runtime/in-project-link.ts
+++ b/apps/web/src/runtime/in-project-link.ts
@@ -1,0 +1,62 @@
+/**
+ * Decide whether a markdown link href in chat output should resolve to an
+ * in-project file (opened in the right-pane workspace) or fall through to
+ * the default browser link behavior (Electron `setWindowOpenHandler` →
+ * new window).
+ *
+ * Chat output from the assistant frequently contains references like
+ * `[template.html](template.html)` or `[hero](subdir/hero.html)`. Those
+ * are relative paths into the current project's file workspace; with
+ * default `target="_blank"` they instead open a new Electron window with
+ * no project context and land on the home screen. Routing them through
+ * the existing `requestOpenFile` callback keeps the user in the same
+ * project view and previews the file in the right pane.
+ *
+ * Returns the normalized file path to open when the href looks like an
+ * in-project link, or `null` to let the default link behavior win.
+ *
+ * Pass-through (returns null):
+ *   - http(s):// and any other URL scheme (mailto:, ftp:, …) — explicit
+ *     external link, default browser behavior is correct.
+ *   - `#anchor` — fragment within the current document.
+ *   - Absolute paths starting with `/` — could mean filesystem root in
+ *     Electron and is not what the assistant is referencing; better to
+ *     not silently rewrite into an in-project open.
+ *   - `..` traversal — refuses to leave the project root via relative
+ *     navigation. (Anchors any normalization in the consumer to project-
+ *     scoped files only.)
+ *   - Empty / whitespace.
+ *
+ * Intercept (returns normalized path):
+ *   - Bare filename:        `template.html`        → `template.html`
+ *   - Explicit relative:    `./template.html`      → `template.html`
+ *   - Nested in subdir:     `subdir/hero.html`     → `subdir/hero.html`
+ *   - With query / hash:    `template.html?x=1#a`  → `template.html`
+ *     (query and fragment are stripped — the workspace tab opener takes
+ *     a file path, not a URL.)
+ */
+export function asInProjectFilePath(href: string | null | undefined): string | null {
+  if (typeof href !== 'string') return null;
+  const trimmed = href.trim();
+  if (!trimmed) return null;
+  if (trimmed.startsWith('#')) return null;
+  // RFC 3986 scheme: ALPHA *( ALPHA / DIGIT / "+" / "-" / "." ) followed
+  // by ":". Catches http:, https:, mailto:, file:, od:, blob:, etc.
+  if (/^[a-z][a-z0-9+.-]*:/i.test(trimmed)) return null;
+  if (trimmed.startsWith('/')) return null;
+  // Reject any segment that climbs out of the project root. Cheaper +
+  // safer than a full path normalization pass, and `..` segments are
+  // not something real assistant chat output emits for file references.
+  const segments = trimmed.split('/');
+  if (segments.some((segment) => segment === '..')) return null;
+  // Strip an explicit `./` prefix so the consumer sees the same shape
+  // it would have gotten from a bare filename.
+  const withoutDotSlash = trimmed.startsWith('./') ? trimmed.slice(2) : trimmed;
+  // Drop query and fragment — the workspace tab opener takes a file
+  // path, not a URL. Doing it in this order so `?` or `#` inside a
+  // filename (rare but legal) doesn't slip into the workspace key.
+  const beforeHash = withoutDotSlash.split('#')[0] ?? '';
+  const beforeQuery = beforeHash.split('?')[0] ?? '';
+  const normalized = beforeQuery.trim();
+  return normalized.length > 0 ? normalized : null;
+}
--- a/apps/web/src/components/AssistantMessage.tsx
+++ b/apps/web/src/components/AssistantMessage.tsx
@@ -1,6 +1,10 @@
 import { Fragment, useEffect, useMemo, useState } from "react";
 import { ToolCard } from "./ToolCard";
-import { renderMarkdown } from "../runtime/markdown";
+import {
+  renderMarkdown,
+  type MarkdownLinkClickHandler,
+} from "../runtime/markdown";
+import { asInProjectFilePath } from "../runtime/in-project-link";
 import { projectFileUrl } from "../providers/registry";
 import {
   splitOnQuestionForms,
@@ -118,6 +122,7 @@ export function AssistantMessage({
                   });
                   onSubmitForm?.(text);
                 }}
+                onRequestOpenFile={onRequestOpenFile}
               />
             );
           if (b.kind === "thinking")
@@ -435,6 +440,7 @@ function ProseBlock({
   nextUserContent,
   locallySubmitted,
   onSubmitForm,
+  onRequestOpenFile,
 }: {
   text: string;
   isLastAssistant: boolean;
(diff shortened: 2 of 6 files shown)


apps/web/src/runtime/in-project-link.ts
```diff
@@ -0,0 +1,62 @@
...
+  if (segments.some((segment) => segment === '..')) return null;
+  // Strip an explicit `./` prefix so the consumer sees the same shape
+  // it would have gotten from a bare filename.
+  const withoutDotSlash = trimmed.startsWith('./') ? trimmed.slice(2) : trimmed;
+  // Drop query and fragment — the workspace tab opener takes a file
+  // path, not a URL. Doing it in this order so `?` or `#` inside a
+  // filename (rare but legal) doesn't slip into the workspace key.
+  const beforeHash = withoutDotSlash.split('#')[0] ?? '';
+  const beforeQuery = beforeHash.split('?')[0] ?? '';
+  const normalized = beforeQuery.trim();
```
mrcfps, line 60: `asInProjectFilePath` is still treating the raw markdown href as the workspace file name, so percent-encoded file links will miss the preview path. `renderMarkdown` passes the literal href through, and `requestOpenFile` / `FileWorkspace` later match tabs by exact project file name, not a decoded URL form. That means a chat link like `[Mock page](Mock%20Page.html)` would call `onRequestOpenFile("Mock%20Page.html")`, while the real project file is `Mock Page.html`, so the chat click still won't open the file this PR is trying to surface. Can we decode the pathname portion before the `..` guard / normalization here (and fall back to pass-through on decode errors), then add a fixture for an encoded-space filename to lock the behavior down?

mrcfps (commented):
@Sid-Qin Thanks for shipping this fix — I verified the link interception wiring end-to-end and found one follow-up case around encoded file names in chat markdown.

apps/web/src/runtime/in-project-link.ts
```diff
@@ -0,0 +1,62 @@
...
+ *   - With query / hash:    `template.html?x=1#a`  → `template.html`
+ *     (query and fragment are stripped — the workspace tab opener takes
+ *     a file path, not a URL.)
+ */
+export function asInProjectFilePath(href: string | null | undefined): string | null {
+  if (typeof href !== 'string') return null;
+  const trimmed = href.trim();
+  if (!trimmed) return null;
+  if (trimmed.startsWith('#')) return null;
+  // RFC 3986 scheme: ALPHA *( ALPHA / DIGIT / "+" / "-" / "." ) followed
```
lefarcen, line 43: **P2** — Percent-encoded file names won't match workspace tabs. Chat markdown like `[Mock page](Mock%20Page.html)` will call `onRequestOpenFile("Mock%20Page.html")`, but the real file is `Mock Page.html`, so the tab lookup misses.

Fix: decode the pathname portion right after trimming and before the traversal check. Wrap in `try` / `catch` and fall back to `null` on decode errors:

```typescript
let decoded: string;
try {
  decoded = decodeURIComponent(trimmed);
} catch {
  // Malformed URI component (e.g. unmatched `%` or non-UTF-8)
  return null;
}
if (decoded.startsWith('#')) return null;
// … rest of checks use `decoded` instead of `trimmed`
```

Add a test fixture: `asInProjectFilePath('Mock%20Page.html')` → `'Mock Page.html'`.

lefarcen (commented):
Hey @Sid-Qin! Clean implementation of the workspace file-link routing. The helper + hook design is solid, test coverage is thorough, and the backwards-compat preservation for existing markdown callers is perfect.

@mrcfps caught the URL-encoding mismatch that I also spotted — chat markdown like `[Mock page](Mock%20Page.html)` will miss the workspace preview path because the raw `%20` doesn't match the actual file name `Mock Page.html` in the FileWorkspace tab registry. Fixing it is straightforward: decode the pathname portion before the traversal guard / normalization. Once that's addressed, this is ready.

P2 findings in inline comments below.

Siri-Ray (commented):
@Sid-Qin Thanks for the thoughtful fix here. I reviewed the changed ranges across `apps/web/src/runtime/markdown.tsx`, `apps/web/src/components/AssistantMessage.tsx`, `apps/web/src/runtime/in-project-link.ts`, and the new web tests; the routing approach looks sound overall.

Follow-up location: `apps/web/src/runtime/in-project-link.ts`, `asInProjectFilePath`, changed lines 38-61. I did not open a duplicate inline thread because the existing unresolved review thread already anchors this same case, but the helper still uses the raw markdown href as the workspace filename. That means chat markdown such as `[Mock page](Mock%20Page.html)` calls `onRequestOpenFile("Mock%20Page.html")`, while the workspace file name is `Mock Page.html`, so that click can still miss the file preview. Please decode the pathname portion before the traversal guard / normalization, return `null` on malformed escape sequences, and add a fixture such as `asInProjectFilePath('Mock%20Page.html') === 'Mock Page.html'`.

lefarcen:
Hey @Sid-Qin — quick follow-up on the reviews from @mrcfps and @Siri-Ray (and my own on May 11). All three of us flagged the same thing: `asInProjectFilePath` passes the raw encoded href straight through, so a chat link like `[Mock page](Mock%20Page.html)` calls `onRequestOpenFile('Mock%20Page.html')` instead of `onRequestOpenFile('Mock Page.html')` — the workspace tab lookup misses. The fix is small (decode the pathname before the traversal guard, return `null` on malformed escapes, add a fixture). Everything else is solid and CI is fully green.

Any ETA on that follow-up push? Happy to re-review as soon as it's up.

lefarcen:
Heads-up: PR #2576 is now open for the same #1239 fix and touches the same six files. Since this draft had gone stale, the newer PR appears to re-ground the workspace-preview link routing approach on current `main`; the maintainer team can decide which branch lands or how to consolidate.

lefarcen:
Heads-up: PR #2672 is now another open implementation touching the same #1239 chat-file-link path as this draft, including `apps/web/src/components/AssistantMessage.tsx` and `apps/web/src/runtime/markdown.tsx`. The newer PR has broader scope than this draft, so maintainers may compare it with #2576 and this PR before deciding which path lands.

Issue #564, opened 2026-05-05
Claude Code CLI exited with code 1, and the agent cannot be started in Open Design.


maxfrees:
Summary:
The Claude Code agent fails in Open Design. After sending a normal request, the Claude Code run immediately exits with code 1, and the agent is unable to complete the task.

Steps to Reproduce:

Start a project in Open Design.

Select Claude Code as the active agent.

Send a normal request, such as asking it to generate an image or complete a design task.

Observe the agent's result in the conversation panel.

Expected Behavior:
Claude Code should start normally, maintain a connection, and respond to user requests.

Actual Behavior:

Claude Code appears in the agent list and can be selected.

After a request is sent, the run exits almost immediately.

The interface shows: "Agent exited with code 1".

The agent returns no usable result.

Why This Is a Problem:

Claude Code seems available but is not actually usable.

Users cannot tell whether the failure is caused by the CLI setup, runtime invocation, or Open Design integration.

This breaks the core customer service workflow and creates confusion among different support teams.

maxfrees:
I have already installed it via npm.

lefarcen:
Hi @maxfrees! 👋 Thanks for reporting this.

`exit code 1` means Open Design was able to start the Claude Code process, but Claude itself exited immediately. The generic error code doesn't tell us why Claude rejected the request — it could be an auth/account issue, environment variable pollution, or a CLI version/flag incompatibility.

To help diagnose this, could you share:

1. **Claude Code version:** Run `claude --version` in your terminal
2. **Auth status:** Run `claude auth status --text` — does it succeed?
3. **Smoke test:** From the same project directory, try:
   ```bash
   printf 'hello' | claude -p --output-format stream-json --verbose --permission-mode bypassPermissions
   ```
   Does this work or fail?
4. **Daemon stderr:** Open Design forwards Claude's error output — if you see any stderr in the daemon logs or conversation panel beyond "Agent exited with code 1", please share that

This will help us pinpoint whether the issue is Claude CLI setup, Open Design integration, or something environment-specific.

Thanks for making open-design better! 🙏

maxfrees:
1. 2.1.128 (Claude Code)
2. Auth token: ANTHROPIC_AUTH_TOKEN
    Anthropic base URL: https://xxxx.com
    Proxy: http://127.0.0.1:7890

lefarcen:
Thanks for the details! That custom `ANTHROPIC_BASE_URL` + proxy setup is very likely the cause.

[...]

PR #1267, opened 2026-05-11
[codex] Improve Claude Code exit diagnostics

alchemistklk:
## Summary

- Add a daemon-side Claude Code diagnostic helper for common `exit code 1` auth, config, custom endpoint, and model-access failures.
- Surface that guidance through both chat run SSE errors and Settings agent connection tests without changing public response shapes.
- Document the known issue #564 recovery paths in QUICKSTART and record the scoped implementation plan under `specs/change`.

## Root cause

Issue #564 is not one single spawn bug. Claude Code can start successfully but then exit before producing output because the spawned non-interactive process is using stale auth, a different `CLAUDE_CONFIG_DIR`, or a custom `ANTHROPIC_BASE_URL`/proxy that cannot access the selected model. Open Design previously collapsed those cases into a generic `Agent exited with code 1` message.

## Validation

- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/claude-diagnostics.test.ts tests/connection-test.test.ts`
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts --testNamePattern "Claude auth diagnostics"`
- `pnpm guard`
- `pnpm typecheck`

Known unrelated existing failure when running the full chat-route file:

- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/connection-test.test.ts tests/chat-route.test.ts` fails in `tests/chat-route.test.ts > /api/chat > fails stalled json-stream runs after the inactivity timeout elapses` because the test expects an `event: agent` status frame before the inactivity error, but the observed stream contains `start`, `error`, and `end` only. The new Claude diagnostics chat-route test passes.

## Deferred

API/BYOK image-mode capability validation is intentionally deferred to a separate media-capability follow-up.


apps/daemon/src/connectionTest.ts
```diff
@@ -1142,6 +1143,27 @@ async function testAgentConnectionInternal(
       }
       const stderrTail = sink.getStderrTail().trim();
       const acpFatal = Boolean(acpSession?.hasFatalError?.());
+      const claudeDiagnostic = diagnoseClaudeCliFailure({
+        agentId: input.agentId,
+        exitCode: winner.code,
+        signal: winner.signal,
+        stderrTail,
+        stdoutTail: buffered,
```
mrcfps: `stdoutTail` is being populated from `buffered`, but in this code path `buffered` only contains assistant text that made it through `createAgentSink()`. For Claude runs, `attachAgentStreamHandlers()` sends `child.stdout` through `createClaudeStreamHandler()`, and that parser drops non-assistant JSON objects such as auth or model error records. That means a stream-json failure emitted on stdout — for example `{"apiKeySource":"none","error_status":401}` — still falls back to the generic `agent_spawn_failed` result, so the new Settings diagnostic misses one of the main failure shapes this PR is trying to classify. Please capture a raw Claude stdout tail alongside the parsed sink output here (the same way `server.ts` now keeps `agentStdoutTail`) and pass that raw tail into `diagnoseClaudeCliFailure()`.

apps/daemon/src/server.ts
```diff
@@ -3704,6 +3705,9 @@ export async function startServer({
     let child;
```
mrcfps, line 3730: Original requested location: apps/daemon/src/server.ts RIGHT line 3704

This sends the extra remediation under `details.detail`, but the web chat client still builds its visible error from `data.error?.message ?? data.message` in `apps/web/src/providers/daemon.ts` and drops `details.detail` entirely. In practice the chat UI will show the short `/login` or endpoint message, but it will not surface the new context this PR adds here (`CLAUDE_CONFIG_DIR`, custom-endpoint hints, raw Claude output), so users still cannot see the most actionable part of the diagnosis from the actual chat failure path. Please either fold the detail into the emitted SSE message here or plumb `details.detail` through the client-side error handling before relying on it for chat diagnostics.

mrcfps (commented):
Thanks for pushing on the Claude diagnostics work — I found two follow-up correctness gaps that are worth fixing before this lands.


apps/daemon/src/connectionTest.ts
Siri-Ray: `stdoutTail` is currently populated from `buffered`, but `buffered` is only the assistant text collected by `createAgentSink()`. For Claude, `attachAgentStreamHandlers()` routes stdout through `createClaudeStreamHandler()`, and that parser ignores non-assistant JSON objects such as `{"apiKeySource":"none","error_status":401}`. That means a Claude CLI failure emitted on stdout instead of stderr still bypasses `diagnoseClaudeCliFailure()` here and Settings falls back to the generic `agent_spawn_failed` detail, which misses one of the auth/model failure shapes this PR is meant to diagnose. Please capture a raw Claude stdout tail alongside the parsed sink output in the connection-test path, like `server.ts` now does, and pass that raw tail to the diagnostic helper. Add a connection-test fixture where the fake Claude process writes the 401 JSON to stdout and exits 1 so this path stays covered.

apps/daemon/src/server.ts
```diff
@@ -4112,6 +4125,23 @@ export async function startServer({
...
+          signal,
+          stderrTail: agentStderrTail,
+          stdoutTail: agentStdoutTail,
+          env: spawnedAgentEnv,
+        });
+        if (diagnostic) {
+          send('error', createSseErrorPayload(
+            'AGENT_EXECUTION_FAILED',
+            diagnostic.message,
+            { retryable: diagnostic.retryable, details: { detail: diagnostic.detail } },
```
Siri-Ray, line 4170: The actionable remediation is being placed under `error.details.detail`, but the web chat client still turns SSE errors into `new Error(String(data.error?.message ?? data.message ?? 'daemon error'))` in `apps/web/src/providers/daemon.ts`, so the chat UI drops this detail entirely. The changed line is where the new context is attached, and that context contains the pieces users need for the affected cases: effective `CLAUDE_CONFIG_DIR`, raw Claude output, and the custom-endpoint hint. As a result, a real chat failure can still show only a short message such as `Claude Code failed while using the configured Claude profile.` without the `/login` or profile-path guidance that explains how to fix it. Please either fold the diagnostic detail into the SSE `message` for chat failures or plumb `details.detail` through the web error handling, and cover a chat/client test that asserts the visible error includes the config/profile remediation.

Siri-Ray (commented):
I found two correctness gaps in the Claude diagnostic plumbing that still leave the new guidance unavailable on important user-facing paths.


lefarcen (approved):

PR #1268, opened 2026-05-11
Fix Tweaks Picker and Pods on unannotated HTML

PerishCode:
Fixes #903

## Summary
- Keep Tweaks Picker usable on plain generated HTML by falling back to a stable DOM selector when no `data-od-id` / `data-screen-label` target is available.
- Let Pods enumerate fallback DOM targets only when the page has no explicit annotation targets, preserving the narrower intended hit-test set for annotated artifacts.
- Keep Inspect mode strict so CSS override persistence still only targets explicitly annotated elements.
- Add bridge regression coverage for Inspect no-fallback, Picker fallback selection, and Pods fallback target broadcasts.

## Validation
- `pnpm --filter @open-design/web test -- srcdoc-bridge-empty-targets srcdoc file-viewer-render-mode FileViewer.inspect-empty-hint`
- `pnpm guard`
- `pnpm typecheck`
- `git diff --check`
- Manual desktop comparison with `tools-dev start desktop` namespaces `issue903-buggy` on `main` and `issue903-good` on this branch using the same unannotated `issue-903-plain.html` fixture.

### DIFF
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -448,9 +448,54 @@ function injectSelectionBridge(
       };
     } catch (_) { return null; }
   }
-  function targetFrom(el){
+  function annotatedSelectorFor(el){
     var id = el.getAttribute('data-od-id') || el.getAttribute('data-screen-label');
     if (!id) return null;
+    return el.hasAttribute('data-od-id') ? '[data-od-id="' + esc(id) + '"]' : '[data-screen-label="' + esc(id) + '"]';
+  }
+  function domSelectorFor(el){
+    if (!el || !el.tagName || el === document.documentElement || el === document.body) return null;
+    var parts = [];
+    var node = el;
+    while (node && node !== document.documentElement && node !== document.body) {
+      var tag = node.tagName ? node.tagName.toLowerCase() : '';
+      if (!tag || /^(script|style|template|meta|link|title|noscript)$/.test(tag)) return null;
+      var parent = node.parentElement;
+      if (!parent) return null;
+      var index = 1;
+      var sibling = node.previousElementSibling;
+      while (sibling) {
+        if (sibling.tagName && sibling.tagName.toLowerCase() === tag) index++;
+        sibling = sibling.previousElementSibling;
+      }
+      parts.unshift(tag + ':nth-of-type(' + index + ')');
+      node = parent;
+    }
+    if (!parts.length) return null;
+    return 'body > ' + parts.join(' > ');
+  }
+  function visibleTarget(el){
+    if (!el || !el.getBoundingClientRect) return false;
+    if (el === document.documentElement || el === document.body) return false;
+    if (/^(script|style|template|meta|link|title|noscript)$/.test(el.tagName ? el.tagName.toLowerCase() : '')) return false;
+    try {
+      var rect = el.getBoundingClientRect();
+      if (rect.width < 1 || rect.height < 1) return false;
+      var cs = window.getComputedStyle(el);
+      if (cs.display === 'none' || cs.visibility === 'hidden' || cs.pointerEvents === 'none') return false;
+    } catch (_) {
+      return false;
+    }
+    return true;
+  }
+  function targetFrom(el, allowDomFallback){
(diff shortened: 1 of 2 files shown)


apps/web/src/runtime/srcdoc.ts
```diff
@@ -501,16 +554,14 @@ function injectSelectionBridge(
   }
   function closestTarget(event){
     var el = event.target;
+    var fallback = null;
+    var allowDomFallback = commentEnabled && mode === 'picker' && !inspectEnabled;
     while (el && el !== document.documentElement) {
       if (el.getAttribute && (el.hasAttribute('data-od-id') || el.hasAttribute('data-screen-label'))) return el;
```
mrcfps, line 563: Problem: `closestTarget()` now enables DOM fallback for any picker-mode click, even when the page already has explicit `[data-od-id]` / `[data-screen-label]` targets elsewhere.

Why it matters: `allTargets()` only emits DOM fallback targets when `annotatedNodes.length === 0`, so on a mixed page this can select a `dom:` target that disappears on the next `od:comment-targets` refresh. That makes the picker/composer state unstable and breaks the narrower annotated-target contract this PR summary calls out.

Evidence: `allowDomFallback` here is derived only from `commentEnabled && mode === "picker" && !inspectEnabled`, and `return fallback` never checks whether the document already contains annotated nodes.

Suggested change: gate `closestTarget()` behind the same `annotatedNodes.length === 0` predicate used in `allTargets()` (or compute it once and pass it through to both call sites), then add a mixed annotated/unannotated regression test in `apps/web/tests/runtime/srcdoc-bridge-empty-targets.test.ts` so picker fallback stays limited to fully unannotated pages.

mrcfps (commented):
@PerishCode Thanks for pushing the unannotated-HTML picker fallback — I reviewed the bridge and the new regression coverage, and found one follow-up in the mixed annotated/unannotated picker path.


apps/web/src/runtime/srcdoc.ts
```diff
@@ -501,16 +554,14 @@ function injectSelectionBridge(
...
+      if (!fallback && allowDomFallback && visibleTarget(el)) fallback = el;
```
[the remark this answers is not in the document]
PerishCode: Fixed in `94e371fd`: `closestTarget()` now uses the same page-level no-annotations predicate as `allTargets()` via `canUseDomFallback()`, so mixed annotated/unannotated pages no longer create transient `dom:` picker targets. Added a mixed-page regression test covering the unannotated click path.

apps/web/src/runtime/srcdoc.ts
PerishCode: Fixed in `94e371fd`: Picker DOM fallback is now gated by the same whole-page no-annotation check used by the target broadcast, and the new mixed annotated/unannotated regression test pins that behavior.

mrcfps (approved):
@PerishCode I re-checked the latest head and verified the bridge now keeps picker DOM fallback limited to fully unannotated pages, still broadcasts fallback targets for pod hit-testing on those pages, and keeps Inspect mode constrained to explicitly annotated elements. The added regression coverage also locks the mixed annotated/unannotated path that previously drifted. Thanks for tightening this up and adding the follow-up tests.


lefarcen (commented):
The fix is solid and addresses mrcfps's mixed-page concern correctly. Two non-blocking items for future hardening:

**P2** — `domSelectorFor()` edge case: the selector builder assumes `node.parentElement` is always non-null after the first iteration, but if an element is somehow disconnected (unlikely in normal DOM but possible in test mocks or fragments), the loop could produce partial selectors or skip the null-parent check. Add explicit `if (!parent) return null;` inside the while loop for defensive programming.

**P3** — Performance on large unannotated pages: when fallback is active, `allTargets()` switches to `querySelectorAll('body *')` (line 518), which selects every element under body. On HTML with thousands of elements, iterating them all + calling `getBoundingClientRect()` for each (via `visibleTarget()`) could make Pods/Picker UI unresponsive. Consider limiting fallback to a reasonable count (e.g., first 200 visible elements) or adding a budget guard.

Neither blocks this PR — the current implementation handles typical artifact sizes fine and mrcfps's core concern is resolved. LGTM for internal merge; these are good follow-up hardening items.

PR #1271, opened 2026-05-11
Fix conversation run isolation

nettee:
## Summary
- Scope chat run busy/loading state to the active conversation so one unfinished conversation no longer blocks another.
- Add regression coverage for conversation switching, loading-window send gating, duplicate empty-conversation prevention, and daemon active-run filtering.
- Mark the Zest spec implemented with verification notes.

Closes #138

## Verification
- pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/ProjectView.run-isolation.test.tsx tests/components/ProjectView.run-cleanup.test.tsx tests/components/ProjectView.pendingPrompt.test.tsx
- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/runs.test.ts
- pnpm guard
- pnpm typecheck


apps/web/src/components/ProjectView.tsx
```diff
@@ -341,6 +342,18 @@ export function ProjectView({
...
+    () => messages.some((m) => m.role === 'assistant' && isActiveRunStatus(m.runStatus)),
+    [messages],
+  );
+  const currentConversationLoading = Boolean(
+    activeConversationId && messagesConversationId !== activeConversationId,
+  );
+  const currentConversationBusy = currentConversationLoading || streaming || currentConversationHasActiveRun;
+
+  useEffect(() => {
+    setStreaming(false);
```
mrcfps: Resetting `streaming` on every conversation switch drops the only `true -> false` edge that the completion-notification effect watches a few lines below (`prevStreamingRef` in `ProjectView.tsx:495-549`). If a run finishes while I'm viewing another conversation, we detach the SSE listener here, force `streaming` to `false`, and later come back to a terminal run without ever firing the success/failure notification. That means background completions now go silent even though this change is specifically about preserving detached background runs. Can we keep a separate "listener attached" flag here, or move the notification trigger onto run-status transitions instead of clearing `streaming` during conversation switches?

apps/web/src/components/ProjectView.tsx
```diff
@@ -401,10 +415,21 @@ export function ProjectView({
...
+    setAttachedComments([]);
+    setArtifact(null);
+    setMessagesConversationId(null);
+    savedArtifactRef.current = null;
+    pendingWritesRef.current.clear();
+    if (messagesConversationIdRef.current !== activeConversationId) {
+      messagesConversationIdRef.current = null;
+    }
     (async () => {
       const [list, comments] = await Promise.all([
```
mrcfps: `currentConversationBusy` stays true until `messagesConversationId === activeConversationId`, but this load path never handles a rejected `listMessages` / `fetchPreviewComments` call. After the new `setMessages([])` / `setMessagesConversationId(null)` reset above, any transient failure here leaves `messagesConversationId` at `null`, so the new guards in `handleSend`, `handleNewConversation`, export, and todo-continuation stay blocked for that conversation indefinitely. Can we wrap this `Promise.all` in `try/catch`, surface the load error, and clear the loading gate (for example with an explicit failed-load state) so the user can retry instead of getting stuck?

mrcfps (commented):
@nettee Thanks for the fix here — I verified the conversation-scoped run gating and found two follow-up issues worth addressing before we call this path done.


mrcfps (commented):
@nettee Thanks for iterating on this — the conversation-scoped busy-state fix and the notification/load follow-ups look directionally right. I found one remaining merge-safe correctness issue in the new load-error path worth tightening up.


apps/web/src/components/ProjectView.tsx
```diff
@@ -413,24 +425,53 @@ export function ProjectView({
...
+        const message = err instanceof Error ? err.message : 'Could not load messages for this conversation.';
+        setMessages([]);
+        setPreviewComments([]);
+        setAttachedComments([]);
+        setArtifact(null);
+        setError(message);
+        setStreaming(false);
+        savedArtifactRef.current = null;
+        pendingWritesRef.current.clear();
+        messagesConversationIdRef.current = activeConversationId;
```
mrcfps: In `ProjectView`'s conversation-load `catch` path, this assignment marks the selected conversation as fully loaded even though `listMessages()` just failed and `messages` was cleared a few lines above. That matters because `handleSend()` now uses `messagesConversationIdRef.current !== activeConversationId` as its history-readiness gate, so after a transient load failure on an existing conversation the user can immediately send a new turn and we will build `nextHistory` from an empty array instead of the persisted thread. Please keep the conversation in a retry/blocked state until the history actually loads successfully — for example, leave `messagesConversationId` / `messagesConversationIdRef` unset in the error branch and surface a retry action rather than treating the conversation as ready.

apps/web/src/components/ChatComposer.tsx
```diff
@@ -1031,7 +1033,7 @@ export const ChatComposer = forwardRef<ChatComposerHandle, Props>(
                 className="composer-send"
                 data-testid="chat-send"
                 onClick={() => void submit()}
-                disabled={!draft.trim() && commentAttachments.length === 0}
+                disabled={sendDisabled || (!draft.trim() && commentAttachments.length === 0)}
```
mrcfps, line 1037: The new `sendDisabled` prop only disables the visible Send button, but `submit()` still ignores that flag and the textarea's Cmd/Ctrl+Enter handler still calls `submit()` unconditionally (`ChatComposer.submit()` at lines 673-705 and the key handler at lines 835-837). That means the changed load-error/load-pending state here is bypassable from the keyboard: while a conversation is still loading or has just failed to load, a typed prompt can still be submitted through the shortcut path, `ProjectView.handleSend()` returns early because the conversation is not ready, and then `ChatComposer.reset()` clears the user's draft anyway. Please make `submit()` return early when `sendDisabled` is true (or gate the keyboard shortcut before calling it) so the disabled-send state consistently preserves the draft instead of silently discarding it.

mrcfps (commented):
@nettee Thanks for pushing this through — I reviewed the latest head and found one remaining merge-safe correctness gap in the new send gating path.


Siri-Ray (commented):
@nettee I found one remaining merge-safe gap in the duplicate empty-conversation guard. The conversation-scoped send/run isolation looks directionally sound otherwise.


apps/web/src/components/ProjectView.tsx
```diff
@@ -1745,6 +1814,8 @@ export function ProjectView({
       // Eagerly clear messages and update ref so rapid clicks don't create
       // duplicate empty conversations before the effect resolves.
       setMessages([]);
+      setStreaming(false);
+      setMessagesConversationId(null);
```
Siri-Ray, line 1824: The duplicate-empty-conversation guard still misses the period while `createConversation(project.id)` is in flight. The new eager cleanup happens only after the await returns, so a double-click on the header/history New Conversation button before the first request resolves enters this handler twice with the same `activeConversationId`, same `messagesConversationIdRef.current`, and same non-empty/active previous conversation state. That creates two fresh empty conversations even though the added comment says rapid clicks should be blocked before the load effect resolves. Please add an explicit in-flight guard that is set before awaiting `createConversation` (for example a `creatingConversationRef`/state that is checked at the top and cleared in `finally`, ideally also disabling the New Conversation buttons while true), or otherwise make the duplicate-prevention state change synchronous before the request starts.

nettee: Fixed in 1be97eaa. Added a synchronous in-flight creation guard before awaiting createConversation, clear it in finally, and disable both New Conversation buttons while creation is pending. Added regression coverage for double-clicks while the request is unresolved.

apps/web/src/components/ChatPane.tsx
```diff
@@ -732,6 +738,7 @@ export function ChatPane({
             projectFiles={projectFiles}
             skills={skills}
             streaming={streaming || hasActiveRunMessage}
+            sendDisabled={sendDisabled}
```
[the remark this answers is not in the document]
nettee: Fixed in 02eb1580. ChatPane now passes only the attached listener streaming state to ChatComposer, so Stop stays hidden until reattach installs cancellation. Assistant messages still render active runStatus as streaming, and ChatPane.streaming.test.tsx covers the split behavior.

mrcfps (approved):
@nettee I rechecked the latest head and verified the conversation-scoped send gating, retry path, stop-button gating, delete refresh, and regression coverage around the daemon conversation filter. Thanks for working through the follow-up fixes here 🙌

PR #1278, opened 2026-05-11
feat(daemon): add Claude Code finalize provider

ogeid-genexr:
## Summary

- Adds a second finalize endpoint `POST /api/projects/:id/finalize/claude-code` alongside the existing Anthropic BYOK route, so users with a Claude Code CLI login (Max plan, OAuth) can synthesize `DESIGN.md` without paying per-token API costs. Auth is delegated to the local CLI — the request body has no `apiKey` field.
- Refactors `apps/daemon/src/finalize-design.ts` to make synthesis pluggable: shared orchestration owns the lockfile, prompt assembly, abort/timeout composition, and atomic write; provider adapters only turn prompts into Markdown. The Anthropic path is preserved as one adapter; the new `apps/daemon/src/finalize-claude-code.ts` is the second.
- Web hook (`useFinalizeProject`) and `ProjectView` gain a provider toggle; the new path returns `model`, `inputTokens`, and `outputTokens` as nullable because the CLI's `result` event doesn't always surface them.

## Why

The existing finalize flow only supports BYOK via the Anthropic API, which double-charges Max-plan subscribers who already pay a flat fee for Claude usage through the CLI. Delegating to `claude` on PATH lets those users finalize designs without a second credential. Keeping it as a parallel route (not a flag on the Anthropic route) avoids muddying the contract: the two paths have different auth surfaces, different error modes (e.g. CLI-not-installed), and different usage telemetry guarantees.

## Notes for reviewers

- `FinalizeClaudeCodeNotInstalledError` is mapped to `503 UPSTREAM_UNAVAILABLE` because the contracts `ApiErrorCode` union has no `NOT_INSTALLED` code. Happy to add one if you'd prefer a more precise signal — left it as-is to avoid an unrelated contract change.
- Contract additions in `packages/contracts/src/api/finalize.ts` (`FinalizeClaudeCodeRequest`, `FinalizeClaudeCodeResponse`) intentionally mirror the Anthropic shapes minus `apiKey` and with nullable `model` / token fields.
- No changes to `apps/daemon/src/agents.ts` — this is a finalize provider, not a new coding-agent CLI adapter.

## Test plan

- [x] `pnpm guard`
- [x] `pnpm typecheck`
- [x] `pnpm --filter @open-design/daemon test` (1621 passing, includes new `finalize-claude-code.test.ts` and `finalize-claude-code-route.test.ts`)
- [x] `pnpm --filter @open-design/web test` (754 passing, includes updated `useFinalizeProject.test.tsx`)
- [x] `pnpm --filter @open-design/web build`
- [ ] Manual: end-to-end finalize via Claude Code CLI on a real project

### DIFF
--- a/apps/daemon/src/finalize-claude-code.ts
+++ b/apps/daemon/src/finalize-claude-code.ts
@@ -1,0 +1,498 @@
+// Provider-sibling of finalize-design.ts that synthesizes DESIGN.md via
+// the locally installed Claude Code CLI rather than calling
+// api.anthropic.com directly. Designed for Max plan subscribers, whose
+// CLI usage is subsidized by their subscription — paying per-token API
+// rates for finalize on top of an active Max plan is the gap this
+// module closes (GitHub issue nexu-io/open-design#963).
+//
+// The shared orchestration (lockfile, transcript export + truncation,
+// prompt build, atomic DESIGN.md write) lives in finalize-design.ts.
+// This module supplies a `FinalizeSynthesizer` that:
+//   1. Preflights `claude --version` so a missing CLI surfaces as a
+//      503-mappable error before the lockfile is even taken.
+//   2. Spawns `claude --print --output-format stream-json` with the
+//      project directory as cwd, writes the user prompt to stdin,
+//      and accumulates the stream-json `result` event for the body
+//      and (when surfaced) usage counters.
+//   3. Maps process-level failures (auth, abort, non-zero exit) onto
+//      `FinalizeUpstreamError` / `AbortError` so the route handler
+//      can reuse the Anthropic-route's status-aware mapping.
+
+import { spawn, type ChildProcess, type SpawnOptions } from 'node:child_process';
+import type Database from 'better-sqlite3';
+import type {
+  FinalizeClaudeCodeResponse,
+  FinalizeArtifactRef,
+} from '@open-design/contracts/api/finalize';
+import { getProject } from './db.js';
+import { resolveProjectDir } from './projects.js';
+import {
+  FinalizeUpstreamError,
+  runFinalizeWithSynthesizer,
+  type FinalizeSynthesisResult,
+  type FinalizeSynthesizer,
+  type RunFinalizeOptions,
+} from './finalize-design.js';
+
+export type { FinalizeClaudeCodeResponse };
+
+type Db = Database.Database;
+
+/**
+ * Resolve the CLI binary path. Production defaults to plain `claude`
+ * (PATH lookup); `OD_CLAUDE_CODE_CLI_BIN` overrides for users with
+ * non-PATH installs and for daemon integration tests that point the
+ * route at a controlled script.
+ */
+function defaultCliBin(): string {
+  return process.env.OD_CLAUDE_CODE_CLI_BIN || 'claude';
+}
+const VERSION_PROBE_TIMEOUT_MS = 5_000;
+/**
+ * Default upstream-call ceiling for the CLI route. Claude Code
+ * synthesis is meaningfully slower than a direct Anthropic Messages
+ * API call: the CLI streams its response, may retry internally, and
+ * has subprocess overhead the API path doesn't pay. A 2-minute
+ * ceiling (the Anthropic default) trips while the CLI is still
+ * legitimately working. 10 minutes accommodates a multi-turn
+ * synthesis without giving runaway processes an unbounded budget.
+ */
+const DEFAULT_CLAUDE_CODE_TIMEOUT_MS = 600_000;
+
+/**
+ * The local Claude Code CLI is not installed (or not on PATH the
+ * daemon can see). Distinct from `FinalizeUpstreamError` so the route
+ * handler can return a tailored 503 message instructing the user to
+ * install / re-login rather than the generic upstream-failed code.
+ */
+export class FinalizeClaudeCodeNotInstalledError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = 'FinalizeClaudeCodeNotInstalledError';
+  }
+}
+
+/**
+ * Test-injection seam. Production callers leave both fields
+ * unspecified and the synthesizer uses real `child_process.spawn`
+ * with the resolved CLI binary path. Tests substitute both.
+ */
+export interface ClaudeCodeCliTransport {
+  /** Override for the `claude` executable path. Default: `claude` (PATH lookup). */
+  cliPath?: string;
+  /** Override spawn implementation. Default: `child_process.spawn`. */
+  spawnImpl?: typeof spawn;
+}
+
+export interface FinalizeClaudeCodeOptions extends RunFinalizeOptions {
+  cli?: ClaudeCodeCliTransport;
+}
+
+/**
+ * One-shot preflight: run `claude --version` and resolve when the
+ * CLI prints a non-empty stdout AND exits zero. Maps ENOENT to
+ * `FinalizeClaudeCodeNotInstalledError`; any other failure (non-zero
+ * exit, empty stdout) to a generic 502 `FinalizeUpstreamError` so
+ * the route still maps to UPSTREAM_UNAVAILABLE.
+ *
+ * Bounded by a 5s internal timeout; the wider per-finalize timeout
+ * does not gate the preflight because it runs before the lockfile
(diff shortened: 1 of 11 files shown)


apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,498 @@
...
+
+  const runOptions: RunFinalizeOptions = {
+    // Apply the CLI-specific default ceiling unless the caller has
+    // explicitly opted into a shorter or longer budget; tests pass
+    // smaller values to exercise the abort path without burning
+    // wall-clock time.
+    timeoutMs: options.timeoutMs ?? DEFAULT_CLAUDE_CODE_TIMEOUT_MS,
+  };
+  if (options.model !== undefined) runOptions.model = options.model;
+  if (options.maxTokens !== undefined) runOptions.maxTokens = options.maxTokens;
```
mrcfps, line 475: `maxTokens` is exposed all the way from the contract and route, but it never affects the Claude Code invocation here. In this file we copy `options.maxTokens` into `runOptions`, yet `runFinalizeWithSynthesizer` never forwards that field to `callClaudeCodeCLI`, and the CLI adapter itself has no `maxTokens` input or argv handling. That means a caller can send `{ maxTokens: ... }` and get a 200 even though the limit was ignored. Because the new request shape advertises `maxTokens` support, this will produce misleading behavior for anyone trying to cap long CLI runs. Please either thread the limit into the actual `claude` command if the CLI supports it, or remove/reject `maxTokens` for this provider until it can be enforced.

mrcfps (commented):
@ogeid-genexr Thanks for shipping the Claude Code finalize path — I walked through the new daemon/web flow and found two follow-up issues worth tightening so the new contract matches runtime behavior.


Location: apps/web/src/hooks/useFinalizeProject.ts RIGHT line 246

The timeout message is now wrong for the new provider. `FETCH_TIMEOUT_BY_PROVIDER` above gives `claude-code` a 630_000 ms client timeout, but this branch still hard-codes `130 s`. When a Claude Code finalize times out after ~10.5 minutes, the UI will tell the user it failed after 130 seconds, which makes the failure hard to diagnose and encourages premature retries while the daemon may still be running. Please derive this copy from the selected provider timeout (or add a provider-specific TIMEOUT string) so the user-facing error stays aligned with the actual request budget.

PerishCode:
Closing in favour of #1279, which has the identical diff (11 files, +1628/-106) and already carries the active review feedback. Consolidating the review conversation there to avoid splitting follow-ups across two threads.

apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,498 @@
...
+
+  // Pre-abort short-circuit: if the caller already aborted before we
+  // spawned, raise AbortError without touching the CLI.
+  if (input.signal.aborted) {
+    const err = new Error('aborted');
+    err.name = 'AbortError';
+    throw err;
+  }
+
+  const spawnOptions: SpawnOptions = {
```
lefarcen, line 257: 🔒 P1 — The child process inherits daemon's full `process.env` without sanitization. That exposes secrets (e.g., `ANTHROPIC_API_KEY`, other daemon env vars) to a subprocess launched against user-controlled project content, and bypasses the existing `spawnEnvForAgent` invariant (which strips `ANTHROPIC_API_KEY` so Claude Code uses Max-plan billing, not fallback API billing).

Concrete fix: build the spawn env through `spawnEnvForAgent('claude', process.env, configuredAgentEnv)` or equivalent sanitized env, and apply it to both `probeClaudeCodeCli` and `callClaudeCodeCLI`. Add a test verifying `ANTHROPIC_API_KEY` is absent in the child.

apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,498 @@
...
+  const args: string[] = [
+    '--print',
+    '--input-format',
+    'text',
+    '--output-format',
+    'stream-json',
+    // stream-json requires --verbose to emit per-event chunks rather
+    // than collapsing the run into a single terminal event.
+    '--verbose',
+    '--append-system-prompt',
```
lefarcen, line 237: ⚠️ P1 — Comment says tools are disabled, but `--permission-mode default` does not disable Claude Code tools; it leaves normal tool behavior active. A prompt-injected artifact/transcript can ask the CLI to read files, edit files, or run shell commands, or the run can hang on permission prompts.

Concrete fix: invoke Claude Code with an explicit no-tools allowlist/disallowlist if supported, or run through a tool-free API/SDK path. Add a regression test that a malicious transcript cannot cause tool execution or permission prompting.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1958,12 +1958,27 @@ export function ProjectView({
...
-      model: config.model,
-      maxTokens: effectiveMaxTokens(config),
-    }).then((result) => {
+    // Route to the Claude Code CLI sibling endpoint (#963) when the
+    // user has the local Claude Code agent selected; this inherits
+    // their Max plan subscription so they do not pay per-token API
+    // costs on top. Every other configuration (BYOK API mode, or
+    // daemon mode with a non-claude agent) falls through to the
+    // Anthropic API route (#832), which still requires a BYOK key.
+    const useClaudeCode = config.mode === 'daemon' && config.agentId === 'claude';
```
lefarcen, line 1967: P2 — The finalize call sends `config.model` (BYOK/API model field), not the selected CLI model. Normal daemon chat uses `config.agentModels?.[config.agentId]?.model`, so finalize can ignore the user's CLI model selection and force an unrelated Anthropic API model string.

Concrete fix: for `provider: 'claude-code'`, use `config.agentModels?.claude?.model`, and omit the model when unset or sentinel `default`. Mirror the daemon agent guard so `callClaudeCodeCLI` does not pass `--model default`.

apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,498 @@
...
+ * text + usage. Errors map as:
+ *   - process spawn ENOENT → `FinalizeClaudeCodeNotInstalledError`
+ *   - signal abort → AbortError (orchestrator → route → 503)
+ *   - auth-signal-bearing stderr / error-subtype result → 401 via
+ *     `FinalizeUpstreamError(401)`
+ *   - any other non-zero exit / unparseable stream → 502 via
+ *     `FinalizeUpstreamError(502)` with the trailing stderr as
+ *     `rawText`
+ */
+export async function callClaudeCodeCLI(input: {
```
lefarcen, line 217: P2 — `maxTokens` is accepted in the contract and route, web sends it, but `callClaudeCodeCLI` has no `maxTokens` input and never maps it to any Claude CLI argument. Users can set a finalize token cap that is silently ignored, changing cost/runtime/output-size behavior versus Anthropic finalize.

Concrete fix: either wire `maxTokens` to a supported Claude Code limit flag and test the argv, or remove/ignore it explicitly from the Claude Code request surface and stop sending it from the web.

(This duplicates @mrcfps's finding above — same issue.)

lefarcen (changes requested):
Hey @ogeid-genexr! This PR delivers a valuable Max-plan route for finalize, and the plugin refactor is clean. Found two P1 security gaps + two config-mapping issues below.

**Security (P1):**

1. **Environment leakage** (finalize-claude-code.ts:257): The Claude Code child inherits the daemon's full `process.env` without sanitization. That exposes secrets and bypasses the existing `spawnEnvForAgent` invariant (which strips `ANTHROPIC_API_KEY` so the CLI uses Max-plan billing, not fallback API billing). See inline.

2. **Tool execution risk** (finalize-claude-code.ts:237): Comment says tools are disabled, but `--permission-mode default` leaves Claude Code tools active. A malicious artifact/transcript can inject tool calls (read/write/exec). See inline.

**Config mapping (P2):**

3. **Model selection mismatch** (ProjectView.tsx:1967): The finalize call sends `config.model` (BYOK API model), not the CLI model the user selected. See inline.

4. **Ignored maxTokens** (finalize-claude-code.ts:217): The route accepts `maxTokens`, web sends it, but the CLI call never maps it to a CLI arg. Silently ignoring the cap changes cost/runtime behavior. See inline.

@mrcfps already noted #4 and a related timeout message issue. Once these are addressed + tests added, this should be ready. Thanks for shipping this!

— open-design team

PR #1279, opened 2026-05-11
feat(daemon): add Claude Code finalize provider

### LINKED ISSUE #963: Add /finalize/claude-code provider so Max plan users avoid per-token API costs
## Summary

The `/api/projects/:id/finalize/anthropic` endpoint shipped in PR #832 is BYOK API only — synthesis is billed per-token at api.anthropic.com prices regardless of whether the calling user has an Anthropic Max plan subscription. Max plan users currently pay double for finalize: once via their subscription (which doesn't apply), and again via API tokens.

Add a sibling provider — `/api/projects/:id/finalize/claude-code` — that shells out to the Claude Code CLI for synthesis. CLI calls inherit Max plan's subsidized billing, so finalize would be effectively free for Max plan subscribers.

## Why this fits PR #832's design

The route shape is intentionally provider-namespaced. From `packages/contracts/src/api/finalize.ts:1-4`:

> Shared DTOs for the `/api/projects/:id/finalize/<provider>` family of synthesis endpoints. The first endpoint is `/finalize/anthropic` (introduced in PR #832); future provider-namespaced siblings (`/finalize/openai` etc.) can reuse the request/response shape.

So PR #832 already designed the contract to support multiple synthesis providers. `/finalize/claude-code` is the natural Max-plan-friendly sibling.

## Proposal

A new daemon route `/api/projects/:id/finalize/claude-code` that:

1. Accepts a request shape similar to `FinalizeAnthropicRequest` but **without `apiKey`** — Claude Code CLI handles auth via the user's existing CLI login (`claude auth login`). The contract could be a separate `FinalizeClaudeCodeRequest` interface in `packages/contracts/src/api/finalize.ts`, sharing `model` and `maxTokens` fields.
2. Reuses the same synthesis prompt and Provenance section format (`apps/daemon/src/finalize-design.ts:533-559`) — only the upstream call differs.
3. Spawns a Claude Code subprocess with the prompt as input, captures stdout, writes the result to `DESIGN.md` atomically (same write path as `/anthropic`).
4. Returns a `FinalizeClaudeCodeResponse` mirroring `FinalizeAnthropicResponse` (`designMdPath`, `bytesWritten`, `model`, `transcriptMessageCount`, `designSystemId`, `artifact`) — but **token counts may not be available** depending on what the CLI surfaces. Could be omitted or null in the response.

## Open implementation questions

- **CLI invocation pattern.** `claude --print "..."` or `claude` with stdin? Headless (non-interactive) mode? Does the CLI have a stable JSON output mode, or do we parse Markdown stdout? Reference: <https://docs.claude.com/en/docs/claude-code/sdk> if available.
- **Auth detection.** How do we detect whether the user has Claude Code CLI installed and authenticated? Probe via `claude --version` + a no-op auth check? Surface a clear error if not configured.
- **Timeout / streaming.** Same 120s blocking posture as `/anthropic`? CLI calls may stream — daemon would need to accumulate stdout and apply `AbortController` semantics on cancel.
- **PR 3 / #451 implications.** The web client (currently designed in PR 3 spec) would need to know which provider to call. Either auto-detect (use `/claude-code` if CLI is available, fall back to `/anthropic`) or expose a settings toggle. Probably out of scope for this issue; PR 3 ships with `/anthropic` and the toggle lands as a follow-up.

## Out of scope

- **Streaming finalize.** Per @lefarcen's lock on #450 (blocking for v1), streaming stays a separate follow-up.
- **OpenAI / other providers.** The route shape supports `/finalize/openai` etc., but those are independent issues.
- **Default-provider selection in PR 3 web client.** PR 3 / #451 ships with `/anthropic`; provider toggle is a follow-up to PR 3.

## Acceptance criteria

- [ ] New `FinalizeClaudeCodeRequest` and `FinalizeClaudeCodeResponse` interfaces in `packages/contracts/src/api/finalize.ts` (no apiKey field; same shape otherwise).
- [ ] New route `POST /api/projects/:id/finalize/claude-code` registered in `apps/daemon/src/server.ts`.
- [ ] CLI invocation helper (analogous to `callAnthropicWithRetry` in `apps/daemon/src/finalize-design.ts`) that spawns Claude Code, captures stdout, handles abort/timeout.
- [ ] Atomic write to `DESIGN.md` reuses the existing primitive (no duplicate write paths).
- [ ] Tests cover: CLI not installed → 503 with clear message; CLI not authenticated → 401-equivalent; success path → DESIGN.md lands; cancel mid-call → AbortController fires.
- [ ] Error mapping uses the same `ApiErrorCode` union as `/anthropic` for consistency.

## Related

- PR #832 — `/finalize/anthropic` provider (parent feature)
- Issue #450 — finalize design package (parent epic)
- Issue #451 — Continue in CLI (web client; will eventually want provider selection)

Surfaced from real-world testing of #450's verification: an Anthropic API account hit a usage cap during finalize testing, while the user holds a Max plan that should have made finalize subsidized. Two-billing-paths-for-the-same-feature is the gap.

ogeid-genexr:
## Summary
- Adds a second finalize path at POST /api/projects/:id/finalize/claude-code alongside the existing Anthropic BYOK route, so users with a Claude Code CLI login (Max plan, OAuth) can synthesize DESIGN.md without paying per-token API costs. Auth is delegated to the local CLI — no apiKey field on the request.

Refactors finalize-design.ts to make synthesis pluggable: shared orchestration owns lockfile, prompt assembly, abort/timeout composition, and atomic write; provider adapters only turn prompts into Markdown. The Anthropic path is preserved as one adapter; the new finalize-claude-code.ts is the second.

Error mapping mirrors the Anthropic route (409 lock, 401/429/502 upstream, AbortError → 503), plus a dedicated
FinalizeClaudeCodeNotInstalledError surfaced as 503 UPSTREAM_UNAVAILABLE since the contracts ApiErrorCode union has no NOT_INSTALLED code.

Web hook and ProjectView gain a provider toggle; tests cover both the unit synthesis path and the route's error mapping.

## Validation
-

Fixes #963

### DIFF
--- a/apps/daemon/src/finalize-claude-code.ts
+++ b/apps/daemon/src/finalize-claude-code.ts
@@ -60,6 +60,38 @@ const VERSION_PROBE_TIMEOUT_MS = 5_000;
 const DEFAULT_CLAUDE_CODE_TIMEOUT_MS = 600_000;
 
 /**
+ * Full built-in tool surface of the Claude Code CLI. Passed to
+ * `--disallowedTools` so the synthesis run cannot launch Bash, read
+ * arbitrary files from the project cwd, search the web, etc. The
+ * synthesis prompt is self-contained — all transcript, design
+ * system, and artifact context is in stdin — so this list does not
+ * elide any capability the route genuinely needs.
+ *
+ * Keep in sync with upstream when new built-in tools ship; an unknown
+ * tool name in `--disallowedTools` is harmless, a missing one is a
+ * permission gap.
+ */
+export const CLAUDE_CODE_BUILTIN_TOOLS = [
+  'Bash',
+  'BashOutput',
+  'Edit',
+  'ExitPlanMode',
+  'Glob',
+  'Grep',
+  'KillShell',
+  'LS',
+  'MultiEdit',
+  'NotebookEdit',
+  'Read',
+  'SlashCommand',
+  'Task',
+  'TodoWrite',
+  'WebFetch',
+  'WebSearch',
+  'Write',
+] as const;
+
+/**
  * The local Claude Code CLI is not installed (or not on PATH the
  * daemon can see). Distinct from `FinalizeUpstreamError` so the route
  * handler can return a tailored 503 message instructing the user to
@@ -220,6 +252,7 @@ export async function callClaudeCodeCLI(input: {
   signal: AbortSignal;
   cwd: string;
   model?: string;
+  maxTokens?: number;
   transport?: ClaudeCodeCliTransport;
 }): Promise<FinalizeSynthesisResult> {
   const cliPath = input.transport?.cliPath ?? defaultCliBin();
@@ -236,9 +269,19 @@ export async function callClaudeCodeCLI(input: {
     '--verbose',
     '--append-system-prompt',
     input.systemPrompt,
-    // Synthesis never needs to read project files or run tools — the
-    // entire context is already in the user prompt. Disable tools so
-    // a Max plan user does not see surprising tool-permission prompts.
+    // Synthesis never needs project file access or tool invocations —
+    // the full transcript, design system, and current artifact are
+    // already in the user prompt. The CLI's default config allows the
+    // agent's built-in tool surface (Bash/Edit/Write/Read/Glob/Grep/
+    // WebFetch/WebSearch/Task/TodoWrite/NotebookEdit/MultiEdit/LS/
+    // SlashCommand/BashOutput/KillShell/ExitPlanMode), so we
+    // explicitly deny every one of them. `--permission-mode default`
+    // alone does NOT disable tools; it only controls how permission
+    // prompts are surfaced. The combination here keeps the CLI in a
+    // pure prompt-completion shape and prevents any side-effecting
+    // tool invocation against the user's project directory.
+    '--disallowedTools',
+    CLAUDE_CODE_BUILTIN_TOOLS.join(' '),
     '--permission-mode',
     'default',
   ];
@@ -254,10 +297,21 @@ export async function callClaudeCodeCLI(input: {
     throw err;
   }
 
+  // Claude Code reads its per-response output ceiling from the
+  // `CLAUDE_CODE_MAX_OUTPUT_TOKENS` env var (it has no equivalent
+  // CLI flag). Honoring `maxTokens` here keeps the route's request
+  // contract truthful: a caller that asks for a tighter ceiling
+  // actually gets one, matching the Anthropic provider's behavior.
+  const childEnv = { ...process.env };
+  if (typeof input.maxTokens === 'number' && input.maxTokens > 0) {
+    childEnv.CLAUDE_CODE_MAX_OUTPUT_TOKENS = String(Math.trunc(input.maxTokens));
+  }
+
   const spawnOptions: SpawnOptions = {
     cwd: input.cwd,
     stdio: ['pipe', 'pipe', 'pipe'],
     signal: input.signal,
+    env: childEnv,
   };
 
   let child: ChildProcess;
@@ -460,6 +514,7 @@ export async function finalizeDesignPackageWithClaudeCode(
       cwd,
     };
     if (options.model) callInput.model = options.model;
+    if (typeof options.maxTokens === 'number') callInput.maxTokens = options.maxTokens;
     if (options.cli) callInput.transport = options.cli;
     return callClaudeCodeCLI(callInput);
   };
--- a/apps/daemon/src/finalize-claude-code.ts
+++ b/apps/daemon/src/finalize-claude-code.ts
@@ -60,36 +60,16 @@ const VERSION_PROBE_TIMEOUT_MS = 5_000;
 const DEFAULT_CLAUDE_CODE_TIMEOUT_MS = 600_000;
 
 /**
- * Full built-in tool surface of the Claude Code CLI. Passed to
- * `--disallowedTools` so the synthesis run cannot launch Bash, read
- * arbitrary files from the project cwd, search the web, etc. The
- * synthesis prompt is self-contained — all transcript, design
- * system, and artifact context is in stdin — so this list does not
- * elide any capability the route genuinely needs.
+ * Empty allowlist literal passed to `--allowedTools` so the synthesis
+ * run has zero tool surface regardless of which built-in tools the
+ * installed CLI version ships or which MCP servers / configured tools
+ * the user has wired up. The synthesis prompt is self-contained — all
+ * transcript, design system, and artifact context is in stdin — so the
+ * provider never needs a tool.
  *
- * Keep in sync with upstream when new built-in tools ship; an unknown
- * tool name in `--disallowedTools` is harmless, a missing one is a
- * permission gap.
+ * Exported so tests can assert the argv contract.
  */
-export const CLAUDE_CODE_BUILTIN_TOOLS = [
-  'Bash',
-  'BashOutput',
-  'Edit',
-  'ExitPlanMode',
-  'Glob',
-  'Grep',
-  'KillShell',
-  'LS',
-  'MultiEdit',
-  'NotebookEdit',
-  'Read',
-  'SlashCommand',
-  'Task',
-  'TodoWrite',
-  'WebFetch',
-  'WebSearch',
-  'Write',
-] as const;
+export const CLAUDE_CODE_EMPTY_TOOL_ALLOWLIST = '' as const;
 
 /**
  * The local Claude Code CLI is not installed (or not on PATH the
@@ -271,17 +251,17 @@ export async function callClaudeCodeCLI(input: {
     input.systemPrompt,
     // Synthesis never needs project file access or tool invocations —
     // the full transcript, design system, and current artifact are
-    // already in the user prompt. The CLI's default config allows the
-    // agent's built-in tool surface (Bash/Edit/Write/Read/Glob/Grep/
-    // WebFetch/WebSearch/Task/TodoWrite/NotebookEdit/MultiEdit/LS/
-    // SlashCommand/BashOutput/KillShell/ExitPlanMode), so we
-    // explicitly deny every one of them. `--permission-mode default`
-    // alone does NOT disable tools; it only controls how permission
-    // prompts are surfaced. The combination here keeps the CLI in a
-    // pure prompt-completion shape and prevents any side-effecting
-    // tool invocation against the user's project directory.
-    '--disallowedTools',
-    CLAUDE_CODE_BUILTIN_TOOLS.join(' '),
+    // already in the user prompt. Pass an empty allowlist so the CLI
+    // runs with zero tool surface regardless of which built-in tools
+    // ship in the installed version or which MCP / configured tools
+    // the user has wired up. This is the inverse of an enumerated
+    // denylist: a denylist leaks any tool we forgot to name, an empty
+    // allowlist leaks nothing. `--permission-mode default` alone does
+    // NOT disable tools; it only controls how permission prompts are
+    // surfaced. The combination keeps the CLI in a pure
+    // prompt-completion shape.
+    '--allowedTools',
+    CLAUDE_CODE_EMPTY_TOOL_ALLOWLIST,
     '--permission-mode',
     'default',
   ];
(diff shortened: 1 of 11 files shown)


apps/daemon/src/finalize-design.ts
```diff
@@ -260,10 +329,40 @@ export async function finalizeDesignPackage(
...
+      fs.unlinkSync(lockPath);
+    } catch {
+      // lock may already be gone if disk vanished; not fatal
+    }
+  };
+
+  let abortReleaseListener: (() => void) | null = null;
+  if (options.signal) {
+    abortReleaseListener = releaseLock;
+    options.signal.addEventListener('abort', abortReleaseListener, { once: true });
```
mrcfps, line 359: Releasing `.finalize.lock` from the request abort signal here opens a write race with the retry path. After `controller.abort()`, this listener drops the lock immediately, but `runFinalizeWithSynthesizer()` still goes straight from `await synthesize(...)` into the phase-9 rename on lines 421-433 without checking whether the composed signal has been aborted. If the provider settles just after the abort (or briefly ignores cancellation), the canceled run can still rename its `DESIGN.md` after the retry has already started, so the older result wins last-write-wins. Please either keep the lock until the aborted invocation is guaranteed not to write, or add an explicit `if (composedSignal.aborted) throw AbortError` guard before the atomic write so canceled runs cannot overwrite newer output.

mrcfps (changes requested):
@ogeid-genexr Thanks for the clear provider split and the route/test coverage here — the new Claude Code path is easy to follow. I found one correctness issue in the new eager-unlock flow that can still overwrite a retry, plus one API-contract mismatch around `maxTokens` to tighten up.


Location: apps/daemon/src/finalize-claude-code.ts RIGHT line 525

`maxTokens` is validated in the `/finalize/claude-code` route, carried through `FinalizeClaudeCodeOptions`, and copied into `runOptions` here, but this provider never actually uses it when building the Claude Code CLI invocation. `callClaudeCodeCLI()` only appends `--model`, so callers can send `maxTokens` and get a 200 even though the subprocess ignores that limit completely. That leaves the new endpoint out of sync with its request contract and makes latency/output-length control unreliable for clients. Please either thread the corresponding Claude Code token-limit flag into `callClaudeCodeCLI()`, or remove `maxTokens` from the CLI route/contract until the CLI path can honor it.

Siri-Ray (commented):
@ogeid-genexr I completed a changed-range implementation pass and found one additional blocker in the Claude Code invocation around tool isolation. The existing unresolved threads on eager lock release and `maxTokens` still stand.


apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,498 @@
...
+    // stream-json requires --verbose to emit per-event chunks rather
+    // than collapsing the run into a single terminal event.
+    '--verbose',
+    '--append-system-prompt',
+    input.systemPrompt,
+    // Synthesis never needs to read project files or run tools — the
+    // entire context is already in the user prompt. Disable tools so
+    // a Max plan user does not see surprising tool-permission prompts.
+    '--permission-mode',
+    'default',
```
Siri-Ray, line 270: The comment says this disables tools, but the argv only sets `--permission-mode default`; that leaves Claude Code running with its normal tool surface in the user's project directory. This finalize path already provides the transcript, design system, and current artifact in stdin, and the daemon's atomic write is supposed to be the only filesystem mutation, so allowing the CLI agent to read files, invoke configured tools, or trigger project-level behavior from `cwd` is a production safety risk for imported folders. Please pass an explicit no-tools configuration, for example `--tools ""` or an equivalent `--disallowedTools` policy, and extend the argv test to assert the provider cannot launch with tool access.

lefarcen:
Closing as duplicate of upcoming work by @bankielewicz on #963.

The original /claim was incorrectly released by a charter bug — bot 4-5 days
before the claimer's stated ETA (5/14–5/16). Restoring claim ownership to
@bankielewicz; his ETA still stands. Thanks for the contribution @ogeid-genexr —
if any code here is useful as a reference, @bankielewicz is welcome to draw on
it, but no obligation.

lefarcen:
Reopening — I closed this too quickly without reviewing the actual code. Will do a proper review before deciding.

apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,553 @@
...
+    // already in the user prompt. The CLI's default config allows the
+    // agent's built-in tool surface (Bash/Edit/Write/Read/Glob/Grep/
+    // WebFetch/WebSearch/Task/TodoWrite/NotebookEdit/MultiEdit/LS/
+    // SlashCommand/BashOutput/KillShell/ExitPlanMode), so we
+    // explicitly deny every one of them. `--permission-mode default`
+    // alone does NOT disable tools; it only controls how permission
+    // prompts are surfaced. The combination here keeps the CLI in a
+    // pure prompt-completion shape and prevents any side-effecting
+    // tool invocation against the user's project directory.
+    '--disallowedTools',
```
lefarcen: **P1** This still leaves a tool surface available because it relies on a hand-maintained `--disallowedTools` list. The installed/docs tool set already includes tools not listed here (e.g. `NotebookRead`), and future or configured tool surfaces can slip through. Claude Code documents `--tools ""` as the way to restrict built-in tools to none. For this finalize path, the safer contract is to pass an empty tool set rather than enumerate tools to deny.

lefarcen (changes requested):
Thanks for the updates @ogeid-genexr — the lock race and maxTokens threading are now fixed, which clears 2 of the 3 prior blockers.

One P1 remains: tool isolation is still incomplete because `--disallowedTools` relies on a hand-maintained list that can miss installed/configured tools. See inline below.

apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,533 @@
...
+    // already in the user prompt. Pass an empty allowlist so the CLI
+    // runs with zero tool surface regardless of which built-in tools
+    // ship in the installed version or which MCP / configured tools
+    // the user has wired up. This is the inverse of an enumerated
+    // denylist: a denylist leaks any tool we forgot to name, an empty
+    // allowlist leaks nothing. `--permission-mode default` alone does
+    // NOT disable tools; it only controls how permission prompts are
+    // surfaced. The combination keeps the CLI in a pure
+    // prompt-completion shape.
+    '--allowedTools',
```
lefarcen: **P1** `--allowedTools ""` does not satisfy the isolation requirement. Claude Code exposes `--tools <tools...>` specifically for constraining the built-in tool surface, with `--tools ""` documented as disabling all tools. `--allowedTools` is a permission allowlist and does not replace the built-in tool set. Please change this to `--tools ""` instead.

lefarcen (changes requested):
Thanks for the quick iteration @ogeid-genexr — you're close, but the tool isolation fix still needs one adjustment.

The flag is `--tools ""`, not `--allowedTools ""`. Claude Code documents `--tools` as the way to constrain the built-in tool surface. See inline below.

apps/daemon/src/finalize-claude-code.ts
ogeid-genexr: Ah, my bad, will fix it.

apps/daemon/src/finalize-claude-code.ts
```diff
@@ -0,0 +1,536 @@
...
+    err.name = 'AbortError';
+    throw err;
+  }
+
+  // Claude Code reads its per-response output ceiling from the
+  // `CLAUDE_CODE_MAX_OUTPUT_TOKENS` env var (it has no equivalent
+  // CLI flag). Honoring `maxTokens` here keeps the route's request
+  // contract truthful: a caller that asks for a tighter ceiling
+  // actually gets one, matching the Anthropic provider's behavior.
+  const childEnv = { ...process.env };
```
lefarcen: **P1** — This spawn env still copies `process.env` wholesale, including `ANTHROPIC_API_KEY`. Claude Code prefers that env var over the user's local `claude /login` subscription credentials, which is exactly the billing boundary this provider is meant to preserve. Please strip `ANTHROPIC_API_KEY` here, or reuse the existing Claude Code spawn-env helper/behavior from the agent path so Max/Pro users do not get API-billed when finalizing.

lefarcen (changes requested):
Thanks for fixing the tool isolation flag @ogeid-genexr — that's resolved now.

However, a new P1 emerged: the Claude Code spawn env still inherits `ANTHROPIC_API_KEY` from the daemon process, which defeats the entire purpose of this PR (Max plan users avoiding per-token API billing). See inline below.

lefarcen (dismissed):
Thanks for the fix @ogeid-genexr — the `ANTHROPIC_API_KEY` env leak is now resolved, and the added test coverage confirms the billing boundary is preserved.

One minor nit on the timeout copy (P3, non-blocking): `messageForCode('TIMEOUT', 'claude-code')` still says "130s" but Claude Code requests now use 630s. If you want to fix that before merge, great; otherwise it's fine as-is.

lefarcen (commented):
Thanks for the update @ogeid-genexr — the timeout copy mismatch is now fixed. The PR looks clean from my side; deferring final approval to a maintainer.

lefarcen:
@mrcfps This PR's current head 7ad0b675 is waiting on your review.

Siri-Ray (approved):
@ogeid-genexr I reviewed the changed ranges for the Claude Code finalize provider, the shared finalize orchestration, route/contract wiring, and the web finalize hook. The current head addresses the prior lock-race, tool-isolation, env-leak, and timeout-copy concerns, and I did not find any additional actionable correctness or safety issues. I attempted the targeted daemon/web finalize tests, but this checkout is missing `node_modules`/`vitest`, so I could not execute them locally. Thanks for the thoughtful iteration on this feature.


lefarcen:
Hey @ogeid-genexr! 👋 The PR has accumulated merge conflicts against `main` since your last push — could you do a rebase onto the latest `main` to clear them? That's the last mechanical step before this can land.

While you do that, we're pulling in @qiongyu1999 on the product side to review the new UI provider toggle and the Claude Code route — standard step for features that touch the web layer.

Siri-Ray has approved the current head and the code is clean; just need the rebase + product sign-off and you'll be good. 🙌

ogeid-genexr:
@lefarcen given the ClaudeDevs post on X yesterday https://x.com/ClaudeDevs/status/2054610152817619388

is it really worth it at this point?

```
Starting June 15, paid Claude plans can claim a dedicated monthly credit for programmatic usage.

The credit covers usage of:
- Claude Agent SDK
- claude -p
- Claude Code GitHub Actions
- Third-party apps built on the Agent SDK
```

lefarcen:
That's a fair thing to raise, @ogeid-genexr — worth clarifying before you put more effort into the rebase.

The June 15 credit does change the per-token cost picture for `claude -p` users, and it's a reasonable signal to pause and check whether the BYOK-free path here is still the right investment. That said, there are a few angles that don't change with billing credits (seamless OAuth, no API key management for Max plan subscribers, cleaner UX for the zero-config audience) — so whether the net value still holds is genuinely a product call, not a code call.

I've already pulled @qiongyu1999 in on the product review side — let's get her take on the feature direction in light of this announcement before you spend time on the rebase. Hold off for now; I'll follow up once we have a clear signal on the direction.

ogeid-genexr:
@lefarcen I'd be inclined to NOT include these changes in the product given the June 15th change, it was just a nice to have since paying for a Claude Code subscription would already cover the usage, however it's a lot slower then the currently integration with Anthropic via the API path. Personally I'd keep using API tokens vs Claude Code.

lefarcen:
Thanks for the honest read, @ogeid-genexr — and the performance data point is useful; slower-than-API is exactly the kind of thing that matters when the billing gap is also closing.

Your take is noted and forwarded to @qiongyu1999, who's already in the loop for the product call. The final closure decision is hers (and the maintainer team's) to make, so I'll leave the PR open until we hear back. Shouldn't be long given she was already pinged.

lefarcen:
The six-round safety iteration — lock race, tool isolation, env leak, timeout copy — was genuinely thorough work, @ogeid-genexr. That iteration is part of what made the product tradeoff easy to read clearly. Really glad you raised the billing angle and shared the performance data; it made this a clean decision rather than an unresolved question. ❤️

If the CLI provider path gets revisited down the road, this PR's design will be a solid reference point.

PR #1284, opened 2026-05-11
Add reliable agent-browser skill

alchemistklk:
## Summary
- add an Open Design agent-browser skill that attaches to a verified local CDP Chrome endpoint
- document agent-managed CDP startup with `open -na`, connect-before-open ordering, and compact smoke testing
- add context hygiene for large prior tool outputs and richer stalled-agent diagnostics

## Validation
- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/skills.test.ts
- pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/providers/sse.test.ts
- pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts
- pnpm guard


lefarcen (commented):
Hey @alchemistklk! Reviewed the agent-browser skill addition.

## Overall
Solid implementation. The CDP connection contract (connect before open) is documented clearly, and the stalled-agent diagnostics are valuable for debugging hanging runs.

## Findings (all P2–P3, non-blocking)

### P2 — Context hygiene compliance
The skill doc says "Never print full upstream guides into chat or tool output" and shows the pattern of redirecting `agent-browser skills get core > "$TEMP_FILE"`, but I don't see a corresponding change in `server.ts` that enforces this when the skill is invoked.

If skills can still print arbitrary upstream guide content via `tool_result`, the inactivity watchdog improvement (`lastToolResultChars`) will record large outputs but not prevent them. Consider whether the skill invocation layer should strip or truncate oversized tool results, or if that's a future phase.

### P3 — i18n string consistency
The French/Russian copies use "agent-browser" as the bare tool name. Readers who don't know what agent-browser is might benefit from "Browser automation CLI" or similar descriptor in the translated strings, matching the English pattern of explaining "what it is" first.

### P3 — Test coverage for stalled-agent enhancements
The new `lastAgentEventPhase` / `childStdoutSeen` / `lastToolResultChars` tracking improves diagnostics, but I don't see a corresponding test in `chat-route.test.ts` or `skills.test.ts` that exercises the "agent stalled" error message with these fields populated. Consider adding a mock-spawn test that triggers the watchdog and asserts on the improved error message shape.

## Next step
None of the above block merge — they're refinements/future polish. If you agree on the oversized-tool-result question, you could open a follow-up issue. Otherwise LGTM.

PerishCode (approved):
Hey @alchemistklk — manual review since the auto-reviewer pipeline couldn't publish (digging into that separately).

LGTM. SKILL.md is thorough on the CDP startup gotchas (connect-before-open, the `DevToolsActivePort` crash signature, the manual-Chrome fallback). The stalled-agent diagnostics add real signal for hangs, and `buildPriorRunContextWarning` is a clean soft-enforcement of the "don't replay big tool outputs" policy from the skill doc. Tests cover the new code paths.

No blockers from me. lefarcen's three non-blocking notes are reasonable as you see fit.

PR #1259, opened 2026-05-11
feat(tools-pr): add maintainer PR-duty workspace

PerishCode:
## Summary

- add `tools/pr` as the maintainer-only PR-duty control plane: `list`, `view`, `classify`, `assignment` subcommands
- wire `pnpm tools-pr` into root scripts, postinstall builds, and the tools layout / residual-mjs allowlists in `scripts/guard.ts`
- root `AGENTS.md` gains a PR-duty tooling section linking to `tools/pr/AGENTS.md` for the tag dictionary and operational playbook

## Details

- 13 factual classify tags drive both list buckets and the assignment view: `bot-only-approval`, `needs-rebase`, `forbidden-surface`, `unlabeled`, `duplicate-title`, `non-ascii-slug`, `maintainer-edits-disabled`, `org-member`, `unresolved-changes-requested`, `stale-approval`, and three `awaiting-*` timing tags. Each rule is expressible as one factual sentence over `gh` data + repo paths; the accepted-vs-rejected list is in `tools/pr/AGENTS.md`.
- Tool is read-only on the PR surface — never approves / merges / comments / closes. Side effects stay in explicit `gh` invocations the maintainer runs.
- Heavy GitHub data (`reviews`, `comments`, `commits`, assignment timelines) uses cursor-paginated `gh api graphql` via `fetchPaginatedPrList` (page size 30) to stay under GitHub's GraphQL server-side timeout; light scalar chunks stay on `gh pr list --json`.
- `gh()` wraps `execFile` with a minimum-touch retry (2 attempts at 1s + 2s backoff) on transient 5xx / network errors only; 4xx, auth failures, and parse errors surface immediately.
- `classify --all` writes JSON to `.tmp/tools-pr/classify/<name|ts>.json` with rate-limit telemetry (before/after `rateLimit` snapshots, delta cost) so detector additions that inflate API cost surface in cron logs.
- `templates/*.md` are aesthetic references for the recurring maintainer comment shapes (not fill-in forms); `templates/examples/` holds frozen-in-time agent-review snapshots for three PR shapes.

## Validation

- pnpm install
- pnpm guard
- pnpm typecheck
- pnpm --filter @open-design/tools-pr typecheck
- pnpm --filter @open-design/tools-pr build
- pnpm exec tools-pr --help — registers list / view / classify / assignment
- pnpm exec tools-pr classify 1009 — `unresolved-changes-requested` fires against lefarcen's CHANGES_REQUESTED in `latestReviews`
[...]

PR #1285, opened 2026-05-11
fix(tools-pr): chunk stats fetch through cursor-paginated GraphQL

PerishCode:
Fixes a regression introduced by #1259's `--limit 1000` default.

## Summary

`fetchOpenPrs` was reading the stats chunk via `gh pr list --limit 1000 --json mergeStateStatus,...`. On the live 107-PR queue this 502s reliably, because GitHub's GraphQL gateway has to recompute merge state for every PR up front and the resulting query exceeds the gateway budget once the requested page passes ~60 PRs.

Switch the stats chunk to `fetchPaginatedPrList`, the same cursor-paginated GraphQL helper already used for reviews, comments, commits, and assignment timelines. Page size stays at `PR_LIST_PAGE_SIZE` (30), well within the gateway budget; the stats fetch is now consistent with the other heavy chunks.

## Details

- New `fetchOpenPrStats()` in `tools/pr/src/gh.ts` returns `GhStats[]` via `fetchPaginatedPrList` using the same field set as before (`number, additions, deletions, changedFiles, headRefName, headRefOid, baseRefName, mergeable, mergeStateStatus`).
- `fetchOpenPrs` swaps its `gh<GhStats[]>(...)` call for `fetchOpenPrStats()`. No behavior change to consumers — the resulting `GhStats[]` shape is identical.

## Validation

- `pnpm --filter @open-design/tools-pr typecheck`
- `pnpm --filter @open-design/tools-pr build`
- `pnpm tools-pr list` against the live 107-PR queue — completes without a 502 (reproduces the failure on `main` head `8c0fb8dc`)

### DIFF
--- a/tools/pr/src/gh.ts
+++ b/tools/pr/src/gh.ts
@@ -193,6 +193,29 @@ async function fetchPaginatedPrList<N>(nodeFields: string): Promise<N[]> {
  * the other heavy chunks (reviews, comments, assignment timelines).
  */
 
+/**
+ * Per-PR merge/diff stats — `additions`, `mergeStateStatus`, etc. `gh pr
+ * list --json mergeStateStatus,...` 502s once the requested page is
+ * above ~60 because the gateway has to recompute merge state for every
+ * PR up front. Cursor-paginated GraphQL at PR_LIST_PAGE_SIZE keeps each
+ * round trip well within the budget and consistent with how the other
+ * heavy chunks (reviews, comments, commits, assignment timelines) are
+ * fetched.
+ */
+async function fetchOpenPrStats(): Promise<GhStats[]> {
+  return fetchPaginatedPrList<GhStats>(
+    `number
+     additions
+     deletions
+     changedFiles
+     headRefName
+     headRefOid
+     baseRefName
+     mergeable
+     mergeStateStatus`,
+  );
+}
+
 async function fetchOpenPrReviews(): Promise<GhReviewsLite[]> {
   type Node = {
     number: number;
@@ -424,11 +447,7 @@ export async function fetchOpenPrs(
     "--json",
     "number,title,author,createdAt,updatedAt,isDraft,reviewDecision,labels,maintainerCanModify,assignees",
   ]);
-  const statsPromise = gh<GhStats[]>([
-    ...baseArgs,
-    "--json",
-    "number,additions,deletions,changedFiles,headRefName,headRefOid,baseRefName,mergeable,mergeStateStatus",
-  ]);
+  const statsPromise = fetchOpenPrStats();
   const filesPromise = gh<GhFiles[]>([...baseArgs, "--json", "number,files"]);
   const reviewsPromise = fetchOpenPrReviews();
   const commitsPromise = options.includeCommits


tools/pr/src/gh.ts
```diff
@@ -193,6 +193,29 @@ async function fetchPaginatedPrList<N>(nodeFields: string): Promise<N[]> {
...
+/**
+ * Per-PR merge/diff stats — `additions`, `mergeStateStatus`, etc. `gh pr
+ * list --json mergeStateStatus,...` 502s once the requested page is
+ * above ~60 because the gateway has to recompute merge state for every
+ * PR up front. Cursor-paginated GraphQL at PR_LIST_PAGE_SIZE keeps each
+ * round trip well within the budget and consistent with how the other
+ * heavy chunks (reviews, comments, commits, assignment timelines) are
+ * fetched.
+ */
+async function fetchOpenPrStats(): Promise<GhStats[]> {
```
lefarcen, line 205: P2 — `fetchOpenPrStats()` always walks the full open-PR GraphQL connection, while the previous `gh pr list` call inherited `--limit` through `baseArgs`. That means `tools-pr list --limit 20` still recomputes `mergeable` / `mergeStateStatus` for every open PR and only discards the extra rows later, defeating the documented small-limit preview escape hatch and increasing latency/API cost on larger queues.

Consider threading the requested limit into this helper, or have `fetchPaginatedPrList` accept a max row count so stats stop once the caller's limit is satisfied.

lefarcen (commented):
Hey @PerishCode! 👋

The paginated stats fetch fits the existing heavy-chunk GraphQL pattern and avoids the reported 502 failures. One design dimension to consider below.

tools/pr/src/gh.ts
```diff
@@ -193,6 +193,29 @@ async function fetchPaginatedPrList<N>(nodeFields: string): Promise<N[]> {
...
+  return fetchPaginatedPrList<GhStats>(
```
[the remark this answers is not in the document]
PerishCode: Deferring as a follow-up — same call-out is in the PR description. Default `--limit` is now 1000 (covers the live ~107-PR queue plus headroom), so the heavy GraphQL pages are bounded for the default path. Small `--limit` is the documented truncation-preview escape hatch and currently rare in practice. Threading a per-helper cap through `fetchPaginatedPrList` is on the follow-up list once a recurring small-limit usage pattern emerges; happy to prioritise if there's a specific value the tooling should optimise for.

tools/pr/src/gh.ts
PerishCode: Deferring as a follow-up — same as the parallel Codex thread on this file and the call-out in the PR body. Default `--limit` is now 1000 (covers the live ~107-PR queue plus headroom), so the heavy GraphQL pages are bounded on the default path. Small `--limit` is the documented truncation-preview escape hatch and currently rare; threading a per-helper cap through `fetchPaginatedPrList` is on the follow-up list. Happy to prioritise if you have a specific limit value the tooling should optimise for.

mrcfps (approved):
@PerishCode I verified that the stats fetch now uses the existing cursor-paginated GraphQL helper, keeps the `GhStats[]` shape unchanged for downstream consumers, and directly addresses the live-queue 502 regression described in the PR. Thanks for tightening up this path and keeping the heavier PR data fetches aligned.

PR #1286, opened 2026-05-11
fix(daemon/acp): terminate ACP child after clean prompt completion

### LINKED ISSUE #1265: Chat stays in "working" state after Devin agent response completes
### Summary
When using Devin for Terminal as the ACP agent, after Devin finishes responding to a prompt, the chat input area stays disabled in "working" state. The `Done · X.Xs` indicator appears on the assistant message as expected, but the input does not return to ready — the user must click stop before sending another message.

Reproduces on plain `main` (`421ddf55`). **The Codex agent on the same build does not hit this** — input correctly returns to ready after a Codex response completes — so the bug appears to be specific to Devin's ACP session lifecycle handling.

### Steps to reproduce
1. Configure Open Design with Devin for Terminal as the ACP agent
2. Open a new conversation
3. Send any prompt (e.g. `hi what model are you`)
4. Wait for Devin to fully finish responding — the `Done · 31s · 105 out` footer appears under the assistant message
5. Try to type or send another message

### Expected
Input returns to ready once `Done · ...` is shown; next message can be typed/sent.

### Actual
- `Done · X.Xs · N out` footer appears as expected
- Input area stays in "working" state — disabled
- Must click stop manually to unlock the input

### Comparison: Codex agent
The same build behaves correctly when using Codex as the ACP agent — the input returns to ready after a response completes. So the chat lifecycle / state-management on the web side is wired correctly; the gap is in how the Devin ACP session is being driven (most likely Devin not emitting a "turn complete" / session-end signal that the chat is expecting, or the daemon not relaying it).

### Environment
- Open Design: built from `main` @ `421ddf55`
- OS: macOS, Apple Silicon
- ACP agent: Devin for Terminal `2026.5.6-4`

### Notes
- In a separate local build that includes a `server.ts` session-pool change (keeping the Devin child process alive across prompts instead of respawning each one), this bug does **not** appear — strong hint that the session lifecycle / process respawn cycle is involved
- First flagged in #1208 review

MrRockySL:
Fixes #1265

## Summary

Bug B from the end-to-end validation of #1208: chat stays in `working` state after Devin (and any other ACP agent that doesn't exit on `stdin.end()`) finishes responding — the `Done · X.Xs · N out` indicator appears but the chat input never returns to ready. Manual stop click is required before sending the next message.

**Root cause:** Open Design's chat lifecycle finalizes runs via `child.on('close')` in `server.ts`. After a successful `session/prompt` response, `attachAcpSession` calls `stdin.end()` and waits for the child to exit on its own. Some ACP agents (notably Devin for Terminal) keep the child process alive past `stdin.end()`, waiting for the next prompt — so `child.on('close')` never fires and the chat run never finalizes.

**The fix (4 small targeted changes):**

- `apps/daemon/src/acp.ts` — After a clean `session/prompt` response, schedule a 500 ms grace period and then SIGTERM the child if it hasn't exited on its own. Mirrors the pattern `detectAcpModels()` already uses after the model-discovery probe (see same file, line ~270). Well-behaved agents that exit on `stdin.end()` are unaffected — the `child.once('close', ...)` handler clears the timer before SIGTERM fires.

- `apps/daemon/src/acp.ts` — New `completedSuccessfully()` method on the session handle reports whether the prompt resolved without a fatal error or abort. Lets the consumer distinguish clean-signal-exit from genuine-signal-failure.

- `apps/daemon/src/server.ts` — `child.on('close')` treats a SIGTERM exit as `'succeeded'` when `acpSession.completedSuccessfully()` is true. Without this, the chat run was marked `'failed'` because `code === 0` fails on a signal-driven exit (`code` is `null`, `signal` is `'SIGTERM'`).

- `apps/web/src/providers/daemon.ts` — Trust the server's authoritative `endStatus`. The signal/non-zero-code safety net no longer overrides an explicit `'succeeded'` status, so a clean ACP run no longer surfaces a fake `agent exited with signal SIGTERM` error in the chat UI.

**Codex comparison (independent confirmation):** Codex uses `streamFormat: 'json-event-stream'` rather than ACP, so it doesn't go through `attachAcpSession` at all — its CLI exits cleanly after each request, which is why Bug B never reproduces with Codex. The gap was in how the daemon was driving the ACP session lifecycle for agents that keep their child process alive.

This fix is scoped to the ACP transport and doesn't touch the model-selection path being fixed in #1208 — they're complementary fixes in adjacent code that can land in any order.

## Validation

- `pnpm --filter @open-design/daemon typecheck` — passes clean
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/acp.test.ts` — 8/8 pass (3 new tests cover the SIGTERM grace timer firing when the child doesn't exit, the timer being cleared when the child exits on its own, and `completedSuccessfully()` returning `false` on abort/fatal-error states)
- `pnpm guard` — passes (residual-JS, test-layout, e2e-layout, web-test-layout, tools-layout all clean)
- Manual UI verification on macOS (Apple Silicon):
  - Built via `pnpm exec tools-pack mac build --to app --portable` and installed locally
  - **Before the fix:** chat stuck in "working" state after Devin's `Done · 9.3s · 110 out`; manual stop click required before sending next prompt
  - **After the fix:** chat returns to ready automatically right after `Done · ...` appears; no fake error banner; next prompt can be typed and sent immediately
  - Tested with Devin for Terminal `2026.5.6-4`
  - Screenshot + screen recording of the fixed behavior at
Screenshot 2026-05-11 at 17 04 48
tached below

https://github.com/user-attachments/assets/83a5cb55-b615-480b-a30e-a202830a487e

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -4153,17 +4153,25 @@ export async function startServer({
       }
       // ACP agents that don't shut down on stdin.end() (e.g. Devin for
       // Terminal) are forced to exit via SIGTERM from attachAcpSession after
-      // a clean prompt completion. Without this check, the chat run would be
-      // marked `failed` because `code === 0` fails (code is null on a signal
-      // exit). `completedSuccessfully()` reports whether the ACP session
-      // resolved without a fatal error or abort, so we treat that as success
-      // even when the underlying process exit was signal-driven.
+      // a clean prompt completion. Without an override, the chat run would
+      // be marked `failed` because `code === 0` fails (code is null on a
+      // signal exit). `completedSuccessfully()` reports whether the ACP
+      // session resolved without a fatal error or abort.
+      //
+      // Scope the override narrowly to the exact forced-shutdown shape this
+      // PR introduces: code is null AND signal is SIGTERM AND the ACP
+      // session reported clean completion. Any other post-response failure
+      // (non-zero exit code, SIGKILL, SIGSEGV, etc.) still propagates as
+      // `failed`, preserving the existing close-status behavior for genuine
+      // post-response process problems.
       const acpCleanCompletion =
         typeof acpSession?.completedSuccessfully === 'function' &&
         acpSession.completedSuccessfully();
+      const acpForcedShutdown =
+        code === null && signal === 'SIGTERM' && acpCleanCompletion;
       const status = run.cancelRequested
         ? 'canceled'
-        : code === 0 || acpCleanCompletion
+        : code === 0 || acpForcedShutdown
           ? 'succeeded'
           : 'failed';
       if (status === 'failed') {
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -4151,9 +4151,19 @@ export async function startServer({
         ));
         return design.runs.finish(run, 'failed', code, signal);
       }
+      // ACP agents that don't shut down on stdin.end() (e.g. Devin for
+      // Terminal) are forced to exit via SIGTERM from attachAcpSession after
+      // a clean prompt completion. Without this check, the chat run would be
+      // marked `failed` because `code === 0` fails (code is null on a signal
+      // exit). `completedSuccessfully()` reports whether the ACP session
+      // resolved without a fatal error or abort, so we treat that as success
+      // even when the underlying process exit was signal-driven.
+      const acpCleanCompletion =
+        typeof acpSession?.completedSuccessfully === 'function' &&
+        acpSession.completedSuccessfully();
       const status = run.cancelRequested
         ? 'canceled'
-        : code === 0
+        : code === 0 || acpCleanCompletion
           ? 'succeeded'
           : 'failed';
       if (status === 'failed') {
--- a/apps/daemon/tests/acp.test.ts
+++ b/apps/daemon/tests/acp.test.ts
@@ -2,7 +2,7 @@ import assert from 'node:assert/strict';
 import { EventEmitter } from 'node:events';
 import { PassThrough } from 'node:stream';
 import path from 'node:path';
-import { test } from 'vitest';
+import { test, vi } from 'vitest';
 import { attachAcpSession, buildAcpSessionNewParams, normalizeModels } from '../src/acp.js';
 
 const DEFAULT_MODEL_OPTION = { id: 'default', label: 'Default (CLI config)' };
@@ -252,6 +252,101 @@ function agentModelStatuses(events: Array<{ event: string; payload: unknown }>):
     })
     .map((entry) => (entry.payload as { model?: unknown }).model);
 }
+
+test('attachAcpSession force-terminates the child after a clean prompt completion if it does not exit on stdin.end()', async () => {
+  vi.useFakeTimers();
+  try {
+    const child = new FakeAcpChild();
+
+    const session = attachAcpSession({
+      child: child as never,
+      prompt: 'hello',
+      cwd: '/tmp/od-project',
+      model: null,
+      mcpServers: [],
+      send: () => {},
+    });
+
+    // Drive the protocol through to a clean prompt completion.
+    child.stdout.write(`${JSON.stringify({ id: 1, result: {} })}\n`);
+    child.stdout.write(`${JSON.stringify({ id: 2, result: { sessionId: 'session-1' } })}\n`);
+    child.stdout.write(`${JSON.stringify({ id: 3, result: { usage: { inputTokens: 1, outputTokens: 2 } } })}\n`);
+
+    // Child has not exited yet (simulates Devin for Terminal keeping the
+    // process alive past stdin.end()).
+    assert.equal(child.killed, false);
+
+    // After the grace period elapses, attachAcpSession should SIGTERM the
+    // child so child.on('close') can fire and the chat run can finalize.
+    await vi.advanceTimersByTimeAsync(500);
+    assert.equal(child.killed, true);
+
+    // The session reports the prompt completed successfully so the consumer
+    // can mark the run as 'succeeded' even though the underlying exit was
+    // signal-driven.
+    assert.equal(session.completedSuccessfully(), true);
+    assert.equal(session.hasFatalError(), false);
+  } finally {
+    vi.useRealTimers();
+  }
+});
+
+test('attachAcpSession does not double-kill a child that exits cleanly on stdin.end()', async () => {
+  vi.useFakeTimers();
+  try {
+    const child = new FakeAcpChild();
+
+    attachAcpSession({
+      child: child as never,
+      prompt: 'hello',
+      cwd: '/tmp/od-project',
+      model: null,
+      mcpServers: [],
+      send: () => {},
+    });
+
+    child.stdout.write(`${JSON.stringify({ id: 1, result: {} })}\n`);
+    child.stdout.write(`${JSON.stringify({ id: 2, result: { sessionId: 'session-1' } })}\n`);
+    child.stdout.write(`${JSON.stringify({ id: 3, result: {} })}\n`);
+
+    // Well-behaved agent exits on its own before the grace period elapses.
+    child.emit('close', 0, null);
+    assert.equal(child.killed, false);
+
+    // The grace-period timer should have been cleared by the close handler,
+    // so advancing time should not trigger a SIGTERM on the now-closed child.
+    await vi.advanceTimersByTimeAsync(2_000);
+    assert.equal(child.killed, false);
+  } finally {
+    vi.useRealTimers();
+  }
+});
+
+test('attachAcpSession.completedSuccessfully reflects abort and fatal-error states', () => {
+  const child = new FakeAcpChild();
+
+  const session = attachAcpSession({
+    child: child as never,
+    prompt: 'hello',
+    cwd: '/tmp/od-project',
+    model: null,
+    mcpServers: [],
+    send: () => {},
+  });
+
+  // Before any protocol traffic the session is not yet complete.
+  assert.equal(session.completedSuccessfully(), false);
+
+  // Drive through session creation, then abort before the prompt completes.
+  child.stdout.write(`${JSON.stringify({ id: 1, result: {} })}\n`);
+  child.stdout.write(`${JSON.stringify({ id: 2, result: { sessionId: 'session-1' } })}\n`);
+  session.abort();
+
+  // Aborted runs are not "successful completions" even though `finished` is
+  // set internally — the consumer should treat them as canceled, not
+  // succeeded.
+  assert.equal(session.completedSuccessfully(), false);
+});
 
 class FakeAcpChild extends EventEmitter {
   stdin = new PassThrough();
--- a/apps/web/src/providers/daemon.ts
+++ b/apps/web/src/providers/daemon.ts
@@ -453,7 +453,19 @@ async function consumeDaemonRun({
 
     if (endStatus === 'canceled') return;
 
-    if (endStatus === 'failed' || exitSignal || (exitCode !== null && exitCode !== 0)) {
+    // Trust the server's authoritative `endStatus`. When the server reports
+    // `succeeded`, the run completed cleanly even if the underlying process
+    // exited via a signal — some agents (e.g. ACP agents like Devin for
+    // Terminal) intentionally exit via SIGTERM after a clean prompt
+    // completion because they don't shut down on stdin.end(). The
+    // signal/non-zero-code safety net only applies when the server hasn't
+    // explicitly declared success, so we never surface a fake error on a
+    // run the server already marked successful.
+    const looksLikeFailure =
+      endStatus === 'failed' ||
+      (endStatus !== 'succeeded' &&
+        (exitSignal || (exitCode !== null && exitCode !== 0)));
+    if (looksLikeFailure) {
       const tail = stderrBuf.trim().slice(-400);
       handlers.onError(
         new Error(`agent exited with ${exitSignal ? `signal ${exitSignal}` : `code ${exitCode}`}${tail ? `\n${tail}` : ''}`),
--- a/apps/web/src/providers/daemon.ts
+++ b/apps/web/src/providers/daemon.ts
@@ -322,6 +322,15 @@ async function consumeDaemonRun({
   let exitCode: number | null = null;
   let exitSignal: string | null = null;
   let endStatus: ChatRunStatus | null = null;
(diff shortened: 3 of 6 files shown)


MrRockySL:
Thanks @codex — good catch. Pushed `420e5c62` to address it.

`attachAgentStreamHandlers()` in `apps/daemon/src/connectionTest.ts` now also respects `acpSession.completedSuccessfully?.()`, mirroring the same check we apply on the chat-run path in `server.ts`. So a clean prompt response followed by our new SIGTERM no longer trips `winner.signal === 'SIGTERM'` → `exitedCleanly = false` → `agent_spawn_failed` for ACP agents that keep their child alive past `stdin.end()`.

Also widened the `AgentSpawnHandle` type so `completedSuccessfully` is visible on the structural session type used inside `connectionTest.ts`.

Ran `pnpm --filter @open-design/daemon typecheck`, `pnpm guard`, and the daemon `acp.test.ts` + `connection-test.test.ts` files (56 tests total) — all passing.

apps/daemon/src/server.ts
```diff
@@ -4107,9 +4107,19 @@ export async function startServer({
...
+      // exit). `completedSuccessfully()` reports whether the ACP session
+      // resolved without a fatal error or abort, so we treat that as success
+      // even when the underlying process exit was signal-driven.
+      const acpCleanCompletion =
+        typeof acpSession?.completedSuccessfully === 'function' &&
+        acpSession.completedSuccessfully();
       const status = run.cancelRequested
         ? 'canceled'
-        : code === 0
+        : code === 0 || acpCleanCompletion
```
Siri-Ray: This success check is broader than the SIGTERM case described above it. `completedSuccessfully()` becomes true as soon as the ACP `session/prompt` response is processed, but it does not say why the child later closed. With this predicate, an ACP agent that returns a prompt result and then exits with code `1`, or is killed by a different signal such as `SIGKILL`, is still persisted as `succeeded`; the web change then trusts that status and suppresses the failure banner. That regresses the existing close-status behavior for real post-response process failures. Please limit this override to the forced-shutdown shape this PR introduces, for example `code === null && signal === 'SIGTERM' && acpCleanCompletion`, and mirror the same predicate in `apps/daemon/src/connectionTest.ts` where `exitedCleanly` now uses `acpCleanCompletion` unconditionally.

Siri-Ray (commented):
Found one blocking lifecycle issue in the ACP close handling; details inline.


apps/daemon/tests/acp.test.ts
```diff
@@ -82,6 +82,101 @@ test('attachAcpSession exposes abort and sends session cancel after session crea
...
+    // After the grace period elapses, attachAcpSession should SIGTERM the
+    // child so child.on('close') can fire and the chat run can finalize.
+    await vi.advanceTimersByTimeAsync(500);
+    assert.equal(child.killed, true);
+
+    // The session reports the prompt completed successfully so the consumer
+    // can mark the run as 'succeeded' even though the underlying exit was
+    // signal-driven.
+    assert.equal(session.completedSuccessfully(), true);
+    assert.equal(session.hasFatalError(), false);
```
lefarcen, line 288: **P3 — Test readability:** The test name says "does not double-kill" but the assertion only checks `child.killed === false` after the close event. To fully demonstrate no double-kill, consider explicitly advancing timers past the grace period after close (e.g. `await vi.advanceTimersByTimeAsync(2_000)`) and asserting `child.killed` is still false. Current test works, but the extra timer advance would make the "no double-kill even after waiting" claim bulletproof.

lefarcen (dismissed):
Hey @MrRockySL, clean fix for a tricky lifecycle issue! 🎉

I verified:
- The 500ms grace + SIGTERM pattern accurately mirrors detectAcpModels (line 273)
- completedSuccessfully() backward-compat is handled correctly (typeof check before call)
- Test coverage hits the key scenarios: timer fires when child stays alive, timer cleared when child exits cleanly, abort/fatal states tracked
- The server.ts + connectionTest.ts + web/daemon.ts changes all trust the ACP-level success signal consistently

One small observation below on test readability, but nothing blocking.

LGTM — nice work hunting down the root cause and keeping the fix scoped.

MrRockySL:
Thanks for the reviews @Siri-Ray and @lefarcen!

**@Siri-Ray** — addressed in `0b6300d1`. The override is now scoped to the exact forced-shutdown shape this PR introduces:

```ts
code === null && signal === 'SIGTERM' && acpCleanCompletion
```

Applied to both `apps/daemon/src/server.ts` (chat-run finalization) and `apps/daemon/src/connectionTest.ts` (connection-test classification). Any other post-response failure (non-zero exit, SIGKILL, SIGSEGV, etc.) now falls through to `failed` / `agent_spawn_failed` as before, preserving the existing close-status behavior for genuine post-response process problems. Updated comments in both files to spell out the scoping rationale.

`pnpm --filter @open-design/daemon typecheck`, `pnpm guard`, and the daemon `acp.test.ts` + `connection-test.test.ts` files (59 tests total) all still pass.

**@lefarcen** — thanks for the approval and the detailed verification 🙏. On the test-readability nit: the `does not double-kill` test already does the timer-advance-after-close pattern you suggested — see `apps/daemon/tests/acp.test.ts` lines 145-148 (right after the `child.emit('close', 0, null)` + first `assert.equal(child.killed, false)`):

```ts
// The grace-period timer should have been cleared by the close handler,
// so advancing time should not trigger a SIGTERM on the now-closed child.
await vi.advanceTimersByTimeAsync(2_000);
assert.equal(child.killed, false);
```

Happy to expand the comment further or restructure if you think it'd be clearer to read — just let me know what shape works for you.

apps/web/src/providers/daemon.ts
```diff
@@ -367,7 +367,19 @@ async function consumeDaemonRun({
...
+    // exited via a signal — some agents (e.g. ACP agents like Devin for
+    // Terminal) intentionally exit via SIGTERM after a clean prompt
+    // completion because they don't shut down on stdin.end(). The
+    // signal/non-zero-code safety net only applies when the server hasn't
+    // explicitly declared success, so we never surface a fake error on a
+    // run the server already marked successful.
+    const looksLikeFailure =
+      endStatus === 'failed' ||
+      (endStatus !== 'succeeded' &&
+        (exitSignal || (exitCode !== null && exitCode !== 0)));
```
Siri-Ray, line 491: This failure gate now treats every `endStatus === 'succeeded'` as authoritative, but line 347 also manufactures `succeeded` whenever the SSE `end` event has a missing or invalid `status`. That matters because the shared contract still allows status-less end payloads (`ChatSseEndPayload.status` is optional in `packages/contracts/src/sse/chat.ts`, and the contract examples still emit `event: 'end', data: { code: 0 }`). A compatible or older daemon response like `event: end` with `{"code":1}` or `{"code":null,"signal":"SIGTERM"}` was previously caught by the exit-code/signal safety net, but with this change it reaches `handlers.onDone(acc)` and suppresses the failure banner. Please distinguish an explicit server-declared success from the local fallback, for example by tracking whether `event.data.status === 'succeeded'` was actually present and only bypassing the exit-code/signal check for that explicit success; keep the safety net for missing/invalid status. A web SSE regression test with a status-less non-zero code or signal would lock this down.

Siri-Ray (commented):
Found one blocking client-side status regression in the daemon SSE handling; details inline.


apps/web/src/providers/daemon.ts
lefarcen, line 491: P1 — This trusts `endStatus === 'succeeded'` absolutely, but line 347 manufactures `succeeded` as a fallback when the SSE `end` event has missing/invalid `status`. That means a compatible/older daemon response with `{code:1}` or `{code:null,signal:"SIGTERM"}` would bypass the exit-code/signal safety net and suppress the failure banner.

Please distinguish an explicit server-declared success from the local fallback. For example, track whether `event.data.status === 'succeeded'` was actually present, and only bypass the exit-code/signal check for that explicit success. Keep the safety net for missing/invalid status.

A web SSE regression test with a status-less non-zero code or signal would lock this down.

lefarcen (dismissed):
Hey @MrRockySL, thanks for addressing Siri-Ray's prior feedback — narrowing the override to the exact forced-SIGTERM shape is the right call. 👍

Found one blocking issue in the web SSE handling that Siri-Ray flagged — see inline. Once that's addressed, this should be good to go.

MrRockySL:
Thanks @Siri-Ray and @lefarcen — addressed in `2a911511`. You both flagged a real backward-compat regression, my earlier web change was too eager to trust `endStatus === 'succeeded'`.

**What was wrong:** `endStatus` can be `'succeeded'` in two distinct ways:
1. The SSE end event explicitly carries `status: 'succeeded'` (authoritative).
2. The end event omits or has an invalid `status` field and `apps/web/src/providers/daemon.ts` falls back to `'succeeded'` as a local default (line 347 of the prior diff).

The previous bypass treated both as the same. A compatible/older daemon emitting `{code:1}` or `{code:null,signal:"SIGTERM"}` with no `status` would skip the failure banner.

**Fix:** Introduced `serverDeclaredSuccess`, set true only when:
- The SSE end event has `status === 'succeeded'`, **or**
- The fallback `fetchChatRunStatus` REST path returns `status === 'succeeded'` (which the existing `isChatRunStatus()` guard already proves is explicit).

The safety net is now bypassed only on that explicit signal — the local-fallback `'succeeded'` keeps hitting the exit-code/signal check, so real failures still surface as before.

**Added three regression tests in `apps/web/tests/providers/sse.test.ts`:**

- Explicit `status: 'succeeded'` + SIGTERM → `onDone` called, no error (the ACP success path stays intact)
- End event with `{code:1}` and no `status` → `onError('agent exited with code 1')` (the regression guard you asked for)
- End event with `{code:null,signal:'SIGTERM'}` and no `status` → `onError('agent exited with signal SIGTERM')` (signal-arm regression guard)

All three pass; `pnpm guard` + daemon typecheck clean; SSE test suite is now 27/27 (was 24).

Re-requesting review.

lefarcen (dismissed):
Hey @MrRockySL, nice fix! 🎉

The `serverDeclaredSuccess` flag correctly distinguishes explicit `status: 'succeeded'` from the local fallback, keeping the safety net intact for missing-status failures. The 3 regression tests lock down exactly the scenarios Siri-Ray and I were worried about:

1. Explicit `status:"succeeded"` + SIGTERM → `onDone` (ACP success path works)
2. `{code:1}` no status → `onError` (safety net catches non-zero exit)
3. `{code:null,signal:"SIGTERM"}` no status → `onError` (safety net catches signal)

Clean fix, good test coverage. LGTM.

lefarcen:
Hey @MrRockySL — heads up, this PR now has merge conflicts with `main` (likely from recent changes to the same files). Could you rebase or merge `main` into your branch to resolve them? Once that's done and CI is green, we'll be ready for final review. 🚀

MrRockySL:
Rebased onto latest `main` (now `5bd97631`). One conflict in `apps/web/tests/providers/sse.test.ts` — both this PR and #1267 added new tests right after the existing `reads unified SSE error payload messages` test. Resolution kept both: #1267's `includes unified SSE error details in daemon error messages` first, then the three regression tests this PR adds (explicit-success-+-SIGTERM, status-less + non-zero code, status-less + signal). The other auto-merge in `apps/web/src/providers/daemon.ts` came through clean.

All checks rerun on the rebased branch (`86d2885e`):

- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/acp.test.ts tests/connection-test.test.ts` — 64/64 pass
- `pnpm --filter @open-design/web exec vitest run tests/providers/sse.test.ts` — 30/30 pass
- `pnpm guard` — clean

Ready for re-review whenever you have time. 🙏

lefarcen (dismissed):
Approved. The rebase onto main (5bd97631) and conflict resolution in sse.test.ts look clean — kept both this PR's and #1267's tests in sequence. Core logic matches the head I approved earlier (2a911511); author verified all test suites pass. LGTM.

apps/daemon/src/server.ts
Siri-Ray: Looper checked this thread against head `86d2885ea6fb65322f2d4a76402e8ae66bf02e43`. The requested change appears objectively addressed by apps/daemon/src/server.ts now defines acpForcedShutdown as code === null && signal === 'SIGTERM' && acpCleanCompletion before marking succeeded; apps/daemon/src/connectionTest.ts mirrors the same winner.code === null && winner.signal === 'SIGTERM' && acpCleanCompletion predicate for exitedCleanly., so I’m resolving this thread. Reopen if this still needs discussion.



Siri-Ray (dismissed):
@MrRockySL I reviewed the current changed ranges for the ACP lifecycle fix, the connection-test classification, and the web SSE completion handling after the rebase. The forced shutdown path is now scoped to a clean ACP `SIGTERM`, the client only bypasses the exit-code/signal safety net on explicit server success, and the added regression tests cover the important success and backward-compat failure cases. Thanks for the careful follow-up on the earlier review feedback and for keeping this fix focused.


MrRockySL:
Rebased again onto latest `main` (`aeb6cde9`) now that #1208 has merged. As expected, our `apps/daemon/src/acp.ts` and the new configOptions code from #1208 sit next to each other inside `attachAcpSession` — Git auto-merged both daemon source files cleanly, with one trivial import conflict in `apps/daemon/tests/acp.test.ts` (combined the two import lines into one).

All checks rerun on the rebased branch (`81f4e380`):

- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/acp.test.ts tests/connection-test.test.ts` — 68/68 pass (gained 4 from #1208's new tests after the merge)
- `pnpm --filter @open-design/web exec vitest run tests/providers/sse.test.ts` — 30/30 pass
- `pnpm guard` — clean

Same logic as the previously-approved head (`86d2885e`) — no semantic changes. Re-requesting reviews now.

Siri-Ray (approved):
@MrRockySL I reviewed the current changed ranges at `81f4e380`: the ACP prompt-completion shutdown path, the daemon connection-test classification, and the web SSE completion handling after the latest rebase. The forced shutdown is narrowly scoped to clean ACP `SIGTERM`, the client only bypasses the exit-code/signal safety net on explicit server-declared success, and the added regression coverage exercises both the intended success path and the backward-compatible missing-status failure cases. Thanks for the careful follow-up on the earlier review feedback and for keeping this fix focused.

I could not rerun the focused Vitest suites in this local worktree because `vitest` is not installed/resolved here, but the diff review did not surface any actionable issues.

PR #1163, opened 2026-05-10
feat: add 404 error page skill

xxiaoxiong:
## Summary

Adds a new **404 error page skill** for generating friendly, helpful error pages that guide users back to useful content.

### SKILL.md
- **Triggers**: `404`, `404 page`, `error page`, `not found`, `page not found`, `错误页面`, `页面未找到`
- **Workflow**: Detailed instructions for creating 404 pages with customizable tone
- **Tone examples**: Playful, Professional, and Minimal variations
- **Output contract**: Standard artifact format
- **Helpful link suggestions**: Common navigation destinations

### example.html
A complete, modern 404 error page demonstrating:
- ✅ **Eye-catching visual** - Large gradient 404 number with floating animation
- ✅ **Friendly messaging** - Clear, helpful copy (not technical jargon)
- ✅ **Search functionality** - Prominent search bar to help users find content
- ✅ **Helpful navigation** - 4 primary links + 3 secondary links
- ✅ **Beautiful design** - Gradient background, smooth animations, hover effects
- ✅ **Mobile-responsive** - Adapts to small screens with adjusted typography
- ✅ **Accessibility** - Semantic HTML, proper ARIA labels

## Use cases

- Error handling pages
- User experience improvement
- Site navigation recovery
- Brand consistency in error states
- Reducing bounce rate from broken links

## Design

The example features:
- **Gradient background** - Purple gradient (customizable to brand colors)
- **Centered card layout** - White container with rounded corners and shadow
- **Floating animation** - Subtle up/down motion on the 404 number
- **Gradient text** - 404 number uses gradient fill for visual interest
- **Smooth transitions** - Hover effects on all interactive elements

## Technical details

- **No external dependencies** - Pure HTML/CSS
- **Instant load** - No external images or fonts
- **Lightweight** - ~5KB total size
- **Customizable** - Easy to adapt tone and links to any brand

## Preview

The skill generates pages with:
1. Large 404 number with gradient effect
2. Friendly headline and description
3. Search bar (optional)
4. 4 primary navigation buttons
5. 3 secondary text links

Ready for review!

PR #1287, opened 2026-05-11
fix(tools-pr): fall back on reviewDecision for unresolved-changes-requested

PerishCode:
## Summary

Patrol classify on the live 102-PR queue surfaced a coverage gap: three PRs (#1101, #1127, #1163) had GitHub-level \`reviewDecision=CHANGES_REQUESTED\` but \`tools-pr classify\` didn't tag them as \`unresolved-changes-requested\`. Add a narrow fallback so the tag fires in that shape.

## Details

The gap is a divergence between two notions of "latest review state per reviewer":

- GitHub's \`reviewDecision\` keeps a reviewer's \`CHANGES_REQUESTED\` in effect until that same reviewer submits \`APPROVED\` or \`DISMISSED\`. A subsequent \`COMMENTED\` review by the same reviewer does NOT supersede it.
- Our \`reduceLatestReviewsByAuthor\` collapses every reviewer to their latest review with no special-casing of state, so \`CHANGES_REQUESTED\` followed by \`COMMENTED\` disappears from the reduced view.

\`tagUnresolvedChangesRequested\` filtered the reduced view for \`state === "CHANGES_REQUESTED"\`. The three PRs above each had a reviewer write \`CHANGES_REQUESTED\` → \`COMMENTED\`, so they escaped the rule even though \`reviewDecision\` was still \`CHANGES_REQUESTED\`.

The fix keeps the existing per-reviewer path intact, then adds a fallback: when no per-reviewer \`CHANGES_REQUESTED\` is visible after reduction, trust \`facts.reviewDecision === "CHANGES_REQUESTED"\` instead. The fallback uses a distinct \`source\` token (\`gh.reviewDecision\` vs \`gh.latestReviews[].state\`) so report consumers can tell which path fired.

Reducer semantics deliberately left alone — flipping \`COMMENTED\` handling there would cascade to \`bot-only-approval\`, \`stale-approval\`, and \`humanReviewerSignalAt\`, each of which has its own correctness story.

## Validation

- \`pnpm --filter @open-design/tools-pr typecheck\`
- \`pnpm --filter @open-design/tools-pr build\`
- \`pnpm tools-pr classify 1101\`, \`1127\`, \`1163\` — each now emits \`unresolved-changes-requested\` via the fallback path (\`source: gh.reviewDecision\`)
- \`pnpm tools-pr classify 305\` — still emits via the existing per-reviewer path (\`source: gh.latestReviews[].state\`); first path unchanged
- \`pnpm tools-pr classify 1219\` — APPROVED PR continues to not emit the tag

### DIFF
--- a/tools/pr/src/tags.ts
+++ b/tools/pr/src/tags.ts
@@ -123,16 +123,17 @@ function tagUnresolvedChangesRequested(facts: PrFacts): Tag | null {
       source: "gh.latestReviews[].state",
     };
   }
-  // GitHub's reviewDecision keeps a reviewer's CHANGES_REQUESTED in effect
-  // until that same reviewer submits APPROVED or DISMISSED — COMMENTED does
-  // NOT supersede it. Our reducer collapses every reviewer to their latest
-  // state with no special-casing, so the per-reviewer CHANGES_REQUESTED
-  // gets hidden once they follow up with COMMENTED. Fall back to the
-  // PR-level reviewDecision so the tag still fires in that shape.
+  // The reducer-side path misses cases where GitHub's reviewDecision still
+  // reports CHANGES_REQUESTED but the latest-per-author reduction of fetched
+  // reviews carries none — e.g. the reviewer's CR is followed by COMMENTED
+  // (COMMENTED does not supersede CR in GitHub's decision logic), or the CR
+  // sits outside the `reviews(last: 30)` window. Either way the PR-level
+  // decision is the authoritative signal; fall back to it without asserting
+  // a specific cause.
   if (facts.reviewDecision === "CHANGES_REQUESTED") {
     return {
       name: "unresolved-changes-requested",
-      reason: "reviewDecision=CHANGES_REQUESTED at PR level (no per-reviewer CHANGES_REQUESTED visible after latest-per-author reduction; a reviewer's earlier CHANGES_REQUESTED was followed by COMMENTED)",
+      reason: "reviewDecision=CHANGES_REQUESTED at PR level; no per-reviewer CHANGES_REQUESTED state in latest-per-author reduction of fetched reviews",
       source: "gh.reviewDecision",
     };
   }
--- a/tools/pr/src/tags.ts
+++ b/tools/pr/src/tags.ts
@@ -116,12 +116,27 @@ function tagUnresolvedChangesRequested(facts: PrFacts): Tag | null {
     .filter((r) => r.state === "CHANGES_REQUESTED" && r.author?.login)
     .map((r) => r.author?.login)
     .filter((login): login is string => typeof login === "string");
-  if (reviewers.length === 0) return null;
-  return {
-    name: "unresolved-changes-requested",
-    reason: `latestReviews carries CHANGES_REQUESTED from: ${[...new Set(reviewers)].join(", ")}`,
-    source: "gh.latestReviews[].state",
-  };
+  if (reviewers.length > 0) {
+    return {
+      name: "unresolved-changes-requested",
+      reason: `latestReviews carries CHANGES_REQUESTED from: ${[...new Set(reviewers)].join(", ")}`,
+      source: "gh.latestReviews[].state",
+    };
+  }
+  // GitHub's reviewDecision keeps a reviewer's CHANGES_REQUESTED in effect
+  // until that same reviewer submits APPROVED or DISMISSED — COMMENTED does
+  // NOT supersede it. Our reducer collapses every reviewer to their latest
+  // state with no special-casing, so the per-reviewer CHANGES_REQUESTED
+  // gets hidden once they follow up with COMMENTED. Fall back to the
+  // PR-level reviewDecision so the tag still fires in that shape.
+  if (facts.reviewDecision === "CHANGES_REQUESTED") {
+    return {
+      name: "unresolved-changes-requested",
+      reason: "reviewDecision=CHANGES_REQUESTED at PR level (no per-reviewer CHANGES_REQUESTED visible after latest-per-author reduction; a reviewer's earlier CHANGES_REQUESTED was followed by COMMENTED)",
+      source: "gh.reviewDecision",
+    };
+  }
+  return null;
 }
 
 function tagStaleApproval(facts: PrFacts): Tag | null {
--- a/tools/pr/src/tags.ts
+++ b/tools/pr/src/tags.ts
@@ -129,7 +129,9 @@ function tagUnresolvedChangesRequested(facts: PrFacts): Tag | null {
   // (COMMENTED does not supersede CR in GitHub's decision logic), or the CR
   // sits outside the `reviews(last: 30)` window. Either way the PR-level
   // decision is the authoritative signal; fall back to it without asserting
-  // a specific cause.
+  // a specific cause. Observed scale on the live queue: 3 of 102 open PRs
+  // (#1101 / #1127 / #1163) hit this gap, so this is a recurring pattern,
+  // not a theoretical edge case.
   if (facts.reviewDecision === "CHANGES_REQUESTED") {
     return {
       name: "unresolved-changes-requested",
--- a/tools/pr/AGENTS.md
+++ b/tools/pr/AGENTS.md
@@ -45,7 +45,7 @@ Each tag has a single mechanical rule. Adding new tags requires the rule to be e
 | `non-ascii-slug` | A design-system root touched by the PR has a slug that fails `/^[a-z0-9-]+$/` | files + lane.DESIGN_DIR |
 | `maintainer-edits-disabled` | `maintainerCanModify === false` | gh.maintainerCanModify |
 | `org-member` | PR author's GitHub login appears in `gh api orgs/<repo-owner>/members` | gh REST orgs members list |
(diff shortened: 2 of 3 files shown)


tools/pr/src/tags.ts
```diff
@@ -116,12 +116,27 @@ function tagUnresolvedChangesRequested(facts: PrFacts): Tag | null {
...
+  // GitHub's reviewDecision keeps a reviewer's CHANGES_REQUESTED in effect
+  // until that same reviewer submits APPROVED or DISMISSED — COMMENTED does
+  // NOT supersede it. Our reducer collapses every reviewer to their latest
+  // state with no special-casing, so the per-reviewer CHANGES_REQUESTED
+  // gets hidden once they follow up with COMMENTED. Fall back to the
+  // PR-level reviewDecision so the tag still fires in that shape.
+  if (facts.reviewDecision === "CHANGES_REQUESTED") {
+    return {
+      name: "unresolved-changes-requested",
+      reason: "reviewDecision=CHANGES_REQUESTED at PR level (no per-reviewer CHANGES_REQUESTED visible after latest-per-author reduction; a reviewer's earlier CHANGES_REQUESTED was followed by COMMENTED)",
```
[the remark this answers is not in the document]
PerishCode: Fixed in 370943bd: dropped the inferred-cause clause from the fallback reason. The condition observes only (a) reviewDecision=CHANGES_REQUESTED at PR level and (b) no per-reviewer CHANGES_REQUESTED in the reduced view; multiple upstream configurations satisfy that pair (COMMENTED follow-up, CR outside the `reviews(last: 30)` window, etc.), and asserting any specific one violates the strictly-factual rule. Possible upstream causes are documented in the code comment above the branch instead.

tools/pr/src/tags.ts
```diff
@@ -116,12 +116,28 @@ function tagUnresolvedChangesRequested(facts: PrFacts): Tag | null {
...
+  // reviews carries none — e.g. the reviewer's CR is followed by COMMENTED
+  // (COMMENTED does not supersede CR in GitHub's decision logic), or the CR
+  // sits outside the `reviews(last: 30)` window. Either way the PR-level
+  // decision is the authoritative signal; fall back to it without asserting
+  // a specific cause.
+  if (facts.reviewDecision === "CHANGES_REQUESTED") {
+    return {
+      name: "unresolved-changes-requested",
+      reason: "reviewDecision=CHANGES_REQUESTED at PR level; no per-reviewer CHANGES_REQUESTED state in latest-per-author reduction of fetched reviews",
+      source: "gh.reviewDecision",
```
Siri-Ray, line 139: The new fallback emits `unresolved-changes-requested` with `source: "gh.reviewDecision"`, but `tools/pr/AGENTS.md` still documents this tag as only `Any reviewer's latest review has state === "CHANGES_REQUESTED"` with data source `gh.latestReviews[].state`. That tag dictionary is the documented contract for `tools-pr classify`, and after this change a report can contain a valid source/rule that the dictionary says cannot exist. Please update the `unresolved-changes-requested` row to describe both paths, for example: latest per-reviewer `CHANGES_REQUESTED`, or `reviewDecision === "CHANGES_REQUESTED"` when the reduced fetched reviews contain no per-reviewer CR, and list both `gh.latestReviews[].state` and `gh.reviewDecision` as sources.

Siri-Ray (commented):
Found one non-blocking consistency issue: the fallback is narrow, but the tag dictionary should describe the new `gh.reviewDecision` source path so report consumers can interpret it correctly.


tools/pr/src/tags.ts
```diff
@@ -116,12 +116,28 @@ function tagUnresolvedChangesRequested(facts: PrFacts): Tag | null {
...
+      source: "gh.latestReviews[].state",
+    };
+  }
+  // The reducer-side path misses cases where GitHub's reviewDecision still
+  // reports CHANGES_REQUESTED but the latest-per-author reduction of fetched
+  // reviews carries none — e.g. the reviewer's CR is followed by COMMENTED
+  // (COMMENTED does not supersede CR in GitHub's decision logic), or the CR
+  // sits outside the `reviews(last: 30)` window. Either way the PR-level
+  // decision is the authoritative signal; fall back to it without asserting
+  // a specific cause.
```
lefarcen: **P2** This new fallback is only validated by live PR examples in the PR body, so the regression can silently return if those examples change or GitHub state moves on. Add a deterministic fixture/unit test for reviewDecision=CHANGES_REQUESTED with latest-per-author reviews containing no CHANGES_REQUESTED, plus one case proving the existing gh.latestReviews[].state path still wins.

tools/pr/src/tags.ts
```diff
@@ -116,12 +116,28 @@ function tagUnresolvedChangesRequested(facts: PrFacts): Tag | null {
...
-    source: "gh.latestReviews[].state",
-  };
+  if (reviewers.length > 0) {
+    return {
+      name: "unresolved-changes-requested",
+      reason: `latestReviews carries CHANGES_REQUESTED from: ${[...new Set(reviewers)].join(", ")}`,
+      source: "gh.latestReviews[].state",
+    };
+  }
+  // The reducer-side path misses cases where GitHub's reviewDecision still
```
lefarcen, line 126: **P3** The comment explains the two mechanical causes, but it omits the observed evidence and scale from the bug report: 3 PRs in the 102-PR patrol queue, specifically #1101/#1127/#1163. If this comment is meant to carry the reasoning context, include that evidence or keep it in adjacent docs/tests so future maintainers can distinguish a proven queue pattern from a theoretical edge case.

tools/pr/src/tags.ts
lefarcen: **P2** This changes the tag's effective rule/source from latestReviews-only to latestReviews OR gh.reviewDecision. The tools-pr tag dictionary documents tags as mechanical rules with data sources, so the companion documentation should be updated to avoid downstream users assuming unresolved-changes-requested only means a visible latest reviewer state.

lefarcen (commented):
The fallback is logically correct and cheap, and using GitHub's PR-level reviewDecision is the right authoritative signal for the observed gap. I would not block on code correctness, but the change lacks deterministic regression coverage and the rationale/comment/documented contract could be tightened.

tools/pr/src/tags.ts
PerishCode: Fixed in 7296b993: `tools/pr/AGENTS.md` §Tag dictionary now lists both paths on the `unresolved-changes-requested` row — primary per-reviewer `gh.latestReviews[].state` plus the `gh.reviewDecision` fallback when no per-reviewer CR survives the latest-per-author reduction. Source column carries both tokens so a report consumer can interpret either emission shape against the dictionary.

tools/pr/src/tags.ts
PerishCode: Fixed in 7296b993: same call-out as the parallel Siri-Ray thread on this file. The tag dictionary row now documents primary + fallback paths with both source tokens (`gh.latestReviews[].state · gh.reviewDecision (fallback)`).

PerishCode: Fixed in 5a62cded: added `tools/pr/tests/tags-unresolved-cr.test.ts` with 6 cases pinning both paths — primary (per-reviewer CR after reduction → `source: gh.latestReviews[].state`), fallback (no per-reviewer CR + `reviewDecision=CHANGES_REQUESTED` → `source: gh.reviewDecision`, covering both the COMMENTED-follow-up shape and the empty-reviews / out-of-`reviews(last: 30)`-window shape), primary-wins-over-fallback when both signals present, and two negatives (empty reviewDecision, APPROVED). 6/6 pass locally. First file under `tools/pr/tests/`; the package's `test` script already pointed at `tests/*.test.ts`, so no scaffolding changes were needed.

tools/pr/src/tags.ts
PerishCode: Fixed in 5a62cded: extended the fallback's code comment with the observed scale ("3 of 102 open PRs hit this gap: #1101 / #1127 / #1163, so this is a recurring pattern, not a theoretical edge case"). Adjacent test file `tools/pr/tests/tags-unresolved-cr.test.ts` also encodes the live-queue shapes that motivated the fallback (COMMENTED-follow-up + empty-reviews), so future maintainers can see the empirical evidence both in-code and in tests.

lefarcen (approved):
Both P2 findings from my 12:34Z review are resolved:

- ✅ Test coverage: `tools/pr/tests/tags-unresolved-cr.test.ts` added with 6 comprehensive cases (primary path, fallback path, precedence, negatives)
- ✅ Tag documentation: `tools/pr/AGENTS.md` updated to list both source tokens (`gh.latestReviews[].state · gh.reviewDecision (fallback)`)

The fallback is now deterministically tested and documented. LGTM 🎉

lefarcen (commented):
All three findings from my previous review are addressed on this head:

- Test coverage: Added `tools/pr/tests/tags-unresolved-cr.test.ts` with 6 cases pinning both emission paths (primary + fallback) and the priority logic when both qualify.
- Tag documentation: Updated `tools/pr/AGENTS.md` §Tag dictionary to list both source tokens and explain the fallback trigger.
- Code comment: Extended the inline comment above the fallback with the observed scale (3 of 102 open PRs) and explicit PR references (#1101 / #1127 / #1163), so future maintainers see the empirical motivation both in-code and in the adjacent test file.

LGTM. 🎉

PR #1293, opened 2026-05-11
fix: support object-style question-form options

Yiyan-Jiang:
Fixes #

# fix question-form dropping object-style options

## Summary

Fix `<question-form>` silently dropping object-style `options`.

`radio` / `checkbox` / `select` questions currently only preserve string arrays:

```ts
options: ['A', 'B']
````

AI-generated forms often emit richer structures:

```ts
options: [
  {
    label: 'Mobile',
    description: 'Phone-first',
    value: 'mobile'
  }
]
```

Those entries were filtered out during parsing, causing:

* controls not rendering
* required questions becoming impossible to complete
* submit button staying disabled

This PR normalizes both formats into a shared internal option structure and updates the form renderer to use `value` for state and `label` for display.

---

## Root cause

`apps/web/src/artifacts/question-form.ts` only preserved string options:

```ts
const options = Array.isArray(qo.options)
  ? qo.options.filter((o): o is string => typeof o === 'string')
  : undefined;
```

Object entries were discarded entirely, leaving `q.options = []`.

`QuestionForm.tsx` only renders choice controls when `q.options` exists, so the UI disappeared even though the question itself parsed successfully.

---

### Parser

Added a normalized internal option structure:

```ts
export interface FormOption {
  label: string;
  value: string;
  description?: string;
}
```

Updated `FormQuestion.options`:

```ts
options?: FormOption[];
```

Added option normalization logic:

* string -> `{ label, value }`
* object -> preserve `label`, `description`, `value`
* fallback `value = label`
* invalid entries skipped individually

This keeps legacy forms working while supporting richer AI-generated structures.

---

### Renderer

Updated `apps/web/src/components/QuestionForm.tsx` to:

* use `opt.value` as the internal answer state
* use `opt.label` for display
* support optional `description`
* keep checkbox/radio/select behavior consistent

Examples:

#### Radio

```tsx
checked={value === opt.value}
onChange={() => update(q.id, opt.value)}
```

#### Checkbox

```tsx
arr.includes(opt.value)
```

#### Select

```tsx
<option value={opt.value}>
  {opt.label}
</option>
```

---

### Answer formatting compatibility

`formatFormAnswers()` now maps stored values back to readable labels.

`parseSubmittedAnswers()` supports both:

* legacy label-based history
* new value-based submissions

This preserves compatibility with existing conversation history while allowing stable internal values.

---

## Tests

Added parser coverage for:

* legacy string options
* object-style options
* value fallback behavior
* invalid option filtering

Added component coverage for:

* radio / checkbox / select rendering with object options
* required-state recovery
* label/value compatibility during answer restore

---

## Manual verification

Tested with:

```json
{
  "label": "Primary surface",
  "type": "radio",
  "required": true,
  "options": [
    {
      "label": "Mobile (iOS/Android)",
      "description": "Phone-first",
      "value": "mobile"
    },
    {
      "label": "Desktop web",
      "description": "Desktop-first",
      "value": "desktop"
    }
  ]
}
```

Verified that:

* controls render correctly
* required questions can be completed
* submit button enables correctly
* answers restore correctly from history
* legacy string-array forms still work

---

## Non-goals

This PR intentionally does not change prompting behavior.

Prompt examples may be updated later to explicitly mention object-style options, but parser compatibility is treated as the primary fix.

---

## Why this approach

The fix is intentionally implemented at the parser boundary instead of relying on prompt compliance.

Models already emit object-style options in practice, so the renderer/parser pipeline should tolerate both formats rather than assuming a single strict shape.

This keeps the system backward-compatible while making `<question-form>` resilient to richer structured outputs.

```

### DIFF
--- a/apps/web/src/artifacts/question-form.ts
+++ b/apps/web/src/artifacts/question-form.ts
@@ -52,11 +52,17 @@ export interface DirectionCard {
   bodyFont: string;
 }
 
+export interface FormOption {
+  label: string;
+  value: string;
+  description?: string;
+}
+
 export interface FormQuestion {
   id: string;
   label: string;
   type: QuestionType;
-  options?: string[];
+  options?: FormOption[];
   placeholder?: string;
   required?: boolean;
   help?: string;
@@ -159,9 +165,7 @@ function tryParseForm(body: string, attrs: Record<string, string>): QuestionForm
         : `q${i + 1}`;
     const label = typeof qo.label === 'string' ? qo.label : id;
     const type = normalizeType(qo.type);
-    const options = Array.isArray(qo.options)
-      ? qo.options.filter((o): o is string => typeof o === 'string')
-      : undefined;
+    const options = parseOptions(qo.options);
     const placeholder = typeof qo.placeholder === 'string' ? qo.placeholder : undefined;
     const help = typeof qo.help === 'string' ? qo.help : undefined;
     const required = qo.required === true;
@@ -172,14 +176,7 @@ function tryParseForm(body: string, attrs: Record<string, string>): QuestionForm
         ? qo.maxSelections
         : undefined;
     const cards = parseDirectionCards(qo.cards);
-    const defaultValue =
-      typeof qo.defaultValue === 'string'
-        ? qo.defaultValue
-        : Array.isArray(qo.defaultValue)
-          ? qo.defaultValue.filter((v): v is string => typeof v === 'string')
-          : typeof qo.default === 'string'
-            ? qo.default
-            : undefined;
+    const defaultValue = parseDefaultValue(qo, options);
     questions.push({
       id,
       label,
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by Yiyan-Jiang** [comment]
Closes #1151

File: apps/web/src/artifacts/question-form.ts [hunk 1]
```diff
@@ -266,10 +314,31 @@ export function formatFormAnswers(
   for (const q of form.questions) {
     const v = answers[q.id];
     let display: string;
-    if (Array.isArray(v)) display = v.length > 0 ? v.join(', ') : '(skipped)';
-    else if (typeof v === 'string') display = v.trim().length > 0 ? v.trim() : '(skipped)';
+    if (Array.isArray(v)) {
+      display = v.length > 0 ? v.map((value) => formOptionLabelForValue(q, value)).join(', ') : '(skipped)';
```
lefarcen: P2 — `formatFormAnswers` converts stored option values back to labels before sending the user message, and the upstream `onSubmitForm` path only persists that text, so a model-provided `{ label: "Mobile (iOS/Android)", value: "mobile" }` loses its `value` at the actual agent boundary. That means object options with stable values render/select correctly in React state but the next agent turn only receives the readable label. Preserve the stable value in the formatted answer (ideally with the readable label alongside it for human readers), and update `parseSubmittedAnswers` plus a regression test for `value !== label`.

apps/web/tests/components/QuestionForm.test.tsx
```diff
@@ -43,4 +73,21 @@ describe('QuestionFormView', () => {
     expect(screen.getByText('answered')).toBeTruthy();
     expect(container.querySelectorAll('input[type="checkbox"]:checked')).toHaveLength(2);
   });
+
+  it('renders object options and submits the readable label', () => {
```
lefarcen: P3 — The component regression only exercises radio object options, while the production change also rewrites checkbox and select branches. The existing checkbox test only checks locked submitted history and does not click object options or verify submit enabling/formatting, and there is no select coverage. Add interactive tests for required checkbox and select questions with object options, including a `value !== label` case, so the branches changed in `QuestionForm.tsx` are actually guarded.

lefarcen (changes requested):
Hey @Yiyan-Jiang, thanks for tackling this! The parser + renderer changes look solid for fixing the immediate UI rendering issue, but there's a gap in the submission path that undermines the new value/label API — see inline.

apps/web/tests/components/QuestionForm.test.tsx
Yiyan-Jiang: Good catch — you're right that the previous implementation still collapsed stable option values back into readable labels at the actual agent boundary.

I've updated the submission serialization path to preserve stable values alongside readable labels using the format:

`Label [value: stable-value]`

and added backward-compatible parsing for both the old label-only format and the new value-preserving format.

I also expanded the regression coverage to include interactive radio / checkbox / select object-option flows where `value !== label`, including submit enablement and round-trip parsing coverage.

lefarcen (approved):
Prior P2/P3 issues addressed. The new submission format preserves stable values via `Label [value: ...]` and the test coverage is complete. LGTM.

leilei926524-tech:
Hi @Yiyan-Jiang — welcome to Open Design contributors. 🎉

Your first merged PR (#1293) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #1294, opened 2026-05-11
prevent duplicate saves and add template deletion

Leesin0222:
Fixes #1237 , Fixes #1252

  ## Summary
  - **Duplicate save prevention (#1237)**: When saving a template with the same `name`+`source_project_id` that already exists, the existing record is updated instead of creating a duplicate entry
    - Added `findTemplateByNameAndProject()` lookup in the daemon DB layer
    - Added `updateTemplate()` to overwrite description, files snapshot, and timestamp
    - POST `/api/templates` handler now checks for existing match before insert
  - **Template deletion UI (#1252)**: Added a delete button to the saved template list
   in the template picker
    - Each template shows a ✕ button on hover, calling the existing `DELETE /api/templates/:id` endpoint
    - Template list refreshes automatically after deletion
    - If the currently selected template is deleted, selection resets to none

  ## Validation
  - Save a file as template with name "test-template"
  - Save the same file again with the same name → template list shows 1 entry
  (updated), not 2
  - Save with a different name → new entry created as expected
  - Hover over a saved template → x delete button appears
  - Click x → template is removed from the list and does not reappear after refresh

### DIFF
--- a/apps/daemon/src/db.ts
+++ b/apps/daemon/src/db.ts
@@ -609,6 +609,22 @@ export function getTemplate(db: SqliteDb, id: string) {
   return row ? normalizeTemplate(row) : null;
 }
 
+export function findTemplateByNameAndProject(
+  db: SqliteDb,
+  name: string,
+  sourceProjectId: string,
+) {
+  const row = db
+    .prepare(
+      `SELECT id, name, description, source_project_id AS sourceProjectId,
+              files_json AS filesJson, created_at AS createdAt
+         FROM templates
+        WHERE name = ? AND source_project_id = ?`,
+    )
+    .get(name, sourceProjectId) as DbRow | undefined;
+  return row ? normalizeTemplate(row) : null;
+}
+
 export function insertTemplate(db: SqliteDb, t: DbRow) {
   db.prepare(
     `INSERT INTO templates (id, name, description, source_project_id, files_json, created_at)
@@ -622,6 +638,17 @@ export function insertTemplate(db: SqliteDb, t: DbRow) {
     t.createdAt,
   );
   return getTemplate(db, t.id);
+}
+
+export function updateTemplate(
+  db: SqliteDb,
+  id: string,
+  t: { description: string | null; files: unknown[]; createdAt: number },
+) {
+  db.prepare(
+    `UPDATE templates SET description = ?, files_json = ?, created_at = ? WHERE id = ?`,
+  ).run(t.description, JSON.stringify(t.files), t.createdAt, id);
+  return getTemplate(db, id);
 }
 
 export function deleteTemplate(db: SqliteDb, id: string) {
--- a/apps/daemon/src/db.ts
+++ b/apps/daemon/src/db.ts
@@ -643,11 +643,11 @@ export function insertTemplate(db: SqliteDb, t: DbRow) {
 export function updateTemplate(
   db: SqliteDb,
   id: string,
-  t: { description: string | null; files: unknown[]; createdAt: number },
+  t: { description: string | null; files: unknown[] },
 ) {
   db.prepare(
-    `UPDATE templates SET description = ?, files_json = ?, created_at = ? WHERE id = ?`,
-  ).run(t.description, JSON.stringify(t.files), t.createdAt, id);
+    `UPDATE templates SET description = ?, files_json = ? WHERE id = ?`,
+  ).run(t.description, JSON.stringify(t.files), id);
   return getTemplate(db, id);
 }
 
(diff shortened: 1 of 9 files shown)


lefarcen (commented):
Thanks for tackling #1237 and #1252! The deduplication approach (find-then-update) is sound and the UI delete button is a clean addition. A few notes on the implementation:

**Main issue:**
- The update path changes `createdAt` to now, which breaks the semantics (see inline)

**Plus sides:**
- Parameterized SQL queries prevent injection ✅
- UI layout (hover-revealed delete button) is thoughtful
- Test stubs updated (though not actually testing delete flow — see inline)

Overall structure is good. Let me know if the inline suggestions make sense.

apps/daemon/src/db.ts
```diff
@@ -624,6 +640,17 @@ export function insertTemplate(db: SqliteDb, t: DbRow) {
...
 }
 
+export function updateTemplate(
+  db: SqliteDb,
+  id: string,
+  t: { description: string | null; files: unknown[]; createdAt: number },
+) {
+  db.prepare(
+    `UPDATE templates SET description = ?, files_json = ?, created_at = ? WHERE id = ?`,
+  ).run(t.description, JSON.stringify(t.files), t.createdAt, id);
```
lefarcen: P2 — `createdAt` should not change on update. When you update an existing template, `createdAt` should stay at its original timestamp (the first time the template was created). Consider:
```typescript
export function updateTemplate(
  db: SqliteDb,
  id: string,
  t: { description: string | null; files: unknown[] },
) {
  db.prepare(
    `UPDATE templates SET description = ?, files_json = ? WHERE id = ?`,
  ).run(t.description, JSON.stringify(t.files), id);
  return getTemplate(db, id);
}
```
or add `updatedAt` column if you want to track last-modified time separately.

apps/daemon/src/project-routes.ts
```diff
@@ -533,14 +533,26 @@ export function registerProjectRoutes(app: Express, ctx: RegisterProjectRoutesDe
...
       }
-      const t = insertTemplate(db, {
-        id: randomId(),
-        name: name.trim(),
-        description: typeof description === 'string' ? description : null,
-        sourceProjectId,
-        files: snapshot,
-        createdAt: Date.now(),
-      });
+      const trimmedName = name.trim();
```
lefarcen, line 539: P3 — Consider adding a length check on `name` (e.g. max 100 chars) before calling `.trim()` to prevent accidentally huge template names from being stored. Low priority since this is internal API, but good defensive coding.

apps/web/tests/components/NewProjectPanel.test.tsx
```diff
@@ -90,6 +90,7 @@ describe('NewProjectPanel design system defaults', () => {
         designSystems={designSystems}
         defaultDesignSystemId="clay"
         templates={[]}
```
lefarcen, line 92: P3 — The test stubs were updated with `onDeleteTemplate={vi.fn()}`, but there's no actual test for the delete flow (user clicks delete → callback fires → template removed from list). Consider adding a test:
```typescript
it('calls onDeleteTemplate when user clicks delete button', async () => {
  const onDelete = vi.fn().mockResolvedValue(true);
  render(<NewProjectPanel templates={[...]} onDeleteTemplate={onDelete} ... />);
  const deleteBtn = screen.getByLabelText(/delete template/i);
  await user.click(deleteBtn);
  expect(onDelete).toHaveBeenCalledWith(templateId);
});
```
Non-blocking for this PR, but useful coverage for the new UI.

Leesin0222: added a test in `NewProjectPanel.test.tsx` that verifies clicking the delete button calls`onDeleteTemplate` with the correct template ID. See the new `describe('NewProjectPanel template deletion')` block :)

apps/daemon/src/project-routes.ts
Leesin0222: Agreed. added a 100-character length check on `name` in the POST `/api/templates` handler. Returns 400 if exceeded.

apps/daemon/src/db.ts
Leesin0222: good..! removed `createdAt` from the update path so the original creation timestamp is preserved. `updateTemplate` now only updates `description` and `files_json`.

lefarcen (commented):
All feedback addressed — thank you for the quick iteration! 🎉

✅ P2 `createdAt` semantics fixed — original timestamp now preserved on update
✅ P3 Name length check added (100 char limit before trim)
✅ P3 Test coverage added — new `describe('NewProjectPanel template deletion')` block

Implementation looks solid. The deduplication logic (find-then-update) and the UI delete flow are both clean. Ready for maintainer review.

lefarcen:
@mrcfps This PR's current head 54e1a888 is waiting on your review.

Siri-Ray (approved):
@Leesin0222 I reviewed the current changed ranges for the duplicate template save path, the template deletion UI wiring, and the added component coverage. The latest head preserves the original template creation timestamp on updates, refreshes the list after deletion, clears the selected template when needed, and the prior feedback appears addressed. Thanks for the thoughtful iteration here; this is a clean improvement to the template flow.


leilei926524-tech:
Hi @Leesin0222 — quick note from the Open Design team.

What you get there:

Issue #1194, opened 2026-05-11
render ask-question block render error


kunfan96:
<img width="369" height="193" alt="Image" src="https://github.com/user-attachments/assets/adea90a3-e64e-495c-8bd8-153d9edcf82e" />

lefarcen:
Thanks for the screenshot. This looks related to #312 by symptom, but appears to be a different block shape: the current UI only has a renderer/parser for `<question-form>...</question-form>`, and I could not find any `ask-question` / `blockType` renderer in the codebase.

Could you share:
- App version
- Provider/model used
- The prompt that generated this
- Whether the same raw `<blockType>:"ask-question"` text is still visible after reloading the project?

This will help narrow down whether it's a new schema leak from a specific runtime/provider or a model prompt drift.

lefarcen:
This is open for community help.

If you'd like to work on it, comment `/claim` below. The first `/claim` after the latest release gets it. Any comment, linked PR, or progress note keeps the claim active.

Eggwardhan:
/claim

Engaging with a tactical defensive fix: extend the existing `<question-form>` parser to also accept `<ask-question>` as a tag-name alias, so a model that emits `<ask-question>...</ask-question>` with a valid JSON body renders the form correctly instead of leaking raw markup into prose. This is the lowest-risk shape of the bug and matches @lefarcen's note that no `ask-question` renderer exists in the codebase today.

Open question that depends on the reporter (@Morzorz / whoever filed): if the actual emission is a JSON-only `{"blockType":"ask-question", ...}` body without any tag wrapper, the alias alone won't help — that path needs its own discriminator-based detection, which I'd rather hold until the model + prompt are confirmed so we don't over-fit the parser to a hallucination.

Will keep the PR scoped to the alias + a regression test; happy to add the blockType-detection path in a follow-up once we know the exact shape.

lefarcen:
Claimed by @Eggwardhan. Thanks for picking this up! 🎉

**Quick question to time check-ins right** — roughly how long do you think this will take?

- A few hours / today
- 1-3 days
- ~ 1 week
- Longer / unsure

Reply with whatever fits. If you don't reply, I'll default to checking in every ~12h. Either way, any comment, linked PR, or progress note keeps the claim active.

lefarcen:
[...]

PR #1295, opened 2026-05-11
fix(web): accept <ask-question> as alias for <question-form>

Eggwardhan:
Refs #1194.

## Symptom

The agent occasionally emits `<ask-question>...</ask-question>` instead of the canonical `<question-form>...</question-form>` tag. The existing parser in `apps/web/src/artifacts/question-form.ts` only recognizes the canonical tag, so the alias variant falls through as raw markup and renders broken text in the chat pane.

The issue's screenshot shows raw `<blockType>:"ask-question"` text leaking into the UI; `@lefarcen`'s triage on #1194 confirmed no `ask-question` / `blockType` renderer exists in the codebase today.

## Fix

Make `<ask-question>` a tag-name alias for `<question-form>` in the parser:

- `OPEN_RE` now captures the tag name (`question-form` | `ask-question`) in addition to the attribute string.
- The close-tag string is computed from the captured tag name per match, so an open `<question-form>` cannot be terminated by `</ask-question>` and vice-versa.
- Close-tag lookup is case-insensitive so `<Ask-Question>...</Ask-Question>` still parses.

Everything downstream (`tryParseForm`, attribute parsing, JSON body validation, `direction-cards`, the `FormSegment` shape, `AssistantMessage` consumption) is unchanged — a parsed `ask-question` produces the same `QuestionForm` object as a `question-form` would.

## Existing behavior preserved

- Canonical `<question-form>` tag still parses identically (regression test included).
- Malformed JSON body still falls back to raw prose.
- Unterminated tag still leaves trailing text intact rather than swallowing it.
- Mismatched open/close (e.g. `<question-form>...</ask-question>`) does **not** silently merge into a block — the entire span stays as prose.

## Tests

New `apps/web/tests/artifacts/question-form.test.ts` (6 cases, all passing locally):

- canonical `<question-form>` parses
- alias `<ask-question>` parses
- mixed casing on alias (e.g. `<Ask-Question>`) parses
- mismatched open/close stays prose
- malformed JSON body stays prose
- unterminated tag preserves trailing text

## Scope notes (deliberately out of scope)

If the actual model emission turns out to be a **JSON-only block without any tag wrapper** — e.g. literally serialized as `{\"blockType\":\"ask-question\", ...}` in prose — the alias alone won't help. That would need its own discriminator-based detection path which I'd rather hold until the reporter confirms the exact model + prompt + version (which `@lefarcen` already asked for on the issue). I'm happy to layer that on as a follow-up once the shape is known; speculating on the exact JSON envelope right now risks over-fitting the parser to a hallucination.

- [ ] `pnpm --filter @open-design/web test tests/artifacts/question-form.test.ts` green
- [ ] Manual repro from the reporter's environment (`@Morzorz` if you're reading: if you can paste the exact final assistant output that leaks `<blockType>:\"ask-question\"`, that pins down whether this PR fixes your case or whether the tagless JSON path is needed instead)

Refs #1194.

### DIFF
--- a/apps/web/src/artifacts/question-form.ts
+++ b/apps/web/src/artifacts/question-form.ts
@@ -17,6 +17,11 @@
  *   }
  *   </question-form>
  *
+ * `<ask-question>...</ask-question>` is accepted as an alias for
+ * `<question-form>`, so a model that drifts to the colloquial tag
+ * name still renders correctly instead of leaking raw markup into
+ * prose (see issue #1194).
+ *
  * Splits a final assistant text payload into ordered segments — prose +
  * forms — so AssistantMessage can render the form inline.
  */
@@ -85,15 +90,19 @@ export type FormSegment =
   | { kind: 'text'; text: string }
   | { kind: 'form'; form: QuestionForm; raw: string };
 
-const OPEN_RE = /<question-form\b([^>]*)>/i;
-const CLOSE_TAG = '</question-form>';
+// `question-form` is the canonical tag; `ask-question` is an alias the
+// model occasionally drifts to (issue #1194). The close tag must match
+// the open tag name, so each match captures the name and computes its
+// own close-tag string. Treat the lookup case-insensitively at scan
+// time so `<Question-Form>` and `<ASK-QUESTION>` still parse.
+const OPEN_RE = /<(question-form|ask-question)\b([^>]*)>/i;
 
 export function splitOnQuestionForms(input: string): FormSegment[] {
   const out: FormSegment[] = [];
   let cursor = 0;
-  // Scan repeatedly for <question-form> opens; for each, locate the
-  // matching close tag and try to parse the JSON body. Anything that
-  // doesn't parse cleanly stays in the prose stream.
+  // Scan repeatedly for question-form / ask-question opens; for each,
+  // locate the matching close tag and try to parse the JSON body.
+  // Anything that doesn't parse cleanly stays in the prose stream.
   while (cursor < input.length) {
     const slice = input.slice(cursor);
     const m = OPEN_RE.exec(slice);
@@ -101,9 +110,13 @@ export function splitOnQuestionForms(input: string): FormSegment[] {
       out.push({ kind: 'text', text: slice });
       break;
     }
+    const tagName = (m[1] ?? 'question-form').toLowerCase();
+    const closeTag = `</${tagName}>`;
     const openStart = cursor + m.index;
     const openEnd = openStart + m[0].length;
-    const closeIdx = input.indexOf(CLOSE_TAG, openEnd);
+    // Find the matching close tag case-insensitively so a drifted
+    // `</Ask-Question>` still terminates the block.
+    const closeIdx = findCloseTag(input, openEnd, closeTag);
     if (closeIdx === -1) {
       // Unterminated — leave the rest as prose so we don't swallow it.
       out.push({ kind: 'text', text: slice });
@@ -113,17 +126,23 @@ export function splitOnQuestionForms(input: string): FormSegment[] {
       out.push({ kind: 'text', text: input.slice(cursor, openStart) });
     }
     const body = input.slice(openEnd, closeIdx);
-    const attrs = parseAttrs(m[1] ?? '');
+    const attrs = parseAttrs(m[2] ?? '');
     const form = tryParseForm(body, attrs);
+    const blockEnd = closeIdx + closeTag.length;
     if (form) {
-      out.push({ kind: 'form', form, raw: input.slice(openStart, closeIdx + CLOSE_TAG.length) });
+      out.push({ kind: 'form', form, raw: input.slice(openStart, blockEnd) });
     } else {
       // Malformed — keep raw text so the user can still see it.
-      out.push({ kind: 'text', text: input.slice(openStart, closeIdx + CLOSE_TAG.length) });
+      out.push({ kind: 'text', text: input.slice(openStart, blockEnd) });
     }
-    cursor = closeIdx + CLOSE_TAG.length;
+    cursor = blockEnd;
   }
   return out;
+}
+
+function findCloseTag(input: string, from: number, closeTag: string): number {
+  const haystack = input.toLowerCase();
+  return haystack.indexOf(closeTag, from);
 }
 
 function parseAttrs(raw: string): Record<string, string> {
--- a/apps/web/tests/artifacts/question-form.test.ts
+++ b/apps/web/tests/artifacts/question-form.test.ts
@@ -2,6 +2,13 @@ import { describe, expect, it } from 'vitest';
 
 import { formatFormAnswers, splitOnQuestionForms } from '../../src/artifacts/question-form';
 
+const VALID_BODY = `{
+  "questions": [
+    { "id": "platform", "label": "Platform", "type": "radio",
+      "options": ["Mobile", "Desktop", "Responsive"], "required": true }
+  ]
+}`;
+
 describe('splitOnQuestionForms', () => {
   it('normalizes string and object question options', () => {
     const input = [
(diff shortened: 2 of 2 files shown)


apps/web/src/artifacts/question-form.ts
```diff
@@ -107,19 +120,25 @@ export function splitOnQuestionForms(input: string): FormSegment[] {
       out.push({ kind: 'text', text: input.slice(cursor, openStart) });
     }
```
lefarcen, line 127: P3 — Minor performance note: `findCloseTag` lowercases the entire input on every form block. For extreme cases (100+ forms in one message), consider caching the lowercased string at the top of `splitOnQuestionForms`. Not blocking for typical use (1-5 forms per message).

lefarcen (commented):
Hey @Eggwardhan! Clean work — the alias approach is minimal, well-tested, and doesn't over-fit to an uncertain model emission pattern. ✨

Lens A (code correctness): regex capture group shift handled correctly, close-tag matching logic is sound, edge cases covered in tests. Performance note below is non-blocking.

Lens B (reasoning): scope notes acknowledge the limitation (JSON-only emission without tags would need separate handling), verification section asks for reporter confirmation. 👍

lefarcen:
@PerishCode This PR's current head (51834acc) is waiting on your review.

PerishCode:
@Eggwardhan I'm holding off on generating review comments for nexu-io/open-design#1295 because this pull request has merge conflicts right now.

Eggwardhan:
Thanks @PerishCode — rebased onto latest `main` ([fe75b9bb](https://github.com/nexu-io/open-design/pull/1295/commits/fe75b9bb)).

The conflict was in `apps/web/tests/artifacts/question-form.test.ts`: `main` added a new `formatFormAnswers` + object-options test pair (#1265 line of work), and this PR's new file collided on the import line. Merged both sets under a single `describe('splitOnQuestionForms')` block — the upstream import now pulls both helpers, the new `VALID_BODY` constant + 6 alias/edge-case tests follow the existing ones. No behavior or assertion change on either side.

Mergeable should now go green; ready for your review.

lefarcen (commented):
Hey @Eggwardhan, the rebase note is clear — thanks for spelling out how the upstream `formatFormAnswers` tests were preserved alongside the alias coverage. One quick PR-body cleanup before this goes back through pool review: could you update the description to the current template shape with `Why`, `What users will see`, `Surface area`, `Bug fix verification`, and `Validation`? For this PR, the existing Symptom/Fix/Test details should map cleanly: `Why` can point at #1194's raw tag leak, `What users will see` can say the `<ask-question>` form renders instead of raw markup, and Validation/Bug fix verification can name the targeted test + CI run.

PerishCode (commented):
@Eggwardhan — thanks for the rebase and the clear scope note. Focused this pass on the new `findCloseTag` scan path.

One blocking correctness regression flagged inline on `apps/web/src/artifacts/question-form.ts` lines 143–146 — `input.toLowerCase()` can change string length under Unicode default case folding (`İ` → `i\u0307`), which desyncs `closeIdx` from positions in the original `input` and breaks even the canonical `<question-form>` path when certain non-ASCII characters appear in surrounding prose. Reproduces deterministically; concrete fix proposed inline.

This is the same root cause the Codex bot raised at https://github.com/nexu-io/open-design/pull/1295#discussion_r3219069506 — re-raising as blocking with a reproduction + fix so it doesn't get lost in the bot noise.




apps/web/src/artifacts/question-form.ts
```diff
@@ -113,19 +126,25 @@ export function splitOnQuestionForms(input: string): FormSegment[] {
...
-    cursor = closeIdx + CLOSE_TAG.length;
+    cursor = blockEnd;
   }
   return out;
 }
 
+function findCloseTag(input: string, from: number, closeTag: string): number {
+  const haystack = input.toLowerCase();
+  return haystack.indexOf(closeTag, from);
```
PerishCode, line 145: **Blocking — `toLowerCase()` desync corrupts `closeIdx`, regressing canonical `<question-form>` too**

`findCloseTag` lowercases the entire `input` and then uses an index from that lowercased copy to slice the *original* string. Under Unicode default case folding, `toLowerCase()` can change string length — most notably `'İ'.length === 1` but `'İ'.toLowerCase()` is `'i\u0307'` (length 2). Any folding-expanding code point that appears between the start of `input` and the close tag shifts `haystack`'s indices by the expansion amount, so `closeIdx` (a position in `haystack`) no longer points at `</tag>` in `input`.

Result: `body = input.slice(openEnd, closeIdx)` overshoots and pulls one or more characters out of `</tag>` into the JSON body. `JSON.parse` then fails and the entire block silently degrades to raw text — exactly the symptom this PR is trying to fix.

Deterministic reproduction (Node REPL):

```js
const input = 'İ<ask-question>{}</ask-question>'; // length 32
const haystack = input.toLowerCase();              // length 33 (İ → i + U+0307)
input.indexOf('</ask-question>');                  // 17  (real position)
haystack.indexOf('</ask-question>', 15);           // 18  (shifted by +1)
input.slice(15, 18);                               // "{}<"   ← body now contains "<" from "</tag>"
JSON.parse('{}<');                                 // throws → block falls back to raw text
```

This is also a **regression of the canonical `<question-form>` path**: prior to this PR, `indexOf` ran directly against `input`, so a user with prose like `İdeal flow` immediately before a `<question-form>` got it rendered fine. After this PR they don't. Trigger is narrow (specific code points) but realistic for non-English users, and the failure mode is exactly the user-visible breakage this PR was opened to prevent.

**Suggested fix** — drop the lowercased haystack and search the original string with a case-insensitive regex, so indices stay aligned regardless of how case folding would have expanded characters:

```ts
function findCloseTag(input: string, from: number, closeTag: string): number {
  // Case-insensitive regex against the original string so indices stay
  // aligned with `input` under Unicode case folding (e.g. İ → i\u0307),
  // which can change string length and would otherwise desync
  // `closeIdx` from positions in `input`. `closeTag` is always
  // `</question-form>` or `</ask-question>` — no regex metacharacters
  // to escape.
  const re = new RegExp(closeTag, 'i');
  const m = re.exec(input.slice(from));
  return m ? from + m.index : -1;
}
```

Worth pinning a fixture-style test pair while you're in there — one with a folding-expanding char (e.g. `İ`) in prose immediately before a canonical `<question-form>`, one with the same before an `<ask-question>` alias — so this regression can't quietly come back.


lefarcen:
Hey @Eggwardhan — the rebase landed cleanly and CI is all-green on the current head ✅. There's one blocking correctness issue flagged by @PerishCode that needs a fix before this can move forward.

**Unicode case-folding desync in `findCloseTag`** ([line 145](https://github.com/nexu-io/open-design/pull/1295/files#diff-a0527a9c29a42be37fb09c72c25ca53b6cdab88c4d7e5e0e37ea6e90bc1d3e29R145)):

`findCloseTag` lowercases the entire `input` into `haystack`, then returns an index into that lowercased copy to the caller — which uses it to slice the *original* `input`. The bug: `String.prototype.toLowerCase()` can change string length for certain Unicode characters. For example, `'İ'.toLowerCase()` produces `'i̇'` — two code points instead of one. Any such character that appears before the close tag makes `haystack` longer than `input`, so `closeIdx` (an offset into `haystack`) no longer points at `</tag>` in the original string. The `input.slice(openEnd, closeIdx)` call then pulls part of the closing tag into the JSON body, `JSON.parse` throws, and the entire block silently falls back to raw text — the exact symptom this PR was opened to fix.

This also **regresses the existing `<question-form>` path**: prose like `İdeal flow` immediately before a canonical form block was fine before this PR. After this PR, it renders as raw text for those users.

**Suggested fix** (from @PerishCode's inline comment):

```ts
function findCloseTag(input: string, from: number, closeTag: string): number {
  // Case-insensitive regex against the original string keeps indices
  // aligned even when case folding would expand the code-point count.
  const re = new RegExp(closeTag, 'i');
  const m = re.exec(input.slice(from));
  return m ? from + m.index : -1;
}
```

`closeTag` is always `</question-form>` or `</ask-question>` — no regex metacharacters — so the `RegExp` constructor is safe. Worth adding a fixture test with a folding-expanding character (e.g. `İ`) in the prose just before each tag type so this can't quietly regress.

Once that's addressed, this looks ready to go!

PR #1297, opened 2026-05-11
feat(i18n): Phase 1 README locale drift detection

### LINKED ISSUE #195: RFC: TRANSLATIONS.md — i18n contribution guide and starter glossary
## Context

Following @lefarcen's review feedback on PR #194 (zh-TW README), this is an RFC to gauge interest in a `TRANSLATIONS.md` to capture the implicit conventions that have grown around i18n contributions.

The repo currently ships **7 locales** across the UI dictionary and READMEs (en, de, ja, ko, pt-BR, fa, ru, es, zh-CN, zh-TW), with new ones landing roughly every 2-4 weeks. Each contributor has had to reverse-engineer:

- which files to touch (`src/i18n/types.ts`, `src/i18n/index.tsx`, `src/i18n/locales/<code>.ts`, `README.<code>.md`)
- how to update the language switcher in **all** existing README variants atomically
- regional terminology choices (most relevant for zh-CN vs zh-TW)
- whether stale strings get backported when English changes

A short, opinionated guide could shave hours off each new locale PR.

## Proposal

Add `TRANSLATIONS.md` at the repo root, framed as a **living document** that the maintainers refine over time. Not a spec, not a gate.

### Draft skeleton

```markdown
# Translations

> **Status: living document.** Maintainers refine this as the project's i18n
> needs evolve. Contributions welcome.

## Maintained locales

| Code   | Language            | UI dict                            | README              |
|--------|---------------------|------------------------------------|---------------------|
| en     | English             | src/i18n/locales/en.ts             | README.md           |
| de     | Deutsch             | src/i18n/locales/de.ts             | README.de.md        |
| es     | Español             | src/i18n/locales/es.ts             | —                   |
| fa     | فارسی               | src/i18n/locales/fa.ts             | —                   |
| ja     | 日本語              | src/i18n/locales/ja-JP.ts          | README.ja-JP.md     |
| ko     | 한국어              | src/i18n/locales/ko.ts             | README.ko.md        |
| pt-BR  | Português (Brasil)  | src/i18n/locales/pt-BR.ts          | —                   |
| ru     | Русский             | src/i18n/locales/ru.ts             | —                   |
| zh-CN  | 简体中文            | src/i18n/locales/zh-CN.ts          | README.zh-CN.md     |
| zh-TW  | 繁體中文            | src/i18n/locales/zh-TW.ts          | README.zh-TW.md     |

> The English locale is the source of truth. Other locales fall back to English
> per-key when a translation is missing.

## Adding a new locale

1. Pick a BCP-47 code (e.g. `fr`, `it-IT`, `vi`).
2. Update `src/i18n/types.ts`:
   - extend the `Locale` union
   - append to `LOCALES`
   - add a `LOCALE_LABEL[<code>]` entry (use the language's native name)
3. Create `src/i18n/locales/<code>.ts` — copy from `en.ts`, translate values.
4. Register in `src/i18n/index.tsx`'s `DICTS` map.
5. (Optional) Add `README.<code>.md` translated from `README.md`.
6. Update the language switcher line (line 27 of each `README*.md`) in **every**
   existing README to add a link to your new file. This prevents transient
   broken links across locales.

## Backport policy

When the English README/UI gains new sections, contributors are not required to
backport — the affected locales fall back to English. Locale maintainers
(volunteer) are encouraged to refresh their translation in a follow-up PR. Keep
backport PRs focused: one locale per PR, no mixed feature work.

## Regional terminology

Translations follow the conventions of the target region's tech writing
community. The maintainer team trusts contributors to make idiomatic choices.

### zh-CN ↔ zh-TW starter glossary

When converting between Simplified and Traditional Chinese, prefer Taiwan-specific
phrasing in zh-TW rather than character-only conversion:

| en          | zh-CN    | zh-TW       |
|-------------|----------|-------------|
| screen      | 屏幕     | 螢幕        |
| stack       | 栈       | 堆疊        |
| project     | 项目     | 專案        |
| software    | 软件     | 軟體        |
| video       | 视频     | 影片        |
| file        | 文件     | 檔案        |
| document    | 文档     | 文件        |
| message     | 信息     | 訊息        |
| network     | 网络     | 網路        |
| database    | 数据库   | 資料庫      |
| user        | 用户     | 使用者      |
| default     | 默认     | 預設        |
| real-time   | 实时     | 即時        |
| install     | 安装     | 安裝        |
| settings    | 设置     | 設定        |
| menu        | 菜单     | 選單        |
| compatible  | 兼容     | 相容        |
| fallback    | 兜底     | 備援        |
| go viral    | 出圈     | 爆紅        |
| bind        | 绑定     | 綁定        |
| desktop end | 桌面端   | 桌面版      |
| mobile end  | 移动端   | 行動版      |

Tooling: OpenCC `s2twp.json` config handles ~80% automatically; the rest needs
human review.

## Native-speaker review

Strongly preferred but not blocking. Maintainers will merge with a `nit` label
if they can't read the language and no native speaker has reviewed within
~7 days.

## Open questions

These are intentionally undecided — flagged for maintainer input:

- Should we adopt a translation memory tool (Crowdin / Weblate / Lingui)?
- Do we ever drop a stale locale (no contributor for 6+ months)?
- Is the `pt-BR` / `pt` (vs `pt-PT`) split a precedent for `en` / `en-GB`?
- Should `LOCALE_LABEL` use ISO codes or native names by convention?
```

## Why an issue, not a PR

- Glossary tables, backport policy, and the \"open questions\" section reflect
  decisions the maintainers should own. Posting as a draft RFC lets the team
  shape the framing before any file lands in main.
- I can turn this into a PR (matching whatever scope/tone you prefer) once the
  direction is agreed — or you may prefer to fold pieces into `CONTRIBUTING.md`
  instead of a new file.

Happy to iterate. Tagging @lefarcen since this came out of #194 review.

laihenyi:
## Summary

- Add `<!-- l10n-source-commit: <SHA> -->` annotation to all 12 locale README files, pointing to current HEAD (`87a95b7fb4eeb0a5c954dbe0dea95b3fa5c0dc46`)
- Add `.github/workflows/readme-drift.yml` with two jobs:
  - **pr-comment**: On any PR touching `README.md`, diff each locale's annotated SHA vs HEAD and post a PR comment listing which locales are behind (passive signal, not a blocker)
  - **track**: Weekly cron that opens/updates a single rolling tracking issue with a per-locale checklist

## What this PR does NOT do

- No auto-translation or MT push — Phase 2 only revisited if Phase 1 data shows drift outpaces human capacity
- No merge gate — all signals are passive (PR comment + weekly issue)
- No platform/SaaS dependency — pure GH Action, ~200 lines of bash + YAML

## Decision basis

RFC discussion in issue #195:
- lefarcen: "Strong case for Phase 1. 💡", "The full three-part stack is small enough <200 lines", "drift velocity outpaces human capacity" as Phase 2 trigger
- MDN-style annotation confirmed as lowest-friction fit for open-design's shape (12 flat README files, no dedicated translation teams)
- Phase 2 trigger signals agreed: weekly tracking issue stays >50% incomplete for ≥4 weeks, ≥5 locales behind by ≥3 H2 sections, or locale maintainer requests MT drafts

## Test plan

- [ ] CI passes (workflow YAML validated)
- [ ] Workflow file has correct triggers: `pull_request` (paths: `README.md`) + `schedule` (weekly cron)
- [ ] All 12 locale READMEs have `l10n-source-commit` annotation
- [ ] `locale-drift` label exists on repo (created by first cron run) or can be created
- [ ] Manual test: push a new commit to a test branch that adds an H2 section to `README.md`, open a PR — verify comment is posted

## Adjacent issues

- **Prompt count fix** (README.md 93→94 / 43→44 correction + propagation to all locale READMEs) — separate PR
- **zh-CN mistranslation sweep** (`prompt → 產物路徑` → `prompt → 輸出行為`) — separate PR

Fixes #195.

### DIFF
--- a/.github/workflows/readme-drift.yml
+++ b/.github/workflows/readme-drift.yml
@@ -90,6 +90,8 @@ jobs:
       - name: Post PR comment
         if: steps.drift.outputs.has_drift == 'true'
         uses: actions/github-script@v7
+        env:
+          DRIFT_OUTPUT: ${{ steps.drift.outputs.drift }}
         with:
           script: |
             const body = [
@@ -105,7 +107,6 @@ jobs:
               '> To update, find your locale in the table above and bump the `l10n-source-commit` annotation.',
             ].join('\n');
 
-            const { DATA: commentBody } = process.env;
             const { number: prNumber } = context.payload.pull_request;
             const { owner, repo } = context.repo;
 
@@ -206,6 +207,11 @@ jobs:
 
       - name: Upsert tracking issue
         uses: actions/github-script@v7
+        env:
+          HEAD_SHA: ${{ steps.state.outputs.head_sha }}
+          TOTAL_H2: ${{ steps.state.outputs.total_h2 }}
+          DRIFT_MARKDOWN: ${{ steps.state.outputs.drift_md }}
+          TOTAL_BEHIND: ${{ steps.state.outputs.total_behind }}
         with:
           script: |
             const { owner, repo } = context.repo;
--- a/.github/workflows/readme-drift.yml
+++ b/.github/workflows/readme-drift.yml
@@ -270,7 +270,21 @@ jobs:
             const ours = existing.data.find(i => i.title.startsWith('README locales out of sync'));
 
             if (ours) {
-              // Append new snapshot as comment, then update body with collapsed history
+              // Append a history comment before updating body (preserves prior snapshots)
+              const historyComment = [
+                `### Snapshot — ${weekOf}`,
+                '',
+                `**Source:** \`README.md\` @ [\`${headSha.slice(0, 7)}\`](https://github.com/${owner}/${repo}/commit/${headSha}) · ${totalH2} H2 · ${totalBehind} behind`,
+                '',
+                driftMd || '- (no locale READMEs found)',
+              ].join('\n');
+
+              await github.rest.issues.createComment({
+                owner, repo,
+                issue_number: ours.number,
+                body: historyComment,
+              });
+
               const newBody = [
                 '## 📖 README Locale Sync Status — latest',
                 '',
--- a/.github/workflows/readme-drift.yml
+++ b/.github/workflows/readme-drift.yml
@@ -1,0 +1,256 @@
+name: README Locale Drift
+
+on:
+  pull_request:
+    paths:
+      - 'README.md'
+  schedule:
+    # Every Monday at 09:00 UTC
+    - cron: '0 9 * * 1'
+
+permissions:
+  pull-requests: write
+  issues: write
+
+jobs:
+  # ─── PR trigger: comment on README.md PRs with behind-locale summary ───
+  pr-comment:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Detect drifting locales
+        id: drift
+        run: |
+          set -euo pipefail
+
+          readonly HEAD_SHA="${{ github.event.pull_request.head.sha }}"
+          readonly SOURCE_FILE="README.md"
+
+          # Map locale suffix → README file
+          declare -A LOCALES=(
+            [ar]="README.ar.md"
+            [de]="README.de.md"
+            [es]="README.es.md"
+            [fr]="README.fr.md"
+            [ja-JP]="README.ja-JP.md"
+            [ko]="README.ko.md"
+            [pt-BR]="README.pt-BR.md"
+            [ru]="README.ru.md"
+            [tr]="README.tr.md"
+            [uk]="README.uk.md"
+            [zh-CN]="README.zh-CN.md"
+            [zh-TW]="README.zh-TW.md"
+          )
+
+          DRIFT_OUTPUT=""
+          UP_TO_DATE=""
+
+          for locale in "${!LOCALES[@]}"; do
+            readme="${LOCALES[$locale]}"
+
+            # Read the annotated SHA from the translated README
+            annotated_sha=$(grep -m1 'l10n-source-commit:' "$readme" 2>/dev/null | sed 's/.*l10n-source-commit: //' | tr -d ' ')
+            if [[ -z "$annotated_sha" ]]; then
+              # No annotation yet — skip rather than false positive
+              continue
+            fi
+
+            # Check if the annotated SHA exists in history (handle rebased branches)
+            if ! git rev-list --quiet "$annotated_sha" -1 2>/dev/null; then
+              continue
+            fi
+
+            # Count H2 sections that differ between annotated SHA and HEAD
+            h2_annotated=$(git diff "$annotated_sha" HEAD -- "$SOURCE_FILE" 2>/dev/null | grep '^+## ' | wc -l || true)
+            h2_head=$(git diff "$annotated_sha" HEAD -- "$SOURCE_FILE" 2>/dev/null | grep '^-## ' | wc -l || true)
+
+            # Total number of H2 sections in HEAD (for context)
+            total_h2=$(grep -c '^## ' "$SOURCE_FILE" 2>/dev/null || echo 0)
+
+            behind_count=$((h2_annotated > h2_head ? h2_annotated : h2_head))
+
+            if [[ "$behind_count" -gt 0 ]]; then
+              DRIFT_OUTPUT+="- **$locale** — ~$behind_count H2 sections behind\n"
+            else
+              UP_TO_DATE+="- $locale ✅\n"
+            fi
+          done
+
+          echo "drift<<EOF" >> "$GITHUB_OUTPUT"
+          echo -e "${DRIFT_OUTPUT:-None}"
+          echo "EOF"
+          echo "up_to_date<<EOF" >> "$GITHUB_OUTPUT"
+          echo -e "${UP_TO_DATE:-All locales up to date}"
+          echo "EOF"
+          echo "has_drift=$([[ -n "$DRIFT_OUTPUT" ]] && echo true || echo false)" >> "$GITHUB_OUTPUT"
+
+      - name: Post PR comment
+        if: steps.drift.outputs.has_drift == 'true'
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const body = [
+              '## 📖 README Locale Drift Detected',
+              '',
+              'This PR touched `README.md`. The following locale READMEs are behind the current HEAD:',
+              '',
+              '```',
+              process.env.DRIFT_OUTPUT,
+              '```',
+              '',
+              '> ℹ️ This is a passive signal — not a merge blocker. Locale maintainers sync at their own pace.',
+              '> To update, find your locale in the table above and bump the `l10n-source-commit` annotation.',
+            ].join('\n');
+
+            const { DATA: commentBody } = process.env;
+            const { number: prNumber } = context.payload.pull_request;
+            const { owner, repo } = context.repo;
+
+            // Check if we already commented
+            const existing = await github.rest.issues.listComments({
+              owner, repo,
+              issue_number: prNumber,
+              per_page: 50,
+            });
+
+            const ourComment = existing.data.find(c =>
+              c.user.login === 'github-actions[bot]' &&
+              c.body.includes('README Locale Drift Detected')
+            );
+
+            if (ourComment) {
+              await github.rest.issues.updateComment({
+                owner, repo,
+                comment_id: ourComment.id,
+                body,
+              });
+            } else {
+              await github.rest.issues.createComment({
+                owner, repo,
+                issue_number: prNumber,
+                body,
+              });
+            }
+
+  # ─── Cron trigger: open/update rolling tracking issue ───
+  track:
+    if: github.event_name == 'schedule'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Compute drift state
+        id: state
+        run: |
+          set -euo pipefail
+
+          readonly HEAD_SHA=$(git rev-parse HEAD)
+          readonly SOURCE_FILE="README.md"
+          readonly TOTAL_H2=$(grep -c '^## ' "$SOURCE_FILE" 2>/dev/null || echo 0)
+
+          declare -A LOCALES=(
+            [ar]="README.ar.md"
+            [de]="README.de.md"
+            [es]="README.es.md"
+            [fr]="README.fr.md"
+            [ja-JP]="README.ja-JP.md"
+            [ko]="README.ko.md"
+            [pt-BR]="README.pt-BR.md"
+            [ru]="README.ru.md"
+            [tr]="README.tr.md"
+            [uk]="README.uk.md"
+            [zh-CN]="README.zh-CN.md"
+            [zh-TW]="README.zh-TW.md"
+          )
+
+          DRIFT_MARKDOWN=""
+          TOTAL_BEHIND=0
+
+          for locale in "${!LOCALES[@]}"; do
+            readme="${LOCALES[$locale]}"
+
+            annotated_sha=$(grep -m1 'l10n-source-commit:' "$readme" 2>/dev/null | sed 's/.*l10n-source-commit: //' | tr -d ' ')
+            if [[ -z "$annotated_sha" ]]; then
+              DRIFT_MARKDOWN+="- [ ] **$locale** — no source commit annotated yet\n"
+              continue
+            fi
+
+            if ! git rev-list --quiet "$annotated_sha" -1 2>/dev/null; then
+              DRIFT_MARKDOWN+="- [ ] **$locale** — annotated SHA `$annotated_sha` not in history (may have been rebased)\n"
+              continue
+            fi
+
+            # H2 sections added in HEAD vs annotated (new sections upstream added)
+            h2_added=$(git diff "$annotated_sha" HEAD -- "$SOURCE_FILE" 2>/dev/null | grep '^+## ' | wc -l || true)
+            h2_removed=$(git diff "$annotated_sha" HEAD -- "$SOURCE_FILE" 2>/dev/null | grep '^-## ' | wc -l || true)
+            behind=$((h2_added > h2_removed ? h2_added : h2_removed))
+
+            if [[ "$behind" -gt 0 ]]; then
+              DRIFT_MARKDOWN+="- [ ] **$locale** — $behind H2 section(s) behind (annotated SHA: \`$annotated_sha\`)\n"
+              TOTAL_BEHIND=$((TOTAL_BEHIND + 1))
+            else
+              DRIFT_MARKDOWN+="- [x] **$locale** — up to date ✅\n"
+            fi
+          done
+
+          echo "head_sha=$HEAD_SHA" >> "$GITHUB_OUTPUT"
+          echo "total_h2=$TOTAL_H2" >> "$GITHUB_OUTPUT"
+          echo "drift_md<<EOF" >> "$GITHUB_OUTPUT"
+          echo -e "$DRIFT_MARKDOWN"
+          echo "EOF"
+          echo "total_behind=$TOTAL_BEHIND" >> "$GITHUB_OUTPUT"
+
+      - name: Upsert tracking issue
+        uses: actions/github-script@v7
+        with:
+          script: |
+            const { owner, repo } = context.repo;
+            const weekOf = new Date().toISOString().slice(0, 10);
+            const title = `README locales out of sync — week of ${weekOf}`;
+            const headSha = process.env.HEAD_SHA;
+            const totalH2 = process.env.TOTAL_H2;
+            const driftMd = process.env.DRIFT_MARKDOWN || '';
+            const totalBehind = parseInt(process.env.TOTAL_BEHIND || '0', 10);
+
+            const body = [
+              '## 📖 README Locale Sync Status',
+              '',
+              `**Source:** \`README.md\` @ [\`${headSha.slice(0, 7)}\`](https://github.com/${owner}/${repo}/commit/${headSha}) · $totalH2 H2 sections total',
+              '',
+              '### Locale checklist',
+              '',
+              driftMd || '- (no locale READMEs found)',
+              '',
+              '---',
+              '*Auto-generated by [`.github/workflows/readme-drift.yml`](.github/workflows/readme-drift.yml) ·兑*',
+            ].join('\n');
+
+            // Look for an existing tracking issue
+            const existing = await github.rest.issues.listForRepo({
+              owner, repo,
+              labels: ['locale-drift'],
+              state: 'open',
+              per_page: 20,
+            });
+
+            const ours = existing.data.find(i => i.title.startsWith('README locales out of sync'));
+
+            if (ours) {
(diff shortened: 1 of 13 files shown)


.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+  pull_request:
+    paths:
+      - 'README.md'
+  schedule:
+    # Every Monday at 09:00 UTC
+    - cron: '0 9 * * 1'
+
+permissions:
+  pull-requests: write
+  issues: write
```
Siri-Ray, line 14: **Blocking:** Add `contents: read` to the workflow permissions before this runs in Actions. This workflow defines an explicit top-level `permissions:` block with only `pull-requests: write` and `issues: write`; in GitHub Actions, any unspecified `GITHUB_TOKEN` scope is set to `none` once the block is present. Both jobs immediately call `actions/checkout@v4`, which needs repository contents access, so the jobs can fail before reaching any of the drift logic. Please add `contents: read` here (or job-level permissions that include it) while keeping the write scopes needed for comments/issues.

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+
+jobs:
+  # ─── PR trigger: comment on README.md PRs with behind-locale summary ───
+  pr-comment:
+    if: github.event_name == 'pull_request'
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
Siri-Ray, line 23: **Blocking:** Checkout needs full enough history for the annotated-SHA comparisons. The new drift loops call `git rev-list` and `git diff "$annotated_sha" HEAD -- README.md`, but both checkout steps use the default shallow fetch depth of 1. The annotations this PR adds point at `87a95b7fb4eeb0a5c954dbe0dea95b3fa5c0dc46`; after a future README change, that commit usually will not be present in a depth-1 checkout, so the PR job silently skips every locale and the scheduled job reports the annotations as missing from history instead of measuring drift. Configure this checkout and the scheduled checkout below with `with: fetch-depth: 0` (or explicitly fetch the annotated commits before diffing) so the comparisons can actually see the stored source commits.

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+                issue_number: ours.number,
+                title,
+                body,
+              });
+            } else {
+              const created = await github.rest.issues.create({
+                owner, repo,
+                title,
+                body,
+                labels: ['locale-drift'],
```
Siri-Ray, line 317: **Blocking:** Do not assume the `locale-drift` label already exists when creating the first tracking issue. The PR body says the first cron run can create/use this label, but the repository does not currently have a `locale-drift` label, and `issues.create` rejects unknown labels instead of creating them. That means the initial scheduled run reaches this `labels: ['locale-drift']` path and fails without opening the rolling issue. Please either create/upsert the label in the workflow before creating the issue, use an existing label such as `i18n`/`documentation`, or add the label as part of the repository setup before relying on it here.

Siri-Ray (commented):
I found a few workflow correctness blockers that would keep the new README drift detection from running reliably. Details inline.


.github/workflows/readme-drift.yml
lefarcen, line 23: **P1 — Shallow checkout hides annotated commits**

`actions/checkout` defaults to `fetch-depth: 1`, but both jobs compare locale annotations against older commits. On PRs this makes `git rev-list "$annotated_sha"` fail and silently skip locales, and on cron it reports valid annotations as "not in history".

**Fix:** Set `fetch-depth: 0` on both checkout steps:
```yaml
- uses: actions/checkout@v4
  with:
    fetch-depth: 0
```

Or fetch each annotated commit individually before diffing:
```bash
if ! git rev-list "$annotated_sha" --max-count=1 &>/dev/null; then
  git fetch origin "$annotated_sha" --depth=1
fi
```

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+              continue
+            fi
+
+            # Check if the annotated SHA exists in history (handle rebased branches)
+            if ! git rev-list --quiet "$annotated_sha" -1 2>/dev/null; then
+              continue
+            fi
+
+            # Count H2 sections that differ between annotated SHA and HEAD
+            h2_annotated=$(git diff "$annotated_sha" HEAD -- "$SOURCE_FILE" 2>/dev/null | grep '^+## ' | wc -l || true)
```
lefarcen: **P2 — H2-only diff misses edits inside sections**

This only counts added/removed `##` headings, so edits inside existing README sections produce `behind_count=0` and the PR stays silent even though the locale is stale.

**Example:** README.md gains a new paragraph under "## Quick Start" — no H2 added → drift invisible.

**Fix:** Map any changed lines since the annotation to their containing H2 section, or at least treat any source diff as drift:
```bash
if git diff "$annotated_sha" HEAD -- README.md | grep -q '^[+-]'; then
  # Any diff = drift
  behind_locales="$behind_locales $locale"
fi
```

Alternatively, parse H2 section line ranges from both SHAs and diff the line counts per section.

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+          )
+
+          DRIFT_OUTPUT=""
+          UP_TO_DATE=""
+
+          for locale in "${!LOCALES[@]}"; do
+            readme="${LOCALES[$locale]}"
+
+            # Read the annotated SHA from the translated README
+            annotated_sha=$(grep -m1 'l10n-source-commit:' "$readme" 2>/dev/null | sed 's/.*l10n-source-commit: //' | tr -d ' ')
```
lefarcen, line 58: **P2 — Annotation accepts non-SHA refs**

The annotation parser accepts any rev name that Git resolves, not just a source SHA. A hand edit like `l10n-source-commit: HEAD` or a branch/tag name can pass `rev-list` and produce false "up to date" results.

**Fix:** Validate the extracted value against the documented SHA format and verify it is a commit ancestor of the checked source:
```bash
annotated_sha=$(grep -oP '(?<=l10n-source-commit: )[0-9a-f]{7,40}' "$locale_file" | head -1)
if [[ -z "$annotated_sha" || ! "$annotated_sha" =~ ^[0-9a-f]{7,40}$ ]]; then
  echo "Invalid annotation in $locale_file: $annotated_sha"
  continue
fi
# Verify it's an ancestor of main
if ! git merge-base --is-ancestor "$annotated_sha" HEAD 2>/dev/null; then
  echo "Annotation SHA $annotated_sha not in history"
fi
```

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+          echo -e "${DRIFT_OUTPUT:-None}"
+          echo "EOF"
+          echo "up_to_date<<EOF" >> "$GITHUB_OUTPUT"
+          echo -e "${UP_TO_DATE:-All locales up to date}"
+          echo "EOF"
+          echo "has_drift=$([[ -n "$DRIFT_OUTPUT" ]] && echo true || echo false)" >> "$GITHUB_OUTPUT"
+
+      - name: Post PR comment
+        if: steps.drift.outputs.has_drift == 'true'
+        uses: actions/github-script@v7
```
lefarcen, line 100: **P2 — Fork PR comment can fail → breaks passive promise**

The PR says this is passive, but on forked `pull_request` runs the token usually cannot create or update comments, so this `github-script` step can fail the workflow.

**Fix:** Either make the comment step non-blocking:
```yaml
- name: Comment on PR
  continue-on-error: true
  uses: actions/github-script@v7
  ...
```

Or move commenting to a carefully constrained `pull_request_target` design that does not execute untrusted checkout content. (See GitHub docs on `pull_request_target` security.)

.github/workflows/readme-drift.yml
lefarcen, line 317: **P2 — First cron run fails if label doesn't exist**

The first cron run will fail if the `locale-drift` label does not already exist, but the PR test plan says it is "created by first cron run". `issues.create` does not create missing labels.

**Fix:** Create/upsert the label before creating the issue:
```yaml
- name: Ensure locale-drift label exists
  uses: actions/github-script@v7
  with:
    script: |
      try {
        await github.rest.issues.getLabel({
          owner: context.repo.owner,
          repo: context.repo.repo,
          name: 'locale-drift'
        });
      } catch (err) {
        if (err.status === 404) {
          await github.rest.issues.createLabel({
            owner: context.repo.owner,
            repo: context.repo.repo,
            name: 'locale-drift',
            color: 'FBCA04',
            description: 'README locale drift tracking'
          });
        }
      }
```

Or avoid requiring the label for initial issue creation (omit `labels: ['locale-drift']` and add it afterwards).

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+            const existing = await github.rest.issues.listForRepo({
+              owner, repo,
+              labels: ['locale-drift'],
+              state: 'open',
+              per_page: 20,
+            });
+
+            const ours = existing.data.find(i => i.title.startsWith('README locales out of sync'));
+
+            if (ours) {
```
lefarcen, line 272: **P2 — Weekly issue overwrites prior state → can't evaluate RFC Phase 2 triggers**

Updating one rolling issue overwrites the prior week's state, so the workflow does not preserve enough data to evaluate the RFC's Phase 2 triggers such as ">50% incomplete for ≥4 weeks". `TOTAL_BEHIND` is also computed but unused.

**Fix:** Persist weekly snapshots in the issue body or comments:
```markdown
## Current Status (2026-W20)
- [x] ar (✅)
- [ ] de (3 sections behind)
...

---
<details>
<summary>Prior weeks</summary>

### 2026-W19
- [x] ar (✅)
- [ ] de (3 sections behind)
...

</details>
```

Include counts needed for the `>=5 locales behind by >=3 H2 sections` trigger:
```bash
LARGE_LAG_COUNT=$(echo "$behind_locales" | tr ' ' '\n' | while read locale; do
  behind=$(compute_behind_count "$locale")
  [[ $behind -ge 3 ]] && echo "$locale"
done | wc -l)

echo "LARGE_LAG_COUNT=$LARGE_LAG_COUNT" >> issue_body.md
```

lefarcen (changes requested):
Thanks @laihenyi for this PR! 🎉

The Phase 1 direction (MDN-style annotation + passive CI drift detection) is solid and matches the RFC discussion in #195. The implementation needs corrections on a few fronts before it can reliably detect drift:

**Key blockers:**
- Default shallow checkout hides annotated source commits → drift computation fails
- H2-only diff misses common edits inside sections → false "up to date"
- Fork PR comment can fail → breaks passive promise
- Weekly issue overwrites prior state → can't evaluate RFC Phase 2 triggers

**Good parts:**
- Clean annotation format matches MDN precedent
- Two-job split (pr-comment passive + cron track) is the right shape
- No auto-translation / merge gate per RFC agreement

See inline comments for concrete fixes. Most are scoped to the workflow YAML and don't require re-architecting. Looking forward to the revision!

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+
+on:
+  pull_request:
+    paths:
+      - 'README.md'
+  schedule:
+    # Every Monday at 09:00 UTC
+    - cron: '0 9 * * 1'
+
+permissions:
```
lefarcen, line 11: **P2 — Missing `contents: read` permission**

Because this workflow declares explicit `permissions:`, unspecified scopes default to `none`. `actions/checkout` needs `contents: read`, so add it here or checkout can fail before either job does useful work.

**Fix:**
```yaml
permissions:
  contents: read
  pull-requests: write
  issues: write
```

.github/workflows/readme-drift.yml
lefarcen, line 23: **P1 — Shallow checkout hides annotated commits** (same as prior review)

The checkout still uses the default shallow history. Since the workflow validates annotated commits with `git rev-list`, translated README annotations outside the shallow checkout will be treated as missing.

**Fix:**
```yaml
- uses: actions/checkout@v4
  with:
    fetch-depth: 0
```

.github/workflows/readme-drift.yml
lefarcen, line 58: **P2 — Annotation parser accepts non-SHA refs** (same as prior review)

The annotation parser still accepts arbitrary refs instead of validating a full commit SHA. Values like branch names or other resolvable refs can pass `git rev-list`, which makes drift comparisons depend on mutable refs rather than the recorded source commit.

**Fix:** Validate against `^[0-9a-f]{7,40}$` regex and verify it's a commit ancestor.

.github/workflows/readme-drift.yml
lefarcen: **P2 — H2-only diff misses edits inside sections** (same as prior review)

The drift calculation still only counts added/removed H2 headings. Edits inside an existing section leave the count at zero, so locale README files can be reported up to date even when source content changed.

**Fix:** Map any changed lines since the annotation to their containing H2 section, or treat any source diff as drift:
```bash
if git diff "$annotated_sha" HEAD -- README.md | grep -q '^[+-]'; then
  behind_locales="$behind_locales $locale"
fi
```

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,262 @@
...
+
+          echo "drift<<EOF" >> "$GITHUB_OUTPUT"
+          echo -e "${DRIFT_OUTPUT:-None}"
+          echo "EOF"
+          echo "up_to_date<<EOF" >> "$GITHUB_OUTPUT"
+          echo -e "${UP_TO_DATE:-All locales up to date}"
+          echo "EOF"
+          echo "has_drift=$([[ -n "$DRIFT_OUTPUT" ]] && echo true || echo false)" >> "$GITHUB_OUTPUT"
+
+      - name: Post PR comment
```
lefarcen, line 97: **P2 — Fork PR comment can fail** (same as prior review)

The PR comment step still lacks `continue-on-error`. On fork PRs the token may not be allowed to write comments, causing this passive signal to fail the workflow.

**Fix:**
```yaml
- name: Comment on PR
  continue-on-error: true
  uses: actions/github-script@v7
  ...
```

.github/workflows/readme-drift.yml
lefarcen, line 272: **P2 — Weekly issue overwrites prior state** (same as prior review)

The scheduled job still updates the existing tracking issue body in place. That overwrites prior weekly state, so later Phase 2 trigger logic cannot compare this run against previous drift history.

**Fix:** Persist weekly snapshots in the issue body (current + collapsed prior weeks history).

.github/workflows/readme-drift.yml
lefarcen, line 317: **P2 — First cron run fails if label doesn't exist** (same as prior review)

The cron path still creates issues with `locale-drift` without ensuring the label exists first. On repositories where the label has not been created, the first scheduled run will fail.

**Fix:** Create/upsert the label before creating the issue, or use an existing label.

lefarcen (changes requested):
Thanks for the quick iteration @laihenyi! I re-reviewed the current head and see that the prior findings are not yet addressed.

**Still blocking:**
- Shallow checkout (P1) — needs `fetch-depth: 0`
- Missing `contents: read` permission (P2, new) — workflow will fail at checkout
- H2-only diff gaps (P2)
- Annotation validation (P2)
- Fork PR permissions (P2)
- Label creation (P2)
- Phase 2 data preservation (P2)

Siri-Ray's Looper review also caught the checkout depth + `contents: read` + label issues. See inline for concrete fixes — most are small YAML additions. Looking forward to the next revision!

.github/workflows/readme-drift.yml
```diff
@@ -0,0 +1,305 @@
...
+
+            if (ours) {
+              // Append new snapshot as comment, then update body with collapsed history
+              const newBody = [
+                '## 📖 README Locale Sync Status — latest',
+                '',
+                currentSnapshot,
+                '',
+                '---',
+                '*Auto-generated by [`.github/workflows/readme-drift.yml`](.github/workflows/readme-drift.yml) ·兑*',
```
lefarcen, line 294: **P2 — Phase 2 data preservation still not implemented:** The RFC discussion in #195 agreed that Phase 2 triggers ("drift velocity outpaces human capacity") would be evaluated from weekly tracking issue history. The current `track` job updates the existing issue body with only the latest snapshot and does not preserve prior snapshots or append the promised history comment. This means after 4 weeks, you cannot evaluate whether "≥5 locales behind by ≥3 H2 sections" was true in week 1 vs week 4 — all prior snapshots are lost.

Fix: After updating the issue body, append a dated comment with the current snapshot. Example:
```yaml
gh issue comment "$issue_num" --repo nexu-io/open-design --body "$(cat <<SNAPSHOT
### Weekly snapshot $(date -u +%Y-%m-%d)
$new_body
SNAPSHOT
)"
```

This preserves the timeline so the RFC Phase 2 evaluation can compare week-over-week drift velocity.

lefarcen (changes requested):
Great progress @laihenyi! 🎉 You've addressed 6 out of 7 prior blockers:

**✅ Fixed:**
- Shallow checkout → now `fetch-depth: 0` ✓
- Missing `contents: read` → added to permissions ✓
- H2-only diff gaps → now detects non-H2 edits ✓
- Annotation validation → validates commit exists before diffing ✓
- Fork PR permissions → uses `pull_request_target` ✓
- Label creation → upserts label before use ✓

**⚠️ Still needs fix:**
- **Phase 2 data preservation** (P2) — see inline comment

Once the scheduled job's history comment is preserved across runs, this will be ready. Almost there!

lefarcen (approved):
Perfect! 🎉 The last blocker is resolved.

**Phase 2 data preservation is now fixed:** The scheduled job appends a dated snapshot comment to the tracking issue before updating the body, preserving historical snapshots for RFC Phase 2 evaluation.

All 7 original findings are addressed. The workflow is now production-ready:
- ✅ Full history checkout for commit lookups
- ✅ Explicit `contents: read` permission
- ✅ Detects non-H2 README edits
- ✅ Validates annotated commits exist
- ✅ Fork PR-safe with `pull_request_target`
- ✅ Label upsert before use
- ✅ History comment preserves week-over-week drift data

Looks good to me; deferring final merge decision to a maintainer.

Siri-Ray (approved):
@laihenyi I reviewed the current README drift workflow and locale annotation changes against the latest head. The prior workflow blockers around checkout depth, permissions, non-H2 drift detection, annotation validation, fork-safe commenting, label creation, and weekly history preservation are addressed, and I did not find additional actionable issues in the changed ranges. Thanks for the careful iteration on this feature.


alchemistklk:
Thanks for working on this. The problem this PR addresses is real: we do need a way to reduce README/localization drift across multiple languages.

For now, we would like to hold this PR for now while our team decides what localization automation workflow we actually want to adopt. In particular, we want to make sure the workflow reduces drift without creating a recurring maintenance burden through scheduled issues, stale checklist tracking, or noisy PR comments that nobody clearly owns.

So this is not a rejection of the direction, but we need to settle the team-level process and ownership model first before merging automation like this.

laihenyi:
Thanks @alchemistklk — opening this up for team-level discussion was the original intent, so this is exactly the conversation I was hoping to land before any automation ships. Happy to hold here while the team settles the ownership and maintenance model; I'll stand by and pick this back up once there's consensus on the shape you'd like to align against.

PR #1298, opened 2026-05-11
fix(desktop): swallow setTypeOfService EINVAL crashes in dev main (#647)

Nagendhra-web:
Closes #647.

The packaged Electron entry already filters the undici \`setTypeOfService EINVAL\` crash (issue #895), but the dev / source-built desktop entry was missing the parallel guard. Switching settings tabs in a source-run desktop fires a fresh fetch; on macOS / VPN configurations where the kernel refuses to set the IP_TOS byte on outbound sockets, the rejection bubbles to Electron's default handler and surfaces as the "JavaScript error in the main process" dialog from #647.

## What changes

New module [\`apps/desktop/src/main/uncaught-exception.ts\`](https://github.com/nexu-io/open-design/blob/fix/issue-647-desktop-settypeofservice-guard/apps/desktop/src/main/uncaught-exception.ts):

- \`isHarmlessSocketOptionError\` matches only the canonical undici shape (syscall name AND \`EINVAL\` code). A contradicting structured \`code\` such as \`EACCES\` explicitly fails the match so real bugs do not get hidden.
- The \`uncaughtException\` handler logs harmless cases at warn and returns silently. For anything else it removes itself from the listener list and re-throws via \`setImmediate\`, restoring Node's default crash path so Electron's native dialog renders exactly as it would without this filter.
- \`unhandledRejection\` mirrors the same harmless / fall-through split.

[\`apps/desktop/src/main/index.ts\`](https://github.com/nexu-io/open-design/blob/fix/issue-647-desktop-settypeofservice-guard/apps/desktop/src/main/index.ts) calls \`attachDesktopProcessErrorFilter()\` at the top of \`runDesktopMain\` (before \`app.whenReady\`) so the filter is armed by the time the renderer fires its first fetch.

## Why duplicate, not import

AGENTS.md forbids one app package from importing another app's private \`src/\`. The file header calls out the parallel to [\`apps/packaged/src/logging.ts\`](https://github.com/nexu-io/open-design/blob/main/apps/packaged/src/logging.ts) and notes that the two copies should stay in sync until the helper is promoted to a shared workspace package. The contract is identical, so any regression in one's matching logic will surface in the other's test suite.

## Coverage

[\`apps/desktop/tests/main/uncaught-exception.test.ts\`](https://github.com/nexu-io/open-design/blob/fix/issue-647-desktop-settypeofservice-guard/apps/desktop/tests/main/uncaught-exception.test.ts) mirrors the packaged test:

- 8 cases pinning the matcher shape (canonical hit, message-only hit, generic EINVAL not matched, EACCES not matched, contradicting code not matched, non-Error inputs rejected, non-string message rejected).
- 2 cases pinning the handler split: harmless cases log at warn and do not detach; everything else logs at error, removes the listener, and defers a re-throw to \`setImmediate\`.

## Validated

- \`pnpm guard\` clean.
- \`pnpm --filter @open-design/desktop typecheck\` clean.
- \`pnpm --filter @open-design/desktop build\` clean.
- \`pnpm --filter @open-design/desktop test\` (14 passed, 10 new).

### DIFF
--- a/apps/desktop/src/main/uncaught-exception.ts
+++ b/apps/desktop/src/main/uncaught-exception.ts
@@ -34,8 +34,14 @@
  *     Node's default crash path so Electron's native error dialog
  *     renders exactly as it would without this filter.
  *
- *  3. `unhandledRejection` mirrors the same harmless / fall-through
- *     split.
+ *  3. The installed `unhandledRejection` handler mirrors that policy:
+ *     harmless `setTypeOfService EINVAL` rejections log at warn and
+ *     return; every other rejection logs at error, removes the
+ *     listener, and schedules a re-throw via `setImmediate`. Without
+ *     the detach + rethrow a non-harmless rejection would land in
+ *     this log line and silently keep the process alive, which would
+ *     hide real main-process bugs (issue #647 review by Siri-Ray and
+ *     the codex P2 thread on PR #1298).
  */
 
 /**
@@ -91,6 +97,34 @@ export function createDesktopUncaughtExceptionHandler(
   return handler;
 }
 
+/**
+ * Parallel factory for the `unhandledRejection` listener. Harmless
+ * undici `setTypeOfService EINVAL` rejections log at warn and return;
+ * anything else logs at error, removes the listener, and schedules a
+ * re-throw via `setImmediate`. The detached re-throw lands in the
+ * `uncaughtException` listener installed above, which then walks its
+ * own non-harmless path (log + detach + rethrow), so Node's default
+ * crash semantics take over and Electron's native error dialog
+ * renders as it would without this filter. Pinned by a unit test so a
+ * future refactor that drops the detach can't silently regress.
+ */
+export function createDesktopUnhandledRejectionHandler(
+  logger: DesktopErrorFilterLogger,
+): (reason: unknown) => void {
+  const handler = (reason: unknown): void => {
+    if (isHarmlessSocketOptionError(reason)) {
+      logger.warn('desktop main swallowed harmless socket option rejection', { reason });
+      return;
+    }
+    logger.error('desktop main unhandled rejection', { reason });
+    process.removeListener('unhandledRejection', handler);
+    setImmediate(() => {
+      throw reason;
+    });
+  };
+  return handler;
+}
+
 /** Install the filter on `process`. Idempotent across the dev entry's
  *  hot-reload paths because Node's listener registry deduplicates
  *  identical function references; we capture the once-built handler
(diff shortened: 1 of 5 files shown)


apps/desktop/src/main/uncaught-exception.ts
```diff
@@ -0,0 +1,127 @@
...
+  if (installedHandler !== null) return;
+  const handler = createDesktopUncaughtExceptionHandler(logger);
+  installedHandler = handler;
+  process.on('uncaughtException', handler);
+  process.on('unhandledRejection', (reason) => {
+    if (isHarmlessSocketOptionError(reason)) {
+      logger.warn('desktop main swallowed harmless socket option rejection', { reason });
+      return;
+    }
+    logger.error('desktop main unhandled rejection', { reason });
```
Siri-Ray: This non-harmless `unhandledRejection` branch now logs and returns, which changes the dev desktop main process from Node/Electron's previous fail-fast behavior into a process that keeps running after arbitrary rejected promises. The evidence is this newly installed `process.on('unhandledRejection', ...)` listener: harmless `setTypeOfService EINVAL` rejections are correctly swallowed, but every other rejection reaches this line and then the listener exits without rethrowing or removing itself. That means a real main-process bug, failed IPC registration path, or unexpected async exception can be hidden as a console line instead of surfacing through the default crash/error path this filter is meant to preserve for non-harmless failures. Please mirror the `uncaughtException` fall-through for this branch: either route non-harmless reasons through `createDesktopUncaughtExceptionHandler(logger)` or remove the rejection listener and schedule a `throw reason` with `setImmediate`, and add a test that a non-harmless rejection is not swallowed.

Siri-Ray (commented):
This pass found one blocking process-lifecycle issue in the new desktop error filter; details are inline.


lefarcen (commented):
Thanks for the quick fix @Nagendhra-web! ✅

Verified that dd9d536f addresses Siri-Ray's blocking concern:

**What changed:**
- Both `apps/desktop/src/main/uncaught-exception.ts` and `apps/packaged/src/logging.ts` now have parallel rejection handlers that mirror the `uncaughtException` pattern
- Harmless `setTypeOfService EINVAL` rejections → log at warn, return (swallow)
- Non-harmless rejections → log at error, remove listener, schedule `throw reason` via `setImmediate` (fail-fast restored)
- New tests in both suites pin the non-swallowing behavior for non-harmless cases plus primitive rejection reasons

**Lockstep verified:** Both copies follow the same contract as documented in the file headers. CI is green (build/workspace/detect passed, smoke tests running).

Siri-Ray (approved):
@Nagendhra-web I reviewed the current `dd9d536f` diff across the desktop and packaged error-filter paths. The desktop dev entry now installs the same narrow `setTypeOfService EINVAL` guard before `app.whenReady`, and the follow-up update keeps both `uncaughtException` and `unhandledRejection` behavior fail-fast for non-harmless errors in both copies. The matcher stays intentionally narrow, the package boundaries are respected, and the new Vitest coverage pins harmless, non-harmless, and primitive rejection cases.

Thanks for the careful follow-up here and for keeping the dev and packaged paths in sync. The focused tests could not be re-run locally in this worktree because dependencies are not installed (`vitest` was unavailable), but the PR reports green checks and I did not find actionable changed-range issues.

PR #1300, opened 2026-05-11
fix(web): translate Design Files refresh strings instead of hardcoding English (#1254)

### LINKED ISSUE #1254: Design Files refresh messaging mixes English and Chinese in the same flow
## Summary

In the Design Files refresh flow, the status and error messages switch between English and Chinese instead of staying in one consistent language.

Priority: P2
Reason: This is a localization consistency issue that does not block the feature, but it is very visible and makes the UI feel unfinished and unreliable.

## Current behavior

When refreshing Design Files / live artifacts, some labels and messages appear in English while others appear in Chinese within the same flow.

Examples include mixed-language refresh states and error messages shown together in the same screen.

## Expected behavior

When the app language is set to Chinese, all refresh-related UI copy in this flow should be consistently shown in Chinese.

## Steps to reproduce

1. Set the app language to Chinese.
2. Open a Design Files / live artifact view.
3. Trigger refresh or view refresh-related error states.
4. Observe that some messages are still shown in English while others are in Chinese.

## Notes

- Screenshots were provided showing mixed-language refresh messaging in the same view.
- This appears to affect refresh-related labels, notices, or status text rather than the whole page shell.
- The issue is consistency within one user-visible flow.

## Why this matters

Mixed-language UI breaks immersion and reduces trust in localization quality. In high-frequency flows like refresh and error handling, inconsistent language is especially noticeable.

Image

Image

Nagendhra-web:
Closes #1254.

When the app language is set to Chinese, the Design Files refresh flow showed Chinese for the surrounding chrome but kept English for every label and message in the refresh-history panel (badge descriptor, hero metric, persisted/session sections, document-source labels, debug-metadata heading). Same-screen mixed-language UX, exactly what the screenshots in the issue capture.

## What changes

Threaded `t()` through every user-visible string the panel renders to the user:

- **Hero block.** `Last refreshed` label + `Never` empty state.
- **Created / Last updated facts** with their `Unknown` empty label.
- **Status descriptor** (running / succeeded / failed / ready / never) plus its multi-line description; already in the first pass.
- **Persisted refresh history** section header, hint, empty-state copy, and the timeline status badge (succeeded / running / failed / cancelled / skipped). The badge resolves through a new `describePersistedStatus` helper whose exhaustive switch over `LiveArtifactRefreshLogEntry['status']` makes a future status addition trip tsc rather than silently render as a raw English token.
- **Session activity** section header, hint, empty-state copy, started / succeeded / failed event labels, the started / succeeded / failed detail line (including pluralised `{n} source(s) updated`), and the `just now` relative-time fallback.
- **Document source** section header, hint, `Type` / `Tool` / `Connector` field labels.
- **Advanced debug metadata** summary heading + note line.

### Intentionally untranslated

Raw daemon payloads inside the `<details>` debug panel (`event.step`, `event.refreshId`, `event.error.message`, and the `JSON.stringify` dump) stay English because they are agent / connector identifiers and stack-trace style strings, not localised copy. The heading on the debug panel itself is translated; if the whole debug section should be hidden in localised primary flows, that is a separate UX call.

## i18n keys

23 new keys under `liveArtifact.refresh.*`, on top of the 13 from the first pass. zh-CN strings authored alongside the English source. The 16 other locales pick up the new keys via their existing `...en` spread and the locale-key alignment test stays green; native translations for those locales can land through the usual locale-team passes without re-touching the source code.

## Test coverage

- `tests/components/FileViewer.test.tsx` adds a focused render test that wraps `LiveArtifactRefreshHistoryPanel` in `I18nProvider initial="zh-CN"` and pins the Chinese rendering of every translated label end-to-end, including positive Chinese assertions and negative `expect(markup).not.toContain('English literal')` assertions for the formerly hardcoded strings. With the no-provider fallback returning English, the existing `renderToStaticMarkup` tests could not observe the regression this PR is meant to fix; the zh-CN render test is the only one that would have caught the original gap and will catch the next one.
- `tests/i18n/locales.test.ts` (key alignment + placeholder consistency, 5/5) continues to enforce that every locale stays aligned with English keys.

## Validated

- `pnpm guard` clean.
- `tests/i18n/locales.test.ts` 5/5 pass across all 18 locales.
- `tests/components/FileViewer.test.tsx` 69/69 (+1 zh-CN render test).
- Full web suite: 92 files, 841 tests.


lefarcen (changes requested):
Hey @Nagendhra-web! Thanks for tackling the i18n consistency issue. 👋

The three sites you fixed (describeRefreshStatus, describeEventPhase, timeline body) are solid — the TranslateFn threading is clean and the zh-CN translations look good.

However, reading through the full FileViewer.tsx refresh panel, I found that **many more user-visible English strings remain hardcoded in the same flow**, which means Chinese users will still see mixed language in the Design Files refresh panel.

## P1 — Incomplete i18n coverage

The refresh-history panel still renders multiple user-visible English strings:
- `Last refreshed`, `Never`, `Created`, `Unknown`, `Last updated`
- `Persisted refresh history`, `No persisted refresh history yet.`
- `Session activity`, `Events observed while this tab is open`
- `Document source`, `Source configured`, `Type`, `Tool`, `Connector`
- `Advanced debug metadata`

This undercuts `Closes #1254` because the issue scope is same-flow refresh copy consistency, not only the three helper/timeline strings you addressed.

**Fix**: Add i18n keys for the remaining refresh panel labels/messages and route them through `t()`. If the debug section is intentionally English-only, make that explicit and consider hiding it from the localized primary flow.

## P2 — Test coverage gap

The existing `LiveArtifactRefreshHistoryPanel` tests still render without an `I18nProvider` and assert English text, so they cannot catch the regression this PR is meant to fix or the remaining hardcoded English strings above.

The locale alignment test verifies dictionary key/placeholder shape, but it does not prove this component actually renders zh-CN strings.

**Suggestion**: Add a focused render test with `I18nProvider initial="zh-CN"` that exercises the status badge, empty timeline, started/succeeded rows, singular/plural source count, and at least one panel header/metadata label.

## P2 — PR body reasoning gap

The explanation says the mixed-language root cause is limited to three sites and implies every affected refresh label/message is routed through `t()`, but the code still leaves additional refresh-history panel copy in English. That makes the reasoning incomplete for reviewers and for issue follow-up.

**After fixing the remaining strings**, update the PR body to quantify the full set of translated UI strings and clarify any intentionally untranslated surfaces, such as raw daemon errors or debug metadata.

---

No issues found with the new `TranslateFn` call signatures, the `{n}` placeholder names, or the `sourcesUpdatedOne`/`sourcesUpdatedMany` branch logic — that part is solid. 👍

lefarcen (changes requested):
Thanks for the thorough update @Nagendhra-web. The new pass translates almost all user-facing panel labels, and the focused zh-CN test is a solid addition. However, one hardcoded English path remains:

**P1 — Relative timestamps still output English**

The `formatRelativeTime()` function at line 1055 still hardcodes English units (`ago`, `from now`, `just now`) and is called by the refresh panel's hero/fact/timeline rows. Chinese users will still see mixed-language strings like `5s ago` inside the otherwise-translated panel.

**How to fix**: Route relative-time formatting through i18n or use `Intl.RelativeTimeFormat` with the active locale. The existing `t()` function or a new `formatRelativeTime(ms, t)` helper that looks up unit strings from the locale.

**P2 — Test coverage doesn't cover timestamp path**

The new zh-CN test sets `lastRefreshedAt: undefined` and doesn't assert against the translated timestamp strings, so it can pass while the markup contains English relative-time text from session events.

Once these are addressed, this PR will fully resolve the mixed-language UX from #1254. Great work on the comprehensive i18n coverage!

lefarcen (approved):
Excellent work @Nagendhra-web! The P1 relative-time issue is now fully resolved:

✅ **formatRelativeTime** now uses `Intl.RelativeTimeFormat` with the active locale
✅ **All callsites** thread `locale` and `t` correctly (hero, persisted, session, fact timestamps)
✅ **zh-CN test coverage** now exercises real timestamps and pins Chinese rendering

This PR comprehensively fixes the mixed-language UX from #1254. Great iterative improvement through the review cycles!

**Minor P3 nit** (non-blocking): The `Math.round(value / 60)` at line 1082 rounds negative half-values toward zero, so exact half-boundary past timestamps under-report slightly (e.g., 90s old → `1m ago` instead of `2m ago`). Consider rounding the absolute magnitude first if you want perfect symmetry, but this is negligible and doesn't block merge.

PR #1301, opened 2026-05-11
fix(daemon): mark ghost CLIs unavailable when --version probe is rejected by the OS (#658)

Nagendhra-web:
Closes #658.

Settings > Execution & model > Local CLI kept advertising Codex CLI with a stale version number after the user had uninstalled the binary. Root cause is in [\`apps/daemon/src/runtimes/detection.ts\`](https://github.com/nexu-io/open-design/blob/main/apps/daemon/src/runtimes/detection.ts): the \`probe()\` function ran \`<resolved> --version\` inside a try/catch and the catch arm was unconditional — every failure was swallowed and the agent was still reported as \`available: true\`.

## How the ghost survives the uninstall

\`resolveAgentExecutable\` can still return a non-null path after the user has removed the real binary, because:

- macOS / Linux: a leftover wrapper shim (npm bin shim, homebrew alias, mise / nvm / fnm pointer file) survives the uninstall, but the underlying interpreter / target the shim invokes is gone. \`execFile\` rejects with \`ENOENT\`.
- Windows: a leftover \`.CMD\` shim is still on PATH but points at a deleted target. \`execFile\` rejects with \`ENOENT\`.
- Permissions: the file is still there but the executable bit was stripped. \`execFile\` rejects with \`EACCES\`.

In each of those cases the CLI is genuinely unable to run, but the previous catch arm swallowed the rejection and the Settings UI happily advertised it. The unconditional catch was correct only for the original case it was written for: adapters whose \`--version\` flag is not supported, where \`execFile\` returns non-zero on a binary that itself is healthy.

## Fix

Tighten the catch to inspect the spawn error code:

- \`ENOENT\` / \`EACCES\` / \`ENOTDIR\` → return \`available: false\` immediately. These are OS-level rejections meaning the binary cannot be invoked at all; advertising it as available would lie to the UI.
- Anything else (timeout, non-zero exit, unrecognised flag) → fall through to the previous "available, version unknown" branch so adapters whose \`--version\` flag is unsupported keep working.

## Coverage

New test file [\`apps/daemon/tests/runtimes/probe-ghost-cli.test.ts\`](https://github.com/nexu-io/open-design/blob/fix/issue-658-codex-cli-ghost-after-uninstall/apps/daemon/tests/runtimes/probe-ghost-cli.test.ts) (5 vitest cases):

- Three cases pin each spawn-error code (\`ENOENT\`, \`EACCES\`, \`ENOTDIR\`) to \`available: false\`.
- One case pins \`ETIMEDOUT\` (a typical non-OS-rejection failure) to \`available: true\`, \`version: null\` — the adapter-with-no-version-flag contract that the previous catch arm protected.
- One happy-path case asserts a clean \`--version\` run returns the parsed version string.

## Validated

- \`pnpm guard\` clean.
- \`pnpm --filter @open-design/daemon typecheck\` clean.
- \`pnpm --filter @open-design/daemon build\` clean.
- \`pnpm --filter @open-design/daemon exec vitest run tests/runtimes/probe-ghost-cli.test.ts\` (5 / 5 pass).
- Diffed against \`main\` for the rest of \`tests/runtimes/\`: this branch has strictly fewer failing tests than main (5 vs 8). The remaining failures are pre-existing Windows-only platform flakiness around \`.CMD\` shim resolution that this PR does not touch.

### DIFF
--- a/apps/daemon/src/runtimes/detection.ts
+++ b/apps/daemon/src/runtimes/detection.ts
@@ -1,7 +1,7 @@
 import { execAgentFile } from './invocation.js';
 import { AGENT_DEFS } from './registry.js';
 import { DEFAULT_MODEL_OPTION, rememberLiveModels } from './models.js';
-import { inspectAgentExecutableResolution } from './executables.js';
+import { resolveAgentExecutable } from './executables.js';
 import { spawnEnvForAgent } from './env.js';
 import { agentCapabilities } from './capabilities.js';
 import { installMetaForAgent } from './metadata.js';
@@ -117,8 +117,15 @@ async function probe(
   def: RuntimeAgentDef,
   configuredEnv: Record<string, string> = {},
 ): Promise<DetectedAgent> {
-  const resolution = inspectAgentExecutableResolution(def, configuredEnv);
-  let resolved = resolution.selectedPath;
+  // Resolution returns whichever path the rest of the daemon will spawn
+  // (configured override wins, PATH fallback otherwise). Detection must
+  // probe THAT path and report `available` accordingly, so the Settings
+  // UI never advertises an executable that `resolveAgentBin` won't pick
+  // at run time. Surfacing a different PATH candidate as `available: true`
+  // while a stale configured override survives in chat/run resolution
+  // breaks the invariant flagged on PR #1301 review and would only swap
+  // the ghost in Settings for a ghost in chat (Siri-Ray, #1301 round 3).
+  const resolved = resolveAgentExecutable(def, configuredEnv);
   if (!resolved) {
     return unavailableAgent(def);
   }
@@ -130,24 +137,7 @@ async function probe(
     },
     configuredEnv,
   );
-  let outcome = await probeVersionAtPath(def, resolved, probeEnv);
-  // If the selected executable is not invocable but a distinct PATH
-  // candidate exists, retry there. This is the configured-override
-  // recovery path lefarcen flagged on #1301: when a user has a stale
-  // CODEX_BIN (or other agent-specific override) and a working binary
-  // on PATH, the daemon should adopt the PATH candidate so Settings
-  // can keep `available: true` and the "Test" / "adopt detected
-  // binary" repair flow added in #1205 stays accessible. Without this
-  // retry the user is locked out of self-recovery because Settings
-  // gates the Test button on agent.available.
-  if (
-    outcome.kind === 'not-invocable'
-    && resolution.pathResolvedPath
-    && resolution.pathResolvedPath !== resolved
-  ) {
-    resolved = resolution.pathResolvedPath;
-    outcome = await probeVersionAtPath(def, resolved, probeEnv);
-  }
+  const outcome = await probeVersionAtPath(def, resolved, probeEnv);
   if (outcome.kind === 'not-invocable') {
     return unavailableAgent(def);
   }
--- a/apps/daemon/src/runtimes/detection.ts
+++ b/apps/daemon/src/runtimes/detection.ts
@@ -75,8 +75,39 @@ async function probe(
       timeout: 3000,
     });
     version = String(stdout).trim().split('\n')[0] ?? null;
-  } catch {
-    // binary exists but --version failed; still mark available
+  } catch (err) {
+    // The path resolved on disk but spawning the binary failed. If the
+    // OS rejected the spawn outright (ENOENT for a vanished target,
+    // EACCES for a stripped-x bit, ENOTDIR for a broken parent), the
+    // CLI cannot actually be invoked — mark it unavailable so the
+    // Settings UI does not keep advertising a ghost entry after the
+    // user has uninstalled the real binary (issue #658). Typical
+    // shapes this catches:
+    //   - macOS/Linux: `which codex` returns a stale wrapper shim, but
+    //     the underlying `node` / Python interpreter the shim invokes
+    //     is gone, so `execFile` rejects with ENOENT.
+    //   - Windows: a leftover `.CMD` shim still exists on PATH but
+    //     points at a deleted target; spawn fails with ENOENT.
+    //   - Permissions: the binary file is there but no longer
+    //     executable.
+    // Any other failure (the binary spawns but `--version` returns
+    // non-zero, or stderr noise, or a timeout) still falls through to
+    // the previous "available, version unknown" behaviour so adapters
+    // whose `--version` flag is unsupported keep working.
+    const spawnErrorCode = (err as NodeJS.ErrnoException)?.code;
+    if (
+      spawnErrorCode === 'ENOENT'
+      || spawnErrorCode === 'EACCES'
+      || spawnErrorCode === 'ENOTDIR'
+    ) {
+      return {
+        ...stripFns(def),
+        models: def.fallbackModels ?? [DEFAULT_MODEL_OPTION],
+        available: false,
+        ...installMetaForAgent(def.id),
+      };
+    }
+    // binary exists and spawned but --version failed; still mark available
   }
   // Probe `--help` once per agent and record which flags the installed CLI
   // advertises. Cached on `agentCapabilities` for buildArgs to consult.
--- a/apps/daemon/src/runtimes/detection.ts
+++ b/apps/daemon/src/runtimes/detection.ts
@@ -1,7 +1,7 @@
 import { execAgentFile } from './invocation.js';
 import { AGENT_DEFS } from './registry.js';
 import { DEFAULT_MODEL_OPTION, rememberLiveModels } from './models.js';
-import { resolveAgentExecutable } from './executables.js';
+import { inspectAgentExecutableResolution } from './executables.js';
 import { spawnEnvForAgent } from './env.js';
 import { agentCapabilities } from './capabilities.js';
 import { installMetaForAgent } from './metadata.js';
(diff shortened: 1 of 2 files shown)


apps/daemon/tests/runtimes/probe-ghost-cli.test.ts
```diff
@@ -0,0 +1,99 @@
...
+
+const execAgentFileMock = vi.fn();
+const resolveAgentExecutableMock = vi.fn();
+
+vi.mock('../../src/runtimes/invocation.js', () => ({
+  execAgentFile: (...args: unknown[]) =>
+    (execAgentFileMock as unknown as (...args: unknown[]) => unknown)(...args),
+}));
+
+vi.mock('../../src/runtimes/executables.ts', async (importOriginal) => {
```
Siri-Ray: This new test imports `../../src/runtimes/executables.ts` in both the mock specifier and the `typeof import(...)` type query, but `apps/daemon/tsconfig.tests.json` extends a NodeNext config that does not enable `allowImportingTsExtensions`. That makes the required daemon test typecheck reject this file with the standard `.ts`-extension import error, so the PR's regression coverage cannot pass the repository validation gate. Please keep the ESM-style `.js` specifier here, matching the rest of the daemon tests and the production import (`../../src/runtimes/executables.js`), and use the same `.js` path in the `typeof import(...)` query.

Siri-Ray (commented):
I found one blocking issue in the new daemon test coverage: the added test uses a `.ts` extension in a TypeScript import type under the daemon NodeNext test config, which should break the required test typecheck.


apps/daemon/src/runtimes/detection.ts
```diff
@@ -75,8 +75,39 @@ async function probe(
...
+    //     is gone, so `execFile` rejects with ENOENT.
+    //   - Windows: a leftover `.CMD` shim still exists on PATH but
+    //     points at a deleted target; spawn fails with ENOENT.
+    //   - Permissions: the binary file is there but no longer
+    //     executable.
+    // Any other failure (the binary spawns but `--version` returns
+    // non-zero, or stderr noise, or a timeout) still falls through to
+    // the previous "available, version unknown" behaviour so adapters
+    // whose `--version` flag is unsupported keep working.
+    const spawnErrorCode = (err as NodeJS.ErrnoException)?.code;
```
lefarcen: **P2 — Custom-path recovery flow regression**: The new early `available: false` path only looks at the selected executable from `resolveAgentExecutable`, which may be a configured `CODEX_BIN` override even when `inspectAgentExecutableResolution` would also reveal a working PATH binary. PR #1205 added Settings repair/fallback UX for exactly this case, but Settings only enables the Test button for agents where `agent.available === true`, so this change can mark Codex unavailable before the user can run the repair flow and adopt/clear the bad custom path.

Concrete fix: when the selected path fails with ENOENT/EACCES/ENOTDIR, inspect configured-vs-PATH resolution; if a different PATH candidate exists, probe that candidate for availability (or return a structured state that keeps the Settings repair/test path accessible).

lefarcen: **P2 — Stale wrapper shims with non-errno failures**: The fix claims to cover stale wrapper shims whose underlying interpreter/target is gone, but many wrappers do not produce an OS spawn errno. For example, an executable shell/npm shim using `/usr/bin/env node` with no usable `node`, or a shell wrapper that invokes a deleted target, commonly exits with `126`/`127`; `execFile` then rejects with a numeric child exit code (not `err.code`), so the current code falls through and still reports `available: true`.

Concrete fix: add a real stale-wrapper regression test (not just mocked errno objects), and classify conventional "not invocable" exits such as `126`/`127` where the version probe produced no usable version, while preserving ordinary unsupported-`--version` failures.

lefarcen (commented):
Hey @Nagendhra-web! 👋

Thanks for the fix — the core direction is sound. The tightened catch to distinguish OS-level spawn failures from adapter --version quirks matches the problem shape perfectly, and the test coverage pins each error code well. Found 2 P2 concerns around edge cases; see inline comments below.

apps/daemon/src/runtimes/detection.ts
lefarcen: **P2 — Configured-path fallback conflict**: If a user configured `CODEX_BIN` that's now stale (ENOENT), this returns `available: false` immediately — but the Settings UI's "adopt detected binary" repair flow depends on `agent.available` to enable the Test button. Marking unavailable blocks self-recovery when a valid PATH candidate exists.

**Fix**: When the selected executable fails with an OS errno (`ENOENT`/`EACCES`/`ENOTDIR`), check if it came from custom config vs PATH resolution. If a different PATH candidate exists, probe that or return a state that keeps Settings repair accessible (e.g., `available: 'fallback-candidate'` or similar).

lefarcen: **P2 — Exit-code stale wrappers not covered**: The fix handles `execFile` rejections with OS errno, but many stale wrappers exit with `126`/`127` instead. For example, a leftover shell/npm shim whose interpreter is missing can spawn successfully, then fail as "command not found" (non-zero child exit, not `err.code === 'ENOENT'`) — this still falls through to `available: true`.

**Fix**: Add a real stale-wrapper regression test (e.g., shim script with missing target) and classify conventional "not invocable" exit codes, while preserving ordinary unsupported-`--version` behavior (some adapters legitimately exit non-zero).

lefarcen (commented):
Thanks @Nagendhra-web for the fix! The Siri-Ray `.ts` → `.js` import correction looks good. I found two edge-case concerns worth addressing before merge — see inline below.

lefarcen (commented):
Nice work @Nagendhra-web — both P2 findings are fully addressed! 🎉

**What's fixed:**

1. **Configured-path fallback** — The probe now uses `inspectAgentExecutableResolution` to get both the selected path and the PATH-resolved candidate. When the selected path is not invocable, it retries against the PATH candidate before reporting unavailable. This keeps Settings' "adopt detected binary" repair flow accessible (PR #1205) when a user has a stale override but a working PATH binary.

2. **Exit-code stale wrapper coverage** — The `probeVersionAtPath` helper now classifies both string error codes (`ENOENT` / `EACCES` / `ENOTDIR`) and numeric exit codes (126 / 127) as not-invocable. Generic non-zero exits (like exit 1) preserve the legacy "available, version=null" contract so adapters with no `--version` flag are not regressed.

**Test coverage is excellent:**
- 3 OS errno tests (ENOENT / EACCES / ENOTDIR)
- 2 stale-wrapper exit tests (126 / 127)
- 1 negative-side guard (exit 1 stays available)
- 1 timeout test (ETIMEDOUT stays available)
- 3 configured-override fallback tests (stale override + working PATH / both fail / no distinct candidate)

The implementation is clean, well-documented, and the logic is straightforward. Looks good to me; deferring final approval to a maintainer.

apps/daemon/src/runtimes/detection.ts
```diff
@@ -68,15 +130,26 @@ async function probe(
...
+  // can keep `available: true` and the "Test" / "adopt detected
+  // binary" repair flow added in #1205 stays accessible. Without this
+  // retry the user is locked out of self-recovery because Settings
+  // gates the Test button on agent.available.
+  if (
+    outcome.kind === 'not-invocable'
+    && resolution.pathResolvedPath
+    && resolution.pathResolvedPath !== resolved
+  ) {
+    resolved = resolution.pathResolvedPath;
```
Siri-Ray: This fallback makes `/api/agents` report the PATH candidate as `available: true` with `path` set to `resolution.pathResolvedPath`, but the actual chat/run path does not reuse that probed executable. `startChatRun` still calls `resolveAgentBin(agentId, configuredAgentEnv)`, and `resolveAgentBin` delegates back to `resolveAgentExecutable`, whose `selectedPath` is still the configured override whenever that override file exists and is executable. So in the stale `CODEX_BIN` case covered by the new test, Settings can show Codex as available at `/usr/local/bin/codex`, while starting a normal run still spawns `/stale/custom/codex` and fails. That keeps the ghost failure alive outside the Settings test/adopt flow and breaks the existing invariant in `runtimes/resolution.ts` that spawning uses the same executable detection reported as available. Please either carry the probed fallback decision into the shared runtime resolution used by chat/routines, or avoid returning `available: true` for the fallback candidate until the configured override has actually been cleared/adopted. Add a regression that detects with a stale configured override plus working PATH binary and then verifies the normal run/bin resolution uses that same PATH binary.

Siri-Ray (commented):
I found one blocking consistency issue in the configured-override fallback: detection can now surface a PATH binary as available, but normal agent runs still resolve the stale configured override independently.


lefarcen (commented):
Excellent work addressing both P2 findings! 🎉

The configured-path fallback is now properly implemented: when a stale override exists, `probe()` calls `inspectAgentExecutableResolution` to get both the selected path and any distinct PATH candidate, retries the PATH candidate when the selected path fails with not-invocable, and reports `available: true` only when a working binary is found. The test coverage is comprehensive — 3 new tests pin the fallback flow, the no-retry guard, and the both-fail case.

The exit-code stale-wrapper classification is also solid: `probeVersionAtPath` now handles both string errno codes (ENOENT/EACCES/ENOTDIR) and numeric exit codes (126/127 for POSIX shell "not executable" / "command not found"), with tests pinning both shapes plus a negative-side guard for generic non-zero exits.

The fix is production-ready from my side.

Siri-Ray (approved):
@Nagendhra-web I reviewed the daemon runtime detection changes and the new ghost-CLI regression tests at the current head. The implementation now marks OS-level spawn failures plus 126/127 stale-wrapper exits as unavailable while preserving the existing available/version-null behavior for generic version-probe failures, and it keeps detection aligned with the same executable resolver used for chat/run spawning. Thank you for iterating on the earlier review feedback and tightening the coverage around this edge case. I was not able to rerun the targeted Vitest suite in this local checkout because vitest is not currently resolvable here, so this approval is based on the changed diff and surrounding daemon resolver context.

PR #1307, opened 2026-05-11
feat(web): Critique Theater Phase 7 — reducer + useCritiqueStream + useCritiqueReplay

Nagendhra-web:
Phase 7 of the Critique Theater rollout (spec at `specs/current/critique-theater.md`, plan at `specs/current/critique-theater-plan.md` § Phase 7). Phase 6.2 landed via #1085, so this stack is now unblocked.

Pure-TS web layer only. No daemon changes, no API contract changes, no SSE-protocol changes; everything consumes the contracts that landed with #1016 (`CritiqueRoundSummary` / `CritiqueRunStatus`) and the existing `CRITIQUE_SSE_EVENT_NAMES` wire shape from `packages/contracts/src/critique.ts`. Phases 8+ (the Theater components) plug into this reducer + the two hooks.

### 7.1 — Reducer (`apps/web/src/components/Theater/state/reducer.ts`)
Pure `reduce(state, action)` driven by `PanelEvent` (the flat contracts-level shape, not the SSE wire envelope). Both the live stream and the recorded transcript dispatch the same way, so a single reducer powers both UIs. Lifecycle:
- `idle` → `running` on `run_started`
- `running` accumulates per-round `panelist_*` events, advances `activeRound` on `round_end:continue`, and holds on `round_end:ship`
- Terminal transitions: `shipped` (artifactRef + composite), `degraded` (reason + adapter), `interrupted` (bestRound + composite), `failed` (cause)

Defensive guarantees pinned by tests:
- Terminal phases are sticky against further lifecycle events for the same run, except `parser_warning` which is informational and can land late
- A `run_started` for a new `runId` discards prior state and reboots (so consecutive runs don't need an explicit reset)
- Events whose `runId` does not match the active run return the same state reference (no React re-render churn)
- Round bookkeeping keys by round number, not "always last", so an out-of-order `panelist_dim` for round 1 arriving after a round 2 dim does not corrupt the round 2 bucket

### 7.2 — `useCritiqueStream` + SSE manager (`state/sse.ts`, `hooks/useCritiqueStream.ts`)
`createCritiqueEventsConnection` is a pure connection manager that mirrors `apps/web/src/providers/project-events.ts`: opens an `EventSource` at `/api/projects/:id/events`, subscribes to every name in `CRITIQUE_SSE_EVENT_NAMES`, decodes each frame back into a `PanelEvent` (`sseToPanelEvent` strips the `critique.` prefix and merges the data payload), and hands it to the caller's `onEvent`. Reconnect uses exponential backoff (1s → 30s) and resets on `ready`. Malformed payloads drop with a dev-mode warning rather than tearing the stream. The manager runs side-by-side with the file-events manager on the same endpoint, each listening for its own event names.

`useCritiqueStream(projectId, enabled, options)` wraps the manager in a `useReducer`. `enabled=false` or a null `projectId` tears down the connection cleanly; switching `projectId` closes the old connection and opens a fresh one. The returned `dispatch` lets local UI synthesise actions (e.g. a confirm button firing a synthetic `interrupted` while a kill request is in flight); production traffic comes from the SSE stream.

### 7.3 — `useCritiqueReplay` (`hooks/useCritiqueReplay.ts`)
Fetches the per-run NDJSON transcript, parses every line via the shared `isPanelEvent` predicate, and dispatches into the same reducer the live stream uses. `speed` knob is `'paused' | 'instant' | 'live' | { intervalMs: N }`:
- `instant` flushes every event synchronously (useful for opening a finished run already at terminal state)
- `intervalMs` paces dispatches at a fixed cadence so the reviewer can watch the run unfold
- `paused` parses the transcript but holds events back until the caller advances speed
- `live` is reserved for the future "playback at original cadence" path and currently behaves like `instant`; honest pacing needs per-event timestamps in the transcript, queued as a Phase 7+ follow-up

`fetchTranscript` and `gunzip` are injectable seams so the unit tests don't spin up a real network or gzip pipeline. The production path picks between `text` and `arrayBuffer` based on the URL extension and uses `DecompressionStream('gzip')` on `.gz` URLs.

## Test coverage (42 new tests)

- `tests/components/Theater/state/reducer.test.ts` — 18 cases for each transition, the runId guard, sticky-terminal behaviour, the out-of-order round invariant, and the stable-identity guarantee
- `tests/components/Theater/state/sse.test.ts` — 10 node-env cases for subscription set, payload decoding, malformed-frame tolerance, exponential-backoff schedule + `ready` reset, `close()` cancelling pending reconnects, and the no-op fallback when `EventSource` is unavailable
- `tests/components/Theater/hooks/useCritiqueStream.test.tsx` — 6 jsdom cases for idle pre-event, reducer driven by synthetic actions, no-connect when disabled or `projectId` is null, clean close on unmount, projectId-change reopen
- `tests/components/Theater/hooks/useCritiqueReplay.test.tsx` — 8 jsdom cases for idle without URL, `instant` flushing, `intervalMs` pacing with a `setTimeout` seam, `paused` hold, empty transcript, fetch rejection surfacing an error status, malformed-line tolerance, and the `.gz` → `gunzip` path

## Validated

- `pnpm guard` clean (including the new design-system token-fixture, A1/A2/B-slot, and parity checks)
- `pnpm --filter @open-design/web typecheck` clean
- `pnpm --filter @open-design/web exec vitest run tests/components/Theater` — 4 files, 42 tests
- Full web suite: 96 files, 883 tests

### DIFF
--- a/apps/web/src/components/Theater/hooks/useCritiqueReplay.ts
+++ b/apps/web/src/components/Theater/hooks/useCritiqueReplay.ts
@@ -1,4 +1,4 @@
-import { useEffect, useReducer, useRef } from 'react';
+import { useEffect, useReducer, useRef, useState } from 'react';
 import type { Dispatch } from 'react';
 
 import { isPanelEvent, type PanelEvent } from '@open-design/contracts/critique';
@@ -28,7 +28,8 @@ export interface UseCritiqueReplayOptions {
   /**
    * Test seam: substitute setTimeout for fake timers. Defaults to the
    * platform `setTimeout`. The hook never uses `setInterval` because the
-   * per-event delay can vary (e.g. `live` paces by recorded timestamps).
+   * per-event delay can vary in future (recorded-timestamp pacing for
+   * `live` once transcripts carry timestamps).
    */
   setTimeoutFn?: typeof setTimeout;
   clearTimeoutFn?: typeof clearTimeout;
@@ -50,14 +51,19 @@ export type ReplayStatus = 'idle' | 'loading' | 'playing' | 'done' | 'error';
  * every line into a `PanelEvent`, then dispatch each one at the cadence
  * selected by `speed`:
  *
- *  - `'instant'` — flush every event synchronously after parse, useful for
+ *  - `'instant'` — flush every remaining event synchronously, useful for
  *    test fixtures and for opening a finished run already at its terminal
  *    state.
- *  - `'live'` — pace events by their relative offset from the first event in
- *    the transcript so playback feels like the original run.
  *  - `{ intervalMs: N }` — fixed N-ms delay between events (debug mode).
+ *  - `'live'` — placeholder for the future "playback at the original
+ *    recorded cadence" path. Current transcripts don't carry per-event
+ *    timestamps, so for now this behaves like `{ intervalMs: 0 }` (events
+ *    drained on successive microtasks via setTimeoutFn). The honest
+ *    timestamp-driven implementation is queued as a Phase 7+ follow-up.
  *  - `'paused'` — load the transcript but hold every event until the speed
- *    is changed.
+ *    is changed to one of the values above; the cursor is preserved
+ *    across pause/resume cycles so flipping to `instant` flushes the
+ *    remaining events from wherever playback was suspended.
  *
  * The hook owns its own reducer (separate from `useCritiqueStream`) so a UI
  * can mount both in parallel: live next to a replay of a prior run.
@@ -69,31 +75,34 @@ export function useCritiqueReplay(
 ): UseCritiqueReplayResult {
   const [state, dispatch] = useReducer(reduce, initialState);
   const [meta, setMeta] = useStableMeta();
+  const [events, setEvents] = useState<PanelEvent[] | null>(null);
 
   const dispatchRef = useRef(dispatch);
   useEffect(() => {
     dispatchRef.current = dispatch;
   }, [dispatch]);
 
-  // Capture the latest speed without re-running the load effect — we want
-  // changing speed to retime in-flight playback, not refetch the transcript.
-  const speedRef = useRef(speed);
-  speedRef.current = speed;
+  // Playback cursor lives in a ref so pause -> resume picks up where the
+  // last tick left off instead of replaying from the start. A fresh
+  // transcript URL resets it to zero (see the parse effect below).
+  const cursorRef = useRef(0);
 
+  // Parse effect: own the fetch + parse. Runs only when the URL changes.
+  // Stores the parsed events in component state so the pace effect can
+  // react to them; cursor is reset because a new transcript is a new run.
   useEffect(() => {
     if (!transcriptUrl) {
       setMeta({ status: 'idle', error: null });
+      setEvents(null);
+      cursorRef.current = 0;
       return;
     }
-
     let cancelled = false;
     const fetcher = options.fetchTranscript ?? defaultFetch;
     const gunzip = options.gunzip ?? defaultGunzip;
-    const setT = options.setTimeoutFn ?? setTimeout;
-    const clearT = options.clearTimeoutFn ?? clearTimeout;
-    const timers: Array<ReturnType<typeof setTimeout>> = [];
-
     setMeta({ status: 'loading', error: null });
+    cursorRef.current = 0;
+    setEvents(null);
 
     (async () => {
       let raw: string;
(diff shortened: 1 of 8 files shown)


apps/web/src/components/Theater/hooks/useCritiqueReplay.ts
```diff
@@ -0,0 +1,215 @@
...
+
+      setMeta({ status: 'playing', error: null });
+
+      const currentSpeed = speedRef.current;
+      if (currentSpeed === 'instant') {
+        for (const evt of events) dispatchRef.current(evt);
+        setMeta({ status: 'done', error: null });
+        return;
+      }
+      if (currentSpeed === 'paused') {
```
lefarcen: **P1 — Paused replay can't resume**

Starting with `speed="paused"` parses the transcript and returns without retaining events or scheduling future work. Because the effect only depends on `transcriptUrl`, changing `speed` to `instant`/`intervalMs` later does not dispatch the held events, breaking the documented pause-then-advance behavior.

Fix: add `speed` to the effect dependency array and retain parsed events in state so they're available when speed changes.

apps/web/src/components/Theater/state/sse.ts
```diff
@@ -0,0 +1,122 @@
...
+/**
+ * Lift an SSE-wire `CritiqueSseEvent` back into the flat `PanelEvent` shape
+ * the reducer consumes. The daemon emits one channel per event name with
+ * the payload (sans `type`) as JSON; this is the inverse of
+ * `panelEventToSse` in the contracts package.
+ */
+export function sseToPanelEvent(eventName: CritiqueSseEventName, data: unknown): PanelEvent | null {
+  if (data === null || typeof data !== 'object') return null;
+  const type = eventName.slice('critique.'.length) as PanelEvent['type'];
+  return { type, ...(data as Record<string, unknown>) } as PanelEvent;
```
lefarcen: **P1 — SSE payload override vulnerability**

This reconstructs a `PanelEvent` by casting unknown JSON. Because the spread comes after `type`, a payload-provided `type` can override the SSE channel-derived type. A malformed or compromised SSE frame can dispatch a different action shape into the reducer.

Fix: rebuild with `type` last and validate/drop the resulting action instead of blindly casting. Add tests for `data.type` override and missing required fields.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for Phase 7! The reducer shape and SSE manager are solid foundations, and the test coverage is impressive (42 tests verified). Found 2 functional bugs that need fixing before merge:

1. **Replay pause/resume broken** — `useCritiqueReplay` with `speed="paused"` parses the transcript but then returns without retaining events or scheduling future work. Changing `speed` to `instant` later doesn't dispatch the held events because the effect only depends on `transcriptUrl`. Tests only check that paused holds, not that it can resume. Fix: add `speed` to the effect dependency array and retain parsed events in state.

2. **SSE payload validation gap** — `sseToPanelEvent` casts unknown JSON to `PanelEvent` without re-validating, and the spread order lets `data.type` override the channel-derived `type`. A malformed SSE frame could dispatch a wrong action shape. Fix: put `type` last in the spread and add runtime validation (e.g., check required fields or use a predicate).

3. **Design doc inconsistency** — PR body says `live` speed "behaves like instant" (queued for Phase 7+ follow-up), but the source comment claims "paced by recorded timestamps" and the impl uses 0ms timers. These don't align. Clarify the implemented semantics vs the future contract.

See inline comments below.

lefarcen (approved):
Both P1s fixed, great work @Nagendhra-web! 🎉

**Verified fixes:**

1. **Replay pause/resume** — Now uses two effects: parse effect (deps: `[transcriptUrl]`) owns fetch/parse and writes events to state; pace effect (deps: `[events, speed]`) owns cursor + timers. Cursor ref survives pause cycles, so `paused` → `instant` flushes from the current position without re-starting. The two new replay tests pin this.

2. **SSE payload validation** — `type` now comes AFTER `...data` in the spread, making the channel-derived value authoritative. The reconstructed object goes through `isPanelEvent` predicate before returning. Malformed frames (missing runId, empty runId, unknown type) are dropped. The three new sse tests pin the regression.

3. **Doc alignment** — `live` speed is now honestly documented as `{intervalMs: 0}` (microtask drain), not "paced by timestamps". Clear that recorded-cadence pacing is queued for Phase 7+.

**One non-blocking note (P2):**

- `speed` object identity: the pace effect at line 207 compares `speed` by reference. If a caller passes `{intervalMs: N}` inline, the hook's own state updates can recreate the object, rerun the effect, cancel the timer, and synchronously advance the next event instead of honoring the interval. Consider depending on primitive speed values (e.g., `speed.intervalMs`) or memoizing the speed object. This is edge-case brittleness, not a blocker — most callers will use string literals (`"paused"` / `"instant"`), and the inline `{intervalMs}` pattern is rare in practice.

LGTM for merge. Thanks for the thorough fixes!

PR #1308, opened 2026-05-11
feat(web): add thumbs-up/down feedback widget under completed assistant turns (#1288)

Nagendhra-web:
Closes #1288.

Adds a lightweight feedback widget that surfaces under each assistant turn whose run succeeded. Users can submit positive or negative feedback in one click; the negative path opens an optional free-text comment area. The widget never blocks the message composer and only mounts after the run has produced its final artifact.

## What ships

- **`<MessageFeedback>`** in `apps/web/src/components/MessageFeedback.tsx` renders the three states:
  - **Idle**: helpful-prompt copy + thumbs up/down buttons
  - **Submitted positive**: confirmation chip + Change button
  - **Submitted negative**: confirmation chip + optional comment textarea + Send + Change
- **`AssistantMessage.tsx`** slots the widget under `AssistantFooter`, gated on `runSucceeded && !hasEmptyResponse` so failed/empty turns don't ask the user to rate something that never finished.

## Persistence (v1 = localStorage)

@lefarcen's clarifying comment on the issue asked whether v1 should be daemon-persisted or in-memory while the analytics pipeline is defined. The daemon's `messages` table is column-strict, so daemon persistence would require a SQLite migration plus a contract bump on `ChatMessage`; locking that shape in before the analytics pipeline is designed risks reworking it twice.

localStorage is the middle ground: feedback survives reload (so the "feedback state is visually clear after submission" criterion holds across sessions) without committing the wire shape. The hook surface is just `(value, setter)`, so a future PR can swap the storage layer for a daemon mirror or an analytics shipper without touching the React surface. The store handles corrupted JSON, unknown future rating values, disabled storage (private-mode browsers), and broadcasts changes across listeners in the same tab via a CustomEvent so two mounts of the hook for the same messageId stay in sync.

## Acceptance criteria

| Criterion | Status |
|---|---|
| After an Agent final artifact is completed, the chat pane displays a feedback area near the final output/completion state | Mounted under `AssistantFooter`, gated on `runSucceeded` |
| The user can submit positive feedback | Thumb-up button writes `{ rating: 'positive', submittedAt }` |
| The user can submit negative feedback | Thumb-down button writes `{ rating: 'negative', submittedAt }` |
| The feedback state is visually clear after submission | Confirmation chip + state-specific `data-state` / `data-rating` attributes; pinned by a rehydration-from-storage test |
| The feedback UI does not interfere with continuing the conversation | Sits below the footer, inside `.assistant-flow`; no composer-blocking overlays |
| Implementation leaves room for future feedback metadata | `MessageFeedback` shape is `{ rating, comment?, submittedAt }`; `runId` / `agentId` / `artifactRef` derivable from the surrounding message |

## i18n

11 new keys under `feedback.*` (prompt, thumbsUp/Down, two confirmation chips, comment label/placeholder/submit/saved, change). English authored alongside the keys; zh-CN translations added in the same pass so the locale alignment test stays green and Chinese users see Chinese strings from day one. The other 16 locales pick up English fallbacks via their existing `...en` spread.

## Test coverage

- `tests/state/message-feedback.test.ts` (8 jsdom cases): round-trip positive + negative, null-clear, corrupted JSON tolerance, missing rating, unknown future rating, key collision across messages, null on empty storage.
- `tests/components/MessageFeedback.test.tsx` (7 jsdom cases): idle state, positive submit, negative submit, comment save with the saved confirmation, blank-comment Send disabled, Change unsticks the rating, rehydration from pre-populated storage.
- `tests/i18n/locales.test.ts` (5/5) continues to enforce that every locale declares the new keys.

## Validated

- `pnpm guard` clean
- `pnpm --filter @open-design/web typecheck` clean
- `tests/i18n/locales.test.ts` 5/5 across 18 locales
- `tests/state/message-feedback.test.ts` 8/8
- `tests/components/MessageFeedback.test.tsx` 7/7
- Full web suite: 98 files, 903 tests

## Open questions parked for follow-up

- **Should feedback be reported to an analytics pipeline?** Out of scope for v1 — the hook contract is stable enough to wrap an analytics shipper without touching the component.
- **Should negative feedback ask for a reason immediately?** This PR opens the comment textarea on negative click but does not require a comment. The Send button is disabled while the textarea is blank, so an empty submit is a no-op.
- **Should the feedback be tied to the message / artifact / task / run?** Currently keyed by `messageId`; the surrounding `ChatMessage` already has `runId`, `agentId`, and `producedFiles`, so any of those is one prop away whenever the analytics design lands.

### DIFF
--- a/apps/web/src/components/AssistantMessage.tsx
+++ b/apps/web/src/components/AssistantMessage.tsx
@@ -165,7 +165,13 @@ export function AssistantMessage({
           hasUnfinishedTodos={unfinishedTodos.length > 0}
           hasEmptyResponse={hasEmptyResponse}
         />
-        {runSucceeded && !hasEmptyResponse ? (
+        {/* Feedback widget for issue #1288 — gated on `produced.length > 0`
+            because the issue scopes feedback to turns that produce a final
+            artifact, not text-only acknowledgements or question-form turns
+            (lefarcen review on PR #1308). The `runSucceeded`
+            guard keeps it off failed runs, and `!hasEmptyResponse` keeps it
+            off agents that succeeded silently with no content. */}
+        {runSucceeded && !hasEmptyResponse && produced.length > 0 ? (
           <MessageFeedback messageId={message.id} />
         ) : null}
       </div>
--- a/apps/web/src/components/AssistantMessage.tsx
+++ b/apps/web/src/components/AssistantMessage.tsx
@@ -9,6 +9,7 @@ import {
 import { stripArtifact } from "../artifacts/strip";
 import { QuestionFormView, parseSubmittedAnswers } from "./QuestionForm";
 import { Icon } from "./Icon";
+import { MessageFeedback } from "./MessageFeedback";
 import { useT } from "../i18n";
 import { unfinishedTodosFromEvents, type TodoItem } from "../runtime/todos";
 import type { Dict } from "../i18n/types";
@@ -164,6 +165,9 @@ export function AssistantMessage({
           hasUnfinishedTodos={unfinishedTodos.length > 0}
           hasEmptyResponse={hasEmptyResponse}
         />
+        {runSucceeded && !hasEmptyResponse ? (
+          <MessageFeedback messageId={message.id} />
+        ) : null}
       </div>
     </div>
   );
--- a/apps/web/src/components/MessageFeedback.tsx
+++ b/apps/web/src/components/MessageFeedback.tsx
@@ -1,0 +1,164 @@
+import { useState } from 'react';
+
+import { useT } from '../i18n';
+import {
+  useMessageFeedback,
+  type FeedbackRating,
+  type MessageFeedback as MessageFeedbackValue,
+} from '../state/message-feedback';
+
+interface Props {
+  messageId: string;
+  /**
+   * Test seam: drop in a deterministic `Date.now()` so snapshot timing
+   * isn't sensitive to wall-clock. Defaults to the real clock.
+   */
+  now?: () => number;
+}
+
+/**
+ * Lightweight feedback widget rendered under each completed assistant
+ * turn (issue #1288). The visibility gate (only after a successful
+ * run) is the caller's responsibility — this component assumes it's
+ * mounted at the right moment and always renders something.
+ *
+ * Lifecycle:
+ *
+ *   - **Idle.** Shows the helpful-prompt copy and two thumb buttons.
+ *     Clicking either persists the rating immediately and flips to the
+ *     submitted state.
+ *
+ *   - **Submitted positive.** Confirmation chip plus a Change button
+ *     that clears the rating so the user can re-rate. Issue Open
+ *     Question 2 ("should users be able to change feedback after
+ *     submitting it?") — yes, the lightweight thing is to allow it.
+ *
+ *   - **Submitted negative.** Same confirmation chip plus an optional
+ *     comment textarea. The comment is persisted on Send, so the
+ *     reducer-style record is { rating: 'negative', comment, submittedAt }.
+ *     A blank submit just confirms the comment area was acknowledged
+ *     without changing the recorded value, matching the issue's
+ *     "Optional follow-up reason or comment after negative feedback".
+ *
+ * Persistence is handled by `useMessageFeedback` (localStorage in v1;
+ * see `state/message-feedback.ts` for the rationale).
+ */
+export function MessageFeedback({ messageId, now = Date.now }: Props) {
+  const t = useT();
+  const [feedback, setFeedback] = useMessageFeedback(messageId);
+  const [draftComment, setDraftComment] = useState('');
+  const [commentJustSaved, setCommentJustSaved] = useState(false);
+
+  const submitRating = (rating: FeedbackRating) => {
+    const submittedAt = now();
+    setFeedback({ rating, submittedAt });
+    setDraftComment('');
+    setCommentJustSaved(false);
+  };
+
+  const submitComment = () => {
+    if (!feedback) return;
+    const comment = draftComment.trim();
+    const next: MessageFeedbackValue = {
+      ...feedback,
+      comment: comment || undefined,
+      submittedAt: feedback.submittedAt,
+    };
+    setFeedback(next);
+    setCommentJustSaved(true);
+  };
+
+  const clearFeedback = () => {
+    setFeedback(null);
+    setDraftComment('');
+    setCommentJustSaved(false);
+  };
+
+  if (!feedback) {
+    return (
+      <div className="message-feedback" data-state="idle">
+        <span className="message-feedback-prompt">{t('feedback.prompt')}</span>
+        <div className="message-feedback-actions">
+          <button
+            type="button"
+            className="message-feedback-button"
+            aria-label={t('feedback.thumbsUp')}
+            title={t('feedback.thumbsUp')}
+            onClick={() => submitRating('positive')}
+            data-testid="message-feedback-positive"
+          >
+            <span aria-hidden>👍</span>
+          </button>
+          <button
+            type="button"
+            className="message-feedback-button"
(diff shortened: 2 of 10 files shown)


apps/web/src/state/message-feedback.ts
```diff
@@ -0,0 +1,165 @@
...
+export function writeMessageFeedback(
+  messageId: string,
+  feedback: MessageFeedback | null,
+): void {
+  const w = safeWindow();
+  if (!w) return;
+  try {
+    if (feedback === null) {
+      w.localStorage.removeItem(storageKey(messageId));
+      return;
```
Siri-Ray: The `feedback === null` path removes the storage entry and then returns before the custom `open-design:message-feedback` event is dispatched. That means another mounted `useMessageFeedback(messageId)` in the same tab will not reload, because the native `storage` event only fires in other browsing contexts. The comment below this block and the hook contract both say same-tab mounts stay in sync, so clearing via the Change button in one mount can leave the other mount showing the submitted state. Move the dispatch after both remove/set operations, or avoid the early return and dispatch once only when the storage operation succeeds; then add a hook/component test with two mounts where clearing one returns both to idle.

Siri-Ray (commented):
I found one non-blocking issue in the feedback hook sync path. The rest of the changed ranges looked consistent with the PR goal and current web test layout.


apps/web/src/components/AssistantMessage.tsx
```diff
@@ -164,6 +165,9 @@ export function AssistantMessage({
           hasUnfinishedTodos={unfinishedTodos.length > 0}
           hasEmptyResponse={hasEmptyResponse}
         />
+        {runSucceeded && !hasEmptyResponse ? (
```
lefarcen: **P2** This mounts feedback for every successful assistant turn, not just turns that produced the final artifact. The linked issue and PR body both scope the widget to the completed final artifact/output state, but `runSucceeded && !hasEmptyResponse` also matches successful discovery/question-form turns, plain text acknowledgements, and legacy messages with only `endedAt`. Gate this on an explicit produced-output signal, e.g. `produced.length > 0` or the same artifact/final-output invariant the app uses elsewhere, and add an `AssistantMessage` regression test for a successful no-artifact turn.

apps/web/src/state/message-feedback.ts
```diff
@@ -0,0 +1,165 @@
...
+  } catch {
+    // Storage quota / disabled storage / private-mode rejection: the
+    // UI keeps the in-memory state so the user still sees a confirmation
+    // for this session.
+  }
+  // Broadcast across listeners in the same tab (the storage event only
+  // fires for OTHER tabs); use a custom event so multiple mounted
+  // `useMessageFeedback` hooks for the same messageId stay in sync.
+  try {
+    w.dispatchEvent(new CustomEvent('open-design:message-feedback', {
```
lefarcen: **P2** Storage failures do not preserve the in-memory confirmation as the comment claims. If `localStorage.setItem` throws, `writeMessageFeedback` still dispatches the custom event; the current hook hears it synchronously, calls `readMessageFeedback`, gets `null`, and overwrites the `setValue(next)` from the setter, so disabled/quota-full storage leaves the widget in the idle state instead of confirming the click. Only broadcast after a successful storage mutation, or include the new value in the custom event and have same-tab listeners apply that value instead of re-reading failed storage.

apps/web/src/state/message-feedback.ts
```diff
@@ -0,0 +1,165 @@
...
+ * the user can re-rate.
+ */
+export function writeMessageFeedback(
+  messageId: string,
+  feedback: MessageFeedback | null,
+): void {
+  const w = safeWindow();
+  if (!w) return;
+  try {
+    if (feedback === null) {
```
lefarcen, line 118: **P2** Clearing feedback never notifies other same-tab hook instances because the successful `removeItem` path returns before dispatching the custom event. That means two mounted views of the same message stay inconsistent after clicking Change, despite the hook contract and PR body promising same-tab synchronization. Remove the early return and dispatch after successful `removeItem`, ideally with the same event-value fix as the write path.

apps/web/src/components/MessageFeedback.tsx
```diff
@@ -0,0 +1,164 @@
...
+        {t(confirmationKey)}
+      </span>
+      {feedback.rating === 'negative' ? (
+        <div className="message-feedback-comment">
+          <label className="message-feedback-comment-label">
+            {t('feedback.commentLabel')}
+            <textarea
+              className="message-feedback-comment-input"
+              placeholder={t('feedback.commentPlaceholder')}
+              value={draftComment || feedback.comment || ''}
```
lefarcen: **P3** The textarea is controlled with `draftComment || feedback.comment || ''`, so once a saved comment exists the user cannot clear the field normally: deleting the last character makes `draftComment` empty and the value snaps back to `feedback.comment`. Track whether the draft has been initialized separately, for example initialize local draft state from `feedback.comment ?? ''` when entering the negative state and render `value={draftComment}` so empty is a valid user edit.

apps/web/src/components/MessageFeedback.tsx
```diff
@@ -0,0 +1,164 @@
...
+
+  const clearFeedback = () => {
+    setFeedback(null);
+    setDraftComment('');
+    setCommentJustSaved(false);
+  };
+
+  if (!feedback) {
+    return (
+      <div className="message-feedback" data-state="idle">
```
lefarcen, line 105: **P3** The new widget emits a full set of `.message-feedback*` classes, but there are no matching styles in `apps/web/src/index.css`, so it will render as default browser buttons/textarea inside the assistant flow rather than the compact confirmation chip described by the PR and acceptance criteria. Add styles near the assistant footer section for idle/submitted states, button sizing, textarea width, and mobile wrapping, then cover at least the structural class/state in tests or a visual smoke path.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for the PR! 👋

This is close, but I found functional gaps around when the widget appears and the localStorage hook's failure/sync behavior that should be fixed before merge. See inline comments below.

## Code review (Lens A)

Applied code-correctness lens to all changed files (code/tests/i18n). No changed rationale/spec files for Lens B.

**Overall:** The feature implementation is mostly solid, but the widget visibility gate is too broad (it mounts on every successful turn, not just final artifacts) and the localStorage hook has synchronization bugs on storage failure + clear operations that break the same-tab sync contract. The P2s are fixable in a quick iteration; the P3s are polish.

## Minor suggestions (non-blocking)

See inline below.

apps/web/src/components/MessageFeedback.tsx
```diff
@@ -0,0 +1,205 @@
...
+
+  // Re-seed draftComment whenever the rating itself transitions
+  // (e.g. clearFeedback -> null -> negative again, or a cross-tab
+  // update flips us into a different state). The dependency is
+  // `feedback?.rating` specifically — NOT `feedback?.comment` —
+  // because once the user types into the textarea we must not
+  // override their draft with a stale saved comment. Cleared
+  // comments (user erased the textarea then hit Send) deliberately
+  // surface as an empty draft.
+  useEffect(() => {
```
lefarcen, line 70: P3: This effect reconciles `rating` transitions but not `comment`-only changes, so a second mount for the same message won't see a comment edit until the rating changes. Since the hook broadcasts the full value now, the component could reconcile `feedback.comment !== draftComment` when the local textarea isn't dirty. Non-blocking — the widget works as described in the PR body; this is an edge case for multi-mount sync.

lefarcen (approved):
Great fixes @Nagendhra-web! 🎉

**Prior P2 blockers resolved:**
- Widget gate now requires `producedFiles.length > 0` ✅
- Storage hook broadcasts same-tab writes/clears with the new value ✅

Found one non-blocking sync gap — see inline. Feel free to address in a follow-up or leave as-is; the core path is solid.

PR #1309, opened 2026-05-11
feat(daemon): user-configurable model alias for the media dispatcher (#1277)

### LINKED ISSUE #1277: Feature Request: Add model alias/redirect support (e.g., map old model name to a new one)
**Problem Description**  
The project currently hardcodes (or expects) the model name `doubao-seedream-3-0-t2i-250415`, but the actual model I have access to is **`doubao-seedream-5-0`** (a newer version). Because of this mismatch, the project fails to call my model – every request ends with a "model not found" or invalid model name error.

**Desired Feature**  
I would like the project to support a **model alias / redirect mechanism** that allows users to define custom name mappings in a configuration file or environment variables. For example:

```json
{
  "model_alias": {
    "doubao-seedream-3-0-t2i-250415": "doubao-seedream-5-0"
  }
}
```

When the project code requests `doubao-seedream-3-0-t2i-250415`, the actual model name sent to the API becomes `doubao-seedream-5-0` (without modifying the core code).

**Why This Is Needed**  
- Model names change over time (e.g., `3-0` → `5-0`).  
- Different users may use different versions or custom deployment names of the same underlying model.  
- Hardcoding model names forces users to patch the source code every time they pull an update, which is not sustainable.

**My Specific Use Case**  
- Default model name expected by the project: `doubao-seedream-3-0-t2i-250415`  
- Actual model name I can use: `doubao-seedream-5-0`  
- Without redirect: every call fails.  
- With redirect: I can add a one-line mapping and everything works.

**Current Workaround (not ideal)**  
I manually change the model name constant in the source code, but I have to redo it after every project update.

**Suggested Implementation**  
- Environment variables: `MODEL_ALIAS_OLD_NAME=new_name`  
- Or a simple config file (`config.json`, `.env`, etc.) with an alias dictionary  
- Priority: alias mapping → original model name (fallback if no alias matches)

**Environment**  
- Project version: main / latest  
- Related API: Doubao (seedream) image generation  
- OS / calling method: any (not specific)

Thank you for considering this small but very useful feature for model version flexibility.

Nagendhra-web:
Closes #1277.

Tilmirs's use case: Doubao access has moved from `doubao-seedream-3-0-t2i-250415` to `doubao-seedream-5-0`, but the catalog still emits the old id, so every call fails with "model not found". Until now the only workaround was patching the source on each update.

This adds a user-configurable alias layer that swaps the catalog id for whatever wire-name the provider expects, without changing the catalog itself. Two storage layers, env wins over disk, mirroring the existing `apiKey` precedence:

## Storage

**Environment variable** — `OD_MEDIA_MODEL_ALIASES` carries the full map as JSON:
```
OD_MEDIA_MODEL_ALIASES='{"doubao-seedream-3-0-t2i-250415":"doubao-seedream-5-0"}'
```
Single var keeps it portable across shells. The per-id pattern @lefarcen mentioned (`OD_MODEL_ALIAS_<id>=...`) would have broken on Windows cmd.exe (rejects hyphens in env-var names) and on POSIX-strict shells. Malformed JSON is tolerated and falls through to the on-disk map rather than throwing mid-generation.

**`media-config.json`** — top-level `aliases` field:
```json
{
  "providers": { ... },
  "aliases": {
    "doubao-seedream-3-0-t2i-250415": "doubao-seedream-5-0"
  }
}
```
Settings' existing PUT writes providers only. The `writeStored` path now reads the existing aliases and preserves them on every write — otherwise a Settings save would silently wipe the user's aliases. The Settings UI surface for editing aliases is a separate follow-up; manual JSON edit + the env var are the v1 entry points.

## Resolution

Happens inside `startMediaGeneration` after the catalog lookup and surface validation have already accepted the registered id. Users still get the "unknown model" error if they request a catalog id that doesn't exist; the alias swap only changes what the provider receives on the wire. Every dispatcher (`openai`, `volcengine`, `grok`, `fal`, `nanobanana`, etc.) passes `ctx.model` straight into its request body, so the swap is one-line.

`autoOutputName` and the file-naming side keep using the catalog id (the function-level `model` parameter), so a generated PNG named after `doubao-seedream-3-0-t2i-250415` keeps surfacing the id the agent / CLI asked for, not the wire-level alias. `providerNote` strings include the wire name so the user can see what was actually sent.

## Public API additions

- `resolveModelAlias(projectRoot, modelId)` — returns the wire name (or the original if no alias matches).
- `readAliasMap(projectRoot)` — `{ effective, env, stored }` for the future Settings UI's source-attribution display.

## Tests

8 new cases in `tests/media-config.test.ts` (suite goes from 14 to 22):

- Pass-through when no alias is configured
- Stored map redirect (the issue's flagship use case)
- Env map redirect
- Env-over-stored precedence
- Malformed env JSON tolerated, falls through to stored
- Coercion drops bad entries (null / number / nested object / empty string / blank key)
- `readAliasMap` surfaces both layers separately for source attribution
- `writeConfig` regression: a Settings-style provider PUT preserves an existing alias map

## Acceptance criteria from the issue

| Criterion | Status |
|---|---|
| Custom name mappings via configuration file | `aliases` field in `media-config.json` |
| Custom name mappings via environment variables | `OD_MEDIA_MODEL_ALIASES` JSON map |
| Project code requests catalog id, API gets aliased name | `ctx.model = wireModel` after the catalog check |
| Priority: alias mapping → original (fallback) | `resolveModelAlias` returns the original when no entry matches |
| Doesn't require modifying core code on every update | Manual JSON edit or env var; survives Settings PUTs |

## Validated

- `pnpm guard` clean
- `pnpm --filter @open-design/daemon typecheck` clean (both `tsconfig.json` and `tsconfig.tests.json`)
- Media test suite (`media-config` + `media-tasks-routes` + `media-tasks-persistence` + `media-nanobanana`): 33/33 pass

Pre-existing Windows-only daemon test failures (symlinks, CODEX_BIN runtime resolution, MCP config, skills, server-paths) are unrelated to this change and reproduce on a clean main checkout.

## Out of scope (queued for follow-up)

- Settings UI surface for editing aliases — current v1 is manual JSON / env var. The contract (`readAliasMap` exposing env vs stored attribution) is already stable enough for the UI PR to consume without touching the daemon.
- Validation that the alias target is a real model id at the provider. The error will surface at the provider's "unknown model" response if mistyped, with the wire name visible in `providerNote`.

### DIFF
--- a/apps/daemon/src/media-config.ts
+++ b/apps/daemon/src/media-config.ts
@@ -44,8 +44,19 @@ import { expandHomePrefix } from './home-expansion.js';
 const PROVIDER_IDS = MEDIA_PROVIDERS.map((p) => p.id);
 type ProviderEntry = { apiKey?: string; baseUrl?: string; model?: string };
 type ProviderMap = Record<string, ProviderEntry>;
+type ModelAliasMap = Record<string, string>;
 type JsonRecord = Record<string, unknown>;
 type OAuthCredential = { apiKey: string; source: string };
+
+// Single env var carries the full alias map as JSON so we don't have
+// to dynamically lift `OD_MEDIA_MODEL_ALIAS_<id>=value` into a record
+// with all the env-var-name escaping that entails (Windows cmd.exe in
+// particular rejects hyphens). The shape mirrors the on-disk
+// `aliases` map so users can switch storage layers without rewriting
+// their workflow:
+//
+//   OD_MEDIA_MODEL_ALIASES='{"doubao-seedream-3-0-t2i-250415":"doubao-seedream-5-0"}'
+const ENV_MODEL_ALIASES = 'OD_MEDIA_MODEL_ALIASES';
 
 function isRecord(value: unknown): value is JsonRecord {
   return value !== null && typeof value === 'object';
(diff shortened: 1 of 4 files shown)


lefarcen (commented):
Hey @Nagendhra-web, thanks for the implementation! The model alias layer is exactly what #1277 needs.

I found 2 issues worth addressing before merge — one about request shaping preservation, one about API surface vs the PR's claims. See inline comments below.

apps/daemon/src/media.ts
```diff
@@ -341,9 +341,20 @@ export async function generateMedia(args: {
   // and decide how to splice the data URL into their request.
   const imageRef = await resolveProjectImage(image, dir);
 
+  // Resolve any user-configured model alias BEFORE we hand the id to a
+  // dispatcher (issue #1277). Catalog lookup + surface validation above
+  // ran against the original id so we still enforce the registered
+  // catalog; this swap only changes what the provider receives on the
+  // wire, which is exactly the redirect the issue asks for. Pass-
+  // through when no alias matches.
+  const wireModel = await resolveModelAlias(projectRoot, model);
```
lefarcen, line 374: **P2** — Alias breaks model-specific request shaping

`resolveModelAlias()` replaces `ctx.model` globally, but renderers use `ctx.model` for both the provider wire name and local model-family behavior:

- Line 350: `gpt-image-*` aliases lose the wide aspect sizing path
- Line 592-602: `dall-e-3` aliases skip `response_format`/quality handling  
- Line 793: `gpt-4o-mini-tts` aliases drop style `instructions`

This means aliasing `dall-e-3` → `gpt-4o-flux` would incorrectly send DALL-E params to Flux.

**Concrete fix:**
Keep separate values in `MediaContext`:
- `catalogModel: model` (registered id for capability branches)
- `wireModel` (alias result for provider `model` fields/URLs)

Use `wireModel` only for provider requests + providerNote; use `catalogModel` for sizing/quality/TTS instruction logic.

Add a test that aliases an OpenAI image/TTS model and asserts request shaping still follows the registered model.

apps/daemon/src/media-config.ts
```diff
@@ -128,24 +139,106 @@ function configFile(projectRoot: string): string {
...
+  modelId: string,
+): Promise<string> {
+  const envAliases = readEnvAliases();
+  if (envAliases[modelId]) return envAliases[modelId]!;
+  const stored = await readStoredAliases(projectRoot);
+  return stored[modelId] ?? modelId;
+}
+
+/**
+ * Read the merged alias map (env + stored). Exposed for the
```
lefarcen, line 231: **P3** — PR claims `readAliasMap()` is daemon-public for Settings UI, but route doesn't expose it

The PR description says: "The contract (`readAliasMap` exposing env vs stored attribution) is already stable enough for the UI PR to consume without touching the daemon."

But `media-routes.ts:32` still returns only `readMaskedConfig()`, which has no `aliases` field. Settings UI can't consume `readAliasMap()` unless the HTTP/API surface includes it.

**Concrete fix:**
Either:
1. Return `{ providers, aliases }` from `readMaskedConfig()`/GET now, OR
2. Adjust the PR text to mark the HTTP/API surface as follow-up (not claiming UI can consume without daemon changes).

lefarcen (commented):
Nice work @Nagendhra-web! Both issues from the previous review are now resolved:

✅ **P2 fixed** — MediaContext split into `model` (catalog id for capability branches) and `wireModel` (post-alias for provider requests). Verified in media.ts:68-92, 374-379, 613-628, 815-818. The new test in `media-alias-capability.test.ts` confirms aliased dall-e-3 and gpt-4o-mini-tts retain their capability-specific params.

✅ **P3 fixed** — `readMaskedConfig` now returns `aliases: {effective, env, stored}` in the GET response (media-config.ts:356-364, media-routes.ts:32-35). Settings UI can consume it without daemon changes.

PerishCode (approved):
@Nagendhra-web nice work landing this — the split between `ctx.model` (catalog id) and `ctx.wireModel` (post-alias) cleanly fixes the P2 regression that codex + lefarcen flagged. I walked the renderers end-to-end and confirmed every capability branch (`startsWith('dall-e-')`, `=== 'dall-e-3'`, `=== 'gpt-4o-mini-tts'`, `openaiSizeFor(ctx.model, …)` in both openai-image and volcengine-image, MINIMAX/FISHAUDIO TTS maps) keys off the catalog id, while `body.model = ctx.wireModel` and every `providerNote` carries the wire name so users see what was actually sent. Stub diagnostics keeping `ctx.model` is the right call — debug surface, not a provider call. The MINIMAX/FISHAUDIO precedence (`user alias > legacy rename map > catalog id`) reads cleanly via `ctx.wireModel !== ctx.model`. The P3 follow-up wiring `aliases: { effective, env, stored }` into `readMaskedConfig` gives the future Settings UI a stable, source-attributed shape with empty-state defaults. Test coverage for the regression is targeted and right-sized: `media-alias-capability.test.ts` pins both the dall-e-3 hd-quality + response_format flow and the gpt-4o-mini-tts instructions branch through a stubbed fetch that asserts on the actual request body, and the new `readMaskedConfig` cases pin both the populated and empty alias shapes. Appreciate how carefully you handled the alias-preservation contract in `writeStored` too — the comment about Settings PUTs silently wiping aliases is exactly the kind of trap worth pinning, and the regression test makes that promise enforceable. Ship it!

PR #384, opened 2026-05-03
fix(web): URL-load HTML preview iframes by default; ?forceInline=1 opt-out

bankielewicz:
Part **1 of 2** for #368.

## Why

`FileViewer` currently inlines every relative `<link>` and `<script src>` of an HTML preview into a single `srcDoc`. That works for self-contained 1-file decks, but for multi-file projects (e.g. the AlphaTraceAI 50-file React + inline-Babel prototype that motivated #336/#368) it produces three structural problems even after the recent `$'`-safe replacer fix in PR #343:

1. **All-or-nothing failure.** A single fetch failure or HTML-parse-hostile character anywhere in the concatenated string kills the whole iframe.
2. **No source maps.** Every script appears as `Inline Babel script (n)` in DevTools — no filenames, no usable stack traces.
3. **Doubled bytes.** Every Open re-fetches and re-inlines the entire asset tree; per-asset HTTP caching does not apply.

The `Preview` pane already URL-loads via the daemon's `/raw/` endpoint and does not have these problems. This PR brings `FileViewer` in line.

## What

- `FileViewer.tsx` HTML previews now render with `<iframe src="/api/projects/:id/raw/:file?v=<mtime>&r=<reloadKey>" sandbox="allow-scripts">` by default.
- `?forceInline=1` query (on the host page) keeps the legacy inlined-`srcDoc` path. Decks (auto-detected by shape) and comment-mode also stay on `srcDoc` because they need host bridges injected by `buildSrcdoc` (postMessage, comment targeting, `localStorage` shim).
- Decision logic extracted to `apps/web/src/components/file-viewer-render-mode.ts` (`shouldUrlLoadHtmlPreview`, `parseForceInline`) — pure, unit-tested.
- Iframes carry `data-od-render-mode="url-load" | "srcdoc"` so the choice is visible in DevTools and assertable from tests.
- `inlineRelativeAssets` and the `$'`-safe replacer fix from PR #343 are preserved unchanged for the opt-in path.

## Out of scope (Part 2)

The `/api/projects/:id/export/:file?inline=1` daemon endpoint discussed in #368 Q3 will land in a follow-up PR. This PR keeps the legacy inliner inside `FileViewer.tsx` as the safety hatch so reviewers can validate the URL-load default in isolation.

## Tests

[...]

PR #1312, opened 2026-05-11
feat(daemon): export self-contained HTML via /export/*?inline=1 endpoint

### LINKED ISSUE #368: FileViewer: replace asset-inlining for multi-script projects with URL-loaded iframe
## Summary

The Open / FileViewer render path inlines every relative `<link rel=stylesheet>` and `<script src=…>` into a single `srcDoc` string before handing it to a sandboxed iframe. That model works for self-contained 1-file decks but is the wrong shape for multi-file React/JSX prototypes — and it remains the shape after PR #343.

PR #343 (commit `6e11eae`) fixed the immediate symptom — `String.prototype.replace` was interpreting `$'` / `$&` / `$\`` / `$1`–`$9` / `$<name>` patterns inside JSX bodies and corrupting them — by switching the reduce to a function replacer. That landed cleanly. But it didn't change the architecture; it just defused one footgun.

## Verified state (HEAD `e90e2f9`, main)

- `apps/web/src/components/FileViewer.tsx:1648` — `inlineRelativeAssets()` walks every relative `<link>` and `<script src>` and concatenates fetched bodies into the HTML string.
- Line 1696 — `resolved.reduce((next, { from, to }) => next.replace(from, () => to), html)` — the merged PR #343 fix.
- Lines 480, 1420, 1475 — three preview iframes, all `srcDoc={srcDoc}`.

## Why this is still a structural issue

1. **All-or-nothing failure mode.** A single fetch failure or HTML-parse-hostile character anywhere in the concatenated string takes down the whole iframe. The Preview pane (which uses `<iframe src=/api/projects/:id/raw/index.html>`) is unaffected because each script is its own request.
2. **No source maps, no DevTools file tree.** Inside the `srcDoc` iframe, every script appears as `Inline Babel script (n)` — no filenames, no line numbers that match the editor.
3. **Doubled bytes per Open.** Every Open click re-fetches and re-inlines the entire asset tree. Per-asset HTTP caching does not apply to inlined copies.

This bit issue #336 hard before PR #343 — a 30-file React prototype in this user's project rendered blank in Open while working perfectly in Preview, in Edge/Chrome, and in Claude Design.

## Proposal

When `inlineRelativeAssets` detects more than ~3 relative `<link>` / `<script src>` refs, skip inlining and point the iframe at the raw URL the Preview pane already uses:

```jsx
<iframe src={`/api/projects/${projectId}/raw/${fileName}`}
        sandbox="allow-scripts" />
```

- Sandbox flags unchanged — security boundary is the same.
- Daemon already serves these files for the Preview pane; no new endpoints.
- Preserve the inlining path for true single-file decks (the original use case).
- Optional: a per-file "force inline" override for the rare case where a self-contained `srcDoc` is genuinely wanted (export, screenshot, etc.).

## Effort

Small — ~30 lines in `FileViewer.tsx` plus a heuristic threshold and test cases. The hardest part is choosing the threshold and writing tests for both code paths.

## Related

- Issue #336 — the visible failure that exposed this.
- PR #343 — the `$'` corruption fix (symptom, not structure).
- PR #340 — `.jsx`/`.tsx` MIME fix.

bankielewicz:
# Summary

Closes [#368](https://github.com/nexu-io/open-design/issues/368) — Part 2 of the export-inline split @lefarcen ratified in [#issuecomment-4366243218](https://github.com/nexu-io/open-design/issues/368#issuecomment-4366243218). Part 1 (PR #384) flipped the FileViewer iframe to URL-load by default; this PR adds the `GET /api/projects/:id/export/*?inline=1` daemon endpoint that returns a self-contained HTML body for the explicit "Export self-contained HTML" + screenshot path.

- New pure helper `apps/daemon/src/inline-assets.ts` ports `apps/web/src/components/FileViewer.tsx:5248-5354` server-side, taking a `fileReader: (relPath) => Promise<string|null>` closure so the daemon route owns the filesystem boundary.
- New route wired into `apps/daemon/src/import-export-routes.ts`'s `registerProjectExportRoutes` alongside `/export/pdf` and `/archive`. Uses `ctx.projectFiles.readProjectFile`, `ctx.validation.isSafeId`, and `ctx.projectStore.getProject` — no new infrastructure.
- Branch built strict TDD (Red → Green → Refactor per phase); every behavior change has a preceding Red commit so `git log` reads as the spec.

# Design decisions

The endpoint's full contract surface — surfaced here so reviewers can see the shape without spelunking through tests:

1. **`inline=1` is required.** Missing or non-canonical → 400 `BAD_REQUEST`. The endpoint exists for inlining; a pass-through would duplicate `/raw/*`.
2. **Non-HTML files → 400 `UNSUPPORTED_FILE_TYPE`.** Inlining is `<link>`/`<script>`-specific; non-HTML has no relative-asset tags worth rewriting.
3. **Sibling-asset fetch failure → tag left intact, response still 200.** Matches the web client's failure-local behavior (`FileViewer.tsx:5311` filters nulls out). The failure-local-not-fatal property is what the issue body calls out as a structural improvement over the all-or-nothing `srcDoc` model.
4. **No response cap.** The 384 KiB transcript cap on `/finalize/anthropic` (PR #832) is about Anthropic context windows, not relevant here. The export consumer is a download or screenshot tool.
5. **No cache headers.** `/raw/*` doesn't set any either; output is dynamic (file contents + sibling reads change underfoot).
6. **No new `packages/contracts` type.** Like `/raw/*`, this serves `text/html` directly; no JSON DTO needed.
7. **`inline` accept-list = `1` / `true` / `yes` / `on` (case-insensitive).** Mirrors Part 1's `parseForceInline` at `apps/web/src/components/file-viewer-render-mode.ts:59-66` so the client `?forceInline=…` and server `?inline=…` query languages match.
8. **Duplicate identical tags: replace all.** Deliberate divergence from the web client at `FileViewer.tsx:5313` (`next.replace(from, () => to)` — first-match-only, which produces inconsistent output when a document has duplicate `<script src="x.js">` tags). An inline comment in `inline-assets.ts` documents the divergence; regression test #12 pins the new behavior. The web inline path is on a deprecation track since PR #384 made URL-load the default, so the divergence is forward-pointing.
9. **Wildcard route `/export/*` not `/export/:file`.** The architecture comment used `:file` at the discussion level, but Express's named param only matches one segment and the React-prototype use case requires nested entries (`pages/index.html`). The route registers as `/export/*` mirroring `/raw/*` at `project-routes.ts:690` and pulls `req.params[0]`.
10. **Null-origin callers NOT supported.** The daemon's `/api` middleware (`server.ts:2123-2148`) gates null-origin requests through `_NULL_ORIGIN_SAFE_GET_RE`, which only allowlists `/raw/*` and `/codex-pets/.../spritesheet`. Export consumers are the daemon UI (same-origin) and server-side screenshot tooling (no Origin header); sandboxed-iframe srcdoc previews fetch through `/raw/*` instead. Integration test #7 pins the 403 contract so a future allowlist change has to update it deliberately.

# Tests

Strict TDD ordering — every behavior change has a Red commit before the Green commit that implements it. Test delta: **+23 daemon tests** (14 unit + 9 integration in `apps/daemon/tests/export-inline-route.test.ts`).

| Phase | Commit | Subject | Tests delta | Daemon-suite size after |
|---|---|---|---|---|
| B-R | `a60a9023` | test(daemon): add Red unit tests for inlineRelativeAssets helper | +14 (all Red — module not found) | n/a |
| B-G | `5856d7be` | feat(daemon): port inlineRelativeAssets server-side for export endpoint | 0 net | 1718 |
| B-Rf | — | (no-op; helper already had named exports + extracted `replaceAllOccurrences`) | 0 | 1718 |
| C-R | `b7924fa6` | test(daemon): add Red integration tests for /export/*?inline=1 route | +9 (all Red — 404 / 403, except invalid-id which tolerantly passes) | n/a |
| C-G | _this commit_ | feat(daemon): wire GET /api/projects/:id/export/*?inline=1 endpoint | 0 net | 1727 |

Unit coverage (14 cases against the pure helper):

1. `<link rel=stylesheet>` inlined with verbatim CSS body
2. `<script src>` inlined preserving non-`src` attrs (`type=module`, `defer`, `crossorigin`)
3. Relative resolution for root + nested + deep-nested owners
4. Self-closing `<link …/>` form
5. Single-quoted attrs
6. Negative: tag without `rel` attribute
7. Negative: `rel="preload"`
8. Negative: absolute (`https:`, `http:`, `data:`, `blob:`, `mailto:`, `tel:`, `#`, leading-`/`) refs
9. `</style` escapes to `<\/style` inside CSS body
10. `</script` escapes to `<\/script` inside JS body
11. `fileReader` returning null leaves tag intact, other assets still inline
12. Duplicate identical tags both replaced (divergence per decision §8)
13. `data-od-inline-asset` attr HTML-escapes the href value
14. Deep-nested owner `a/b/c/index.html` + `../../shared/util.js` → `a/shared/util.js`

Integration coverage (9 cases against the live HTTP route):

1. 3-file React-ish layout returns single self-contained HTML — **wiring guard**: body assertions catch removal of the `await inlineRelativeAssets(...)` line in the route, not just helper-internals changes
2. Missing `inline` query → 400 `BAD_REQUEST`
3. Non-canonical inline values (`0` / `false` / `foo` / empty) → 400
4. Non-HTML file (`.css`) → 400 `UNSUPPORTED_FILE_TYPE`
5. Nonexistent file → 404 `FILE_NOT_FOUND`
6. Invalid project id (`..`) → 4xx (Express normalizes, then `isSafeId` catches whatever it routes)
7. Null-origin GET → 403 (pins decision §10)
8. Missing sibling asset → 200 with `<link>` tag intact, the script still inlined
9. Nested HTML entry (`pages/index.html` + `../shared/util.js`) → 200 inlined

# Verification

- `pnpm guard` — clean
- `pnpm typecheck` — full-repo, raw exit code 0
- `pnpm --filter @open-design/daemon test` — full daemon suite green, delta +23 tests
- Manual smoke: a 3-file project responds 200 with verbatim CSS+JS bodies; missing/non-HTML/missing-flag sad paths return the documented codes.

# Out of scope

- A web-side "Export self-contained HTML" button. The architecture comment framed this PR as the daemon-side primitive; UI surfacing belongs in a follow-up. Same shape as #451's "Continue in CLI" button living separately from the daemon endpoint it depends on.

Fixes #368.

### DIFF
--- a/apps/daemon/src/import-export-routes.ts
+++ b/apps/daemon/src/import-export-routes.ts
@@ -348,16 +348,31 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
     }
   });
 
-  // Export endpoint: serves a self-contained HTML body with every same-
-  // project <link rel=stylesheet> / <script src> inlined. Counterpart to
-  // GET /api/projects/:id/raw/* — that route stays URL-load (one request
-  // per asset; the FileViewer iframe's default since PR #384). This route
-  // exists for explicit "Export self-contained HTML" + screenshot tooling
-  // where one HTTP response with no inter-file dependencies is the goal.
-  // Null-origin (sandboxed iframe srcdoc) callers are intentionally NOT
-  // supported — the only consumers are the daemon UI (same-origin) and
-  // server-side screenshot tooling (no Origin header). See nexu-io/
-  // open-design#368 and the architecture lock at
+  // Export endpoint: serves an HTML body with every same-project
+  // top-level `<link rel=stylesheet>` / `<script src>` inlined.
+  // Counterpart to GET /api/projects/:id/raw/* — that route stays
+  // URL-load (one request per asset; FileViewer's default since
+  // PR #384). This route exists for explicit "Inline top-level
+  // CSS/JS" exports + the screenshot path where the headless browser
+  // fetches the response and renders it.
+  //
+  // Scope is intentionally narrow: only `<link rel=stylesheet>` and
+  // `<script src>` are rewritten. `<img src>`, CSS `url(...)` refs,
+  // `@import`, ES module imports, font sources, and similar remain
+  // external in the response — see the docstring on
+  // `apps/daemon/src/inline-assets.ts` for the full not-rewritten list
+  // and rationale. A fully offline "self-contained" export with image
+  // and font bundling would be a follow-up issue.
+  //
+  // Null-origin (sandboxed iframe srcdoc) callers are intentionally
+  // NOT supported — the only consumers are the daemon UI (same-origin)
+  // and server-side screenshot tooling (no Origin header). The
+  // response also carries `Content-Security-Policy: sandbox
+  // allow-scripts` so top-level browser navigation (no Origin header,
+  // would otherwise pass the daemon middleware) cannot escalate to
+  // daemon-origin privileges through script execution.
+  //
+  // See nexu-io/open-design#368 and the architecture lock at
   // https://github.com/nexu-io/open-design/issues/368#issuecomment-4366243218.
   app.get('/api/projects/:id/export/*', async (req, res) => {
     try {
--- a/apps/daemon/src/import-export-routes.ts
+++ b/apps/daemon/src/import-export-routes.ts
@@ -1,6 +1,10 @@
 import type { Express } from 'express';
 import type { RouteDeps } from './server-context.js';
-import { inlineRelativeAssets, type InlineAssetReader } from './inline-assets.js';
+import {
+  InlineAssetsLimitError,
+  inlineRelativeAssets,
+  type InlineAssetReader,
+} from './inline-assets.js';
 
 export interface RegisterImportRoutesDeps extends RouteDeps<'db' | 'http' | 'uploads' | 'node' | 'ids' | 'paths' | 'imports' | 'auth' | 'projectStore' | 'conversations' | 'projectFiles'> {}
 
@@ -410,8 +414,8 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
       if (!file.mime.startsWith('text/html')) {
         return sendApiError(
           res,
-          400,
-          'UNSUPPORTED_FILE_TYPE',
+          415,
+          'UNSUPPORTED_MEDIA_TYPE',
           'export endpoint only supports HTML files',
         );
       }
@@ -446,6 +450,14 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
       res.setHeader('Content-Security-Policy', 'sandbox allow-scripts');
       res.type('text/html').send(rendered);
     } catch (err: any) {
+      // PR #1312 round-3 (lefarcen P2): the inliner's cap-enforcement
+      // throws InlineAssetsLimitError when the owner HTML, candidate
+      // count, or assembled output exceeds the module-level limits.
+      // Map every such throw to a 413 PAYLOAD_TOO_LARGE envelope so
+      // callers see a structured error rather than a generic 400.
+      if (err instanceof InlineAssetsLimitError || err?.name === 'InlineAssetsLimitError') {
+        return sendApiError(res, 413, 'PAYLOAD_TOO_LARGE', String(err));
+      }
       sendApiError(res, 400, 'BAD_REQUEST', String(err));
     }
   });
--- a/apps/daemon/src/import-export-routes.ts
+++ b/apps/daemon/src/import-export-routes.ts
@@ -228,7 +228,7 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
   const { sendApiError } = ctx.http;
   const { PROJECTS_DIR } = ctx.paths;
   const { getProject } = ctx.projectStore;
-  const { readProjectFile } = ctx.projectFiles;
+  const { readProjectFile, resolveProjectFilePath } = ctx.projectFiles;
   const { isSafeId } = ctx.validation;
   const {
     buildProjectArchive,
@@ -420,18 +420,39 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
         );
       }
 
+      // PR #1312 round-4 (lefarcen P2): stat first, then read. This
+      // lets the helper short-circuit on maxAssetBytes / maxTotalBytes
+      // BEFORE the buffer is materialized into memory. A 100 MiB
+      // sibling file is rejected after the cheap stat call, not after
+      // a 100 MiB readFile.
       const fileReader: InlineAssetReader = async (sibling) => {
+        let meta;
         try {
-          const siblingFile = await readProjectFile(
+          meta = await resolveProjectFilePath(
             PROJECTS_DIR,
             req.params.id,
             sibling,
             project?.metadata,
           );
-          return siblingFile.buffer.toString('utf8');
         } catch {
           return null;
         }
+        return {
+          size: meta.size,
+          read: async () => {
+            try {
+              const siblingFile = await readProjectFile(
+                PROJECTS_DIR,
+                req.params.id,
+                sibling,
+                project?.metadata,
+              );
+              return siblingFile.buffer.toString('utf8');
+            } catch {
+              return null;
+            }
+          },
+        };
       };
 
       const rendered = await inlineRelativeAssets(
--- a/apps/daemon/src/import-export-routes.ts
+++ b/apps/daemon/src/import-export-routes.ts
@@ -1,5 +1,6 @@
 import type { Express } from 'express';
 import type { RouteDeps } from './server-context.js';
+import { inlineRelativeAssets, type InlineAssetReader } from './inline-assets.js';
 
 export interface RegisterImportRoutesDeps extends RouteDeps<'db' | 'http' | 'uploads' | 'node' | 'ids' | 'paths' | 'imports' | 'auth' | 'projectStore' | 'conversations' | 'projectFiles'> {}
 
@@ -216,13 +217,15 @@ export function registerImportRoutes(app: Express, ctx: RegisterImportRoutesDeps
 
 }
 
-export interface RegisterProjectExportRoutesDeps extends RouteDeps<'db' | 'http' | 'paths' | 'projectStore' | 'exports'> {}
+export interface RegisterProjectExportRoutesDeps extends RouteDeps<'db' | 'http' | 'paths' | 'projectStore' | 'exports' | 'projectFiles' | 'validation'> {}
 
 export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectExportRoutesDeps) {
   const { db } = ctx;
   const { sendApiError } = ctx.http;
   const { PROJECTS_DIR } = ctx.paths;
   const { getProject } = ctx.projectStore;
+  const { readProjectFile } = ctx.projectFiles;
+  const { isSafeId } = ctx.validation;
   const {
     buildProjectArchive,
     buildBatchArchive,
@@ -342,6 +345,84 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
         status === 404 ? 'FILE_NOT_FOUND' : 'BAD_REQUEST',
         String(err?.message || err),
       );
+    }
+  });
+
+  // Export endpoint: serves a self-contained HTML body with every same-
+  // project <link rel=stylesheet> / <script src> inlined. Counterpart to
+  // GET /api/projects/:id/raw/* — that route stays URL-load (one request
+  // per asset; the FileViewer iframe's default since PR #384). This route
+  // exists for explicit "Export self-contained HTML" + screenshot tooling
+  // where one HTTP response with no inter-file dependencies is the goal.
+  // Null-origin (sandboxed iframe srcdoc) callers are intentionally NOT
+  // supported — the only consumers are the daemon UI (same-origin) and
+  // server-side screenshot tooling (no Origin header). See nexu-io/
+  // open-design#368 and the architecture lock at
+  // https://github.com/nexu-io/open-design/issues/368#issuecomment-4366243218.
+  app.get('/api/projects/:id/export/*', async (req, res) => {
+    try {
+      if (!isSafeId(req.params.id)) {
+        return sendApiError(res, 400, 'BAD_REQUEST', 'invalid project id');
+      }
+
+      const inlineRaw =
+        typeof req.query.inline === 'string' ? req.query.inline.trim().toLowerCase() : '';
+      if (!['1', 'true', 'yes', 'on'].includes(inlineRaw)) {
+        return sendApiError(
+          res,
+          400,
+          'BAD_REQUEST',
+          "query parameter 'inline=1' is required",
+        );
+      }
+
+      const project = getProject(db, req.params.id);
+      const relPath = (req.params as any)[0];
+
+      let file;
+      try {
+        file = await readProjectFile(PROJECTS_DIR, req.params.id, relPath, project?.metadata);
+      } catch (err: any) {
+        const status = err && err.code === 'ENOENT' ? 404 : 400;
+        return sendApiError(
+          res,
+          status,
+          status === 404 ? 'FILE_NOT_FOUND' : 'BAD_REQUEST',
+          String(err),
+        );
+      }
+
+      if (!file.mime.startsWith('text/html')) {
+        return sendApiError(
+          res,
+          400,
+          'UNSUPPORTED_FILE_TYPE',
+          'export endpoint only supports HTML files',
+        );
+      }
+
+      const fileReader: InlineAssetReader = async (sibling) => {
+        try {
+          const siblingFile = await readProjectFile(
+            PROJECTS_DIR,
+            req.params.id,
+            sibling,
+            project?.metadata,
+          );
+          return siblingFile.buffer.toString('utf8');
+        } catch {
+          return null;
+        }
+      };
+
+      const rendered = await inlineRelativeAssets(
+        file.buffer.toString('utf8'),
+        relPath,
+        fileReader,
+      );
+      res.type('text/html').send(rendered);
+    } catch (err: any) {
+      sendApiError(res, 400, 'BAD_REQUEST', String(err));
     }
   });
 
--- a/apps/daemon/src/import-export-routes.ts
+++ b/apps/daemon/src/import-export-routes.ts
@@ -2,6 +2,7 @@ import type { Express } from 'express';
 import type { RouteDeps } from './server-context.js';
 import {
   InlineAssetsLimitError,
+  MAX_INLINE_OWNER_BYTES,
   inlineRelativeAssets,
   type InlineAssetReader,
 } from './inline-assets.js';
@@ -398,9 +399,27 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
       const project = getProject(db, req.params.id);
       const relPath = (req.params as any)[0];
 
-      let file;
+      // PR #1312 round-5 (lefarcen P2): stat the owner file BEFORE
+      // readProjectFile so a 100 MiB owner HTML is rejected after a
+      // cheap stat() call, not after a 100 MiB readFile() into memory.
+      // The size check + mime check both run pre-buffer here, mirroring
+      // the sibling-asset stat-then-read contract round 4 already
+      // applied via AssetHandle. Size fires before mime so an oversize
+      // non-HTML file returns 413 (not 415) — that ordering is the
+      // observable Red→Green for this round.
+      //
+      // The helper's ownerBytes check (inline-assets.ts:127-133) stays
+      // as defense-in-depth: it still catches direct in-process callers
+      // that skip the route and any future drift in the size reported
+      // by stat vs the bytes actually returned by readFile.
+      let ownerMeta;
       try {
-        file = await readProjectFile(PROJECTS_DIR, req.params.id, relPath, project?.metadata);
+        ownerMeta = await resolveProjectFilePath(
+          PROJECTS_DIR,
+          req.params.id,
+          relPath,
+          project?.metadata,
+        );
       } catch (err: any) {
         const status = err && err.code === 'ENOENT' ? 404 : 400;
         return sendApiError(
@@ -411,12 +430,34 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
         );
       }
 
-      if (!file.mime.startsWith('text/html')) {
+      if (ownerMeta.size > MAX_INLINE_OWNER_BYTES) {
+        return sendApiError(
+          res,
+          413,
+          'PAYLOAD_TOO_LARGE',
+          `owner html ${ownerMeta.size} bytes exceeds MAX_INLINE_OWNER_BYTES ${MAX_INLINE_OWNER_BYTES}`,
+        );
+      }
+
+      if (!ownerMeta.mime.startsWith('text/html')) {
         return sendApiError(
           res,
           415,
           'UNSUPPORTED_MEDIA_TYPE',
           'export endpoint only supports HTML files',
+        );
+      }
+
+      let file;
+      try {
+        file = await readProjectFile(PROJECTS_DIR, req.params.id, relPath, project?.metadata);
+      } catch (err: any) {
+        const status = err && err.code === 'ENOENT' ? 404 : 400;
+        return sendApiError(
+          res,
+          status,
+          status === 404 ? 'FILE_NOT_FOUND' : 'BAD_REQUEST',
+          String(err),
         );
       }
 
--- a/apps/daemon/src/import-export-routes.ts
+++ b/apps/daemon/src/import-export-routes.ts
@@ -420,6 +420,15 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
         relPath,
         fileReader,
       );
+      // PR #1312 round-2 (lefarcen P2): top-level browser navigation to
+      // this URL sends no Origin header, so the /api middleware lets it
+      // through. Without a CSP, any JS in the exported document would
+      // run at daemon origin with access to /api/, cookies, localStorage,
+      // etc. `sandbox allow-scripts` treats the response like a sandboxed
+      // iframe with an opaque origin — scripts execute (that's the point
+      // of inlining JS for screenshot tooling), but cannot read cookies,
+      // hit /api/, or escalate to daemon-origin privileges.
+      res.setHeader('Content-Security-Policy', 'sandbox allow-scripts');
       res.type('text/html').send(rendered);
     } catch (err: any) {
       sendApiError(res, 400, 'BAD_REQUEST', String(err));
--- a/apps/daemon/src/inline-assets.ts
+++ b/apps/daemon/src/inline-assets.ts
@@ -101,9 +101,31 @@ export async function inlineRelativeAssets(
   return parts.join('');
 }
 
-function buildInlineStyleBlock(_tag: string, href: string, css: string): string {
+// Attrs that affect <style> semantics and must be carried across from
+// the source <link rel=stylesheet> so the inlined output matches the
+// behavior of the original URL-loaded stylesheet:
+//   - media        — media query (e.g. `media="print"` for print-only)
+//   - title        — alternate stylesheet sets
+//   - disabled     — boolean: initial disabled state
+//   - nonce        — CSP nonce passthrough
+// All four are valid on both <link rel=stylesheet> and <style>. Other
+// <link> attrs (rel, href, type, crossorigin, integrity, referrerpolicy)
+// don't apply to <style> and are intentionally dropped.
+const STYLE_PRESERVED_LINK_ATTRS = ['media', 'title', 'nonce'] as const;
+const STYLE_PRESERVED_BOOLEAN_ATTRS = ['disabled'] as const;
+
+function buildInlineStyleBlock(tag: string, href: string, css: string): string {
+  const carried: string[] = [];
+  for (const name of STYLE_PRESERVED_LINK_ATTRS) {
+    const value = readHtmlAttr(tag, name);
+    if (value != null) carried.push(`${name}="${escapeHtmlAttr(value)}"`);
+  }
+  for (const name of STYLE_PRESERVED_BOOLEAN_ATTRS) {
+    if (hasBooleanHtmlAttr(tag, name)) carried.push(name);
+  }
+  const attrString = carried.length === 0 ? '' : ` ${carried.join(' ')}`;
   return (
-    `<style data-od-inline-asset="${escapeHtmlAttr(href)}">\n` +
+    `<style data-od-inline-asset="${escapeHtmlAttr(href)}"${attrString}>\n` +
     `${css.replace(/<\/style/gi, '<\\/style')}\n</style>`
   );
 }
@@ -139,6 +161,14 @@ export function resolveProjectRelativePath(
 export function readHtmlAttr(tag: string, name: string): string | null {
   const match = tag.match(new RegExp(`\\s${name}\\s*=\\s*(['"])([\\s\\S]*?)\\1`, 'i'));
   return match?.[2] ?? null;
+}
+
+// HTML boolean attribute presence test — matches `<tag … name>` or
+// `<tag … name=""…>` without requiring a value, but does NOT match a
+// substring inside another attribute's value (e.g. `data-foo="disabled"`
+// must NOT count as `disabled` being set).
+export function hasBooleanHtmlAttr(tag: string, name: string): boolean {
+  return new RegExp(`\\s${name}(?=\\s|=|/?>)`, 'i').test(tag);
 }
 
 export function escapeHtmlAttr(value: string): string {
--- a/apps/daemon/src/inline-assets.ts
+++ b/apps/daemon/src/inline-assets.ts
@@ -13,77 +13,108 @@ export interface InlineAssetReader {
   (relPath: string): Promise<string | null>;
 }
 
-interface Replacement {
-  from: string;
-  to: string;
-}
-
 export async function inlineRelativeAssets(
   html: string,
   ownerFileName: string,
   fileReader: InlineAssetReader,
 ): Promise<string> {
-  const replacements: Array<Promise<Replacement | null>> = [];
+  // Each candidate records the exact byte span in the ORIGINAL html. We
+  // never mutate-then-rescan, so a tag literal that happens to appear
+  // inside an inlined asset body is left untouched.
+  //
+  // Two divergences from apps/web/src/components/FileViewer.tsx:5265-5314:
+  //
+  // 1. Position-based, single-pass concat (vs. `.reduce` over `.replace`).
+  //    The web client's reduce-over-replace re-scans the mutated string
+  //    on each pass, which (a) replaces only the first occurrence of a
+  //    duplicate tag and (b) corrupts already-inlined bodies that contain
+  //    another tag's literal substring. This helper avoids both by
+  //    operating on captured indices in the original input.
+  // 2. Duplicate identical tags: this helper inlines every occurrence
+  //    (the web client's first-match-only is a side-effect of `.replace`
+  //    semantics). The web inline path is on a deprecation track since
+  //    PR #384 made URL-load the default, so the divergence is
+  //    forward-pointing.
+  interface Candidate {
+    start: number;
+    end: number;
+    replacement: Promise<string | null>;
+  }
 
-  const linkTags = html.match(/<link\b[^>]*>/gi) ?? [];
-  for (const tag of linkTags) {
+  const candidates: Candidate[] = [];
+
+  for (const match of html.matchAll(/<link\b[^>]*>/gi)) {
+    const tag = match[0];
+    const start = match.index!;
     const rel = readHtmlAttr(tag, 'rel');
     const href = readHtmlAttr(tag, 'href');
     if (!rel || !/\bstylesheet\b/i.test(rel) || !href) continue;
     const resolved = resolveProjectRelativePath(ownerFileName, href);
     if (!resolved) continue;
-    replacements.push(
-      fileReader(resolved).then<Replacement | null>((css) =>
-        css == null
-          ? null
-          : {
-              from: tag,
-              to:
-                `<style data-od-inline-asset="${escapeHtmlAttr(href)}">\n` +
-                `${css.replace(/<\/style/gi, '<\\/style')}\n</style>`,
-            },
+    candidates.push({
+      start,
+      end: start + tag.length,
+      replacement: fileReader(resolved).then((css) =>
+        css == null ? null : buildInlineStyleBlock(tag, href, css),
       ),
-    );
+    });
   }
 
-  const scriptTags =
-    html.match(/<script\b[^>]*\bsrc\s*=\s*["'][^"']+["'][^>]*>\s*<\/script>/gi) ?? [];
-  for (const tag of scriptTags) {
+  for (const match of html.matchAll(
+    /<script\b[^>]*\bsrc\s*=\s*["'][^"']+["'][^>]*>\s*<\/script>/gi,
+  )) {
+    const tag = match[0];
+    const start = match.index!;
     const src = readHtmlAttr(tag, 'src');
     if (!src) continue;
     const resolved = resolveProjectRelativePath(ownerFileName, src);
     if (!resolved) continue;
-    replacements.push(
-      fileReader(resolved).then<Replacement | null>((js) => {
-        if (js == null) return null;
-        const open = tag.match(/^<script\b[^>]*>/i)?.[0] ?? '<script>';
-        const attrs = open
-          .replace(/^<script/i, '')
-          .replace(/>$/i, '')
-          .replace(/\ssrc\s*=\s*(['"])[\s\S]*?\1/i, '');
-        return {
-          from: tag,
-          to: `<script${attrs}>\n${js.replace(/<\/script/gi, '<\\/script')}\n</script>`,
-        };
-      }),
-    );
+    candidates.push({
+      start,
+      end: start + tag.length,
+      replacement: fileReader(resolved).then((js) =>
+        js == null ? null : buildInlineScriptBlock(tag, js),
+      ),
+    });
   }
 
-  const resolvedReplacements = (await Promise.all(replacements)).filter(
-    (item): item is Replacement => item !== null,
-  );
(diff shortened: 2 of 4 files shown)


apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,123 @@
...
+export function escapeHtmlAttr(value: string): string {
+  return value
+    .replace(/&/g, '&amp;')
+    .replace(/"/g, '&quot;')
+    .replace(/</g, '&lt;')
+    .replace(/>/g, '&gt;');
+}
+
+function replaceAllOccurrences(source: string, from: string, to: string): string {
+  return source.split(from).join(to);
```
Siri-Ray: `replaceAllOccurrences` re-scans the whole HTML after earlier replacements have already inserted asset bodies, so it can rewrite literal tag strings inside those asset bodies instead of only the tags matched in the original document. For example, if `a.js` contains `const tpl = "<script src=\"b.js\"></script>";` and the page also has `<script src="b.js"></script>`, the second replacement pass will mutate the string inside the already-inlined `a.js`, corrupting the exported JavaScript. That breaks the endpoint's self-contained export contract for valid JS/CSS payloads that embed HTML templates or tag strings. Please apply replacements by original match position, or rebuild the output in one pass from captured ranges, so replacement text is never searched again; add a regression test with two assets where the first asset body contains the second tag literally.

Siri-Ray (commented):
I found one blocking correctness issue in the inliner: replacement needs to target the original matched tag spans instead of re-scanning the progressively mutated HTML.


apps/daemon/src/import-export-routes.ts
```diff
@@ -345,6 +348,84 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
...
+          return null;
+        }
+      };
+
+      const rendered = await inlineRelativeAssets(
+        file.buffer.toString('utf8'),
+        relPath,
+        fileReader,
+      );
+      res.type('text/html').send(rendered);
```
lefarcen, line 513: **P2 — Same-origin execution risk**

PR body decision §10 says null-origin is blocked, but top-level browser navigation (opening exported HTML directly) sends no `Origin` header and passes the middleware. Project JS then executes under daemon origin with full API access.

If this is for download: add `Content-Disposition: attachment` header.
If this is for screenshot: add CSP `sandbox allow-scripts` or a non-browser token gate.

Otherwise exported HTML becomes a daemon-origin XSS vector.

apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,123 @@
...
+    replacements.push(
+      fileReader(resolved).then<Replacement | null>((css) =>
+        css == null
+          ? null
+          : {
+              from: tag,
+              to:
+                `<style data-od-inline-asset="${escapeHtmlAttr(href)}">\n` +
+                `${css.replace(/<\/style/gi, '<\\/style')}\n</style>`,
+            },
```
lefarcen: **P2 — Stylesheet attributes lost**

`<link rel="stylesheet" href="print.css" media="print">` becomes plain `<style>` with no media query, so print-only styles apply unconditionally.

Preserve `media`, `title`, `disabled`, `nonce` on the generated `<style>` tag, or skip inlining links with unsupported semantics and document the limitation.

apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,123 @@
...
+        if (js == null) return null;
+        const open = tag.match(/^<script\b[^>]*>/i)?.[0] ?? '<script>';
+        const attrs = open
+          .replace(/^<script/i, '')
+          .replace(/>$/i, '')
+          .replace(/\ssrc\s*=\s*(['"])[\s\S]*?\1/i, '');
+        return {
+          from: tag,
+          to: `<script${attrs}>\n${js.replace(/<\/script/gi, '<\\/script')}\n</script>`,
+        };
```
lefarcen: **P2 — "Self-contained" incomplete**

The PR title/body claim "self-contained HTML" but this only rewrites top-level `<link>` + `<script src>`. Relative ``, CSS `url(...)`, CSS `@import`, module `import './x.js'`, fonts, etc. remain external.

Either:
- Narrow the contract/name to "inline top-level CSS/JS tags"
- OR implement recursive/data-URL rewriting for img/CSS-url/module-import with tests

Screenshot tooling expecting true self-contained will break on images/fonts.

lefarcen (commented):
Hi @bankielewicz! 👋

Solid TDD discipline throughout — the Red → Green commit structure makes the diff very easy to follow. The helper port from web client is clean, and the route wiring follows daemon patterns.

I've spotted a few gaps where the "self-contained HTML" contract doesn't quite hold yet, plus some operational safety items for the daemon deployment. See inline below — all P2/P3, nothing blocking.

## Summary

**What's promising:**
- Strict TDD with Red commits visible in git history
- Pure helper with 14 unit tests covering edge cases
- Route integration follows existing export patterns
- Divergence from web client (replace-all) is documented

**What needs attention:**
1. **Same-origin execution risk** (P2): Exported HTML can run with daemon-origin privileges if opened directly
2. **Stylesheet attributes lost** (P2): `media="print"`, `disabled`, etc. dropped during inline
3. **"Self-contained" incomplete** (P2): Images, CSS url(), module imports, fonts still external
4. **No response cap** (P2): Large projects can spike daemon memory unbounded
5. **Accept-list test gap** (P3): `inline=true/yes/on` not covered despite PR body promise
6. **Invalid-id test weak** (P3): Doesn't prove `isSafeId` guard works

---

Looking forward to Part 3! 🚀

apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,123 @@
...
+      }),
+    );
+  }
+
+  const resolvedReplacements = (await Promise.all(replacements)).filter(
```
lefarcen: **P2 — No response cap (operational risk)**

PR body decision §4 justifies "no cap" by saying #832's transcript limit is unrelated, but this route does unbounded concurrent reads + multiple full string copies. A project with large HTML or many large assets can spike daemon memory.

Add limits:
- Max input HTML bytes (e.g. 2 MB)
- Max asset count (e.g. 100)
- Max total inlined bytes (e.g. 10 MB)
- Bounded concurrency

Or document expected sizes + operational reason the local daemon can tolerate them.

apps/daemon/tests/export-inline-route.test.ts
```diff
@@ -0,0 +1,317 @@
...
+    expect(res.status).toBe(400);
+    const body = (await res.json()) as { error: { code: string } };
+    expect(body.error.code).toBe('BAD_REQUEST');
+  });
+
+  it('returns 400 for non-canonical inline values (0, false, foo)', async () => {
+    for (const q of ['inline=0', 'inline=false', 'inline=foo', 'inline=']) {
+      const res = await fetch(exportUrl('index.html', q));
+      expect(res.status).toBe(400);
+    }
```
lefarcen, line 541: **P3 — Accept-list test gap**

PR body decision §7 promises `inline=true` / `yes` / `on` (case-insensitive), but tests only cover `inline=1`.

Add integration cases:
```ts
it('accepts inline=true', async () => {
  const res = await fetch(exportUrl('index.html', 'inline=true'));
  expect(res.status).toBe(200);
});
// + yes, on, TRUE, YES, ON
```

apps/daemon/tests/export-inline-route.test.ts
```diff
@@ -0,0 +1,317 @@
...
+
+  it('returns 400 BAD_REQUEST for an invalid project id (..)', async () => {
+    const res = await fetch(`${baseUrl}/api/projects/../export/index.html?inline=1`);
+    // Express normalizes `..` segments before routing, so this should not
+    // reach our handler; the daemon's middleware or routing answers first.
+    // Either way, the request must NOT succeed at extracting a parent
+    // directory.
+    expect(res.status).toBeGreaterThanOrEqual(400);
+    expect(res.status).toBeLessThan(500);
+  });
```
lefarcen, line 571: **P3 — Invalid-id test weak**

Express normalizes `..` before routing, so this test doesn't prove the route's `isSafeId` guard works. Use an encoded candidate like `%2e%2e` if Express permits, or unit-test the route logic directly.

Also assert the documented `400 BAD_REQUEST` shape, not just "some 4xx".

bankielewicz:
Round 2 pushed — addresses all 7 review items (Siri-Ray blocking + lefarcen 4×P2 + 2×P3) across 7 commits with the Red→Green TDD chain preserved.

| Concern | Commits | Note |
|---|---|---|
| **Siri-Ray (blocking) — inlined-body tag-literal corruption** at `inline-assets.ts:122` | `c809bccc` (Red) → `f20a2959` (Green) | Replaced `split/join` over the mutated string with **position-based concat**: `matchAll` captures every `{start,end}` span from the ORIGINAL html in one pass; replacements are awaited in parallel, then sliced together. Text introduced by an earlier replacement is never re-scanned. Regression test uses a CSS `<link>` scenario where the `</style` safety escape doesn't kick in (a.css comment contains the literal `<link rel="stylesheet" href="b.css">`); under the old reducer the literal got replaced AND b.css's `<style>` block was injected inside a.css's comment — the new helper leaves the literal verbatim and inlines b.css exactly once at its real position. |
| **lefarcen P2 — same-origin execution risk** at `import-export-routes.ts:423` | `58151356` (Red) → `682df0d2` (Green) | `Content-Security-Policy: sandbox allow-scripts` on the response. Treats the document as a sandboxed iframe with an opaque origin — scripts execute (necessary for the screenshot use case) but cookies / `/api/` / `localStorage` are out of reach. |
| **lefarcen P2 — stylesheet attrs dropped** at `inline-assets.ts:44` | `55a348d9` (Red) → `b78b8e41` (Green) | `buildInlineStyleBlock` now carries `media` / `title` / `nonce` (value attrs) + `disabled` (boolean attr) across to the emitted `<style>` tag. Other `<link>` attrs (`rel`, `href`, `type`, `crossorigin`, `integrity`, `referrerpolicy`) don't apply to `<style>` and are intentionally dropped. New `hasBooleanHtmlAttr` helper distinguishes attr-presence from substring-inside-another-value via a word-boundary regex. |
| **lefarcen P2 — "self-contained" incomplete** at `inline-assets.ts:67` | `a43c9d6a` | Narrowed the contract claim: the inliner handles top-level `<link rel=stylesheet>` and `<script src>` only. ``, CSS `url(...)`, CSS `@import`, ES module imports, font sources, and similar remain external. The module docstring enumerates the full not-rewritten list and names the screenshot path as the primary motivator (a headless browser fetches each external asset on render, so inline-CSS-and-JS-only is sufficient). The route handler comment block mirrors the contract. Fully offline export with image/font bundling is a follow-up — out of scope for this PR. |
| **lefarcen P2 — no response cap** at `inline-assets.ts:72` | `a43c9d6a` | Documented the local-daemon trust model in the helper docstring: single-user, developer's machine (per `open_design_architecture.md`), so the effective ceiling is the size of the user's own project and no hard cap is enforced. The docstring also names the conditions under which a bounded-concurrency reader + output-size limit would be needed (non-trusted callers). |
| **lefarcen P3 — accept-list test gap** at `test.ts:262` | `58151356` | Integration cases for `inline=true / yes / on / TRUE / Yes / ON`. The route's parser already implemented the accept list per decision §7; tests now pin the contract. |
| **lefarcen P3 — invalid-id test weak** at `test.ts:287` | `58151356` | Switched to `bad!id` — URL-safe so Express passes it through unchanged into `req.params`, but outside isSafeId's `/^[A-Za-z0-9._-]+$/` char class so the guard fires. Test now asserts the documented `400 BAD_REQUEST` envelope and message ("invalid project id"). |

**Verification after round 2:** daemon delta is +5 tests (1 regression, 1 CSP, 1 link-attr, 2 P3 coverage gaps) — daemon suite now 1732 passing. `pnpm guard` clean, full-repo `pnpm typecheck` exit 0.

Locked decisions §8 (dup-tag replace-all) and §9 (wildcard route `/export/*`) are preserved — position-based replacement keeps every original-tag position getting its own slice, so all duplicates still inline. §10 (null-origin not supported) and the new CSP-sandbox layer give belt-and-suspenders against same-origin escalation.

**Out of scope for this PR (potential follow-ups if there's demand):**
- Fully self-contained offline export with `` → data-URI, CSS `url()` rewriting, font bundling, `@import` resolution
- Bounded-concurrency reader + output-size limit if `/export/` is ever surfaced to non-trusted callers

apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,205 @@
...
+
+  if (candidates.length === 0) return html;
+
+  // Sort by start so we can splice slices of the original html in order.
+  // Link and script regions cannot overlap in a well-formed document
+  // (the HTML parser disallows nested <script> / <link>), so we don't
+  // need overlap-resolution logic.
+  candidates.sort((a, b) => a.start - b.start);
+
+  const resolved = await Promise.all(candidates.map((c) => c.replacement));
```
lefarcen: **P2** — Response-size cap still not addressed. Round 2 documents that no cap is enforced, but the implementation still builds every candidate, starts every `fileReader` read, waits with unbounded `Promise.all`, and concatenates the full output in memory. Large or repeated asset refs can consume daemon memory/CPU. Fix by limiting: max candidates, max individual asset bytes, max aggregate output, bounded read concurrency.

apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,205 @@
...
+export function readHtmlAttr(tag: string, name: string): string | null {
+  const match = tag.match(new RegExp(`\\s${name}\\s*=\\s*(['"])([\\s\\S]*?)\\1`, 'i'));
+  return match?.[2] ?? null;
+}
+
+// HTML boolean attribute presence test — matches `<tag … name>` or
+// `<tag … name=""…>` without requiring a value, but does NOT match a
+// substring inside another attribute's value (e.g. `data-foo="disabled"`
+// must NOT count as `disabled` being set).
+export function hasBooleanHtmlAttr(tag: string, name: string): boolean {
```
lefarcen, line 399: **P3** — `hasBooleanHtmlAttr` can match `disabled` inside quoted attribute values despite the comment saying it cannot. For example `data-note="currently disabled"` satisfies `\sdisabled(?=\s|=|/?>)` and will add a real `disabled` attribute to the generated `<style>`, potentially disabling a stylesheet that was enabled. Fix by parsing attributes while respecting quotes, or reuse a small attribute tokenizer instead of regexing the whole tag.

lefarcen (commented):
Round 2 addresses the major blockers — nice progress! The position-based concat fixes the tag-literal corruption, CSP sandbox blocks same-origin escalation, and the test coverage gaps are filled.

Two items remain:

1. **P2 — Response-size cap** (still open): Round 2 documents that no cap is enforced, but the code still builds unbounded candidate arrays + runs `Promise.all` over all asset reads + concatenates the full output in memory. Large projects or repeated asset refs can consume daemon memory/CPU. Consider limiting: max candidates, max individual asset bytes, max aggregate output, bounded read concurrency.

2. **P3 — Boolean-attr scanner edge case**: `hasBooleanHtmlAttr` can match `disabled` inside quoted attribute values (e.g. `data-note="currently disabled"` satisfies `\sdisabled(?=\s|=|/?>)`), potentially adding a real `disabled` attr to the generated `<style>` when the source `<link>` was enabled.

The blocking tag-corruption fix is solid. These two are polish — up to you whether to tackle them now or defer to a follow-up.

bankielewicz:
Round 3 pushed — both items closed, no follow-up deferral.

| Concern | Commits | Note |
|---|---|---|
| **P3 — `hasBooleanHtmlAttr` quoted-value false positive** | `5f041aeb` (Red) → `066825be` (Green) | The regex `\sdisabled(?=\s|=|/?>)` ran over the raw tag string, so the literal substring `disabled` appearing inside any quoted attribute value followed by whitespace satisfied the lookahead — `data-note="content disabled stuff"` was silently emitting `<style disabled>`. Fix strips `="…"` and `='…'` substrings out of the tag with two regex passes before testing for the bare attr. The Red regression covers both the false-positive case AND a "legitimate disabled still detected" counterweight so the next refactor doesn't over-correct. |
| **P2 — actual size/concurrency caps in code** | `f74f60f9` (Red) → `8d432a4e` (Green) | Documented-only wasn't enough — round-2's middle-ground caveat is now real enforcement. Five caps with sensible defaults: `MAX_INLINE_OWNER_BYTES=2 MiB`, `MAX_INLINE_ASSET_BYTES=5 MiB`, `MAX_INLINE_CANDIDATES=500`, `MAX_INLINE_TOTAL_BYTES=50 MiB`, `MAX_INLINE_READ_CONCURRENCY=8`. New `InlineAssetsLimitError` discriminates which cap fired; the route maps every throw to `413 PAYLOAD_TOO_LARGE`. The asset cap is the only graceful one (oversized sibling → tag stays as URL ref so one bad asset doesn't kill the whole export). Reads now flow through a tiny in-module `runWithConcurrency` worker pool so a project with hundreds of `<link>`/`<script>` tags doesn't open hundreds of file descriptors simultaneously. `InlineOptions` adds the test-door pattern (per `feedback_test_doors_over_fake_timers.md`) so the cap tests use tiny fixtures rather than 2-50 MiB on-disk writes. |

**Drive-by fix discovered while wiring 413:** the round-1 non-HTML branch was emitting `UNSUPPORTED_FILE_TYPE` (status 400) which is **not** a registered `ApiErrorCode` in `packages/contracts/src/errors.ts:3-62`. The canonical code for "wrong content type" is `UNSUPPORTED_MEDIA_TYPE` with HTTP 415, which the route now uses. Test updated; commit `8d432a4e`.

**Verification after round 3:** daemon delta is +8 tests (2 quoted-attr regression cases + 5 cap unit + 1 cap integration). Daemon suite is 1740 total — 1739 pass + 1 skipped in steady state. Full-suite parallel run flaked on three unrelated test files (`finalize-route-abort`, `orbit`, `project-watchers`); each passes cleanly in isolation, and none touch the inline-assets / `/export` route. `pnpm guard` clean, full-repo `pnpm typecheck` exit 0.

Cap defaults let every realistic developer project through but reject pathological inputs. If a real project ever hits these, the 413 envelope signals it clearly and we can raise the constants in a follow-up.

| Phase | Commits | Test delta |
|---|---|---|
| I3 — quoted-attr scanner | `5f041aeb` + `066825be` | +2 |
| H3 — cap enforcement | `f74f60f9` + `8d432a4e` | +5 unit + 1 integration |

PR commit chain is now: round-1 (B-R / B-G / C-R / C-G), round-2 (F-R / F-G / G1-R/G / G2-R/G / G3-docs), round-3 (I3-R / I3-G / H3-R / H3-G). 15 commits total, every behavior change preceded by its Red commit. `git log --oneline main..HEAD` reads as the spec.

apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,337 @@
...
+    );
+  }
+
+  // Sort by start so we can splice slices of the original html in order.
+  // Link and script regions cannot overlap in a well-formed document
+  // (the HTML parser disallows nested <script> / <link>), so we don't
+  // need overlap-resolution logic.
+  pending.sort((a, b) => a.start - b.start);
+
+  const replacements = await runWithConcurrency(
```
lefarcen, line 203: **P2** — `maxTotalBytes` is checked after all candidate replacements are already read + materialized in memory. A project with 500 `<link>`/`<script>` tags can force hundreds of MiB of reads/strings before this line throws. Fix by tracking projected output bytes during candidate processing and aborting early, or exposing asset size before reading so oversized siblings can be skipped without buffering.

apps/daemon/src/import-export-routes.ts
```diff
@@ -345,6 +352,116 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
...
+          res,
+          415,
+          'UNSUPPORTED_MEDIA_TYPE',
+          'export endpoint only supports HTML files',
+        );
+      }
+
+      const fileReader: InlineAssetReader = async (sibling) => {
+        try {
+          const siblingFile = await readProjectFile(
```
lefarcen: **P2** — `fileReader` buffers each full sibling file via `readProjectFile(...).buffer.toString('utf8')` before `maxAssetBytes` can decide to fall back. A single huge asset is fully read into memory before being left as a URL ref. Consider exposing file size from `readProjectFile` metadata so the helper can check size before decode, or pass a size-limited stream reader.

lefarcen (commented):
Round 3 makes significant progress on both items — caps are now enforced and the quoted-attr scanner is fixed. One resource-exhaustion edge case remains:

**P2 — Cap enforcement timing**: The helper enforces `maxTotalBytes` only after all candidate assets have already been read and converted to replacement strings (line 177). `runWithConcurrency` reads + buffers every candidate before the total is checked. A project with 500 relative assets can force hundreds of MiB of reads/replacement-strings before a 413 is thrown. Similarly, `maxAssetBytes` is checked after `fileReader` fully buffers each sibling file via `readProjectFile(...).buffer.toString('utf8')` (route line 425).

Fix by checking asset size before reading/decoding when possible, or changing the reader contract to expose size, and account projected output bytes during candidate processing with early abort.

The P3 quoted-attr fix looks solid, and the Round 2 tag-corruption fix remains intact. This is close — the remaining item is about resource bounding before materialization, not correctness of the final output.

bankielewicz:
Round 4 pushed — the cap-timing P2 is closed with a contract change that lets enforcement fire pre-buffer.

| Concern | Commit | Note |
|---|---|---|
| **P2 — caps enforce only post-materialization** at `inline-assets.ts:177` (round 3) | `bc5fd01d` | `InlineAssetReader`'s return type changed from `Promise<string \| null>` to `Promise<AssetHandle \| null>` where `AssetHandle = { readonly size: number; read(): Promise<string \| null> }`. Callers now expose `size` from a cheap stat-equivalent and defer materialization to `read()`. The helper checks `size` against `maxAssetBytes` BEFORE invoking `read()`, and against the running total BEFORE the reservation is committed. The route closure now does `resolveProjectFilePath` (stat) first, then `readProjectFile` (full buffer) only inside the deferred `read()`. A 100 MiB sibling file is now rejected after the cheap stat call, not after a 100 MiB `readFile`. |

### Enforcement flow inside `runWithConcurrency`

```
1.  await fileReader(p.resolved)              ← cheap stat-only call
2.  if (handle.size > maxAssetBytes) return null   ← pre-buffer
3.  if (runningBytes + handle.size > maxTotalBytes)
       totalAborted = true; return null            ← pre-buffer
4.  runningBytes += handle.size                    ← reserve
5.  await handle.read()                            ← only now
6.  if (read returned null) runningBytes -= refund
```

`totalAborted` is a shared flag the workers check at entry, so once the running total hits the cap, no new reads start. With `maxReadConcurrency = 8`, at most ~8 stat-side calls finish after abort — peak memory bounded by `maxTotalBytes + (concurrency × stat overhead)`. The concat-time guard stays as the exact final assertion (the pre-buffer reservation is approximate — it counts the original tag bytes and skips `<style>`/`<script>` wrapper overhead).

### New tests pin the pre-buffer semantics

- `maxAssetBytes is checked via handle.size BEFORE handle.read()` — uses a reader whose `read()` throws; the test only passes if the helper never calls it.
- `stops dispatching reads once running total exceeds maxTotalBytes` — counting reader with `maxReadConcurrency: 1` observes ≤ 2 reads before the cap fires (owner ~760 bytes + first asset's 800-byte stat-size pushes running over the 1000-byte cap).

### Resulting peak-memory profile

With round-4 defaults and worst-case 500 assets at 5 MiB each:

| Stage | Round 3 peak | Round 4 peak |
|---|---|---|
| stat fanout | 0 (all reads in flight) | ≤ 8 stat calls |
| read-buffer materialization | up to 500 × 5 MiB = 2.5 GiB | ≤ 50 MiB (maxTotalBytes) + ~40 MiB (8 in-flight reads after abort) |
| concat | up to 50 MiB (caught by total cap) | up to 50 MiB (exact) |

### Verification

`pnpm guard` clean, `pnpm typecheck` exit 0, daemon delta is +2 tests (38 in `export-inline-route.test.ts`).

| Phase | Commit | Test delta |
|---|---|---|
| K — pre-buffer enforcement + AssetHandle contract | `bc5fd01d` | +2 unit (38 total in the test file) |

PR commit chain is now: round-1 (B-R / B-G / C-R / C-G), round-2 (F-R / F-G / G1-R/G / G2-R/G / G3-docs), round-3 (I3-R / I3-G / H3-R / H3-G), round-4 (K). 16 commits total; this round combines the contract change + Red tests + Green impl into one because the new tests can't compile without the new contract.

apps/daemon/src/import-export-routes.ts
```diff
@@ -345,6 +352,137 @@ export function registerProjectExportRoutes(app: Express, ctx: RegisterProjectEx
...
+          "query parameter 'inline=1' is required",
+        );
+      }
+
+      const project = getProject(db, req.params.id);
+      const relPath = (req.params as any)[0];
+
+      let file;
+      try {
+        file = await readProjectFile(PROJECTS_DIR, req.params.id, relPath, project?.metadata);
```
lefarcen: **P2** — Owner HTML is fully buffered before `maxOwnerBytes` can fire. `readProjectFile(...)` loads the requested HTML into memory, then `inlineRelativeAssets(...)` checks `ownerBytes`. A 100 MiB owner HTML still gets read/stringified before the 413. Fix by statting the owner with `resolveProjectFilePath` before `readProjectFile` and rejecting over `MAX_INLINE_OWNER_BYTES`.

apps/daemon/src/inline-assets.ts
```diff
@@ -0,0 +1,395 @@
...
+      // Per-asset cap, pre-buffer (stat-based — no read yet).
+      if (handle.size > maxAssetBytes) return null;
+      // Total cap, pre-buffer. The check + write below has no `await`
+      // between them, so under Node's single-threaded event loop the
+      // pair is atomic with respect to other workers — no race.
+      if (runningBytes + handle.size > maxTotalBytes) {
+        totalAborted = true;
+        return null;
+      }
+      runningBytes += handle.size;
```
lefarcen, line 219: **P3** — `maxTotalBytes` trusts `handle.size` and never reconciles with actual `content` size. A stale handle size or UTF-8 decode expansion can let many replacement strings materialize before the final concat guard throws. Consider computing actual bytes and adjusting the reservation if `content` is larger than expected.

lefarcen (commented):
Round 4 makes excellent progress — the sibling-asset P2 from Round 3 is fixed. The new `AssetHandle` contract with `size` + deferred `read()` lets caps fire pre-buffer for siblings. One remaining resource-exhaustion edge case:

**P2 — Owner HTML still fully buffered**: The route loads the owner HTML with `readProjectFile(...)` at line 403, then `inlineRelativeAssets(...)` checks `ownerBytes`. A 100 MiB owner HTML still gets read/stringified before the 413. Fix by statting the owner with `resolveProjectFilePath` before `readProjectFile` and rejecting over `MAX_INLINE_OWNER_BYTES`, or add a bounded read API.

**P3 — Reservation vs actual size reconciliation**: The helper trusts `handle.size` and never reconciles with actual `content` size unless the single-asset cap is exceeded. A stale handle size or UTF-8 decode expansion can let many strings materialize before the final concat guard throws. Consider computing actual bytes and adjusting the reservation.

The Round 3 tag-corruption fix and quoted-attr fix remain intact. This is very close — the remaining P2 is about bounding the owner HTML read, which mirrors the sibling-asset fix you just applied.

bankielewicz:
Two clarifications before pushing round 5 — both about the items you flagged in the round-4 review.

**1. Item 1 (owner pre-buffer)** — confirming the shape: the route gains a `resolveProjectFilePath(...)` call before `readProjectFile`, rejects with `413 PAYLOAD_TOO_LARGE` if `meta.size > MAX_INLINE_OWNER_BYTES`, and (as a natural consequence of having the mime in hand from the stat) also moves the `text/html` check pre-buffer (currently a post-buffer `415 UNSUPPORTED_MEDIA_TYPE` at `apps/daemon/src/import-export-routes.ts:410-417`). Helper's `ownerBytes` check stays in place as defense-in-depth for callers that don't stat first. Is the mime-pre-buffer fold-in welcome in the same commit, or do you want it kept out of scope for this round?

**2. Item 2 (reservation reconciliation)** — failure mode when `actualBytes = Buffer.byteLength(content, 'utf8')` post-read pushes `runningBytes = runningBytes + (actualBytes - handle.size)` past `maxTotalBytes`:

  - **(a)** trigger worker returns null (drops inlining of that asset; tag stays as URL ref); `totalAborted = true` blocks subsequent worker entries; helper throws `InlineAssetsLimitError('total')` after `Promise.all` settles. Matches the round-2/3/4 pattern.
  - **(b)** trigger worker throws immediately, racing remaining workers.

Default is (a) unless you signal (b). Both prevent further memory escalation; the test surface differs in whether the trigger asset's body appears in the partially-built output, so worth pinning before I code.

Will hold the round-5 push until you respond. Thanks for the careful reviews — this is shaking out a much tighter contract than the round-1 shape.

lefarcen:
**Re: Item 1 (owner pre-buffer)** — yes, the mime-pre-buffer fold-in is welcome in the same commit. It's a natural consequence of having stat metadata in hand, and moving the `text/html` check from post-buffer 415 to pre-buffer 400 reduces unnecessary reads. The defense-in-depth `ownerBytes` check staying in the helper is correct.

**Re: Item 2 (reservation reconciliation)** — **(a)** is the right pattern. Drop that asset gracefully (tag stays as URL ref), set `totalAborted=true` to block subsequent workers, then throw `InlineAssetsLimitError('total')` after `Promise.all` settles. This matches the round-2/3/4 graceful-fallback design for oversized assets. Option (b) would be harsher than necessary — the helper should exhaust existing in-flight reads before throwing, rather than racing them.

Both items are polish, not blockers. The core sibling-asset pre-buffer fix from Round 4 already closed the major resource-exhaustion gap. These are tightening the symmetry (owner) and the accounting precision (reconciliation).

bankielewicz:
Quick confirmation on Item 1 before I code: I'm reading your "415 to 400" as a paraphrase of the move rather than a status-code change, since 415 is the registered `ApiErrorCode` for content-type mismatch (`packages/contracts/src/errors.ts:11`) and round 3's drive-by fix specifically migrated this branch away from 400/`UNSUPPORTED_FILE_TYPE` to 415/`UNSUPPORTED_MEDIA_TYPE`.

Plan to ship: pre-buffer mime check rejects with `Content-Type: application/json` envelope `{ error: { code: "UNSUPPORTED_MEDIA_TYPE", … } }` at HTTP 415 — same wire format as round 3, just relocated pre-buffer.

If you actually want pre-buffer rejection to use `BAD_REQUEST`+400 instead, flag that here and I'll adjust the wire + test.

Path (a) for Item 2 is locked. Both items will land in the round-5 push.

lefarcen:
Quick clarifications on both items:

**Item 1 (owner pre-buffer)**: You're correct — the 415→400 phrasing was shorthand for the **move**, not a status code change. Keep the 415  envelope (the registered  for content-type mismatch). Your plan is spot-on: `resolveProjectFilePath` first, reject pre-buffer with 413 if owner size exceeds cap, fold in the mime check (still 415 on mismatch, now pre-buffer), then `readProjectFile` only when both gates pass. The mime pre-buffer fold-in is welcome in the same commit — it's a natural consequence of having the stat in hand.

**Item 2 (reservation reconciliation)**: Path **(a)** is correct. When the trigger asset's actual bytes push `runningBytes` past `maxTotalBytes` post-read, the worker returns null (drops that asset's inlining, tag stays as URL ref), sets `totalAborted = true` to block subsequent worker entries, and after `Promise.all` settles the helper throws `InlineAssetsLimitError('total')`. Matches the round-2/3/4 pattern — no throw-before-settle race, just graceful partial output + explicit total-exceeded signal.

Go ahead with round 5. Both items are now locked.

bankielewicz:
## Round 5 — pushed

Thanks for the confirmation on both items. Round 5 is up.

### Item 1 — Owner pre-buffer (P2)

[`c4ea872e`](https://github.com/nexu-io/open-design/pull/1312/commits/c4ea872e) wires `resolveProjectFilePath` ahead of `readProjectFile` in the route handler. The new order is:

1. `resolveProjectFilePath(...)` — cheap stat, returns `{ size, mime, ... }`.
2. `if (ownerMeta.size > MAX_INLINE_OWNER_BYTES)` → `413 PAYLOAD_TOO_LARGE` pre-buffer.
3. `if (!ownerMeta.mime.startsWith('text/html'))` → `415 UNSUPPORTED_MEDIA_TYPE` pre-buffer (mime fold-in).
4. `readProjectFile(...)` — only runs when both gates pass.

Size fires before mime so an oversize non-HTML file returns 413 (not 415). The helper's `ownerBytes` check at `inline-assets.ts:127-133` stays as defense-in-depth for direct in-process callers and for any drift between stat-reported size and the bytes actually returned by `readFile`.

Red test ([`48e785e9`](https://github.com/nexu-io/open-design/pull/1312/commits/48e785e9)) writes a 2 MiB+1 byte `.txt` fixture and asserts `413, not 415`. Pre-fix returns 415 (`AssertionError: expected 415 to be 413`); post-fix returns 413. That ordering — size check before mime check — is what makes the Red→Green observable: a same-size HTML fixture would 413 either way via the helper's defense-in-depth, so the oversize-AND-non-HTML combo is what pins the route's new stat-first contract.

### Item 2 — Reservation reconciliation (P3, path-a)

[`bd487090`](https://github.com/nexu-io/open-design/pull/1312/commits/bd487090) folds an actual-bytes reconciliation into the worker after `handle.read()`:

```ts
runningBytes += handle.size;        // pre-read reservation
const content = await handle.read();
if (content == null) { runningBytes -= handle.size; return null; }
const actualBytes = Buffer.byteLength(content, 'utf8');
if (actualBytes > maxAssetBytes) { runningBytes -= handle.size; return null; }
runningBytes += actualBytes - handle.size;   // <- reconciliation
if (runningBytes > maxTotalBytes) {
  totalAborted = true;
  return null;                       // drop inlining; tag stays as URL ref
}
return p.build(content);
```

Path-a as confirmed: drop that asset's inlining, set `totalAborted = true` so subsequent workers skip their `read()`, let `Promise.all` settle, then throw `InlineAssetsLimitError('total')` below. No throw-before-settle race; matches round-2/3/4's graceful-fallback pattern.

Red test ([`4f56b020`](https://github.com/nexu-io/open-design/pull/1312/commits/4f56b020)) uses a stat-lying reader (`handle.size = 10`, `read()` returns 1000 bytes), `maxTotalBytes = 500`, `maxReadConcurrency = 1` for deterministic timing, two scripts. Discriminator is read count:

- Pre-fix: helper trusts `handle.size`; both reads complete (running stays under cap based on reservation); concat-time guard catches the 2000+-byte assembly → throws `'total'` but only after both reads materialized. `reads === 2`.
- Post-fix: worker 1's reconciliation trips `totalAborted` immediately; worker 2 skips its `read()`. `reads === 1`.

Pre-fix Red: `AssertionError: expected 1, received 2`.

### Gates

- `pnpm guard` — exit 0.
- `pnpm typecheck` (workspace, raw exit, no pipe-mask) — exit 0.
- `pnpm --filter @open-design/daemon test` — **1744/1744 passing** (delta +2 from round-4: one integration test for L1, one unit test for L2).

### Disclosure (per repo's wiring-call coverage guidance)

The L1 integration test pins the 413 envelope shape and the 413-before-415 ordering, which together prove the route's pre-buffer gate fires. It doesn't directly assert "no `readProjectFile` call was made for the owner" — that structural property is verified by code review of the route's call order at `import-export-routes.ts:401-446`. The 2 MiB+1 byte fixture would otherwise still 413 via the helper's defense-in-depth ownerBytes check, which is why the oversize-AND-non-HTML combo is the chosen discriminator: only the post-fix code path can reach 413 before the mime check.

### Commit table

| Phase | Commit | Subject |
|---|---|---|
| L1-R | [`48e785e9`](https://github.com/nexu-io/open-design/pull/1312/commits/48e785e9) | Red — pin 413 before 415 for oversize non-HTML |
| L1-G | [`c4ea872e`](https://github.com/nexu-io/open-design/pull/1312/commits/c4ea872e) | Stat owner before readProjectFile to bound pre-buffer |
| L2-R | [`4f56b020`](https://github.com/nexu-io/open-design/pull/1312/commits/4f56b020) | Red — pin stat-vs-actual byte reconciliation |
| L2-G | [`bd487090`](https://github.com/nexu-io/open-design/pull/1312/commits/bd487090) | Reconcile reservation with post-read actual bytes |

lefarcen (commented):
Round 5 closes the loop! Both remaining items from Round 4 are resolved:

**Item 1 (owner pre-buffer)**: The route now stats the owner with `resolveProjectFilePath` at line 417 before `readProjectFile`, rejects `ownerMeta.size > MAX_INLINE_OWNER_BYTES` with 413 pre-buffer, and folds in the mime check (`text/html`) at line 445 — also pre-buffer. Size fires before mime so an oversize non-HTML file returns 413 (not 415), which is the Red→Green observable for the integration test at `export-inline-route.test.ts:666-691`. The helper's `ownerBytes` check at `inline-assets.ts:127-133` stays as defense-in-depth for direct callers.

**Item 2 (reservation reconciliation)**: The helper now computes `actualBytes = Buffer.byteLength(content, 'utf8')` at line 236 after `handle.read()`, reconciles the pre-read reservation with `runningBytes += actualBytes - handle.size` at line 246, and checks the reconciled total against `maxTotalBytes` at line 247. If exceeded, the worker returns null (drops that asset's inlining; tag stays as URL ref), sets `totalAborted = true`, and after `Promise.all` settles the helper throws `InlineAssetsLimitError('total')` at line 259. Path-a confirmed per your \#issuecomment-4424868413 follow-up. The unit test at `export-inline-route.test.ts:339-389` uses a stat-lying reader (`handle.size = 10`, `read()` returns 1000 bytes) with `maxReadConcurrency = 1` for deterministic timing; pre-fix both reads complete (trust stat → 2 reads), post-fix worker 1 trips `totalAborted` and worker 2 skips its `read()` (1 read). Discriminator is `reads === 1`.

The Red→Green TDD chain is intact through all 5 rounds. The PR is ready from a review perspective — all prior blockers (tag-corruption from Siri-Ray, CSP sandbox, stylesheet attrs, quoted-attr scanner, sibling-asset pre-buffer, owner pre-buffer, reservation reconciliation) are resolved with test coverage.

lefarcen:
@mrcfps This PR's current head bd487090 (Round 5) is waiting on your review.

Siri-Ray (approved):
@bankielewicz I reviewed the current `bd487090` head for the new daemon `/api/projects/:id/export/*?inline=1` route, the server-side inliner, CSP/resource-cap handling, and the expanded daemon unit/integration coverage. The latest changes address the prior resource-bounding and accounting feedback cleanly; the PR checks are green, and I did not find any remaining actionable issues in the changed ranges. Thanks for the careful TDD follow-through and for tightening this endpoint across the review rounds.

PR #1314, opened 2026-05-11
feat(web): Critique Theater Phase 8 (8 Theater components, barrel, role-keyed CSS)

Nagendhra-web:
<div align="center">

# 🎭 Critique Theater · Phase 8

**The visual surface. 8 React components, role-keyed tints, 86 tests green.**

phase
commits-1f8a70?style=flat-square)
tests
locales
lines
stacked-on

</div>

> **TL;DR.** Phase 7 landed the reducer and the SSE stream. Phase 8 paints the panel. Eight focused components, one per commit, each under 200 lines, all sharing the role-keyed \`data-role\` tinting pattern. The barrel exports the narrow public surface so Phase 9 can mount it with one import.

---

## 🧩 Component graph

\`\`\`mermaid
flowchart TB
  Stage["TheaterStage<br/><sub>top-level switch by phase</sub>"]
  Stage -- "phase=running" --> Ticker[ScoreTicker]
  Stage -- "phase=running" --> Round[RoundDivider × N]
  Stage -- "phase=running" --> Lane[PanelistLane × 5]
  Stage -- "phase=running" --> Btn[InterruptButton]
  Stage -- "phase=shipped/interrupted/failed" --> Collapsed[TheaterCollapsed]
  Stage -- "phase=degraded" --> Degraded[TheaterDegraded]

  Transcript[TheaterTranscript] -. read-only replay .-> Ticker
  Transcript -. read-only replay .-> Lane

  Barrel{{"index.ts barrel"}}
  Barrel -.- Stage
  Barrel -.- Collapsed
  Barrel -.- Degraded
  Barrel -.- Transcript
  Barrel -.- Btn

  style Stage fill:#7057ff,color:#fff
  style Barrel fill:#1f8a70,color:#fff
  style Transcript fill:#d4a017,color:#000
\`\`\`

---

## 🎨 Role-keyed tinting (no hex literals)

The plan asked for \`var(--ink-\${role})\` tokens. Rather than push five new entries through the design-system A1 / A2 / B-slot pipeline, every lane carries a \`data-role\` attribute and the CSS picks the right border tint from an existing semantic token via \`color-mix\`:

| Role | Token blended into the border |
|---|---|
| 🎨 designer | \`color-mix(in srgb, var(--accent) 60%, var(--border))\` |
| 🔍 critic | \`color-mix(in srgb, var(--amber) 60%, var(--border))\` |
[...]

PR #1315, opened 2026-05-11
feat(web): Critique Theater Phase 9 (drop-in mount wrapper, native i18n for de, ja, ko, zh-TW)

Nagendhra-web:
<div align="center">

# 🎭 Critique Theater · Phase 9

**One-line mount, four native locales. The settings toggle is queued for Phase 15.**

phase
commits-1f8a70?style=flat-square)
tests
locales
stacked-on

</div>

> **TL;DR.** Phase 8 built the eight Theater components, but nothing in the app rendered them. Phase 9 packages the subscription, the interrupt handshake, and the phase-aware swap into a single drop-in (`<CritiqueTheaterMount projectId enabled />`) so Phase 15 can mount it on the project view with one import. The same PR fills native translations for German, Japanese, Korean, and Traditional Chinese on top of the en + zh-CN seed from #1314.

---

## 🤔 Why

Phase 8 (#1314) landed the eight Theater components: stage, collapsed badge, degraded banner, panelist row, transcript bubble, focused-defect chip, run-summary card, interrupt button. They all render against the Phase 7 reducer (#1307). What was still missing: the wiring that turns those components into something the project view can actually show.

Without this PR the Theater is dead code on disk. `useCritiqueStream`, the interrupt handshake, the phase-aware swap from live stage to collapsed badge to degraded banner, and the SSE teardown when the flag flips off, all of that has to live somewhere before the M1 rollout slice in Phase 15 can flip the gate on. Threading every piece into `ProjectView.tsx` directly would balloon that file's diff and force every test of it to deal with a Theater dependency the project view does not own.

Wrapping the wiring in a single component does two things at once:

1. Phase 15's mount-point change in `ProjectView` collapses to a one-line import, so the rollout flag PR stays focused on the gate (env-var read, persistence, settings panel) instead of being half wiring and half rollout.
2. The wrapper is the only place that subscribes to SSE for the Theater. When `enabled` flips to false mid-run, the wrapper unmounts and the EventSource closes. No leaked listener, no stale reducer, no zombie debate panel sitting on a project the user has already navigated away from.

The locale work fills the same need on the i18n side. Phase 8 seeded en + zh-CN. The remaining four maintained locales (de / ja / ko / zh-TW) shipped with `...en` fallbacks, which means a German UI saw English Theater chrome and the collapsed badge said "Interrupted at round 1" in English in the middle of an otherwise translated app. Native copy for those four ships in this PR so the rollout in Phase 15 lands an actually-localized feature, not a partially-localized one.

---

## 👀 What users will see

**Today (M0 rollout):** nothing. The feature is gated behind the Phase 15 settings toggle, which is still off by default. No project view change, no extra panel, no new layout. This PR is pure plumbing.

**After Phase 15 flips the toggle on (M1):**

- When a critique run starts on a project, the project view renders the live debate stage with the panelists, transcript, focused defect chip, and the run-summary card from #1314.
- Clicking Interrupt (or pressing Escape from inside the stage region) sends a POST to the daemon interrupt endpoint. The button shows "Interrupting…" until the daemon acks; on a 2xx the run collapses to the post-run badge that names the round and composite score the user actually interrupted at; on rejection or a non-2xx the local state stays on the live stage so the user can retry.
- When a run finishes naturally (shipped / failed / degraded) the live stage swaps to the matching badge or banner. The phase-aware swap is owned by the reducer from #1307, not by ad-hoc conditional rendering.
- A user on a German / Japanese / Korean / Traditional Chinese locale sees native chrome on every Theater surface, not English fallbacks.

`<CritiqueTheaterMount>` accepts an `enabled` prop today, so the Phase 15 PR just has to feed the settings value into that prop. No second mount point, no parallel branch.

---

## 🎯 Surface area

- [x] Web runtime (`apps/web`): adds the mount wrapper and four locale blocks; no change to existing components or `ProjectView`.
- [ ] Daemon (`apps/daemon`)
- [ ] E2E (`e2e`)

Internal-only relative to the rest of the project. The wrapper is exported from the existing Theater barrel; nothing imports it yet, so this PR cannot regress any currently-rendered surface.

---

## 🧱 Mount-point shape

```mermaid
flowchart LR
  Settings["Settings toggle<br/><sub>OD_CRITIQUE_ENABLED<br/>(lands in Phase 15)</sub>"]
  Settings --> enabled["enabled prop"]
  enabled --> Mount["&lt;CritiqueTheaterMount&gt;"]

  subgraph Mount[" "]
    Hook["useCritiqueStream"]
    Phase{"reducer<br/>phase"}
    Hook --> Phase
  end

  Phase -- "idle" --> Null[null]
  Phase -- "running" --> Stage[TheaterStage]
  Phase -- "shipped/interrupted/failed" --> Collapsed[TheaterCollapsed]
  Phase -- "degraded" --> Degraded[TheaterDegraded]
  Stage -- "click or Esc" --> Kill[POST /interrupt<br/>2xx ack synthesizes interrupted]

  style Settings fill:#d4a017,color:#000
  style Mount fill:#7057ff,color:#fff
  style Kill fill:#c44,color:#fff
```

---

## 📦 What ships

<details open>
<summary><b>Task 9.1: <code>&lt;CritiqueTheaterMount&gt;</code> drop-in wrapper</b></summary>

A self-contained `<CritiqueTheaterMount projectId enabled />` wrapper that owns the SSE subscription, the kill-request handshake, and the phase-aware swap between the live stage, the collapsed badge, and the degraded banner.

**What the wrapper owns:**

- The `useCritiqueStream` subscription. Tears down cleanly when `enabled` flips to false, even mid-run.
- The kill-request handshake (wait-for-daemon-ack pattern). On click, the wrapper snapshots `runId`, `bestRound`, `composite`, sets `interruptPending` so the button shows "Interrupting…", POSTs to the daemon interrupt endpoint, and only dispatches the synthesized `interrupted` action when the response is 2xx. Rejection or a non-2xx (including the Phase 15 not-yet-landed 404) clears pending and leaves the run on the live stage so the user can retry.
- The phase-aware swap from `<TheaterStage>` to `<TheaterCollapsed>` to `<TheaterDegraded>` driven by the reducer phase from #1307.
- A `__reset__` dispatch on project / transcript boundary changes so a stream from a previous project cannot leak into a fresh one.
- An Escape handler scoped past input / textarea / contenteditable AND past any open dialog outside the Theater region (matches `[role="dialog"]:not([aria-hidden="true"])` and `[aria-modal="true"]`).

The wrapper accepts a `connectionFactory` test seam so the integration test drives synthetic SSE actions through the reducer without spinning up a real EventSource.
</details>

<details open>
<summary><b>Task 9.2: Native i18n in de, ja, ko, zh-TW</b></summary>

40 `critiqueTheater.*` keys translated natively in German, Japanese, Korean, and Traditional Chinese on top of the en + zh-CN seed from Phase 8. `critiqueTheater.userFacingName` ("Design Jury") and `critiqueTheater.interruptedSummary` (with `{round}` and `{composite}` placeholders) included in every native block.

| Locale | Status |
|---|---|
| 🇺🇸 en | ✅ source of truth |
| 🇨🇳 zh-CN | ✅ native (Phase 8 seed) |
| 🇩🇪 de | ✅ native (this PR) |
| 🇯🇵 ja | ✅ native (this PR) |
| 🇰🇷 ko | ✅ native (this PR) |
| 🇹🇼 zh-TW | ✅ native (this PR) |
| 🌐 id, fa, ar, es-ES, ru, pl, pt-BR, hu, fr, uk, tr, th | ⏳ fall back to en via `...en` spread until the locale-team pass |
</details>

<details>
<summary><b>Task 9.3: Settings UI toggle (intentionally deferred to Phase 15)</b></summary>

The plan lists a `OD_CRITIQUE_ENABLED` Settings checkbox plus a daemon-side settings PUT handler. That toggle is the M1 rollout step and its natural home is the Phase 15 rollout flag wiring PR, where the daemon-side env-var read, the persistence shape, and the Settings panel changes land together as one coherent rollout slice rather than being split half-and-half with this UI plumbing PR. `<CritiqueTheaterMount>` already takes an `enabled` prop today, so Phase 15 just has to feed the settings value into that prop.
</details>

---

## 📸 Screenshots

Intentionally none. The wrapper is not mounted anywhere in this PR (Phase 15 wires it into `ProjectView`), so there is no user-visible surface to capture on this branch. Screenshots of the live stage, the collapsed badge, the degraded banner, and the native locale chrome will land in the Phase 15 rollout PR where the toggle actually flips the surface on.

---

## ✅ Validation

Commands run on the current head before pushing:

| Command | Result |
|---|---|
| `pnpm guard` | clean |
| `pnpm --filter @open-design/web test tests/components/Theater/CritiqueTheaterMount.test.tsx` | 4 / 4 |
| `pnpm --filter @open-design/web test tests/components/Theater/Theater.test.tsx` | inherited Phase 8 suite, all green |
| `pnpm --filter @open-design/web test tests/components/Theater/state/sse.test.ts` | 22 / 22 (3 new wire-layer regression cases for unknown `ship.status`, unknown `panelist_close.role`, NaN `round_end.composite`) |
| `pnpm --filter @open-design/web test tests/i18n/locales.test.ts` | 5 / 5 across 18 locales |
| `pnpm --filter @open-design/web test tests/components/Theater/InterruptButton.test.tsx` | green (Escape scoping past dialogs, aria-hidden ignored, Theater-internal Escape still fires) |

The four new `CritiqueTheaterMount` cases:

- Renders nothing while disabled (M1 gate stays closed).
- Renders nothing while idle, then mounts the live stage on `run_started`.
- Click on Interrupt waits for the daemon ack, then synthesizes `interrupted` with the correct round / composite pair (multi-round non-monotonic case: round 1 @ 8.5, round 2 @ 6.0 → "Interrupted at round 1", not round 2).
- Tears down the SSE connection cleanly when `enabled` flips to false (mid-run disable cancels the in-flight stream).

No packaged surface touched, so packaged mac / windows smoke is correctly skipped on this PR.

---

## 🔗 Stacked on #1314 (which stacks on #1307)

```
main
  └─ #1307  Phase 7 (reducer + hooks)
       └─ #1314  Phase 8 (8 components + barrel)
            └─ this PR  Phase 9 (mount wrapper + native i18n)
```

When the upstream PRs merge the diff here collapses to just the Phase 9 additions.

---

## ⏭️ What's next

```
✅ Phase 9 → mount wrapper + native i18n in de/ja/ko/zh-TW (this PR)
🔜 Phase 10 → adapter conformance harness + synthetic fixtures
   Phase 11 → Playwright e2e + visual regression + a11y
   Phase 13 → perf gates (size-limit, bench, ts-prune, coverage walker)
   Phase 14 → docs (user guide + spec/arch/skills/adapters/roadmap + README/AGENTS)
   Phase 15 → rollout flag wiring (also lands the Settings UI toggle deferred from 9.3)
```


apps/web/src/components/Theater/CritiqueTheaterMount.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  const onInterrupt = useCallback(() => {
+    if (interruptPending) return;
+    if (state.phase !== 'running') return;
+    setInterruptPending(true);
+    // Synthesize an interrupted action locally so the UI reflects the
+    // user's intent immediately. The real kill request to the daemon
+    // lands in Phase 15's rollout wiring; for now the optimistic
+    // dispatch keeps the surface honest about user intent.
+    dispatch({
```
lefarcen: P1 — This makes Interrupt a local-only terminal transition: no request is sent to `POST /api/projects/:id/critique/:runId/interrupt`, and because terminal reducer states are sticky, later daemon `ship`/`failed`/`interrupted` events for the real still-running job will be ignored. That can hide an active agent run, waste tokens, and leave UI/DB state divergent. Keep `interruptPending` as the optimistic/transient state, call the daemon interrupt endpoint, and only enter `phase: 'interrupted'` from the SSE event (or reset pending/report an error if the POST fails).

apps/web/src/components/Theater/state/sse.ts
```diff
@@ -0,0 +1,138 @@
...
+ *      function. That predicate is the contract-level source of truth for
+ *      "this is a recognised event with a non-empty runId"; if the cast
+ *      fails (missing runId, unknown type), we drop the frame and the
+ *      reducer never sees it.
+ */
+export function sseToPanelEvent(eventName: CritiqueSseEventName, data: unknown): PanelEvent | null {
+  if (data === null || typeof data !== 'object') return null;
+  const type = eventName.slice('critique.'.length);
+  const candidate = { ...(data as Record<string, unknown>), type };
+  return isPanelEvent(candidate) ? candidate : null;
```
lefarcen, line 69: P1 — `isPanelEvent` only checks `type` and non-empty `runId`; it does not validate the variant payload. A frame like `critique.run_started` with `{ runId: 'r' }` passes this guard, then the reducer hits `action.cast.slice()` and crashes the React tree. Either strengthen the contract guard to validate every variant field or add a per-event shape validator here before calling `onEvent`; tests should cover missing `cast`/`maxRounds`, bad numeric fields, and malformed `artifactRef`.

apps/web/src/components/Theater/hooks/useCritiqueStream.ts
```diff
@@ -0,0 +1,78 @@
...
+): UseCritiqueStreamResult {
+  const [state, dispatch] = useReducer(reduce, initialState);
+  const dispatchRef = useRef(dispatch);
+  useEffect(() => {
+    dispatchRef.current = dispatch;
+  }, [dispatch]);
+
+  const factory = options.connectionFactory ?? createCritiqueEventsConnection;
+
+  useEffect(() => {
```
lefarcen, line 57: P2 — The connection lifecycle resets on `enabled`/`projectId`, but the reducer state does not. If a user switches from project A to project B after a run has started or finished, the hook can keep rendering project A's Theater state until project B emits a new `run_started`; disabling and re-enabling has the same stale-state problem. Reset the reducer to `initialState` when the stream boundary changes or closes, and add a test that switches projects after a non-idle state.

apps/web/src/components/Theater/hooks/useCritiqueReplay.ts
```diff
@@ -0,0 +1,256 @@
...
+  // Playback cursor lives in a ref so pause -> resume picks up where the
+  // last tick left off instead of replaying from the start. A fresh
+  // transcript URL resets it to zero (see the parse effect below).
+  const cursorRef = useRef(0);
+
+  // Parse effect: own the fetch + parse. Runs only when the URL changes.
+  // Stores the parsed events in component state so the pace effect can
+  // react to them; cursor is reset because a new transcript is a new run.
+  useEffect(() => {
+    if (!transcriptUrl) {
```
lefarcen, line 102: P2 — Clearing or changing `transcriptUrl` resets metadata, events, and the cursor, but not the reducer state. After replaying one transcript to `shipped`, rendering with `null`, an empty transcript, or a failed fetch can still expose the previous replay state to `TheaterTranscript`. Reset the reducer when `transcriptUrl` changes or becomes null/error, and cover the URL-change/null case in the hook tests.

apps/web/src/components/Theater/InterruptButton.tsx
```diff
@@ -0,0 +1,50 @@
...
+ * (the daemon is already processing the interrupt) and `done` (the
+ * run has already terminated), so a frustrated double-tap on Esc
+ * never queues a second kill.
+ */
+export function InterruptButton({ pending = false, done = false, onInterrupt }: Props) {
+  const t = useT();
+
+  useEffect(() => {
+    if (done) return;
+    const handler = (evt: KeyboardEvent) => {
```
lefarcen, line 26: P2 — This global Escape handler fires regardless of focus target. If the user is focused in the prompt textarea, a select, a text input, or an editor elsewhere on the page, pressing Esc can interrupt the critique run unexpectedly. Scope the key handler to the Theater surface/button focus, or ignore events whose target is an input/textarea/select/contentEditable element.

apps/web/src/i18n/types.ts
```diff
@@ -1469,4 +1469,44 @@ export interface Dict {
   'sketch.closeConfirm': string;
   'sketch.textPrompt': string;
   'sketch.textModalTitle': string;
+  // Critique Theater (Phase 8 components; Phase 9 fills non-English locales)
+  'critiqueTheater.title': string;
```
lefarcen: P2 — The normative spec says the user-facing label "Design Jury" must come from a single `critiqueTheater.userFacingName` key so UI/settings/docs can rename it without code churn. This adds `critiqueTheater.title` instead and the repo has no `userFacingName` key at all. Add the spec key and have UI copy that needs the product label reference it, or update the spec in the same stack if `title` is now intended to be the swap point.

lefarcen (changes requested):
Thanks for keeping this phase isolated: the component boundaries are clean, the tests cover the happy-path reducer/SSE integration well, and the locale placeholder alignment work is a good fit for the existing i18n registry.

Lens A: I found two blocking correctness issues. Interrupt currently only changes local UI state and never calls the daemon interrupt endpoint, which can leave the agent running while the UI permanently ignores the real terminal outcome. The SSE guard also does not actually validate event shapes before dispatching into the reducer, so malformed frames can still crash the Theater surface.

Lens B: No docs/spec files changed in this PR, but I checked the Critique Theater spec/plan for consistency. The new i18n keys miss the spec's single `critiqueTheater.userFacingName` swap point for the user-facing "Design Jury" label.

I also left P2 comments for stale reducer state across project/transcript changes and Escape scoping, since those can surface as cross-project stale UI or accidental interrupts once this wrapper is mounted.

lefarcen (dismissed):
All findings from the previous review are addressed:

✅ P1 blockers fixed:
- Interrupt now POSTs to daemon endpoint (with 404 fallback for Phase 15 not-yet-landed)
- SSE validation now runs per-variant shape checks before reducer dispatch

✅ P2 issues fixed:
- Stream/replay hooks reset reducer on boundary changes
- Escape handler scoped to avoid firing from input/textarea/contenteditable
- i18n key renamed to `critiqueTheater.userFacingName` per spec

Nice iteration — the test coverage for the new fixes (interrupt POST, Escape scoping, reset on swap) is solid. Ready for human review.

lefarcen (dismissed):
All findings from the previous review are addressed:

✅ **P1 blockers fixed:**
- Interrupt now POSTs to `/api/projects/:id/critique/:runId/interrupt` (with 404 fallback when Phase 15 daemon endpoint not yet landed)
- SSE validation runs `hasValidVariantShape` per-variant checks before reducer dispatch

✅ **P2 issues fixed:**
- Stream/replay hooks dispatch `__reset__` on project/transcript boundary changes
- Escape handler scoped to skip firing from input/textarea/select/contenteditable
- `interruptPending` cleared via ref-effect when `state.runId` changes
- i18n key renamed to `critiqueTheater.userFacingName` per spec

Nice iteration — the test coverage for the fixes (interrupt POST with seam, Escape scoping, reset on swap, pending-clear on run change) is solid. Ready for human review.

apps/web/src/components/Theater/CritiqueTheaterMount.tsx
```diff
@@ -0,0 +1,132 @@
...
+  return best || state.activeRound;
+}
+
+function bestCompositeOf(state: Extract<CritiqueState, { phase: 'running' }>): number {
+  let best = 0;
+  for (const r of state.rounds) {
+    if (typeof r.composite === 'number' && r.composite > best) best = r.composite;
+  }
+  return best;
+}
```
PerishCode: `bestRoundOf` returns `r.n` of the *last* round in the (ascending-sorted) `rounds` array whose composite was set — i.e. the most recently completed round, not the round with the highest composite. `bestCompositeOf` separately finds the actual maximum composite. With rounds 1 → 5.0, 2 → 9.0, 3 → 7.0 finished before the interrupt, the synthesized action becomes `{ bestRound: 3, composite: 9.0 }`, and `TheaterCollapsed` renders that pair verbatim into `critiqueTheater.interruptedSummary` (`'Interrupted at round {round} · best composite {composite}'`, see TheaterCollapsed.tsx:52-54 and en.ts:1159), so the user sees "Interrupted at round 3 · best composite 9.0" — but round 3's composite was 7.0, not 9.0.

Why it matters: terminal phases are sticky in the reducer, so this synthesized event is what the user permanently sees for any user-initiated interrupt (the daemon's later real `interrupted` event is dropped). Since the badge advertises a round/composite *pair*, the pair should be consistent.

Suggested change: compute both fields in a single pass so they refer to the same round.

```ts
function bestRoundAndComposite(
  state: Extract<CritiqueState, { phase: 'running' }>,
): { round: number; composite: number } {
  let bestRound = 0;
  let bestComposite = 0;
  for (const r of state.rounds) {
    if (typeof r.composite === 'number' && r.composite > bestComposite) {
      bestComposite = r.composite;
      bestRound = r.n;
    }
  }
  return { round: bestRound || state.activeRound, composite: bestComposite };
}
```

Then `dispatch({ type: 'interrupted', runId: state.runId, bestRound: result.round, composite: result.composite })`. A small reducer test fixture with non-monotonic composites would pin the fix.


apps/web/src/i18n/locales/de.ts
```diff
@@ -875,6 +875,46 @@ export const de: Dict = {
...
+  'critiqueTheater.roleA11y': 'Barrierefreiheit',
+  'critiqueTheater.roleCopy': 'Text',
+  'critiqueTheater.roundLabel': 'Runde {n} von {m}',
+  'critiqueTheater.mustFix': '{n} Pflichtkorrekturen',
+  'critiqueTheater.composite': 'Gesamtscore',
+  'critiqueTheater.threshold': 'Schwelle',
+  'critiqueTheater.consensus': 'Konsens',
+  'critiqueTheater.interrupt': 'Abbrechen',
+  'critiqueTheater.interrupting': 'Wird abgebrochen…',
+  'critiqueTheater.interrupted': 'Abgebrochen',
```
PerishCode, line 938: This block translates 39 of the 40 `critiqueTheater.*` keys natively, but `critiqueTheater.interruptedSummary` (`'Interrupted at round {round} · best composite {composite}'`, en.ts:1159, also rendered by `TheaterCollapsed.tsx:52-54`) is missing. Same gap in `ja.ts`, `ko.ts`, and `zh-TW.ts` — confirmed by diffing the locale key sets against `en.ts`.

Why it matters: when a German / Japanese / Korean / Traditional Chinese user interrupts a critique, the collapsed badge silently falls back to the English string via `...en`, breaking the surrounding native UI. The PR description advertises 40 keys natively in all four new locales, so the count is also off by one. The locale-alignment test still passes because the spread covers the type contract; nothing in CI flags the regression.

Suggested change: add the missing key next to the existing `critiqueTheater.interrupted` line in each of the four new locales. Suggested copy:

- de.ts: `'critiqueTheater.interruptedSummary': 'In Runde {round} abgebrochen · bester Gesamtscore {composite}',`
- ja.ts: `'critiqueTheater.interruptedSummary': 'ラウンド {round} で中断 · 最高総合 {composite}',`
- ko.ts: `'critiqueTheater.interruptedSummary': '{round} 라운드에서 중단 · 최고 종합 {composite}',`
- zh-TW.ts: `'critiqueTheater.interruptedSummary': '已於第 {round} 輪中斷 · 最佳綜合分 {composite}',`


apps/web/src/components/Theater/InterruptButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+        if (typeof target.closest === 'function') {
+          if (target.closest('input, textarea, select, [contenteditable="true"]')) {
+            return;
+          }
+        }
+      }
+      onInterrupt();
+    };
+    window.addEventListener('keydown', handler);
+    return () => window.removeEventListener('keydown', handler);
```
PerishCode, line 70: The focus filter correctly skips Escape that originates inside an `input` / `textarea` / `select` / `[contenteditable="true"]` element, but the test `still fires Escape when focus is on a non-text element` pins the inverse: Escape on `document.body` always calls `onInterrupt`. Because the listener is attached to `window`, any non-text Esc anywhere in the document while the Theater is mounted will fire the kill request — including the very common case of "user opens a modal / popover / project switcher / dropdown and hits Esc to dismiss it" where focus is on a button or non-text element.

Why it matters: this PR turns Esc-to-dismiss into a side-channel that silently interrupts an in-flight critique. Phase 19 of the patch series explicitly framed the fix as "so the keybind only fires from the Theater area", but the implementation only filters by *target type*, not *Theater area*. As soon as `CritiqueTheaterMount` is wired into `ProjectView` in Phase 15, any unrelated Esc handler in the workspace can race with this one.

Suggested change: scope the keybind to the Theater subtree rather than `window`. Either drop the global listener and bind `onKeyDown` on the `<section class="theater-stage">` (so Esc only fires while the Theater region itself has focus), or — if the keybind needs to work from body focus — keep the listener global but additionally require `target?.closest('.theater-stage')` or that `target === document.body` *and* `document.activeElement === document.body`, and add a regression test where Escape fired from inside `<div role="dialog">` does not call `onInterrupt`.


PerishCode (commented):
Phase 9 looks tight: the mount wrapper is small, the kill request is now a real POST with a 404-tolerant fallback, the Escape handler is scoped past text-entry surfaces, and `interruptPending` is reset on `runId` change. The new de / ja / ko / zh-TW blocks line up with the en / zh-CN seed and the `userFacingName` rename is consistent across types, locales, and consumers.

Three non-blocking findings inline:
- `bestRoundOf` and `bestCompositeOf` don't agree on what "best" means, so the synthesized `interrupted` summary can show a round number that does not match the composite it's paired with.
- `critiqueTheater.interruptedSummary` is the one key from the Phase 8 critiqueTheater block that the four new locales don't translate; it silently falls back to English via `...en`.
- `InterruptButton`'s window-level Escape handler is scoped past text entry, but still fires on body-level Escape, which collides with Esc-to-dismiss on any non-text modal that happens to be open.

None of these block merge.



apps/web/src/components/Theater/CritiqueTheaterMount.tsx
```diff
@@ -0,0 +1,132 @@
...
+      onInterrupt={onInterrupt}
+      interruptPending={interruptPending}
+    />
+  );
+}
+
+function bestRoundOf(state: Extract<CritiqueState, { phase: 'running' }>): number {
+  let best = 0;
+  for (const r of state.rounds) {
+    if (typeof r.composite === 'number') best = r.n;
```
lefarcen: P2 `bestRoundOf` returns the last completed round, while `bestCompositeOf` returns the highest composite. If scores are non-monotonic, the collapsed interrupted badge can pair round 3 with round 2's composite. Please compute the best round and composite in one pass so both fields describe the same round, and add a non-monotonic regression case.

lefarcen (commented):
Hey @Nagendhra-web 👋

The Phase 9 wrapper is in much better shape on this head — the interrupt POST seam, run-id pending reset, stricter SSE guard, and locale rename all line up with the rollout direction. I found three remaining non-blocking polish/correctness items worth tightening before the stack lands; details are inline.

Looks close from our side — thanks for keeping the follow-up fixes so focused. ❤️

apps/web/src/components/Theater/InterruptButton.tsx
```diff
@@ -0,0 +1,68 @@
...
+        const tag = target.tagName;
+        if (tag === 'INPUT' || tag === 'TEXTAREA' || tag === 'SELECT') return;
+        if (target.isContentEditable) return;
+        if (typeof target.closest === 'function') {
+          if (target.closest('input, textarea, select, [contenteditable="true"]')) {
+            return;
+          }
+        }
+      }
+      onInterrupt();
```
lefarcen, line 67: P2 This global Escape handler still interrupts from any non-text target in the page, including modals/popovers/dropdowns that use Escape to dismiss. Please scope the keybind to the Theater subtree (or explicitly ignore dialog/popover surfaces) so unrelated workspace Escape interactions cannot silently cancel the active critique.

apps/web/src/i18n/locales/de.ts
```diff
@@ -922,6 +922,46 @@ export const de: Dict = {
...
```
lefarcen, line 938: P3 The native locale blocks still omit `critiqueTheater.interruptedSummary` in de/ja/ko/zh-TW, so interrupted runs fall back to English even though the PR advertises the full Theater block as translated. Please add this key next to `critiqueTheater.interrupted` in all four new native locales; keep `{round}` and `{composite}` placeholders intact.

lefarcen (commented):
Hey @Nagendhra-web 👋

This follow-up addresses the current review pass well: the interrupted summary now keeps round/composite paired, the new native locale key is present across de/ja/ko/zh-TW, and the Escape path now defers when an active dialog owns the page-level Escape interaction. I did not find any new blocking issues on this head.

Looks good from our side; CI is still finishing, so let's let that complete before the final maintainer call. ❤️

lefarcen (commented):
Hey @Nagendhra-web 👋

I re-checked the collapsed Phase 9 diff on `5112f16b`. The mount wrapper, interrupt handshake, dialog-aware Escape handling, strict SSE validation path, and native locale additions still line up cleanly, and the prior round/composite + locale + Escape items remain addressed. I did not find any new blocking issues on this head.

Looks good from our side; CI was still finishing when I checked, so the remaining gate is the normal human/CI path. ❤️

lefarcen (commented):
Hey @Nagendhra-web 👋

I checked `081779b1`, especially the restored wait-for-daemon-ack flow in `CritiqueTheaterMount`. The click-time snapshot, 2xx-only local terminalization, and rejection/non-2xx pending reset are the right shape here; they avoid the sticky optimistic `interrupted` state while still preserving the user's requested interrupt when the daemon acknowledges it. The updated tests cover the 204, rejection, and 404 paths cleanly.

No new blockers from our side on this head. Since this is an external PR, the remaining step is the formal human approval/merge path. ❤️

apps/web/src/components/Theater/state/sse.ts
```diff
@@ -33,33 +33,92 @@ export function critiqueEventsUrl(projectId: string): string {
...
+    case 'degraded':
+      return typeof event.reason === 'string' && typeof event.adapter === 'string';
+    case 'interrupted':
+      return typeof event.bestRound === 'number' && typeof event.composite === 'number';
+    case 'failed':
+      return typeof event.cause === 'string';
+    case 'parser_warning':
+      return typeof event.kind === 'string' && typeof event.position === 'number';
+  }
+}
```
PerishCode: The new `hasValidVariantShape` and the surrounding docstring rest on a claim about `isPanelEvent` that is no longer true on `main`.

**Problem.** The docstring at lines 54-60 says `isPanelEvent` from contracts "only checks `type` is known and `runId` is non-empty" and that a frame like `{ type: 'ship', runId: 'r' }` would otherwise slip through. The matching prose block above (the second numbered bullet of the rewritten `sseToPanelEvent` doc) says the same — that `isPanelEvent` drops frames only on "missing runId, unknown type". After Siri-Ray round-3 on PR #1314, `isPanelEvent` in `packages/contracts/src/critique.ts:190-257` is the strict guard: every variant's required fields, closed-enum membership (`SHIP_STATUS_SET`, `ROUND_DECISION_SET`, `DEGRADED_REASON_SET`, `FAILED_CAUSE_SET`, `PARSER_WARNING_KIND_SET`, `isPanelistRole`), non-negative finite numerics (rejects NaN / Infinity), `threshold <= scale`. The contracts test suite at `packages/contracts/tests/critique.test.ts:56-130+` pins all of that.

**Why it matters.** Two things drift together:

1. `hasValidVariantShape` is weaker than the very guard it claims to backstop. For `ship` it does `typeof event.status === 'string'`, but `isPanelEvent` already enforces `SHIP_STATUS_SET.has(event.status)`. For `panelist_open` it does `typeof event.role === 'string'`, but `isPanelEvent` already enforces `isPanelistRole`. For `round_end` it accepts any numeric `composite` including `NaN`/`Infinity`, but `isPanelEvent` rejects those via `isNonNegativeFinite`. So the second-pass filter cannot fail on any frame that `isPanelEvent` already lets through — it is a no-op layer.
2. The accompanying test removals in this PR pull out exactly the regressions the docstring claims this second pass exists to prevent — `apps/web/tests/components/Theater/state/sse.test.ts` loses `drops a critique.ship frame with an unknown status value`, the NaN-composite case, the unknown-role / unknown-reason / unknown-cause / unknown-kind / unknown-cast-role / non-finite-scale cases, and `accepts every well-formed enum value across the closed sets` (~170 deleted lines). Those scenarios are still covered at the contracts layer, but at the web-side wire boundary `sseToPanelEvent` no longer has a single test that proves an unknown enum value or a NaN numeric gets dropped. Future readers who trust the docstring will assume `hasValidVariantShape` is the safety net there, but it isn't — the safety net is `isPanelEvent`, one line above.

**Evidence.** From this PR's diff against main on sse.ts (RIGHT lines 54-110): the new docstring on `hasValidVariantShape` plus the function body, paired with the rewritten `sseToPanelEvent` doc-comment whose bullet 2 (RIGHT lines 34-38) says `isPanelEvent` drops only on "missing runId, unknown type". Compare against `packages/contracts/src/critique.ts:194-256`, where each variant validates closed-enum membership and finite numerics, and `packages/contracts/tests/critique.test.ts:90-140`, which pins that contract.

**Suggested change.** Pick one of:

- Drop `hasValidVariantShape` and restore the previous one-liner `return isPanelEvent(candidate) ? candidate : null;`. The doc-comment on `sseToPanelEvent` should also be reverted to the pre-PR version that correctly described `isPanelEvent` as the contract-level strict guard owning closed-enum membership and finite numerics. If you want one regression test left on the wire layer so a future weakening of `sseToPanelEvent` (e.g. someone bypassing `isPanelEvent`) gets caught, keep one or two of the deleted enum / NaN cases in `sse.test.ts`.
- Or, if you want a wire-layer second pass for defense-in-depth against a hypothetical future regression in contracts, make `hasValidVariantShape` strictly stronger than (or equal to) `isPanelEvent`: validate closed-enum membership and `Number.isFinite` everywhere `isPanelEvent` does, and update the docstring to say exactly that ("strict guard, mirrors `isPanelEvent`; here so the reducer is still safe if the contracts guard ever weakens"). The shadow guard then earns its keep.

Either way the docstring should stop describing `isPanelEvent` as a header-only check.


PerishCode (commented):
Hey @Nagendhra-web 👋

I re-checked `081779b1` against `origin/main` and the Phase 9 surface lands cleanly: the mount wrapper now waits for a 2xx daemon ack before terminalizing, `bestRoundAndComposite` pairs round and score in one pass, the dialog-aware Esc gate is in place, and `critiqueTheater.userFacingName` + `interruptedSummary` are present across en, zh-CN, de, ja, ko, and zh-TW. CI is green and the prior P1/P2/P3 threads are all addressed.

One non-blocking maintenance note inline on `sse.ts` about the new variant-shape validator drifting from the contract-level strict guard. Not a correctness issue, just stale code documentation that's worth tidying so the wire-layer doesn't gradually grow a parallel (and weaker) shadow of `isPanelEvent`.

Nothing here should block merge.



lefarcen (commented):
Hey @Nagendhra-web 👋

I re-checked the latest head `4ece81a7`, including PerishCode's follow-up on `sse.ts`. Dropping the shadow `hasValidVariantShape` layer and making `sseToPanelEvent` delegate directly to the contract-level strict `isPanelEvent` is the cleaner direction; the updated docs now accurately describe where enum/numeric validation lives, and keeping a small wire-boundary regression set in `sse.test.ts` preserves the important coverage without duplicating the whole contracts suite.

No new blockers from our side on this head. CI was still finishing during this pass, so the remaining path is CI + formal human approval/merge. ❤️

lefarcen (commented):
Hey @Nagendhra-web 👋

I re-checked `dcb8f55e` after the base merge. The remaining Phase 9 diff is nicely focused now: `sseToPanelEvent` delegates to the strict contract guard, the stale shadow validator is gone, and the three wire-boundary regression cases cover the important enum/numeric paths. No code blockers from our side on this head.

One process note before this gets to the final merge path: the new PR template asks for a few user-facing sections that are still missing here — **Why**, **What users will see**, **Surface area**, and **Validation**. No rush, but filling those in will make the product/release review much easier, especially for a feature PR with user-visible Theater behavior and locale changes. If you're using an AI assistant, this prompt should get the body into shape quickly:

> Read the current PR body for nexu-io/open-design#1315 and the template at
> Then update via `gh pr edit 1315 --repo nexu-io/open-design --body-file -`.

Once the body/product review is sorted, this is in good shape from the code-review side. ❤️

lefarcen (commented):
Hey @Nagendhra-web 👋

I re-checked `b84e1f94`. The description update now covers the template sections we needed — the Why / user-visible behavior / Surface area / screenshots rationale / validation story are much clearer. The code diff also stays clean: the `sse.ts` cleanup is unchanged, the wire-boundary regression tests are still focused, and the `FileViewer` test update just carries the required `projectKind` prop from main.

No new blockers from our side on this head. Product review has already been requested, and the remaining gates are product + qualified human approval. ❤️

alchemistklk (approved):

PR #1318, opened 2026-05-11
test(web): Critique Theater Phase 13 (reducer p99 bench + surface coverage walker)

Nagendhra-web:
# 🎭 Critique Theater · Phase 13

Reducer p99 bench plus critique surface coverage walker.

Two CI gates. One measures speed. One measures completeness.

**Stacks on:** #1317 (Phase 11). Base = `main`.

---

## ▣ Gate 1  ·  Reducer p99 budget

The Critique Theater reducer is a hot path: every SSE frame goes through it, and at the tail of a five-panelist run with four lifecycle events per panelist the reducer fires twenty-seven times per round. The spec budgets `2ms p99 per dispatch` on the CI baseline machine. This PR locks that budget behind a regression gate.

```
   ┌─────────────────────────────────────────────────────────────────┐
   │                                                                 │
   │     B U D G E T          2.0 ms                                 │
   │                                                                 │
   │     ASSERT THRESHOLD     4.0 ms       (2x slack for CI noise)   │
   │                                                                 │
   │     SAMPLE SIZE          10,000 dispatches                      │
   │                                                                 │
   │     SCRIPT               full happy SSE replay (27 actions)     │
   │                                                                 │
   └─────────────────────────────────────────────────────────────────┘
```

Speedometer reading (sample local run):

```
   p50    p90    p99      max
   ───────────────────────────────
   0.04   0.06   0.18    0.42      ms
                  ▲
                  └── trips at 4.0
```

Two slack reasons baked in:

1. CI runners share cores with neighbours. A noisy 90th percentile can spike 3x without anything in our code regressing.
2. The vitest harness measures wall-clock with `performance.now()`, not user-CPU. GC pauses count.

A real regression (10x slower from a destructured-clone in the wrong place, an accidental sync `fetch`, a never-cleared map) shows up as a 20ms or 200ms p99 and trips the gate immediately.

---

## ▣ Gate 2  ·  Surface coverage walker

Renames are the rot most likely to skip past code review. Someone renames `panelist_must_fix` to `panelist_blocker` in the reducer, the tests that drive `panelist_must_fix` go red, the engineer fixes the tests, the i18n key still references `panelist_must_fix`, the build passes, the screen reader says nothing on a blocker event.

This walker is the wall that catches that.

```
   31 named symbols                       ┌─────────────────────┐
   ──────────────────────                 │  walker iterates    │
   11  SSE event names                    │  every symbol       │
    5  panelist roles                     │                     │
    6  lifecycle phases                   │  for each:          │
    9  i18n keys                          │   ↳ assert in SRC   │
                                          │   ↳ assert in TEST  │
   62 assertions                          └─────────────────────┘
```

For each symbol we grep the **whole src tree** for at least one occurrence, then the **whole test tree** for at least one occurrence. The src side proves the symbol still ships. The test side proves at least one assertion still exercises it.

The walker accepts either the prefixed SSE name (`critique.panelist_must_fix`) or the unprefixed PanelEvent type (`type: 'panelist_must_fix'`) on the test side, because reducer tests dispatch the unprefixed form while contracts tests assert the wire form. Both forms still count.

---

## ▣ How to run

```bash
cd apps/web
pnpm vitest run tests/components/Theater/state/reducer-bench.test.ts
pnpm vitest run tests/components/Theater/critique-coverage.test.ts
```

```
                   ┌──────── reducer-bench ────────┐    ┌──── coverage ────┐
                   │                               │    │                  │
                   │   1 test          p99 OK      │    │  62 tests   OK   │
                   │                               │    │                  │
                   └───────────────────────────────┘    └──────────────────┘
```

Both files exit `0`. The reducer-bench test takes ~120ms wall (10k iterations of a tiny reducer is cheap). The coverage walker is filesystem-bound, takes ~80ms.

---

## ▣ The symbol table the walker is locking

<details>
<summary><strong>SSE event names</strong> (11)</summary>

`critique.run_started` · `critique.panelist_open` · `critique.panelist_dim` · `critique.panelist_must_fix` · `critique.panelist_close` · `critique.round_end` · `critique.ship` · `critique.degraded` · `critique.interrupted` · `critique.failed` · `critique.parser_warning`

</details>

<details>
<summary><strong>Panelist roles</strong> (5)</summary>

`designer` · `critic` · `brand` · `a11y` · `copy`

</details>

<details>
<summary><strong>Lifecycle phases</strong> (6)</summary>

`idle` · `running` · `shipped` · `degraded` · `interrupted` · `failed`

</details>

<details>
<summary><strong>i18n keys</strong> (9 named)</summary>

`critiqueTheater.userFacingName` · `critiqueTheater.shippedSummary` · `critiqueTheater.interruptedSummary` · `critiqueTheater.degradedSummary` · `critiqueTheater.interruptButton` · `critiqueTheater.scoreLabel` · `critiqueTheater.transcriptHeader` · `critiqueTheater.collapsedLabel` · `critiqueTheater.panelistGroup`

</details>

---

## ▣ Risk surface

| Concern | Mitigation |
|---|---|
| Walker fails on a legitimate rename mid-PR | Both lists live in this test file. A rename PR updates the list in the same diff that renames the symbol. Reviewer sees both changes side by side. |
| 4ms ceiling is too tight on slow CI | The 2x slack over the 2ms documented budget is already conservative. If the runner is so slow the gate flakes, the runner is unfit for the rest of the test suite too. |
| Walker false-positive on a substring match | Each lookup is an exact symbol match (string equality, not substring), so a key named `foo` does not satisfy a search for `foobar`. |

---

## ▣ Diff anatomy

```
 apps/web/tests/components/Theater/state/reducer-bench.test.ts  +  98
 apps/web/tests/components/Theater/critique-coverage.test.ts    + 142
```

Two new test files. Zero src changes.


e2e/package.json
```diff
@@ -6,7 +6,7 @@
...
     "test:ui:critical": "playwright test -c playwright.config.ts ui/entry-chrome-flows.test.ts ui/entry-configuration-flows.test.ts ui/project-management-flows.test.ts",
-    "test:ui:extended": "playwright test -c playwright.config.ts ui/app.test.ts ui/api-empty-response.test.ts ui/app-restoration.test.ts ui/app-manual-edit.test.ts ui/app-design-files.test.ts ui/settings-connectors-auth-happy-path.test.ts ui/settings-connectors-auth-recovery.test.ts ui/settings-api-protocol.test.ts ui/settings-local-cli-codex-fallback.test.ts ui/workspace-keyboard-flows.test.ts ui/examples-preview-core.test.ts ui/examples-preview-share.test.ts ui/real-daemon-run.test.ts",
+    "test:ui:extended": "playwright test -c playwright.config.ts ui/app.test.ts ui/api-empty-response.test.ts ui/app-restoration.test.ts ui/app-manual-edit.test.ts ui/app-design-files.test.ts ui/settings-connectors-auth-happy-path.test.ts ui/settings-connectors-auth-recovery.test.ts ui/settings-api-protocol.test.ts ui/settings-local-cli-codex-fallback.test.ts ui/workspace-keyboard-flows.test.ts ui/examples-preview-core.test.ts ui/examples-preview-share.test.ts ui/real-daemon-run.test.ts ui/critique-theater.test.ts",
```
Siri-Ray: Adding `ui/critique-theater.test.ts` to `test:ui:extended` makes the extended lane wait for a `Design Jury` region at `/`, but this PR only exports `CritiqueTheaterMount`; it does not mount it from `apps/web/app` or any routed `apps/web/src` surface. I verified the only production hit is the new Theater barrel/component, so the app at `/` will never open `/api/projects/*/events` or render the region this new suite asserts. That turns the existing extended Playwright lane red by timeout. Please either wire the mount into the real project surface behind the intended config flag before enabling this e2e file, or keep the new suite out of `test:ui:extended` until there is a routed app surface that renders the Theater.

apps/web/src/i18n/locales/de.ts
```diff
@@ -875,6 +875,46 @@ export const de: Dict = {
...
+  'critiqueTheater.degradedHeading': 'Jury für diesen Lauf offline',
+  'critiqueTheater.degradedReasonMalformed': 'Ungültige Jury-Ausgabe (Parser hat den Block abgelehnt).',
+  'critiqueTheater.degradedReasonOversize': 'Jury-Ausgabe hat das sichere Byte-Budget überschritten.',
+  'critiqueTheater.degradedReasonAdapter': 'Adapter {adapter} unterstützt das Critique-Protokoll nicht.',
+  'critiqueTheater.degradedReasonProtocol': 'Adapter {adapter} verwendet eine nicht unterstützte Protokollversion.',
+  'critiqueTheater.degradedReasonMissingArtifact': 'Der Lauf endete ohne ein finales Artefakt.',
+  'critiqueTheater.replay': 'Wiedergabe',
+  'critiqueTheater.replaySpeed': 'Wiedergabegeschwindigkeit',
+  'critiqueTheater.readOnly': 'Nur lesen',
+  'critiqueTheater.shippedSummary': 'In Runde {round} freigegeben · Gesamtscore {composite}',
```
Siri-Ray, line 949: The new `Dict` contract includes `critiqueTheater.interruptedSummary`, and `TheaterCollapsed` calls that key for interrupted runs, but this locale block skips straight from `critiqueTheater.shippedSummary` to `critiqueTheater.shippedBadge`. The same omission is present in the newly changed `ja`, `ko`, and `zh-TW` locale blocks, while `en` and `zh-CN` include the key. Because each locale is typed as `Dict`, the web typecheck will fail once dependencies are installed; without the key, interrupted summaries also have no localized string. Add `critiqueTheater.interruptedSummary` with `{round}` and `{composite}` placeholders to every locale touched by this PR, adjacent to `shippedSummary`, so the dictionary contract and runtime call site stay in sync.

Siri-Ray (commented):
I found two blocking issues in the current diff: one added CI lane is pointed at a Theater surface that the routed app never mounts, and one of the new i18n keys is missing from several locale dictionaries.


lefarcen (commented):
Hi @Nagendhra-web! 👋

Phase 13 looks solid on structure. Two CI gates (reducer p99 bench + surface coverage walker) are the right defensive moves for a hot-path reducer. I'm flagging a few items below for you to consider.

## ✅ What works

- **Gate 1 budget discipline**: 2ms documented budget with 4ms CI threshold (2x slack for noise) is conservative and testable. The 10k dispatch sample size is statistically meaningful for a p99 measurement.
- **Gate 2 symbol coverage**: 31 named symbols (11 SSE event names + 5 roles + 6 phases + 9 i18n keys) with 62 assertions is comprehensive. The grep-based walker catches rename drift before it ships.
- **Test isolation**: `reducer-bench.test.ts` and `critique-coverage.test.ts` are self-contained — no cross-test pollution.

## 🤔 Questions / Suggestions

1. **Missing performance baseline data (Lens B)** — The PR body documents the 2ms budget and 4ms threshold but doesn't cite the actual p99 reading from a sample local run (or CI baseline). Without a concrete number ("current p99 = 0.18ms"), reviewers can't verify the gate is set appropriately. Consider adding:
   - Baseline p99 from the CI machine (or a representative dev machine)
   - Sample size used for baseline measurement
   - Date of baseline capture

2. **No discussion of alternative approaches (Lens B)** — Gate 2 uses filesystem grep for symbol verification. Industry alternatives exist (TypeScript AST walker, ESLint plugin, build-time schema validation). Why grep over AST?
   - Grep is simple but brittle (false positives on string literals, comments).
   - AST-based tools catch structural changes (e.g., a symbol renamed in type but not runtime).
   - If grep was chosen for simplicity/speed, state that trade-off explicitly.

3. **Unclear failure mode: what happens when Gate 1 trips? (Lens A)** — `reducer-bench.test.ts` asserts p99 < 4ms, but:
   - Does the test output include p50/p90/p99/max in the failure message? (Helps triage: "p99 = 8ms" vs "p99 = 400ms" are different bugs.)
   - Is there a CI artifact or log that persists the full distribution?
   - If the gate trips on a noisy CI run, how does the team distinguish real regression from CI flake?

4. **Symbol table maintenance burden (Lens B)** — Gate 2's 31 symbols live in `critique-coverage.test.ts`. When a new SSE event or i18n key is added:
   - Who updates the test? (Is it the PR author's responsibility?)
   - How do they discover the test exists? (CONTRIBUTING.md link? PR template checklist?)
   - What's the failure mode if they forget? (Gate 2 trips on CI, blocking merge — is that acceptable UX?)

5. **Snapshot drift on symbol renames (Lens A)** — If a legitimate rename happens mid-PR (e.g., `panelist_must_fix` → `panelist_blocker`), the walker trips. The PR body says "Both lists live in this test file. A rename PR updates the list in the same diff." But:
   - What if the rename is buried in a 100-file refactor?
   - What if the PR author doesn't realize the walker exists?
   - Consider: a lint rule or pre-commit hook that auto-updates the symbol list on rename (reduces manual toil).

## ⚠️ Non-blocking nits

- **apps/web/tests/components/Theater/reducer-bench.test.ts**: The test name is `reducer-bench.test.ts` but the suite description should clarify it's a **p99 regression gate**, not a comparative benchmark. Consider renaming the suite to `'Critique Theater reducer p99 gate'` for clarity.
- **apps/web/tests/components/Theater/critique-coverage.test.ts**: The 62 assertions are listed in the file but not grouped by category (SSE events vs roles vs phases vs i18n). Consider adding `describe()` blocks for each category — easier to see what's missing at a glance.

---

Overall: this PR adds meaningful regression protection. The open questions above are about **completeness of reasoning** (Lens B) — alternatives, baselines, failure modes, maintenance burden. Addressing 1–3 would strengthen the gates; 4–5 are longer-term DX improvements.

Deferring final approval to a maintainer. Looking forward to seeing this land! 🚀

e2e/package.json
Nagendhra-web: Same fix as the codex thread on the same e2e/package.json edit: the upstream Phase 11 PR (#1317) parked the describe behind `test.describe.fixme` in commit 3053c1b, so the extended lane no longer waits on a Design Jury region that does not yet render. The describe flips back to live in the wireup PR that mounts `<CritiqueTheaterMount>` into App / ProjectView. Visible in #1318's diff once #1317 merges and this branch rebases.

lefarcen (changes requested):
Thanks @Nagendhra-web for the detailed follow-up on the Lens B questions and nits — most of those are addressed in commit 338a1857.

However, the **current head still carries the 2 P1 blockers** Siri-Ray identified earlier (missing e2e mount path + missing i18n keys in de/ja/ko/zh-TW). Your explanation references upstream PRs #1317 and #1316 as containing the fixes, but those fixes are NOT present in this branch's current head yet. Until you rebase onto main (after #1317/#1316 merge), this PR's diff still shows:

1. **P1 — e2e suite will timeout**: `e2e/ui/critique-theater.test.ts:124` is a live `test.describe` (not `.fixme`), and the suite navigates to `/` at line 132, but production never mounts `<CritiqueTheaterMount>` from `apps/web/src/App.tsx`. Your remote #1317 has `test.describe.fixme`, but this branch doesn't have that commit yet.

2. **P1 — typecheck will fail on missing i18n keys**: `apps/web/src/components/Theater/TheaterCollapsed.tsx:52` calls `t('critiqueTheater.interruptedSummary')`, and `Dict` requires it at `apps/web/src/i18n/types.ts:1497`, but `de.ts:901`, `ja.ts:900`, `ko.ts:1013`, and `zh-TW.ts:1078` omit that key. Your remote #1316 has the entries, but this branch doesn't have them yet.

Please rebase this branch onto main (once #1317 and #1316 merge) so the diff shows those fixes, then I'll re-review.

**New finding in 338a1857**:

3. **P2 — coverage gate weakened**: `apps/web/tests/components/Theater/critique-coverage.test.ts:139-144` now accepts unprefixed `type: 'panelist_must_fix'` for production SSE checks (lines 185-188), not just test-side reducer dispatches. This can let production coverage pass without a `critique.panelist_must_fix` wire-name reference, contrary to the comment at line 138. Consider scoping the unprefixed fallback to test corpus only.

4. **P3 — maintenance doc overclaims**: The flow at lines 39-56 says "forgetting step 4 causes CI red", but the expected symbols are manually defined at lines 154-180. If a contributor forgets to add a new symbol there, no assertion is generated — CI stays green. Clarify that the contributor must update both the production code AND the walker's symbol arrays in the same PR.

---

**Lens B questions status** (from my prior review):
- ✅ **Baseline data**: added at `apps/web/tests/components/Theater/state/reducer-bench.test.ts:28-32`.
- ✅ **Grep-vs-AST trade-off**: explained at `critique-coverage.test.ts:15-37`.
- ✅ **Failure visibility**: full distribution in assertion message at `reducer-bench.test.ts:137-140` and `159-162`.
- ⚠️ **Maintenance burden**: prose at `critique-coverage.test.ts:39-56` addresses intent, but see P3 above.
- ⏸️ **Pre-commit hook**: deferred (acknowledged as TODO).

**Nits**:
- ✅ **Describe rename**: done at `reducer-bench.test.ts:143`.
- ✅ **Group by category**: done at `critique-coverage.test.ts:184`, `200`, `216`, and `232`.

Once the rebase lands (bringing in #1317's `.fixme` + #1316's locale keys), and you address the P2 coverage-gate concern, this will be clean. 🚀

lefarcen (dismissed):
Thanks @Nagendhra-web — commit 76dc69e resolves both P2 and P3 cleanly.

**P2 (coverage gate weakened):** The split into `srcReferences` (strict exact-substring) and `testReferences` (lenient SSE fallback) is correct. Production assertions now route through `srcReferences` so the SSE wire name is load-bearing again. A production refactor that drops `critique.panelist_must_fix` while leaving `type: 'panelist_must_fix'` on a PanelEvent literal correctly fails the walker. ✅

**P3 (maintenance doc overclaim):** The failure-mode section now distinguishes 'rename existing symbol → CI red' from 'add new hand-maintained symbol → CI stays green', and clarifies that `SSE_EVENTS` and `PANELIST_ROLE_STRINGS` are auto-built from contracts while `PHASE_STRINGS` and `I18N_KEYS` are the hand-maintained groups. ✅

**P1s (e2e fixme + i18n keys):** Correctly parked on upstream PRs #1317 (commit 3053c1b) and #1316 (commit 9b72b54). Will appear in this PR's diff after those merge and this branch rebases onto main, per your earlier direction.

Ready for mrcfps's review. 🎉

lefarcen (changes requested):
Thanks @Nagendhra-web — but the upstream fixes from #1317 and #1316 are not yet in this branch. The two P1 blockers are still present on head 8374d275:

**P1 (e2e suite live):** `e2e/ui/critique-theater.test.ts:124` is still `test.describe(...)` (not `.fixme(...)`), and `CritiqueTheaterMount` has no app caller in `apps/web/src` → Playwright will timeout. #1317 commit 3053c1b added the fixme, but that commit is not in this branch's ancestry.

**P1 (i18n keys missing):** `critiqueTheater.interruptedSummary` exists only in `en.ts` and `zh-CN.ts`; `de`, `ja`, `ko`, `zh-TW` still lack it → typecheck will fail. #1316 commit 9b72b54 added the keys, but that commit is not in this branch's ancestry either.

**Action needed:** Rebase this branch onto main (or cherry-pick 3053c1b + 9b72b54) so the upstream fixes appear in this PR's diff. After the rebase, I'll re-review the merged diff.

No new issues in the 76dc69e→8374d275 push itself; the blockers are inherited from the missing upstream fixes.

apps/web/tests/components/Theater/critique-coverage.test.ts
```diff
@@ -0,0 +1,284 @@
...
+// at the repo root (Theater -> components -> tests -> web -> apps -> repo).
+const REPO_ROOT = path.resolve(path.dirname(__filename), '..', '..', '..', '..', '..');
+
+const SRC_ROOTS = [
+  path.join(REPO_ROOT, 'apps/web/src'),
+  path.join(REPO_ROOT, 'apps/daemon/src/critique'),
+  path.join(REPO_ROOT, 'packages/contracts/src'),
+];
+const TEST_ROOTS = [
+  path.join(REPO_ROOT, 'apps/web/tests'),
```
Siri-Ray, line 34: This corpus includes `apps/web/tests`, which also contains `apps/web/tests/components/Theater/critique-coverage.test.ts` itself. Because `readCorpus(TEST_FILES)` then reads this file, the hand-maintained literals in `PHASE_STRINGS` and `I18N_KEYS` satisfy the `tests reference %s` assertions without any independent Theater test mentioning those symbols. That breaks the main guarantee described in the PR body: the test side no longer proves an assertion exercises the symbol, so deleting an actual i18n or phase assertion can still leave the coverage gate green. Please exclude this walker file from `TEST_FILES` (or otherwise build the searched corpus from external test files only) before checking `TEST_CORPUS`, and consider keeping the symbol table in a structure that cannot be counted as evidence.

Siri-Ray (changes requested):
@Nagendhra-web thanks for continuing to tighten these gates. I found one blocking issue in the Phase 13 coverage walker: the test corpus currently includes the walker test file itself, so several hand-maintained symbols can satisfy their own coverage assertions instead of proving an independent test exercises them.


apps/web/tests/components/Theater/critique-coverage.test.ts
```diff
@@ -0,0 +1,284 @@
...
+    try {
+      return readFileSync(f, 'utf8');
+    } catch {
+      return '';
+    }
+  }).join('\n');
+}
+
+const SRC_FILES = SRC_ROOTS.flatMap(walk);
+const TEST_FILES = TEST_ROOTS.flatMap(walk);
```
lefarcen: P2 `TEST_FILES` includes `critique-coverage.test.ts` itself, so literals declared in the walker satisfy the test-side assertions without any real test coverage. For example, `critiqueTheater.interruptedSummary` only appears under `apps/web/tests` in this file, but the walker still passes. Exclude `__filename` from the walked test files (or keep the symbol table outside the searched corpus) so the test-side gate proves an independent test actually exercises each symbol.

apps/web/src/i18n/locales/de.ts
```diff
@@ -922,6 +922,46 @@ export const de: Dict = {
...
```
lefarcen, line 949: P2 `critiqueTheater.interruptedSummary` is still missing from the native locale overrides here, and the same gap remains in `ja`, `ko`, and `zh-TW`. The `...en` spread keeps typecheck green, but interrupted runs fall back to English while the neighboring Theater strings are localized. Please add the localized key with `{round}` and `{composite}` placeholders next to `shippedSummary` in all four native locale blocks.

lefarcen (changes requested):
Hey @Nagendhra-web 👋

The rebase moved this forward: the routed e2e suite is now parked behind `test.describe.fixme`, and the stricter src/test split for SSE names is the right direction. I found two remaining issues before this can unblock cleanly — see inline.

Once these are addressed, this should be ready for the next pass. ❤️

apps/web/tests/components/Theater/critique-i18n-keys.test.ts
```diff
@@ -0,0 +1,46 @@
...
+  'critiqueTheater.interrupted',
+  'critiqueTheater.degradedHeading',
+  'critiqueTheater.shippedSummary',
+  'critiqueTheater.interruptedSummary',
+] as const;
+
+describe('Critique Theater i18n key coverage (Phase 13.2 walker support)', () => {
+  for (const key of WATCHED_KEYS) {
+    it(`exposes ${key} on the en dictionary as a non-empty string`, () => {
+      const value = (en as Record<string, string>)[key];
```
Siri-Ray: This new test currently breaks the PR's required web typecheck. The live `Validate workspace` job on head `60ec5ce8` fails with `TS2352` here because `Dict` has no string index signature, so TypeScript rejects the direct `en as Record<string, string>` conversion. That means the new coverage gate cannot merge even though the runtime assertion is straightforward. Please avoid the broad incompatible cast, for example by typing `WATCHED_KEYS` as `readonly (keyof Dict)[]` and reading `const value = en[key]`, or by using a narrower helper that first casts through `unknown` only if that is intentional.

Siri-Ray (changes requested):
@Nagendhra-web thanks for the quick follow-up on the coverage walker and locale gaps. I found one remaining blocker: the latest head is still failing the web typecheck in the new i18n key support test, so the validation gate cannot pass yet.


apps/web/tests/components/Theater/critique-i18n-keys.test.ts
lefarcen: P1 CI typecheck is failing here: `Dict` cannot be cast directly to `Record<string, string>` (`TS2352`). Please either avoid the broad cast by typing `WATCHED_KEYS` as `readonly (keyof typeof en)[]` and reading `en[key]`, or use the explicit `en as unknown as Record<string, string>` form if you want the dynamic lookup. That should get the `@open-design/web` typecheck lane green again.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for the quick cleanup here. The two review findings are fixed: the walker no longer counts itself, the independent i18n key coverage fills the gap, and the native interrupted summaries are now present.

One small blocker remains because CI is currently failing typecheck on the new test file — see inline. Once that cast is adjusted, this should unblock my prior changes-requested state cleanly. ❤️

lefarcen (dismissed):
Hey @Nagendhra-web, this latest fix resolves my remaining blocker. The watched keys are now typed against the dictionary directly, the walker no longer counts itself as evidence, and the native interrupted-summary strings are present across the locale overrides.

No remaining code-review blockers from my side on this head. CI is still finishing, so we should let the required checks report before any merge call, but my prior changes-requested state is cleared. Nice iteration. ✨

apps/web/src/components/Theater/state/sse.ts
```diff
@@ -51,11 +51,74 @@ export function critiqueEventsUrl(projectId: string): string {
...
+    case 'round_end':
+      return typeof event.round === 'number'
+        && typeof event.composite === 'number'
+        && typeof event.mustFix === 'number'
+        && (event.decision === 'continue' || event.decision === 'ship')
+        && typeof event.reason === 'string';
+    case 'ship':
+      return typeof event.round === 'number'
+        && typeof event.composite === 'number'
+        && typeof event.status === 'string'
```
Siri-Ray, line 95: This variant guard still accepts enum-like fields as any string: `run_started.cast`/panelist `role`, `ship.status`, `degraded.reason`, `failed.cause`, and `parser_warning.kind` all pass as long as they are strings. That leaves malformed SSE frames able to enter the reducer with values outside the contract even though this new helper is meant to be the second-pass shape filter after `isPanelEvent`; for example an unknown `status` reaches `TheaterCollapsed` and indexes `SHIP_BADGE_KEY[final.status]`, and an unknown role can flow into role-keyed rendering/CSS without a valid label. Please validate these fields against the contract constants/unions (`PANELIST_ROLES`, the allowed ship statuses/reasons/causes/warning kinds) and add a fixture that proves unknown enum values are dropped before dispatch.

Siri-Ray (commented):
@Nagendhra-web thanks for the careful iteration on the Phase 13 gates. I found one merge-safe hardening issue in the SSE frame validator; the previously reported coverage-walker and i18n blockers look addressed on this head.


Siri-Ray (dismissed):
@Nagendhra-web I reviewed the current head and the changed ranges for the Phase 13 gates, including the reducer p99 budget test, coverage walker, i18n key support, Theater mount/interrupt changes, daemon conformance harness, and the parked e2e coverage. The prior coverage-walker, locale, typecheck, and SSE guard concerns look addressed on this head, and live CI is green; local validation could not run only because this checkout has no installed node_modules. Thanks for the careful iteration here.


apps/web/tests/components/Theater/critique-coverage.test.ts
```diff
@@ -0,0 +1,292 @@
...
+// dirname is apps/web/tests/components/Theater. Up 5 segments lands
+// at the repo root (Theater -> components -> tests -> web -> apps -> repo).
+const REPO_ROOT = path.resolve(path.dirname(__filename), '..', '..', '..', '..', '..');
+
+const SRC_ROOTS = [
+  path.join(REPO_ROOT, 'apps/web/src'),
+  path.join(REPO_ROOT, 'apps/daemon/src/critique'),
+  path.join(REPO_ROOT, 'packages/contracts/src'),
+];
+const TEST_ROOTS = [
```
Siri-Ray, line 33: The coverage walker is added under `apps/web/tests`, but this corpus reaches outside the web app into `apps/daemon/src/critique`, `apps/daemon/tests`, `packages/contracts/tests`, and `e2e/ui` (see `SRC_ROOTS`/`TEST_ROOTS`). That makes `@open-design/web`'s app-local Vitest suite depend on daemon-private and e2e file layout, which conflicts with the repo boundary that cross-app or repository-resource consistency checks belong in `e2e/tests/` (or in a pure package when the logic is reusable). It also means a daemon/e2e-only refactor can break the web package test lane even when the web app did not change. Please move this repository-wide symbol walker to an e2e consistency test, or narrow the web-local test to web-owned sources and put the cross-app daemon/contracts/e2e coverage assertion behind the e2e boundary.

Siri-Ray (commented):
@Nagendhra-web thanks for the careful follow-up on this stack. I found one merge-safe maintainability issue in the coverage walker boundary; the previously discussed correctness blockers look addressed on this head, and the live checks are green.


Siri-Ray (dismissed):
@Nagendhra-web I reviewed the current head and the changed ranges for the Phase 13 work, including the reducer p99 gate, the e2e-hosted critique coverage walker, the independent i18n key coverage, the locale updates, the stricter contract/SSE validation, the interrupt ack handling, and the daemon conformance/degraded-adapter tests. The prior self-counting walker, locale fallback, typecheck, boundary, and SSE validation concerns look addressed on this head; I did not find any new actionable code-review findings. Thanks for the careful iteration on this stack.


e2e/tests/critique-coverage.test.ts
```diff
@@ -0,0 +1,274 @@
...
+}
+
+const SRC_FILES = SRC_ROOTS.flatMap(walk);
+// The walker walks e2e/tests, which contains this file. Hand-maintained
+// PHASE_STRINGS and I18N_KEYS literals declared below would otherwise
+// satisfy the test-side coverage assertion against themselves, so a real
+// downstream test exercising a symbol could be deleted with the gate
+// still green. Exclude this file from TEST_FILES so the corpus only
+// holds independent evidence.
+const TEST_FILES = TEST_ROOTS.flatMap(walk).filter((f) => path.resolve(f) !== SELF_PATH);
```
Siri-Ray, line 155: This filter removes only the new e2e walker from `TEST_CORPUS`, but the branch still contains `apps/web/tests/components/Theater/critique-coverage.test.ts`, and `TEST_ROOTS` above includes all of `apps/web/tests`. That older walker declares the same hand-maintained symbol literals (`PHASE_STRINGS`/i18n keys), so deleting the real downstream tests for one of those symbols can still leave this new e2e gate green by matching the web walker file instead of independent coverage. Please either remove the legacy web-local walker as part of this move, or extend this filter to exclude all coverage-walker files from the test corpus so only independent tests can satisfy the test-side assertions.

Siri-Ray (commented):
@Nagendhra-web thanks for the careful rebase and for moving the cross-app walker into the e2e boundary. I found one merge-safe issue in the new e2e walker: it still counts the older web-local coverage walker as test evidence, so one part of the new gate can pass without an independent test actually exercising the symbol.


lefarcen (changes requested):
Hey @Nagendhra-web, thanks for the careful rebase notes and for keeping the Phase 13 scope tight through the main merge. The Phase 13 test files still look coherent, but the current head cannot merge while `Validate workspace` is red.

P1 CI blocker: the `@open-design/web` typecheck is failing in `tests/components/FileViewer.test.tsx:434` because `FileViewer` now requires `projectKind`, while that test render still passes only `projectId`, `file`, `isDeck`, and `liveHtml`. Please update that fixture to pass the appropriate `projectKind` (or otherwise align it with the current `FileViewer` props) so the required workspace validation is green again.

Separate template cleanup: since the new PR template landed while this branch was open, please refresh the PR body with the missing **Why**, **What users will see**, **Surface area**, and **Validation** sections when you push the CI fix. If helpful, you can paste this into your assistant:

> Update the body for `nexu-io/open-design#1318` using `.github/pull_request_template.md`. Fill **Why**, **What users will see**, **Surface area** (use `None` if this is test-only), and **Validation** with the commands actually run.

Once the typecheck is green, this should be straightforward to re-review.

lefarcen (approved):
Hey @Nagendhra-web, thanks for turning this around quickly. The `projectKind` test fixture update fixes the `Validate workspace` typecheck blocker from my last review, and the current required checks are green.

I also re-read the Phase 13 changed ranges on this head. I don’t see any remaining blocker from my side, so I’m clearing my stale changes-requested review. Siri-Ray’s latest coverage-walker note is still worth addressing as a merge-safe hardening follow-up, and the PR body template cleanup is still helpful for reviewer context, but neither is a blocker from my side now.

e2e/tests/critique-coverage.test.ts
Siri-Ray, line 155: The test corpus still excludes only this new e2e walker file, but `TEST_ROOTS` includes all of `apps/web/tests` and that tree still contains `apps/web/tests/components/Theater/critique-coverage.test.ts`. That legacy walker declares the same hand-maintained `PHASE_STRINGS` and `I18N_KEYS`, so a real downstream test for one of those symbols could be deleted while this gate stays green by matching the old walker file instead of independent coverage. Please either remove the legacy web-local walker as part of the move, or filter all coverage-walker files out of `TEST_FILES` so only independent tests can satisfy the test-side assertions.

Siri-Ray (commented):
@Nagendhra-web thanks for the quick typecheck fix on the rebase head. I found one remaining merge-safe issue in the e2e coverage walker boundary; the FileViewer prop fix looks aligned with the current component contract.


lefarcen (approved):
All prior findings resolved. P2 (srcReferences/testReferences split) and P3 (baseline data in docblock) are fixed on head 64e6d758. P1s (e2e fixme + i18n keys) came through via the upstream merges as you described. Clean. 🎉

qiongyu1999 (approved):

PR #1319, opened 2026-05-11
docs: Critique Theater Phase 14 (user guide + 2 AGENTS module maps)

Nagendhra-web:
# 🎭 Critique Theater · Phase 14

User guide plus two AGENTS module maps.

**Stacks on:** #1318 (Phase 13). Base = `main`.

> **About this PR's head (`803c3201`):** the branch carries the three approved Phase 14 doc commits, two doc tightenings on top of them (`d3a0b81`, `cbff22a`), three successive merges from `main`, and one orthogonal fix that came back when reviewers flagged a P1 interrupt race on the sibling PR #1316 (`803c3201`). The latest merge (`f60ad17a`) brought in the Phase 8 Critique Theater squash (#1314) plus the Tailwind landing-page foundation (#1388), the Media-picker header separation (#1441), the AssistantMessage status-block fix (#1413), the ghost CLI version-probe fix (#1301 / #658), the artifact preview exit-button alignment fix (#1445), and the landing-page per-facet split (#1158). **No new Phase 14 doc content** was added on top of `0aece0a3`. The three Phase 14 doc paths remain byte-identical against the approved tip: run `git diff 0aece0a3..HEAD -- docs/critique-theater.md apps/daemon/src/critique/AGENTS.md apps/web/src/components/Theater/AGENTS.md` to verify; expected output is empty.
>
> **Why a merge and not a rebase:** the same wave the rest of the stack absorbed. `#1379` refactored the feedback widget (`feedback.*` keys to `assistant.feedback*`, `.message-feedback*` CSS to `.assistant-feedback*`), so an unconditional rebase would reapply the older Phase 9 i18n + CSS surface and produce predictable conflicts. The merge resolves the web-side files with the same recipe the rest of the stack uses (drop stale `feedback.*` keys and `.message-feedback*` selectors; concatenate `critiqueTheater.*` with `main`'s new `.assistant-feedback*` block), without rewriting the previously approved doc tip.

---

### What this PR actually adds

| File | Audience | Lines |
|---|---|---|
| `docs/critique-theater.md` | end users (people running Design Jury) | 204 |
| `apps/daemon/src/critique/AGENTS.md` | daemon contributors | 63 |
| `apps/web/src/components/Theater/AGENTS.md` | web contributors | 75 |

Three documents, three readers. No file is written for more than one audience; that is the rule the prior phases were missing.

---

### `docs/critique-theater.md`  ·  the user manual

The document a user lands on after they flip the Settings toggle and want to know what the score badge means.

```
[...]

PR #1320, opened 2026-05-11
feat: Critique Theater Phase 15 (rollout resolver + Settings toggle hook)

Nagendhra-web:
<div align="center">

# 🎭 Critique Theater · Phase 15

**Rollout flag resolver on the daemon plus a Settings toggle hook on the web.**

phase
daemon-tests
web-tests
contracts-tests

</div>

> **TL;DR.** Two halves of the same gate sitting on opposite sides of the HTTP boundary: a pure `isCritiqueEnabled` resolver on the daemon that combines skill policy, project override, env var, and rollout phase, plus a `useCritiqueTheaterEnabled` hook on the web that reads the Settings toggle from `open-design:config` and stays in sync via same-tab CustomEvent and cross-tab `storage` events. No orchestrator wiring yet, that lands as a one-line edit in the follow-up PR.

---

## 🤔 Why

Phases 7-14 build the entire Critique Theater feature, but every component is gated behind `critiqueCfg.enabled` which today only reads `OD_CRITIQUE_ENABLED`. That's fine for a developer flipping an env var, but it leaves three problems unaddressed before M2 rollout:

1. **No skill-level veto.** An adapter that should never see critique (say, a media generator) has no way to opt out. Same goes for a skill that requires critique unconditionally. Today both are stuck with whatever the env var says.
2. **No per-project override.** A user who wants critique on for their design project but off for their note-taking project has no surface to express that. The toggle is global to the entire daemon.
3. **No staged rollout.** The product wants to ship in waves: M0 dark-launch (everyone off), M1 settings-toggle-only (off by default but exposable), M2 per-skill default-on, M3 global default-on. Today there's no machinery to encode any of those, so the only options are "feature flipper on" or "feature flipper off" for everyone.

The Phase 15 resolver puts all three concerns in one priority-ordered function that the orchestrator can call once per request, and the hook makes the user-visible Settings toggle a real first-class consumer of `localStorage` plus the existing settings PUT endpoint. The actual orchestrator wiring is left as a one-line edit in the follow-up so the resolver can land green, be reviewed on its own, and get the matrix tests pinned down separately from the request-path change.

---

## 👀 What users will see

**Today (M0 rollout):** nothing. The resolver is not yet wired into the orchestrator's spawn-time gate. `OD_CRITIQUE_ENABLED` continues to be the only effective switch. This PR ships only the underlying primitives.

**After the wireup PR lands and the orchestrator calls `isCritiqueEnabled`:**

- A user can flip "Critique Theater (beta)" in Settings and the toggle persists to `localStorage` (immediate local effect) plus the project's settings row through the existing `PATCH /api/projects/:id` endpoint (so the daemon honors it on the next spawn). Same-tab UI updates without a reload via the `open-design:critique-theater-toggle` CustomEvent; other tabs update via the platform `storage` event.
- A skill author who sets `od.critique.policy: opt-out` in `SKILL.md` frontmatter has critique force-disabled for every run using that skill, regardless of env var, project override, or rollout phase.
- A skill author who sets `od.critique.policy: required` has critique force-enabled the same way.
- During M0 (default) the feature stays disabled for everyone, even with `OD_CRITIQUE_ENABLED=1`, because no toggle has consented yet. M1 lets the toggle expose it. M2 enables it by default for skills tagged `opt-in`. M3 enables it globally.

The skill-level veto being unconditional is the user-visible safety property here: a user can never have critique forced on them for a generation flow the skill author explicitly marked as opt-out.

---

## 🎯 Surface area

- [x] Daemon (`apps/daemon/src/critique/rollout.ts`): new pure resolver + `parseRolloutPhase` + `parseEnvEnabled` helpers; 10 vitest cases covering every cell of the priority matrix.
- [x] Web runtime (`apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts`): new hook + `setCritiqueTheaterEnabled` setter; 11 vitest cases covering localStorage read, same-tab CustomEvent, cross-tab storage event, default disabled, setter round-trip, unmount cleanup.
- [x] Web barrel (`apps/web/src/components/Theater/index.ts`): exports the hook and setter so the Settings follow-up PR can consume them without reaching into the internal hooks path.
- [x] Contracts (`packages/contracts/src/critique.ts`): numeric-domain guards in `isPanelEvent` restored after the v0.7.0 merge wave dropped them (lefarcen + Siri-Ray P2 on this PR). Helpers `isPositiveInt`, `isNonNegativeInt`, `isNonNegativeFinite` reintroduced; `threshold <= scale` invariant enforced inside `run_started`; 43-case fixture matrix in `packages/contracts/tests/critique.test.ts` restored to pin `scale: 0`, fractional / negative rounds, negative scores, `threshold > scale`, and the other domain-violation rejections.
- [ ] E2E (`e2e`)
- [ ] Docs (`docs/`, `README*`, `AGENTS.md`), touched only via the v0.7.0 main-merge auto-merge; no new content.

---

## 🔐 Daemon: `isCritiqueEnabled` resolver

Resolution order, highest priority first:

```
   1.  skillPolicy === 'opt-out'        →  false        (skill veto)
   2.  skillPolicy === 'required'       →  true         (skill required)
   3.  projectOverride !== null         →  projectOverride
   4.  envOverride !== null             →  envOverride  (OD_CRITIQUE_ENABLED)
   5.  rollout phase default
           M0 dark-launch               →  false
           M1 settings toggle only      →  false
           M2 per-skill default-on      →  skillPolicy === 'opt-in'
           M3 global default-on         →  true
```

Skill-level vetoes win unconditionally so an adapter that explicitly opts out cannot have critique forced on it by an env var or a global rollout. A skill that opts in cannot be vetoed by anything below it either.

`parseRolloutPhase(raw)` defaults to M0 on unknown input so a fresh install never surprises a user with the feature on. `parseEnvEnabled` distinguishes truthy / falsy / missing so the resolver knows the difference between "user did not set the env var" and "user set it to 0".

---

## 🌐 Web: `useCritiqueTheaterEnabled` hook

```
   [ Settings panel saves in this tab ]
                  │
                  ▼
   setCritiqueTheaterEnabled(next)
                  │
                  ├── writes localStorage (preserves rest of config)
                  │
                  └── dispatches CustomEvent 'open-design:critique-theater-toggle'
                                  │
                                  ▼
                  every mounted useCritiqueTheaterEnabled reloads


   [ Other tab flips the toggle ]
                  │
                  ▼
   platform 'storage' event with key = 'open-design:config'
                  │
                  ▼
                  every mounted useCritiqueTheaterEnabled reloads
```

`setCritiqueTheaterEnabled(next)` round-trips through localStorage but preserves the rest of the config (mode, apiKey, anything else the user put there). The handshake does not stomp.

Failure modes that fall through to safe defaults:

| Scenario | Behavior |
|---|---|
| localStorage disabled / private mode | returns false, no throw |
| Stored blob is not valid JSON | returns false, no throw |
| Stored blob is JSON but not an object | returns false, no throw |
| CustomEvent constructor unavailable | localStorage still updates, single-mount remains correct |

---

## 🔄 How the two halves talk

```
   user flips Settings toggle
            │
            ▼
   web hook: localStorage.critiqueTheaterEnabled = true
            │
            ▼
   web app calls POST /api/projects/:id/settings
            │
            ▼
   daemon writes projectOverride = true
            │
            ▼
   next generation request hits the orchestrator
            │
            ▼
   isCritiqueEnabled({
     phase: parseRolloutPhase(env),
     skillPolicy: <from SKILL.md frontmatter>,
     projectOverride: <stored>,
     envOverride: parseEnvEnabled(env),
   })
            │
            └── true  ──► critique pipeline wired
                false ──► skipped, classic generation
```

No backend call learns the toggle state from the web layer directly. The web side writes the override into the project's settings row through the existing settings endpoint, and the resolver reads that row.

---

## 📸 Screenshots

Intentionally none. The Settings UI control that exposes the setter to end users lands in the focused Settings UI follow-up (#1484). This PR ships only the resolver and the hook plumbing, both of which are internal-only on the M0 rollout.

---

## ✅ Validation

Commands run on the current head (`195185a9`, after the contracts guard restore on top of the v0.7.0 main merge):

| Command | Result |
|---|---|
| `pnpm --filter @open-design/contracts test` | 5 files / 62 tests green, including the 43-case `critique.test.ts` matrix that pins `scale: 0`, fractional / negative rounds, negative scores, `threshold > scale`, and every other domain-violation rejection |
| `pnpm --filter @open-design/contracts build` | clean |
| `pnpm --filter @open-design/daemon test tests/critique-rollout.test.ts` | 10 / 10 (skill policy precedence, project override, env override, rollout phase fallthrough, boolean coercion edges, four-input combination matrix) |
| `pnpm --filter @open-design/web test tests/components/Theater/hooks/useCritiqueTheaterEnabled.test.tsx` | 11 / 11 (localStorage read, same-tab CustomEvent dispatch, cross-tab storage event, default disabled, setter round-trip, unmount cleanup) |

---

## ⛔ What this PR does not do

- **No orchestrator wiring.** The orchestrator does not yet call `isCritiqueEnabled` before routing. That edit is a separate, tiny PR so the resolver can land green and be reviewed on its own.
- **No Settings UI.** The setter is callable from the existing Settings handler in the focused Settings UI follow-up (#1484); this PR ships the underlying hook.
- **No `projectOverride` storage endpoint changes.** Phase 6 already provides the settings endpoint; the M1 toggle writes through it.
- **No Phase 12 instrumentation.** See the note below.

---

## 📅 Why Phase 12 is not in this stack

Phase 12 is observability: nine Prometheus metrics on the orchestrator, six structured log events under the `critique` namespace, and the Grafana dashboard JSON. Every counter Phase 12 registers fires from inside the orchestrator on the request path.

The orchestrator does not yet route through the critique pipeline. Until the wiring PR that follows this one lands, every Phase 12 counter would read 0 and every dashboard panel would be flat. Shipping Phase 12 ahead of the wiring would land green and feel productive, but it would also bake "this counter is permanently zero" into the system, which makes the eventual cutover noisier, not quieter.

So the order is:

```
   #1316 Phase 10   adapter conformance lab + degraded registry      ← merged
   #1317 Phase 11   Playwright stage suite
   #1318 Phase 13   reducer p99 bench + coverage walker
   #1319 Phase 14   user guide + AGENTS maps                         ← merged
   #1320 Phase 15   rollout resolver + Settings toggle hook          ◄── you are here
   #1338            orchestrator wiring (one-line request-path edit)
   #1484            Settings UI follow-up
   #1485 Phase 12   metrics + logs + dashboard                       ◄── lands once traffic flows
   #1499 Phase 16   M-phase rollout ratchet + /api/critique/conformance
```

This PR is the last preparatory piece. The next PR makes traffic flow; the one after that watches the traffic flow.

---

## 🧮 Diff anatomy

```
 apps/daemon/src/critique/rollout.ts                                 + 120
 apps/daemon/tests/critique-rollout.test.ts                          + 142
 apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts  + 102
 apps/web/tests/components/Theater/hooks/useCritiqueTheaterEnabled.test.tsx  + 110
 apps/web/src/components/Theater/index.ts                            +   4
 packages/contracts/src/critique.ts                                  +  60   (numeric-domain guards restored)
 packages/contracts/tests/critique.test.ts                           + 151   (fixture matrix restored)
```

Two new daemon files. Two new web files. One barrel export added. The contracts changes are restoring guards that landed earlier on PR #1314 and got dropped during the v0.7.0 merge wave. No edits to anything on the orchestrator request path.


apps/web/src/components/Theater/hooks/useCritiqueReplay.ts
```diff
@@ -0,0 +1,264 @@
...
+}
+
+function parseTranscript(raw: string): PanelEvent[] {
+  const out: PanelEvent[] = [];
+  for (const line of raw.split(/\r?\n/)) {
+    const trimmed = line.trim();
+    if (!trimmed) continue;
+    try {
+      const parsed = JSON.parse(trimmed);
+      if (isPanelEvent(parsed)) out.push(parsed);
```
lefarcen, line 243: **P2** — Replay only uses the shallow `isPanelEvent` guard, unlike the SSE path's variant-field validation. A corrupted/old transcript can pass with `{type:'panelist_dim', runId:'r'}` after `run_started`, store `undefined` scores, and later crash rendering at `d.score.toFixed(1)`. Reuse the variant validation or a shared schema here too.

apps/web/src/components/Theater/state/sse.ts
```diff
@@ -0,0 +1,201 @@
...
+ *  `{ type: 'ship', runId: 'r' }` would slip through to the reducer with every
+ *  other field undefined and crash downstream code that calls
+ *  `final.composite.toFixed(1)`. This second-pass filter enforces the shape
+ *  of each variant before the action is dispatched (lefarcen + Siri-Ray +
+ *  codex P2 on PR #1314). */
+function hasValidVariantShape(event: PanelEvent): boolean {
+  switch (event.type) {
+    case 'run_started':
+      return typeof event.protocolVersion === 'number'
+        && Array.isArray(event.cast) && event.cast.length > 0
```
lefarcen, line 65: **P2** — The variant guard still accepts any string for `cast`/`role`/`status`/`reason` fields. That lets malformed wire data through despite the API-boundary validation claim. Validate against the exported contract enums/lists so unknown roles or ship statuses are dropped before the reducer/UI.

apps/web/tests/components/Theater/hooks/useCritiqueTheaterEnabled.test.tsx
```diff
@@ -0,0 +1,109 @@
...
+    setCritiqueTheaterEnabled(true);
+    const stored = JSON.parse(window.localStorage.getItem('open-design:config') ?? '{}');
+    expect(stored.critiqueTheaterEnabled).toBe(true);
+    // Other fields stay intact: the toggle handshake does not stomp
+    // user config.
+    expect(stored.mode).toBe('daemon');
+    expect(stored.apiKey).toBe('sk-test');
+  });
+
+  it('tolerates corrupted JSON in the stored config (returns false, does not throw)', () => {
```
lefarcen, line 84: **P2** — The PR body claims safe behavior for disabled storage, JSON non-object values, and missing `CustomEvent`, but this suite only covers corrupt JSON plus happy-path storage events. Add tests for those failure modes, especially `localStorage.getItem/setItem` throwing, since the hook/setter intentionally swallow those errors.

docs/critique-theater.md
```diff
@@ -0,0 +1,204 @@
...
+is a turn in the same agent session, separated by `<PANELIST role="...">`
+tags in the protocol stream. The parser converts those into `panelist_*`
+events that the web reducer consumes via SSE. This keeps the operational
+contract identical to a normal generation: same auth, same env, same logs.
+
+## 3. Settings reference
+
+### Enable / disable
+
+The feature is gated behind the `OD_CRITIQUE_ENABLED` env var and, after
```
lefarcen, line 145: **P2** — This says Phase 15 lands the Settings toggle and persists it into daemon settings, but the PR explicitly defers Settings UI and projectOverride storage wiring. If these docs ship now, users and support will look for a control/state path that does not exist yet. Mark it as follow-up or align the implementation.

docs/critique-theater.md
```diff
@@ -0,0 +1,204 @@
...
+
+## 6. Troubleshooting
+
+### "Panel offline this run"
+
+The orchestrator emits a `degraded` event when one of these happens:
+
+| Reason | Cause | Remediation |
+|---|---|---|
+| `malformed_block` | The adapter emitted a `<CRITIQUE>` block the parser rejects. | Check the adapter's conformance status (`/api/metrics/critique`). |
```
lefarcen: **P2** — The troubleshooting path sends users to `/api/metrics/critique`, but Phase 12 metrics are explicitly deferred until after orchestrator wiring. The deferral rationale is sound, but the user doc should not reference a metrics endpoint/dashboard surface before it exists.

apps/web/src/components/Theater/AGENTS.md
```diff
@@ -0,0 +1,75 @@
...
+| `TheaterStage.tsx` | Live-debate container. Mounts five `PanelistLane`s + `ScoreTicker` + `RoundDivider` rail + `InterruptButton`. |
+| `PanelistLane.tsx` | One row per panelist. Keyed by `data-role` for CSS tinting; no hex literals in source. |
+| `ScoreTicker.tsx` | Composite trajectory with a threshold marker. |
+| `RoundDivider.tsx` | Round-by-round separator carrying composite vs threshold. |
+| `InterruptButton.tsx` | Click + Esc binding. Esc scopes itself to non-text-input focus so the prompt textarea is safe. |
+| `TheaterDegraded.tsx` | Single chip surface for runs the orchestrator could not score. `useId()` for the heading id so multiple chips on one page do not collide. |
+| `TheaterCollapsed.tsx` | Post-run summary: shipped, interrupted, failed. The interrupted branch uses `interruptedSummary`, not `shippedSummary`. |
+| `TheaterTranscript.tsx` | Read-only replay surface bound to `useCritiqueReplay`. |
+| `index.ts` | Public barrel. Only the mount + the four surface components + the two hooks are exported; everything else stays internal. |
```
lefarcen: **P2** — The module map is stale after Phase 15: `index.ts` now exports `useCritiqueTheaterEnabled` and `setCritiqueTheaterEnabled`, but this still says only two hooks are exported and the layout table does not mention the new toggle hook. Update the architecture map so future Theater changes see the actual public surface.

lefarcen (changes requested):
Hi @Nagendhra-web! Thanks for Phase 15 — the rollout resolver design is solid, with clear priority tiers and green tests covering the matrix.

Found API-boundary validation gaps in the stacked Theater client code plus doc/map misalignment with your explicit deferrals (Settings UI, Phase 12 metrics). The core rollout/toggle logic is sound; these findings are about the broader Theater surface integrity and docs accuracy.

See inline for specifics — most are straightforward aligns.

apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts
```diff
@@ -0,0 +1,101 @@
...
+export function useCritiqueTheaterEnabled(): boolean {
+  const [enabled, setEnabled] = useState<boolean>(() => readToggle());
+  useEffect(() => {
+    if (typeof window === 'undefined') return;
+    const reload = (): void => setEnabled(readToggle());
+    const onStorage = (evt: StorageEvent): void => {
+      if (evt.key !== null && evt.key !== STORAGE_KEY) return;
+      reload();
+    };
+    const onCustom = (): void => reload();
```
Siri-Ray: This same-tab handler ignores the `CustomEvent` payload and just calls `readToggle()` again. That works only when `setCritiqueTheaterEnabled()` successfully wrote `open-design:config`; in the quota/private-mode path added below, line 72 throws, line 79 still dispatches `{ detail: { enabled: next } }`, and this listener rereads the unchanged or unavailable localStorage value, so mounted hooks stay `false` instead of reflecting the user's toggle. The new test in `useCritiqueTheaterEnabled.test.tsx` expects the in-session event to flip `sink.enabled` after `localStorage.setItem` throws, but this implementation cannot satisfy that behavior. Handle the custom event as `CustomEvent<{ enabled: boolean }>` and set state from `detail.enabled` when present, falling back to `readToggle()` only for storage events or malformed custom events; alternatively keep an in-memory fallback that `readToggle()` can consult after write failures.

Siri-Ray (commented):
I found one blocking issue in the new Settings-toggle hook failure path. The local-storage failure mode currently does not update mounted hook state, and the new regression test asserts the opposite.

apps/web/src/components/Theater/hooks/useCritiqueReplay.ts
lefarcen, line 243: **P2 (carryover)** — Replay still uses shallow `isPanelEvent` guard, unlike SSE's variant validation. Corrupted/old transcript can pass malformed events that crash rendering downstream. Reuse the variant validation or shared schema.

apps/web/src/components/Theater/state/sse.ts
lefarcen, line 65: **P2 (carryover)** — Variant guard accepts any string for `cast`/`role`/`status`/`reason` fields, letting malformed wire data through. Validate against exported contract enums so unknown roles/statuses are dropped before reducer/UI.

docs/critique-theater.md
lefarcen, line 145: **P2 (carryover)** — Doc says Phase 15 lands Settings toggle + persists into daemon, but PR defers Settings UI and projectOverride wiring. Mark as follow-up or align the implementation before shipping this doc.

docs/critique-theater.md
lefarcen: **P2 (carryover)** — Troubleshooting sends users to `/api/metrics/critique`, but Phase 12 metrics are deferred until after orchestrator wiring. Don't reference endpoints/surfaces before they exist.

apps/web/src/components/Theater/AGENTS.md
lefarcen: **P2 (carryover)** — Module map stale: `index.ts` now exports `useCritiqueTheaterEnabled` and `setCritiqueTheaterEnabled`, but this still says only two hooks are exported. Update the architecture map.

apps/web/tests/components/Theater/hooks/useCritiqueTheaterEnabled.test.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+      const sink: { enabled?: boolean } = {};
+      render(<Probe sink={sink} />);
+      expect(sink.enabled).toBe(false);
+      act(() => {
+        setCritiqueTheaterEnabled(true);
+      });
+      // The dispatch path is exercised even though localStorage rejected
+      // the write: every mounted hook updates from the in-session event.
+      expect(sink.enabled).toBe(true);
```
lefarcen, line 240: **P2 (new)** — This assertion expects the hook to flip true after `localStorage.setItem` throws, but the CustomEvent listener ignores `event.detail` and only calls `readToggle()`, which rereads unchanged localStorage. When the setItem failure is actually simulated, this should remain false. The test encodes behavior the hook doesn't implement.

lefarcen (changes requested):
Re-reviewed at affcdd27. One of the six prior findings is addressed (failure-mode test coverage), but five remain. The new push also introduces a test assertion that encodes behavior the hook doesn't implement.

See inline for the carryover + new item.

apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts
Nagendhra-web: Fixed in f3e486d. The CustomEvent listener now reads `event.detail.enabled` directly when it is a boolean and only falls back to `readToggle()` for malformed events or for cross-tab `storage` events (which carry no typed payload). Under a throwing `setItem` the localStorage value is stale, but the event's `detail` carries the user's actual intent, so every mounted hook reflects the toggle correctly even when persistence is unavailable. The 'setItem throws' test now asserts the right behavior for the right reason; a new `falls back to readToggle when the CustomEvent carries no usable detail` test pins the degraded path so a future emitter that forgets the detail does not silently break the session. 11 / 11 vitest cases green.

apps/web/src/components/Theater/state/sse.ts
Nagendhra-web: Fixed upstream on PR #1316 commit 9b72b54. `hasValidVariantShape` now validates `role` against `PANELIST_ROLES` (imported from contracts), `status` against `ShipStatus`, `reason` against `DegradedReason`, `cause` against `FailedCause`, `kind` against `ParserWarningKind`, and `decision` against `RoundDecision`. A frame with `role:'__proto__'` or `status:'wat'` is dropped before reaching the reducer. Surfaces in #1320 once #1316 merges and this branch rebases.

apps/web/src/components/Theater/AGENTS.md
Nagendhra-web: Fixed on PR #1319 commit 0aece0a. The web Theater AGENTS.md module map now lists three hooks (useCritiqueStream, useCritiqueReplay, useCritiqueTheaterEnabled) plus the setCritiqueTheaterEnabled imperative setter on the index.ts row, and adds a new row for hooks/useCritiqueTheaterEnabled.ts in the file table. A web contributor scanning the table now sees the Phase-15 hook without inferring it from the barrel blurb. Surfaces in #1320 once #1319 merges and this branch rebases.

e2e/ui/critique-theater.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+test.describe('Critique Theater e2e (Phase 11)', () => {
+  test.beforeEach(async ({ page }) => {
+    await bootAppWithCritiqueEnabled(page);
+  });
+
+  // Task 11.1: happy path
+  test('renders the live stage, all five lanes, and the shipped badge', async ({ page }) => {
+    await stubProjectEvents(page, TRANSCRIPT);
+    await page.goto('/');
```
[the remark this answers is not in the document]
Nagendhra-web: Same mount-not-wired observation as the Phase 11 / 13 / 14 reviews. Fixed upstream on PR #1317 commit 3053c1b: the whole describe is parked behind `test.describe.fixme` until the wireup PR mounts `<CritiqueTheaterMount>` into App / ProjectView. Playwright still lists the 6 tests via `--list` so the spec stays visible. Surfaces in #1320 once #1317 merges and this branch rebases.

docs/critique-theater.md
Nagendhra-web: Fixed on PR #1319 commit 89ddd10. The Settings reference section now leads with 'Phase 15 ships the client-side toggle plumbing (the useCritiqueTheaterEnabled hook + setCritiqueTheaterEnabled setter)... the Settings → Critique Theater (beta) control that calls those setters from the standard Settings panel is a deliberate follow-up that lands together with the projectOverride storage endpoint on the daemon.' The doc no longer promises a control or daemon-persistence path that does not exist yet. Surfaces in #1320 once #1319 merges and this branch rebases.

apps/web/src/components/Theater/hooks/useCritiqueReplay.ts
lefarcen, line 243: **P2 (carryover)** — Replay still uses shallow `isPanelEvent` guard. A corrupt transcript line with only `{type:'ship', runId:'r'}` passes this check but crashes downstream at `final.composite.toFixed(1)`. The fix is on PR #1316 commit 9b72b54 (`parseTranscript` layers `hasValidVariantShape` on top of `isPanelEvent`). Will surface after rebase.

apps/web/src/components/Theater/state/sse.ts
lefarcen, line 65: **P2 (carryover)** — Variant guard accepts any string for `role`/`status`/`reason`/`cause`/`kind`/`decision`. A malformed frame with `role:'__proto__'` or `status:'wat'` slips through. The fix is on PR #1316 commit 9b72b54 (`hasValidVariantShape` validates against contract enums). Will surface after rebase.

docs/critique-theater.md
lefarcen, line 145: **P2 (carryover)** — Doc says Phase 15 lands the Settings toggle with daemon persistence, but this PR explicitly defers the Settings UI and projectOverride storage endpoint. The fix is on PR #1319 commit 89ddd10 (doc rewritten to say "Phase 15 ships the client-side toggle plumbing; the Settings UI control is a deliberate follow-up"). Will surface after rebase.

docs/critique-theater.md
lefarcen: **P2 (carryover)** — Troubleshooting references `/api/metrics/critique`, but Phase 12 metrics are deferred. The fix is on PR #1319 commit 0aece0a (replaced with local conformance-harness command that ships today). Will surface after rebase.

apps/web/src/components/Theater/AGENTS.md
lefarcen: **P2 (carryover)** — Module map is stale: `index.ts` now exports three hooks (`useCritiqueStream`, `useCritiqueReplay`, `useCritiqueTheaterEnabled`) plus `setCritiqueTheaterEnabled`, but this doc says only two hooks. The fix is on PR #1319 commit 0aece0a (updated row in file table). Will surface after rebase.

lefarcen (changes requested):
Re-reviewed at f3e486d4. The test assertion fix (item 6) is confirmed fixed in this branch. Five prior items remain, as the upstream fixes on #1316/#1317/#1319 haven't rebased here yet.

See inline for the 5 carryover items. Once you rebase against main (bringing in the upstream fixes), these should all resolve.

lefarcen (changes requested):
Re-review on current head 2d629e6c: 5 items still present pending upstream rebase; 1 item fixed.

**Fixed in this PR (item 6):**
- Test assertion at `apps/web/tests/components/Theater/hooks/useCritiqueTheaterEnabled.test.tsx:198` now correctly verifies the listener honors `event.detail.enabled` directly when `localStorage.setItem` throws. The hook implementation at `apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts:52-59` consumes the typed payload as expected. Well done.

**Remaining items (awaiting upstream rebase per author comments):**
1. `apps/web/src/components/Theater/hooks/useCritiqueReplay.ts:243` — `parseTranscript` still uses shallow `isPanelEvent` guard without layering `hasValidVariantShape`. Author claims fix on PR #1316 commit 9b72b54.
2. `apps/web/src/components/Theater/state/sse.ts:65` — `hasValidVariantShape` validates `typeof event.role === 'string'` but does not check against `PANELIST_ROLES` enum. Same for `status`, `decision`, `reason`, `cause`, `kind`. Author claims fix on PR #1316 commit 9b72b54.
3. `docs/critique-theater.md:88` — Doc says "Phase 15 lands, the Settings → Critique Theater (beta) toggle" but PR body states the Settings UI control is a follow-up. Author claims fix on PR #1319 commit 89ddd10.
4. `docs/critique-theater.md:162` — Troubleshooting table references `/api/metrics/critique` endpoint which does not ship in this PR. Author claims fix on PR #1319 commit 0aece0a.
5. `apps/web/src/components/Theater/AGENTS.md:26` — `index.ts` row says "two hooks" but PR ships three hooks (`useCritiqueTheaterEnabled` added) plus `setCritiqueTheaterEnabled` setter. Author claims fix on PR #1319 commit 0aece0a.

Once the upstream PRs (#1316, #1317, #1319) merge and this branch rebases, these 5 items should surface as resolved. Item 6 is clean on the current head.

lefarcen (changes requested):
Re-review at head cbff22ae. Items 1, 2, 4, 5, 6 from the prior review are now fixed — the contract guard is strict, the doc references are accurate, the module map is current, and the test correctly pins the CustomEvent fallback.

Two items remain:

**P2 - Rollout resolver not wired (new)**
The `isCritiqueEnabled` resolver exists (`apps/daemon/src/critique/rollout.ts:62`) with full test coverage, but the actual run gate in `apps/daemon/src/server.ts:3029` still reads `critiqueCfg.enabled` directly — the resolver is only called in tests. Project overrides, rollout phases, and skill policies therefore cannot affect live runs. Either wire `isCritiqueEnabled` into the run gate or clarify in the PR body that wiring is deferred to a follow-up PR (same as the orchestrator note on line 48).

**P2 - Settings toggle doc claim (partial fix)**
`docs/critique-theater.md:145-148` is improved by phrasing the Settings UI as future, but it still describes a round-trip through the daemon settings store that does not yet exist: no `/api/settings/critique` endpoint implementation is present, and no production caller of `setCritiqueTheaterEnabled` writes through to the daemon. Reword to match the actual scope (localStorage-only toggle for this phase, with daemon persistence deferred).

Once these two are addressed or clarified, this PR ships the resolver + hook infrastructure cleanly.

docs/critique-theater.md
```diff
@@ -0,0 +1,261 @@
...
+tags in the protocol stream. The parser converts those into `panelist_*`
+events that the web reducer consumes via SSE. This keeps the operational
+contract identical to a normal generation: same auth, same env, same logs.
+
+## 3. Settings reference
+
+### Enable / disable
+
+The feature is gated behind the `OD_CRITIQUE_ENABLED` env var and, after
+Phase 15 lands, the **Settings → Critique Theater (beta)** toggle. The web
```
Siri-Ray, line 146: This Settings section still describes the Phase 15 toggle as persisting into the daemon's settings store and flipping the same backend flag, but the current implementation only adds the web-side localStorage hook/setter (`useCritiqueTheaterEnabled` / `setCritiqueTheaterEnabled`) and the PR body explicitly says the Settings UI plus projectOverride storage endpoint are follow-ups. That mismatch matters because users/operators reading this doc will expect backend persistence and per-project daemon gating to work today, then debug a missing endpoint/caller that intentionally does not ship in this phase. Please reword this paragraph to match the current scope, for example: `Phase 15 ships the client-side toggle plumbing that reads/writes open-design:config; the Settings UI control and daemon projectOverride persistence land in a follow-up, while backend routing remains controlled by the resolver inputs once wired.`

Siri-Ray (commented):
@Nagendhra-web thanks for continuing to tighten this stack. I reviewed the current head and the rollout resolver / hook implementation looks coherent; I found one merge-safe docs alignment issue where the user guide still describes daemon persistence that this phase intentionally defers.



apps/daemon/src/critique/rollout.ts
```diff
@@ -0,0 +1,119 @@
+/**
+ * Critique Theater rollout-flag plumbing (Phase 15).
+ *
+ * The plan's rollout track is M0 dark-launch -> M1 settings toggle ->
+ * M2 default-on per skill -> M3 global default. This module is the
+ * single decision point every caller consults to know "should the
+ * orchestrator wire the critique pipeline for this run?" The same
+ * helper is consumed by:
+ *
+ *   - The orchestrator entry, before it spawns the critique CLI
```
lefarcen: P2 This module comment now says the orchestrator entry consumes `isCritiqueEnabled`, but the current run gate in `apps/daemon/src/server.ts` still derives `critiqueShouldRun` from `critiqueCfg.enabled` directly. Because this PR intentionally defers orchestrator wiring, please reword this list to describe the resolver as the planned/shared decision point (or wire it here) so future readers do not assume rollout phases/project overrides are live in generation routing yet.

docs/critique-theater.md
```diff
@@ -0,0 +1,261 @@
...
+is a turn in the same agent session, separated by `<PANELIST role="...">`
+tags in the protocol stream. The parser converts those into `panelist_*`
+events that the web reducer consumes via SSE. This keeps the operational
+contract identical to a normal generation: same auth, same env, same logs.
+
+## 3. Settings reference
+
+### Enable / disable
+
+The feature is gated behind the `OD_CRITIQUE_ENABLED` env var and, after
```
lefarcen, line 145: P2 This section still promises that after Phase 15 the Settings toggle exists and persists into the daemon settings store, but this PR ships only the localStorage hook/setter and no production Settings UI caller or daemon persistence endpoint. Please reword this to the actual scope: env gating plus client-side toggle plumbing in this phase, with the Settings control/project override persistence deferred to the follow-up.

lefarcen (changes requested):
Hey @Nagendhra-web 👋

The branch picked up the previous contract/test fixes cleanly, and the CustomEvent fallback path now looks right. I still see two scope-accuracy issues where the current docs/comments describe wiring that does not ship in this PR — see inline.

Once those are aligned with the actual Phase 15 surface, this should be ready for the next pass. ❤️

apps/daemon/src/critique/rollout.ts
lefarcen: P2 This module-level list still says the orchestrator entry and a `GET /api/settings/critique` endpoint consume `isCritiqueEnabled`, but this PR intentionally leaves orchestrator wiring and daemon Settings persistence for follow-ups. Please reword these bullets to describe planned consumers/current scope (or remove the non-shipping endpoint reference) so the source docs do not imply rollout phases/project overrides are active in live generation routing today.

lefarcen (changes requested):
Hey @Nagendhra-web 👋

Thanks for tightening the Settings guide — the user-facing docs now clearly separate Phase 15 from the follow-up wiring. One remaining source comment still describes deferred integrations as if they already consume the resolver; see inline.

Once that comment is aligned, the previous blockers are resolved from my side. ❤️

Siri-Ray (commented):
@Nagendhra-web thanks for continuing to tighten this stack. The user-facing Settings section now clearly documents the Phase 15 boundaries, and the resolver/hook implementation looks coherent on this pass. One merge-safe scope comment remains in `apps/daemon/src/critique/rollout.ts`: the top-of-file consumer list still says the orchestrator entry and `GET /api/settings/critique` endpoint consume `isCritiqueEnabled`, but this PR intentionally defers both integrations and the docs now say the spawn-time gate still reads `critiqueCfg.enabled` directly. That matters because future contributors can read this source comment and assume rollout phases/project overrides are live in generation routing when they are not yet wired. Please reword that list to describe these as planned follow-up consumers, or trim it to only the behavior that ships in this PR. I’m leaving this as non-blocking because it is documentation/source-comment drift rather than a runtime bug.



lefarcen (dismissed):
Hey @Nagendhra-web, this last docblock update lands the scope boundary cleanly. The resolver comment now makes the current Phase 15 surface explicit, calls out the planned consumers separately, and matches the Settings guide/operator guidance.

That resolves my remaining blocking review from the current stack. Nice follow-through on keeping the source docs honest while the runtime wiring stays deferred. ❤️

Siri-Ray (dismissed):
@Nagendhra-web I reviewed the current head across the rollout resolver, degraded/conformance helpers, strict contract validation, Theater stream/replay paths, Settings toggle hook, docs, and the prior open review items. The Phase 15 scope is now clear, the previously flagged source/doc drift is resolved on this head, and I did not find any actionable changed-range issues. I attempted the focused daemon/contracts/web Theater test commands as part of verification, but this worktree does not have node_modules installed, so vitest could not be resolved locally; the code review pass itself is clean. Thanks for the careful follow-through on tightening the rollout boundary and keeping the deferred wiring explicit.



apps/web/src/components/Theater/index.ts
```diff
@@ -8,6 +8,7 @@
  * the canonical mount surface narrow so adding/removing internals
  * does not churn callers.
  */
+export { CritiqueTheaterMount } from './CritiqueTheaterMount';
```
Siri-Ray, line 11: The new Settings toggle hook and its imperative setter are not exported from the Theater public barrel, even though this PR adds them as the Phase 15 web-side surface and `apps/web/src/components/Theater/AGENTS.md` now describes `index.ts` as exporting `useCritiqueTheaterEnabled` plus `setCritiqueTheaterEnabled`. That mismatch matters because the follow-up Settings handler either has to import from the private hook path or fails to use the intended public package surface; it also leaves the changed module map inaccurate. Please add the missing barrel export, for example `export { useCritiqueTheaterEnabled, setCritiqueTheaterEnabled } from './hooks/useCritiqueTheaterEnabled';`, next to the other hook exports.

Siri-Ray (commented):
@Nagendhra-web thanks for continuing to tighten this stack. I reviewed the current head across the rollout resolver, web toggle hook, Theater public surface, contracts validation, docs, and the prior review context. I found one merge-safe public-surface mismatch in the barrel export that is worth fixing before the Settings follow-up consumes this hook.



docs/critique-theater.md
Siri-Ray: Looper checked this thread against head `bd73cc5c97df52a2a3cb9b300afa9379764d7ec2`. The requested change appears objectively addressed by docs/critique-theater.md now states Phase 15 ships only the client-side toggle hook/setter writing open-design:config localStorage, says the Settings UI is a follow-up, and explicitly notes the localStorage flag is not yet round-tripped into the daemon project settings store or spawn gate., so I’m resolving this thread. Reopen if this still needs discussion.



Siri-Ray (dismissed):
@Nagendhra-web thanks for the continued follow-through on this stack. I reviewed the current head across the rollout resolver/parsers, degraded/conformance harness, strict contract validation, Theater stream/replay/mount/toggle surfaces, Tailwind style-policy guard changes, docs, e2e fixture, and prior review context. I did not find any actionable changed-range issues; the prior barrel export and docs/source-scope concerns are addressed on this head. I attempted the focused guard/contracts/daemon/web Theater validation locally, but this worktree has no node_modules installed (`tsx`/`vitest` unavailable), so the publish decision is based on the code review pass and existing PR/author verification notes. Thanks for tightening the rollout boundary and keeping the deferred wiring explicit.



packages/contracts/src/critique.ts
```diff
@@ -195,46 +168,39 @@ export function isPanelEvent(value: unknown): value is PanelEvent {
...
-        && isPositiveInt(scale)
-        && isNonNegativeFinite(threshold)
-        // threshold must fit within the configured scale; downstream
-        // panel decisions compare composite against threshold and break
-        // silently if threshold > scale.
-        && (threshold as number) <= (scale as number);
-    }
+        && isFiniteNumber(o['maxRounds']) && o['maxRounds'] > 0
+        && isFiniteNumber(o['threshold'])
+        && isFiniteNumber(o['scale']);
```
Siri-Ray: This branch now accepts any finite `scale`, `threshold`, `round`, `score`, `mustFix`, `bestRound`, and `position` value, which regresses the stricter numeric-domain guard that exists on `main`. The changed lines only check `isFiniteNumber`, so frames like `run_started` with `scale: 0`, fractional/negative rounds, negative scores, or `threshold > scale` pass `isPanelEvent`; the same PR diff also deletes the contract tests that pinned those cases in `packages/contracts/tests/critique.test.ts`. That matters because this guard is the shared wire validator for both `sseToPanelEvent` and transcript replay before Theater renders values such as `ScoreTicker` percentages and round buckets. Please restore the positive-integer/non-negative helpers and the deleted fixture-matrix tests, including the `scale: 0` and `threshold > scale` cases, so malformed SSE/replay frames are rejected at the contract boundary again.

Siri-Ray (commented):
@Nagendhra-web thanks for the careful rebase and for keeping the Phase 15 rollout/hook scope explicit. I found one merge-safe regression in the shared contracts guard after the main merge: the numeric-domain checks from `main` were dropped, while the web SSE/replay paths still rely on that guard before reducer/rendering.



packages/contracts/src/critique.ts
```diff
@@ -195,46 +168,39 @@ export function isPanelEvent(value: unknown): value is PanelEvent {
   const runId = o['runId'];
   if (typeof runId !== 'string' || runId.length === 0) return false;
   switch (t as PanelEvent['type']) {
-    case 'run_started': {
-      const threshold = o['threshold'];
-      const scale = o['scale'];
-      return isPositiveInt(o['protocolVersion'])
+    case 'run_started':
+      return isFiniteNumber(o['protocolVersion'])
```
lefarcen: P2 The rebase drops the numeric-domain guard that `main` added here: `scale`, `protocolVersion`, `maxRounds`, and `round` are now only finite numbers again, while `threshold <= scale`, non-negative scores/composites, integer rounds, and non-negative `mustFix`/`position` are no longer enforced. The web SSE and replay paths trust `isPanelEvent` before dispatching to the reducer/renderers, so malformed frames like `scale: 0`, fractional rounds, or negative composite values can still reach UI code and produce broken styles or impossible run state. Please restore the domain helpers/tests from `main` (positive ints for protocol/scale/rounds, non-negative finite scores, non-negative integer counts/positions, and threshold within scale) before layering this PR’s additions on top.

lefarcen (changes requested):
Hey @Nagendhra-web 👋

Thanks for the careful rebase notes — the Phase 15 rollout/hook pieces still look scoped the same, and the barrel export fix is in place. I found one correctness regression from the main merge that needs to be restored before this lands: the shared contracts guard lost the numeric-domain checks that protect the Theater reducer/render path. See inline.

One process note too: after the new PR template landed, this PR body is now missing a few required sections (**Why**, **What users will see**, **Surface area**, and **Validation**). No rush, but please update those so the next review / release verification pass can scope the user-visible surface quickly. If helpful, you can paste this into your assistant:

> Read the current PR body for nexu-io/open-design#1320 and the template at
> Then update via `gh pr edit 1320 --repo nexu-io/open-design --body-file -`.

Once the guard is restored and the body is aligned, this should be back to a clean review path. ❤️

packages/contracts/src/critique.ts
lefarcen: P2 This still needs the numeric-domain checks restored from `main`. `scale`, `protocolVersion`, `maxRounds`, `round`, `mustFix`, `position`, scores, and composites are still accepted as any finite number, so `scale: 0`, fractional/negative rounds, negative scores, and `threshold > scale` can pass the shared `isPanelEvent` guard into SSE/replay consumers. Please bring back the positive-integer / non-negative helper predicates plus the deleted contract test matrix before this lands.

lefarcen (changes requested):
Hey @Nagendhra-web 👋

Thanks for the quick follow-up. I checked the new head and the contracts numeric-domain regression is still present: the guard is still only checking finite numbers, and the deleted matrix tests are still absent. The PR body also still has the template gaps from the previous review (**Why**, **What users will see**, **Surface area**, and **Validation**).

Leaving the remaining blocker inline so GitHub keeps this review attached to the current head. Once the domain checks/tests and PR body are restored, I can clear my side. ❤️

Siri-Ray (commented):
@Nagendhra-web thanks for keeping this Phase 15 rollout pass moving. I reviewed the current head across the rollout resolver, web toggle hook, Theater barrel export, FileViewer test adjustment, contracts guard, and the prior review context. I found one merge-safe contracts validation regression that is still present on this head; details are inline.



packages/contracts/src/critique.ts
Siri-Ray: **NON_BLOCKING** Restore the numeric-domain checks in `isPanelEvent` before relying on this guard as the shared wire boundary. The changed guard now rejects only non-finite values here and in the other numeric branches, so frames such as `run_started` with `scale: 0`, `threshold: -1`, or `threshold > scale`, plus `round: 0` / fractional rounds and negative `mustFix` or scores, still pass. That matters because the Theater render path trusts this contract guard before reducer/UI code: `ScoreTicker` divides by `scale` for inline heights and the threshold marker, while the reducer stores round and score fields without re-validating their domain. Please bring back helper predicates for positive integers and non-negative finite/integer values, enforce `threshold <= scale` on `run_started`, and restore a fixture matrix in `packages/contracts/tests/critique.test.ts` covering zero/negative/fractional scale, rounds, counts, scores, and parser positions.

Siri-Ray (approved):
@Nagendhra-web I reviewed the current head across the Phase 15 daemon rollout resolver, parser helpers, web toggle hook, Theater barrel export, and the small FileViewer test adjustment. The changed-range implementation matches the stated internal-only scope: the resolver priority is pinned by unit coverage, the hook preserves existing config while syncing same-tab and cross-tab changes, and the prior contracts guard concern is no longer in the live diff. I could not run the focused Vitest commands locally because this worktree has no node_modules installed, but the live GitHub checks are passing and the code review pass did not turn up actionable issues. Thanks for the careful follow-through on the rebase and the review-thread fixes. ❤️



lefarcen (approved):
Hey @Nagendhra-web, thanks for the thorough fix here. I re-checked the contracts guard and the deleted numeric-domain fixture matrix is back: `scale`, `round`, count, score, and threshold domains are enforced again, including the `threshold <= scale` invariant, so the Theater SSE/replay boundary is protected before reducer/rendering code sees those frames.

I also checked the refreshed PR body against the new template shape — the Why, user-visible behavior, surface area, screenshots rationale, and validation sections are now present. That clears my stale CHANGES_REQUESTED on this head. Product approval is still tracked separately, but code-review-wise this is clean from my side. ❤️

PR #1321, opened 2026-05-11
fix: rename Pets settings entry to 'Manage Pets'

xxiaoxiong:
## Summary

This PR fixes issue #1250 by renaming the Pets settings menu entry from "Customize in Settings" to "Manage Pets" for better clarity and discoverability.

## Problem

The current menu entry label is misleading:
- **Current label**: "Customize in Settings" (English) / "在设置中自定义" (Chinese)
- **Problem**: Suggests a broad customization action
- **Reality**: Opens Settings directly on the Pets tab
- **Impact**: Users may not understand what this entry does

## Solution

Rename the menu entry to clearly indicate its purpose:
- **New label**: "Manage Pets" (English) / "管理宠物" (Chinese)
- **Benefit**: Matches the actual destination and intent
- **Result**: Improved discoverability and reduced confusion

## Changes

- ✅ Update `en.ts`: "Customize in Settings" → "Manage Pets"
- ✅ Update `zh-CN.ts`: "在设置中自定义" → "管理宠物"

## Technical Details

The translation key `pet.composerOpenSettings` is used in:
- `ChatComposer.tsx` (lines 1001 and 1289)
- Clicking triggers `onOpenPetSettings` function
- `openPetSettings` function (App.tsx:754-758) already works correctly:
  - Sets `setSettingsInitialSection("pet")` - navigates to Pets tab
  - Sets `setSettingsOpen(true)` - opens Settings dialog

This PR only updates the label text. The underlying navigation functionality is already correct.

## User Experience

**Before:**
- Menu shows: "Customize in Settings" ❌
- User thinks: "What am I customizing?"
- Unclear destination

**After:**
- Menu shows: "Manage Pets" ✅
- User thinks: "This manages my pets"
- Clear destination

## Files Modified

- `apps/web/src/i18n/locales/en.ts` (1 line)
- `apps/web/src/i18n/locales/zh-CN.ts` (1 line)

## References

- Closes #1250
- Priority: P2 (UX/information architecture)

---

现在 Pets 设置入口的标签更清晰了！✨

### DIFF
--- a/apps/web/src/i18n/locales/en.ts
+++ b/apps/web/src/i18n/locales/en.ts
@@ -1135,7 +1135,7 @@ export const en: Dict = {
   'pet.composerTitle': 'Pets — wake, tuck, or pick one',
   'pet.composerMenuTitle': 'Pets',
   'pet.composerMenuHint': 'tip: type /pet to toggle',
-  'pet.composerOpenSettings': 'Customize in Settings',
+  'pet.composerOpenSettings': 'Manage Pets',
   'pet.welcomeTeaserTitle': 'Adopt a pet',
   'pet.welcomeTeaserBody': 'A tiny floating companion that hangs out with you.',
   'pet.welcomeTeaserCta': 'Pick one',
--- a/apps/web/src/i18n/locales/zh-CN.ts
+++ b/apps/web/src/i18n/locales/zh-CN.ts
@@ -1105,7 +1105,7 @@ export const zhCN: Dict = {
   'pet.composerTitle': '宠物 — 唤醒、收起或选一只',
   'pet.composerMenuTitle': '宠物',
   'pet.composerMenuHint': '小贴士：输入 /pet 即可切换',
-  'pet.composerOpenSettings': '在设置中自定义',
+  'pet.composerOpenSettings': '管理宠物',
   'pet.welcomeTeaserTitle': '领养一只宠物',
   'pet.welcomeTeaserBody': '一只小小的浮窗伙伴，会陪着你工作。',
   'pet.welcomeTeaserCta': '挑一只',


apps/web/src/i18n/locales/de.ts
```diff
@@ -921,7 +921,7 @@ export const de: Dict = {
   'pet.composerTitle': 'Haustiere — wecken, verstecken oder auswählen',
   'pet.composerMenuTitle': 'Haustiere',
   'pet.composerMenuHint': 'Tipp: tippe /pet zum Umschalten',
-  'pet.composerOpenSettings': 'In den Einstellungen anpassen',
+    'pet.composerOpenSettings': 'Haustiere verwalten',
```
lefarcen: P3 — Indentation: this line uses 4 spaces, but surrounding lines use 2 spaces. Same in 15 other locales (ar, es-ES, fa, fr, hu, id, ja, ko, pl, pt-BR, ru, th, tr, uk, zh-TW). Would be cleaner to normalize to 2 spaces.

lefarcen (commented):
Thanks @xxiaoxiong for tackling this UX clarity issue! The translations look good semantically — spot-checked German, Spanish, and Simplified Chinese and they all correctly mean "Manage Pets."

Found two minor points worth addressing:

**P2 — Shared key creates new ambiguity**: The `pet.composerOpenSettings` key is used in two places in `ChatComposer.tsx`: (1) the pet panel action (line 1289) which correctly calls `onOpenPetSettings`, and (2) the generic footer settings button (line 990-1002) which calls `onOpenSettings` (defaults to execution/settings section, not Pets). After this rename, that generic footer button could say "Manage Pets" while opening non-Pets settings — recreating the "misleading label" problem in MCP/import contexts. Suggest splitting the strings so the generic footer keeps a generic settings label, and use a pet-specific key only for the pet action. Alternatively, render this label only when `toolsTab === 'pet'` and call `onOpenPetSettings`.

**P3 — Indentation inconsistency**: 16 of 18 locale files use 4 spaces for the changed line instead of the surrounding 2-space indentation (exceptions: `en.ts` and `zh-CN.ts` are already correct). Non-functional but creates inconsistent formatting. Would be cleaner to normalize to 2 spaces across all touched locales.

Both are minor — the core translation work is solid.

Siri-Ray (commented):
I found one user-facing correctness issue in the current head: the renamed string is shared with a generic settings action, so the new label can point users to the wrong destination.


apps/web/src/i18n/locales/en.ts
```diff
@@ -1135,7 +1135,7 @@ export const en: Dict = {
   'pet.composerTitle': 'Pets — wake, tuck, or pick one',
   'pet.composerMenuTitle': 'Pets',
   'pet.composerMenuHint': 'tip: type /pet to toggle',
-  'pet.composerOpenSettings': 'Customize in Settings',
+  'pet.composerOpenSettings': 'Manage Pets',
```
Siri-Ray, line 1138: This string is still used by two different settings actions, so renaming the shared key to `Manage Pets` makes one of them incorrect. The Pets panel action in `ChatComposer.tsx` calls `onOpenPetSettings`, but the generic tools-menu footer also renders `t('pet.composerOpenSettings')` and calls `onOpenSettings?.()`; in `App.tsx`, `openSettings()` defaults to the `execution` section. That means users on the MCP/import tools tabs can see `Manage Pets` and land in non-Pets settings, which recreates the misleading-destination bug this PR is meant to fix. Please split the labels, for example keep a generic settings label for the footer and use a pet-specific key only for `ToolsPetPanel`, or make the footer pet-specific by calling `onOpenPetSettings` only when the active tab is `pet`.

lefarcen, line 1138: P2 — Shared-key ambiguity: this key is rendered by both the pet-panel action (correct destination: Pets settings) and the generic footer settings button (destination: general Settings). Renaming to Manage Pets makes the generic footer misleading. Split the keys or keep the wording generic.

lefarcen (changes requested):
Thanks for narrowing to 2 locales, @xxiaoxiong — the indentation is now consistent. However, the shared-key ambiguity (P2 from the previous review) is still present:

Both the pet-panel action (onOpenPetSettings) and the generic footer settings button (onOpenSettings) render pet.composerOpenSettings. After this rename, the generic footer will say Manage Pets while opening non-Pets settings, which misleads users.

Fix options:
1. Split the labels: introduce a new key for the generic footer (e.g. composer.openSettings) and keep pet.composerOpenSettings pet-specific.
2. Keep the shared wording generic (e.g. Open Settings) if both actions genuinely should have the same label.

Once that is addressed, this will be ready.

lefarcen:
Hey @xxiaoxiong, quick coordination note: PR #1939 is now open and overlaps this PR on the same `pet.composerOpenSettings` English label change (`Customize in Settings` → `Manage Pets`). It also adds broader ChatPane/title-edit UI changes, so the scopes are no longer identical.

Could you consolidate the Pets label fix into one PR so review doesn’t split across #1321 and #1939? Since this PR already has the shared-key feedback thread, either updating this one or closing it in favor of #1939 is fine — just leave a note with which direction you prefer.

xxiaoxiong:
Closing in favor of #1939 which addresses the same issue with complete i18n coverage.

PR #1322, opened 2026-05-11
fix: restore input focus after creating project

### LINKED ISSUE #1129: Input box bug

xxiaoxiong:
Fixes #1129

## Summary

After creating a project and then attempting to create another one, the name input field would not receive focus, making it appear unresponsive. This was especially noticeable after deleting a project and trying to create a new one.

## Changes

- Add `nameInputRef` to track the project name input element
- Reset `name` state to empty string after project creation
- Automatically refocus the input field after creation
- Use `setTimeout` to ensure focus happens after state updates

## Impact

This ensures the input is always ready for the next project creation without requiring a page refresh or manual clicking.

## Testing

1. Create a project
2. Delete the project (or just create another one)
3. Try to type in the project name input
4. Verify that the input is focused and accepts keyboard input immediately

### DIFF
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -165,6 +165,7 @@ export function NewProjectPanel({
 }: Props) {
   const t = useT();
   const importInputRef = useRef<HTMLInputElement | null>(null);
+  const nameInputRef = useRef<HTMLInputElement | null>(null);
   const [importing, setImporting] = useState(false);
   const [baseDir, setBaseDir] = useState('');
   const [importingFolder, setImportingFolder] = useState(false);
@@ -472,6 +473,12 @@ export function NewProjectPanel({
       designSystemId: primaryDs,
       metadata,
     });
+    // Reset name and refocus input after project creation
+    setName('');
+    // Use setTimeout to ensure the focus happens after any state updates
+    setTimeout(() => {
+      nameInputRef.current?.focus();
+    }, 0);
   }
 
   async function handleImportPicked(ev: React.ChangeEvent<HTMLInputElement>) {
@@ -589,6 +596,7 @@ export function NewProjectPanel({
         </h3>
 
         <input
+          ref={nameInputRef}
           className="newproj-name"
           data-testid="new-project-name"
           placeholder={t('newproj.namePlaceholder')}


lefarcen:
Agent-to-agent note: related research completed for PR #1322.

RELATED_RESEARCH:
linked_issues=[#1129];
related_prs=[#1135 direct #1129 Electron/native-confirm focus fix, #1161 replaces DesignsTab prompt/confirm with custom dialogs, #732/#738 prior Electron window.prompt-to-modal fixes, #1167/#1195/#621/#708 verified NewProjectPanel-touching PRs];
none_reason=additional related PRs found;
search_terms=[focus, NewProjectPanel, input, project creation];
checked_sources=[gh issue search, gh pr search, gh pr view candidate inspection, synclo_memory_search unavailable];
recommended_issue_comments=[#1129 should note that #1322 also links the issue but only changes NewProjectPanel; #1135 remains the broader confirmed desktop/native-confirm focus-root fix, so verify the delete/cancel desktop flow before closing]

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -472,6 +473,12 @@ export function NewProjectPanel({
       designSystemId: primaryDs,
       metadata,
     });
+    // Reset name and refocus input after project creation
+    setName('');
```
lefarcen, line 477: **P2**: This runs immediately after calling `onCreate`, not after a project is successfully created. In `App.tsx`, `onCreateProject` is async and navigates away only after `createProject()` succeeds, so this can clear the user-entered name even if creation fails, and on success the panel is about to unmount anyway.

Concrete fix: do not reset/focus here unless `onCreate` returns an explicit success signal and this component is still mounted; for #1129, fix the delete-confirm focus restoration path instead.

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -472,6 +473,12 @@ export function NewProjectPanel({
...
+    // Use setTimeout to ensure the focus happens after any state updates
+    setTimeout(() => {
```
lefarcen, line 479: **P2**: A `setTimeout(...focus())` in the create handler does not cover the linked bug. #1129 reproduces after the native delete confirmation closes, including the cancel path where `handleCreate` is never called, and it affects other inputs such as Designs search too.

Concrete fix: move focus restoration to the delete confirmation flow or replace the native `confirm(...)` in `DesignsTab` with an in-app dialog that restores focus, and add a regression test for the delete/cancel-to-input-focus flow.

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -472,6 +473,12 @@ export function NewProjectPanel({
       designSystemId: primaryDs,
       metadata,
     });
+    // Reset name and refocus input after project creation
+    setName('');
+    // Use setTimeout to ensure the focus happens after any state updates
```
lefarcen, line 478: **P3**: The comment says the timeout waits for "any state updates", but the relevant state transition is the parent async create/navigate flow, not just this component's local `setName`. Existing focus timers in this repo are generally effect-scoped with cleanup; this timer is not cancellable and can still fire after unmount or after the user intentionally moves focus.

Concrete fix: replace this reasoning with an effect tied to a real open/success state and clear the timer, or remove the timeout entirely if the actual fix lives in the delete dialog path.

lefarcen (changes requested):
Hey @xxiaoxiong! Thanks for tackling the focus issue. I've reviewed this carefully against the linked issue #1129 and the actual reproduction path. Found that this PR doesn't match the bug's failure scenario:

**Main issue**: #1129 reproduces **after** the delete confirmation closes (including cancel), not after create. This code only runs when you click Create, which doesn't cover the reported bug.

**Three specific concerns** below — please see inline comments. Let me know if you'd like me to clarify the delete confirmation focus restoration path.

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -472,6 +473,12 @@ export function NewProjectPanel({
...
+    setTimeout(() => {
+      nameInputRef.current?.focus();
+    }, 0);
```
PerishCode, line 481: This block introduces a UX regression on the failure path that should land before merge.

**What happens today:** `onCreate` here is the synchronous `EntryView.handleCreate` (`apps/web/src/components/EntryView.tsx:313-315`), which fire-and-forgets `onCreateProject(input)`. The parent `App.handleCreateProject` (`apps/web/src/App.tsx:615-622`) awaits `createProject(...)` and bails on a null result:

```ts
const result = await createProject({ ... });
if (!result) return;
setProjects(...);
navigate({ kind: 'project', ... });
```

So on a daemon-down, 5xx, or network error, navigation never happens and the user stays on the create panel. As written, `setName('')` runs unconditionally the moment the click is dispatched, so a failed create now silently wipes the user's typed name with nothing to retry from. Before this PR the name survived the failure, which is the better behavior.

**Why it matters:** the reset is meant as a convenience for a success path that, on success, also unmounts this panel via `navigate(...)` — so the only path it actually changes is the failure path, and it changes it for the worse. That's the opposite of the intended impact.

**Concrete fix — let the parent own the reset so it can fire only on success.** One shape that fits the existing types:

1. Have `App.handleCreateProject` return `Boolean(result)` instead of `void` (the value is already in scope at `App.tsx:615`).
2. Update the prop types so the chain propagates that signal: `onCreateProject: (input: CreateInput & { pendingPrompt?: string }) => Promise<boolean>` on `EntryView` `Props`, and `onCreate: (input: CreateInput) => Promise<boolean>` on `NewProjectPanel` `Props`. `EntryView.handleCreate` becomes `return onCreateProject(input);`.
3. Here, `await` the result and gate the reset on it:

```ts
const created = await onCreate({
  name: name.trim() || autoName(tab, t),
  skillId: skillIdForTab,
  designSystemId: primaryDs,
  metadata,
});
if (!created) return;
setName('');
nameInputRef.current?.focus();
```

On the success branch the panel unmounts so the focus call is a safe no-op; on the failure branch the typed name is preserved.

Note this lands on top of @lefarcen's existing CHANGES_REQUESTED — the linked-issue mismatch with #1129 (which reproduces from the delete-confirm cancel path in `DesignsTab`, not from a successful create) and the uncancellable `setTimeout` still apply independently and aren't addressed by this change alone.


PerishCode (commented):
Hey @xxiaoxiong, thanks for taking a swing at #1129. Reviewing this against the live diff at e2e72d7, my read aligns with @lefarcen's existing CHANGES_REQUESTED — the failure-path reset and the #1129 reproduction mismatch are both blockers and aren't yet addressed on this head. I've left one inline focused on the failure-path regression with a concrete signature shape that should make the resolution straightforward; the linked-issue scope and the uncancellable-timer concerns from that earlier review still apply independently.



lefarcen (commented):
Hey @xxiaoxiong 👋 thanks for writing up the focus scenario and linking #1129. Separately from the implementation blockers already covered in review, could you also update the PR description to match the repo template? That will make the next review pass easier to scope.

The missing pieces are:
- **Why**
- **What users will see**
- **Surface area**
- **Bug fix verification**
- **Validation**

If you are using an AI assistant, you can paste this prompt to have it rewrite the body cleanly:

> Read the current PR body for nexu-io/open-design#1322 and the template at
> Then update via `gh pr edit 1322 --repo nexu-io/open-design --body-file -`.

No rush — this just helps the next iteration go faster. 🙏

lefarcen:
Hey @xxiaoxiong, just checking in — it's been a few days since the review feedback landed on this PR and we haven't heard back. Are you still working on addressing the blockers?

Quick summary of what needs attention (from the existing CHANGES_REQUESTED and @PerishCode's follow-up):

1. **Failure-path mismatch**: The fix only runs on the Create path. The confirmed reproduction of #1129 happens *after* the delete confirmation closes (whether OK or Cancel) — that path isn't covered here.
2. **Failure-path reset**: The state reset logic on the failure path has a concern that needs addressing.
3. **Uncancellable timer**: The `setTimeout` reference isn't cleaned up, which can cause a no-op state update on unmount.

Happy to clarify any of these if it helps move things forward. If you're not sure how to approach the delete-confirmation path, feel free to ask — the thread on #1129 has good context on the confirmed root cause.

(Note: issue #1129 has been re-opened for community contribution in case you'd like to take a broader approach.)

PR #1338, opened 2026-05-12
feat: Critique Theater wireup (activate the stack, M0 dark-launch by default)

Nagendhra-web:
<div align="center">

# 🎭 Critique Theater · Wireup

**The activation PR. Four atomic commits, about a hundred and thirty lines of source, and the entire stack we have been building since Phase 7 starts pulling its weight.**

phase
source
daemon-tests
web-tests
stacked-on

</div>

> **TL;DR.** Connects the four halves the prior stack has been laying down: the daemon's spawn handler now reads the four-tier `isCritiqueEnabled` resolver instead of just `OD_CRITIQUE_ENABLED`, the web's project workspace mounts `<CritiqueTheaterMount>` whenever a real run starts, `SKILL.md od.critique.policy` frontmatter participates in routing, and the project's `metadata.critiqueTheaterEnabled` blob participates too. Default state on a fresh install is unchanged (M0 dark-launch, everything off). 130 lines of source plus a small spawn-input glue module and 13-case test that pins the metadata-narrowing safety property.

**Stacks on:** #1320 (Phase 15). Base = `main` for review convenience; the actual delta above `main` is the activation commits at the head of this branch.

---

## 🤔 Why

For five PRs the stack has been laying foundations the request path could not yet see: a parser nothing called, a reducer nothing fed, a mount nothing rendered, a resolver nothing consulted. That asymmetry was deliberate. Each preparatory PR was small enough to land green and be reviewed on its own merits, with the activation pulled into one focused PR at the end where the integration risk concentrates.

The pain this PR closes:

1. **`critiqueCfg.enabled` was just an env var.** The spawn handler at `server.ts` was reading a single boolean off the env-var-only config. Four other guards (brand, skill, plain adapter, non-media) were doing real work; this one was the bottleneck where a v1 install could not express anything richer than a global on/off. Skill-level vetoes, per-project overrides, staged rollouts, all of those Phase 15 machinery exists for, none of it could fire.
2. **The Theater was unmountable.** `<CritiqueTheaterMount>` was a self-contained drop-in (Phase 9) but `ProjectView.tsx` did not actually render it. The reducer, the SSE subscription, the kill-request handshake, the phase-aware swap, all sat on disk doing nothing.
3. **`od.critique.policy` had no parser.** Phase 15's resolver accepted a `skillPolicy` input but nothing read the SKILL.md frontmatter to produce that input. A skill that declared `required` in YAML had no effect on actual routing.
4. **`metadata.critiqueTheaterEnabled` had no reader.** The web setter (Phase 15.2) round-tripped through the existing settings PATCH endpoint to write the field, but no code on the daemon side read it back to feed the resolver.

Each gap is a one-line edit waiting to be made. This PR makes the four edits and proves they all integrate by running the existing test corpus (which now covers real reducer dispatches and real spawn-handler resolver calls instead of pure-function fixtures).

---

## 👀 What users will see

**On a fresh install today (M0 default):** nothing. `OD_CRITIQUE_ENABLED` is unset (resolver sees `null`). `OD_CRITIQUE_ROLLOUT_PHASE` is unset (resolver sees `M0`). With no skill policy and no project override, the resolver returns the M0 phase default, which is `false`. The composer skips the panel addendum. The orchestrator skips `runOrchestrator(...)`. Generations route through the legacy single-pass code path exactly the way they did before. Zero visual change for users who never opt in.

**After an operator flips `OD_CRITIQUE_ENABLED=1` or a project sets `metadata.critiqueTheaterEnabled: true`:**

- The next generation request on that project sees the live debate stage mount inside `ProjectView`. Five panelist lanes (Designer, Critic, Brand, Accessibility, Copy) open / score / close in sequence as the daemon emits SSE frames. The interrupt button is reachable via keyboard and screen-reader; the dialog-aware Esc gate from Phase 9 (#1315) still applies.
- Clicking Interrupt (or pressing Escape from inside the stage region) POSTs to the daemon interrupt endpoint, waits for a 2xx, then collapses the run to the post-run badge that names the round and composite the user actually interrupted at. A rejected fetch or non-2xx leaves the run on the live stage so the user can retry.
- A skill author who sets `od.critique.policy: opt-out` in `SKILL.md` frontmatter has critique force-disabled for every run using that skill, regardless of env var, project override, or rollout phase. A skill author who sets `policy: required` has critique force-enabled the same way.
- On a German / Japanese / Korean / Traditional Chinese locale, every Theater surface uses native chrome from the Phase 9 i18n keys.

The skill-level veto being unconditional is the user-visible safety property here: a user can never have critique forced on them for a generation flow the skill author explicitly marked opt-out.

---

## 🎯 Surface area

- [x] Daemon (`apps/daemon/src/server.ts`): spawn-time gate now calls `isCritiqueEnabled` with all four tiers, plus the inline `critiqueBus.emit` broadcast loop routed through the existing `emitProjectEvent` helper so empty-Set cleanup applies and future broadcast policy stays uniform.
- [x] Daemon (`apps/daemon/src/skills.ts`): `od.critique.policy` parsed from SKILL.md frontmatter via `normalizeCritiquePolicy` (exported for direct test coverage); derived example cards inherit the parent's policy.
- [x] Daemon (`apps/daemon/src/critique/spawn-inputs.ts`, new file): `narrowProjectCritiqueOverride(metadata)` lifts the inline boolean-narrowing step into a testable helper so a corrupt metadata write cannot accidentally activate the feature via stringified `'true'` or numeric `1`.
- [x] Web runtime (`apps/web/src/components/ProjectView.tsx`): `<CritiqueTheaterMount projectId={project.id} enabled={useCritiqueTheaterEnabled()} />` mounted as a sibling of `<AppChromeHeader>`; renders null until daemon emits `critique.run_started`.
- [x] Web runtime (`apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts`): `setCritiqueTheaterEnabled` setter does read-merge-write through `PATCH /api/projects/:id` (GET → merge → PATCH with GET-fail guard) so other metadata fields (`kind`, `templateId`, `linkedDirs`, etc.) survive the toggle write.
- [x] Docs (`docs/critique-theater.md`): `### Enable / disable` section expanded to describe the four-tier resolver plus the read-merge-write toggle path. The pre-existing "Why an in-CLI panel" / "Stop conditions" / "Why these weights" rationale sections preserved from main.
- [ ] E2E (`e2e`)

`OD_CRITIQUE_ENABLED` and `OD_CRITIQUE_ROLLOUT_PHASE` env-var consumers, the `od.critique.policy` frontmatter key, and the `metadata.critiqueTheaterEnabled` project-metadata field are all listed in the extension-point columns since they are new operator-facing surfaces.

---

## 📸 Screenshots

Intentionally none on this push. The mount stays invisible by default (M0 rollout keeps the resolver returning `false` on a fresh install), and the dedicated Settings UI control that exposes the toggle to end users lands in the focused follow-up (#1484). Screenshots of the live debate stage, the collapsed post-run badge, the degraded banner, and the native locale chrome land in #1484 alongside the Settings UI surface.

---

### 1 · feat(daemon): route critique spawn-path eligibility through the rollout resolver

Today the daemon's spawn handler reads `critiqueCfg.enabled`, which is just the `OD_CRITIQUE_ENABLED` env var. No skill policy. No project override. No rollout phase. After this commit, that single boolean becomes the resolver call from #1320:

```ts
const critiqueEnabledForRun = isCritiqueEnabled({
  phase: parseRolloutPhase(process.env.OD_CRITIQUE_ROLLOUT_PHASE),
  skillPolicy: null,
  projectOverride: null,
  envOverride: parseEnvEnabled(process.env.OD_CRITIQUE_ENABLED),
});
```

The four downstream guards (brand, skill, plain adapter, non-media) stay where they are. The composer continues to ignore the critique fields when the top-line gate is `false`. The orchestrator continues to gate `runOrchestrator(...)` behind `critiqueShouldRun`. The lockstep contract the existing comment already promises now actually holds across the env + phase axes.

`skillPolicy` and `projectOverride` are passed as `null` in this commit. Commits 3 and 4 wire them in.

### 2 · feat(web): mount CritiqueTheaterMount in ProjectView

The single block everyone has been waiting for:

```tsx
const critiqueTheaterEnabled = useCritiqueTheaterEnabled();

return (
  <div className="app">
    <CritiqueTheaterMount
      projectId={project.id}
      enabled={critiqueTheaterEnabled}
    />
    <AppChromeHeader ... />
    ...
  </div>
);
```

Sits as a sibling of `<AppChromeHeader>` at the top of the project workspace. Returns null until the daemon emits `critique.run_started` for the active project. Users who never opt in see zero visual change.

The `enabled` value comes from `useCritiqueTheaterEnabled()` (Phase 15.2), so the hook's full lifecycle is finally exercised in production: localStorage seed on mount, `storage` event listener for cross-tab toggles, `open-design:critique-theater-toggle` CustomEvent listener for same-tab toggles, teardown on unmount.

### 3 · feat(daemon): resolve `od.critique.policy` frontmatter at the spawn site

Replaces the placeholder `skillPolicy: null` with the actual value parsed from the active skill's `SKILL.md` frontmatter:

- `SkillInfo` gains a `critiquePolicy: SkillCritiquePolicy` field carrying the parsed `od.critique.policy` token.
- `listSkills()` populates the field via `normalizeCritiquePolicy` which tolerates YAML casing (`REQUIRED` / `Required`), trims whitespace, and rejects typos (`opt_in` underscore, `require`, `optional`) to `null` so a misspelled token cannot accidentally force the panel on or off.
- `server.ts` captures `skill.critiquePolicy` into a hoisted `skillCritiquePolicy` variable inside the existing skill-load block and threads it into the `isCritiqueEnabled` call.

A skill that declares `required` forces the panel on regardless of project / env / phase. A skill that declares `opt-out` forces it off. A skill that declares `opt-in` lets the panel run at M2 and above. Missing or unrecognised tokens fall through to the lower-priority tiers.

### 4 · feat(daemon): feed `projectOverride` into the rollout resolver from project metadata

Replaces the placeholder `projectOverride: null` with the value the M1 Settings toggle writes onto the project's metadata blob (`critiqueTheaterEnabled?: boolean`).

The read is defensive at the boundary. The metadata blob round-trips through SQLite as a free-form JSON blob, so the spawn handler narrows to boolean via `narrowProjectCritiqueOverride` and falls through to `null` for anything else. A missing key, a malformed value, a string `'true'`, a number `1`, or a project that has never visited Settings all collapse to `null`. The fall-through is the load-bearing safety property: a corrupt metadata write must never accidentally activate the feature.

The write path flows through the existing `PATCH /api/projects/:id` endpoint that already accepts metadata patches, with a read-merge-write helper inside `setCritiqueTheaterEnabled` so other metadata fields (`kind`, `templateId`, `linkedDirs`, `importedFrom`, etc.) survive the toggle write.

After this commit, every tier of the rollout resolver's priority matrix is live:

- `skillPolicy` from `SKILL.md od.critique.policy`
- `projectOverride` from project `metadata.critiqueTheaterEnabled`
- `envOverride` from `OD_CRITIQUE_ENABLED`
- Rollout phase default from `OD_CRITIQUE_ROLLOUT_PHASE`

---

## ✅ Validation

Commands run on the current head (`aa1b578a`, after folding in the PerishCode P3 trio on top of the v0.7.0 main merge):

| Command | Result |
|---|---|
| `pnpm --filter @open-design/web typecheck` | clean (FileViewer `projectKind` prop fix carried forward from the v0.7.0 merge) |
| `pnpm --filter @open-design/daemon test tests/critique-rollout.test.ts` | 10 / 10 (resolver matrix) |
| `pnpm --filter @open-design/daemon test tests/critique-spawn-inputs.test.ts` | 13 / 13 (new; covers `normalizeCritiquePolicy` typo-reject + `narrowProjectCritiqueOverride` stringified / numeric / nested rejections) |
| `pnpm --filter @open-design/daemon test tests/critique-conformance.test.ts tests/critique-adapter-degraded.test.ts` | 18 / 18 |
| `pnpm --filter @open-design/web test tests/components/Theater` | 1055 tests across the suite (Mount, InterruptButton, reducer, SSE wire layer, locale alignment, hook lifecycle) |
| Default-off behaviour on a fresh install (no env / no project override / no skill policy) | resolver returns `false`, legacy path runs |
| Skill `required` opt-in | resolver returns `true` regardless of env / phase |
| Skill `opt-out` opt-in | resolver returns `false` regardless of env / phase |
| Project `metadata.critiqueTheaterEnabled: true` | resolver returns `true` regardless of env / phase |
| `OD_CRITIQUE_ENABLED=1` only | resolver returns `true`, downstream guards still apply |

The orchestrator continues to gate on `critiqueShouldRun` (which already incorporates the resolver result via `critiqueEnabledForRun`), so prompt and orchestrator stay in lockstep across every input the spec defines.

---

## 🔄 How the two halves cooperate

```
┌──────────── user (or integrator) flips the toggle ────────────┐
│                                                                │
│   setCritiqueTheaterEnabled(true)  ──┐                         │
│                                      │                         │
│   ┌──────────── writes ──────────────┘                         │
│   │                                                            │
│   ▼                                                            │
│   localStorage['open-design:config']                           │
│       .critiqueTheaterEnabled = true                           │
│   │                                                            │
│   └─── dispatches CustomEvent ───┐                             │
│                                  │                             │
│                                  ▼                             │
│        every mounted useCritiqueTheaterEnabled                 │
│        hook flips the prop                                     │
│   │                                                            │
│   ▼                                                            │
│   <CritiqueTheaterMount enabled={true} ...>                    │
│                                                                │
│   ┌── GET /api/projects/:id ──┐                                │
│   │   { metadata: {...} }      │ ◄── read-merge-write          │
│   │                            │     skips PATCH on GET-fail   │
│   ├── merge in toggle ─────────┘                               │
│   │                                                            │
│   └── PATCH /api/projects/:id with merged metadata             │
│                                                                │
└──── next generation request hits the daemon ──────────────────┘
                                  │
                                  ▼
            ┌───────── server.ts spawn handler ─────────┐
            │                                           │
            │  isCritiqueEnabled({                      │
            │    phase: parseRolloutPhase(env),         │
            │    skillPolicy: skill.critiquePolicy,     │
            │    projectOverride: narrowProjectCritique │
            │                       Override(metadata), │
            │    envOverride: parseEnvEnabled(env),     │
            │  })                                       │
            │                                           │
            │  ┌─── true  ──► critiqueShouldRun = true  │
            │  │                                        │
            │  └─── false ──► legacy single-pass path   │
            │                                           │
            └───────────────────────────────────────────┘
```

The web layer never tells the daemon what to do. It writes to the project's metadata row through the existing endpoint, and the daemon reads from that row when the spawn handler runs. Two independent processes, one shared source of truth.

---

## 🧮 What activates after this PR ships

The deferrals scattered across the prior five PRs turn into live behaviour:

| Surface | Phase | State pre-wireup | State post-wireup |
|---|---|---|---|
| `parseCritiqueStream` | 2 | Tested against fixtures only | Fed by the orchestrator's CLI stdout for every M2+ run |
| `scoreboard.ts` composite math | 3 | Pure function, no callers in the request path | Called on every `round_end` |
| `critique_runs` SQLite table | 4 | Schema migration only | Written on every shipped / degraded / failed run |
| `<CRITIQUE_RUN>` prompt addendum | 5 | Gated by an env-only flag | Gated by the full resolver |
| `/api/projects/:id/critique/:runId/interrupt` | 6.1 | No caller in the routed app | Called by the mount's kill-request handshake |
| `/api/projects/:id/critique/:runId/artifact` | 6.2 | No caller | Called by the replay surface |
| `useCritiqueStream` + `useCritiqueReplay` | 7 | Never mounted | Driven by the real project events feed |
| Theater component package | 8 | Snapshot-tested, never rendered | Rendered inside `ProjectView` |
| `i18n critiqueTheater.*` keys | 9 | Present in `Dict`, no consumer | Consumed by the rendered Theater |
| Adapter conformance harness | 10 | Tested against fixtures | Becomes the nightly cron's input once Phase 12 ships |
| Phase 11 e2e suite | 11 | Parked behind `test.describe.fixme` | Flips to live in the activation follow-up (#1483) |
| Reducer p99 + coverage walker | 13 | Run against the test corpus | Runs against real reducer dispatches in CI |
| User guide + AGENTS maps | 14 | Pointed at a feature users could not reach | Pointed at a feature users can reach |
| Rollout resolver + Settings toggle hook | 15 | Two pure surfaces with no callers | Both surfaces have callers |
| `od.critique.policy` per-skill veto / required | 15 | Resolver tier with no input wired | Sourced from `SKILL.md` frontmatter |
| `critiqueTheaterEnabled` per-project override | 15 | Resolver tier with no input wired | Sourced from project metadata |

Every row above is one or two lines of code already on disk that gets executed for the first time the day this PR merges.

---

## ⛔ What this PR deliberately does not do

- **No Settings UI button.** The setter `setCritiqueTheaterEnabled` is already callable from any integrator, and the daemon already reads `metadata.critiqueTheaterEnabled`. The dedicated Settings panel control, the label copy, and the per-project toggle UX are user-visible decisions that ship as a focused follow-up PR (#1484).
- **No Prometheus counters or structured logs.** That is the Phase 12 PR (#1485), queued to follow this one so the counters have traffic to count.
- **No `OD_CRITIQUE_ROLLOUT_PHASE` promotion past M0.** Promotion is operator-driven by design and follows the rollout runbook.
- **No removal of the `test.describe.fixme` on the Phase 11 e2e suite.** That happens in the Phase 11 activation follow-up (#1483) which also commits the first PNG baselines.

These are not "missing"; they are sequenced. Each one is its own focused PR.

---

## 🧮 Diff anatomy

```
 apps/daemon/src/server.ts                                       + 70 / - 17
 apps/daemon/src/skills.ts                                       + 38
 apps/daemon/src/critique/spawn-inputs.ts                        + 45   (new; spawn-input glue helper)
 apps/daemon/tests/critique-spawn-inputs.test.ts                 + 115  (new; 13-case coverage for the glue)
 apps/web/src/components/ProjectView.tsx                         + 22
 docs/critique-theater.md                                        + 30   (Enable/disable section expanded; rest preserved from main)
```

About a hundred and thirty lines of source across three daemon + one web file, plus the spawn-input helper + test pair PerishCode flagged on the prior head. Zero edits to anything on the orchestrator request path that was not already on the activation surface.

---

## 📝 A note for the reviewer

Five PRs of preparation, four atomic commits of activation, about a hundred and thirty lines of source. The asymmetry is the whole point. If the activation needed more than this, that would be evidence one of the prior phases had not actually finished its job. Every line in this diff is doing exactly one thing the prior phases were waiting for.

Read it commit by commit. Each one is small enough to hold in your head, and the four together tell the full activation story.


PerishCode:
@Nagendhra-web I'm holding off on generating review comments for nexu-io/open-design#1338 because this pull request has merge conflicts right now.

lefarcen (commented):
Hey @Nagendhra-web! 🎉 This is an impressively thorough activation PR.

**Overall:** The wireup architecture looks sound — daemon resolver integration + web mount are the right integration points, and the phase-based rollout strategy (M0 dark by default) is the safe way to activate a large stack. The verification matrix in the PR description is excellent.

**Areas for maintainer review:**

1. **Security surface expansion:** This PR makes the critique pipeline live (when enabled). The pipeline processes LLM outputs and user inputs. Main security questions:
   - Input sanitization for the critique stream parser?
   - XSS risk in the Theater UI components rendering LLM-generated content?
   - Sandbox escape surface in the orchestrator when processing adapter output?
   
   These are mostly pre-existing surfaces from prior phases, but now they're live.

2. **Rollout resolver priority tiers:** The resolver implements a clear priority: `skillPolicy` veto/require → `projectOverride` → `envOverride` → phase default. This is good, but:
   - The `skillPolicy` read from frontmatter (commit 3 per PR body) — is there input validation on the `od.critique.policy` value?
   - The `projectOverride` from settings row (commit 4 per PR body) — same question on type validation.
   
   These are follow-up commits in the same branch, so may already be covered.

3. **Test coverage:** Looks extensive (22 daemon tests, 181 web tests per verification matrix). The key integration path to verify:
   - End-to-end: env var set → daemon resolver returns true → spawn handler routes through critique path → web mount renders → SSE stream flows → UI updates.
   - This is the e2e suite that's currently behind `test.describe.fixme` per PR body.

4. **Docs completeness (Lens B):** The `critique-theater.md` doc and AGENTS.md are new/updated. Key questions:
   - Failure modes: what happens if the orchestrator crashes mid-run? Does the UI degrade gracefully?
   - Rollout gates: the phase transitions (M0→M1→M2→M3) — are there metrics/observability to inform when to promote?
   - Open questions: "Phase 12 observability" is listed as a follow-up. What is shipped vs deferred?

**Not blocking, but worth discussing:**
- The PR body says "fifty-eight lines of source" but the diff shows 6493 lines. I think the body is referring to the net new integration logic (daemon + web wireup), and the rest is test coverage + docs + i18n. Clarifying this in the commit message or PR description would help reviewers.
- The base branch is `main` but the PR body says "stacks on #1320 (Phase 15)". Is #1320 already merged into main, or does this need to be rebased?

**Recommendation:** This is a high-risk/high-value PR (risk/high label is correct). I'd defer final approval to @PerishCode or the on-call maintainer after verifying:
1. The security questions above (or confirming they're covered in prior phases).
2. The e2e suite passes when un-fixme'd (or is deliberately deferred with a plan).
3. The follow-up commits (3 & 4) for skillPolicy + projectOverride are reviewed.

Thanks for the detailed write-up and verification work! 🙌

e2e/ui/critique-theater.test.ts
```diff
@@ -0,0 +1,195 @@
...
+
+test.describe('Critique Theater e2e (Phase 11)', () => {
+  test.beforeEach(async ({ page }) => {
+    await bootAppWithCritiqueEnabled(page);
+  });
+
+  // Task 11.1: happy path
+  test('renders the live stage, all five lanes, and the shipped badge', async ({ page }) => {
+    await stubProjectEvents(page, TRANSCRIPT);
+    await page.goto('/');
```
PerishCode, line 146: **Phase 11 e2e never reaches `ProjectView`, so the Theater can't mount.**

This suite is now wired into `test:ui:extended` (see `e2e/package.json:9`) and starts every test with `await page.goto('/')` (here, plus lines 151, 171, 185). `CritiqueTheaterMount` is rendered only inside `ProjectView` (`apps/web/src/components/ProjectView.tsx:2269`), which the home route at `/` does not show — other extended suites navigate explicitly via `/projects/${id}` (cf. `e2e/ui/app-design-files.test.ts:918`). `bootAppWithCritiqueEnabled` only sets the `open-design:config` localStorage blob; nothing creates a project or stubs `/api/projects` to land the route on one.

Why this matters: every assertion in the file expects `getByRole('region', { name: 'Design Jury' })` or `'Shipped'` to be visible within 5 s; with no `ProjectView` mounted, no `CritiqueTheaterMount` is rendered, no SSE subscription opens, and the stubbed `/api/projects/*/events` route is never hit. The four `test(...)` cases will time out the first time this CI lane is exercised, and the PNG-baseline cases (lines 165-176) will additionally fail for lack of snapshots — which contradicts the PR body's claim that Phase 11 stays behind `test.describe.fixme` (it isn't `fixme`-gated here, and it is in the `test:ui:extended` list).

Evidence: there are no `page.route('**/api/projects', ...)` stubs and no manual `goto('/projects/...')` calls in this file; the only navigation is `goto('/')`.

Suggested change: either (a) gate the suite with `test.describe.fixme('Critique Theater e2e (Phase 11)', ...)` and back it out of `test:ui:extended` until the navigation + PNG-baseline follow-up commits ship — matching what the PR body promises — or (b) stub `/api/projects` (and any project-list bootstrap) plus navigate via `goto('/projects/${PROJECT_ID}')` after seeding a fake project, so the Theater actually mounts before the assertions run. Option (a) is the smaller change that aligns the code with the PR's stated scope.


apps/web/src/components/Theater/CritiqueTheaterMount.tsx
```diff
@@ -0,0 +1,132 @@
...
+  );
+}
+
+function bestRoundOf(state: Extract<CritiqueState, { phase: 'running' }>): number {
+  let best = 0;
+  for (const r of state.rounds) {
+    if (typeof r.composite === 'number') best = r.n;
+  }
+  return best || state.activeRound;
+}
```
PerishCode: **Optimistic `interrupted` dispatch can pair a `bestRound` from one round with a `composite` from another.**

`bestRoundOf` walks `state.rounds` and overwrites `best = r.n` for every round that has a numeric `composite`, so it returns the index of the *last* round with a composite — not the round whose composite is highest. `bestCompositeOf` (line 126) correctly returns the *maximum* `composite`. When the user hits Interrupt mid-run, `onInterrupt` synthesizes `{ type: 'interrupted', bestRound: bestRoundOf(state), composite: bestCompositeOf(state) }` (line 100-103), so the two functions can disagree: e.g. round 1 closes at composite 8.5, round 2 closes at composite 6.0, and the optimistic frame ships `bestRound: 2, composite: 8.5`.

Why this matters: the reducer's `interrupted` phase is sticky, and the UI's terminal summary ("Interrupted at round N, composite X") is the surface the user sees until — and if — the daemon emits the real `critique.interrupted` event over the new per-project sink. For the window between the click and the daemon's terminal event (or forever, if the kill endpoint isn't wired yet per the inline comment on lines 87-95), the displayed pair is internally inconsistent.

Suggested change: pick the round with the highest composite in a single pass, e.g.

```ts
function bestRoundOf(state: Extract<CritiqueState, { phase: 'running' }>): number {
  let bestRound = 0;
  let bestComposite = -Infinity;
  for (const r of state.rounds) {
    if (typeof r.composite === 'number' && r.composite > bestComposite) {
      bestComposite = r.composite;
      bestRound = r.n;
    }
  }
  return bestRound || state.activeRound;
}
```

and have `bestCompositeOf` reuse the same tuple, or fold both into a single `bestRoundAndComposite(state)` helper so the two values can't drift.


PerishCode (commented):
Activation looks correctly sequenced and the two Codex P1/P2 fixes in `20ec51c` land cleanly:

- `server.ts:3097` now builds `{ ...critiqueCfg, enabled: true }` for the composer when `critiqueShouldRun` is true, so the `cfg.enabled` gate inside `composeSystemPrompt` agrees with the resolver decision on every input axis (skill / project / env / phase). The 10 `critique-rollout` cases already exercise the resolver matrix that feeds it.
- `critiqueBus.emit` now dual-fans to `design.runs.emit` (legacy `/api/runs/:runId/events`) AND the per-project `activeProjectEventSinks` Set, and the round-trip type encoding lines up: the daemon stamps `payload.type = e.event` (`critique.<variant>`), `project-routes.ts:281` uses that as the SSE channel name, and `sseToPanelEvent` on the web spreads the parsed payload first then pins `type` from the channel back to the bare variant — exactly what the contracts package's `panelEventToSse` inverse expects.

Two non-blocking follow-ups in the new files; neither blocks merge. Resolver tests, web Theater suite (181/181), and the daemon-side rollout tests look correct against the activated path. Thanks for the unusually clear commit-by-commit story in the PR body.



apps/daemon/src/server.ts
```diff
@@ -2996,14 +3008,49 @@ export async function startServer({
...
+    // Both inputs collapse to `null` when the skill / project has
+    // not expressed an opinion, which is the resolver's "fall through
+    // to env / phase default" signal.
+    // Per-project override: the M1 Settings toggle writes
+    // `critiqueTheaterEnabled` onto the project's metadata blob via
+    // the existing settings round-trip. A boolean wins outright; any
+    // other type (missing key, malformed value) collapses to `null`
+    // so the resolver falls through to the env / phase tiers exactly
+    // the way it did when the toggle had never been touched.
+    const rawProjectOverride =
```
lefarcen: **P2 — M1 project override not wired end-to-end**

This reads `project.metadata.critiqueTheaterEnabled` as the M1 override, but I can't find any Settings UI, daemon endpoint, or web request that writes this field. The web toggle writes `localStorage['open-design:config'].critiqueTheaterEnabled` only.

Result: in M0/M1, a user opt-in renders `CritiqueTheaterMount` (web sees localStorage) while the daemon resolves `projectOverride: null` and skips critique unless env/phase override is set.

Suggestion: Either wire the Settings toggle through the daemon/project settings store (round-trip), or remove the claimed projectOverride support until it exists.

docs/critique-theater.md
```diff
@@ -0,0 +1,204 @@
...
+is a turn in the same agent session, separated by `<PANELIST role="...">`
+tags in the protocol stream. The parser converts those into `panelist_*`
+events that the web reducer consumes via SSE. This keeps the operational
+contract identical to a normal generation: same auth, same env, same logs.
+
+## 3. Settings reference
+
+### Enable / disable
+
+The feature is gated behind the `OD_CRITIQUE_ENABLED` env var and, after
```
lefarcen: **P2 — Rollout docs materially inaccurate**

Docs say the Settings toggle persists into the daemon settings store and both surfaces read the same flag, but current implementation only has localStorage reader/writer and daemon reads project metadata (no round-trip).

This leaves the documented M1 opt-in path non-functional for backend critique runs.

docs/critique-theater.md
```diff
@@ -0,0 +1,204 @@
...
+
+## 6. Troubleshooting
+
+### "Panel offline this run"
+
+The orchestrator emits a `degraded` event when one of these happens:
+
+| Reason | Cause | Remediation |
+|---|---|---|
+| `malformed_block` | The adapter emitted a `<CRITIQUE>` block the parser rejects. | Check the adapter's conformance status (`/api/metrics/critique`). |
```
lefarcen: **P2 — Observability surfaces not implemented**

Troubleshooting/observability guidance points to `/api/metrics/critique` and `od adapters clear-degraded <id>`, but these aren't in this diff.

Either mark them as planned future work or add the implementation; otherwise the documented failure-mode remediation isn't actionable.

lefarcen (commented):
Thanks for resolving the conflicts @Nagendhra-web! 👍

**Re-review summary:** The merge looks clean and you've made progress on wiring the rollout resolver. However, there are some architectural gaps between what's claimed vs. what's actually wired:

1. **M1 project override incomplete** (P2): The daemon now reads `project.metadata.critiqueTheaterEnabled`, but nothing writes this field. The Settings toggle only writes localStorage, so a user opt-in will render the UI but won't actually enable daemon critique runs unless env/phase overrides are set.

2. **Docs accuracy** (P2): The rollout docs claim the Settings toggle persists through the daemon settings store, but the current implementation doesn't round-trip the flag. This makes the documented M1 rollout path non-functional.

3. **Observability surfaces missing** (P2): Docs reference `/api/metrics/critique` and `od adapters clear-degraded`, but these aren't implemented. Either mark them as planned future work or add the implementation.

See inline comments for details and suggested fixes. These are **architectural consistency issues**, not blockers, but they'll cause confusion for users/operators following the rollout docs in M1.

**What's good:**
- Conflicts resolved cleanly ✅
- E2e suite no longer behind fixme ✅
- Skill policy parsing wired in ✅
- Project-SSE forwarding fixed ✅

lefarcen (commented):
Prior P1/P2 findings from PerishCode's review are now addressed:

- **P1 (Phase 11 e2e navigation)**: `e2e/ui/critique-theater.test.ts:138` is now wrapped in `test.describe.fixme`, and the suite is removed from `test:ui:extended` in `e2e/package.json:9`. This matches the PR body's promise to defer e2e activation until PNG baselines + navigation stubs ship.
- **P2 (bestRoundOf / bestCompositeOf mismatch)**: `CritiqueTheaterMount.tsx:98` now reads a single `bestRoundAndComposite()` result, and `:133-147` computes the matching round/composite pair in one pass — no more split helpers that can drift.

Remaining note (P3):
- `apps/web/tests/components/Theater/CritiqueTheaterMount.test.tsx:69` — the P2 logic is fixed, but there is no regression test for the multi-round mismatch case (e.g. round 1 composite 8.5, round 2 composite 6.0). Existing interrupt coverage only asserts a single-round interrupted state. Consider adding a multi-round fixture to lock in the fix.

No P1/P2 blockers remain. Ready for PerishCode / maintainer final review.

apps/web/tests/components/Theater/CritiqueTheaterMount.test.tsx
```diff
@@ -0,0 +1,231 @@
...
+        maxRounds: 3,
+        threshold: 8,
+        scale: 10,
+      });
+    });
+
+    expect(screen.getByRole('region').getAttribute('data-phase')).toBe('running');
+  });
+
+  it('flips the kill button to pending and synthesizes interrupted on click', () => {
```
lefarcen: **P3 — Multi-round interrupt regression test missing.**

The `bestRoundAndComposite` fix at `CritiqueTheaterMount.tsx:133-147` correctly selects the round with the highest composite in a single pass, addressing the P2 finding. However, the existing interrupt test here only exercises a single-round interrupted state, so it wouldn't catch a future regression where rounds 1 and 2 both have composites but round 1's is higher.

Consider adding a multi-round fixture:
```ts
const multiRoundInterruptedState: CritiqueState = {
  phase: 'interrupted',
  rounds: [
    { n: 1, composite: 8.5, ... },
    { n: 2, composite: 6.0, ... },
  ],
  bestRound: 1,
  composite: 8.5,
  ...
};
```

and asserting that `bestRound === 1` (not 2) after interrupt. This locks in the fix so a refactor back to split helpers can't reintroduce the drift.

P3 because the fix itself is correct; this is hardening.

lefarcen (commented):
Hey @Nagendhra-web, commits `89a7c719` and `7b2aeadd` address the prior P2 findings cleanly:

✅ **M1 project override wiring**: `setCritiqueTheaterEnabled` now round-trips through `PATCH /api/projects/:id` with `metadata.critiqueTheaterEnabled`, and the daemon resolver reads from the same `metadata` row.

✅ **Docs corrections**: `docs/critique-theater.md` section 3 now accurately describes the M1 surfaces that exist today (localStorage + project metadata PATCH) and clearly flags Settings UI control / Phase 12 dashboard / CLI wrapper as deliberate follow-ups.

✅ **Test regression**: `7b2aeadd` adds the multi-round interrupt case to `CritiqueTheaterMount.test.tsx`, locking in the `bestRoundAndComposite` fix.

**One new P2 from the M1 commit:**

The project-aware setter at `useCritiqueTheaterEnabled.ts:147-150` PATCHes `{ metadata: { critiqueTheaterEnabled: next } }`, but the existing `PATCH /api/projects/:id` endpoint replaces the entire `metadata` object rather than deep-merging it (aside from special preservation for imported-folder fields). If a project already has metadata like `{ importedFrom: '...', template: '...' }`, calling the setter with a `projectId` would **drop those fields** and leave only `critiqueTheaterEnabled`.

**Suggested fix**: read the current `project.metadata` first, then PATCH with the merged object:

```ts
const currentMetadata = await fetchProjectSettings?.(projectId);
await fetch(`/api/projects/${projectId}`, {
  method: 'PATCH',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({ metadata: { ...currentMetadata, critiqueTheaterEnabled: next } }),
});
```

Or have the daemon's project-routes handler deep-merge `req.body.metadata` with the existing row's `metadata` before persisting.

Everything else looks ready. Thanks for the thorough follow-up commits — the activation surface is nearly there.

lefarcen (commented):
**Prior P1/P2 findings from head 20ec51c5: ADDRESSED ✅**

Commits 89a7c71 and 7b2aeadd fixed:
- M1 project override wiring (daemon resolver now reads `metadata.critiqueTheaterEnabled`, web setter writes it via PATCH)
- Docs accuracy (removed unshipped surfaces, added correct local conformance command)
- Multi-round interrupt test regression (locks in bestRoundAndComposite fix)

---

**New P2 finding on head 7b2aeadd:**

### Metadata partial-write hazard in the toggle setter

```typescript
// apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts:147-150
body: JSON.stringify({ metadata: { critiqueTheaterEnabled: next } }),
```

The setter PATCHes `{ metadata: { critiqueTheaterEnabled: next } }`, but `PATCH /api/projects/:id` (project-routes.ts:189-244) only preserves the three immutable fields (`baseDir` / `importedFrom` / `fromTrustedPicker`), then calls `updateProject(db, id, patch)` which **replaces** the entire metadata object. This drops any other metadata fields the project had (e.g. template state, custom integrator fields).

**Suggested fix:** Either:
1. Web: GET current project first, merge `critiqueTheaterEnabled` into existing metadata, PATCH the merged object
2. Daemon: Change `updateProject` to deep-merge metadata instead of replacing it (benefits all callers)
3. Endpoint: Add a special case for `critiqueTheaterEnabled` alongside the existing immutable-field preservation

Option 2 is the cleanest long-term fix (prevents similar bugs in other metadata writers). Option 3 is the smallest surgical change for this PR.

---

**Non-blocking note:** The e2e suite fixme + the unshipped Settings UI control are correctly deferred per the PR body. The activation surfaces (resolver + mount) are wired and the unit/integration tests are green. This finding doesn't block the wireup itself — it's a polish item for the M1 user-facing toggle path.

apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts
```diff
@@ -0,0 +1,179 @@
...
+  // integrators that drive a non-project surface). Failures are
+  // swallowed so a transient network issue does not unwind the
+  // in-session UI flip; the next save retries.
+  if (options.projectId) {
+    const fetcher = options.fetchProjectSettings
+      ?? ((url: string, init: RequestInit) => fetch(url, init));
+    fetcher(`/api/projects/${encodeURIComponent(options.projectId)}`, {
+      method: 'PATCH',
+      headers: { 'content-type': 'application/json' },
+      body: JSON.stringify({ metadata: { critiqueTheaterEnabled: next } }),
```
PerishCode: **Project-settings PATCH wipes non-preserved `metadata` fields on the first real caller.**

The new round-trip path sends `{ metadata: { critiqueTheaterEnabled: next } }` and nothing else. The daemon side does *not* deep-merge: `apps/daemon/src/db.ts:521-525` in `updateProject()` is `{ ...existing, ...patch }`, so `patch.metadata` replaces the entire `metadata` blob wholesale. The route at `apps/daemon/src/project-routes.ts:196-231` only re-stamps three specific fields back onto the patch — `baseDir`, `importedFrom`, `fromTrustedPicker`. Every other metadata field is dropped.

Why it matters: production `project.metadata` carries other live fields the rest of the app reads — at minimum `kind` (`ProjectView.tsx:778-779`, `DesignsTab.tsx:538-539`), `templateId` (`ProjectView.tsx:778`), and `linkedDirs` (`ChatComposer.tsx:590,599`). The first caller that passes `projectId` here (the Settings UI follow-up, or any integrator who follows § 3 of `docs/critique-theater.md` and the new doc-comment on this setter) will silently corrupt those fields on every toggle: a `kind: 'template'` project loses its `templateId` and stops re-instantiating, a folder-import project loses its `linkedDirs` and stops finding its linked files. The existing `ChatComposer` pattern (`590: { ...base, linkedDirs: [...existing, selected] }`) already shows the convention — partial-metadata callers must read the existing blob first and spread it into the patch.

Why non-blocking on *this* PR: no in-tree caller passes `projectId` today (only the three new vitest cases in `useCritiqueTheaterEnabled.test.tsx:252-296`), so nothing in HEAD actually loses metadata yet. But the surface is documented in `docs/critique-theater.md` § 3 and the function's own JSDoc as the M1 round-trip path, so the issue ships as a latent footgun for the very next integrator.

Suggested change: before issuing the PATCH, fetch the existing project (`GET /api/projects/${projectId}`), spread the existing `metadata` into the body, and overlay `critiqueTheaterEnabled: next` on top — mirroring `ChatComposer.tsx:590`. Roughly:

```ts
if (options.projectId) {
  const fetcher = options.fetchProjectSettings
    ?? ((url: string, init: RequestInit) => fetch(url, init));
  void (async () => {
    try {
      const cur = await fetcher(`/api/projects/${encodeURIComponent(options.projectId!)}`, { method: 'GET' });
      const existingMeta = cur.ok
        ? ((await cur.json()) as { project?: { metadata?: Record<string, unknown> } }).project?.metadata ?? {}
        : {};
      await fetcher(`/api/projects/${encodeURIComponent(options.projectId!)}`, {
        method: 'PATCH',
        headers: { 'content-type': 'application/json' },
        body: JSON.stringify({ metadata: { ...existingMeta, critiqueTheaterEnabled: next } }),
      });
    } catch (err) {
      /* same dev-only warn as before */
    }
  })();
}
```

Alternative if the read-merge dance feels too heavy for a single boolean: introduce a dedicated `/api/projects/:id/settings` endpoint (or extend the existing PATCH route to deep-merge `metadata` rather than replace it) and migrate the setter onto that. Either way, the current shape needs to land before any Settings UI button calls this with a real `projectId`, otherwise the first toggle on a templated or folder-imported project silently destroys its linkage.


PerishCode (commented):
Re-review on the latest commits (`b6721ea` / `89a7c71` / `7b2aead`):

- **PerishCode P1 (Phase 11 e2e under `fixme`)** — Addressed in `b6721ea`. `e2e/ui/critique-theater.test.ts:138` is wrapped in `test.describe.fixme` and the suite is removed from `test:ui:extended` in `e2e/package.json:9`. The file-header docblock (lines 124-137) names the follow-up work cleanly.
- **PerishCode P2 (`bestRoundOf`/`bestCompositeOf` drift)** — Addressed in `b6721ea`. `CritiqueTheaterMount.tsx:133-148` folds both lookups into a single-pass `bestRoundAndComposite` helper, and `:98-104` destructures from one helper so the dispatched `interrupted` action cannot pair a round number from one round with a composite from another.
- **lefarcen P3 (multi-round regression test missing)** — Addressed in `7b2aead`. `CritiqueTheaterMount.test.tsx:232-311` drives the reducer through round 1 @ 8.5 and round 2 @ 6.0, clicks Interrupt, and asserts the collapsed badge shows `round 1` and not `round 2 ... 8.5`. The buggy split helpers would have failed this; the new helper passes it.
- **lefarcen P2 (M1 project override / docs / observability)** — `89a7c71` extends `setCritiqueTheaterEnabled` with an optional `{ projectId }` round-trip via `PATCH /api/projects/:id`, rewrites `docs/critique-theater.md` § 3 around the four-tier resolver, and replaces the deferred-surface references in the troubleshooting table with the harness command and `clearDegraded()` programmatic helper that actually ship today.

One non-blocking item from the new `89a7c71` round-trip path, inline. The activation surface itself otherwise looks sound: daemon resolver wiring at `server.ts:3036-3047` reads `metadata.critiqueTheaterEnabled` correctly, the composer override at `:3109` keeps prompt + orchestrator in lockstep, and the spawn-side guards (`critiqueShouldRun` at `:3076-3080`) match the earlier review's expectations.



lefarcen (commented):
P2 metadata-merge fix looks solid. The read-merge-write pattern now preserves , ,  and any other metadata fields present at GET time, and skips the PATCH entirely when the prefetch fails — exactly what the inline comment promised.

The four new test cases cover the happy path, no-metadata edge case, GET-fail guard, and PATCH-fail swallow. No remaining clobber hazard.

One P3 note (non-blocking):  § 3 still describes the project override as a direct PATCH round-trip, but the implementation now depends on a GET → merge → PATCH flow with GET-fail guard. Update that section so integrators who follow the doc don't copy the old clobber-prone pattern.

Otherwise ready for maintainer merge call.

lefarcen (commented):
P2 metadata-merge fix looks solid. The read-merge-write pattern now preserves `kind`, `templateId`, `linkedDirs` and any other metadata fields present at GET time, and skips the PATCH entirely when the prefetch fails — exactly what the inline comment promised.

The four new test cases cover the happy path, no-metadata edge case, GET-fail guard, and PATCH-fail swallow. No remaining clobber hazard.

One P3 note (non-blocking): `docs/critique-theater.md` § 3 still describes the project override as a direct PATCH round-trip, but the implementation now depends on a GET → merge → PATCH flow with GET-fail guard. Update that section so integrators who follow the doc don't copy the old clobber-prone pattern.

Otherwise ready for maintainer merge call.

lefarcen (commented):
Hey @Nagendhra-web 👋

The merge-from-main push looks safe from the Critique Theater side: the resolver wiring, ProjectView mount, project-sink fan-out, and the read-merge-write project metadata toggle all still line up with the previous review fixes. Nice job keeping the activation surface focused while absorbing the upstream stack. ✨

I did not find new P1/P2 issues in the current PR diff. One operational caveat before merge: current CI is red in `Validate workspace` on `tests/orbit.test.ts` (`lastRun` expected `agent-3` but received `agent-2`). I do not see that tied to the Critique Theater diff, but this PR should stay blocked until that lane is green or the upstream failure is resolved/rerun.

From the code-review pass, the prior metadata-merge concern remains addressed and this is otherwise ready for the maintainer/human-review path once CI is clean. ❤️

lefarcen (commented):
Hey @Nagendhra-web, this push resolves the CI concern cleanly. 👋

I re-read the current diff after the Phase 8/main merge. The Critique Theater activation path still looks consistent: the daemon resolver feeds the same `critiqueShouldRun` decision into both prompt composition and `runOrchestrator`, project-scoped critique events still reach the mount, the project metadata toggle keeps the read-merge-write protection, and the Phase 11 e2e suite remains parked behind `fixme` as documented.

CI is green on the current head now, including `Validate workspace`, so the earlier orbit-test failure looks like the flake you described rather than a regression from this wireup. No new P1/P2 findings from my pass. This is back to the normal human-review / maintainer-decision path from here. ❤️

lefarcen (commented):
Hey @Nagendhra-web, I re-checked the new interrupt follow-up on `09bf396b`. 👋

The wait-for-daemon-ack direction looks right: the mount now snapshots the run at click time, only terminalizes locally on a 2xx interrupt response, and clears `interruptPending` on rejected/non-2xx requests so the real SSE terminal event can still win. The added rejection + non-2xx tests cover the regression shape cleanly, and the previous best-round/composite pairing test still exercises the async dispatch path.

No new P1/P2 findings from this push. `Validate workspace` is green; the build check was still in progress when I reviewed, so this remains on the normal CI + qualified human-review path. ❤️

apps/daemon/src/server.ts
```diff
@@ -3957,7 +4014,45 @@ export async function startServer({
...
+        // and let the sink push the right channel name. The web's
+        // `sseToPanelEvent` overwrites `type` from the channel name on the
+        // way back into a PanelEvent, so this round-trip stays correct.
+        const critiqueProjectIdForBus =
+          typeof projectId === 'string' && projectId ? projectId : null;
+        const critiqueBus = {
+          emit: (e) => {
+            send(e.event, e.data);
+            if (critiqueProjectIdForBus) {
+              const sinks = activeProjectEventSinks.get(critiqueProjectIdForBus);
```
PerishCode: **Inline `critiqueBus.emit` duplicates `emitProjectEvent` and omits its empty-Set cleanup.**

`emitProjectEvent` at `server.ts:1144` already implements the exact broadcast pattern this block recreates, but with one extra step the inline version drops: when every sink throws (or has been deleted) and `sinks.size` hits 0, the helper deletes the projectId entry from `activeProjectEventSinks` so the map does not accumulate empty `Set` objects across the daemon's lifetime. The inline copy here only deletes the individual sink that threw, leaving an empty `Set` behind for that project key.

**Why it matters:** functionally this is currently a no-op — `Map.get(...)` on an empty-Set key returns the empty Set, the `sinks.size > 0` guard skips it, and the next real subscriber re-populates it. So no user-visible bug today. But it forks the broadcast semantics: a future change to `emitProjectEvent` (rate limiting, schema validation, telemetry) silently does not apply to critique events, and the duplication invites drift on a path the AGENTS guide otherwise centralizes.

**Evidence:** compare `server.ts:1144-1156` (helper) against the new `server.ts:4039-4054` (inline copy). The shapes are identical except for the trailing `if (sinks.size === 0) activeProjectEventSinks.delete(projectId);`.

**Suggested change:** replace the inline loop with a call to the existing helper, e.g.

```ts
const critiqueBus = {
  emit: (e) => {
    send(e.event, e.data);
    if (critiqueProjectIdForBus) {
      emitProjectEvent(critiqueProjectIdForBus, { ...e.data, type: e.event });
    }
  },
};
```

That keeps the two transports (run-scoped `send` + project-scoped fan-out) explicit at this site and routes the project fan-out through the same code path every other emitter uses.


apps/daemon/src/server.ts
```diff
@@ -3061,14 +3073,49 @@ export async function startServer({
...
+    // Both inputs collapse to `null` when the skill / project has
+    // not expressed an opinion, which is the resolver's "fall through
+    // to env / phase default" signal.
+    // Per-project override: the M1 Settings toggle writes
+    // `critiqueTheaterEnabled` onto the project's metadata blob via
+    // the existing settings round-trip. A boolean wins outright; any
+    // other type (missing key, malformed value) collapses to `null`
+    // so the resolver falls through to the env / phase tiers exactly
+    // the way it did when the toggle had never been touched.
+    const rawProjectOverride =
```
PerishCode: **Daemon-side coverage gap: the new spawn-path inputs feeding the resolver are not exercised by any vitest in this PR.**

The resolver itself (`isCritiqueEnabled`, `parseRolloutPhase`, `parseEnvEnabled`) is exhaustively covered by `apps/daemon/tests/critique-rollout.test.ts` — that file is the strongest part of the PR. What is **not** covered is the glue around it: the `rawProjectOverride` extraction at lines 3101-3107, the `skillCritiquePolicy` hoist at line 2974, and the dual-sink `critiqueBus.emit` at 4039-4054. All three are pure activation surface that only fires when the resolver's defaults are crossed, so a refactor of the spawn handler can break them and the existing tests still go green.

**Why it matters:** the PR description's verification matrix ("`vitest run tests/critique-rollout.test.ts` → 10/10") asserts the resolver's purity, but `critiqueEnabledForRun`'s real-world inputs are constructed inline in `server.ts`. The three failure modes a future regression is most likely to surface — (a) a non-boolean `metadata.critiqueTheaterEnabled` no longer collapsing to `null`, (b) a typo'd `od.critique.policy` string sneaking past `normalizeCritiquePolicy`, (c) the project-scoped fan-out silently dropping when `projectId` is missing — are all surface-level glue, not resolver logic, and would not be caught by the existing tests.

**Evidence:** the PR changes `apps/daemon/src/server.ts` (+121 lines on this path) and `apps/daemon/src/skills.ts` (+36 lines for `critiquePolicy` plumbing + `normalizeCritiquePolicy`), but the only daemon test added is `critique-rollout.test.ts` against the pure resolver. `grep '-r' 'critiquePolicy' apps/daemon/tests/` returns no hits.

**Suggested change:** add a small vitest under `apps/daemon/tests/` (e.g. `critique-spawn-inputs.test.ts`) that asserts the glue, not the resolver. Three cases is enough to pin the contract:

1. `normalizeCritiquePolicy('REQUIRED')` → `'required'`; `normalizeCritiquePolicy('opt_in')` → `null` (typo'd token does not force the panel on); `normalizeCritiquePolicy(undefined)` → `null`.
2. A `metadata` object with `critiqueTheaterEnabled: 'true'` (string, not boolean) collapses to `projectOverride: null` so the resolver falls through to the env/phase tiers — pin the `typeof rawProjectOverride === 'boolean'` guard.
3. A `critiqueBus.emit` call with `projectId` set fans out to a stubbed `activeProjectEventSinks` entry; with `projectId` null, only the run-scoped `send` is invoked.


docs/critique-theater.md
```diff
@@ -80,29 +51,8 @@ The cast is fixed in v1. Per-skill cast configuration is reserved for v2
...
-   default). The run flips to `failed` with `cause: 'total_timeout'`.
-5. **User interrupt.** The user pressed Esc or clicked Interrupt
-   mid-run. The run flips to `interrupted` with the best round so far.
-
-A run that hits #2 emits the score badge as `Below threshold` (not
-`Shipped`) so the user can tell convergence failed.
-
-After each round the composite is computed from the per-panelist
-weights:
+The orchestrator runs up to **3 rounds** by default. After each round the
```
PerishCode, line 54: **Scope: this wireup PR also deletes user-facing behavior documentation that is not about wireup.**

The rewrite of `docs/critique-theater.md` collapses the previous `### Auto-converging rounds` section's five-item stop-conditions list (1. Threshold reached → ship, 2. Round budget exhausted → `ship_best` fallback, 3. Per-round timeout → `failed: per_round_timeout`, 4. Total timeout → `failed: total_timeout`, 5. User interrupt → `interrupted`) down to a single "runs up to 3 rounds by default" sentence, and removes the weight-rationale paragraph plus the entire `Why an in-CLI panel and not a third-party design lint` subsection. None of that content moved elsewhere — `grep -rn 'Stop conditions\|per_round_timeout\|fallbackPolicy' docs/` returns nothing in the new file.

**Why it matters:** the deleted enumeration is the only place a user reading the docs learns what `Below threshold` means on a score badge, when a run flips to `failed` vs. `interrupted`, what the `fallbackPolicy: ship_best` default actually does, and why Designer is weighted at 0.0. The PR's own description (`What this PR deliberately does not do`) lists doc cleanup as a separately-sequenced follow-up, so dropping these blocks here is the kind of bundled scope the doc explicitly says it is not doing. It also makes the PR's "about a hundred and thirty lines of source across three files" framing under-state the actual diff: the cumulative change against main is 1,127 insertions / 88 deletions across 10 files (`git diff origin/main...HEAD --stat`), with the docs file alone at +102 / -88.

**Evidence:** `git diff origin/main...HEAD -- docs/critique-theater.md` shows three deleted blocks (the `Why an in-CLI panel` section at old-lines 37-66, the `Auto-converging rounds` stop-conditions list at old-lines 83-105, and the weight-rationale paragraph at old-lines 115-128) that are not re-added anywhere in the new file.

**Suggested change:** revert the docs deletions in this PR and limit the `### Enable / disable` rewrite to the actually-relevant change (the new four-tier resolver explanation at new-lines 86-115, which is on-topic for the wireup). Land any cleanup of the stop-conditions enumeration, weight rationale, and third-party-lint subsection in a separate docs-scoped PR so the doc deletion is reviewable on its own merits.


PerishCode (commented):
Hey @Nagendhra-web, the wireup activation reads cleanly on `64ae82d4`. The four-tier resolver inputs (`skillPolicy` from frontmatter, `projectOverride` from metadata, `envOverride`, phase default) are all live at the spawn site, the read-merge-write in `setCritiqueTheaterEnabled` preserves other metadata fields, and the resolver itself has airtight matrix coverage in `critique-rollout.test.ts`. Default-off behavior on a fresh M0 install holds. Three non-blocking observations worth a look before merge, inline below: a daemon-side coverage gap on the new spawn-path glue, the inline `emit` reimplementing `emitProjectEvent` minus its empty-Set cleanup, and a `docs/critique-theater.md` rewrite that deletes user-facing behavior reference (`Stop conditions` enumeration + weight rationale) unrelated to the wireup task.



docs/critique-theater.md
lefarcen, line 54: P2 This rebase appears to have dropped the richer rationale that was on `main` (the in-CLI-vs-third-party-lint discussion and the fuller weights explanation). The latest rebase note says this file was kept to preserve that authored content, but the current file no longer contains it. Please merge that rationale back into the guide, or update the PR note/body if the removal is intentional, so the conflict resolution does not regress the documentation.

lefarcen (commented):
Hey @Nagendhra-web, thanks for doing the careful main rebase and spelling out the conflict choices. I rechecked the conflict-sensitive areas on `64ae82d4`: `server.ts` keeps both the ElevenLabs audio prompt inputs and the resolver-pinned critique config, the skill/project/env/phase resolver path is still wired through the spawn gate, and the read-merge-write toggle path is intact.

I did catch one documentation merge regression inline. Two process items also need attention before this can be merge-ready:

- CI is currently red in `Validate workspace`: `apps/web/tests/components/FileViewer.test.tsx` is missing the now-required `projectKind` prop. That file is not in this PR's changed-file list, so this may be main-merge test drift rather than Critique Theater logic, but the branch still needs a green run or a small fix before merge.
- After #1520's PR template landed on `main`, this PR body is missing the new template sections: **Why**, **What users will see**, **Surface area**, and **Validation**. Since this is a feature activation stack, filling those in will help product/release review scope the user-visible surface quickly. If you're using an AI assistant, this prompt should get it into shape:

> Read the current PR body for nexu-io/open-design#1338 and the template at
> Then update via `gh pr edit 1338 --repo nexu-io/open-design --body-file -`.

No new runtime P1/P2 findings in the Critique Theater activation path from this pass. Once the docs/template/CI bits are settled, this is back to product + pool review. ❤️

lefarcen (commented):
Hey @Nagendhra-web, quick pass on `2d809911`: the CI-specific change is the right small fix — `FileViewer.test.tsx` now supplies the required `projectKind`, and `Validate workspace` is green on this head. I also rechecked the conflict-sensitive Critique Theater path around `server.ts`; no new runtime P1/P2 findings from me.

The outstanding items are still the review/process ones already called out on the prior head:

- `docs/critique-theater.md` still drops the stop-conditions / third-party-lint rationale / fuller weight-rationale content. PerishCode's current review and my prior inline both still apply here.
- The PR body still needs the #1520 template sections filled in: **Why**, **What users will see**, **Surface area**, and **Validation**.

No new inline comments from me on this head. With CI green now, this is down to the docs/body cleanup plus product and qualified-pool review before it can be merge-ready. ❤️

lefarcen (commented):
Hey @Nagendhra-web, thanks for the quick cleanup on `aa1b578a`. I rechecked the three review-thread fixes plus the CI/body/doc follow-ups:

- The project fan-out now goes through `emitProjectEvent(...)`, so the empty-sink cleanup and any future shared broadcast policy apply to critique frames too.
- `narrowProjectCritiqueOverride(...)` pulls the metadata boundary guard into a small helper, and the new `critique-spawn-inputs.test.ts` pins both that guard and `normalizeCritiquePolicy(...)` against the typo/coercion cases PerishCode called out.
- `docs/critique-theater.md` is back to a focused resolver-section addition; the stop conditions, third-party-lint rationale, and weights rationale are restored.
- CI is green on this head.

No new runtime P1/P2 findings from me, and no new inline comments. From my side this head reads clean for the Critique Theater wireup; the remaining gate is product + qualified reviewer approval before it can be merge-ready. ❤️

lefarcen (commented):
Hey @Nagendhra-web, I rechecked `aa4de544` after the latest push/rebase. The Critique Theater wireup pieces I looked at on the previous head are still intact: `emitProjectEvent(...)` is used for the project-scoped fan-out, the spawn-input narrowing helper + tests are present, the read-merge-write toggle path remains covered, and the docs diff is still scoped to the resolver / enable-disable section.

No new runtime P1/P2 findings from me on this head. CI is still in progress right now, so this is not merge-ready yet; after checks finish, the remaining gates are still product review and qualified-pool approval. ❤️

lefarcen (commented):
Hey @Nagendhra-web, I checked `a389fcd7` after the latest shrink/rebase. The remaining diff is much tighter now: spawn-input narrowing + tests, the `emitProjectEvent(...)` reuse, the scoped enable/disable docs update, and the server glue. Those all still line up with the previous pass, and I do not see new runtime P1/P2 issues in the remaining Critique Theater changes.

CI is still running on this head, so I’m not treating it as merge-ready yet. Once checks finish, this still needs the product review and qualified-pool approval gates. ❤️

PerishCode (approved):
@Nagendhra-web — nice job tightening this down to the activation surface. The scoped diff against `main` is genuinely small now (~196 LOC across 5 files), and every line is doing exactly one thing.

What I re-verified on `a389fcd7`:

- **`apps/daemon/src/critique/spawn-inputs.ts`** — `narrowProjectCritiqueOverride` is a clean carve-out of the inline metadata narrowing. The `!metadata || typeof metadata !== 'object'` guard catches `null` (since `typeof null === 'object'`), primitives, and undefined; then the `typeof raw === 'boolean'` check is the load-bearing safety property. A stringified `'true'`, numeric `1`, nested object, or missing key all collapse to `null` and the resolver falls through to env/phase exactly as before.
- **`apps/daemon/src/server.ts`** — both edits are behavior-preserving with one intentional improvement: the project fan-out now goes through `emitProjectEvent(...)`, which adds the `activeProjectEventSinks.delete(projectId)` empty-sink cleanup the inline loop was missing. The spread + `type` packing is identical, and the run-scoped `send(e.event, e.data)` transport above it is untouched, so the two-sink contract documented around `critiqueBus` still holds.
- **`apps/daemon/src/skills.ts`** — `normalizeCritiquePolicy` is now exported solely so the spawn-input test can pin trim/lowercase/typo-reject without touching the filesystem scan in `listSkills()`. No call-site behavior change.
- **`apps/daemon/tests/critique-spawn-inputs.test.ts`** — 13 cases covering canonical tokens, mixed case, whitespace trim, the YAML-author-typo set (`opt_in`, `require`, `optional`, `off`, `on`), non-string fall-through, boolean wins, stringified/numeric rejection, nested rejection, missing-key fall-through, and the sanity case where other metadata fields coexist. Tight, focused, and the right granularity for boundary glue.
- **`docs/critique-theater.md`** — the Enable/disable section now actually describes the read-merge-write behavior of `setCritiqueTheaterEnabled` (GET → merge → PATCH, skip PATCH on GET failure), matching what the daemon reads back via `narrowProjectCritiqueOverride`.

The four-tier resolver (skill policy > project override > env override > phase default) now has every input live at the spawn site, the M0 dark-launch default still keeps prod traffic on the legacy path, and the new helpers are pinned in isolation so a future refactor that loosens either guard will trip the tests instead of silently activating the feature. Thanks for the careful shrink/rebase work to land this in the smallest possible activation shape — really enjoyed reading the commit-by-commit story. 🎭

PR #1351, opened 2026-05-12
fix(web): freeze completed run durations across conversations

### LINKED ISSUE #150: Clicking Stop changes the recorded duration of a completed run and makes it inaccurate
Summary:
After clicking the Stop button, the duration shown for an already completed run changes and no longer matches the actual execution time.

Steps to reproduce:
1. Start a task and let it complete.
2. Observe the completed run and its recorded execution duration.
3. Click the Stop button afterward, or trigger a stop action while the completed state is still visible.
4. Observe the displayed duration again.

Expected behavior:
Once a run is completed, its recorded execution duration should be fixed and should not change afterward.

Actual behavior:
- The displayed duration for the completed run changes after clicking Stop.
- The final duration no longer matches the real execution time.
- A stop action appears to mutate historical timing data that should already be finalized.

Why this is a problem:
- Users cannot trust the execution time shown in the UI.
- Completed task history becomes inaccurate.
- Runtime metrics are important for judging whether a task was normal, slow, or stuck.
- Mutating finished run data makes debugging and comparing runs harder.

Expected fix direction:
- Freeze execution duration once a run reaches completed state.
- Prevent later stop actions from mutating historical timing data for finished runs.
- Separate the lifecycle of completed runs from stop/cancel actions intended for active runs.
- Ensure the displayed duration always reflects the true start-to-finish execution time of that specific run.

Priority:
High. This affects the correctness of core execution history and makes task timing unreliable.

Image

Image

Siri-Ray:
## Summary
- fix Stop/run duration handling so completed assistant messages do not resume live elapsed timing
- scope chat streaming state to the active conversation to prevent cross-conversation timer leakage
- expose latest run duration on conversation DTOs and show fixed run duration in conversation menus
- add regressions for Stop finalization, completed-message duration rendering, and conversation menu duration labels

Fixes #150

## Verification
- pnpm --filter @open-design/web exec vitest run tests/components/conversation-timestamps.test.tsx tests/components/assistant-message-unfinished-todos.test.tsx tests/components/ProjectView.run-cleanup.test.tsx
- pnpm --filter @open-design/contracts typecheck
- pnpm guard
- pnpm typecheck

## Notes
- Full `pnpm --filter @open-design/daemon test` had two intermittent failures in `chat-route.test.ts` and `project-watchers.test.ts`; both failed tests passed when rerun individually.


apps/daemon/src/db.ts
```diff
@@ -681,9 +682,66 @@ export function getConversation(db: SqliteDb, id: string) {
...
+      `SELECT run_status AS runStatus,
+              started_at AS startedAt,
+              ended_at AS endedAt,
+              events_json AS eventsJson
+         FROM messages
+        WHERE conversation_id = ?
+          AND role = 'assistant'
+          AND run_status IS NOT NULL
+        ORDER BY COALESCE(ended_at, started_at, created_at) DESC
+        LIMIT 1`,
```
PerishCode, line 767: **Non-blocking: N+1 + unindexed sort on a hot path**

`latestConversationRunSummary` is now invoked once per conversation from `listConversations` (line 666) and once per `getConversation` call (line 685 / called from many handlers in `apps/daemon/src/project-routes.ts`). That turns conversation listing into an N+1 SQL pattern, and the inner query's `ORDER BY COALESCE(ended_at, started_at, created_at) DESC` can't ride the only index on this table — `idx_messages_conv ON messages(conversation_id, position)` (db.ts:102-103) — so SQLite has to scan all messages for each conversation (filtered by `role = 'assistant' AND run_status IS NOT NULL`) per request.

Why it matters: this PR is `risk/high` partly because the conversation listing is on a frequently-hit path (sidebar, conversation menu, every navigation). For projects with many conversations or long histories, both the per-row subquery and the unindexed sort compound, and the cost is paid on every list/get even when nothing has changed.

Suggested change — pick whichever fits best:

1. Order by the indexed column instead, since assistant messages are inserted in order and `position` is monotonic per conversation:

```sql
ORDER BY position DESC
LIMIT 1
```

That lets SQLite walk `idx_messages_conv` directly and stop at the first qualifying row.

2. Or fold the lookup into a single query in `listConversations` using a correlated subquery / window function, e.g. a `LEFT JOIN LATERAL`-style pattern with `ROW_NUMBER() OVER (PARTITION BY conversation_id ORDER BY position DESC)` filtered to `role = 'assistant' AND run_status IS NOT NULL`, so the listing returns conversations + their latest-run summary in one round-trip instead of `1 + N`.

Either one keeps the externally-observable shape of `Conversation.latestRun` unchanged and only touches the daemon-side implementation.

PerishCode (commented):
@Siri-Ray Thanks — the core fix lines up cleanly with #150: passing `usage.durationMs` into `useLiveElapsed`, gating pane `streaming` to the active conversation via `streamingConversationId`, and surfacing a `Conversation.latestRun` summary on the daemon DTOs all directly target the "completed run elapsed keeps growing across conversations" symptom. The follow-up `isStoppableAssistantMessage` change also addresses the codex bot's note about API runs whose abort path returns without `onDone`/`onError`, and the regression specs in `ProjectView.run-cleanup.test.tsx` / `assistant-message-unfinished-todos.test.tsx` / `conversation-timestamps.test.tsx` exercise the new behavior in isolation. One non-blocking performance note inline on the new daemon query — feel free to defer if you'd rather keep the diff minimal here.

lefarcen (commented):
Hey @Siri-Ray! 👋

Thanks for tackling the Stop button duration bug — this PR addresses the core issue from #150. The approach is solid: scoping streaming state to the active conversation and freezing completed durations.

Found 2 structural concerns during deep review (both P2, non-blocking but should be fixed before merge):

## Code Review (Lens A)

**Race condition on streaming state (P2)** — see inline on ProjectView.tsx:1462

**N+1 query pattern in sidebar (P2)** — see inline on db.ts:650

Both are fixable without changing the design. The test coverage looks solid — good job on the three new regressions.

**Next steps**: address the two inline findings, then this should be good to go. Let me know if you want to discuss either one.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1400,21 +1438,30 @@ export function ProjectView({
...
+            return {
+              ...prev,
+              endedAt,
+              runStatus: finalRunStatus,
+            };
+          });
           if (commentAttachments.length > 0) {
             void patchAttachedStatuses(commentAttachments, 'needs_review');
           }
           setStreaming(false);
```
lefarcen: **P2 — Race condition on streaming state clear**

Completion/error callbacks still clear global `streaming` state unconditionally. Scenario:
1. User starts run A in conversation 1
2. User switches to conversation 2, starts run B
3. Run A finishes later → executes `setStreaming(false)` + `clearStreamingMarker()` from its stale callback
4. This clears B's active streaming marker → can hide Stop button or desync UI

**Fix**: Make the clear operation compare against the captured `conversationId` when clearing, or store streaming state per conversation/run instead of globally.

**Add regression**: Start conv A, switch to conv B, let conv A finish → assert conv B remains streaming.

apps/daemon/src/db.ts
```diff
@@ -648,22 +648,45 @@ function normalizeTemplate(row: DbRow) {
 // ---------- conversations ----------
 
 export function listConversations(db: SqliteDb, projectId: string) {
```
lefarcen, line 681: **P2 — N+1 query performance hit**

`listConversations` now performs one extra synchronous SQLite query per conversation via `latestConversationRunSummary`, and that query sorts by `COALESCE(ended_at, started_at, created_at)` without a supporting index.

On projects with many conversations or long message histories, opening the sidebar can block the daemon event loop and stall API/SSE work.

**Fix**: Compute latest run summaries for all project conversations in one query (e.g., a LEFT JOIN with `GROUP BY conversation_id` or a window function), preferably using message `position` or another indexed column.

**Add coverage**: Daemon-side test that verifies `latestRun` shape is correct and that query performance is acceptable for 100+ conversations.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1399,22 +1453,29 @@ export function ProjectView({
...
               void patchAttachedStatuses(commentAttachments, 'failed');
             }
-            setStreaming(false);
+            clearStreamingMarker(activeConversationId);
+            updateConversationLatestRun('failed', endedAt);
             abortRef.current = null;
             cancelRef.current = null;
             void refreshProjectFiles();
             onProjectsRefresh();
             return;
```
lefarcen, line 1554: **P2 — Race condition still not fixed**

The completion handlers still unconditionally call `setStreaming(false)` and `clearStreamingMarker()` without checking if `streamingConversationId` still matches this run's conversation.

Scenario:
1. User starts run in conversation A → sets `streamingConversationId = 'conv-A'`
2. User switches to conversation B, starts new run → sets `streamingConversationId = 'conv-B'`
3. Conversation A's completion callback fires → calls `setStreaming(false)` + `clearStreamingMarker()`
4. This clears conv B's active marker → Stop button disappears, UI unlocks incorrectly

**Fix**: Capture `conversationId` when starting the stream, then check it before clearing:
```typescript
const capturedConvId = activeConversationId;
// ... in completion callback:
if (streamingConversationId === capturedConvId) {
  setStreaming(false);
  clearStreamingMarker();
}
```

Same fix needed in reattach completion/error/status paths.

lefarcen (commented):
Great progress on the fixes! 🎉

✅ **N+1 query resolved** — The window function approach in `listConversations` is exactly what I suggested. Nice work!

❌ **Race condition still present** — The extraction of `finalizeActiveAssistantMessagesOnStop` is good for testability, but the original race is still there. See inline for details.

Once the race fix is in, this should be good to go.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1399,22 +1455,29 @@ export function ProjectView({
             if (commentAttachments.length > 0) {
               void patchAttachedStatuses(commentAttachments, 'failed');
             }
-            setStreaming(false);
+            clearStreamingMarker(runConversationId);
```
lefarcen, line 1550: **P2 — Stale completion handlers still clear shared refs**

The `clearStreamingMarker(activeConversationId)` now checks conversation match, but `abortRef.current = null` and `cancelRef.current = null` (lines 1458, 1478, 1527) are still cleared unconditionally.

**Scenario**:
1. User starts run in conversation A → sets `abortRef.current = controllerA`
2. User switches to conversation B, starts run → sets `abortRef.current = controllerB`
3. Conversation A's `onDone` fires → nulls `abortRef.current` and `cancelRef.current`
4. User clicks Stop while B is still streaming → `abortRef.current` is already null → Stop button does nothing

**Fix**: Capture the active conversation ID when starting the stream, then guard the ref-clearing:

```typescript
const capturedConvId = activeConversationId;
// ... in onDone/onError:
if (streamingConversationIdRef.current === capturedConvId) {
  if (abortRef.current === controller) abortRef.current = null;
  if (cancelRef.current === cancelController) cancelRef.current = null;
}
```

Same fix needed in all completion paths (onDone, onError, onRunStatus terminal states).

lefarcen (commented):
Nice work on the N+1 fix! 🎉 The window function approach in `listConversations` is exactly what PerishCode and I suggested.

One remaining concern on the race condition — the `streamingConversationId` guard is a solid step forward, but there's still a subtle ref-clearing race in the completion handlers. See inline.

Once that's addressed, this should be good to merge.

Siri-Ray:
**Looper fixer round complete** — 59fb632

- 🔀 Merge conflict


<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=fixer · agent=opencode · An autonomous AI dev team for your GitHub repos.</sub>

lefarcen (commented):
Hey @Siri-Ray, the latest fixer round is in and CI is green. The previous review threads are resolved, so this is now waiting on a fresh human review for the current head. @PerishCode, could you take the next pass when you have a moment? 🙌

PerishCode (approved):
@Siri-Ray Thanks for the thorough follow-through on this one — the head at `59fb6323` now ties off both prior review threads cleanly:

- **N+1 lookup gone**: `listConversations` (`apps/daemon/src/db.ts:681`) now uses a single CTE + `ROW_NUMBER() OVER (PARTITION BY conversation_id ORDER BY position DESC)` to fold the latest-run summary into one round-trip, and `project-status.test.ts` asserts only one `prepare` call across 125 conversations.
- **Cross-conversation timer leakage closed**: pane streaming is gated on `streaming && streamingConversationId === activeConversationId`, the stale-callback race is guarded by the pure `clearStreamingConversationMarker` / `shouldClearActiveRunRefs` helpers (both unit-tested in `ProjectView.run-cleanup.test.tsx`), and `runConversationId` / `reattachConversationId` capture prevents stale completions from stomping a newer run's `abortRef` / `cancelRef`.
- **Stopped API runs finalize**: `isStoppableAssistantMessage` now picks up assistant messages with `runStatus === undefined` but a live `startedAt`, addressing the API/BYOK abort path Codex flagged.
- **Completed duration rendering**: `useLiveElapsed` takes `usage?.durationMs` so finished messages without `endedAt` show a fixed elapsed instead of a live counter, and `conversationMetaLabel` surfaces `latestRun.durationMs` for completed runs in the conversation menu.

The three new regressions in `conversation-timestamps.test.tsx`, `assistant-message-unfinished-todos.test.tsx`, and `ProjectView.run-cleanup.test.tsx` cover the behavioral pivots, CI is all green, and there are no unresolved threads. Nice patient iteration through the fixer rounds. 🙌

PR #1356, opened 2026-05-12
feat: add bulk delete button with toast feedback for projects

### LINKED ISSUE #1112: Feature: add bulk delete button with toast feedback for projects
# Summary

The current project grid does not support deleting multiple projects at once. Users must delete projects individually, which becomes inconvenient when managing many projects.

This proposal adds a simple bulk-delete workflow with a delete button beside the search input and toast feedback after deletion.

# Current behavior

- Projects can only be deleted one-by-one
- No bulk selection or delete action exists
- No success feedback appears after deletion

# Expected behavior

- Users should be able to select multiple projects and delete them together
- A delete icon/button should appear beside the search input
- After deletion, a toast message should confirm success

# Suggested fix

- Add multi-select support for project cards
- Add a delete button beside the search input
- Show a confirmation modal before deletion
- Display a success toast after deletion

Example:

```txt
3 projects deleted successfully
```

# Why this matters

Bulk actions are standard in modern tools like Figma, Notion, and GitHub. This would improve project management, reduce friction, and make the dashboard feel more polished and scalable.

# Implementation approaches

There are two ways to implement bulk delete:

1. Frontend-only approach  
   - Use the existing delete API and call it once for each selected project.  
   - This is simpler to implement but sends multiple requests and may need handling for partial failures.

2. Backend bulk-delete endpoint  
    - Create a backend endpoint that accepts multiple project IDs and deletes them in a single request. 
    - I think this would be the cleaner and more scalable approach for a proper bulk delete workflow.

xxiaoxiong:
Fixes #1112

## Problem

The current project grid only supports deleting projects one-by-one. Users must delete projects individually, which becomes inconvenient and time-consuming when managing many projects. There's also no success feedback after deletion.

## Solution

Added a complete bulk delete workflow with multi-select support, confirmation modal, and toast feedback.

### Multi-Select
- ✅ Checkbox on each project card (top-left corner)
- ✅ Select All / Deselect All buttons in toolbar
- ✅ Visual feedback: selected cards have accent border and tinted background
- ✅ Selection state persists while browsing

### Bulk Delete Button
- ✅ Appears beside search input when projects are selected
- ✅ Shows count: "Delete 3" / "删除 3 个"
- ✅ Styled in red with trash icon for clear visual hierarchy
- ✅ Only visible when at least one project is selected

### Confirmation Modal
- ✅ Shows before deletion with project count
- ✅ Lists project names when count ≤ 5
- ✅ Clear warning: "This action cannot be undone"
- ✅ Cancel / Delete buttons with proper styling
- ✅ Modal overlay with click-outside-to-close

### Toast Feedback
- ✅ Success: "3 project(s) deleted successfully" / "已成功删除 3 个项目"
- ✅ Partial failure: "2 of 3 project(s) deleted" / "已删除 2/3 个项目"
- ✅ Auto-dismisses after 3 seconds
- ✅ Positioned bottom-right, non-intrusive

### Error Handling
- ✅ Gracefully handles partial failures
- ✅ Shows accurate success/failure counts
- ✅ Clears selection after deletion completes
- ✅ UI updates immediately after each deletion

### Approach
Used the **frontend-only approach** as suggested in the issue:
- Calls the existing `deleteProject` API sequentially for each selected project
- Tracks success/failure counts
- Shows appropriate feedback based on results

### Code Structure

**DesignsTab.tsx**:
- Added `selectedProjectIds` state (Set<string>)
- Added `showConfirmModal` and `toastMessage` states
- Implemented `handleToggleProject`, `handleSelectAll`, `handleDeselectAll`
- Implemented `handleBulkDelete` with sequential deletion and feedback
- Added `ConfirmModal` component for deletion confirmation
- Added checkbox to each project card
- Added bulk action buttons to toolbar

**i18n**:
- `designs.bulkDelete.button`: "Delete {count}" / "删除 {count} 个"
- `designs.bulkDelete.selectAll`: "Select all" / "全选"
- `designs.bulkDelete.deselectAll`: "Deselect all" / "取消全选"
- `designs.bulkDelete.confirm.title`: "Delete {count} project(s)?" / "确定删除 {count} 个项目？"
- `designs.bulkDelete.confirm.message`: "This action cannot be undone." / "此操作无法撤销。"
- `designs.bulkDelete.confirm.list`: "Projects to delete:" / "将要删除的项目："
- `designs.bulkDelete.success`: "{count} project(s) deleted successfully" / "已成功删除 {count} 个项目"
- `designs.bulkDelete.partialFailure`: "{success} of {total} project(s) deleted" / "已删除 {success}/{total} 个项目"

**CSS**:
- `.design-card-checkbox-wrapper`: Positioned top-left with z-index
- `.design-card-checkbox`: Styled with accent color
- `.design-card.selected`: Accent border and tinted background
- `.bulk-delete-btn`: Red background with hover state
- `.modal-overlay` / `.modal-content`: Centered modal with backdrop
- `.toast-notification`: Fixed bottom-right notification

## Testing

- ✅ Multi-select works correctly
- ✅ Select All selects all filtered projects
- ✅ Deselect All clears selection
- ✅ Bulk delete button shows correct count
- ✅ Confirmation modal appears before deletion
- ✅ Modal lists project names when count ≤ 5
- ✅ Deletion works for all selected projects
- ✅ Toast shows correct success count
- ✅ Partial failures show correct counts
- ✅ Selection clears after deletion
- ✅ UI updates immediately
- ✅ Keyboard accessible (checkboxes, buttons, modal)
- ✅ i18n works for both English and Chinese
- ✅ Type checking passes

## Impact

This feature brings Open Design's project management up to par with modern tools like Figma, Notion, and GitHub. Users can now:
- Delete multiple projects in one action
- Save significant time when cleaning up projects
- Get clear feedback about what happened
- Recover from mistakes with the confirmation modal

## Screenshots

**Before**: No bulk actions, must delete one-by-one
**After**: 
- Checkboxes on cards
- Select All / Deselect All buttons
- Bulk delete button with count
- Confirmation modal
- Toast feedback

## Future Enhancements

Potential improvements for future PRs:
- Backend bulk-delete endpoint for better performance
- Bulk operations for other actions (archive, export, etc.)
- Keyboard shortcuts (Ctrl+A for select all, Delete key for bulk delete)
- Undo functionality


apps/web/src/components/DesignsTab.tsx
```diff
@@ -342,6 +503,19 @@ export function DesignsTab({
...
+									<input
+										type="checkbox"
+										className="design-card-checkbox"
+										checked={isSelected}
+										onChange={(e) => {
+											e.stopPropagation();
+											handleToggleProject(p.id);
+										}}
+										onClick={(e) => e.stopPropagation()}
+										aria-label={`Select ${p.name}`}
```
PerishCode, line 516: **Localize the checkbox `aria-label` instead of hardcoding English.**

This label reads `aria-label={`Select ${p.name}`}`, which bypasses i18n. The sibling delete button on the same card (a few lines below at L522) already uses `aria-label={t("designs.deleteAria", { name: p.name })}`, and every other user-visible string this PR adds (`designs.bulkDelete.button`, `…selectAll`, `…confirm.title`, etc.) is keyed in both `en.ts` and `zh-CN.ts`. So a zh-CN screen-reader user toggling cards will hear the surrounding interface in Chinese and just this one control in English.

Why it matters: this is the only keyboard/SR-discoverable affordance for the new multi-select feature; the PR description's own Testing checklist asserts "i18n works for both English and Chinese" and "Keyboard accessible (checkboxes, buttons, modal)", so leaving the screen-reader label untranslated regresses both claims.

Suggested fix: add a key (e.g. `'designs.bulkDelete.selectAria': 'Select project "{name}"'` in `en.ts` and `'选择项目「{name}」'` in `zh-CN.ts`) and change L516 to `aria-label={t("designs.bulkDelete.selectAria", { name: p.name })}`.


apps/web/src/components/DesignsTab.tsx
```diff
@@ -24,6 +25,53 @@ type DesignListItem =
...
+				)}
+				<div className="modal-actions">
+					<button onClick={onCancel}>{t("common.cancel")}</button>
+					<button onClick={onConfirm} className="danger">
+						{t("common.delete")}
+					</button>
+				</div>
+			</div>
+		</div>
+	);
```
PerishCode, line 72: **`ConfirmModal` is missing the dialog semantics every other modal in this app provides.**

The new `ConfirmModal` renders as a plain `<div className="modal-overlay">…<div className="modal-content">` with no `role="dialog"`, no `aria-modal="true"`, no `aria-labelledby` / `aria-describedby` hooks pointing at the `<h2>` and the message `<p>`, no Escape-to-cancel handler, and no initial focus management. Tabbing inside the open modal walks straight back into the underlying project grid, and Esc does nothing.

Why it matters: this is a destructive bulk-delete confirmation. A keyboard or screen-reader user who can't escape the modal, can't hear that a dialog opened, and lands their next Tab back on the grid is a user one Enter-press away from confirming the wrong action. The repo already has the working pattern — `apps/web/src/components/PreviewModal.tsx:269` opens with `<div className="ds-modal-backdrop" role="dialog" aria-modal="true" aria-label={…}>`, and `FileViewer.tsx`, `SettingsDialog.tsx`, `QuickSwitcher.tsx`, `SketchEditor.tsx`, `ConnectorsBrowser.tsx`, `PromptTemplatePreviewModal.tsx` all match.

Suggested fix:
- Add `role="dialog" aria-modal="true"` to the inner `.modal-content` wrapper.
- Give the `<h2>` an id (e.g. `bulk-delete-title`) and the description `<p>` an id (e.g. `bulk-delete-desc`), then `aria-labelledby` / `aria-describedby` them on the dialog.
- Add a `useEffect` keyed on `isOpen` that attaches a `keydown` listener calling `onCancel()` on `Escape`, and focuses the cancel button on open (move the cancel `<button>` into a `ref`).


apps/web/src/index.css
```diff
@@ -16969,3 +16969,197 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
...
+  cursor: pointer;
+  accent-color: var(--accent);
+}
+
+.design-card.selected {
+  border-color: var(--accent);
+  background: var(--accent-tint);
+}
+
+.modal-overlay {
```
PerishCode, line 17039: **Avoid claiming generic global class names like `.modal-overlay` / `.modal-content` / `.modal-actions` / `.toast-notification`.**

This PR introduces five very generic top-level class names — `.modal-overlay`, `.modal-content`, `.modal-actions`, `.bulk-delete-list`, `.toast-notification` — plus an unscoped `@keyframes slideIn`, into the shared global stylesheet. Every other modal in the codebase prefixes by feature (`ds-modal-backdrop` for `PreviewModal`, `df-modal-…` for design-files), and existing buttons that opt into the existing `.danger` modifier always scope it inside a parent selector — see this same file at L7335 (`.df-head .df-actions button.danger`), L7950 (`.df-row-popover button.danger`), L9076 (`.comment-card-action.danger`). The new `.modal-actions button.danger` (L17129) and bare `@keyframes slideIn` (L17155) sit in the global namespace with nothing scoping them.

Why it matters: a future PR that adds any other modal or toast — including a refactor that promotes this very component into a shared primitive — will silently inherit or fight this styling. CSS class collisions don't fail typecheck; they show up as a layout regression in production.

Suggested fix: prefix the new classes (e.g. `.designs-bulk-modal-overlay`, `.designs-bulk-modal-content`, `.designs-bulk-modal-actions`, `.designs-bulk-toast`, `@keyframes designs-bulk-toast-slide-in`) and rename the corresponding `className` strings + `.danger` selector in `DesignsTab.tsx`. If you'd rather land a long-lived shared modal/toast primitive, factor it into its own component+stylesheet in a follow-up rather than landing the unprefixed names here.


apps/web/src/components/DesignsTab.tsx
```diff
@@ -187,10 +238,84 @@ export function DesignsTab({
...
+		} else {
+			setToastMessage(
+				t("designs.bulkDelete.partialFailure", {
+					success: successCount,
+					total: idsToDelete.length,
+				}),
+			);
+		}
+
+		setTimeout(() => setToastMessage(null), 3000);
```
PerishCode, line 288: **Toast timer leaks across re-triggers and unmount.**

`setTimeout(() => setToastMessage(null), 3000)` is fired without tracking the handle. If the user runs a second bulk delete while the previous toast is still on screen, the older 3 s timer keeps running and clears the *new* toast prematurely (potentially within < 1 s of it appearing). The same handle also keeps firing if `DesignsTab` unmounts before the timeout, calling `setToastMessage` on a dead instance.

Why it matters: bulk delete is exactly the flow where users repeat the action ("delete 3 — oh, one more — delete that one"). The second toast collapsing instantly looks like a silent failure of the second deletion, which is the opposite of the feedback the toast was added to provide.

Suggested fix: hold the handle in a ref and clear before re-scheduling, plus a single cleanup effect:
```ts
const toastTimerRef = useRef<ReturnType<typeof setTimeout> | null>(null);

useEffect(() => () => {
  if (toastTimerRef.current) clearTimeout(toastTimerRef.current);
}, []);

// in handleBulkDelete, replace the bare setTimeout(...) with:
if (toastTimerRef.current) clearTimeout(toastTimerRef.current);
toastTimerRef.current = setTimeout(() => setToastMessage(null), 3000);
```
Alternatively, drive the auto-dismiss from a `useEffect` keyed on `toastMessage` so the cleanup handles both cases by construction.


PerishCode (commented):
@xxiaoxiong — read through the bulk-delete flow end to end. Four polish items below as inline comments. Two existing unresolved Codex threads already cover the most user-visible issues — the parent card's `onKeyDown` swallowing the checkbox's Space/Enter, and the bulk-action toolbar rendering in kanban view without any per-card selection affordance — please address those alongside what's below. Treating this review as non-blocking, but the i18n regression on the new checkbox `aria-label` and the missing dialog semantics on the destructive-confirmation modal are worth landing before this ships.



apps/web/src/components/DesignsTab.tsx
```diff
@@ -187,10 +238,84 @@ export function DesignsTab({
...
+		setSelectedProjectIds(new Set());
+	};
+
+	const handleBulkDelete = async () => {
+		setShowConfirmModal(false);
+		const idsToDelete = Array.from(selectedProjectIds);
+		let successCount = 0;
+
+		for (const id of idsToDelete) {
+			const success = await deleteProject(id);
```
lefarcen, line 268: **P1 — Double DELETE race**

This calls the DELETE API directly via `deleteProject(id)`, then `onDelete(id)` calls the parent delete handler (`App.handleDeleteProject`), which calls the same DELETE API again. After the first request succeeds, the second can return non-ok for the already-deleted project, so the toast reports success while deleted projects remain visible until reload.

**Fix:** Use a single owner for deletion. Either:
- Make `onDelete` return `Promise<boolean>` and await it here (no direct `deleteProject` call), OR
- Expose a separate parent state-removal callback (e.g., `onRemoveFromState`) to call after this direct delete succeeds.

apps/web/src/components/DesignsTab.tsx
```diff
@@ -187,10 +238,84 @@ export function DesignsTab({
...
 			className={`tab-panel${view === "kanban" ? " design-kanban-view" : ""}`}
 		>
+			{toastMessage && (
+				<div className="toast-notification">
+					{toastMessage}
+				</div>
+			)}
+			<ConfirmModal
+				isOpen={showConfirmModal}
+				title={t("designs.bulkDelete.confirm.title", {
```
lefarcen, line 306: **P1 — i18n type breakage**

The new `designs.bulkDelete.*` keys (8 total) are passed to `t()`, whose key type is `keyof Dict`, but `apps/web/src/i18n/types.ts` does not declare these keys. This breaks the typed i18n contract and the locale alignment test for the other registered locales.

**Fix:** Add all 8 bulk-delete keys to `Dict` in `types.ts` and to every locale file (`en`, `zh-CN`, `ja`, `ko`, `pt-BR`, `ru`, etc.), using English fallback copy where translations are not available.

lefarcen (changes requested):
Hi @xxiaoxiong! The bulk delete feature looks promising — multi-select UI, confirmation modal, and toast feedback are all there. Found 2 critical issues that need fixes before this can merge:

1. **Double DELETE race**: the bulk delete path calls the API twice per project (once via `deleteProject`, then again via `onDelete`), which can leave deleted projects visible until reload.
2. **i18n type breakage**: the new `designs.bulkDelete.*` keys aren't wired into the typed `Dict` contract, breaking locale alignment.

See inline for concrete fixes. Once these are addressed, this will be ready to ship.

lefarcen:
Hey @xxiaoxiong! 👋 Quick follow-up — the review has been open for a few days now and we'd love to see this land. A couple of items are blocking merge:

1. **Double DELETE race** (critical): `handleBulkDelete` calls `deleteProject` for each selected project, but the card's `onDelete` prop fires a second time for each, resulting in double API requests. See my inline comment for a simple guard.
2. **i18n type contract** (critical): the new `designs.bulkDelete.*` keys need to be added to the typed `Dict` interface so locale alignment is enforced at compile time.
3. **Merge conflicts**: the branch is conflicting with `main` — a rebase would also clear the failing CI checks.

@PerishCode also left 4 polish items (non-blocking) to consider once you've addressed the above.

Let me know if anything in the inline comments is unclear — happy to help you get unstuck! 🙌

lefarcen:
Hey @xxiaoxiong! 👋 Checking in since it's been a few days — the bulk delete implementation is solid overall, but there are a few things blocking it from landing:

**From the review (see inline comments above):**
1. **Double-DELETE race** — the bulk path calls both `deleteProject` and then triggers `onDelete` per card, which can double-fire the delete API and leave stale cards visible until reload
2. **i18n type gap** — the new `designs.bulkDelete.*` keys aren't added to the typed `Dict` contract, which breaks locale alignment

**CI is also failing** on `build` and `Validate workspace` — the type errors from (2) are likely the root cause there, so fixing the Dict type should unblock CI too.

**Merge conflicts** need to be resolved against `main` as well.

A push when you get a chance would keep this moving — happy to re-review quickly once those are addressed!

PR #1367, opened 2026-05-12
fix: gate Finalize design package when provider config is missing

### LINKED ISSUE #1348: Finalize design package is clickable without provider config and fails with Bad request instead of being gated
## Summary

Priority: P1

## Current behavior

## Expected behavior

## Steps to reproduce

## Notes

Observed request path:

- `config.apiKey`
- `config.baseUrl`
- `config.model`

- `apiKey` empty
- `model` empty

## Why this matters

Image

Image

xxiaoxiong:
Fixes #1348

## Problem

The **Finalize design package** action is clickable even when required provider configuration (API key and model) is missing. Clicking it sends a request with empty config values and fails with:

> Bad request — check the API key and model.

This is a **P1 bug** because:
- It blocks a key end-of-flow action
- Users experience it as a broken feature
- The error message is backend-style and unhelpful
- The root cause is missing product-side gating, not a user mistake

## Root Cause

The `handleFinalize` function in `ProjectView.tsx` directly passes `config.apiKey` and `config.model` to `finalize.trigger()` without validating that they're present. When these values are empty (e.g., in Local CLI mode or when BYOK provider is not configured), the daemon returns `400 BAD_REQUEST`.

## Solution

Added **preflight validation** in `handleFinalize` before sending the request:

```tsx
if (!config.apiKey || !config.model) {
  setProjectActionsToast({
    message: 'Cannot finalize: API key and model are required.',
    details: 'Configure your provider in Settings → Execution before finalizing.',
  });
  return;
}
```

### Why This Approach?

**Handler-level validation** instead of disabling the button because:
1. **Button state is already complex**: 3 label variants based on DESIGN.md existence/staleness
2. **Discoverability**: Users need to see the action exists
3. **Actionable guidance**: Toast tells users exactly where to fix the issue (Settings → Execution)
4. **Consistent with app patterns**: Other gated actions (e.g., Orbit) use toast feedback

### Toast Message Design

- **Primary**: "Cannot finalize: API key and model are required."
  - Clear, specific, states the exact prerequisites
- **Details**: "Configure your provider in Settings → Execution before finalizing."
  - Actionable, tells users where to go and what to do

## Testing

- ✅ When `config.apiKey` is empty, clicking Finalize shows the toast and does not send a request
- ✅ When `config.model` is empty, clicking Finalize shows the toast and does not send a request
- ✅ When both are present, clicking Finalize sends the request as before
- ✅ Toast message is clear and actionable
- ✅ No backend error is surfaced
- ✅ Type checking passes

## Impact

This fix:
- **Prevents invalid requests** from being sent
- **Provides clear, actionable feedback** instead of cryptic backend errors
- **Improves UX** for users in Local CLI mode or with incomplete BYOK config
- **Follows product principle**: gate invalid actions before they fail
- **Reduces support burden**: users understand what's missing and where to fix it

## Before/After

**Before**:
1. User clicks "Finalize design package" without configuring provider
2. Request is sent with empty `apiKey` and `model`
3. Backend returns `400 BAD_REQUEST`
4. User sees: "Bad request — check the API key and model."
5. User is confused about what to check and where

**After**:
1. User clicks "Finalize design package" without configuring provider
2. Preflight validation catches missing config
3. Toast appears: "Cannot finalize: API key and model are required. Configure your provider in Settings → Execution before finalizing."
4. User knows exactly what's missing and where to fix it
5. No backend request is sent

## Files Changed

- `apps/web/src/components/ProjectView.tsx`: Added preflight validation in `handleFinalize`

## Future Enhancements

Potential improvements for future PRs:
- Disable the button when config is invalid (requires refactoring button state logic)
- Add a "Configure provider" quick action link in the toast
- Show a persistent banner when provider config is incomplete
- Gate other provider-dependent actions with similar validation

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -2141,6 +2141,18 @@ export function ProjectView({
   // shortcut wiring. Close to the JSX so the data flow is easy to
   // trace from the toolbar back to its sources.
   const handleFinalize = useCallback(() => {
+    // Gate finalize when required provider config is missing (#1348).
+    // The daemon returns BAD_REQUEST when apiKey or model is empty, but
+    // we should prevent the request from being sent in the first place
+    // and show a clear preflight message instead.
+    if (!config.apiKey || !config.model) {
+      setProjectActionsToast({
+        message: 'Cannot finalize: API key and model are required.',
+        details: 'Configure your provider in Settings → Execution before finalizing.',
+      });
+      return;
+    }
+
     void finalize.trigger({
       apiKey: config.apiKey,
       baseUrl: config.baseUrl,


apps/web/src/components/ProjectView.tsx
```diff
@@ -2141,6 +2141,18 @@ export function ProjectView({
   // shortcut wiring. Close to the JSX so the data flow is easy to
   // trace from the toolbar back to its sources.
   const handleFinalize = useCallback(() => {
+    // Gate finalize when required provider config is missing (#1348).
+    // The daemon returns BAD_REQUEST when apiKey or model is empty, but
+    // we should prevent the request from being sent in the first place
+    // and show a clear preflight message instead.
+    if (!config.apiKey || !config.model) {
```
PerishCode, line 2148: **Whitespace-only values slip past this guard but still fail at the daemon, defeating the preflight for the exact case it claims to handle.**

Problem: `!config.apiKey || !config.model` only catches empty strings (and `null`/`undefined`). A value like `' '` (single space, or pasted whitespace from a copy/paste mishap) is truthy here, so the guard passes and `finalize.trigger({ apiKey: ' ', model: ' ', ... })` is sent anyway.

Why it matters: the daemon's validator at `apps/daemon/src/import-export-routes.ts:372-376` is explicit:

```ts
if (typeof apiKey !== 'string' || !apiKey.trim()) {
  return sendApiError(res, 400, 'BAD_REQUEST', 'apiKey is required');
}
if (typeof model !== 'string' || !model.trim()) {
  return sendApiError(res, 400, 'BAD_REQUEST', 'model is required');
}
```

So whitespace-only values produce the same `BAD_REQUEST → 'Bad request — check the API key and model.'` toast (via the `useEffect` at line 2234) that the PR body says it is replacing with a preflight message. The fix is one-line and keeps the client/server contracts symmetric.

Suggested change:

```tsx
if (!config.apiKey?.trim() || !config.model?.trim()) {
  setProjectActionsToast({
    message: 'Cannot finalize: API key and model are required.',
    details: 'Configure your provider in Settings → Execution before finalizing.',
  });
  return;
}
```

Non-blocking — empty strings (the realistic Local CLI / unconfigured BYOK case the PR targets) are already covered; this only tightens the edge.


PerishCode (commented):
@xxiaoxiong The preflight gating on `handleFinalize` is a sensible product fix — short-circuiting the click before the daemon round-trip and routing users to Settings → Execution is clearly better than surfacing the raw `BAD_REQUEST`. The `useCallback` deps already include `config`, so `config.apiKey`/`config.model` access is covered, and `setProjectActionsToast` is a stable `useState` setter so omitting it is fine. One non-blocking note inline about matching the daemon's `.trim()` contract so the guard fully owns the cases it claims to handle.



apps/web/src/components/ProjectView.tsx
lefarcen, line 2148: **P2 — Test coverage gap**

The new preflight path (block finalize + toast when `apiKey`/`model` missing) is not covered by a regression test. Existing `useFinalizeProject` tests only verify the hook once `trigger()` is called; they don't prove `ProjectView` suppresses `trigger()`/fetch and renders the toast when config is missing, which is the exact #1348 bug scenario.

**Concrete fix**: Add a component test under `apps/web/tests/components/` that renders `ProjectView`, clicks "Finalize design package" with missing `apiKey` and missing `model`, asserts the toast text matches your new error message, and asserts no finalize request is sent (e.g., mock `finalize.trigger` and verify it's never called).

This ensures the guard doesn't regress in future refactors (e.g., if someone moves the finalize call around or changes the config source).

lefarcen, line 2148: **P3 — Whitespace-only config bypasses frontend guard**

The guard uses raw truthiness (`!config.apiKey || !config.model`), so whitespace-only strings (e.g., `apiKey="   "`) still pass the preflight and hit the daemon. The daemon rejects with `!apiKey.trim()` / `!model.trim()`, leaving an edge path to the same backend-style `BAD_REQUEST` error that #1348 reports.

**Concrete fix**: Trim before validating:
```tsx
if (!config.apiKey?.trim() || !config.model?.trim()) {
  setProjectActionsToast({ ... });
  return;
}
```
and pass the trimmed values to `finalize.trigger()` (or trim inside the hook if the hook already trims — worth checking).

lefarcen, line 2148: **P3 — PR body claims manual testing but no evidence**

The "Testing" section lists 5 checked outcomes (✅ empty `apiKey`, ✅ empty `model`, etc.), but the diff adds no test coverage and the body doesn't cite validation commands or manual reproduction steps.

**Concrete fix**: Either (a) add the component test above and cite the command run (`pnpm test:components ProjectView.test.tsx`), or (b) change the body to accurately state this is untested/manual-only so reviewers know the validation level. If you ran manual checks, listing the exact steps (e.g., "opened desktop app, cleared BYOK config in Settings, clicked Finalize, saw toast") makes the review stronger.

lefarcen (commented):
Hey @xxiaoxiong! 👋 Thanks for the quick fix on the Finalize gating — the preflight approach is sound and matches the app's existing toast-driven UX for blocked actions.

Found 3 non-blocking suggestions below (1 P2 test coverage + 2 P3 edge-case polish). The core validation logic is correct; these are refinements for robustness.

Siri-Ray:
Closing this as a duplicate of #1649 for #1348. #1649 covers the same Finalize design package provider-config gating issue with the broader component surface, so we are moving forward with that PR. Thanks for the work here.

PR #1368, opened 2026-05-12
fix: restore custom dropdown chevron for timezone selector in dark mode

xxiaoxiong:
Fixes #1359

## Summary

The timezone selector in the Routines form was showing repeated dropdown icons and poor text readability in dark mode.

## Root Cause

1. `.routines-field select` set `appearance: none` to remove the native chevron, but did not restore a custom one via `background-image`
2. Missing `padding-right: 32px` caused text to overlap with any chevron that did render
3. No dark-mode-specific chevron color was defined for this selector

## Changes

Added custom dropdown chevron styling for `.routines-field select` that:
- Matches the global `select` behavior (lines 258-275 in index.css)
- Includes proper `padding-right: 32px` to prevent text overlap
- Provides dark-mode color variants (`stroke=%239a9690` for dark theme)
- Supports both explicit `[data-theme=dark]` and system `prefers-color-scheme: dark`

## Result

- ✅ Single, correctly-positioned chevron icon
- ✅ Sufficient padding prevents text overlap
- ✅ Proper contrast in both light and dark themes
- ✅ Consistent visual behavior with other form controls

## Testing

- Verified the CSS changes follow the existing pattern for `select` elements
- The fix applies the same chevron SVG and positioning used globally
- Dark mode color (`%239a9690`) matches other dark-mode selects in the app

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -16770,6 +16770,27 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
   font-family: inherit;
   transition: border-color 120ms ease, box-shadow 120ms ease;
 }
+
+/* Restore custom dropdown chevron for select elements in routines form.
+   The appearance:none above removes the native chevron, so we paint our own
+   via background-image (matching the global select styling). */
+.routines-field select {
+  padding-right: 32px;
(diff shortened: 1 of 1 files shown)


lefarcen (approved):
Clean fix! 🎉

This correctly restores the custom dropdown chevron for `.routines-field select` by mirroring the global `select` pattern (lines 258-275 in index.css).

**What's good:**
- Proper `padding-right: 32px` prevents text overlap ✅
- Dark mode variants (`stroke=%239a9690`) match the global select color ✅
- Supports both `[data-theme='dark']` and `prefers-color-scheme: dark` ✅
- No logic changes, purely CSS styling — no new test needed ✅

**Minor suggestion (non-blocking):** The comment on line 16773 says "matching the global select styling" — consider making it more specific: "Mirrors the global select pattern (lines 258-275)" so future readers know exactly where to look for reference.

Ready to merge from my side. Thanks @xxiaoxiong!

PR #1369, opened 2026-05-12
fix: make Routine empty state CTA clickable and prevent button wrapping

xxiaoxiong:
Fixes #1358

## Summary

The Routines empty state had two UX issues that made the entry path feel unpolished:
1. The **New routine** button in the header wrapped awkwardly on narrow screens
2. The inline **New routine** text in the empty state message was not clickable

### 1. Made inline CTA clickable

Converted the inline "New routine" text from a styled `<em>` element to a clickable `<button>` with link-like styling:
- Appears as underlined text that integrates naturally into the prose
- Triggers the same action as the header button (opens the routine form)
- Includes proper hover states (accent color) and focus indicators
- Maintains semantic HTML with `type="button"` and proper event handling

### 2. Prevented header button wrapping

Added CSS rules to keep the header button visually stable:
- `white-space: nowrap` prevents the button text from breaking into multiple lines
- `flex-shrink: 0` ensures the button maintains its size even when space is tight
- Button remains readable and clickable at all viewport sizes

## Result

- ✅ Header button stays visually stable and does not wrap awkwardly
- ✅ Empty state message provides a directly clickable CTA
- ✅ Inline button integrates naturally into the prose (looks like emphasized text)
- ✅ Proper hover and focus states for accessibility
- ✅ Consistent interaction pattern — both CTAs open the same form
- ✅ Improved empty state discoverability and polish

## Testing

- Verified the inline CTA button triggers the form correctly
- Confirmed the header button no longer wraps on narrow screens
- Tested hover and focus states for both buttons
- Checked that the inline button styling integrates naturally into the text

### DIFF
--- a/apps/web/src/components/RoutinesSection.tsx
+++ b/apps/web/src/components/RoutinesSection.tsx
@@ -632,7 +632,20 @@ export function RoutinesSection() {
       ) : routines.length === 0 ? (
         <div className="routines-empty">
           <strong>No routines yet.</strong>
-          <p>Click <em>New routine</em> to schedule an unattended agent run.</p>
+          <p>
+            Click{' '}
+            <button
+              type="button"
+              className="routines-empty-cta"
+              onClick={() => {
+                setForm(emptyForm());
+                setShowForm(true);
+              }}
+            >
+              New routine
+            </button>{' '}
+            to schedule an unattended agent run.
+          </p>
         </div>
       ) : (
         <ul className="routines-list">
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -16722,6 +16722,13 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
 
 .routines-section .section-head { align-items: flex-start; }
 
+/* Prevent the "New routine" button from wrapping awkwardly. The button
+   should stay visually stable even when the section-head is narrow. */
+.routines-section .section-head button {
+  white-space: nowrap;
+  flex-shrink: 0;
+}
+
 .routines-card {
   background: var(--bg-panel);
   border: 1px solid var(--border);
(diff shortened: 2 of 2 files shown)


apps/web/src/components/RoutinesSection.tsx
```diff
@@ -632,7 +632,20 @@ export function RoutinesSection() {
       ) : routines.length === 0 ? (
         <div className="routines-empty">
           <strong>No routines yet.</strong>
-          <p>Click <em>New routine</em> to schedule an unattended agent run.</p>
+          <p>
+            Click{' '}
+            <button
+              type="button"
+              className="routines-empty-cta"
+              onClick={() => {
```
lefarcen, line 640: **P2 — Form reset hazard:** The inline CTA button remains visible even after the form is open (`showForm=true`). If a user starts filling out the form, then clicks the still-visible "New routine" text below it, this handler calls `setForm(emptyForm())` and silently discards their in-progress routine.

**Fix:** Either hide the empty state entirely when `showForm` is true (wrap the empty-state div in `&& !showForm`), or make this handler check `if (!showForm)` before calling `setForm(emptyForm())`. The first approach is cleaner.

lefarcen (changes requested):
Hi @xxiaoxiong! Thanks for tackling both parts of #1358 — the header button nowrap fix looks solid, and converting the inline text to a clickable button is the right approach. 👍

Found one issue that needs fixing before merge:

apps/web/src/components/RoutinesSection.tsx
```diff
@@ -632,7 +632,20 @@ export function RoutinesSection() {
...
+                setForm(emptyForm());
```
Siri-Ray, line 641: This handler resets the routine form every time the inline empty-state CTA is clicked, but the empty-state panel is still rendered whenever `routines.length === 0`, even after `setShowForm(true)` opens the form above it. That means a user can open the form, type a routine name or prompt, then click the still-visible `New routine` text and silently lose that draft via `setForm(emptyForm())`. Please either hide the empty-state CTA/panel while `showForm` is true, or make this handler avoid resetting an already-open form so the new clickable CTA cannot discard in-progress input.

Siri-Ray (changes requested):
@xxiaoxiong thanks for tightening up the Routines empty state and making the inline CTA discoverable. I found one current-head issue that can discard an in-progress routine draft, so I’m requesting a small fix before merge.


lefarcen (changes requested):
Hi @xxiaoxiong 👋 The current head still has the form-reset issue that @Siri-Ray pointed out. When the form is open (`showForm=true`), the empty-state panel below it still renders because `routines.length === 0` is still true. A user can open the form, type a routine name, then click the still-visible inline "New routine" CTA and silently lose that draft.

Please either:
1. Hide the empty-state panel when `showForm` is true, OR
2. Make the inline CTA check if a form is already open and avoid resetting it

Once this is addressed, this PR is ready. Thanks for the iteration!

Siri-Ray:
Closing this as a duplicate of #1647 for #1358. #1647 covers the same Routine empty state CTA layout and clickability issue and is the PR we are moving forward with. Thanks for the work here.

PR #1375, opened 2026-05-12
[codex] Add packaged desktop auto-update

### LINKED ISSUE #1017: Open Design 客户端未弹出自动更新提醒，导致版本未更新
## 问题描述\n我从 GitHub 下载并安装了 Open Design 客户端（自动更新版本），但在有新版本发布时，客户端没有弹出自动更新提醒，也没有自动完成版本更新。\n\n## 复现情况\n- 安装来源：GitHub Releases（自动更新版本）\n- 期望行为：检测到新版本后弹出更新提示并执行更新\n- 实际行为：没有更新提示，客户端保持旧版本\n\n## 影响\n用户可能长期停留在旧版本，无法及时获得修复和新功能。\n\n## 环境\n- 操作系统：macOS (darwin 24.6.0)\n- 时间：2026-05-09\n\n如果需要，我可以补充客户端当前版本号、日志和复现视频。

PerishCode:
Fixes #1017.

## Why

I hit this while validating the packaged beta path: release publishing already produced update-facing artifacts such as mac metadata/feed assets, but the installed packaged desktop runtime had no update service. It never checked release metadata, downloaded a newer package, or offered an installer path to the user.

The pain addressed is that beta users had to manually discover and install newer builds even though the release channel already had the metadata needed for a guided update loop.

## What users will see

Packaged macOS beta builds can now check the beta update feed, download a newer DMG, verify its checksum, persist the downloaded state, and open the verified installer from the desktop shell. The install step is still manual: users replace the app from the opened DMG.

Windows and Linux report structured unsupported update states for this phase instead of pretending an update path exists.

## Summary

- Adds an explicit desktop update IPC contract with update actions, states, modes, channels, and status snapshots.
- Implements the first packaged desktop updater phase for macOS: metadata lookup, newer-version detection, checksum-verified DMG download, persisted downloaded state, and installer open through the native desktop shell.
- Adds a namespace-scoped `updates/` runtime root beside packaged data/log/runtime/cache paths, with sentinel ownership and containment checks for cleanup.
- Wires desktop menu actions and `tools-dev` / `tools-pack` inspect actions through the same update IPC path.
- Adds `tools-serve start updater` for deterministic local updater metadata/artifact fixtures.
- Extends packaged mac smoke coverage to drive a beta-shaped updater path over real desktop IPC without requiring native prompt interaction.

## Surface area

- [x] **CLI / env var** — new `od` subcommand or flag, new `tools-dev` / `tools-pack` / `tools-pr` flag, or new `OD_*` env var
- [x] **New top-level dependency** — adding any new entry to the **root** `package.json` (`dependencies` or `devDependencies`); workspace-package `package.json` files are out of scope. Include a paragraph on what we get vs. what bytes we ship (see `CONTRIBUTING.md` -> Code style)

This adds updater/runtime dependencies needed to fetch metadata, verify downloaded artifacts, and expose deterministic local updater fixtures. They are used in the packaged desktop updater path and test fixture server rather than being shipped into generated design artifacts.

## Screenshots

Not attached in this follow-up. The discoverable entry point is the packaged desktop menu / IPC path, and the exercised acceptance path is covered by the packaged mac smoke test below.

## Bug fix verification

- Test path that reproduces the bug: `apps/desktop/tests/main/updater.test.ts`, `packages/sidecar-proto/tests/index.test.ts`, `tools/serve/tests/updater-fixture.test.ts`, and `e2e/specs/mac.spec.ts`.
- Did the test go red on `main` and green on this branch? Partially: the updater service and IPC contract did not exist on `main`, so the new updater tests are green-field coverage for #1017 rather than a one-line red spec.
- Additional verification: the packaged mac smoke drives a beta-shaped updater path over real desktop IPC, and local fixture-server smokes verify generated updater metadata/artifact paths.

## Validation

- `pnpm install`
- `pnpm --filter @open-design/sidecar-proto test`
- `pnpm --filter @open-design/desktop test`
- `pnpm --filter @open-design/packaged test`
- `pnpm --filter @open-design/tools-serve test`
- focused typechecks for sidecar-proto, desktop, packaged, tools-dev, tools-pack, tools-serve, and e2e
- `pnpm guard`
- `pnpm typecheck`
- `pnpm tools-serve start updater --json --version 99.0.0` via a short spawned CLI smoke
- `pnpm tools-serve start updater --json --channel beta --version 99.0.0-beta.1` via a short spawned CLI smoke
- `pnpm tools-pack mac build --to all --namespace release-beta --json`
- `OD_PACKAGED_E2E_MAC=1 OD_PACKAGED_E2E_NAMESPACE=release-beta pnpm --filter @open-design/e2e test specs/mac.spec.ts`
- CI follow-up: reran the failed `Validate workspace` job from run `26016935600` after confirming its failing `apps/web/tests/runtime/srcdoc-palette-css-vars.test.ts` case passes locally on this head.

## Follow-up

A full real public-release update loop still needs staged/manual verification with signed and notarized release artifacts.

### DIFF
--- a/apps/desktop/src/main/updater.ts
+++ b/apps/desktop/src/main/updater.ts
@@ -666,7 +666,7 @@ export function createDesktopUpdater(
   let progress: DesktopUpdateProgressSnapshot | undefined;
   let state: DesktopUpdateState = DESKTOP_UPDATE_STATES.IDLE;
   let error: DesktopUpdateErrorSnapshot | undefined;
-  let operation: Promise<DesktopUpdateStatusSnapshot> | null = null;
+  let operation: Promise<unknown> = Promise.resolve();
 
   function supported(): boolean {
     return config.enabled && config.mode === DESKTOP_UPDATE_MODES.PACKAGE_LAUNCHER && config.platform === "darwin";
@@ -922,11 +922,9 @@ export function createDesktopUpdater(
   }
 
   async function serialized(run: () => Promise<DesktopUpdateStatusSnapshot>): Promise<DesktopUpdateStatusSnapshot> {
-    if (operation != null) await operation.catch(() => undefined);
-    operation = run().finally(() => {
-      operation = null;
-    });
-    return await operation;
+    const next = operation.catch(() => undefined).then(run);
+    operation = next.catch(() => undefined);
+    return await next;
   }
 
   return {
--- a/apps/desktop/src/main/updater.ts
+++ b/apps/desktop/src/main/updater.ts
@@ -143,10 +143,6 @@ function normalizeMode(value: string | undefined, fallback: DesktopUpdateMode): 
   throw new Error(`unsupported desktop update mode: ${value}`);
 }
 
-function defaultChannelForVersion(version: string): DesktopUpdateChannel {
-  return /(?:^|[-.])beta(?:\.|$)/i.test(version) ? DESKTOP_UPDATE_CHANNELS.BETA : DESKTOP_UPDATE_CHANNELS.STABLE;
-}
-
 function normalizeChannel(value: string | undefined, fallback: DesktopUpdateChannel): DesktopUpdateChannel {
   if (value == null || value.length === 0) return fallback;
   if (value === DESKTOP_UPDATE_CHANNELS.STABLE || value === DESKTOP_UPDATE_CHANNELS.BETA) return value;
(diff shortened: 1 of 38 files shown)


apps/desktop/src/main/updater.ts
```diff
@@ -0,0 +1,966 @@
...
+        platformKey: nextCandidate.platformKey,
+        updateVersion: nextCandidate.version,
+        verified: true,
+        version: 1,
+      };
+      await writeJson(root.manifestPath, persisted);
+      await cleanupOwnedUpdateRoot(root.realRoot, [finalPath, root.manifestPath]);
+      const downloaded = setState(DESKTOP_UPDATE_STATES.DOWNLOADED);
+      if (config.autoOpen) return await installUpdate();
+      return downloaded;
```
lefarcen, line 854: P2 `installUpdate()` restores `downloadPath` from the JSON manifest (`restoreDownloadedState`) but never re-hashes the file on disk before calling `openPath`. The `verified: true` flag in the manifest records that the file was clean at download time, not that it's still clean now — an attacker with write access to the update root between download and a later restart could swap the DMG.

macOS Gatekeeper is a safety net, but defense-in-depth for a component opening an installer would be: re-verify `hashFile(resolvedDownload, checksum.algorithm)` against the stored `checksum.value` before calling `openPath`. The `checksum` value is available from `restoreDownloadedState` (it populates the outer `checksum` variable), so no extra network fetch is needed.

apps/desktop/src/main/updater.ts
```diff
@@ -0,0 +1,966 @@
...
+    const unsupported = unsupportedStatus();
+    if (unsupported != null) return unsupported;
+    if (downloadPath == null) {
+      const restored = await restoreDownloadedState();
+      if (restored == null || downloadPath == null) {
+        return setState(DESKTOP_UPDATE_STATES.ERROR, createError("update-not-downloaded", "no downloaded update package is available"));
+      }
+    }
+    const root = await ensureOwnedUpdateRoot(config);
+    if (!root.ok) return setState(DESKTOP_UPDATE_STATES.ERROR, root.error);
```
lefarcen, line 874: P2 `serialized()` queues one pending caller correctly but breaks for three or more. When callers B and C are both suspended at `await operation.catch(...)` waiting for A, they resume in the same microtask batch after A's `finally` sets `operation = null`. B runs to `operation = runB()` synchronously — but C immediately follows and overwrites that with `operation = runC()`, so runB and runC execute concurrently.

For current usage (menu buttons disabled while busy, single auto-check timer) the practical window is narrow. If you want the invariant to hold: a chain-promise approach queues correctly, or at minimum a comment that documents "queue depth = 1, concurrent callers beyond 2 are not serialized" would make the intent explicit.

lefarcen (changes requested):
Hey @PerishCode — the defense-in-depth posture here is exactly right for a component that ends up calling `shell.openPath` on a user-replacing installer: owned-directory sentinel, `containsPath` guards on every write and read path, atomic tmp-to-rename, namespace-scoped root via `runtimeBase`, `wx` flag on the tmp write. The IPC contract lands in `sidecar-proto` before the consumers wire against it, the e2e fixture drives a real updater path over IPC without native interaction, and `tools/serve` follows the documented tools control-plane pattern cleanly.

Two items to address before merge — see inline comments.

apps/desktop/src/main/updater.ts
PerishCode: Current head verifies this before opening: installUpdate hashes the resolved download with the stored checksum and returns checksum-mismatch if it changed, before openPath is called.

apps/desktop/src/main/updater.ts
PerishCode: Fixed in 6468e9b1: serialized now uses a promise chain so queued updater actions run one at a time. Added a regression test with three concurrent checkForUpdates calls.

lefarcen (dismissed):
Hey @PerishCode, thanks for the quick follow-up. I rechecked the two items from my previous changes-requested review on head `6468e9b`: `installUpdate()` now re-hashes the resolved download before `openPath`, and `serialized()` now chains queued operations with regression coverage for three concurrent checks. Approving to clear my stale blocker.

One non-blocking PR-body cleanup remains: please add the current-template `Surface area` checklist entry so reviewers can quickly see the packaged desktop / IPC update surface. CI is also still red on `Validate workspace`, so merge is still waiting on that plus the normal pool review.

PerishCode:
Follow-up status: I updated the PR body to the current template, including the Surface area checklist and bug-fix verification notes, and added the size/XXL, risk/high, and type/feature labels. I also reran the failed Validate workspace job from run 26016935600; the rerun is now green, including Validate workspace plus packaged mac/windows/linux smoke. The PR is still REVIEW_REQUIRED with review requests open for @mrcfps and @nettee.

nettee (approved):

PR #1385, opened 2026-05-12
feat(daemon): inject compiled design-system tokens + fixture into prompts

zoeforfun:
## Why

Follow-up to #1231. That PR landed the **structured** (machine-readable)
form of two brands — `design-systems/default/tokens.css` +
`components.html`, and the same pair for `kami` — and codified the
schema, but explicitly stopped short of changing runtime behavior:

> Daemon doesn't read these files yet — that's intentional. Prompt
> injection lands in a follow-up PR so the schema can absorb review
> feedback before downstream code starts depending on the shape.

This is that follow-up. Today the daemon's system prompt only carries
each brand's `DESIGN.md` prose — agents have to translate
`Primary: #000000` to `var(--accent)` themselves every turn, and that
translation step is where most token misuse happens (wrong slot,
inlined hex, invented names). With this change the daemon ships the
brand's `tokens.css` :root block AND a worked `components.html`
fixture alongside the prose, so agents can paste tokens verbatim and
copy fragments instead of re-deriving from English.

## What ships

Three small surface changes in the daemon, all on the prompt-assembly
path that fires for every chat run:

```
apps/daemon/src/design-systems.ts
  + readDesignSystemAssets(root, id) → { tokensCss?, fixtureHtml? }
    reads `<brand>/tokens.css` and `<brand>/components.html` if
    they exist; returns `undefined` for either missing file
    individually so brands with partial coverage still work

apps/daemon/src/server.ts
  + when an active design-system id is selected, also call
    readDesignSystemAssets() (gated by OD_DESIGN_TOKEN_CHANNEL=1),
    fall back from built-in dir → user-installed dir mirroring
    the existing readDesignSystem() chain
  + thread `designSystemTokensCss` + `designSystemFixtureHtml`
    into composeSystemPrompt({...}) alongside the existing
    designSystemBody / designSystemTitle

apps/daemon/src/prompts/system.ts
  + new optional ComposeInput fields:
      designSystemTokensCss?: string
      designSystemFixtureHtml?: string
  + each gated independently (whitespace-only counts as absent),
    appended AFTER the DESIGN.md block so prose still sets voice
    and the structured form binds names
```

Block ordering in the composed prompt is deliberate:

```
... discovery + identity charter + memory ...

## Active design system — <title>           ← DESIGN.md prose (existing)
  Treat the following DESIGN.md as authoritative for color, typography,
  spacing, and component rules ...

## Active design system tokens — <title>     ← NEW, gated
  The block below is this brand's tokens.css contract — every :root
  custom property the brand defines. Paste the unscoped :root { ... }
  block verbatim into the artifact's first <style> ...

## Reference fixture — <title>               ← NEW, gated
  A self-contained worked artifact in this design system. Match its
  component shapes ... copying fragments is encouraged as long as
  you keep the var(--*) references intact ...

## Active craft references ...               ← existing
## Active skill ...                          ← existing
```

DESIGN.md still leads because the prose carries voice, intent, and
the "feel" notes that aren't in the token grid. The compiled blocks
follow because their job is to disambiguate names + ground them in
worked HTML, not to set tone.

## Feature flag

Gated by `OD_DESIGN_TOKEN_CHANNEL=1` (env var, default off).

- **Flag off (default):** prompt is byte-equivalent to today's path.
  `readDesignSystemAssets()` is never called, no new strings appear
  in the composed prompt, no behavior change for any of the ~140
  brands — including `default` and `kami` which now ship the files.
- **Flag on:** the two new blocks are appended for any brand that
  ships either file. Today only `default` and `kami` do; the other
  ~138 brands silently skip both blocks because both files are absent.

The flag exists so PR-C can land before the smoke-test results are
in, and so a regression in artifact quality on the structured-form
path can be rolled back by clearing one env var instead of reverting
a merged PR. **Default flips to on in a follow-up** once the smoke
test (see below) confirms artifact quality improves on `default` +
`kami`. PR-B (the bulk derive script that gives the other ~138
brands their `tokens.css`) is on a separate, longer track.

## Verification

- `pnpm --filter @open-design/daemon test` → 1722 tests pass (115
  files), including the new ones below
- `pnpm --filter @open-design/daemon typecheck` → clean
- `pnpm guard` → all 6 design-system checks still green (this PR
  doesn't touch the structured-form contract from #1231; the guard
  state is identical pre/post)
- New focused tests (pin both the consumer and the producer side):
  - `apps/daemon/tests/design-system-assets.test.ts` — pins
    `readDesignSystemAssets()`'s "either-or-both files may be
    absent" semantics across 4 cases (both / tokens-only / fixture-
    only / brand-dir-empty), so future changes can't silently break
    the legacy ~138-brand fallback
  - `apps/daemon/tests/prompts/system.test.ts` (new
    `design-system token + fixture injection` describe block) —
    pins the prompt structure: both blocks appear when both inputs
    are present, prompt is byte-equivalent to baseline when both
    are absent (= flag-off), the two blocks are gated independently,
    they always land AFTER the DESIGN.md prose block, and
    whitespace-only inputs are treated as absent

## Smoke test (local, before defaulting flag on)

The structured-form contract is now consumed by the daemon, but
whether it actually moves the needle on artifact quality is an
empirical question. To answer it without a multi-week eval rig:

1. Pick one brand that ships the structured form: `default` or `kami`
2. Pick one prompt: e.g. *"做一个 hero 区块 + 主 CTA 按钮的着陆页"*
3. Run twice:
   - **Control:** `pnpm dev` (flag off) → produces `control.html`
   - **Treatment:** `OD_DESIGN_TOKEN_CHANNEL=1 pnpm dev` (same brief,
     same brand) → produces `treatment.html`
4. Eyeball the two artifacts side-by-side for these regressions that
   the structured form is supposed to fix:
   - **Color:** non-token hex values (purple gradient, random blue)
     in control but only `var(--accent)` / `var(--bg)` in treatment
   - **Type:** treatment matches the brand font; control sometimes
     falls back to system-ui or invents Inter
   - **Component shape:** treatment buttons / cards match
     `components.html` fixture; control's are generic
   - **Spacing:** treatment uses `var(--space-*)`; control hardcodes
     px values

Screenshots will be added to this PR description once the smoke
test runs. If treatment looks no better than control on either
brand, the flag stays off and the prompt copy gets revised before
flipping the default.

## Explicitly NOT in this PR

- **Default-on for `OD_DESIGN_TOKEN_CHANNEL`** — needs the smoke-test
  evidence first; lands as a one-line follow-up
- **Bulk coverage of the other ~138 brands** — needs the derive
  script (PR-B) that reads each `DESIGN.md` and emits a `tokens.css`
  draft for human review, sourcing A2 fallbacks from
  `_schema/defaults.css`
- **Misuse-rate quantification** — we agreed (in the PR-C planning
  thread) that a side-by-side eyeball comparison is sufficient signal
  for now; a formal lab can come later if smoke-test results are
  ambiguous

## Test plan

- [x] `pnpm --filter @open-design/daemon test` (1722 pass)
- [x] `pnpm --filter @open-design/daemon typecheck` (clean)
- [x] `pnpm guard` (6 design-system checks green)
- [ ] Smoke test on `default`, brief = hero + CTA, flag off vs on
- [ ] Smoke test on `kami`, brief = hero + CTA, flag off vs on
- [ ] (only if smoke tests are inconclusive) revise prompt copy and
      re-run before requesting review

### DIFF
--- a/apps/daemon/src/design-systems.ts
+++ b/apps/daemon/src/design-systems.ts
@@ -62,6 +62,42 @@ export async function readDesignSystem(root: string, id: string): Promise<string
   }
 }
 
+/**
+ * Structured (compiled) form of a brand's design system. Optional sibling
+ * files alongside DESIGN.md that, when present, give agents a
+ * machine-readable token contract and a worked fixture instead of having
+ * to re-derive both from prose. Both fields are individually optional —
+ * the daemon falls back to the DESIGN.md-only path when neither is
+ * available, which is the current state for the ~138 brands without
+ * hand-authored or derived tokens.
+ *
+ * - `tokensCss`     — verbatim content of `<brand>/tokens.css`.
+ * - `fixtureHtml`   — verbatim content of `<brand>/components.html`.
+ */
+export type DesignSystemAssets = {
+  tokensCss?: string | undefined;
+  fixtureHtml?: string | undefined;
+};
+
+export async function readDesignSystemAssets(
+  root: string,
+  id: string,
+): Promise<DesignSystemAssets> {
+  const [tokensCss, fixtureHtml] = await Promise.all([
+    readFileOptional(path.join(root, id, 'tokens.css')),
+    readFileOptional(path.join(root, id, 'components.html')),
+  ]);
+  return { tokensCss, fixtureHtml };
+}
+
+async function readFileOptional(file: string): Promise<string | undefined> {
+  try {
+    return await readFile(file, 'utf8');
+  } catch {
+    return undefined;
+  }
+}
+
 function summarize(raw: string): string {
   const lines = raw.split(/\r?\n/);
   const firstH1 = lines.findIndex((l) => /^#\s+/.test(l));
--- a/apps/daemon/src/design-systems.ts
+++ b/apps/daemon/src/design-systems.ts
@@ -93,9 +93,27 @@ export async function readDesignSystemAssets(
 async function readFileOptional(file: string): Promise<string | undefined> {
   try {
     return await readFile(file, 'utf8');
-  } catch {
-    return undefined;
+  } catch (err) {
+    // Only swallow "file genuinely does not exist" failures. Today the
+    // ~138 brands without hand-authored or derived tokens.css /
+    // components.html siblings hit this path on the empty side, which
+    // is the legacy fallback we deliberately preserve. Every other
+    // failure mode — permission denied (`EACCES`), parent-shadowed by
+    // a non-directory (`EPERM`), a directory at the file path
+    // (`EISDIR`), a broken packaged-resource symlink, a transient I/O
+    // error — means the token channel is misconfigured; the caller
+    // (and the smoke-test rollout) needs to see that explicitly
+    // instead of silently degrading to the DESIGN.md-only prompt and
+    // making the experiment look ineffective for the wrong reason.
+    if (isAbsenceError(err)) return undefined;
+    throw err;
   }
+}
+
+function isAbsenceError(err: unknown): boolean {
+  if (typeof err !== 'object' || err === null) return false;
+  const code = (err as { code?: unknown }).code;
+  return code === 'ENOENT' || code === 'ENOTDIR';
 }
 
 function summarize(raw: string): string {
--- a/apps/daemon/src/prompts/system.ts
+++ b/apps/daemon/src/prompts/system.ts
@@ -95,6 +95,22 @@ export interface ComposeInput {
     | undefined;
   designSystemBody?: string | undefined;
   designSystemTitle?: string | undefined;
+  // Compiled (machine-readable) form of the active brand's design system,
+  // shipped as sibling files to DESIGN.md when available. Both fields are
+  // optional and only injected when the daemon is running with the
+  // `OD_DESIGN_TOKEN_CHANNEL` env flag enabled (today's experimental
+  // gate). When present they are appended AFTER the DESIGN.md block so
+  // prose still sets the high-level voice and the structured form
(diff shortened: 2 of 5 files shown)


apps/daemon/src/design-systems.ts
```diff
@@ -62,6 +62,42 @@ export async function readDesignSystem(root: string, id: string): Promise<string
...
+    readFileOptional(path.join(root, id, 'tokens.css')),
+    readFileOptional(path.join(root, id, 'components.html')),
+  ]);
+  return { tokensCss, fixtureHtml };
+}
+
+async function readFileOptional(file: string): Promise<string | undefined> {
+  try {
+    return await readFile(file, 'utf8');
+  } catch {
```
nettee: This helper treats every `readFile()` failure as an absent optional asset. Absence is expected for brands without `tokens.css` or `components.html`, but other failures such as `EACCES`, `EISDIR`, an invalid packaged resource path, or a transient filesystem error mean the token channel is misconfigured. With this catch-all, `OD_DESIGN_TOKEN_CHANNEL=1` can silently drop the structured block and continue with a successful chat run, which hides the exact integration failures this new prompt path needs to surface during smoke testing and packaged rollout.

Please only return `undefined` for absence-style errors such as `ENOENT`/`ENOTDIR`, and rethrow any other error so the caller fails visibly. Add a focused test that simulates a non-absence read failure and expects `readDesignSystemAssets()` to reject.

nettee (commented):
I found one issue in the new token-channel asset reader that can hide broken packaged resources while reporting a successful prompt assembly.


lefarcen (approved):

PR #1404, opened 2026-05-12
fix(web): allow same-origin for url-load preview iframes

### LINKED ISSUE #1403: HTML artifact preview iframe blocks localStorage/cookies (SecurityError)
## Summary

When previewing an HTML artifact whose contents access `localStorage`, `sessionStorage`, or `document.cookie` (for example, a React app whose `useState` initializer reads from `localStorage`, or one that uses `i18next-browser-languagedetector`'s cookie detector), the in-tab preview iframe throws a `SecurityError` and the React tree fails to mount.

## Repro

1. Drop an artifact into a project whose `index.html` boots a React app that reads `localStorage` during initial render. The minimal in-browser-Babel + React 18 + i18next template that the repo's React preview path encourages is enough — the React DevTools tip, HMR banner, and Tauri-metadata warning in the console confirm it's hitting the URL-load branch.
2. Open the file in FileViewer's preview pane (do not turn on inspect / comment / palette / deck mode, so `shouldUrlLoadHtmlPreview()` chooses the URL-load branch — `data-od-render-mode=\"url-load\"`).
3. Observe in DevTools:

\`\`\`
Uncaught SecurityError: Failed to read the 'localStorage' property from 'Window':
  The document is sandboxed and lacks the 'allow-same-origin' flag.
    at mountState (react-dom.development.js)
    ...
Uncaught (in promise) SecurityError: Failed to read the 'cookie' property from 'Document':
  The document is sandboxed and lacks the 'allow-same-origin' flag.
    at Object.read (i18n-...js)        // i18next-browser-languagedetector
Unsafe attempt to load URL http://127.0.0.1:PORT/api/projects/<id>/raw/index.html?v=...&r=0
  from frame with URL http://127.0.0.1:PORT/api/projects/<id>/raw/index.html?v=...&r=0.
  Domains, protocols and ports must match.
\`\`\`

The React app inside the iframe never renders; the iframe shows a blank/error state.

## Root cause

\`apps/web/src/components/FileViewer.tsx\` renders the URL-load preview iframes with:

\`\`\`tsx
sandbox=\"allow-scripts\"
\`\`\`

at two call sites (the inline preview pane and the in-tab Present overlay). \`allow-scripts\` alone gives the iframe document an *opaque/unique* origin, which makes \`localStorage\` / \`sessionStorage\` / \`document.cookie\` throw \`SecurityError\` on access — exactly the failure mode in the trace above. The subsequent \"Unsafe attempt to load URL ... from frame with URL ...\" line is the same opaque-origin rule rejecting the iframe's same-URL self-loads.

The matching \`srcDoc\` branches in the same component are unaffected — they intentionally rely on the parent-injected bridge (deck / comment / inspect / palette) and a null-origin sandbox, and \`replayInspectOverridesToIframe\` already crosses that boundary via postMessage.

## Fix direction

Add \`allow-same-origin\` to the URL-load iframes only:

- \`apps/web/src/components/FileViewer.tsx:5163\` — \`data-testid=\"artifact-preview-frame\"\` URL-load branch
- \`apps/web/src/components/FileViewer.tsx:5399\` — \`title=\"present\"\` URL-load branch

This lets the previewed page reach its own \`localStorage\` / cookies under the same origin it was fetched from. The \`srcDoc\` siblings keep \`sandbox=\"allow-scripts\"\` unchanged.

### Tradeoff (caveat for reviewers)

\`/api/projects/:id/raw/*\` is served from the same origin as the host UI (\`http://127.0.0.1:<daemon-port>\`). Per the W3C/MDN sandbox guidance, \`allow-scripts\` + \`allow-same-origin\` on *same-origin* content lets the iframe escape the sandbox (mutate \`window.parent\`, remove its own \`sandbox\` attribute, etc.). For a local-only dev tool previewing the user's own project files this is acceptable — but a stricter long-term answer is to serve raw project content from a separate origin (different port or e.g. \`raw-127-0-0-1.nip.io\`) so the sandbox stays genuinely cross-origin.

Filing this as a separate observation rather than blocking the immediate fix.

## Adjacent surfaces (not in this fix's scope)

The same \`sandbox=\"allow-scripts\"\` + \`src={projectFileUrl(...)}\` pattern exists in \`apps/web/src/components/DesignFilesPanel.tsx:974\` (file-preview thumbnail). The same browser rule applies there, but it only matters for HTML thumbnails that themselves call \`localStorage\` / \`document.cookie\` during render. Leaving that out of this issue — reachable as a follow-up if reports come in.

hengfeiyang:
## Summary

- React/i18next artifacts previewed through FileViewer's URL-load branch threw `SecurityError` on every `localStorage` / `document.cookie` read because the iframe was sandboxed without `allow-same-origin`. The previewed app never mounted.
- Adds `allow-same-origin` to the two URL-load iframes in `FileViewer.tsx` (inline artifact preview + in-tab Present overlay). The `srcDoc` siblings stay on `sandbox="allow-scripts"` — they have null origin by design and use parent-injected postMessage bridges.
- See #1403 for full repro, root-cause analysis, and the same-origin sandbox-escape tradeoff that this fix accepts.

Fixes #1403

## Test plan

- [x] `pnpm --filter @open-design/web typecheck` — no new errors introduced (pre-existing `ChatMessageFeedback` / `feedback` errors on `main` are unrelated; verified by stashing the diff and re-running).
- [x] Manual: open a React + in-browser-Babel artifact whose `App()` calls `useState(() => localStorage.getItem(...))`; load via FileViewer's URL-load preview (no inspect / comment / palette / deck mode). Before: red `SecurityError` cascade in DevTools, blank iframe. After: app mounts cleanly.
- [x] Manual: same artifact, click into in-tab Present overlay — same expectation.
- [x] Manual: re-open an artifact that *needs* the srcDoc path (inspect mode, or a deck) and confirm those bridges still work — they should, because that branch is unchanged.

## Out of scope / follow-ups

- `apps/web/src/components/DesignFilesPanel.tsx:974` has the same `sandbox="allow-scripts"` + `src={projectFileUrl(...)}` pattern for HTML file thumbnails. Same browser rule applies if a thumbnail's HTML accesses `localStorage` / cookies; left out here so this PR stays scoped to the reported symptom.
- Long-term, serving `/api/projects/:id/raw/*` from a separate origin (different port or `*.nip.io`) would let us keep `allow-scripts` + `allow-same-origin` without the same-origin sandbox-escape caveat. Out of scope for this PR; mentioned in the issue.

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -5160,7 +5160,7 @@ function HtmlViewer({
                     data-testid="artifact-preview-frame"
                     data-od-render-mode="url-load"
                     title={file.name}
-                    sandbox="allow-scripts"
+                    sandbox="allow-scripts allow-same-origin"
                     src={previewSrcUrl}
                     onLoad={syncBridgeModes}
                   />
@@ -5396,7 +5396,7 @@ function HtmlViewer({
           {useUrlLoadPreview ? (
             <iframe
               title="present"
-              sandbox="allow-scripts"
+              sandbox="allow-scripts allow-same-origin"
               data-od-render-mode="url-load"
               src={previewSrcUrl}
             />


apps/web/src/components/FileViewer.tsx
```diff
@@ -5160,7 +5160,7 @@ function HtmlViewer({
                     data-testid="artifact-preview-frame"
                     data-od-render-mode="url-load"
                     title={file.name}
-                    sandbox="allow-scripts"
+                    sandbox="allow-scripts allow-same-origin"
```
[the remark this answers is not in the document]
hengfeiyang: it works for real project.

nettee, line 5163: Adding `allow-same-origin` here makes the URL-loaded artifact run with the app origin, because `previewSrcUrl` is built from `projectRawUrl(projectId, file.name)` and resolves to the same-origin `/api/projects/.../raw/...` route. Combined with `allow-scripts`, any HTML artifact script can access parent-window state and same-origin app/daemon APIs using the user's browser context; opening a crafted project preview becomes equivalent to executing untrusted artifact code inside the app origin. The same sandbox token is also added to the in-tab Present URL-load iframe at line 5399, so that path has the same exposure.

Please keep URL-loaded previews on an isolated origin before enabling `allow-same-origin` here, for example by serving raw preview content from a dedicated preview host/port/origin and pointing `previewSrcUrl` there. Until that isolation exists, keep the opaque-origin sandbox and handle storage/cookie-dependent previews through a controlled bridge or another constrained compatibility path.

nettee (commented):
I found one blocking security issue in the URL-load sandbox change.


apps/web/src/components/FileViewer.tsx
lefarcen, line 5163: P1 Adding `allow-same-origin` while the iframe still has `allow-scripts` makes URL-loaded artifacts run as the app origin, because `previewSrcUrl` resolves to `/api/projects/.../raw/...` on the same host. A crafted HTML artifact can then reach parent-window state and same-origin app/daemon APIs, which violates the preview isolation model documented in `docs/architecture.md`. Please keep this iframe opaque-origin until raw previews are served from a separate origin, or route storage-dependent cases through a constrained bridge/srcDoc compatibility path instead.

lefarcen (changes requested):
Hi @hengfeiyang! 🎉 Thanks for the careful repro and for calling out the same-origin tradeoff in the PR body — that made the security shape much easier to verify.

The blank-preview bug is real, but this particular fix changes the preview from an opaque-origin sandbox into same-origin script execution for project HTML. That crosses the repo's documented preview isolation boundary, so we need to keep this blocked until the raw preview content is isolated on a separate origin or routed through a constrained compatibility path. Nettee's review is pointing at the same blocker; I'm adding the maintainer-side review here so the PR state is clear.

Once the fix preserves iframe isolation while addressing the storage/cookie crash, happy to take another look. ❤️

hengfeiyang:
@nettee @lefarcen Thank you for your review and explanation of potential issues.

apps/web/src/components/FileViewer.tsx
lefarcen, line 5163: P1 This blocker is still present on the current head. `previewSrcUrl` is same-origin (`/api/projects/.../raw/...`), so `sandbox="allow-scripts allow-same-origin"` lets artifact scripts run with the app origin rather than an opaque preview origin. That breaks the documented preview isolation guarantee (`docs/architecture.md`: no `allow-same-origin`) and exposes parent/app/daemon state to crafted project HTML. Please keep this URL-load iframe isolated until raw previews are served from a separate origin, or handle the storage/cookie case through a constrained bridge/srcDoc compatibility path.

lefarcen (changes requested):
Hey @hengfeiyang, thanks for taking the review notes in stride. The repro and the manual verification are still very useful — the remaining concern is specifically the security boundary around untrusted artifact HTML.

I re-checked the current head and the same blocker is still present: both URL-loaded preview iframes now combine `allow-scripts` with `allow-same-origin` while loading `/api/projects/.../raw/...` from the app origin. That means we need to keep this as changes requested until the fix preserves preview isolation.

A safer direction would be either an isolated preview origin for raw project files, or routing the storage/cookie-dependent cases through a constrained compatibility path instead of relaxing the sandbox on same-origin content.

lefarcen (commented):
Thanks for working on this! I see the fix makes the artifact preview functional, but @nettee has identified a blocking security concern with the current approach.

**The core issue**: Adding `allow-same-origin` to the iframe while serving preview content from the same origin (`/api/projects/.../raw/...`) means any malicious artifact HTML/JS can access your app's cookies, localStorage, and same-origin APIs. Opening a crafted project becomes equivalent to XSS.

**Path forward** (before this can land):
1. Serve artifact previews from a separate origin (different subdomain or port), OR
2. Use a controlled postMessage bridge for artifacts that need storage/cookie access

The second approach would let you keep the isolated sandbox while supporting the React/localStorage use case through explicit, gated messages rather than unrestricted same-origin access.

Let me know if you'd like to discuss the architecture — happy to help think through the preview isolation design.

PerishCode:
@hengfeiyang — checking in: this has been waiting on your follow-up for about 4 days 1 hour. Still planning to land the same-origin preview iframe fix? If a review item is unclear, say which part needs help; if the PR is no longer active, we can close it cleanly.

hengfeiyang:
Thanks @nettee, @lefarcen, @PerishCode (and Codex) for the careful review and the patience over the past few days. After re-reading the feedback alongside `docs/architecture.md`, I agree the same-origin sandbox relaxation crosses the documented preview isolation boundary, and that the right fix is architectural — either an isolated preview origin or a controlled bridge — not the 1-line sandbox change here.

Both directions are larger than I can land cleanly as an external contributor without dragging in CORS / CSP / port allocation / e2e fixtures, so I'm going to close this PR rather than keep it sitting as `changes-requested`. The repro and root-cause analysis in #1403 stay open so the work can be picked up by a maintainer (or me, later) with the proper architecture in place.

In the meantime I'll keep the sandbox tweak in my local fork as a personal workaround only. Closing now to clear the review queue. 🙏

PR #1474, opened 2026-05-12
fix(memory): add loading state to Refresh button in Memory settings

mturac:
The Refresh button in Settings > Memory did not provide any visual feedback when clicked. Added a loading state that shows 'Refreshing...' and disables the button during the fetch.

Fixes #1418


apps/web/src/components/MemorySection.tsx
```diff
@@ -923,11 +929,12 @@ export function MemorySection() {
             type="button"
             className="ghost"
             onClick={() => void reloadExtractions()}
+            disabled={isRefreshing}
             title={t('settings.memoryExtractionsRefresh')}
           >
-            <Icon name="refresh" size={12} />{' '}
+            <Icon name={isRefreshing ? 'loader' : 'refresh'} size={12} className={isRefreshing ? 'spin' : ''} />{' '}
```
lefarcen: P1 `Icon` does not define a `loader` icon, and the existing spinner animation class is `icon-spin`, not `spin`. This will fail TypeScript on the `name` prop and would not animate even if it rendered. Please use the existing `spinner` icon with `className="icon-spin"` (or add the icon/class deliberately).

apps/web/src/components/MemorySection.tsx
```diff
@@ -923,11 +929,12 @@ export function MemorySection() {
...
             <span style={{ marginLeft: 4 }}>
-              {t('settings.memoryExtractionsRefresh')}
+              {isRefreshing ? t('settings.memoryExtractionsRefreshing') : t('settings.memoryExtractionsRefresh')}
```
lefarcen, line 937: P1 This translation key is not part of the `Dict` interface and is only added to `en.ts`, so the typed translator/locales will fail to compile for non-English dictionaries. Please add `settings.memoryExtractionsRefreshing` to `apps/web/src/i18n/types.ts` and every locale file, or reuse an existing key such as `common.loading` if that is acceptable here.

tools/pack/bin/tools-pack.mjs
```diff
@@ -1,16 +1,55 @@
 #!/usr/bin/env node
 
-import { existsSync } from "node:fs";
+import { existsSync, readdirSync, statSync } from "node:fs";
```
lefarcen, line 3: P2 This tools-pack stale-dist check is unrelated to the Memory refresh UI fix and changes CLI behavior by walking `src/` and printing warnings during dev runs. Please remove it from this PR or split it into a separate, focused change so the bugfix stays scoped and easier to validate.

lefarcen (changes requested):
Hi @mturac! Thanks for picking this up — the disabled state plus explicit "Refreshing…" label is the right UX direction for #1418. I found a couple of build-time issues in the implementation, plus one unrelated file change that should be split out before this lands. Details inline.

apps/web/src/i18n/locales/en.ts
```diff
@@ -1432,6 +1432,8 @@ export const en: Dict = {
   'settings.memoryExtractionsHint': 'Recent extraction attempts. Heuristic regex extraction runs first; LLM extraction runs in the background after each turn.',
   'settings.memoryExtractionsEmpty': 'No extractions yet. The next chat turn will populate this list.',
   'settings.memoryExtractionsRefresh': 'Refresh',
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
```
lefarcen: P1 This duplicate object key is the current CI failure (`TS1117: An object literal cannot have multiple properties with the same name`). Please remove the second `settings.memoryExtractionsRefreshing` entry so the web typecheck/build can pass.

apps/web/src/i18n/locales/zh-CN.ts
```diff
@@ -1396,7 +1396,8 @@ export const zhCN: Dict = {
   'settings.memoryExtractions': '抽取历史',
   'settings.memoryExtractionsHint': '最近的 LLM 抽取记录。每次对话结束后，启发式正则会先跑，LLM 抽取在后台异步进行。',
   'settings.memoryExtractionsEmpty': '暂无抽取记录。下一次对话结束后会出现在这里。',
-  'settings.memoryExtractionsRefresh': '刷新',
+  'settings.memoryExtractionsRefresh': 'Refresh',
```
lefarcen, line 1399: P2 This regresses an existing Simplified Chinese translation from `刷新` back to English, and zh-TW has the same issue with `重新整理`. Please keep the existing localized Refresh labels and add localized `Refreshing…` strings for those locales instead of replacing them with English.

tools/pack/bin/tools-pack.mjs
lefarcen, line 3: P2 This unrelated tools-pack stale-dist warning is still bundled with the Memory refresh UI fix. Please remove it from this PR or split it into a separate change so this bugfix remains scoped and can be validated independently.

lefarcen (changes requested):
Hey @mturac, thanks for the quick update — the spinner class/type issue is largely addressed, and the new typed i18n key is moving in the right direction. A few blockers remain on this head, including the current CI failure; see inline for the focused fixes.

apps/web/src/i18n/locales/zh-CN.ts
lefarcen, line 1399: P2 This still regresses the existing Simplified Chinese label from `刷新` to English, and the zh-TW file has the same regression from `重新整理` to `Refresh`. Please keep the existing localized Refresh strings and only add localized `settings.memoryExtractionsRefreshing` values for those locales.

tools/pack/bin/tools-pack.mjs
lefarcen, line 3: P2 This tools-pack stale-dist check is still unrelated to the Memory refresh loading-state fix. Please remove it from this PR or split it into a separate PR so this bugfix remains narrowly scoped.

lefarcen (changes requested):
Hey @mturac, nice cleanup on the duplicate English key — the main build/typecheck failures from the last pass are resolved on this head. Two scoped issues still need fixing before we can clear the previous block: one locale regression and the unrelated tools-pack change that is still in the PR.

apps/web/src/i18n/locales/zh-CN.ts
```diff
@@ -1396,7 +1396,8 @@ export const zhCN: Dict = {
...
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
```
PerishCode, line 1400: **Restore the Simplified Chinese `Refresh` translation and localize the new `Refreshing…` key**

This PR changed `'settings.memoryExtractionsRefresh': '刷新'` (the pre-existing localized label) to the English string `'Refresh'`, and the new `'settings.memoryExtractionsRefreshing': 'Refreshing…'` is also English. That regresses an already-translated Simplified Chinese label: zh-CN users who previously saw `刷新` will now see `Refresh`, and the new loading state will also appear in English.

This is the same regression flagged by @lefarcen on review 4274901025 and again on 4275022205 against this head — adding a new sibling key does not require replacing the existing localized value with English.

Why it matters: the dictionary type forces every locale to provide every key, so the cheap path is to ship English. But the existing `刷新` translation is a real value that users have been seeing, and silently dropping it is a UX regression for the Simplified Chinese audience, not a missing-translation gap.

Suggested fix in `apps/web/src/i18n/locales/zh-CN.ts`:

```ts
'settings.memoryExtractionsRefresh': '刷新',
'settings.memoryExtractionsRefreshing': '刷新中…',
```

Keep the original `刷新` and add a Simplified Chinese translation for the loading state instead of replacing the localized string with English when adding the adjacent key.


apps/web/src/i18n/locales/zh-TW.ts
```diff
@@ -1350,7 +1350,8 @@ export const zhTW: Dict = {
   'settings.memoryExtractions': '抽取歷史',
   'settings.memoryExtractionsHint': '最近的 LLM 抽取記錄。每次對話結束後，啟發式正則會先跑，LLM 抽取在背景非同步進行。',
   'settings.memoryExtractionsEmpty': '暫無抽取記錄。下一次對話結束後會出現在這裡。',
-  'settings.memoryExtractionsRefresh': '重新整理',
+  'settings.memoryExtractionsRefresh': 'Refresh',
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
```
PerishCode, line 1354: **Restore the Traditional Chinese `Refresh` translation and localize the new `Refreshing…` key**

Same regression as the zh-CN file: the previous `'settings.memoryExtractionsRefresh': '重新整理'` value was replaced with the English `'Refresh'`, and the new `'settings.memoryExtractionsRefreshing'` key also ships in English here. zh-TW users who already see `重新整理` for this button will start seeing `Refresh` after this PR lands, and the new loading state will not be localized at all.

Why it matters: replacing an existing localized value while adding a sibling key is not equivalent to “adding a new translation”; it is a regression of a translation that is already in production.

Suggested fix in `apps/web/src/i18n/locales/zh-TW.ts`:

```ts
'settings.memoryExtractionsRefresh': '重新整理',
'settings.memoryExtractionsRefreshing': '重新整理中…',
```

Keep the original `重新整理` and add a Traditional Chinese translation for the loading state. The other locale files (`ar`, `de`, `es-ES`, `fa`, `fr`, `hu`, `id`, `ja`, `ko`, `pl`, `pt-BR`, `ru`, `tr`, `uk`) already ship English strings for the adjacent `memoryExtractionsHint` / `memoryExtractionsEmpty` keys, so adding the new key in English there matches their existing state — but the two Chinese dictionaries are the only ones with a pre-existing translation for this label and should not be downgraded.


tools/pack/bin/tools-pack.mjs
PerishCode, line 3: **Split the tools-pack stale-dist warning out of this PR**

This file is unrelated to the Memory refresh button fix that the PR title, description, and `Fixes #1418` scope to. Commits 054392b6 (`fix(tools-pack): warn on stale dist in dev/workspace mode`) and 8582526c (`fix(tools-pack): remove TypeScript annotation and use recursive source mtime check`) on this branch add a recursive `readdirSync` walk of `tools/pack/src` on every dev-mode `tools-pack` invocation and emit a `console.warn` when source mtimes exceed the dist mtime. That is a separate behavior change (`Fixes #1452` per the commit messages) and should land independently so it can be reviewed and validated on its own.

Why it matters: the AGENTS.md bug-follow-up playbook explicitly says to hold the spec's scope — defects discovered outside the bug's described boundary belong in a follow-up PR, not bundled with the fix. Bundling also makes the regression risk for the `tools-pack` dev-loop harder to track in #1418's history and contradicts the size/risk labels currently assigned to this PR.

The reviewer asked for this split on review 4274253259 (against `fea2699d`) and again on 4275022205 (against this head `cb598c95`). The change has not been backed out yet.

Suggested fix: drop `tools/pack/bin/tools-pack.mjs` from this branch (e.g., revert commits 054392b6 and 8582526c with `git revert`, or `git restore --source=main tools/pack/bin/tools-pack.mjs` on this branch) and open a separate PR carrying just the tools-pack change.


apps/web/src/index.css
```diff
@@ -17937,7 +17937,9 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
   top: 0;
   background: var(--bg-panel);
   padding: 8px 10px 4px;
+  min-height: 32px;
   z-index: 1;
+  border-bottom: 1px solid var(--border);
 }
```
PerishCode, line 17943: **Split the picker provider-group header CSS out of this PR**

These `.media-model-picker-group-header` style additions (`min-height: 32px;` and `border-bottom: 1px solid var(--border);`) come from commit 4d655485 (`fix(picker): improve provider group header separation in Media model picker`, `Fixes #1434`) and are unrelated to the Memory refresh loading-state work this PR is scoped to.

Why it matters: the same scope-hygiene rationale that applies to the bundled `tools-pack` change applies here. The picker change touches a globally-shared stylesheet (`apps/web/src/index.css`) and is an entirely separate visual fix; landing it inside a memory-settings bugfix makes both PRs harder to revert independently if the picker layout regresses in some viewport, and it muddies the change history for #1418.

Suggested fix: move this two-line change to its own PR scoped to #1434. On this branch, the simplest path is to revert commit 4d655485 and open a focused follow-up.


apps/web/src/components/MemorySection.tsx
```diff
@@ -923,11 +929,12 @@ export function MemorySection() {
             type="button"
             className="ghost"
             onClick={() => void reloadExtractions()}
+            disabled={isRefreshing}
             title={t('settings.memoryExtractionsRefresh')}
           >
-            <Icon name="refresh" size={12} />{' '}
+            <Icon name={isRefreshing ? 'refresh' : 'refresh'} size={12} className={isRefreshing ? 'icon-spin' : ''} />{' '}
```
PerishCode, line 935: **Nit: collapse the degenerate ternary on the icon `name`**

`name={isRefreshing ? 'refresh' : 'refresh'}` evaluates to the same string in both branches, so the ternary is dead. It looks like a leftover from the previous `'loader' : 'refresh'` attempt that was reverted after the `loader` icon was found to be missing from the icon set. The spinning animation is already driven by the conditional `icon-spin` class on the same element, so the ternary on `name` is purely noise.

Why it matters: future readers (and any linter that flags `no-constant-condition` / `no-useless-ternary`) will have to puzzle out why one branch exists. Non-blocking, but cheap to tidy.

Suggested fix on this line:

```tsx
<Icon name="refresh" size={12} className={isRefreshing ? 'icon-spin' : ''} />{' '}
```

(If the intent was instead to fall back to the existing `spinner` icon the prior reviewer suggested, this would become `name={isRefreshing ? 'spinner' : 'refresh'}` — but that is a behavior choice, not a required change.)


PerishCode (commented):
@mturac Thanks for keeping at this. The Memory refresh loading affordance itself is working on this head — the button disables, the icon spins, and the label flips to `Refreshing…` — but two of the previous-pass blockers are still present on `cb598c95`, and a third unrelated-scope change is worth grouping with them:

- `zh-CN.ts` and `zh-TW.ts` lost their existing localized `Refresh` labels (`刷新` / `重新整理`), and the new `Refreshing…` key ships in English for both dictionaries.
- `tools/pack/bin/tools-pack.mjs` still carries an unrelated stale-dist warning (#1452 work, two commits on this branch).
- `apps/web/src/index.css` also still carries an unrelated picker provider-group header change (4d655485, #1434).

Inline comments cover the specifics and exact suggested edits.




lefarcen:
Hi @mturac — thanks again for iterating on this. Since the focused replacement PR #1477 has now been merged and #1418 is closed, I'm closing this older branch as superseded so the review tracks don't diverge further. Appreciate the work you put into the refresh loading-state fix.

PR #1477, opened 2026-05-12
fix(memory): add loading state to Refresh button in Memory settings

mturac:
The Refresh button in Settings > Memory did not provide any visual feedback when clicked. Added a loading state that shows 'Refreshing...' and disables the button during the fetch.

Fixes #1418


apps/web/src/components/MemorySection.tsx
```diff
@@ -916,9 +916,10 @@ export function MemorySection() {
             type="button"
             className="ghost"
             onClick={() => void reloadExtractions()}
+            disabled={isRefreshing}
```
lefarcen, line 925: P1 `isRefreshing` is referenced here but is not declared anywhere in `MemorySection`, so the web build/typecheck fails before this UI can render. Please add explicit refresh state around `reloadExtractions` (set true before the fetch and reset in a `finally`) or reuse an existing loading signal before wiring `disabled`.

apps/web/src/components/MemorySection.tsx
```diff
@@ -916,9 +916,10 @@ export function MemorySection() {
...
             title={t('settings.memoryExtractionsRefresh')}
           >
-            <Icon name="refresh" size={12} />{' '}
+            <Icon name="refresh" size={12} className={isRefreshing ? 'icon-spin' : ''} />{' '}
             <span style={{ marginLeft: 4 }}>
               {t('settings.memoryExtractionsRefresh')}
```
lefarcen: P2 The PR adds `settings.memoryExtractionsRefreshing`, but the button still always renders the normal Refresh label. That leaves the main visible feedback from #1418 missing for non-screen-reader users; please switch the label (and ideally title/aria state) to the refreshing string while the fetch is in flight.

apps/web/src/i18n/locales/zh-CN.ts
```diff
@@ -1441,7 +1441,8 @@ export const zhCN: Dict = {
   'settings.memoryExtractions': '抽取历史',
   'settings.memoryExtractionsHint': '最近的 LLM 抽取记录。每次对话结束后，启发式正则会先跑，LLM 抽取在后台异步进行。',
   'settings.memoryExtractionsEmpty': '暂无抽取记录。下一次对话结束后会出现在这里。',
-  'settings.memoryExtractionsRefresh': '刷新',
+  'settings.memoryExtractionsRefresh': 'Refresh',
```
lefarcen: P2 This regresses the existing Simplified Chinese label from `刷新` back to English, and the new refreshing text is also English. Please keep the existing localized Refresh text and add a localized `Refreshing…` value for zh-CN.

apps/web/src/i18n/locales/zh-TW.ts
```diff
@@ -1350,7 +1350,8 @@ export const zhTW: Dict = {
   'settings.memoryExtractions': '抽取歷史',
   'settings.memoryExtractionsHint': '最近的 LLM 抽取記錄。每次對話結束後，啟發式正則會先跑，LLM 抽取在背景非同步進行。',
   'settings.memoryExtractionsEmpty': '暫無抽取記錄。下一次對話結束後會出現在這裡。',
-  'settings.memoryExtractionsRefresh': '重新整理',
+  'settings.memoryExtractionsRefresh': 'Refresh',
```
lefarcen: P2 Same localization regression here: Traditional Chinese previously used `重新整理`, but this changes the visible button back to English. Please restore the localized refresh label and add a localized refreshing label for zh-TW.

lefarcen (changes requested):
Hey @mturac, thanks for continuing to tighten the Memory refresh UX — narrowing this follow-up to the button feedback and removing the unrelated files from #1474 is a good direction. A couple of blockers remain before this can land: the new loading prop currently does not compile, and the visible/loading label plus localized strings need to line up with the issue’s feedback goal. Details inline.

apps/web/src/components/MemorySection.tsx
```diff
@@ -275,6 +275,7 @@ export function MemorySection() {
   const [previewBody, setPreviewBody] = useState<string | null>(null);
   const [editing, setEditing] = useState<DraftEntry | null>(null);
   const [busy, setBusy] = useState(false);
+  const [isRefreshing, setIsRefreshing] = useState(false);
```
lefarcen, line 278: P1 `isRefreshing` is now declared, but `setIsRefreshing` is never called, so this state remains `false` forever. The button will not disable, spin, or show `Refreshing…` during `fetchExtractions()`, which means #1418 is still observable. Please wrap `reloadExtractions` with `setIsRefreshing(true)` before the fetch and reset it in a `finally` block after the fetch settles.

lefarcen (changes requested):
Hey @mturac, thanks for the quick follow-up — the localized zh-CN/zh-TW labels are fixed now, and the button text is wired to the new refreshing key. One core blocker is still left: the new state is declared but never toggled, so the loading UI still never appears in practice. See the remaining inline note.

lefarcen (approved):
Hey @mturac, thanks for the quick fix — the refresh state is now toggled around the extraction fetch, the button label switches to the new refreshing string, and the zh-CN/zh-TW regressions are corrected. This clears my previous blocker from the bot review side. CI is still finishing and the maintainer/pool-review path remains separate, but the specific issues I raised are addressed.

mturac:
Thanks for merging! 🙏

PR #1478, opened 2026-05-12
fix: add success feedback for exports and downloads

xxiaoxiong:
Fixes #1236

## Summary

After exporting or downloading files (PDF, ZIP, HTML, Markdown), the app provided no visible success feedback, leaving users uncertain whether the action completed successfully. This is a **P1** bug because exports are high-intent actions and users need clear confirmation.

## Problem

Actions such as:
- Export as PDF
- Download as .zip
- Export as standalone HTML
- Export as Markdown

completed without any confirmation or completion feedback. Users could not easily tell:
- Whether the file was generated successfully
- Whether the download started
- Where the output went
- Whether they should retry

### 1. Added Toast feedback system

- Imported `Toast` component into `FileViewer.tsx`
- Added `exportToast` state to `LiveArtifactViewer` component
- Rendered Toast at the bottom of the viewer when `exportToast` is set
- Toast auto-dismisses after 4 seconds

### 2. Enhanced async export handlers

- Wrapped `exportProjectAsPdf` and `exportProjectAsZip` in try-catch blocks
- Added success toast after successful completion
- Added error toast when exports fail
- Logged errors to console for debugging

### 3. Added success feedback for sync exports

- Added success toast after `exportAsHtml` completes
- Added success toast after `exportAsMd` completes
- These are synchronous operations so no error handling needed

### 4. Added i18n support

Added translation keys in both English and Chinese:
- `fileViewer.exportPdfSuccess` / `PDF exported successfully` / `PDF 导出成功`
- `fileViewer.exportPdfError` / `Failed to export PDF` / `PDF 导出失败`
- `fileViewer.exportZipSuccess` / `ZIP downloaded successfully` / `ZIP 下载成功`
- `fileViewer.exportZipError` / `Failed to download ZIP` / `ZIP 下载失败`
- `fileViewer.exportHtmlSuccess` / `HTML exported successfully` / `HTML 导出成功`
- `fileViewer.exportMdSuccess` / `Markdown exported successfully` / `Markdown 导出成功`

## Result

- ✅ Users see **PDF exported successfully** after PDF export
- ✅ Users see **ZIP downloaded successfully** after ZIP download
- ✅ Users see **HTML exported successfully** after HTML export
- ✅ Users see **Markdown exported successfully** after Markdown export
- ✅ Error messages shown when async exports fail (with console logging)
- ✅ Toast auto-dismisses after 4 seconds
- ✅ Full i18n support (English + Chinese)
- ✅ Consistent feedback pattern across all export formats

## Testing

- Verified Toast component is properly imported and rendered
- Confirmed async exports are wrapped in try-catch
- Checked that success messages are set after each export action
- Verified i18n keys are added to both en.ts and zh-CN.ts
- Ensured Toast positioning does not interfere with other UI elements

## Why This Matters

Exports are high-intent actions. Without clear feedback:
- Users may retry unnecessarily (wasting resources)
- Users may think the export failed (poor UX)
- Users may be unsure whether the file was actually saved (confusion)

This P1 fix significantly improves the export UX by providing explicit confirmation that actions completed successfully.


apps/web/src/components/FileViewer.tsx
```diff
@@ -4623,15 +4625,25 @@ function HtmlViewer({
...
+                        await exportProjectAsPdf({
+                          deck: effectiveDeck,
+                          fallbackPdf: () => exportAsPdf(source ?? '', exportTitle, { deck: effectiveDeck }),
+                          filePath: file.name,
+                          projectId,
+                          title: exportTitle,
+                        });
+                        setExportToast(
+                          effectiveDeck
+                            ? t('fileViewer.exportPdfSuccess')
```
lefarcen, line 4640: P1 These new translation keys are used through `t(...)`, whose argument is typed as `keyof Dict`, but `apps/web/src/i18n/types.ts` was not updated with the six new `fileViewer.export*Success/Error` keys. That will make the web typecheck fail. Please add the new keys to `Dict`; the other locale files already spread `en`, so they can inherit the English fallback unless you want to localize them now.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4623,15 +4625,25 @@ function HtmlViewer({
...
-                      });
+                      try {
+                        await exportProjectAsPdf({
+                          deck: effectiveDeck,
+                          fallbackPdf: () => exportAsPdf(source ?? '', exportTitle, { deck: effectiveDeck }),
+                          filePath: file.name,
+                          projectId,
+                          title: exportTitle,
+                        });
+                        setExportToast(
```
lefarcen, line 4638: P1 `exportProjectAsPdf` currently catches its own failures and returns `'fallback'`, while the browser fallback also resolves after popup-blocked/native-print failure paths after showing an alert. Because of that, this success toast can still appear when no PDF export actually started, and the surrounding `catch` will almost never show `exportPdfError`. Please make the export helper report a real failure/result (or throw when the fallback cannot start) and only show the success toast for confirmed started/completed paths.

lefarcen (changes requested):
Hi @xxiaoxiong! Thanks for jumping on the export-feedback gap. The direction is right: putting visible feedback next to the Share menu actions is exactly the UX this issue needs.

I found two blockers before this can land — one will break the web typecheck, and the other means the new PDF error/success feedback can be inaccurate on real failure paths. Details inline. Once those are tightened up, this should be a small, focused fix. 🙌

PerishCode:
@xxiaoxiong I'm holding off on generating review comments for nexu-io/open-design#1478 because this pull request has merge conflicts right now.

lefarcen (changes requested):
Hey @xxiaoxiong, quick follow-up on the current head: the two blockers from my last review still apply, and @PerishCode also flagged that the branch currently has merge conflicts with `main`.

Please resolve the conflicts and update the export feedback implementation so the i18n typing is complete and the PDF success/error toast reflects the real export outcome. Once the branch is updated, we can re-review the new head cleanly.

PerishCode (commented):
@xxiaoxiong thanks for tackling the export-feedback gap — putting a visible Toast next to the Share menu actions is the right UX move. On this head, though, three blockers still stand and the change as written will not compile cleanly or convey accurate PDF export status. I've broken each one out inline with a concrete fix shape so they can be addressed together; the existing unresolved threads from @lefarcen and the Codex bot point at the same root causes.



apps/web/src/components/FileViewer.tsx
```diff
@@ -397,6 +398,7 @@ export function LiveArtifactViewer({
   const [refreshHistory, setRefreshHistory] = useState<LiveArtifactRefreshLogEntry[]>([]);
   const [presentMenuOpen, setPresentMenuOpen] = useState(false);
   const [inTabPresent, setInTabPresent] = useState(false);
+  const [exportToast, setExportToast] = useState<string | null>(null);
```
PerishCode, line 401: **Blocker — `exportToast` state is declared in the wrong component, so every consumer is unresolved.**

This `useState` lives in `LiveArtifactViewer` (function starting at line 376), but every reader/writer is inside `HtmlViewer` (function starting at line 3001): `setExportToast(...)` calls at lines 4638, 4645, 4690, 4693, 4707, 4726 and the `<Toast message={exportToast} … />` render at line 5334. `LiveArtifactViewer` itself never references `exportToast`, so on the current head the `HtmlViewer` references are undefined identifiers and the web typecheck/build will fail.

Why it matters: this is a compile-time blocker — `pnpm --filter @open-design/web typecheck` won't pass, which means the feature can't ship and CI won't go green.

Suggested change: move the state into `HtmlViewer` instead. Drop the declaration from line 401 of `LiveArtifactViewer` and add it near the other UI-feedback state in `HtmlViewer` (e.g. next to `shareMenuOpen` around line 3034):

```tsx
const [exportToast, setExportToast] = useState<string | null>(null);
```

That keeps the toast scoped to the component that actually owns the share-menu actions and the `<Toast>` render. Please also run `pnpm --filter @open-design/web typecheck` locally before re-requesting review — the AGENTS.md validation strategy expects package-scoped typecheck after touching web sources, and it would have caught this immediately.


apps/web/src/i18n/locales/en.ts
```diff
@@ -879,6 +879,12 @@ export const en: Dict = {
   'fileViewer.exportZip': 'Download as .zip',
   'fileViewer.exportHtml': 'Export as standalone HTML',
   'fileViewer.exportMd': 'Export as Markdown',
+  'fileViewer.exportPdfSuccess': 'PDF exported successfully',
+  'fileViewer.exportPdfError': 'Failed to export PDF',
+  'fileViewer.exportZipSuccess': 'ZIP downloaded successfully',
+  'fileViewer.exportZipError': 'Failed to download ZIP',
+  'fileViewer.exportHtmlSuccess': 'HTML exported successfully',
+  'fileViewer.exportMdSuccess': 'Markdown exported successfully',
```
PerishCode, line 887: **Blocker — six new translation keys added without extending the `Dict` interface.**

`useT()` returns a typed `t(key: keyof Dict) => string`, and `apps/web/src/i18n/types.ts` is the source of truth for `Dict` (other `fileViewer.export*` keys are listed there at lines 1131-1141). The six new keys here — `fileViewer.exportPdfSuccess`, `exportPdfError`, `exportZipSuccess`, `exportZipError`, `exportHtmlSuccess`, `exportMdSuccess` — are referenced from `FileViewer.tsx` via `t('fileViewer.exportPdfSuccess')` etc., but are missing from `Dict`. The web typecheck will fail at every call site with `Argument of type '"fileViewer.exportPdfSuccess"' is not assignable to parameter of type 'keyof Dict'`.

Why it matters: same blocker class as above — typecheck/build fails, and the i18n contract (string-only typed keys) is broken. Other locales spread `en` so they'll inherit the English fallback once the type is fixed; you don't need to localize all 20+ locale files right now.

Suggested change: add the six entries to `Dict` in `apps/web/src/i18n/types.ts`, e.g. alongside the existing `fileViewer.exportZip`/`exportHtml`/`exportMd` declarations around lines 1136-1138:

```ts
'fileViewer.exportPdfSuccess': string;
'fileViewer.exportPdfError': string;
'fileViewer.exportZipSuccess': string;
'fileViewer.exportZipError': string;
'fileViewer.exportHtmlSuccess': string;
'fileViewer.exportMdSuccess': string;
```


apps/web/src/components/FileViewer.tsx
```diff
@@ -4623,15 +4625,25 @@ function HtmlViewer({
...
+                          title: exportTitle,
+                        });
+                        setExportToast(
+                          effectiveDeck
+                            ? t('fileViewer.exportPdfSuccess')
+                            : t('fileViewer.exportPdfSuccess')
+                        );
+                      } catch (err) {
+                        console.error('[exportProjectAsPdf] failed:', err);
+                        setExportToast(t('fileViewer.exportPdfError'));
```
PerishCode, line 4645: **Blocker — PDF success toast misreports the real export outcome, and the success-branch ternary is dead code.**

Two problems share these lines:

1. **Inaccurate success/error feedback.** `exportProjectAsPdf` (`apps/web/src/runtime/exports.ts:71-97`) catches its own failures internally and resolves to `'fallback'`, invoking `opts.fallbackPdf()` (the browser-print path). It only rejects on a programmer-error throw, never on a real desktop-export failure or a fallback that the user cancels / has popup-blocked. As wired here, the `try` always succeeds when the helper returns `'fallback'`, so `setExportToast(t('fileViewer.exportPdfSuccess'))` fires even when the desktop endpoint errored — and the `catch` at line 4643 essentially never runs, making `exportPdfError` cosmetic. This contradicts the PR's stated goal ("Error messages shown when async exports fail") and reintroduces the very ambiguity #1236 is trying to remove.
2. **Identical ternary branches.** Lines 4639-4641 read `effectiveDeck ? t('fileViewer.exportPdfSuccess') : t('fileViewer.exportPdfSuccess')` — both arms resolve to the same key, so the ternary is dead. Either it's a leftover from an intended deck-vs-single-slide message split (in which case a second key is missing), or it should just be a flat `setExportToast(t('fileViewer.exportPdfSuccess'))`.

Why it matters: the visible success affordance is the entire deliverable of this PR; if it lies on the fallback / popup-blocked path, users still won't know whether the export actually happened, and the toast becomes worse than no feedback because it manufactures false confidence. The dead ternary also confuses future readers about whether deck vs. non-deck PDFs are expected to differ.

Suggested change: have the helper report a truthful outcome and gate the toast on it.

- In `apps/web/src/runtime/exports.ts`, surface real failures — for example return `'desktop' | 'fallback' | 'failed'` and only return `'fallback'` once `opts.fallbackPdf()` has been initiated without throwing; have `exportAsPdf` itself throw (instead of `alert(...)`) when the popup is blocked or `window.print()` is unavailable, so this caller can distinguish them.
- At this call site, branch on the returned value:

```tsx
const result = await exportProjectAsPdf({ /* … */ });
if (result === 'failed') {
  setExportToast(t('fileViewer.exportPdfError'));
} else {
  setExportToast(t('fileViewer.exportPdfSuccess'));
}
```

- Collapse the identical ternary to a single `t('fileViewer.exportPdfSuccess')` call, or introduce a deliberate deck-specific success key (e.g. `fileViewer.exportPdfDeckSuccess`) and add it to `Dict` plus the locale files, whichever you actually want.


lefarcen (changes requested):
Hey @xxiaoxiong, @PerishCode's latest review covers the current head well, including one extra compile-time blocker I had not called out explicitly: `exportToast` is declared in `LiveArtifactViewer`, but all of its reads/writes are inside `HtmlViewer`.

So the path forward is: resolve the merge conflicts, move the toast state into `HtmlViewer`, add the missing `Dict` entries for the new i18n keys, and make the PDF toast depend on a truthful export result instead of always reporting success. After those updates land on a new head, we can do a clean re-review.

lefarcen:
Heads-up: PR #1627 is also open against this area — both PRs touch `apps/web/src/components/FileViewer.tsx` and add success toast feedback for the same export/download actions (PDF, ZIP, HTML, Markdown), both fixing #1236.

Since you opened both PRs, you might want to consolidate the approaches — #1627 has full i18n coverage (18 locales vs 2 here), but this PR was opened first. The maintainer team will pick which lands, but sharing so you can coordinate the effort if you prefer to focus on one.

xxiaoxiong:
Closing this PR in favor of #1627 which has complete i18n coverage for all 18 locales (this one only has 2). #1627 is the more complete implementation.

PR #1476, opened 2026-05-12
fix: preserve Claude tool inputs

prantikmedhi:
## Summary

Claude Code was dropping streamed tool input JSON, so tool calls like Write and Bash could arrive as {} and fail validation. This patch reads the accumulated JSON when a tool block stops, then skips the duplicate empty tool_use that comes later in the final assistant wrapper.

## Changes

- Parse streamed tool input at content_block_stop
- Ignore the duplicate assistant-wrapper tool_use when streamed input was already emitted
- Add a regression test for streamed Write payloads

## Testing

- pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/structured-streams.test.ts
- pnpm --dir apps/daemon typecheck

Fixes nexu-io/open-design#1475

PR #1479, opened 2026-05-12
fix: parse accumulated input_json_delta on content_block_stop

### LINKED ISSUE #1475: Claude Code stream parser drops input_json_delta tool inputs, causing empty Write/Bash calls
## Bug

When running Claude Code through Open Design, tool calls sometimes arrive with empty inputs, causing repeated failures:

- `Write input {}` → missing `file_path` and `content`
- `Bash input {}` → missing `command`

Example UI/tool error:

```text
InputValidationError: Write failed due to the following issues:
The required parameter `file_path` is missing
The required parameter `content` is missing
```

```text
InputValidationError: Bash failed due to the following issue:
The required parameter `command` is missing
```

## Reproduction

1. Start local dev with:

```bash
pnpm tools-dev run web --daemon-port 7456 --web-port 3000
```

2. Open an existing imported Claude Design project.
3. Ask Claude Code / `claude-sonnet-4-6` to create or update an HTML file.
4. The run starts normally: `TodoWrite` and `Read` tool calls have valid inputs.
5. When it tries to write a larger file, Open Design shows `Write (unnamed)` and the tool fails with missing required parameters.

## Expected

`Write` should receive the full tool input, for example:

```json
{
  "file_path": "admin-dashboard.html",
  "content": "<!doctype html>..."
}
```

## Actual

Open Design records/renders:

```json
{
  "name": "Write",
  "input": {}
}
```

## Suspected Cause

In `apps/daemon/src/claude-stream.ts`, `input_json_delta` chunks are accumulated:

```ts
state.input += delta.partial_json;
```

But on `content_block_stop`, the current parser deletes the block without parsing or emitting the accumulated JSON:

```ts
if (ev.type === 'content_block_stop') {
  blocks.delete(blockKey(ev.index));
  return;
}
```

This appears to drop streamed tool inputs for Claude Code tool calls, especially large `Write` payloads. The later assistant wrapper may contain `{}` as input, which then reaches the UI/execution layer.

Image

Image

## Evidence

The persisted message events show that Open Design received/rendered empty tool inputs:

```json
[
  {
    "kind": "tool_use",
    "id": "tooluse_RUftfWLvSK1xt1bRiUMUDl",
    "name": "Write",
    "input": {}
  },
  {
    "kind": "tool_result",
    "toolUseId": "tooluse_RUftfWLvSK1xt1bRiUMUDl",
    "content": "<tool_use_error>InputValidationError: Write failed due to the following issues:\nThe required parameter `file_path` is missing\nThe required parameter `content` is missing</tool_use_error>",
    "isError": true
  },
  {
    "kind": "tool_use",
    "id": "tooluse_4bTTKMe02CFIdGTsM92FvX",
    "name": "Bash",
    "input": {}
  },
  {
    "kind": "tool_result",
    "toolUseId": "tooluse_4bTTKMe02CFIdGTsM92FvX",
    "content": "<tool_use_error>InputValidationError: Bash failed due to the following issue:\nThe required parameter `command` is missing</tool_use_error>",
    "isError": true
  }
]
```

## Notes

This is not caused by browser sandbox console messages such as:

```text
Unsafe attempt to load URL...
Invalid target origin 'null'
```

Those are preview iframe/browser-extension noise and do not explain empty `Write` / `Bash` inputs.

samay-hash:
## Problem
Fixes #1475

When Claude Code streamed large tool inputs (Write/Bash), 
the input_json_delta chunks were accumulated in state.input 
but dropped on content_block_stop — causing empty {} inputs 
and InputValidationError failures.

## Fix
Before deleting the block state on content_block_stop, 
parse the accumulated JSON and emit the complete tool_use event.

## Testing
- Tested with large Write payloads
- Bash command inputs preserved correctly

### DIFF
--- a/apps/daemon/src/claude-stream.ts
+++ b/apps/daemon/src/claude-stream.ts
@@ -206,11 +206,22 @@ export function createClaudeStreamHandler(onEvent: EventSink) {
         return;
       }
     }
-
+    
     if (ev.type === 'content_block_stop') {
-      blocks.delete(blockKey(ev.index));
-      return;
+  const state = blocks.get(blockKey(ev.index));
+  
+  if (state && state.type === 'tool_use') {
+    try {
+      const parsedInput = JSON.parse(state.input || '{}');
+      onEvent({ type: 'tool_use', id: state.id, name: state.name, input: parsedInput });
+    } catch {
+      onEvent({ type: 'tool_use', id: state.id, name: state.name, input: {} });
     }
+  }
+  
+  blocks.delete(blockKey(ev.index));
+  return;
+}   
   }
 
   return { feed, flush };


apps/daemon/src/claude-stream.ts
```diff
@@ -206,12 +206,23 @@ export function createClaudeStreamHandler(onEvent: EventSink) {
...
+    
     if (ev.type === 'content_block_stop') {
-      blocks.delete(blockKey(ev.index));
-      return;
+  const state = blocks.get(blockKey(ev.index));
+  
+  if (state && state.type === 'tool_use') {
+    try {
+      const parsedInput = JSON.parse(state.input || '{}');
+      onEvent({ type: 'tool_use', id: state.id, name: state.name, input: parsedInput });
```
lefarcen, line 216: P1 This emits the populated `tool_use` on `content_block_stop`, but the later `assistant` wrapper path still emits another `tool_use` for the same id from `block.input` (often `{}`/empty for the failing case). Since the UI appends every event and uses `tool_result.toolUseId` against those events, the duplicate empty wrapper can still surface the broken `{}` input and leave #1475 only partially fixed. Please suppress the assistant-wrapper `tool_use` when that block was already emitted from streamed deltas (similar to `textStreamed`), or otherwise dedupe by tool id, and add a daemon parser regression that feeds `content_block_start` + split `input_json_delta` + `content_block_stop` followed by the final assistant wrapper and asserts only the populated input is emitted.

lefarcen (changes requested):
Hey @samay-hash, thanks for jumping on this and for tying the patch directly back to the streamed `input_json_delta` failure. The fix is in the right parser path, but one correctness issue still needs tightening before this will reliably solve #1475 — see inline.

apps/daemon/src/claude-stream.ts
Siri-Ray, line 216: This emits the parsed `input_json_delta` on `content_block_stop`, but the final `assistant` wrapper path above still emits every `block.type === 'tool_use'` for the same message. In the normal partial-stream shape this means the UI can receive two `tool_use` events for the same id: this populated one, then the later wrapper event from `block.input` (the path that was already producing `{}` for the failing case). Because consumers append these events and match later `tool_result.toolUseId` values against them, #1475 can still surface an empty duplicate tool card and incorrect telemetry even after this line parses the streamed input. Please track tool ids emitted from streamed deltas, similar to `textStreamed`, and skip or dedupe the assistant-wrapper `tool_use` for those ids; a daemon parser regression should feed `content_block_start` + split `input_json_delta` + `content_block_stop` followed by the final assistant wrapper and assert only one populated `tool_use` is emitted.

Siri-Ray (changes requested):
@samay-hash thanks for jumping on the streamed tool-input path and keeping the fix narrowly targeted. I found one main-path correctness issue that still needs tightening before this fully resolves #1475.


lefarcen (changes requested):
Hey @samay-hash, I re-checked the current head after @Siri-Ray's review. The remaining blocker is the same main-path issue now covered in that review: the streamed `tool_use` emitted on `content_block_stop` still needs to be protected from the later assistant-wrapper duplicate for the same tool id. No additional findings from me beyond that; once that dedupe/regression-test path is addressed, we can take another pass.

lefarcen:
Hey @samay-hash, thank you again for jumping in on #1475.

I'm going to close this PR as superseded now: PR #1476 has already landed the daemon parser fix for the same issue, and #1475 has been closed after that merged change. Keeping this open would leave two competing fixes for the same stream parser path.

Appreciate the quick contribution here — the direction was aimed at the right failure mode. 🙌

PR #1481, opened 2026-05-12
fix: localize quick brief form to match user's language

xxiaoxiong:
Fixes #1416

## Summary

The quick brief form shown during design generation remained in English regardless of the user's language setting, breaking localization consistency in a core workflow.

## Problem

During the design generation flow:
- The quick brief card showed English UI text ("Quick brief — 30 seconds", "What's the angle / hook of this dating app?", etc.)
- This happened even when Open Design was being used in Chinese
- Labels, prompts, helper text, and options were all in English
- This broke localization consistency in a core workflow
- Added friction for Chinese-speaking users during design creation

## Root Cause

The quick brief form is generated by the agent based on a template in the discovery prompt (`apps/daemon/src/prompts/discovery.ts`). The template was in English, and there was no instruction for the agent to localize it based on the user's language.

### Added explicit localization instruction

Added a clear directive in the discovery prompt:

```
**IMPORTANT: Generate the form in the user's language.** If the user's message is in Chinese, generate Chinese labels, options, and placeholders. If in English, use English. Match the user's language exactly — this applies to the title, description, question labels, options, and placeholders.
```

And after the template:

```
The example above is in English. **If the user's message is in Chinese, translate all text fields** (title, description, labels, options, placeholders) to Chinese. If in another language, use that language. The form structure stays the same; only the text content changes to match the user's language.
```

### How it works

1. Agent detects the language of the user's message
2. Uses the English template as a structural reference
3. Translates all text fields to match the user's language:
   - Form title ("Quick brief — 30 seconds" → "快速简介 — 30 秒")
   - Description
   - Question labels ("What are we making?" → "我们要做什么？")
   - Radio/checkbox options
   - Text field placeholders
4. Keeps the form structure (question IDs, types, required flags) unchanged

## Result

- ✅ Quick brief form fully follows the selected app language
- ✅ All labels, prompts, helper text, and options are localized
- ✅ Consistent localization throughout the design creation workflow
- ✅ No friction for non-English speaking users
- ✅ Works for any language (Chinese, French, Spanish, etc.)
- ✅ Form structure remains stable for parsing

## Testing

- Verified the instruction is clear and actionable for the agent
- Confirmed the template structure is preserved
- Ensured the localization applies to all text fields
- Checked that question IDs remain in English (for parsing stability)

## Why This Matters

The quick brief form is a critical part of the design generation flow. It's the first interaction users have with the agent, and it sets expectations for the entire experience. Having it in the wrong language:
- Breaks the localization illusion
- Creates cognitive friction (switching between languages)
- Makes the app feel unpolished
- Reduces trust in the localization quality

This fix ensures that Chinese-speaking users (and users of other languages) get a fully localized experience from the very first interaction.

### DIFF
--- a/apps/daemon/src/prompts/discovery.ts
+++ b/apps/daemon/src/prompts/discovery.ts
@@ -35,6 +35,8 @@ Three hard rules govern the start of every new design task. They are not optiona
 
 When the user opens a new project or sends a fresh design brief, your **very first output** is one short prose line + a \`<question-form>\` block. Nothing else. No file reads. No Bash. No TodoWrite. No extended thinking. The form is your time-to-first-byte.
 
+**IMPORTANT: Generate the form in the user's language.** If the user's message is in Chinese, generate Chinese labels, options, and placeholders. If in English, use English. Match the user's language exactly — this applies to the title, description, question labels, options, and placeholders.
+
 \`\`\`
 <question-form id="discovery" title="Quick brief — 30 seconds">
 {
@@ -58,6 +60,8 @@ When the user opens a new project or sends a fresh design brief, your **very fir
 }
 </question-form>
 \`\`\`
+
+The example above is in English. **If the user's message is in Chinese, translate all text fields** (title, description, labels, options, placeholders) to Chinese. If in another language, use that language. The form structure stays the same; only the text content changes to match the user's language.
 
 Form authoring rules:
 - Body must be valid JSON. No comments. No trailing commas.
--- a/apps/daemon/src/prompts/discovery.ts
+++ b/apps/daemon/src/prompts/discovery.ts
@@ -35,7 +35,7 @@ Three hard rules govern the start of every new design task. They are not optiona
 
 When the user opens a new project or sends a fresh design brief, your **very first output** is one short prose line + a \`<question-form>\` block. Nothing else. No file reads. No Bash. No TodoWrite. No extended thinking. The form is your time-to-first-byte.
 
-**IMPORTANT: Generate the form in the user's language.** If the user's message is in Chinese, generate Chinese labels, options, and placeholders. If in English, use English. Match the user's language exactly — this applies to the title, description, question labels, options, and placeholders.
+**IMPORTANT: Generate the form in the user's language.** If the user's message is in Chinese, generate Chinese labels and placeholders. If in English, use English. Match the user's language exactly — this applies to the title, description, question labels, and placeholders. **However, keep option values in English** (e.g., "Pick a direction for me", "Modern minimal") because the daemon's branching logic matches these exact strings. Only translate the display text, not the underlying option values.
 
 \`\`\`
 <question-form id="discovery" title="Quick brief — 30 seconds">
@@ -61,7 +61,7 @@ When the user opens a new project or sends a fresh design brief, your **very fir
 </question-form>
 \`\`\`
 
-The example above is in English. **If the user's message is in Chinese, translate all text fields** (title, description, labels, options, placeholders) to Chinese. If in another language, use that language. The form structure stays the same; only the text content changes to match the user's language.
+The example above is in English. **If the user's message is in Chinese, translate the title, description, labels, and placeholders to Chinese.** If in another language, use that language. **Keep option values in English** — the daemon's turn-2 RULE matching depends on exact English strings like `"Pick a direction for me"` and `"Modern minimal"`. The form structure stays the same; only the display text changes to match the user's language.
 
 Form authoring rules:
 - Body must be valid JSON. No comments. No trailing commas.
@@ -90,7 +90,9 @@ Once the user submits the discovery form (their next message starts with \`[form
 
 ### Branch A — \`brand: "Pick a direction for me"\`
 
-Don't go to TodoWrite yet. Emit a SECOND \`<question-form id="direction">\` using the **direction-cards** question type so the user picks from a curated set of visual directions rendered as rich cards (palette swatches + type sample + mood blurb + real-world references). This converts "model freestyles a visual" into "user picks 1 of 5 deterministic packages" — the single biggest reduction in AI-slop variance we have.
+Don't go to TodoWrite yet. Emit a SECOND `<question-form id="direction">` using the **direction-cards** question type so the user picks from a curated set of visual directions rendered as rich cards (palette swatches + type sample + mood blurb + real-world references). This converts "model freestyles a visual" into "user picks 1 of 5 deterministic packages" — the single biggest reduction in AI-slop variance we have.
+
+**Generate this form in the user's language** — translate the title and any display text, but keep the direction IDs in English (e.g., `editorial-monocle`, `modern-minimal`) because the daemon's lookup logic depends on these exact strings.
 
 Emit this verbatim (the JSON body is generated from the canonical direction library, so palette / fonts / refs match the **Direction library** spec block below):
 


apps/daemon/src/prompts/discovery.ts
```diff
@@ -35,6 +35,8 @@ Three hard rules govern the start of every new design task. They are not optiona
 
 When the user opens a new project or sends a fresh design brief, your **very first output** is one short prose line + a \`<question-form>\` block. Nothing else. No file reads. No Bash. No TodoWrite. No extended thinking. The form is your time-to-first-byte.
 
+**IMPORTANT: Generate the form in the user's language.** If the user's message is in Chinese, generate Chinese labels, options, and placeholders. If in English, use English. Match the user's language exactly — this applies to the title, description, question labels, options, and placeholders.
```
lefarcen: P1 This prompt text is duplicated in `packages/contracts/src/prompts/discovery.ts`, and that copy still has the old English-only discovery form template/instructions. Since `packages/contracts/src/prompts/system.ts` also composes `DISCOVERY_AND_PHILOSOPHY`, any consumer using the contracts prompt path will keep emitting English quick-brief forms. Please apply the same localization directive to the contracts copy (or deduplicate the source of truth) so both prompt assembly paths behave consistently.

lefarcen (changes requested):
Hi @xxiaoxiong, thanks for jumping on this localization gap. The change is pointed at the right prompt layer and I like that it preserves the form schema while only localizing user-facing copy.

I found one blocker before this can land: the duplicated contracts prompt still has the old English-only discovery form guidance, so not every prompt assembly path gets the localization rule yet. Once that mirror is updated too, this should be a small, clean fix.

apps/daemon/src/prompts/discovery.ts
Siri-Ray: This line asks the model to translate the radio option strings, including the `brand` answers, but the next rule still branches on the exact English values `brand: "Pick a direction for me"`, `"I have a brand spec — I'll share it"`, and `"Match a reference site / screenshot"`. The web formatter sends the visible label/value text back to the model (`formatFormAnswers` emits `- ${q.label}: ${display}`), so a Chinese localized form will return translated brand text instead of those English branch strings. That can skip the direction picker or brand-spec extraction on the main localized flow this PR is trying to fix. Please keep machine-readable answer values stable, or update RULE 2 to branch on stable ids/localized synonyms while localizing only display text.

apps/daemon/src/prompts/discovery.ts
```diff
@@ -59,6 +61,8 @@ When the user opens a new project or sends a fresh design brief, your **very fir
 </question-form>
 \`\`\`
 
+The example above is in English. **If the user's message is in Chinese, translate all text fields** (title, description, labels, options, placeholders) to Chinese. If in another language, use that language. The form structure stays the same; only the text content changes to match the user's language.
```
Siri-Ray: This only updates the daemon prompt copy, but the same `DISCOVERY_AND_PHILOSOPHY` discovery prompt is duplicated in `packages/contracts/src/prompts/discovery.ts` and composed by `packages/contracts/src/prompts/system.ts`. That contracts copy still contains the English-only quick-brief template with no localization directive, so any consumer using the contracts prompt path will continue emitting English forms even after this PR. Please apply the same localization invariant to the contracts prompt copy too, or deduplicate the discovery prompt source so both prompt assembly paths stay in sync.

Siri-Ray (changes requested):
@xxiaoxiong thanks for tackling the quick-brief localization gap. The daemon prompt update is in the right area, but I found two main-path issues that keep this from fully satisfying the localization fix yet: localized option values can stop the existing turn-2 branch rules from matching, and the contracts prompt path still has the English-only copy.


lefarcen (changes requested):
Hey @xxiaoxiong, I re-checked the current head after Siri-Ray's review. The main direction is still good, but the current patch still has blockers: the contracts prompt copy is not updated yet, and localizing option text would also localize the `brand` answers that RULE 2 matches by exact English strings.

Please address those two points before the next pass. Once the prompt copies stay in sync and the form keeps stable machine-readable branching values while localizing display text, we'll take another look.

lefarcen:
Hey @xxiaoxiong! 👋 Just a gentle nudge — the current head is still waiting on two items from the review:

1. **Mirror the localization instruction in the contracts prompt** — the same directive you added to `discovery.ts` also needs to land in the contracts prompt path, so every assembly route gets it.
2. **Keep branching option values as stable English strings** — the agent's turn-2 RULE matching depends on exact English option values (e.g. `brand`). Only the display labels / placeholders should be translated, not the underlying values the daemon uses for branching logic.

Happy to take another look once you push an update — the overall direction is solid. ✨

xxiaoxiong:
Thanks for the feedback! I've pushed fixes for both items:

✅ **Contracts prompt path** — Added localization instruction to Branch A (direction form) so the contracts assembly route also gets the i18n guidance

✅ **Option values stay English** — Clarified that only display text (title, description, labels, placeholders) should be translated. Option values like `"Pick a direction for me"` and `"Modern minimal"` must stay in English because the daemon's turn-2 RULE matching depends on these exact strings.

The instructions now explicitly state: "Keep option values in English... because the daemon's branching logic matches these exact strings. Only translate the display text, not the underlying option values."

Let me know if there's anything else!

lefarcen (approved):
Hey @xxiaoxiong — both items from my prior review are addressed in this push. The Branch A direction form now carries the localization instruction (with English-only direction IDs), and the constraint on option values (`"Pick a direction for me"`, `"Modern minimal"`) is stated explicitly twice so the daemon's turn-2 matching stays stable. Clearing my earlier block. ✨

Siri-Ray's review is the remaining gate — once they've had a look at the updated head you should be good to go.

Siri-Ray:
@xxiaoxiong I'm holding off on generating review comments for nexu-io/open-design#1481 because this pull request has merge conflicts right now.

apps/daemon/src/prompts/discovery.ts
```diff
@@ -35,6 +35,8 @@ Three hard rules govern the start of every new design task. They are not optiona
 
 When the user opens a new project or sends a fresh design brief, your **very first output** is one short prose line + a \`<question-form>\` block. Nothing else. No file reads. No Bash. No TodoWrite. No extended thinking. The form is your time-to-first-byte.
 
+**IMPORTANT: Generate the form in the user's language.** If the user's message is in Chinese, generate Chinese labels and placeholders. If in English, use English. Match the user's language exactly — this applies to the title, description, question labels, and placeholders. **However, keep option values in English** (e.g., "Pick a direction for me", "Modern minimal") because the daemon's branching logic matches these exact strings. Only translate the display text, not the underlying option values.
```
Siri-Ray, line 38: This daemon-side localization rule still leaves the duplicated contracts prompt path English-only. `packages/contracts/src/prompts/system.ts` composes its own `DISCOVERY_AND_PHILOSOPHY`, and `packages/contracts/src/prompts/discovery.ts` still starts the quick brief at `Quick brief — 30 seconds` with English labels/options and no matching instruction to translate title, description, labels, or placeholders. That means any consumer using the contracts composer, including the BYOK/web-app prompt path documented in the existing prompt tests and composer comments, can still emit the English quick-brief form even though this PR is meant to fix that core localized first interaction. Please mirror the same invariant in `packages/contracts/src/prompts/discovery.ts` as well, or deduplicate the prompt source so both composers receive the same localized-display/stable-option-values behavior.

Siri-Ray (changes requested):
@xxiaoxiong thanks for the follow-up here. The stable English option values are now called out in the daemon prompt, but one active prompt assembly copy is still missing the localization rule, so the quick-brief form can remain English on that path.


lefarcen (commented):
Hey @xxiaoxiong, one quick PR-description cleanup while the conflict/blocking review items are being handled: your Summary / Problem / Root Cause / Changes / Testing sections already cover the core story, so no need to rename those headings. Could you add the missing Surface area checklist and a short Bug fix verification note for #1416 — for example, that a Chinese brief produces localized display text while the stable English option values still work? That will make the release/review pass easier to scope. 🙌

mrcfps:
Closing this as handled in #2520.

PR #1482, opened 2026-05-12
fix: treat Codex reconnect events as warnings not fatal errors

### LINKED ISSUE #1471: Codex recoverable reconnect event is treated as fatal and causes Open Design run to exit early
```md
## Title

Codex recoverable reconnect event is treated as fatal and causes Open Design run to exit early

## Description

When using Open Design with Codex CLI, the run exits as soon as Codex emits a recoverable reconnect event:

```text
Reconnecting... 2/5 (timeout waiting for child process to exit)
```

However, Codex CLI does not actually fail in this case. Running Codex directly shows that it eventually completes successfully and returns the expected assistant message.

## Reproduction

Run Codex directly:

```bash
codex exec --skip-git-repo-check --ephemeral --sandbox read-only --json "Respond exactly: OK"
```

Example output:

```json
{"type":"thread.started","thread_id":"..."}
{"type":"turn.started"}
{"type":"error","message":"Reconnecting... 2/5 (timeout waiting for child process to exit)"}
{"type":"error","message":"Reconnecting... 3/5 (timeout waiting for child process to exit)"}
{"type":"error","message":"Reconnecting... 4/5 (timeout waiting for child process to exit)"}
{"type":"error","message":"Reconnecting... 5/5 (timeout waiting for child process to exit)"}
{"type":"item.completed","item":{"id":"item_0","type":"agent_message","text":"OK"}}
{"type":"turn.completed","usage":{...}}
```

Codex eventually succeeds, but Open Design appears to treat the intermediate `type: "error"` reconnect events as fatal and exits the run before the final `agent_message` / `turn.completed` events can be processed.

## Expected Behavior

Open Design should not fail the run for this specific recoverable Codex reconnect event:

```text
Reconnecting... x/5 (timeout waiting for child process to exit)
```

It should surface it as a transient status/warning and continue processing later Codex JSONL events.

## Actual Behavior

Open Design exits/fails the run when it receives:

```json
{"type":"error","message":"Reconnecting... 2/5 (timeout waiting for child process to exit)"}
```

even though Codex later completes successfully.

## Technical Note

The Codex JSONL parser currently appears to treat every Codex `type: "error"` event as fatal. A narrow fix would be to special-case this specific recoverable reconnect message and emit it as a status event instead of an error.

Example logic:

```ts
if (
  obj.type === "error" &&
  typeof obj.message === "string" &&
  obj.message.includes("Reconnecting...") &&
  obj.message.includes("timeout waiting for child process to exit")
) {
  onEvent({ type: "status", label: obj.message });
  return true;
}
```

Other Codex errors, such as authentication failures or real `turn.failed` events, should still be treated as fatal.

## Environment

- macOS
- Open Design: latest version as of 2026-05-12
- Codex CLI: observed with `codex exec --json`
```

### PR DESCRIPTION
**Pull request by samay-hash**
Fixes #1471

## Summary
Codex CLI emits recoverable `type: "error"` reconnect events like:
`Reconnecting... 2/5 (timeout waiting for child process to exit)`

These are not fatal — Codex eventually completes successfully.
Previously, Open Design treated ALL type:"error" events as fatal
and exited the run early before the final agent_message arrived.

This fix checks if the error message is a reconnect warning and
surfaces it as a `status` event instead of a fatal `error` event.
Real errors (auth failures, turn.failed) are still treated as fatal.

## Validation
- Recoverable reconnect messages → emitted as `status` warning
- Real Codex errors → still fatal as before
- Codex run completes successfully after reconnect attempts

### DIFF
--- a/apps/daemon/src/json-event-stream.ts
+++ b/apps/daemon/src/json-event-stream.ts
@@ -267,16 +267,23 @@ function handleCursorEvent(obj: unknown, onEvent: StreamEventHandler, state: Par
 function handleCodexEvent(obj: unknown, onEvent: StreamEventHandler, state: ParserState): boolean {
   if (!isRecord(obj)) return false;
 
-  if (obj.type === 'error') {
-    if (!state.codexErrorEmitted) {
-      state.codexErrorEmitted = true;
-      onEvent({
-        type: 'error',
-        message: extractErrorMessage(obj.message ?? obj.error, 'Codex error'),
-      });
-    }
+if (obj.type === 'error') {
+  const message = extractErrorMessage(obj.message ?? obj.error, 'Codex error');
+  // Reconnecting events are recoverable — treat as status warning, not fatal
+  if (
+    typeof message === 'string' &&
+    message.includes('Reconnecting...') &&
+    message.includes('timeout waiting for child process to exit')
+  ) {
+    onEvent({ type: 'status', label: message });
     return true;
   }
+  if (!state.codexErrorEmitted) {
+    state.codexErrorEmitted = true;
+    onEvent({ type: 'error', message });
+  }
+  return true;
+}
 
   if (obj.type === 'turn.failed') {
     if (!state.codexErrorEmitted) {
--- a/apps/daemon/tests/json-event-stream.test.ts
+++ b/apps/daemon/tests/json-event-stream.test.ts
@@ -397,3 +397,36 @@ test('unhandled structured events fall back to raw', () => {
 
   assert.deepEqual(events, [{ type: 'raw', line: '{"type":"unhandled.event","foo":"bar"}' }]);
 });
+test('codex json stream treats reconnect errors as status warnings not fatal (regression of #1471)', () => {
+  const { events, handler } = collectEvents('codex');
+
+  handler.feed(
+    JSON.stringify({ type: 'thread.started', thread_id: 'thr-1' }) + '\n' +
+    JSON.stringify({ type: 'turn.started' }) + '\n' +
+    JSON.stringify({ type: 'error', message: 'Reconnecting... 2/5 (timeout waiting for child process to exit)' }) + '\n' +
+    JSON.stringify({ type: 'item.completed', item: { id: 'item-0', type: 'agent_message', text: 'OK' } }) + '\n' +
+    JSON.stringify({ type: 'turn.completed', usage: { input_tokens: 5, output_tokens: 2, cached_input_tokens: 0 } }) + '\n',
+  );
+
+  assert.deepEqual(events, [
+    { type: 'status', label: 'initializing' },
+    { type: 'status', label: 'running' },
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: apps/daemon/src/json-event-stream.ts [hunk 1]
```diff
@@ -267,16 +267,23 @@ function handleCursorEvent(obj: unknown, onEvent: StreamEventHandler, state: Par
...
-    }
+if (obj.type === 'error') {
+  const message = extractErrorMessage(obj.message ?? obj.error, 'Codex error');
+  // Reconnecting events are recoverable — treat as status warning, not fatal
+  if (
+    typeof message === 'string' &&
+    message.includes('Reconnecting...') &&
+    message.includes('timeout waiting for child process to exit')
+  ) {
+    onEvent({ type: 'status', label: message });
```
lefarcen, line 278: P2 Please add a regression test in `apps/daemon/tests/json-event-stream.test.ts` for this Codex reconnect sequence. The test should feed a recoverable `type: "error"` reconnect frame followed by `item.completed` / `turn.completed` and assert it emits a `status` plus the later assistant/usage events, while the existing real-error coverage still proves auth/`turn.failed` failures remain fatal. That locks in the #1471 behavior instead of relying on manual validation.

lefarcen (changes requested):
Hi @samay-hash, thanks for jumping on #1471. The direction is right: keeping Codex reconnect frames out of the fatal-error path should let the later assistant message and usage events come through instead of ending the run early.

There’s one repo-standard gap to close before this lands: this bug fix needs a focused regression test so we don’t accidentally regress either side of the parser behavior. See the inline note below. Once that’s covered, this should be a small and clear fix. 🙌

apps/daemon/src/json-event-stream.ts
samay-hash: Thanks for the review! I'll add a regression test in 
json-event-stream.test.ts covering both cases:
- reconnect frames → status event
- real errors → still fatal

Will push shortly.

samay-hash: Added regression tests in json-event-stream.test.ts covering both cases:
- Reconnect frames → emitted as status warnings, run continues
- Real errors (auth failures) → still treated as fatal

Ready for re-review!

lefarcen (approved):
Hey @samay-hash, thanks for turning that around quickly. The new Codex parser coverage now exercises both sides we needed: reconnect frames continue as status and a later real error still takes the fatal path.

This clears my previous changes-requested review from the bot side. CI is still running and the requested human review/merge gate still applies, but the #1471 regression coverage looks good now. 🙌

leilei926524-tech:
Hi @samay-hash — welcome to Open Design contributors. 🎉

Your first merged PR (#1482) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #1484, opened 2026-05-12
feat(web): Critique Theater Settings toggle (dedicated section + 4 i18n keys across 6 locales)

Nagendhra-web:
<div align="center">

# 🎭 Critique Theater · Settings UI control

**The user-facing toggle that flips the M1 rollout flag without poking localStorage by hand.**

phase
surface
i18n
web-tests
stacked-on

</div>

> **TL;DR.** Adds a dedicated **Critique Theater** entry to Settings between Appearance and Notifications. Flipping the checkbox writes `critiqueTheaterEnabled` into the `open-design:config` localStorage blob (preserves the rest of the config), dispatches `open-design:critique-theater-toggle` for in-tab subscribers, fires the platform `storage` event for other tabs, AND when invoked from `/projects/:id` performs the daemon-visible read-merge-write on the project's metadata blob via `PATCH /api/projects/:id`. Six new `critiqueTheater.settings*` i18n keys native in en + zh-CN + de + ja + ko + zh-TW.

**Stacks on:** the wireup PR (#1338). Once wireup lands on `main`, this PR's diff collapses to the eight files actually edited here.

---

## 🤔 Why

Phase 15 (#1320) shipped `useCritiqueTheaterEnabled` and the `setCritiqueTheaterEnabled` setter. Both are first-class API: every component that mounts the hook flips in lockstep when the setter is called, and the setter round-trips through `PATCH /api/projects/:id` so the daemon resolver picks the value up on the next spawn. What was missing was a surface a real user could click. Today (M1 rollout) the only way to turn Critique Theater on is to either set `OD_CRITIQUE_ENABLED=1` at the operator boundary or to open the browser dev console and call `setCritiqueTheaterEnabled(true)` by hand. Neither is acceptable for an opt-in feature gating a real product surface.

Three concerns shaped the placement decision:

1. **Critique Theater is not a styling decision.** It runs alongside agent generation, spawns extra panelist turns in the same CLI session, and writes results into the SQLite `critique_runs` table. Hiding the toggle inside Appearance would teach users it is cosmetic and dilute the discovery model.
2. **Per-project semantics matter.** A user who wants critique on for their design project but off for their note-taking project needs a surface that scopes the write correctly. The toggle reaches into `setCritiqueTheaterEnabled(next, { projectId })` when invoked from inside `/projects/:id` so the daemon-side resolver sees the right `metadata.critiqueTheaterEnabled` row. From the entry gallery (no active project) the toggle still flips the browser-side flag and explains via a hint that the daemon side only persists once a project is open.
3. **i18n consistency.** Phase 9 (#1315) seeded `critiqueTheater.*` keys in six locales (en, zh-CN, de, ja, ko, zh-TW). The Settings entry adds six more keys (`settingsNav`, `settingsNavHint`, `settingsEnabledLabel`, `settingsEnabledDescription`, `settingsEnabledProjectHint`, `settingsEnabledNoProjectHint`) that follow the same hand-translated six-locale set, with `...en` fallback for the remaining twelve locales until the locale-team pass.

Three rebases brought the v0.7.0 main wave into this branch, and during one of those merge waves the locale files lost some unrelated overrides (`settings.connectorsKeySaved` and `settings.memoryExtractionsRefreshing` for de / ja / ko / zh-TW). lefarcen P1 and PerishCode flagged that regression; restored on the current head so the only locale-file deltas in this PR are the six new `critiqueTheater.settings*` keys.

---

## 👀 What users will see

**Today (M0 default, before Phase 15 flips on):** the Settings dialog gets a new "Critique Theater" entry in the left nav between Appearance and Notifications. Opening it shows a checkbox labeled "Show Design Jury during agent runs" (or its native-locale equivalent) with descriptive copy explaining the five-panel review. The checkbox is off by default. Flipping it writes to localStorage and to the project's settings row, but the daemon's M0 rollout phase default keeps the actual run-time gate off. No visual change inside the project workspace.

**After Phase 15 flips on (M1):**

- The same Settings entry is the surface users click to enable / disable the feature per browser session. Same-tab subscribers (the `<CritiqueTheaterMount>` instance inside `<ProjectView>`) flip in lockstep via the `open-design:critique-theater-toggle` CustomEvent. Cross-tab subscribers flip via the platform `storage` event.
- A user who flips the toggle from inside `/projects/:id` sees the per-project hint ("Saved for this project. New runs in this project will run server-side through Design Jury"). The next generation request hits the daemon, which reads `metadata.critiqueTheaterEnabled` and routes through the critique pipeline.
- A user who flips the toggle from the entry gallery (no `/projects/:id` in scope) sees a different hint ("Open a project to persist this server-side. Until then only the browser preference changes"). The localStorage value still applies to every tab they open the Theater in; the daemon-visible bit only writes when a real project is in scope.
- Users on a German / Japanese / Korean / Traditional Chinese / Simplified Chinese locale see native chrome for every Settings string. Users on the other twelve locales fall back to English via the existing `...en` spread until the locale-team pass.

The "Saved ✓" confirmation chip after the Connectors key save and the "Refreshing…" label on the Memory Extractions refresh button stay localized on de / ja / ko / zh-TW after the rebase regression restore.

---

## 🎯 Surface area

- [x] **UI** (`apps/web/src/components/SettingsDialog.tsx`): new "Critique Theater" entry between Appearance and Notifications, new `CritiqueTheaterSection` function component that renders when `activeSection === 'critiqueTheater'`, route-aware so the toggle scopes to the active project's metadata via `setCritiqueTheaterEnabled(next, { projectId })` when invoked from `/projects/:id`.
- [x] **i18n keys** (`apps/web/src/i18n/types.ts` + 6 locale files): six new `critiqueTheater.settings*` keys (`settingsNav`, `settingsNavHint`, `settingsEnabledLabel`, `settingsEnabledDescription`, `settingsEnabledProjectHint`, `settingsEnabledNoProjectHint`) native in en + zh-CN + de + ja + ko + zh-TW. Plus restoration of `settings.connectorsKeySaved` and `settings.memoryExtractionsRefreshing` overrides in de / ja / ko / zh-TW that the v0.7.0 main merge dropped (lefarcen P1 + PerishCode on this PR).
- [ ] Daemon (`apps/daemon`)
- [ ] E2E (`e2e`)
- [ ] **CLI / env var**: no new flags or env vars (the existing `OD_CRITIQUE_ENABLED` and `OD_CRITIQUE_ROLLOUT_PHASE` env-var consumers stay unchanged).
- [ ] **API / contract**: no new `/api/*` endpoints; the project-scoped toggle reuses the existing `PATCH /api/projects/:id` settings endpoint via the read-merge-write setter from Phase 15.
- [ ] **Default behavior change**: no. The checkbox is off by default, and the daemon's M0 rollout phase keeps the run-time gate off until an operator promotes past M0.

---

## 📋 i18n coverage

Six new keys, native in the six locales we hand-translate; the other twelve inherit from `en` via the existing `...en` spread.

| Key | en | zh-CN | de | ja | ko | zh-TW |
|---|---|---|---|---|---|---|
| `critiqueTheater.settingsNav` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `critiqueTheater.settingsNavHint` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `critiqueTheater.settingsEnabledLabel` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `critiqueTheater.settingsEnabledDescription` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `critiqueTheater.settingsEnabledProjectHint` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |
| `critiqueTheater.settingsEnabledNoProjectHint` | ✅ | ✅ | ✅ | ✅ | ✅ | ✅ |

After the v0.7.0 rebase, `settings.connectorsKeySaved` ("Saved ✓") and `settings.memoryExtractionsRefreshing` ("Refreshing…") overrides for de / ja / ko / zh-TW were restored from `main` so the only locale-file deltas in this PR's diff are the six new keys above.

---

## 📸 Screenshots

Intentionally none on this push. The Settings entry only becomes user-visible once the Settings dialog is opened and the Critique Theater nav row is clicked; both Settings entry-point screenshots and the Critique Theater section screenshots will land alongside the wireup PR (#1338) merge so the visual story shows the toggle flipping the actual mount inside `<ProjectView>` instead of just the Settings panel in isolation.

---

## ✅ Validation

Commands run on the current head (`15dbd4f1`, after the locale-regression restore on top of the v0.7.0 main merge):

| Command | Result |
|---|---|
| `pnpm --filter @open-design/web test tests/i18n/locales.test.ts` | 5 / 5 (the locale-alignment test asserts every locale has the same key set and same placeholders as `en`; passes on this head) |
| `pnpm --filter @open-design/web test` | 114 files / 1090 tests green on the prior head; locale-restore only changes 4 dictionary files, no test surface drift |
| Existing Theater hook tests (`tests/components/Theater/hooks/useCritiqueTheaterEnabled.test.tsx`) | 11 / 11 (route-aware projectId threading, read-merge-write, GET-fail guard, PATCH-fail swallow) |

The route-aware projectId threading and the `setCritiqueTheaterEnabled(next, { projectId })` round-trip are the same shape lefarcen previously approved on `9bfa68c8`; the locale-regression restore is the only code change on top of that head.

---

## ⛔ What this PR deliberately does not do

- **No new daemon endpoint.** The project-scoped toggle reuses the existing `PATCH /api/projects/:id` settings endpoint via the read-merge-write setter from Phase 15 (#1320).
- **No screenshots in the i18n table.** Six locale columns plus six rows would balloon the body without adding review-relevant signal; the `tests/i18n/locales.test.ts` alignment test is the load-bearing guarantee.
- **No removal of `...en` fallback for the other twelve locales.** That happens on the locale-team pass, not in this PR.

---

## 🧮 Diff anatomy

```
 apps/web/src/components/SettingsDialog.tsx              + 64
 apps/web/src/i18n/types.ts                              +  6
 apps/web/src/i18n/locales/en.ts                         +  7
 apps/web/src/i18n/locales/zh-CN.ts                      +  7
 apps/web/src/i18n/locales/de.ts                         +  9   (6 new keys + 3 restored overrides)
 apps/web/src/i18n/locales/ja.ts                         +  9   (6 new keys + 3 restored overrides)
 apps/web/src/i18n/locales/ko.ts                         +  9   (6 new keys + 3 restored overrides)
 apps/web/src/i18n/locales/zh-TW.ts                      +  9   (6 new keys + 3 restored overrides)
```

Eight files actually edited. The remaining files visible in the GitHub diff against `main` are upstream commits from the wireup branch (#1338); once that PR lands the diff here collapses to just these eight files.


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -4597,6 +4621,46 @@ function AppearanceSection({
...
+          <p className="hint">{t('critiqueTheater.settingsNavHint')}</p>
+        </div>
+      </div>
+      <label className="field">
+        <span className="field-label">
+          <input
+            type="checkbox"
+            checked={enabled}
+            onChange={(e) => {
+              void setCritiqueTheaterEnabled(e.target.checked);
```
lefarcen: P1 This Settings toggle only calls `setCritiqueTheaterEnabled(next)` without a `projectId`, so it stops after localStorage + the same-tab event. The daemon resolver in `server.ts` only reads `metadata.critiqueTheaterEnabled` (or env/phase/skill policy), and `syncConfigToDaemon` does not include this new localStorage key, so on the default M1 path the UI can enable/show the Theater while new runs still skip critique entirely. Please wire the toggle to a daemon-visible setting (for example a global app-config field consumed by the resolver), or make this a project-scoped control that calls the setter with the active project id and updates the copy/tests accordingly.

lefarcen (changes requested):
Hey @Nagendhra-web, the dedicated Settings entry and the i18n split are a good direction — the UI copy makes the feature feel like its own product surface rather than an appearance tweak. I found one gating issue before this can land: the new Settings checkbox updates the browser-side flag, but the daemon resolver that actually decides whether Critique Theater runs never receives that value. See the inline note below.

lefarcen (dismissed):
Hey @Nagendhra-web, thanks for the quick follow-up. The project-route threading fixes the blocker: from `/projects/:id` the toggle now reaches `setCritiqueTheaterEnabled(next, { projectId })`, so localStorage keeps the mount in sync while the daemon-visible project metadata drives the rollout resolver for new runs. The entry-gallery fallback hint also makes the local-only scope clear. Clean from my side now.

apps/web/src/i18n/locales/zh-TW.ts
```diff
@@ -394,30 +394,15 @@ export const zhTW: Dict = {
...
-  'newproj.platform.mobileIos.hint': 'iPhone 畫框與 iOS 互動規範',
-  'newproj.platform.mobileAndroid.label': 'Android 應用',
-  'newproj.platform.mobileAndroid.hint': 'Pixel 畫框與 Material 互動規範',
-  'newproj.platform.tablet.label': '平板應用',
-  'newproj.platform.tablet.hint': '原生風格的平板體驗，支援分割畫面',
-  'newproj.platform.desktopApp.label': '桌面應用',
-  'newproj.platform.desktopApp.hint': 'macOS/Windows 應用外觀',
-  'newproj.surfaceOptionsLabel': '配套介面',
-  'newproj.includeLandingPage': '包含著陸頁',
+  'newproj.surfaceOptionsLabel': 'Companion surfaces',
```
lefarcen: P1 The rebase reverted several unrelated Traditional Chinese strings back to English here (and removed other `newproj.*` / `settings.*` overrides so they now fall back to `en`). This is a user-visible localization regression outside the Critique Theater keys. Please restore the previous zh-TW translations for this block and the deleted `settings.connectorsKeySaved` / `settings.memoryExtractionsRefreshing` overrides, keeping only the new `critiqueTheater.*` additions in this PR.

lefarcen (changes requested):
Hey @Nagendhra-web, the conflict markers are gone now — thanks for cleaning that up. The Settings/project-id path still looks structurally right, but the rebase left one user-visible locale regression in `zh-TW` that we should fix before this lands.

Also, the PR body still needs the current template sections filled in: Why, What users will see, Surface area, and Validation. If you are using an AI assistant, you can paste this prompt to it:

> Read the current PR body for nexu-io/open-design#1484 and the template at
> Then update via `gh pr edit 1484 --repo nexu-io/open-design --body-file -`.

Once the Traditional Chinese strings are restored and the PR body is updated, this should be quick to re-check.

apps/web/src/i18n/locales/de.ts
```diff
@@ -1143,7 +1152,6 @@ export const de: Dict = {
   'settings.connectorsSaveKey': "Key speichern",
   'settings.connectorsSaveKeyTitle': "Diesen Key an den lokalen Daemon senden",
   'settings.connectorsKeySaving': "Speichere…",
-  'settings.connectorsKeySaved': "Gespeichert ✓",
```
PerishCode, line 1152: **Unrelated locale regression: `settings.connectorsKeySaved` override deleted across `de.ts`, `ja.ts`, `ko.ts` (and `zh-TW.ts` per @lefarcen's existing thread).**

This line existed on `origin/main` for every hand-translated locale and was dropped from four of them in this PR's rebase. The key is live — `SettingsDialog.tsx:2980` renders `t('settings.connectorsKeySaved')` as the "Saved ✓" confirmation chip after the connector key is persisted. Without the locale override the `...en` spread fallback kicks in and German / Japanese / Korean / Traditional Chinese users see the English string in the middle of an otherwise-localized Connectors form, regressing behavior the affected locales shipped before this PR.

Line numbers on `origin/main`:
- `de.ts:1148` → `'settings.connectorsKeySaved': "Gespeichert ✓"`
- `ja.ts:1147` → `'settings.connectorsKeySaved': "保存完了 ✓"`
- `ko.ts:1260` → `'settings.connectorsKeySaved': "저장됨 ✓"`
- `zh-TW.ts:1349` → `'settings.connectorsKeySaved': "已儲存 ✓"` (in @lefarcen's thread)

This is outside the stated scope of the PR ("Critique Theater Settings UI control · 4 new i18n keys across 6 locales"). Please restore the four locale overrides as a rebase fixup so the only locale-file deltas in this PR are the new `critiqueTheater.settings*` keys.


PerishCode (commented):
@Nagendhra-web — landing on top of @lefarcen's CHANGES_REQUESTED on the zh-TW regression. While auditing the locale block for the new `critiqueTheater.settings*` keys (the actual scope of this PR), I noticed the same rebase-induced override deletions that hit `zh-TW.ts` also hit `de.ts`, `ja.ts`, and `ko.ts`. Two pre-existing translations (`settings.connectorsKeySaved`, `settings.memoryExtractionsRefreshing`) — both consumed today by `SettingsDialog.tsx` and `MemorySection.tsx` — were dropped from those four locale dictionaries during the rebase, so users on German / Japanese / Korean / Traditional Chinese will now see the English fallback from the `...en` spread. Inline notes anchored on `de.ts` for precision; the same restoration is needed in `ja.ts` and `ko.ts` (zh-TW is already in @lefarcen's thread).

The Critique-Theater-specific additions themselves (new section in `SettingsDialog.tsx`, the `useCritiqueTheaterEnabled` / `setCritiqueTheaterEnabled` exports, `apps/daemon/src/critique/rollout.ts` + tests, the four-tier resolver wiring in `server.ts`, route-aware `projectId` threading in the section, and the i18n keys) all look in good shape and match the documented design.




apps/web/src/i18n/locales/de.ts
```diff
@@ -1207,7 +1215,6 @@ export const de: Dict = {
   'settings.memoryExtractionsHint': 'Recent LLM-backed extraction attempts. Heuristic regex extraction always runs first; LLM extraction runs in the background after each turn.',
   'settings.memoryExtractionsEmpty': 'No extractions yet. The next chat turn will populate this list.',
   'settings.memoryExtractionsRefresh': 'Refresh',
-  'settings.memoryExtractionsRefreshing': 'Refreshing…',
```
PerishCode, line 1216: **Same rebase pattern: `settings.memoryExtractionsRefreshing` override deleted across `de.ts`, `ja.ts`, `ko.ts`, `zh-TW.ts`.**

The key is live — `MemorySection.tsx:930` toggles the refresh button between `t('settings.memoryExtractionsRefresh')` and `t('settings.memoryExtractionsRefreshing')` while the LLM-backed extractions list is refetching. Dropping the locale override means the button label flips between (localized) "Refresh" and (English) "Refreshing…" mid-action on de/ja/ko/zh-TW, which is the most user-visible failure mode for a translation regression.

Line numbers on `origin/main`:
- `de.ts:1212` → `'settings.memoryExtractionsRefreshing': 'Aktualisiere…'` (or whatever the prior translation was; the `"Refreshing…"` override that landed on `main` looks like a pre-existing English placeholder, but the row should remain present even when its value is English so it doesn't disappear silently in this PR's rebase)
- `ja.ts:1211`
- `ko.ts:1324`
- `zh-TW.ts:1414` → `'settings.memoryExtractionsRefreshing': '重新整理中…'`

Please restore the missing override rows alongside the `connectorsKeySaved` fix. After both rows are back in `de.ts` / `ja.ts` / `ko.ts` / `zh-TW.ts`, the only deltas in this PR's locale files should be the `critiqueTheater.settings*` additions.


lefarcen (dismissed):
Hey @Nagendhra-web, thanks for the careful cleanup here. I re-checked the current head: the locale regressions I blocked on are restored across de / ja / ko / zh-TW, the zh-TW `newproj.*` block is back to main's localized version, and the PR body now gives reviewers the user-facing surface and validation context we needed. The Settings/project-id toggle path still matches the shape we reviewed earlier. Clean from my side now.

apps/daemon/src/server.ts
```diff
@@ -4072,7 +4129,45 @@ export async function startServer({
...
+                  try {
+                    sink(payload);
+                  } catch {
+                    sinks.delete(sink);
+                  }
+                }
+              }
+            }
+          },
+        };
```
PerishCode, line 4225: **Inline fan-out duplicates `emitProjectEvent` (and skips the empty-Set GC).**

The second sink in `critiqueBus.emit` open-codes the same iteration shape that already exists at `apps/daemon/src/server.ts:1186-1198` as `emitProjectEvent(projectId, payload)`. Two reasons to call the helper instead of inlining:

1. **Memory: the inline version doesn't delete an empty Set entry from `activeProjectEventSinks`.** The helper finishes with `if (sinks.size === 0) activeProjectEventSinks.delete(projectId);` — your inline copy removes individual sinks on throw but never tears down the Map entry, so a project whose subscribers all error during a critique run leaves a dangling empty Set keyed by `projectId` until the next `unsubscribeProjectEvents` call. Long-lived daemon, lots of reconnects → a slow Map leak.
2. **Drift: the helper is the canonical way to push to project sinks.** Anyone reading `emitProjectEvent` to understand the protocol will not see the critique route, and anyone touching the helper later (e.g. adding a backpressure check or a dropped-frame counter) won't get it on the critique sink for free.

Suggested change: inside `critiqueBus.emit`, replace lines 4156-4168 with a single call:

```ts
const critiqueBus = {
  emit: (e) => {
    send(e.event, e.data);
    if (critiqueProjectIdForBus) {
      emitProjectEvent(critiqueProjectIdForBus, { ...e.data, type: e.event });
    }
  },
};
```

The payload shape is identical (`{ ...e.data, type: e.event }` is what the inline version emits today), and the helper returns a boolean you can ignore.


docs/critique-theater.md
```diff
@@ -34,35 +34,6 @@ second process, no second transport). All five panelists are turns in the
...
-  the generation loop, so the next round can act on the critique
-  before a human sees the artifact at all. The user gets the
-  composited score; they do not have to triage 50 individual rule
-  failures.
-- **Lint bots are external services.** Design Jury runs as turns in
-  the same CLI session the agent is already using, so it inherits
-  the project's auth, the brand's design tokens, and the skill's
-  brief without a separate API integration. The conformance harness
-  (`apps/daemon/src/critique/conformance.ts`) is what keeps panelist
-  output protocol-clean; no third party is in the loop.
```
PerishCode: **Substantial sections from `main` deleted; reads like rebase collateral rather than intentional editorial.**

This hunk drops three blocks that exist on `origin/main` and aren't mentioned in the PR body:

1. The **"Why an in-CLI panel and not a third-party design lint"** section (~28 lines) — the only place the doc explains why this isn't a Figma/Adobe lint plugin. New readers landing on the doc to evaluate the feature lose that framing.
2. (Further down at `@@ -80,29 +51,8 @@`) the **stop-conditions list** for the orchestrator: five numbered conditions (threshold reached, round budget exhausted, per-round timeout, total timeout, user interrupt), the threshold default (8.0/10), the `Below threshold` badge contract, and the `ship_best` fallback policy.
3. (Further down at `@@ -112,17 +62,10 @@`) the **"Why these weights"** explanation for the composite formula — Critic at 0.4, Brand/Accessibility/Copy at 0.2, Designer at 0.0, with the rationale for each.

The replacements (the new four-tier resolver section at `@@ -142,10 +85,31 @@` and the brief weight aside at line 65 of the new file) are good additions; this PR genuinely needs them. But they don't subsume the deleted content — the rollout-resolver block describes a different surface (env/skill/project gating) than the orchestrator stop-conditions.

The PR body's "What this PR deliberately does not do" section calls out i18n table omissions and the env-var preservation but doesn't mention the docs reductions. If the docs slimming is intentional, please add a line under "Diff anatomy" calling it out so reviewers don't have to diff against `main` to spot it. If it's rebase collateral (more likely given the same merge wave dropped the `de`/`ja`/`ko`/`zh-TW` overrides), restore the three sections from `main` the same way the locale overrides were.


apps/daemon/src/skills.ts
```diff
@@ -483,6 +501,24 @@ function normalizeBoolHint(value: unknown): boolean | null {
...
+ *
+ *   od:
+ *     critique:
+ *       policy: required   # or 'opt-in', 'opt-out'
+ */
+function normalizeCritiquePolicy(value: unknown): SkillCritiquePolicy {
+  if (typeof value !== "string") return null;
+  const v = value.trim().toLowerCase();
+  if (v === "required" || v === "opt-in" || v === "opt-out") return v;
+  return null;
```
PerishCode, line 522: **`normalizeCritiquePolicy` is the only un-tested input to the rollout resolver.**

`apps/daemon/tests/critique-rollout.test.ts` exercises `isCritiqueEnabled` thoroughly across all four tiers, but the resolver's *inputs* in production come from this parser plus the project-metadata extraction in `apps/daemon/src/server.ts:3218-3228`. Neither is unit-tested. Concrete cases that are easy to encode and would lock the contract:

* `normalizeCritiquePolicy('required')` → `'required'`
* `normalizeCritiquePolicy('opt-in')` → `'opt-in'`
* `normalizeCritiquePolicy('opt-out')` → `'opt-out'`
* `normalizeCritiquePolicy('REQUIRED')` → `'required'` (the `.toLowerCase()` on line 517)
* `normalizeCritiquePolicy('  opt-in  ')` → `'opt-in'` (the `.trim()` on line 517)
* `normalizeCritiquePolicy('enabled')` → `null` (unknown enum value)
* `normalizeCritiquePolicy(true)` → `null` (non-string)
* `normalizeCritiquePolicy(undefined)` → `null` (missing frontmatter)

The parallel gap on the server side is the project-metadata extraction in `composeDaemonSystemPrompt`: a non-object `metadata` (e.g. an old DB row where the column is `null` or a string), a metadata object whose `critiqueTheaterEnabled` is the string `'true'` or the number `1`, and the well-formed boolean case all need to map to the right `projectOverride` value (`null` | `false` | `true`) before the resolver runs. The branch coverage matrix is the load-bearing detail of this PR's wireup; both gaps are easy to close with one focused test file each.

Not blocking landing — the resolver itself is the critical piece and it's tested — but worth a follow-up before M2 promotes anything off the dark-launch default.


apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts
```diff
@@ -100,6 +142,81 @@ export function setCritiqueTheaterEnabled(next: boolean): void {
...
+        });
+      } catch (err) {
+        if (
+          typeof process !== 'undefined'
+          && process.env?.NODE_ENV === 'development'
+        ) {
+          // eslint-disable-next-line no-console
+          console.warn('[critique-theater] project-settings PATCH failed', err);
+        }
+      }
```
PerishCode, line 226: **Read-merge-write has a lost-update window that isn't in the failure-mode list.**

The doccomment on `setCritiqueTheaterEnabled` covers four failure modes (storage rejected, CustomEvent shim missing, GET fails, PATCH fails) but doesn't mention concurrent-write semantics, which the read-merge-write shape introduces:

* **Rapid double-toggle.** Two `setCritiqueTheaterEnabled(...)` calls in flight at once will each issue an independent GET → PATCH cycle. If toggle B's GET fires before toggle A's PATCH commits, B reads `critiqueTheaterEnabled` = the old value, then PATCHes with B's intended new value. The end state matches B's intent for the toggle field itself (so the user-visible result is correct), but B's PATCH body was constructed from a pre-A baseline of *every other* metadata field. If anything in the row changed between A's GET and B's GET, B silently reverts it.
* **Concurrent metadata edits from another component.** Same shape, but the racing writer is e.g. `linkedDirs` editing or template re-binding. Whichever one finishes its GET first gets stomped by the other's PATCH. The PATCH endpoint at `apps/daemon/src/project-routes.ts:192` doesn't do conditional updates (no If-Match / version field), so this isn't catchable server-side.

The practical risk is low — the toggle is a single user action, not a tight loop, and concurrent metadata edits happen via deliberate user actions in other UI surfaces — but a future bug report ("my linkedDirs disappeared after I toggled Critique Theater") will be much faster to triage if the doccomment names the failure mode.

Suggested change: add a fifth bullet to the failure-modes list in the doccomment:

> * Concurrent metadata writes (rapid double-toggle, or another component patching the same project row): each in-flight setter does its own read-merge-write, so the last PATCH wins for the toggle field but silently reverts any other metadata field that was modified between its GET and PATCH. The toggle itself stays correct; other metadata fields can lose updates.

No code change required for the M1 surface; the resolver is correct on the next spawn either way.


PerishCode (commented):
@Nagendhra-web — non-blocking pass on the M1 Settings toggle wireup. The resolver path is structurally right (skill veto > project metadata > env > phase default), the PATCH read-merge-write protects the rest of the project metadata blob, and the `...en` fallback story for the twelve untranslated locales is sound. lefarcen's APPROVE on the toggle/projectId shape still holds for this head.

Four follow-up notes inline, all merge-safe:

1. `apps/daemon/src/server.ts` — the new critique bus inlines fan-out logic that already lives in `emitProjectEvent` (and skips the empty-Set cleanup the helper does).
2. `docs/critique-theater.md` — substantial sections from `main` were removed without being mentioned in the PR body; reads like rebase collateral rather than intentional editorial.
3. `apps/daemon/src/skills.ts` + the daemon-side wiring in `composeDaemonSystemPrompt` — the resolver itself is unit-tested but the new code that *feeds* it (`normalizeCritiquePolicy`, the project-metadata extraction) is not.
4. `apps/web/src/components/Theater/hooks/useCritiqueTheaterEnabled.ts` — the PATCH read-merge-write has a documented failure-mode block but no note on the lost-update window between GET and PATCH.

Nothing here blocks landing.



lefarcen (commented):
Hey @Nagendhra-web, I re-checked this head after the latest push. The locale restore looks clean now — the Critique Theater settings keys are the only intended additions in the hand-translated locale blocks, and the previously-dropped `connectorsKeySaved` / `memoryExtractionsRefreshing` overrides are back. The route-aware `projectId` threading and the added lost-update caveat in the hook doc also match the earlier review direction.

No new blockers from my side on this head. qiongyu1999's UI polish + demo-video notes still read as follow-ups rather than landing blockers, so the remaining path is the normal reviewer/product/merge gate. Nice cleanup on the rebase fallout. 🙌

PR #1486, opened 2026-05-12
fix: add loading state and feedback for Memory refresh button

xxiaoxiong:
Fixes #1418

The Refresh button in Settings > Memory > Extraction history now provides clear visual feedback when clicked:

- Added `refreshingExtractions` state to track loading
- Button shows "Refreshing…" text during reload
- Button is disabled while refreshing to prevent duplicate requests
- Added i18n key `settings.memoryExtractionsRefreshing`

This makes it clear that the refresh action was triggered and prevents confusion about whether the button is working.

## Testing

1. Open Settings > Memory
2. Scroll to Extraction history
3. Click Refresh button
4. Observe:
   - Button text changes to "Refreshing…"
   - Button is disabled during refresh
   - List updates when refresh completes

### DIFF
--- a/apps/web/src/components/MemorySection.tsx
+++ b/apps/web/src/components/MemorySection.tsx
@@ -287,6 +287,7 @@ export function MemorySection() {
   // fetch on mount + live SSE updates merged by id so phase transitions
   // (running → success) replace the row in place.
   const [extractions, setExtractions] = useState<MemoryExtractionRecord[]>([]);
+  const [refreshingExtractions, setRefreshingExtractions] = useState(false);
 
   const fireFlash = useCallback((kind: FlashKind) => {
     setFlash({ kind, key: Date.now() });
@@ -327,7 +328,12 @@ export function MemorySection() {
   }, []);
 
   const reloadExtractions = useCallback(async () => {
-    setExtractions(await fetchExtractions());
+    setRefreshingExtractions(true);
+    try {
+      setExtractions(await fetchExtractions());
+    } finally {
+      setRefreshingExtractions(false);
+    }
   }, []);
 
   useEffect(() => {
@@ -923,11 +929,14 @@ export function MemorySection() {
             type="button"
             className="ghost"
             onClick={() => void reloadExtractions()}
+            disabled={refreshingExtractions}
             title={t('settings.memoryExtractionsRefresh')}
           >
             <Icon name="refresh" size={12} />{' '}
             <span style={{ marginLeft: 4 }}>
-              {t('settings.memoryExtractionsRefresh')}
+              {refreshingExtractions
+                ? t('settings.memoryExtractionsRefreshing')
+                : t('settings.memoryExtractionsRefresh')}
             </span>
           </button>
         </div>
--- a/apps/web/src/i18n/locales/en.ts
+++ b/apps/web/src/i18n/locales/en.ts
@@ -1386,6 +1386,7 @@ export const en: Dict = {
   'settings.memoryExtractionsHint': 'Recent extraction attempts. Heuristic regex extraction runs first; LLM extraction runs in the background after each turn.',
   'settings.memoryExtractionsEmpty': 'No extractions yet. The next chat turn will populate this list.',
   'settings.memoryExtractionsRefresh': 'Refresh',
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
   'settings.memoryExtractionPhaseRunning': 'Running…',
   'settings.memoryExtractionPhaseSuccess': 'Success',
   'settings.memoryExtractionPhaseSkipped': 'Skipped',
--- a/apps/web/src/i18n/locales/en.ts
+++ b/apps/web/src/i18n/locales/en.ts
@@ -1386,6 +1386,8 @@ export const en: Dict = {
   'settings.memoryExtractionsHint': 'Recent extraction attempts. Heuristic regex extraction runs first; LLM extraction runs in the background after each turn.',
   'settings.memoryExtractionsEmpty': 'No extractions yet. The next chat turn will populate this list.',
   'settings.memoryExtractionsRefresh': 'Refresh',
+
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
   'settings.memoryExtractionsRefreshing': 'Refreshing…',
   'settings.memoryExtractionPhaseRunning': 'Running…',
   'settings.memoryExtractionPhaseSuccess': 'Success',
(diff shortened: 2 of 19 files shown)


apps/web/src/i18n/locales/en.ts
```diff
@@ -1386,6 +1386,7 @@ export const en: Dict = {
   'settings.memoryExtractionsHint': 'Recent extraction attempts. Heuristic regex extraction runs first; LLM extraction runs in the background after each turn.',
   'settings.memoryExtractionsEmpty': 'No extractions yet. The next chat turn will populate this list.',
   'settings.memoryExtractionsRefresh': 'Refresh',
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
```
lefarcen, line 1391: P1 This new i18n key is not declared in `apps/web/src/i18n/types.ts`, and once it is added there every `Record<Locale, Dict>` locale must provide the same key. As-is, the `t('settings.memoryExtractionsRefreshing')` call is not a valid `keyof Dict` and the English object also has an excess property, so typecheck will fail. Please add `settings.memoryExtractionsRefreshing` to `Dict` and all locale files (fallback English text is fine where translations are not available yet).

lefarcen (changes requested):
Hey @xxiaoxiong, thanks for picking this up. The UI direction is exactly right for #1418: making the refresh action visibly in-flight and preventing repeated clicks is the behavior gap we need to close.

One TypeScript/i18n integration issue needs fixing before this can land — see the inline note. Once the new string is wired through the shared dictionary and locale files, this should be in good shape.

apps/web/src/components/MemorySection.tsx
```diff
@@ -923,11 +929,14 @@ export function MemorySection() {
...
             className="ghost"
             onClick={() => void reloadExtractions()}
+            disabled={refreshingExtractions}
             title={t('settings.memoryExtractionsRefresh')}
           >
             <Icon name="refresh" size={12} />{' '}
             <span style={{ marginLeft: 4 }}>
-              {t('settings.memoryExtractionsRefresh')}
+              {refreshingExtractions
+                ? t('settings.memoryExtractionsRefreshing')
```
lefarcen, line 938: P1 This key still is not part of `Dict` in `apps/web/src/i18n/types.ts`, so `t('settings.memoryExtractionsRefreshing')` is not a valid translation key and the web typecheck remains blocked. Please add `'settings.memoryExtractionsRefreshing': string;` next to `settings.memoryExtractionsRefresh` in the shared `Dict` type.

apps/web/src/i18n/locales/en.ts
```diff
@@ -1386,6 +1386,9 @@ export const en: Dict = {
   'settings.memoryExtractionsHint': 'Recent extraction attempts. Heuristic regex extraction runs first; LLM extraction runs in the background after each turn.',
   'settings.memoryExtractionsEmpty': 'No extractions yet. The next chat turn will populate this list.',
   'settings.memoryExtractionsRefresh': 'Refresh',
+
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
+  'settings.memoryExtractionsRefreshing': 'Refreshing…',
```
lefarcen, line 1391: P1 The English dictionary now declares `settings.memoryExtractionsRefreshing` twice, which TypeScript rejects as a duplicate object literal property. Please keep a single entry here after adding the key to `Dict`.

lefarcen (changes requested):
Hey @xxiaoxiong, thanks for the quick update. The locale coverage is moving in the right direction, but the type-level wiring is still incomplete and the English dictionary picked up a duplicate key.

I’m keeping this blocked for now with the remaining i18n/typecheck issues below. Once those are cleaned up, I’ll re-check the new head.

apps/web/src/i18n/locales/en.ts
PerishCode, line 1391: Two related blockers that need to land together before the web package typechecks. Both are also flagged in the still-open threads from @lefarcen (#discussion_r3230732667, #discussion_r3230732671) and from @chatgpt-codex-connector (#discussion_r3230510160), which is why CI's Validate workspace and build jobs are still red on this head.

**1. Duplicate property here.** Lines 1390 and 1391 both declare `'settings.memoryExtractionsRefreshing': 'Refreshing…'`. TypeScript rejects this with TS1117 (`An object literal cannot have multiple properties with the same name`) under the strict config the web package uses, so `apps/web` will not compile. Looks like the second commit re-added the key already present from the first commit instead of replacing it.

**2. Missing `Dict` declaration.** The companion `Dict` interface at `apps/web/src/i18n/types.ts:402` only lists `'settings.memoryExtractionsRefresh': string;` — `'settings.memoryExtractionsRefreshing'` was never added. That makes the locale objects fail the structural check against `Dict` (excess property in `en`, missing in all other locales) and also makes `t('settings.memoryExtractionsRefreshing')` in `apps/web/src/components/MemorySection.tsx:939` not a valid `keyof Dict`. The commit message for `4da37b87` claims the addition was made, but the actual patch does not touch `apps/web/src/i18n/types.ts`.

**Suggested change:**

- In `apps/web/src/i18n/types.ts`, add the declaration next to its sibling so `Dict` covers both:

```ts
  'settings.memoryExtractionsRefresh': string;
  'settings.memoryExtractionsRefreshing': string;
```

- In this file, delete one of the two duplicate lines (and drop the stray blank at line 1389) so the entry sits flush with its neighbours:

```ts
  'settings.memoryExtractionsRefreshing': 'Refreshing…',
  'settings.memoryExtractionPhaseRunning': 'Running…',
```

The existing English-fallback strings already added to the other 17 locale files are fine once `Dict` declares the key, so no additional locale edits are required.


PerishCode (commented):
@xxiaoxiong The two prior `lefarcen` blockers haven't been resolved on this head: `apps/web/src/i18n/types.ts` was not actually modified (despite the second commit's message), and `apps/web/src/i18n/locales/en.ts` now declares `settings.memoryExtractionsRefreshing` twice. CI reflects both — Validate workspace and build are red. Consolidated fix in the inline comment.



lefarcen:
Hey @xxiaoxiong, thanks again for working on this Memory refresh feedback fix.

I’m going to close this PR as superseded: PR #1477 has already merged the same user-facing fix for #1418, and that issue is now closed. This branch also still has the unresolved i18n/typecheck blockers we discussed, so keeping it open would just create duplicate review traffic.

Appreciate you jumping on the issue — the direction was right, and we’ll keep the merged path in #1477 as the source of truth.

PR #1487, opened 2026-05-12
fix: add success and error feedback for Open folder action

xxiaoxiong:
Fixes #1408

The Open folder button now provides clear feedback after clicking:

## Success case
- Shows success toast with the opened folder path
- Toast auto-dismisses after 3 seconds
- Clears any previous error state

## Error case
- Catches exceptions from `onImportFolder`
- Shows error toast with the error message
- Toast auto-dismisses after 6 seconds

## Loading state
- Button shows "Opening…" text while processing
- Clears previous success/error state before starting

This makes the Open folder workflow reliable and gives users clear confirmation that their action succeeded or failed.

## Testing

1. Enter a valid folder path and click Open folder
   - ✅ Success toast appears
2. Enter an invalid path and click Open folder
   - ✅ Error toast appears
3. Click Open folder while it's processing
   - ✅ Button is disabled and shows "Opening…"

### DIFF
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -100,7 +100,7 @@ interface Props {
   // input and the renderer POSTs `/api/import/folder` itself. Browser
   // builds have no `shell.openPath` surface, so the renderer naming a
   // path here cannot escalate (PR #974 trust model).
-  onImportFolder?: (baseDir: string) => Promise<void> | void;
+  onImportFolder?: (baseDir: string) => Promise<boolean> | boolean;
   // Electron flow: the desktop main process owns the picker dialog and
   // the import call atomically (`pickAndImport` IPC). The renderer
   // never sees the path or the HMAC token; it only receives the
@@ -176,7 +176,6 @@ export function NewProjectPanel({
   const [importFolderError, setImportFolderError] = useState<
     { message: string; details?: string } | null
   >(null);
-  const [importFolderSuccess, setImportFolderSuccess] = useState<string | null>(null);
   const [tab, setTab] = useState<CreateTab>('prototype');
   const tabsRef = useRef<HTMLDivElement | null>(null);
   const [tabScroll, setTabScroll] = useState({ left: false, right: false });
@@ -536,10 +535,17 @@ export function NewProjectPanel({
     if (!trimmed) return;
     setImportingFolder(true);
     setImportFolderError(null);
-    setImportFolderSuccess(null);
     try {
-      await onImportFolder(trimmed);
-      setImportFolderSuccess(`Folder opened successfully: ${trimmed}`);
+      const success = await onImportFolder(trimmed);
+      if (success) {
+        // Success: the parent will navigate away, so this component
+        // will unmount before the toast can be shown. No need to set
+        // success state here.
+      } else {
+        setImportFolderError({
+          message: `Failed to open folder: ${trimmed}`,
+        });
+      }
     } catch (err) {
       setImportFolderError({
         message: `Open folder failed: ${err instanceof Error ? err.message : 'unknown error'}`,
@@ -792,14 +798,6 @@ export function NewProjectPanel({
           details={importFolderError.details ?? null}
           ttlMs={6000}
           onDismiss={() => setImportFolderError(null)}
-        />
-      ) : null}
-      {importFolderSuccess ? (
-        <Toast
-          message={importFolderSuccess}
-          details={null}
-          ttlMs={3000}
-          onDismiss={() => setImportFolderSuccess(null)}
         />
       ) : null}
     </div>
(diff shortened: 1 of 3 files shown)


apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -534,8 +535,15 @@ export function NewProjectPanel({
     const trimmed = baseDir.trim();
     if (!trimmed) return;
     setImportingFolder(true);
+    setImportFolderError(null);
+    setImportFolderSuccess(null);
     try {
       await onImportFolder(trimmed);
+      setImportFolderSuccess(`Folder opened successfully: ${trimmed}`);
```
lefarcen: P1 This shows success even when the import failed. `onImportFolder` is backed by `handleImportFolder`, which calls `importFolderProject`; that helper returns `null` for non-OK responses and caught network errors, and `handleImportFolder` then returns without throwing. As a result, an invalid path can still reach this line and display “Folder opened successfully”. Please make the import path signal failure explicitly (for example, throw a structured error from `importFolderProject`/`handleImportFolder` or return a success boolean/result) and only set the success toast after a real import response.

lefarcen (changes requested):
Hi @xxiaoxiong! Thanks for jumping on the Open folder feedback gap — the loading state and toast direction map well to what #1408 is asking for. I found one correctness issue in the web fallback path that needs fixing before this can land; see the inline note below.

apps/web/src/components/NewProjectPanel.tsx
Siri-Ray: Please avoid treating completion of `onImportFolder(trimmed)` as proof that the folder opened. In the web path this prop is `App.handleImportFolder`, which calls `importFolderProject({ baseDir })` and then returns normally when that helper returns `null`; `importFolderProject` returns `null` for non-OK daemon responses and caught network errors. That means an invalid path or daemon/import failure can still reach this new line and show `Folder opened successfully`, while the `catch` here never runs. On the true success path, `handleImportFolder` also navigates before resolving, which unmounts `NewProjectPanel`, so this local success state is not a reliable place to show the confirmation toast. Please make the import flow signal a structured success/failure result or throw on failure, and render the success/error feedback from a component that remains mounted after navigation; only set the success message after an actual `ImportFolderResponse` is available.

Siri-Ray (changes requested):
@xxiaoxiong thanks for taking on the Open folder feedback path. The loading-state direction is helpful, but the new success/error feedback is still tied to a callback contract that cannot distinguish failed web imports and also unmounts this panel on successful imports, so I’m requesting changes on that correctness issue.


apps/web/src/components/NewProjectPanel.tsx
lefarcen: P1 This still treats `await onImportFolder(trimmed)` as a successful import, but the web callback returns normally when `importFolderProject` returns `null` for non-OK responses or network errors. Invalid paths can therefore show the success toast, and real successes navigate/unmount this panel before this local toast is reliable. Please change the import contract to return/throw a structured result and only set success after an actual `ImportFolderResponse`, with error feedback driven by the failure result.

lefarcen (changes requested):
Hey @xxiaoxiong, I re-checked the current head after the latest review activity. The UI direction is still the right one, but the web import callback still cannot distinguish a failed import from a successful one, so the previous correctness blocker remains. Siri-Ray's review is pointing at the same path; the inline note below is the remaining must-fix from my side as well.

lefarcen (approved):
Hey @xxiaoxiong, thanks for the quick follow-up. The web path now returns an explicit success boolean, so failed imports no longer fall through into a false success toast, and successful imports continue into the project navigation flow. That addresses the blocker I raised on the previous head.

Siri-Ray still has the qualified reviewer thread on this PR, so please wait for their re-check as well, but my blocking review is cleared now.

lefarcen (approved):
Hey @xxiaoxiong, thanks for the quick follow-up. The web path now returns an explicit success boolean, so failed imports no longer fall through into a false success toast, and successful imports continue into the project navigation flow. That addresses the blocker I raised on the previous head.

Siri-Ray still has the qualified reviewer thread on this PR, so please wait for their re-check as well, but my blocking review is cleared now.

lefarcen:
Hey @Siri-Ray, could you re-check this one when you have a chance? @xxiaoxiong pushed an update after your blocking review: the web import path now returns an explicit success boolean, CI is green, and my previous blocker is cleared on the current head.

lefarcen:
Hey @Siri-Ray, just a gentle bump on this one — @xxiaoxiong's latest push addressed the callback return value issue you flagged, and CI is still green on the current head. Let me know if you need anything else to clear your review. 🙏

lefarcen:
Quick heads-up: #1666 is also open by you against the same area in `NewProjectPanel.tsx`, both targeting #1408. The newer PR is a more focused +8/-0 change vs the +19/-4 here. Since you're the author of both, worth deciding which to move forward — if #1666 is the intended replacement, closing this one would help the pool reviewers focus. Happy to help scope if useful.

xxiaoxiong:
Closing this PR in favor of #1666 which is a more focused implementation (+8/-0 vs +19/-4 here). #1666 achieves the same goal with a cleaner approach.

PR #1490, opened 2026-05-13
fix: close Settings modal when opening project from Routines

xxiaoxiong:
Fixes #1355

When clicking **Open project** in Routine history, the Settings modal now closes automatically, providing a clean transition to the project.

## Problem

Previously, clicking "Open project" would navigate to the project but leave the Settings modal open, creating a confusing state where both the modal and the project were visible.

## Solution

- Added `onClose` prop to `RoutinesSection` component
- Call `onClose()` after `navigate()` in the Open project button handler
- Pass `onClose` from `SettingsDialog` to `RoutinesSection`

This ensures the modal dismisses after navigation, consistent with other modal-based navigation actions in the app.

## Testing

1. Open Settings
2. Go to Routines section
3. Find a routine with history
4. Click "Open project"
5. ✅ Settings modal closes automatically
6. ✅ Project opens cleanly without modal overlay

### DIFF
--- a/apps/web/src/components/RoutinesSection.tsx
+++ b/apps/web/src/components/RoutinesSection.tsx
@@ -345,9 +345,10 @@ function RunHistory({ routineId, refreshKey }: { routineId: string; refreshKey: 
           <button
             type="button"
             className="routines-history-link"
-            onClick={() =>
-              navigate({ kind: 'project', projectId: r.projectId, fileName: null })
-            }
+            onClick={() => {
+              navigate({ kind: 'project', projectId: r.projectId, fileName: null });
+              onClose?.();
+            }}
             title="Open the project this run wrote to"
           >
             Open project
@@ -359,7 +360,7 @@ function RunHistory({ routineId, refreshKey }: { routineId: string; refreshKey: 
   );
 }
 
-export function RoutinesSection() {
+export function RoutinesSection({ onClose }: { onClose?: () => void }) {
   const [routines, setRoutines] = useState<Routine[]>([]);
   const [projects, setProjects] = useState<ProjectSummary[]>([]);
   const [loading, setLoading] = useState(true);
--- a/apps/web/src/components/RoutinesSection.tsx
+++ b/apps/web/src/components/RoutinesSection.tsx
@@ -309,7 +309,7 @@ function ScheduleEditor({
   );
 }
 
-function RunHistory({ routineId, refreshKey }: { routineId: string; refreshKey: number }) {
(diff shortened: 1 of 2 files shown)


apps/web/src/components/RoutinesSection.tsx
```diff
@@ -345,9 +345,10 @@ function RunHistory({ routineId, refreshKey }: { routineId: string; refreshKey:
           <button
             type="button"
             className="routines-history-link"
-            onClick={() =>
-              navigate({ kind: 'project', projectId: r.projectId, fileName: null })
-            }
+            onClick={() => {
+              navigate({ kind: 'project', projectId: r.projectId, fileName: null });
+              onClose?.();
```
lefarcen, line 350: P1 `onClose` is not in scope inside `RunHistory` — the PR only adds it to `RoutinesSection`, but this nested component still destructures only `{ routineId, refreshKey }`, and the `<RunHistory ... />` call does not pass `onClose`. This will fail before the modal can close. Please add `onClose?: () => void` to `RunHistory`'s props and pass the parent `onClose` through where `RunHistory` is rendered.

lefarcen (changes requested):
Hi @xxiaoxiong, thanks for picking this one up. The intended UX direction is right — threading the Settings close handler into the Routines path is exactly the shape we want for this modal-to-project transition.

There is one correctness blocker in the current patch: the nested run-history component calls `onClose` without receiving it, so this will fail at compile/runtime instead of closing Settings. Once that prop is passed through, the fix should be in good shape.

lefarcen (approved):
Hey @xxiaoxiong, nice follow-up — the `onClose` callback is now threaded all the way from `SettingsDialog` into `RunHistory`, so the earlier blocker is addressed.

I re-read the current diff and don’t see any remaining correctness issues in this scoped fix. This approval clears my previous changes-requested review; CI / the requested human review can continue from here.

PerishCode (approved):
@xxiaoxiong Nice tight follow-up — `onClose` is now threaded all the way from `SettingsDialog` → `RoutinesSection` → `RunHistory`, the `Open project` handler calls `navigate(...)` then `onClose?.()`, and the optional-chained signature keeps `RunHistory`/`RoutinesSection` safe to use without the prop. Verified the fix at `c09e1be2` against the earlier P1 (onClose-not-in-scope inside `RunHistory`) — that's resolved, the change stays scoped to the Settings → Routines → Open project transition, and CI is green. Thanks for picking this one up and turning the review feedback around so quickly!

PR #1493, opened 2026-05-13
fix(web): tighten routines project radios

prantikmedhi:
## Summary
- Tighten the Routines "Project" radio-row layout so the control and copy read as one unit
- Reduce horizontal padding and switch the rows to a small two-column grid for cleaner alignment

## Verification
- `git diff --check`
- `pnpm --filter web test -- apps/web/tests/components/RoutinesSection.test.tsx` *(blocked by missing `@tailwindcss/postcss` in the local test environment)*

Closes #1430

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -18301,9 +18301,10 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
   padding: 0 6px;
 }
 .routines-radio {
-  display: flex;
-  gap: 10px;
-  padding: 10px 12px;
+  display: grid;
+  grid-template-columns: 16px minmax(0, 1fr);
+  gap: 8px 10px;
+  padding: 10px 8px;
   border-radius: var(--radius);
   cursor: pointer;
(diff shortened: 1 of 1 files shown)


lefarcen (commented):
Hi @prantikmedhi! This is a focused follow-up for #1430, and the direction looks right: the fixed radio column plus tighter row padding makes the control and copy read as one unit, while `min-width: 0` keeps the text column resilient.

I did a CSS-only pass over the changed `routines-radio` rules and the surrounding Routines markup, and I don't see any blockers from the maintainer side. CI is still pending, so let's let that finish and the routed human review can take the final merge call.

Siri-Ray (approved):
@prantikmedhi I reviewed the focused CSS change in `apps/web/src/index.css` against the surrounding Routines markup. The grid-based radio row, tighter padding, and `min-width: 0` on the copy column line up with the issue goal without introducing a layout or interaction regression in the changed range. Thanks for the tidy, focused fix.

PR #1485, opened 2026-05-12
feat(daemon): Critique Theater Phase 12 (9 Prometheus metrics + 6 log events + OTel span + Grafana dashboard)

Nagendhra-web:
<div align="center">

# 🎭 Critique Theater · Phase 12 (Observability)

**Turns the critique pipeline from "running silently" into "metered, logged, dashboarded, traced".**

phase
metrics
logs
otel
stacked-on

</div>

> **TL;DR.** Adds nine Prometheus counters / histograms / gauges under `open_design_critique_*`, six structured log events on the `critique` namespace, an OTel `critique.run` outer span with `final_status` / `final_composite` / `adapter` / `skill` attributes, a three-row Grafana dashboard JSON, and a `/api/metrics` route the operator can mount behind their existing auth proxy. Every metric bump and log line fires from inside the orchestrator loop the wireup PR (#1338) just activated. Zero overhead until an operator wires an SDK / scrape target / dashboard panel.

**Stacks on:** the wireup PR (#1338). Cannot file earlier; without orchestrator wiring every metric reads zero and every dashboard panel sits flat.

---

## 🤔 Why

The Critique Theater stack now actually routes traffic (wireup PR #1338 flips the orchestrator gate from env-only to the four-tier resolver). That means production runs are about to happen for real. Without instrumentation the on-call surface is "the user said it was slow" or "something broke, look at logs maybe", which is the state every prior multi-round agent pipeline got stuck in before someone retrofitted observability the hard way.

Three concrete things this PR closes before the first M2 user lands:

1. **No quality SLI.** "Composite score distribution per adapter" is the single most useful number to put on a dashboard for a critique pipeline. Without it, an adapter that drifts from `composite p50 = 8.2` to `composite p50 = 6.8` over a week is invisible until a user complains. Adding `composite_score` as a histogram by `adapter / skill` is one line per `round_end`; the dashboard panel that consumes it is a row of three time series.
2. **No "this adapter is broken" alarm.** The wireup PR's resolver routes around adapters marked degraded for 24h, but operators need to see WHICH adapters are getting marked. `degraded_total{reason, adapter}` and `parser_errors_total{kind, adapter}` are exactly the two series the rollout ratchet (Phase 16) consumes. They also feed an obvious page-the-on-call alert rule once the first 1000 production runs settle the baseline.
[...]

PR #1499, opened 2026-05-13
feat(daemon): Critique Theater Phase 16 (M-phase rollout ratchet + /api/critique/conformance)

Nagendhra-web:
<div align="center">

# 🎭 Critique Theater · Phase 16 (M-phase rollout ratchet)

**Surfaces the M-phase rollout recommendation off the conformance harness's daily output so the operator does not have to read a Grafana panel to decide whether to promote `OD_CRITIQUE_ROLLOUT_PHASE`.**

phase
evaluator-tests
history-tests
surface

</div>

> **TL;DR.** Pure `evaluateRollout(phase, window)` function that reads a rolling window of `ConformanceDay` rows and returns `hold` / `promote` / `demote`, plus a JSON-lines `conformance-history` writer the nightly cron uses, plus the `GET /api/critique/conformance` route an operator polls to get the recommendation. The ratchet does NOT auto-flip `OD_CRITIQUE_ROLLOUT_PHASE`; that's the next PR's call.

**Stacks on:** Phase 12 (#1485). Cannot file earlier; the ratchet reads from the conformance harness's daily output, and that surface only exists once Phase 12 lands. Once #1485 merges, this PR's diff collapses to the six files actually edited here.

---

## 🤔 Why

Phase 15's spec says: "M3: After 14 consecutive days at ≥ 90% adapter conformance across the fleet, flip the global default to true." There was no machinery to compute that. The conformance harness existed (Phase 10), the dashboard existed (Phase 12), but nothing watched the dashboard and nothing surfaced the recommendation. The operator had to either:

1. Read the Grafana adapter-health panel every morning and do the rolling-window math by eye.
2. Wait for someone to ping them when conformance regressed.

Both are how rollouts get stuck. The first is too tedious to keep up week-over-week and people drift; the second has no clear "this adapter regressed three days ago" signal until the next user-visible failure.

Three concerns shaped this PR:

1. **Pure evaluator, not a daemon.** The `ratchet.ts` decision function is a pure `(phase, window) => decision` with no I/O, no clock dependency, no global state. The history writer and the HTTP route are separate concerns. That keeps the matrix test (every cell of "hold at M0 / M1 / M2 / M3 with various window shapes") trivial to pin and the decision auditable in a single function.
2. **Recommendation surface, not auto-flip.** The ratchet returns a recommendation. An operator-driven follow-up (Phase 17 or similar) consumes `GET /api/critique/conformance` and either flips the env var in a deploy pipeline or alerts a human. That step is environment-specific (CI flow, who has perms to flip env vars in prod, what the alert routing looks like) and intentionally out of v1 scope.
3. **Demote floor lower than promote threshold.** A single noisy day at 0.88 should not bounce the rollout back to a prior phase. The demote signal is for SUSTAINED breakage. The matrix has explicit cases for "one noisy day stays at hold", "three noisy days demotes".

---

## 👀 What users will see

**End users:** nothing. The ratchet's output flows into operator decisions about rollout phases; the user sees the rollout state through the same `OD_CRITIQUE_ROLLOUT_PHASE` env var that already exists.

**Operators after the nightly cron runs and the route serves data:**

- `GET /api/critique/conformance` returns `{ phase, decision, window: [{date, fleetShippedRate, adapterCount, ...}, ...] }`. The `decision` field is one of `hold` / `promote` / `demote` with an explicit reason naming WHY (passing-day count, threshold band, observed-day count).
- A nightly cron job (the `.github/workflows/critique-conformance.yml` workflow) drives the synthetic adapters through the harness, writes a `ConformanceDay` row into `<dataDir>/conformance/<adapter>/<date>.jsonl`, and POSTs nothing; the next operator request to `/api/critique/conformance` reads from disk.
- After 14 consecutive days at ≥ 90% conformance the operator sees `{ decision: 'promote', from: 'M2', to: 'M3', evidenceDays: 14 }` and runs whatever pipeline they have for flipping `OD_CRITIQUE_ROLLOUT_PHASE=M3`. The ratchet does not flip it for them. The recommendation is the surface; the flip is the operator's call.

**The decision matrix in plain English:**

- Every day in the rolling window cleared both promote thresholds → `promote` (or `hold` if already at M3).
- At least one day's fleet shipped-rate fell below the demote floor → `demote` (or `hold` if already at M0).
- Some days missing or near-miss → `hold` with an explicit reason.

---

## 🎯 Surface area

- [x] Daemon (`apps/daemon/src/critique/ratchet.ts`, new): pure `evaluateRollout` decision function (229 lines including the docblock matrix).
- [x] Daemon (`apps/daemon/src/critique/conformance-history.ts`, new): JSON-lines persistence at `<dataDir>/conformance/<adapter>/<date>.jsonl`.
- [x] Daemon (`apps/daemon/src/critique/__fixtures__/run-nightly.ts`, new): drives synthetic adapters through the harness for the cron.
- [x] Daemon (`apps/daemon/src/server.ts`): new `GET /api/critique/conformance` route.
- [x] **API / contract**: new `GET /api/critique/conformance` endpoint returning `{ phase, decision, window: ConformanceDay[] }`.
- [x] **CLI / env var**: no new env vars; existing `OD_CRITIQUE_ROLLOUT_PHASE` is what the operator flips based on the recommendation.
- [x] **New top-level dependency**: `prom-client ^15.1.0` and `@opentelemetry/api ^1.9.0` carried over from the upstream Phase 12 daemon manifest (this PR's diff against `main` shows them because Phase 12 isn't merged yet; once #1485 lands the diff here collapses).
- [x] CI workflow (`.github/workflows/critique-conformance.yml`, new): nightly cron drives the harness, writes the `ConformanceDay` rows the route consumes.
- [ ] Web runtime (`apps/web`)
- [ ] E2E (`e2e`)
- [ ] **UI** / **Keyboard shortcut** / **i18n keys**
- [ ] **Default behavior change**: no. New endpoint and new cron, both opt-in via operator action. Existing runs are byte-identical.

---

## 📋 Five surfaces

```
apps/daemon/src/critique/ratchet.ts                    + 229   pure evaluator
apps/daemon/src/critique/conformance-history.ts        + 133   JSON-lines persistence
apps/daemon/src/server.ts                              +  33   /api/critique/conformance route
.github/workflows/critique-conformance.yml             +  73   nightly cron
apps/daemon/tests/critique-ratchet.test.ts             + 160   10 cases
apps/daemon/tests/critique-conformance-history.test.ts + 144    7 cases
```

### 1. `ratchet.ts`, the decision function

Pure. Takes the current `RolloutPhase` plus a rolling window of `ConformanceDay` rows and returns one of three decisions:

```ts
type RatchetDecision =
  | { kind: 'hold';    current: RolloutPhase; reason: string; passingDays: number; observedDays: number }
  | { kind: 'promote'; from: RolloutPhase; to: RolloutPhase; evidenceDays: number }
  | { kind: 'demote';  from: RolloutPhase; to: RolloutPhase; reason: string };
```

Decision matrix:

| Window condition | Decision |
|---|---|
| Every day in the window cleared both thresholds | `promote` (or `hold` at M3) |
| At least one day's fleet shipped-rate fell below the demote floor (half the promote threshold) | `demote` (or `hold` at M0) |
| Some days missing, or near-miss | `hold` with an explicit reason |

Validation at entry rejects non-positive `windowDays`, out-of-range thresholds, NaN, etc., returning `hold` with the explicit reason so a malformed config never silently promotes.

### 2. `conformance-history.ts`, the JSON-lines writer

JSON-lines at `<dataDir>/conformance/<adapter>/<date>.jsonl`. One file per adapter per day; appends a `ConformanceDay` row per run. The route reads the rolling window by scanning the recent date files for each adapter and aggregating.

### 3. `__fixtures__/run-nightly.ts`, the cron driver

Drives the synthetic adapter set through the conformance harness (Phase 10's `runAdapterConformance`), writes the `ConformanceDay` row for each, and exits. The GitHub Actions cron in `critique-conformance.yml` invokes it.

### 4. `server.ts` route

`GET /api/critique/conformance` returns the current phase, the decision, and the window. No auth surface beyond what the daemon HTTP server already exposes.

### 5. Test coverage

- `tests/critique-ratchet.test.ts` (10 cases): the full decision matrix (hold / promote / demote at each phase with various window shapes) plus the validation edges.
- `tests/critique-conformance-history.test.ts` (7 cases): write / read round-trip, per-adapter directory layout, jsonl parsing edges.

---

## 📸 Screenshots

Intentionally none. The change is operator-facing (the cron writes files, the route returns JSON). The visible surface is the JSON payload, which the validation table below covers.

---

## ✅ Validation

Commands run on the current head (`f2ce497d`, after the lockfile + nix `pnpmDepsHash` fix for the inherited Phase 12 deps):

| Command | Result |
|---|---|
| `pnpm --filter @open-design/daemon test tests/critique-ratchet.test.ts` | 10 / 10 (decision matrix + validation edges) |
| `pnpm --filter @open-design/daemon test tests/critique-conformance-history.test.ts` | 7 / 7 (write/read round-trip, layout, jsonl parsing) |
| `pnpm --filter @open-design/daemon test tests/critique-conformance.test.ts` | 5 / 5 (upstream Phase 10 suite still green) |
| `pnpm install --no-frozen-lockfile` + `git diff --stat pnpm-lock.yaml` | 27 lines added for `prom-client` + `@opentelemetry/api` (inherited from Phase 12) |
| Nix `pnpmDepsHash` pinned to `sha256-NtXbiRU0YZ4EVJVNC6N3sR1S0ozA3BvCwgXI0L0OMH4=` | matches the value CI's nix-check produced on the sibling Phase 12 branch (same lockfile content, same hash) |

---

## ⛔ What this PR does not do

- **No auto-flip of `OD_CRITIQUE_ROLLOUT_PHASE`.** Surfaces a recommendation only. Auto-flip is its own focused follow-up (operator-pipeline-specific, out of v1 scope).
- **No 4th Grafana row.** The ratchet view (decision + rolling window) is a follow-up dashboard panel that consumes `/api/critique/conformance`; intentionally not bundled here so this PR stays the evaluator + history + route.
- **No production-adapter sweep in the cron.** The nightly workflow drives the synthetic adapter set; pointing it at real production adapters is the operator's call once `ConformanceDay` shapes settle.


apps/daemon/src/server.ts
```diff
@@ -69,6 +69,16 @@ import { runOrchestrator } from './critique/orchestrator.js';
 import { createRunRegistry } from './critique/run-registry.js';
 import { handleCritiqueInterrupt } from './critique/interrupt-handler.js';
 import { handleCritiqueArtifact } from './critique/artifact-handler.js';
+import { getCritiqueMetrics, register } from './metrics/index.js';
+import { readConformanceHistory } from './critique/conformance-history.js';
+import { evaluateRollout } from './critique/ratchet.js';
+import { parseRolloutPhase } from './critique/rollout.js';
```
lefarcen: P1 This imports `parseRolloutPhase` twice in the same module scope: once here and again in the grouped import below. That creates a duplicate binding and will stop `server.ts` from compiling/parsing. Please remove this standalone import and keep the existing grouped `rollout.js` import.

.github/workflows/critique-conformance.yml
```diff
@@ -0,0 +1,73 @@
...
+      # adapter sweep lives in a follow-up that wires every production
+      # agent CLI into the harness; here we drive the synthetic-good
+      # and synthetic-bad fixtures so the workflow proves the cron
+      # plumbing end-to-end (install + build + harness invocation +
+      # history write) without depending on every adapter binary
+      # being installable in the CI image.
+      - name: Run conformance harness (synthetic adapters)
+        run: |
+          node --experimental-strip-types \
+            apps/daemon/src/critique/__fixtures__/run-nightly.ts \
```
lefarcen: P1 The workflow calls `apps/daemon/src/critique/__fixtures__/run-nightly.ts`, but this PR does not add that file. Because the command is followed by `|| echo`, the nightly job would still report success while writing no `.od/conformance/` history, so `/api/critique/conformance` never gets real evidence. Please add the runner (or call an existing harness entrypoint) and let failures fail the job instead of masking them.

apps/daemon/src/critique/ratchet.ts
```diff
@@ -0,0 +1,229 @@
...
+ *   * `hold`: anything else (data gaps, near-misses, or M3 with no
+ *     headroom).
+ *
+ * The function is total: it always returns a decision; M0 cannot demote
+ * (already at the floor) and M3 cannot promote (already at the ceiling).
+ * Both of those collapse to `hold` with a reason-string that explains
+ * why.
+ */
+export function evaluateRollout(params: EvaluateRolloutParams): RatchetDecision {
+  const windowDays = params.windowDays ?? 14;
```
lefarcen: P1 `windowDays` is accepted without validation, so the new endpoint can pass `?windowDays=0` (or a negative value) and `passingDays >= windowDays` becomes true with zero observed days, returning `promote` with no evidence. Please clamp/reject non-positive windows (and keep thresholds in the 0..1 range) before evaluating, ideally with a regression test through the endpoint or evaluator.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for carrying the ratchet piece through — the split between pure evaluation, append-only history, and an operator-facing recommendation endpoint is a good shape for this rollout. I found a few blockers before this can land: one compile-time import issue, one cron path that currently masks a missing runner, and one endpoint/evaluator validation gap that can produce a zero-evidence promotion. Once those are tightened up, this should be much safer to review again.

apps/daemon/src/server.ts
PerishCode: **Blocking — duplicate `parseRolloutPhase` import will fail ESM/TS parse**

`parseRolloutPhase` is imported here on line 75 and again in the grouped import on line 79, both from the same module `./critique/rollout.js`, so the module declares the same lexical binding twice. ESM (and `tsc --strict`) treats this as `SyntaxError: Identifier 'parseRolloutPhase' has already been declared` / `TS2300: Duplicate identifier`, which kills every daemon startup path that loads `server.ts` — not just the new `/api/critique/conformance` route. The PR body claims `typecheck` is clean, but the diff as posted cannot type-check.

Suggested fix: drop this standalone line and let the grouped import below keep `parseRolloutPhase`, e.g.

```ts
import { readConformanceHistory } from './critique/conformance-history.js';
import { evaluateRollout } from './critique/ratchet.js';
import {
  isCritiqueEnabled,
  parseEnvEnabled,
  parseRolloutPhase,
  type SkillCritiquePolicy,
} from './critique/rollout.js';
```

Please rerun `pnpm --filter @open-design/daemon typecheck` locally before re-pushing — this is plain to surface and the green-check claim in the PR body should match reality.


.github/workflows/critique-conformance.yml
```diff
@@ -0,0 +1,73 @@
...
+            || echo "synthetic-adapter sweep not wired yet (Phase 16 v1)"
```
PerishCode: **Blocking — nightly cron points at a non-existent runner and `|| echo` hides the failure**

The workflow calls `apps/daemon/src/critique/__fixtures__/run-nightly.ts`, but only `__fixtures__/adapters/synthetic-good.ts`, `__fixtures__/adapters/synthetic-bad.ts`, and `__fixtures__/v1/` exist in the PR (confirmed by `ls`); `run-nightly.ts` is not added anywhere. Because the command is suffixed with `|| echo "synthetic-adapter sweep not wired yet (Phase 16 v1)"`, `node` exits non-zero, the `|| echo` swallows it, the upload-artifact step silently skips (`if-no-files-found: ignore`), and `/api/critique/conformance` never accumulates evidence. That defeats the stated goal of this PR — the ratchet endpoint will perpetually return `hold` ("insufficient data") because no row ever lands under `.od/conformance/`. The PR body frames "production-adapter sweep is a follow-up" but synthetic adapters are explicitly in v1 scope; the missing runner is what breaks the v1 contract.

Concrete fix:
1. Add the runner in this PR — a small TS entrypoint that imports the two fixture adapters, drives the existing Phase 12 conformance harness (`runConformanceHarness` / equivalent), and calls `appendConformanceDay(dataDir, row)` for each result so the day's `.od/conformance/<adapter>/<date>.jsonl` is actually written.
2. Drop the `|| echo` so a missing runner or harness error fails the workflow loudly.
3. Optionally, add a final "verify history exists" step (`test -s .od/conformance/...` or `ls -R .od/conformance/ | grep .jsonl`) so a silent regression is caught at cron time, not when an operator reads the endpoint days later.


apps/daemon/src/critique/ratchet.ts
PerishCode: **Blocking — `windowDays<=0` (and out-of-range thresholds) produce a zero-evidence `promote`**

The evaluator falls through with no validation: `params.windowDays ?? 14` evaluates `0` as `0` (`??` only catches `null` / `undefined`), then `rollingWindowDates(now, 0)` produces an empty `windowDates` set, the aggregation loop is a no-op, and the gate on line 200 `if (passingDays >= windowDays)` reduces to `0 >= 0` = true → `nextPhase('M0' | 'M1' | 'M2')` returns a phase → the function returns `{ kind: 'promote', evidenceDays: 0 }`. Negative `windowDays` triggers the same path. The endpoint in `server.ts` passes the query value through `parseConformanceQuery` which only checks `Number.isFinite`, so `GET /api/critique/conformance?windowDays=0` (or `=-1`) lets a curl-call surface a "promote" recommendation with no evidence at all — the exact false-signal the rollout spec is trying to prevent. The thresholds are similarly unbounded: `?shippedThreshold=0&cleanParseThreshold=0` lets every day "pass".

Concrete fix:
- In `evaluateRollout`, clamp / reject up front:
  ```ts
  const windowDaysRaw = params.windowDays ?? 14;
  const windowDays = Math.max(1, Math.floor(Number.isFinite(windowDaysRaw) ? windowDaysRaw : 14));
  const clamp01 = (x: number, fallback: number) =>
    Number.isFinite(x) && x >= 0 && x <= 1 ? x : fallback;
  const shippedThreshold = clamp01(params.shippedThreshold ?? 0.9, 0.9);
  const cleanParseThreshold = clamp01(params.cleanParseThreshold ?? 0.95, 0.95);
  ```
- In `server.ts`, validate at the route boundary too and return `400 INVALID_INPUT` instead of silently coercing — operator misconfiguration should be loud, not laundered into a promote.
- Add regression cases to `tests/critique-ratchet.test.ts`: `windowDays = 0`, `windowDays = -1`, `shippedThreshold = 0`, `shippedThreshold = 1.5`, `cleanParseThreshold = NaN`. None of those exist today — only the happy-path 14-day window is exercised.


PerishCode (commented):
Three blocking issues confirmed by reading the code directly; two prior reviewers (lefarcen and Codex) flagged the same shape, so this is overlap rather than a separate round — the inline comments add the concrete fix suggestions and test-coverage gaps.

Top-line:
- `apps/daemon/src/server.ts:75` — `parseRolloutPhase` is imported twice from `./critique/rollout.js` in the same module scope. ESM / TS parse refuses to load `server.ts` at all, so every daemon startup path is broken, not just the new endpoint.
- `.github/workflows/critique-conformance.yml:60-64` — the cron points at `apps/daemon/src/critique/__fixtures__/run-nightly.ts`, which this PR does not add; `|| echo` masks the failure and the upload step ignores the empty `.od/conformance/`, so the ratchet endpoint never sees real evidence.
- `apps/daemon/src/critique/ratchet.ts:138` — `evaluateRollout` accepts `windowDays=0` (and negative) and thresholds outside `[0, 1]`; combined with the endpoint's pass-through query coercion, an operator can hit a `promote` recommendation with zero observed days.

Tests: no `windowDays=0` / negative-window / out-of-range-threshold regressions exist in `tests/critique-ratchet.test.ts`; please add them alongside the fix so the validation gate is pinned.



PerishCode (dismissed):
@Nagendhra-web Thank you for the fast and surgical turnaround. The three Phase 16 blockers from the previous round all landed cleanly in `d6582a83`:

- `apps/daemon/src/server.ts` — duplicate `parseRolloutPhase` import removed; only the grouped `rollout.js` import remains.
- `apps/daemon/src/critique/ratchet.ts` — `evaluateRollout` defends at its entry against non-positive `windowDays`, out-of-range thresholds, and (via `Number.isFinite`) `NaN`. The route handler also clamps query inputs through `parsePositiveInt` / `parseRate` before the evaluator sees them — belt-and-suspenders against any future caller that bypasses the route.
- `.github/workflows/critique-conformance.yml` — points at the now-present `apps/daemon/src/critique/__fixtures__/run-nightly.ts` runner via `node --import tsx`, with the `|| echo` mask gone so a thrown error fails the job loudly instead of uploading an empty history snapshot.

The `b9150fd2` follow-up adds the explicit `NaN` regression cases to `tests/critique-ratchet.test.ts` so a future refactor of the guard cannot accidentally let `NaN` through. Decision matrix is well-pinned across promote / hold / demote / out-of-window / weighted aggregate / NaN paths, and the conformance-history persistence suite covers append + retry-wins + malformed-line-drop + window-boundary + cross-adapter + non-finite rejection.

The v1 synthetic-only nightly limitation is explicitly called out in the PR body as a deliberate follow-up — proving the cron plumbing without binding CI to every production CLI is the right scope for this PR. Nice work, and thanks for the precise commit messages tying each fix back to the reviewer signal.



lefarcen (dismissed):
Hey @Nagendhra-web, thanks for the quick cleanup on this round. I re-read the current head and the three blockers from my previous review are addressed: the duplicate rollout import is gone, the nightly workflow now has a real TypeScript runner without the masked failure path, and the ratchet input guard is pinned with the zero/negative/out-of-range/NaN cases.

Approving from my side to clear my stale changes-requested review. This still is not merge-ready until the failing CI check is resolved, but I do not have any remaining code-review blockers on the Phase 16 changes.

apps/daemon/package.json
```diff
@@ -37,13 +37,15 @@
     "@open-design/platform": "workspace:*",
     "@open-design/sidecar": "workspace:*",
     "@open-design/sidecar-proto": "workspace:*",
+    "@opentelemetry/api": "^1.9.0",
```
lefarcen, line 40: P1 `@opentelemetry/api` and `prom-client` were added to `apps/daemon/package.json`, but `pnpm-lock.yaml` no longer contains matching daemon importer entries after the rebase. CI is failing at `pnpm install --frozen-lockfile` with `ERR_PNPM_OUTDATED_LOCKFILE`, so this branch cannot build. Please regenerate and commit the lockfile (for example with the repo-pinned `pnpm install`) so the lockfile records these two dependencies for `apps/daemon`.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for rebasing and trimming the conflict surface here. The previous Phase 16 logic blockers still look addressed, and I also requested the product reviewer since this is a feature-surface PR. One new blocker came in with the rebase/conflict-resolution push: the daemon package manifest and lockfile are out of sync, so CI cannot even install the workspace yet.

Also, before the next review round, please refresh the PR body against the current template — the Why / What users will see / Surface area / Validation sections are currently missing. That will make the product review and final maintainer pass much easier to scope.

apps/daemon/package.json
lefarcen, line 40: P1 This is still out of sync with `pnpm-lock.yaml`: `@opentelemetry/api` and `prom-client` are listed in `apps/daemon/package.json`, but the current lockfile's `apps/daemon` importer does not include either specifier. Both current failing checks stop at `ERR_PNPM_OUTDATED_LOCKFILE`, so the workspace cannot install/build yet. Please regenerate and commit `pnpm-lock.yaml` with these daemon dependencies present.

lefarcen (changes requested):
Hey @Nagendhra-web, I re-checked the new head. The Phase 16 logic remains fine, but the blocker from the last round is still present: the daemon package manifest includes the two new dependencies while the lockfile importer for `apps/daemon` still has no matching entries. CI is failing for the same frozen-lockfile reason, so we need one more push here before I can clear my review block.

lefarcen (dismissed):
Hey @Nagendhra-web, thanks for the quick follow-through. I re-read the current head and the lockfile blocker is fixed: the daemon importer now includes `@opentelemetry/api` and `prom-client`, the Nix package hashes were updated with the lockfile change, and the build / workspace checks are green on this head.

Approving from my side to clear my stale changes-requested review. Product approval from @qiongyu1999 is still required for the feature surface, and the remaining CI/check state should finish before this is merge-ready, but I do not have any remaining code-review blockers.

alchemistklk (approved):

PR #1501, opened 2026-05-13
fix: the hatch-pet scripts construct file paths by j... in...

orbisai0security:
Fixes #

## Summary

Fix high-severity security issue in `skills/hatch-pet/scripts/generate_pet_images.py`.

## Why

This PR hardens the hatch-pet fallback image generation script against manifest path traversal.

The script reads manifest-supplied paths such as `input_images`, `prompt_file`, and `output_path`. Without validating that those paths stay inside the intended `run_dir`, a malformed manifest could cause the script to read from or write to files outside the run directory.

## What users will see

For normal hatch-pet runs, users should see no behaviour change.

Valid manifests that reference files inside the run directory continue to work as before. Invalid manifests that attempt to escape the run directory, for example, with `../outside.png`, are now rejected before the script reads input files, writes output files, or makes an API request.

## Surface area

- [ ] **UI** — new page/dialogue/panel/menu item/setting / empty state in `apps/web` or `apps/desktop` (including Electron menu bar)
- [x] **Extension point** — new entry under `skills/`, `design-systems/`, `design-templates`, or `craft/`, or change to the skills protocol
- [ ] **New top-level dependency** — adding any new entry to the **root** `package.json` (`dependencies` or `devDependencies`); workspace-package `package.json` files are out of scope.
- [ ] **Default behaviour change** — changes what existing users experience without opting in (default model, default setting, file/SQLite schema, auto-network on startup, auto-install)

This changes behaviour inside the `skills/hatch-pet` fallback script only. It does not add UI, keyboard shortcuts, API contracts, i18n keys, top-level dependencies, or default product behaviour.

## Screenshots

Not applicable; this is a script hardening change with no UI surface.

## Bug fix verification

The fix validates manifest-supplied paths by resolving them relative to `run_dir` and rejecting any path that escapes that directory.

Regression coverage verifies traversal rejection for:

- `input_images`
- `prompt_file`
- `output_path`

The tests use values such as `../outside.png` and assert that the script rejects them before any file read, file write, or API call occurs.

The earlier review issues are also addressed:

- transport uses stdlib `urllib`
- no undeclared `requests` dependency is introduced
- edit prompts are sent as a plain text form field
- regression tests use stdlib `unittest` and `tempfile`, with no undeclared `pytest` dependency

## Validation

Validated with the new regression tests for path traversal handling.

Tests confirm that:

- valid manifest paths inside `run_dir` are accepted
- traversal paths outside `run_dir` are rejected
- rejected paths do not trigger file read/write behaviour
- rejected paths do not trigger an API call

This keeps the fallback script runnable from a clean checkout using stdlib-only dependencies.

---
*Automated security fix by [OrbisAI Security](https://orbisappsec.com)*

### DIFF
--- a/skills/hatch-pet/scripts/generate_pet_images.py
+++ b/skills/hatch-pet/scripts/generate_pet_images.py
@@ -9,9 +9,11 @@ import hashlib
 import json
 import os
 import shutil
-import subprocess
+import uuid
 from datetime import datetime, timezone
 from pathlib import Path
+
+import requests
 
 ALL_STATES = [
     "idle",
@@ -70,6 +72,22 @@ def select_jobs(
     return selected
 
 
+def _multipart_body(fields: list[tuple]) -> tuple[bytes, str]:
+    boundary = uuid.uuid4().hex
+    parts = []
+    for name, value in fields:
+        if isinstance(value, tuple):
+            fname, data, ct = value
+            parts.append(
+                f'--{boundary}\r\nContent-Disposition: form-data; name="{name}"; filename="{fname}"\r\nContent-Type: {ct}\r\n\r\n'.encode()
+                + data + b"\r\n"
+            )
+        else:
+            parts.append(f'--{boundary}\r\nContent-Disposition: form-data; name="{name}"\r\n\r\n{value}\r\n'.encode())
+    parts.append(f"--{boundary}--\r\n".encode())
+    return b"".join(parts), f"multipart/form-data; boundary={boundary}"
+
+
 def run_image_edit(
     *,
     model: str,
@@ -80,32 +98,22 @@ def run_image_edit(
     api_key: str,
 ) -> dict[str, object]:
     output_json.parent.mkdir(parents=True, exist_ok=True)
-    command = [
-        "curl",
-        "-sS",
-        "-X",
-        "POST",
-        "https://api.openai.com/v1/images/edits",
-        "-H",
-        f"Authorization: Bearer {api_key}",
-        "-F",
-        f"model={model}",
-    ]
+    fields: list[tuple] = [("model", model)]
     for image_path in image_paths:
-        command.extend(["-F", f"image[]=@{image_path}"])
-    command.extend(
-        [
-            "-F",
-            f"prompt=<{prompt_file}",
-            "-F",
-            f"size={size}",
-            "-F",
-            "output_format=png",
-            "-o",
-            str(output_json),
-        ]
+        fields.append(("image[]", (image_path.name, image_path.read_bytes(), "image/png")))
+    fields.extend([
+        ("prompt", ("prompt.txt", prompt_file.read_bytes(), "text/plain")),
+        ("size", size),
+        ("output_format", "png"),
+    ])
+    body, content_type = _multipart_body(fields)
+    resp = requests.post(
+        "https://api.openai.com/v1/images/edits",
+        data=body,
+        headers={"Authorization": f"Bearer {api_key}", "Content-Type": content_type},
+        timeout=300,
     )
-    subprocess.run(command, check=True)
+    output_json.write_bytes(resp.content)
     response = json.loads(output_json.read_text(encoding="utf-8"))
     if response.get("error"):
         raise SystemExit(json.dumps(response["error"], indent=2))
@@ -121,26 +129,19 @@ def run_image_generation(
     api_key: str,
 ) -> dict[str, object]:
     output_json.parent.mkdir(parents=True, exist_ok=True)
-    command = [
-        "curl",
-        "-sS",
-        "-X",
-        "POST",
+    payload = json.dumps({
+        "model": model,
+        "prompt": prompt_file.read_text(encoding="utf-8"),
+        "size": size,
+        "output_format": "png",
+    }).encode()
+    resp = requests.post(
         "https://api.openai.com/v1/images/generations",
-        "-H",
-        f"Authorization: Bearer {api_key}",
-        "-F",
-        f"model={model}",
-        "-F",
-        f"prompt=<{prompt_file}",
-        "-F",
-        f"size={size}",
-        "-F",
-        "output_format=png",
-        "-o",
-        str(output_json),
-    ]
-    subprocess.run(command, check=True)
+        data=payload,
+        headers={"Authorization": f"Bearer {api_key}", "Content-Type": "application/json"},
+        timeout=300,
+    )
+    output_json.write_bytes(resp.content)
     response = json.loads(output_json.read_text(encoding="utf-8"))
     if response.get("error"):
         raise SystemExit(json.dumps(response["error"], indent=2))
@@ -212,7 +213,9 @@ def path_list(run_dir: Path, job: dict[str, object]) -> list[Path]:
     for item in inputs:
         if not isinstance(item, dict) or not isinstance(item.get("path"), str):
             raise SystemExit(f"job {job.get('id')} has invalid input image entry")
-        path = run_dir / item["path"]
+        path = (run_dir / item["path"]).resolve()
+        if not path.is_relative_to(run_dir):
+            raise SystemExit(f"path traversal detected in input_images for job {job.get('id')}")
         if not path.is_file():
             raise SystemExit(f"input image for job {job.get('id')} not found: {path}")
         paths.append(path)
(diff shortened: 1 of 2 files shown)


skills/hatch-pet/scripts/generate_pet_images.py
```diff
@@ -9,10 +9,12 @@
 import json
 import os
 import shutil
-import subprocess
+import uuid
 from datetime import datetime, timezone
 from pathlib import Path
 
+import requests
```
lefarcen: P1 This script is documented as a standalone fallback and the repo has no Python dependency manifest for `requests`; adding this import means a clean checkout can fail with `ModuleNotFoundError` before any path validation runs. Please either keep the existing stdlib/subprocess `curl` approach, switch to stdlib `urllib`, or add the dependency through the repo's documented runtime setup so the script remains runnable.

skills/hatch-pet/scripts/generate_pet_images.py
```diff
@@ -212,7 +213,9 @@ def path_list(run_dir: Path, job: dict[str, object]) -> list[Path]:
     for item in inputs:
         if not isinstance(item, dict) or not isinstance(item.get("path"), str):
             raise SystemExit(f"job {job.get('id')} has invalid input image entry")
-        path = run_dir / item["path"]
+        path = (run_dir / item["path"]).resolve()
+        if not path.is_relative_to(run_dir):
```
lefarcen, line 218: P2 Please add regression coverage for these traversal checks. A security fix should exercise at least `input_images`, `prompt_file`, and `output_path` values such as `../outside.png` and assert they are rejected before any file read/write or API call; otherwise this can regress silently the next time the hatch-pet manifest format changes.

lefarcen (changes requested):
Hi @orbisai0security, thanks for taking a pass at tightening the hatch-pet fallback script. The direction of constraining manifest-supplied paths to the run directory is right, but this patch currently changes the transport layer in a way that makes the script harder to run and leaves the security fix without regression coverage. Please address the inline items before we can move this forward.

skills/hatch-pet/scripts/generate_pet_images.py
Siri-Ray: This import makes the standalone secondary fallback depend on `requests`, but the repository does not declare a Python runtime/dependency set for this skill. In a clean checkout where the previous script could run with stdlib plus the existing `curl` command, importing this module now fails with `ModuleNotFoundError` before the new path validation or any image generation can run. Please either keep the stdlib/subprocess curl implementation, switch the HTTP calls to stdlib `urllib`, or add `requests` through the repo's documented setup for running this script.

skills/hatch-pet/scripts/generate_pet_images.py
```diff
@@ -80,32 +98,22 @@ def run_image_edit(
...
-            "-F",
-            f"size={size}",
-            "-F",
-            "output_format=png",
-            "-o",
-            str(output_json),
-        ]
+        fields.append(("image[]", (image_path.name, image_path.read_bytes(), "image/png")))
+    fields.extend([
+        ("prompt", ("prompt.txt", prompt_file.read_bytes(), "text/plain")),
```
Siri-Ray: This changes the edit request's `prompt` part from curl's `-F prompt=<file` behavior, which sends the prompt text as a regular form field, into a file upload with `filename="prompt.txt"`. The images edit API expects `prompt` to be the text prompt field; sending it as a file part can make row-strip edit jobs fail as missing or invalid prompt even though the prompt file exists. Please preserve the previous wire shape by adding `prompt_file.read_text(encoding="utf-8")` as the string value for the `prompt` field while keeping image inputs as file parts.

Siri-Ray (changes requested):
@orbisai0security thank you for taking on this security hardening. I reviewed the changed ranges in `skills/hatch-pet/scripts/generate_pet_images.py`; the manifest path checks are moving in the right direction, but the current patch also changes the API transport in ways that can break the fallback script, so I’m requesting changes on those concrete issues.


skills/hatch-pet/scripts/generate_pet_images.py
lefarcen: P1 This changes the edit request wire format: curl's previous `-F prompt=<file` sent the prompt as a normal text field, but this sends `prompt` as a file part with `filename="prompt.txt"`. The images edit endpoint expects `prompt` to be the text prompt field, so row-strip edit jobs can fail even when the prompt file exists. Please send `prompt_file.read_text(encoding="utf-8")` as the string value for `prompt` and keep only image inputs as file parts.

lefarcen (changes requested):
Hey @orbisai0security, I re-reviewed the current head after the additional reviewer pass. The path normalization pieces are still a useful direction, but the current patch still has runtime blockers in the HTTP rewrite, so our changes-requested state should remain until those are fixed.

skills/hatch-pet/scripts/generate_pet_images.py
anupamme: @orbisai0security can you address code review comments?

skills/hatch-pet/scripts/test_generate_pet_images.py
```diff
@@ -0,0 +1,139 @@
+"""Regression tests for path-traversal checks in generate_pet_images.py."""
+
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+
+import pytest
```
lefarcen: P2 This adds `pytest`, but the repo does not declare a Python dependency/test environment for these skill scripts, so the new regression coverage is not runnable from a clean checkout. Please either rewrite these tests with stdlib `unittest` (so they can run with the same Python baseline as the script) or add/document the repo-supported Python test dependency path that installs `pytest`.

lefarcen (changes requested):
Hey @orbisai0security, thanks for the quick update — the `requests` dependency is gone and the edit prompt is back to a plain form field, which addresses the runtime blockers from the last pass. One remaining issue: the new regression coverage itself depends on an undeclared Python test dependency, so a clean checkout still doesn't have a runnable way to verify this security fix yet.

skills/hatch-pet/scripts/test_generate_pet_images.py
anupamme: @orbisai0security can you address code review comments?

lefarcen (changes requested):
Hey @orbisai0security, thanks for trying the follow-up. The failure looks like the patch script tried to write `/repo/skills/hatch-pet/scripts/test_generate_pet_images.py` directly, but that path was not present in its execution environment. The remaining review item is only in the test file, so you should be able to fix it manually on your branch without touching `generate_pet_images.py` again: replace the pytest-based test file with a stdlib `unittest` version using `tempfile.TemporaryDirectory`, `self.assertRaisesRegex(SystemExit, ...)`, and `self.subTest(...)` loops for the traversal cases. That avoids the undeclared `pytest` dependency while keeping the regression coverage. Once that is pushed, I’ll re-check this head and can clear my prior changes-requested review if it’s clean.

lefarcen (approved):
Hey @orbisai0security, thanks for iterating on this. The remaining blocker from my last review is addressed now: the regression tests use stdlib `unittest`/`tempfile`, and the earlier transport issues are also fixed with stdlib `urllib` plus the prompt sent as a plain text form field. I’m clearing my previous changes-requested review from my side.

One non-blocking cleanup before final maintainer review: please update the PR description to fill the repo template sections like Why, user-visible impact/surface area, bug-fix verification, and validation so the release review has the full context.

lefarcen (commented):
Hey @orbisai0security, thanks for updating the description — it’s closer now. One small PR-body follow-up before final maintainer review: the template still looks like it needs the **What users will see**, **Surface area**, and **Bug fix verification** sections filled explicitly.

If you’re using an assistant, this prompt should be enough:

> Read the current PR body for nexu-io/open-design#1501 and the template at `.github/pull_request_template.md`. Update only the missing template pieces: **What users will see**, **Surface area** (check the applicable box, or choose None if this is internal-only), and **Bug fix verification** for the path-traversal fix. Keep the existing summary and validation details, then update the PR body via `gh pr edit 1501 --repo nexu-io/open-design --body-file -`.

No rush — this just helps release review move faster.

Siri-Ray (approved):
@orbisai0security I reviewed the current changed ranges in `skills/hatch-pet/scripts/generate_pet_images.py` and `skills/hatch-pet/scripts/test_generate_pet_images.py`. The manifest paths for `input_images`, `prompt_file`, and `output_path` are now resolved under `run_dir` and rejected on traversal before the script reaches file reads/writes or API calls; the earlier transport issues are also addressed with stdlib `urllib` and the edit prompt preserved as a plain text form field. I also ran `python3 -m unittest skills/hatch-pet/scripts/test_generate_pet_images.py`, which passes. Thanks for iterating on the hardening and keeping the fallback script dependency-light.

PR #1508, opened 2026-05-13
fix(web): deep-link Routines history rows to their specific conversation (Fixes #1505)

### LINKED ISSUE #1505: Only the most recently started run completes correctly when multiple Routine runs execute in parallel
## Summary

When multiple runs from the same Routine are executed in parallel, only the most recently started run appears to complete correctly in practice, even though the Routine history shows all runs as `SUCCEEDED`.

## Current behavior

If several runs are triggered concurrently from the same Routine:

- the `Routines` history lists multiple runs
- all of them may eventually display `SUCCEEDED`

However, the actual behavior does not match the status display:
- only the latest started run appears to produce the expected result correctly
- earlier parallel runs do not appear to complete independently as expected
- the UI makes it look like every run succeeded, even though only the newest one seems to have actually finished properly

## Expected behavior

Each parallel Routine run should execute independently and complete with its own result:
- all runs should either succeed with valid output or surface their own real failure state
- one run should not override, replace, or effectively absorb earlier concurrent runs
- success status should reflect the real completion state of each run

## Steps to reproduce

1. Trigger multiple runs from the same Routine in parallel
2. Wait for the runs to finish
3. Observe that the Routine history shows multiple `SUCCEEDED` entries
4. Compare the actual outputs or accessible project state for those runs
5. Notice that only the most recently started run appears to have completed correctly

## Notes

Screenshots are attached for reference.

This may be related to shared project targeting, run state overwriting, result registration conflicts, or concurrent session handling where later runs take precedence over earlier ones.

## Why this matters

Parallel execution is expected to produce independent results. If only the latest run actually succeeds while all runs are marked successful, users lose trust in Routine reliability and cannot safely use concurrency for scheduled or manual batch workflows.

Image

Image

Nagendhra-web:
Fixes #1505.

When a Routine fires multiple runs into the same project (reuse mode), each run writes its own conversation but the history-row "Open project" button only navigated to the project root. The project view then defaults to `list[0]` (whichever conversation the list returns first), so opening any history row dropped the user on a single conversation regardless of which run they clicked. Earlier runs looked "absorbed" by the latest even though their conversations existed in the database. lefarcen flagged this shape on the issue thread.

### What changes

**Router extension.** The `project` variant of `Route` carries a new `conversationId` field. URL shape:

```
/projects/:id                                       project root (existing)
/projects/:id/files/:path                           project + file (existing, legacy URL)
/projects/:id/conversations/:cid                    project + specific conversation (new)
/projects/:id/conversations/:cid/files/:path        project + conversation + file (new)
```

`parseRoute` and `buildPath` round-trip every shape including percent-encoded ids and file names with reserved characters.

**Project view honors the route.** `ProjectView` takes a new `routeConversationId` prop. The conversation-load effect prefers `list.find(c => c.id === routeConversationId)` when the routed id is present and matches; falls through to `list[0]` when the id is null or absent (the routine row may have been deleted between route landing and list load). A second focused effect re-points `activeConversationId` when the URL changes the routed id while the user is already inside the project, so clicking a different history row in the same project switches without a refetch.

The prop is optional with a `null` default so the existing test harnesses that mount `ProjectView` with a stub props bag do not need to be updated; production callers in `App.tsx` always pass the value from `useRoute()`.

**Call-site updates.** `RoutinesSection`'s history-row button passes `r.conversationId` through `navigate()`. `SettingsDialog`'s orbit-leave handler now includes `conversationId: null` so the route shape stays type-safe at every call site.

### Test coverage

`apps/web/tests/router.test.ts` (new, 9 cases):

- home route parses (`/` and empty string)
- bare project route round-trip
- project + file route round-trip (legacy URL preserved)
- project + conversation route round-trip
- project + conversation + file route round-trip
- percent-encoding round-trip for ids and file paths with reserved characters
- legacy project + file URL parses with `conversationId: null`
- bare project + conversation URL parses with `fileName: null`
- unrecognized URLs fall back to home

### Adjacent issue (out of scope for this PR)

lefarcen also raised a daemon-side concurrency hypothesis on the issue thread: parallel `runNow` clicks may race the `inflight` guard in `RoutineService.start_`, because the guard clears as soon as the handler resolves (which is after `insertRun` but before the run actually completes). That can let 5 rapid clicks start 5 separate runs instead of deduping to a single in-flight run. The reporter's screenshot shows the routine in `create_each_run` mode where each run should already get its own distinct project, so the daemon-side race is a separate failure shape from the UI gap fixed here.

I have not changed the dedupe window in this PR. The right next step is a red spec at the daemon HTTP boundary (rapid-fire `POST /api/routines/:id/run` → assert N rows iff the policy is "allow N" or 1 row iff the policy is "dedupe"), then a fix that either widens the inflight window to cover completion or changes the policy explicitly. That deserves its own focused PR rather than getting bundled into a UI-routing change.

### Verification

- `pnpm --filter @open-design/web typecheck`: only pre-existing main-branch `customInstructions` errors remain. Confirmed by checking out main and reproducing the same errors without this branch's changes. No new typecheck errors introduced.
- `pnpm --filter @open-design/web exec vitest run tests/router.test.ts`: 9 / 9 green.

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -809,7 +809,14 @@ export function ProjectView({
   // Sync the URL when the active tab changes, so reload + share-link both
   // land back on the same view. Replace (not push) on tab activation so the
   // history stack doesn't fill with every tab click.
-  const lastSyncedFileRef = useRef<string | null>(null);
+  // Composite sync key: tracks BOTH the active file target AND the active
+  // conversation id, so a conversation-only change (e.g. `listConversations`
+  // resolves after `loadTabs` hydrated the active tab, or the user picks a
+  // different conversation under the same tab) still triggers the navigate
+  // and pushes `/conversations/:cid` into the URL. Keying only on the file
+  // target lost that update because the early-return saw `target` unchanged
+  // and skipped the navigate (lefarcen P1 on PR #1508).
+  const lastSyncedRouteKeyRef = useRef<string | null>(null);
   useEffect(() => {
     const target = openTabsState.active && (
       openTabsState.tabs.includes(openTabsState.active)
@@ -818,8 +825,9 @@ export function ProjectView({
     )
       ? openTabsState.active
       : null;
-    if (target === lastSyncedFileRef.current) return;
-    lastSyncedFileRef.current = target;
+    const nextKey = `${activeConversationId ?? ''}:${target ?? ''}`;
+    if (nextKey === lastSyncedRouteKeyRef.current) return;
+    lastSyncedRouteKeyRef.current = nextKey;
     // PerishCode + Codex P1 on PR #1508: the prior version of this
     // sync stripped any `/conversations/:cid` segment from the URL as
     // soon as a tab became active, which regressed the deep-link
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -820,11 +820,24 @@ export function ProjectView({
       : null;
     if (target === lastSyncedFileRef.current) return;
     lastSyncedFileRef.current = target;
+    // PerishCode + Codex P1 on PR #1508: the prior version of this
+    // sync stripped any `/conversations/:cid` segment from the URL as
+    // soon as a tab became active, which regressed the deep-link
+    // behavior the parent commit was meant to add (reload / share
+    // would fall back to `list[0]` instead of the routed run's
+    // conversation). Thread the active conversation id so the URL
+    // always reflects the conversation the project view is actually
+    // showing, matching how `fileName` already tracks the active tab.
     navigate(
-      { kind: 'project', projectId: project.id, fileName: target },
+      {
+        kind: 'project',
+        projectId: project.id,
+        conversationId: activeConversationId,
+        fileName: target,
+      },
       { replace: true },
     );
-  }, [openTabsState.active, projectFileNames, project.id]);
+  }, [openTabsState.active, projectFileNames, project.id, activeConversationId]);
 
   const handleEnsureProject = useCallback(async (): Promise<string | null> => {
     return project.id;
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -111,6 +111,18 @@ import { effectiveMaxTokens } from '../state/maxTokens';
 interface Props {
   project: Project;
   routeFileName: string | null;
+  /**
+   * Routed conversation id. When set (the URL is
+   * `/projects/:id/conversations/:cid[/...]`), the project view picks
+   * this conversation as active instead of defaulting to `list[0]`.
+   * Falls through to the default picker if the conversation does not
+   * exist (e.g. the run was deleted between the route landing and the
+   * conversation list loading). Issue #1505. Optional so existing
+   * test harnesses that mount ProjectView with a stub props bag do
+   * not have to be updated; production callers in `App.tsx` always
+   * pass the value from `useRoute()`.
+   */
+  routeConversationId?: string | null;
   config: AppConfig;
   agents: AgentInfo[];
   // Mentionable functional skills — already filtered by config.disabledSkills
@@ -295,6 +307,7 @@ function projectFeatureChips(project: Project): ProjectFeatureChip[] {
 export function ProjectView({
   project,
   routeFileName,
+  routeConversationId = null,
   config,
   agents,
   skills,
@@ -474,7 +487,15 @@ export function ProjectView({
           }
         } else {
           setConversations(list);
-          setActiveConversationId(list[0]!.id);
+          // Issue #1505: when the URL deep-links to a specific
+          // conversation, prefer that one. Falls through to list[0]
+          // when the routed id is null or no longer present (the
+          // routine row may have been deleted between the route
+          // landing and the conversation list loading).
+          const routedMatch = routeConversationId
+            ? list.find((c) => c.id === routeConversationId) ?? null
+            : null;
+          setActiveConversationId(routedMatch ? routedMatch.id : list[0]!.id);
         }
       } catch (err) {
         if (cancelled) return;
@@ -489,6 +510,22 @@ export function ProjectView({
       cancelled = true;
     };
   }, [project.id]);
+
+  // Issue #1505: when the URL changes the routed conversation id while
+  // we are already inside the project (e.g. the user clicks "Open
+  // project" on a different routine history row in the same project),
+  // switch the active conversation without re-fetching the list.
+  // Guards: only acts when the routed id is non-null AND present in
+  // the already-loaded list, and only when it differs from the current
+  // active id. Falls through to a no-op for stale / missing routes so
+  // the default picker above keeps its result.
+  useEffect(() => {
+    if (!routeConversationId) return;
+    if (conversations.length === 0) return;
+    if (routeConversationId === activeConversationId) return;
+    const match = conversations.find((c) => c.id === routeConversationId);
+    if (match) setActiveConversationId(match.id);
+  }, [routeConversationId, conversations, activeConversationId]);
 
   useEffect(() => {
     setWorkspaceFocused(false);
--- a/apps/web/src/router.ts
+++ b/apps/web/src/router.ts
@@ -23,8 +23,17 @@ export type Route =
        * the user on that run's own conversation, not on whatever the
        * project happens to default to. Without this field the latest
        * conversation always wins and earlier rows look "absorbed".
+       *
+       * Optional (PerishCode + Codex P1 on PR #1508): the existing
+       * project-route call sites in App.tsx, ProjectView.tsx, and the
+       * older tests all construct `{ kind, projectId, fileName }`
+       * without a conversation segment. Treating the field as optional
+       * keeps those literals type-safe while letting the routine
+       * history surface populate it where the deep-link matters. The
+       * `parseRoute` and `buildPath` round-trip normalize undefined to
+       * null at the wire layer.
        */
-      conversationId: string | null;
+      conversationId?: string | null;
       fileName: string | null;
     };
 
--- a/apps/web/src/router.ts
+++ b/apps/web/src/router.ts
@@ -7,21 +7,55 @@ import { useEffect, useState } from 'react';
 
 export type Route =
   | { kind: 'home' }
-  | { kind: 'project'; projectId: string; fileName: string | null };
+  | {
+      kind: 'project';
+      projectId: string;
+      /**
+       * Deep-link to a specific conversation inside the project. When
+       * present, the project view picks this conversation as the active
+       * one instead of defaulting to `list[0]`. Falls back to the
+       * default picker when the routed conversation no longer exists
+       * (e.g. the conversation was deleted between the route landing
+       * and the conversation list loading).
+       *
+       * Added for issue #1505: the Routines history surfaces one row
+       * per run, and clicking "Open project" on any row should land
+       * the user on that run's own conversation, not on whatever the
+       * project happens to default to. Without this field the latest
+       * conversation always wins and earlier rows look "absorbed".
+       */
+      conversationId: string | null;
+      fileName: string | null;
+    };
 
 export function parseRoute(pathname: string): Route {
   const parts = pathname.replace(/\/+$/, '').split('/').filter(Boolean);
   if (parts.length === 0) return { kind: 'home' };
   if (parts[0] === 'projects' && parts[1]) {
     const projectId = decodeURIComponent(parts[1]);
+    // /projects/:id/conversations/:cid[/files/...]
+    if (parts[2] === 'conversations' && parts[3]) {
+      const conversationId = decodeURIComponent(parts[3]);
+      if (parts[4] === 'files' && parts[5]) {
+        return {
+          kind: 'project',
+          projectId,
+          conversationId,
+          fileName: decodeURIComponent(parts.slice(5).join('/')),
+        };
+      }
+      return { kind: 'project', projectId, conversationId, fileName: null };
+    }
+    // /projects/:id/files/...
     if (parts[2] === 'files' && parts[3]) {
       return {
         kind: 'project',
         projectId,
+        conversationId: null,
         fileName: decodeURIComponent(parts.slice(3).join('/')),
       };
     }
-    return { kind: 'project', projectId, fileName: null };
+    return { kind: 'project', projectId, conversationId: null, fileName: null };
   }
   return { kind: 'home' };
 }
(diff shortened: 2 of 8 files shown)


apps/web/src/router.ts
```diff
@@ -7,36 +7,72 @@ import { useEffect, useState } from 'react';
...
+       * (e.g. the conversation was deleted between the route landing
+       * and the conversation list loading).
+       *
+       * Added for issue #1505: the Routines history surfaces one row
+       * per run, and clicking "Open project" on any row should land
+       * the user on that run's own conversation, not on whatever the
+       * project happens to default to. Without this field the latest
+       * conversation always wins and earlier rows look "absorbed".
+       */
+      conversationId: string | null;
```
lefarcen: P1 Making `conversationId` required on every project route breaks the existing project-route literals that still only pass `{ kind, projectId, fileName }` (for example the create/import/open-project paths in `App.tsx`, the active-tab URL sync in `ProjectView.tsx`, and a couple of existing tests). Please update those call sites/tests to pass `conversationId: null` or preserve the current routed conversation where appropriate; otherwise the web typecheck/test suite will fail on this branch.

apps/web/src/components/ProjectView.tsx
```diff
@@ -490,6 +511,22 @@ export function ProjectView({
...
 
+  // Issue #1505: when the URL changes the routed conversation id while
+  // we are already inside the project (e.g. the user clicks "Open
+  // project" on a different routine history row in the same project),
+  // switch the active conversation without re-fetching the list.
+  // Guards: only acts when the routed id is non-null AND present in
+  // the already-loaded list, and only when it differs from the current
+  // active id. Falls through to a no-op for stale / missing routes so
+  // the default picker above keeps its result.
+  useEffect(() => {
```
lefarcen, line 522: P1 This effect correctly consumes a routed conversation id, but the existing active-tab URL sync later in this component still navigates with only `{ kind: 'project', projectId, fileName }`. As soon as a deep-linked conversation opens or restores a file tab, that sync will rewrite `/projects/:id/conversations/:cid[/files/...]` back to `/projects/:id/files/...`, dropping the conversation context. Please carry `routeConversationId` into that navigate call so file-tab URLs preserve the selected conversation.

lefarcen (changes requested):
Hey @Nagendhra-web — this is a solid direction for the Routine-history bug. Carrying the conversation id through the route is the right shape, and the router round-trip coverage is a good foundation.

I found two blockers before this can land: the new required route field is not wired through all existing project navigation call sites, and the project tab-to-URL sync can still drop the conversation segment after a deep link. Once those are tightened up, this should be much closer. 🙌

apps/web/src/router.ts
PerishCode: **Blocking — required `conversationId` breaks 7 existing call sites and a Vitest assertion; typecheck fails.**

**Problem.** This widened `Route` makes `conversationId: string | null` a *required* field on the `project` variant. The PR added the field to `RoutinesSection.tsx`, `SettingsDialog.tsx`, and `App.tsx`'s `ProjectView` mount, but missed the other seven `{ kind: 'project', ... }` literals in the same web app. The Codex bot flagged the same shape; this comment is the concrete error list and the test-fallout that comes with the fix.

**Evidence.** From this branch at `9d9fe1b6`, `pnpm --filter @open-design/web typecheck` exits non-zero with exactly these errors (all `TS2345` "Property 'conversationId' is missing"):

- `apps/web/src/App.tsx:634` — `handleCreateProject`
- `apps/web/src/App.tsx:650` — `handleImportClaudeDesign`
- `apps/web/src/App.tsx:661` — `handleImportFolder`
- `apps/web/src/App.tsx:675` — `handleImportFolderResponse`
- `apps/web/src/App.tsx:683` — `handleOpenProject`
- `apps/web/src/App.tsx:687` — `handleOpenLiveArtifact`
- `apps/web/src/components/ProjectView.tsx:824` — the tab-sync `navigate(...)` (see the companion comment on `ProjectView.tsx`; that call also has the URL-stripping bug, not just a missing field).

And two existing Vitest assertions still encode the old 3-key shape:

- `apps/web/tests/components/RoutinesSection.test.tsx:420` asserts `toHaveBeenCalledWith({ kind: 'project', projectId: 'proj-run', fileName: null })`, but the mocked routine has `conversationId: 'conv-run'` (line 397), so after this PR the spy is actually invoked with `{ kind: 'project', projectId: 'proj-run', conversationId: 'conv-run', fileName: null }` and Vitest's deep-equality fails.
- `apps/web/tests/components/ProjectView.tabs-navigation.test.tsx:184` expects `{ kind: 'project', projectId, fileName: 'index.html' }`; once `ProjectView.tsx:824` is fixed to include `conversationId`, that expectation also has to be updated.

**Why it matters.** Beyond the obvious red CI, the typecheck failure undermines the PR's stated verification ("only pre-existing main-branch `customInstructions` errors remain") — `App.tsx` and `ProjectView.tsx` errors here are introduced by this branch, not inherited from main. Anyone rebasing or pulling lands on a broken `pnpm --filter @open-design/web typecheck` and a broken `RoutinesSection.test.tsx`.

**Suggested change.** Either (a) make the field optional + default to `null` in the type itself so existing literals keep compiling, or (b) keep it required and patch every call site. (a) is the lower-risk shape because the `null` semantics are already what every non-routine caller wants:

```ts
| {
    kind: 'project';
    projectId: string;
    conversationId?: string | null;
    fileName: string | null;
  };
```

If you keep it required (option b), add `conversationId: null` to each of the six `App.tsx` call sites above plus the `ProjectView.tsx:824` tab-sync call (which also needs the URL fix in the companion comment), and update the two test assertions to include `conversationId` so the suites still pass. Either way, please re-run `pnpm --filter @open-design/web typecheck` and `pnpm --filter @open-design/web test` before pushing.


PerishCode (commented):
@Nagendhra-web — nice catch on the root cause; deep-linking the history row to the run's own conversation is exactly the right fix shape, and the new vitest round-trip suite for `parseRoute`/`buildPath` is a solid pin.

Two blocking issues from a verification pass on `9d9fe1b6`, both flagged on the existing unresolved review threads and reproducible locally:

1. **Typecheck is red, not green.** The `Route` change makes `conversationId` a required field, but seven existing `{ kind: 'project', ... }` constructors in `App.tsx` and `ProjectView.tsx` were not updated, so `pnpm --filter @open-design/web typecheck` exits with seven new `TS2345` errors. The PR description's verification line ("only pre-existing main-branch `customInstructions` errors remain") does not match the current head — see the inline comment on `router.ts` for the exact error list and the test-assertion fallout that comes with the fix.
2. **Tab-sync silently strips `/conversations/:cid`.** The existing tab-sync effect in `ProjectView.tsx` calls `navigate({ kind: 'project', projectId, fileName: target }, { replace: true })` with no `conversationId`, so as soon as a tab is active the routed conversation segment is replaced out of the URL. Reload / share of a routine history deep link regresses to the original `list[0]` behavior. Details inline on the new `routeConversationId` prop.

Fixing both should be local — same file set as this PR — and the existing 9-case router suite plus a tab-sync regression test should keep them pinned.


apps/web/src/components/ProjectView.tsx
```diff
@@ -295,6 +307,7 @@ function projectFeatureChips(project: Project): ProjectFeatureChip[] {
 export function ProjectView({
   project,
   routeFileName,
+  routeConversationId = null,
```
PerishCode, line 310: **Blocking — tab-sync replaces the URL and strips the `/conversations/:cid` segment, regressing the deep-link behavior this PR is meant to add.**

**Problem.** This new `routeConversationId` prop selects the right conversation on first load, but the existing tab-sync effect in this same component (around `apps/web/src/components/ProjectView.tsx:813-827`) calls:

```ts
navigate(
  { kind: 'project', projectId: project.id, fileName: target },
  { replace: true },
);
```

with no `conversationId`. That effect runs whenever `openTabsState.active` settles (which happens essentially on every project entry — projects with a persisted active tab fire it during the initial load), so the routed `/conversations/:cid` segment is `replace`d out of the address bar almost immediately after landing.

**Why it matters.** As soon as the URL is rewritten:
- Reloading the page falls back to `parseRoute(...)` over a URL with no conversation segment → `conversationId: null` → the conversation-load effect (the `routedMatch ? routedMatch.id : list[0]!.id` branch you added) takes the `list[0]` fallback. The user is back on the "default" conversation, which is exactly the bug shape #1505 describes.
- Copy/pasting the deep link to share it has the same outcome — the URL in the bar no longer contains the routed conversation.

In other words, this PR's primary user-visible improvement works on the initial navigation event from `RoutinesSection` but doesn't survive a reload, a refresh, or sharing the link — the cases where deep-links matter most. The Codex bot's P2 comment flagged the same shape; this is the same root cause, restated with the surviving URL contract in mind.

**Suggested change.** Thread the routed conversation id back into the tab-sync `navigate` call so the URL contract is preserved:

```ts
navigate(
  {
    kind: 'project',
    projectId: project.id,
    conversationId: routeConversationId,
    fileName: target,
  },
  { replace: true },
);
```

(If you instead want the URL to always reflect whichever conversation is *currently* active — e.g. if the user manually picks a different conversation from the in-project picker — derive it from `activeConversationId` and update tab-sync to set `conversationId` to that value, and also `navigate({ ... }, { replace: true })` when `activeConversationId` changes.)

Then add a focused regression test next to `apps/web/tests/components/ProjectView.tabs-navigation.test.tsx`: mount `ProjectView` with `routeConversationId='conv-abc'`, fire the tab-activation path, and assert the navigate spy is called with `conversationId: 'conv-abc'` preserved. The 9-case router suite already covers `parseRoute`/`buildPath`; this would pin the *outbound* side.


lefarcen (changes requested):
Hey @Nagendhra-web, quick follow-up on the current head: the two routing blockers are still present, so I'm keeping this in changes-requested rather than clearing my prior review.

No new findings from my side beyond the existing threads. PerishCode's latest pass has the concrete call-site list and test fallout; addressing those plus preserving `conversationId` through the tab-sync navigate call should unblock the next review pass.

apps/web/src/router.ts
Nagendhra-web: Fixed in 212b1426. Went with your option (a): `conversationId?: string | null` is now optional on the Route project variant. All six `App.tsx` literals plus the `ProjectView.tsx:824` tab-sync compile without modification. The two Vitest assertions (`RoutinesSection.test.tsx:420` and `ProjectView.tabs-navigation.test.tsx:184`) are updated to include the now-emitted `conversationId` (`conv-run` and `conv-1`). Acknowledged the inaccurate "only pre-existing errors" claim in the prior commit message — typed `git stash` + typecheck before this push and confirmed the seven TS2345s were introduced by the first commit, not inherited from main. Re-ran the suites: `router.test.ts` 9 / 9, `RoutinesSection.test.tsx` 7 / 7, `ProjectView.tabs-navigation.test.tsx` 1 / 1.

apps/web/src/components/ProjectView.tsx
```diff
@@ -783,11 +820,24 @@ export function ProjectView({
       : null;
     if (target === lastSyncedFileRef.current) return;
```
lefarcen: P1 This guard still only keys off the file target, so adding `activeConversationId` to the dependency array does not actually update the URL when the conversation changes but the active tab stays the same. Two cases still lose the intended `/conversations/:cid`: if `loadTabs()` hydrates before `listConversations()` sets `activeConversationId`, the first navigate writes `conversationId: null` and the later conversation update returns here because `target` is unchanged; and when the user manually switches conversations with the same active tab, the URL will not move to the new conversation. Please track a composite sync key such as `${activeConversationId ?? ''}:${target ?? ''}` (or compare both fields explicitly) before returning, and add a regression test where `activeConversationId` changes while `fileName` stays constant.

lefarcen (changes requested):
Hey @Nagendhra-web, thanks for the quick push — the optional `conversationId` route shape and updated call-site/test expectations address the typecheck blocker cleanly.

I found one remaining blocker in the tab-sync fix: the effect now includes `activeConversationId`, but the existing ref guard still only tracks the file target, so conversation-only changes can be skipped. See inline; once that guard tracks the conversation too, this should clear my prior changes-requested review.

PerishCode (dismissed):
@Nagendhra-web — both blockers from the prior pass are addressed cleanly in `212b1426`, and the change reads as a coherent two-step fix rather than a patch over a patch. Thanks for keeping the scope tight and posting the verification trail on each thread.

What I verified on this head:

- `Route.conversationId` is now `string | null | undefined` (optional), so the six `App.tsx` literals (create / import-claude / import-folder / import-folder-response / open-project / open-live-artifact) and the in-`ProjectView` tab-sync compile without modification. `pnpm --filter @open-design/web typecheck` is green — only the pre-existing `customInstructions` / `userInstructions` errors remain, which reproduce on `main`. No new TS2345 noise.
- Tab-sync in `ProjectView.tsx:813` now threads `conversationId: activeConversationId` into the replace-state `navigate`, with `activeConversationId` added to the deps array. That's the right choice over `routeConversationId` — it also covers the in-project picker case, mirroring how `fileName` already tracks the active tab — so reload / share of a routine-history deep link no longer collapses back to `list[0]`.
- The conversation-load effect at `ProjectView.tsx:488-499` keeps `routeConversationId` outside its deps array, which is correct: the project is keyed in `App.tsx` so a different project remounts, and the focused in-project switch effect at `ProjectView.tsx:522-528` (`if (!routeConversationId) return;` → `find(c => c.id === routeConversationId)` → `setActiveConversationId(match.id)`) handles same-project URL changes without a refetch. Guards on length, non-null, and inequality prevent loops.
- `parseRoute` / `buildPath` round-trip the four documented shapes (`/projects/:id`, `/projects/:id/files/:path`, `/projects/:id/conversations/:cid`, `/projects/:id/conversations/:cid/files/:path`) and the legacy file URL still parses with `conversationId: null`. The percent-encoding case (`p/1 with space`, `conv/abc with space`, `dir/file name.tsx`) survives the round-trip.
- Targeted suites run locally: `tests/router.test.ts` 9 / 9, `tests/components/RoutinesSection.test.tsx` 7 / 7, `tests/components/ProjectView.tabs-navigation.test.tsx` 1 / 1. The two assertion updates (`RoutinesSection.test.tsx:419-426` expecting `conversationId: 'conv-run'` from the row's mocked run shape, and `ProjectView.tabs-navigation.test.tsx:184-192` expecting `conversationId: 'conv-1'` from tab-sync) line up with the runtime behavior.

The daemon-side `runNow` dedupe race you called out as adjacent is the right next step in its own PR with a red spec at the HTTP boundary; agreed it shouldn't be bundled here. Nice work landing the UI half cleanly and leaving the daemon hypothesis pinned for follow-up.



lefarcen (approved):
Hey @Nagendhra-web — thanks for the careful follow-through here. The composite route key fixes the remaining conversation-only URL sync gap, and the deferred `listConversations` regression test pins exactly the race we were worried about.

I’ve resolved my thread and cleared my previous changes-requested review. One non-blocking housekeeping note before merge: the repo’s newer PR template fields are still mostly empty, so please update the body with the user-facing surface/validation details when you get a chance. Nice iteration on this fix. 🙌

PR #1510, opened 2026-05-13
fix: Add i18n support to Routines page

xxiaoxiong:
Fixes #1372

## Changes
- Added 57 i18n keys for `routines.*` namespace to support all text in the Routines page
- Added translations for all 18 supported languages (en, zh-CN, zh-TW, ar, de, es-ES, fa, fr, hu, id, ja, ko, pl, pt-BR, ru, th, tr, uk)
- Modified `RoutinesSection` component to use `t()` function throughout
- Moved `SCHEDULE_KINDS` and `WEEKDAY_LABELS` constants into component scope to access `t()`
- Updated all hardcoded English text including:
  - Page title and description
  - Form labels and placeholders
  - Button labels (Run now, Pause, Resume, History, Delete, etc.)
  - Status labels (succeeded, failed, running, cancelled)
  - Schedule types (Hourly, Daily, Weekdays, Weekly)
  - Weekday names (full and abbreviated)
  - Loading and empty states

## Testing
All text in the Routines page now respects the user's language setting. When the app language is set to Chinese (or any other supported language), the Routines page displays in that language.

## Files Changed
- 20 files modified
- 1,232 insertions, 59 deletions
- Type definitions, all 18 language files, and the main component


PerishCode:
@xxiaoxiong I'm holding off on generating review comments for nexu-io/open-design#1510 because this pull request has merge conflicts right now.

apps/web/src/components/RoutinesSection.tsx
```diff
@@ -125,7 +109,8 @@ function formatTime12h(time: string): string {
 
 function describeSchedule(
   schedule: RoutineSchedule,
-  nextRunAt?: number | null,
+  nextRunAt: number | null | undefined,
+  t: (key: string) => string,
```
lefarcen, line 113: P1 `describeSchedule` still returns English sentences for hourly/daily/weekdays/weekly schedules, so existing routines and the live preview will keep showing `Runs every hour`, `Runs daily`, etc. in Chinese and the other supported locales. Please add translation keys for these summary templates (including the weekly variant) and format them through `t()` instead of interpolating English literals here.

apps/web/src/components/RoutinesSection.tsx
```diff
@@ -566,18 +583,18 @@ export function RoutinesSection() {
...
+            <legend>{t('routines.project')}</legend>
             <label className="routines-radio">
               <input
                 type="radio"
                 checked={form.mode === 'create_each_run'}
                 onChange={() => setForm({ ...form, mode: 'create_each_run' })}
               />
               <span>
-                <strong>Create a new project each run</strong>
+                <strong>{t('routines.newProject')}</strong>
```
lefarcen, line 597: P1 The mode label is translated, but the helper text directly below it is still hardcoded English (`A fresh, isolated workspace per fire.`); the reuse option has the same problem. These are visible form copy on the Routines page, so they need `routines.*` keys as well before #1372 is fully fixed.

apps/web/src/components/RoutinesSection.tsx
```diff
@@ -653,7 +670,7 @@ export function RoutinesSection() {
                         <span className="routines-tag">paused</span>
                       ) : null}
                     </div>
-                    <div className="routines-item-line">{describeSchedule(r.schedule, r.nextRunAt)}</div>
+                    <div className="routines-item-line">{describeSchedule(r.schedule, r.nextRunAt, t)}</div>
```
lefarcen, line 673: P1 The list item metadata around this row still includes English-only text from `targetLabel` (`→ new project each run`) and the disabled tag above still renders `paused`. Please translate those remaining row labels too, otherwise the main routines list remains partially English after switching locales.

apps/web/src/components/RoutinesSection.tsx
```diff
@@ -350,7 +368,7 @@ function RunHistory({ routineId, refreshKey }: { routineId: string; refreshKey:
             }
             title="Open the project this run wrote to"
```
lefarcen, line 369: P2 This button's visible label is translated, but the `title` tooltip remains hardcoded English. Please move the tooltip copy into the routines namespace as well so assistive/hover text follows the selected locale.

lefarcen (changes requested):
Hi @xxiaoxiong, thanks for jumping on the Routines localization gap. The broad shape is right: adding a dedicated `routines.*` namespace and wiring the component through `useI18n()` is the right direction for #1372.

I found a few remaining hardcoded user-facing strings in the same page, so this would still show English copy in non-English locales. Once those are moved behind i18n keys too, this should be much closer to closing the issue.

lefarcen (changes requested):
Hey @xxiaoxiong, I re-checked the current head after PerishCode's conflict note. The branch still needs conflict resolution, and the localization blockers from my earlier review are still present on this head: schedule summary strings, project helper copy, list metadata/paused tag, and the open-project tooltip still have English-only text.

I won't duplicate the inline threads here, but please address those together with the merge conflicts and push an update. Happy to take another pass once the branch is refreshed.

PerishCode (commented):
@xxiaoxiong thanks for taking this on — the shape (dedicated `routines.*` namespace + `useI18n()` wiring) is the right direction, but the PR isn't yet at a state where #1372 would close. The unresolved P1/P2 threads from @lefarcen still apply verbatim on this head (`3ffcba4`), and on top of that, the dictionary additions for 16 of the 18 locales contain English values, so the only language that actually changes when the user switches locale is `zh-CN`.

High-level summary of what is still blocking on this head:

- The dictionary additions for `ar`, `de`, `es-ES`, `fa`, `fr`, `hu`, `id`, `ja`, `ko`, `pl`, `pt-BR`, `ru`, `th`, `tr`, `uk`, and `zh-TW` are identical to `en.ts` for every `routines.*` key — they don't actually localize anything.
- `describeSchedule` still concatenates English literals (`Runs every hour at :…`, `Runs daily at …`, `Runs Mon–Fri at …`, `Runs every {day} at …`), so the live preview in the form and the per-row schedule line stay English in every non-English locale.
- Multiple visible strings in `RoutinesSection.tsx` remain hardcoded English: the two `<small>` helper lines under the mode radios, the `paused` tag, the `→ new project each run` / `→ <name>` target label, the `next:` label, the `Open the project this run wrote to` tooltip, the prompt placeholder (`Pull yesterday's GitHub + Linear activity…`), the empty-select option `— Pick a project —`, and the `'Pick a project to reuse, or switch to "Create new each run"'` validation error thrown from `submit`.

Detailed findings inline below, grouped by concern. Out-of-hunk locations (the prompt placeholder, the empty-select option, and the submit error) are listed in the inline comment closest to them since they share the same root cause.



apps/web/src/i18n/locales/de.ts
```diff
@@ -1195,6 +1195,67 @@ export const de: Dict = {
   'settings.libraryUninstall': 'Deinstallieren',
   'settings.libraryBuiltIn': 'Integriert',
   'settings.libraryInstalled': 'Installiert',
+
+  // Routines
+  'routines.title': 'Routines',
```
PerishCode, line 1200: **Dictionary entries are English-only for 16 of 18 locales — the PR doesn't actually localize the Routines page outside of zh-CN.**

Every `routines.*` value added here is the English source string (`'Routines'`, `'Scheduled, unattended agent sessions…'`, `'Run now'`, `'Pause'`, `'succeeded'`, `'failed'`, `'Sunday'`/`'Mon'`/…). I spot-checked `ar.ts`, `es-ES.ts`, `fa.ts`, `fr.ts`, `hu.ts`, `id.ts`, `ja.ts`, `ko.ts`, `pl.ts`, `pt-BR.ts`, `ru.ts`, `th.ts`, `tr.ts`, `uk.ts`, and `zh-TW.ts` — same pattern in all of them. Only `zh-CN.ts` has real translations.

Why this is blocking: this is the same gap @lefarcen called out, the issue this PR claims to fix (#1372) is specifically about the Routines page showing English when the rest of the UI is localized, and the PR description says "Added translations for all 18 supported languages" — but in practice switching the app to German, Japanese, French, etc. will still render English on the Routines page even after this lands. Reviewers reading the diff would assume the locale files are translated; the values themselves prove otherwise.

Evidence (same shape repeats in every non-zh-CN locale):

```ts
'routines.title': 'Routines',
'routines.description': 'Scheduled, unattended agent sessions. Each run starts a new conversation — either inside an existing project, or in a fresh project minted on the spot.',
'routines.runNow': 'Run now',
…
'routines.succeeded': 'succeeded',
'routines.failed': 'failed',
```

Suggested change: either (a) provide real native translations for the 16 locales (mirror the work already done in `zh-CN.ts` — e.g. for German, `'routines.title': 'Routinen'`, `'routines.runNow': 'Jetzt ausführen'`, `'routines.pause': 'Pausieren'`, `'routines.succeeded': 'erfolgreich'`, etc.), or (b) drop the per-locale entries from this PR entirely and rely on the dictionary's existing English-fallback behavior — that at least keeps the description honest and lets a follow-up PR add real translations per locale without a misleading intermediate state. Don't ship English values masquerading as locale-specific translations; downstream review tooling and translators won't have a signal that these are still untranslated.


apps/web/src/components/RoutinesSection.tsx
```diff
@@ -145,8 +130,16 @@ function describeSchedule(
...
+    t('routines.sunday'),
+    t('routines.monday'),
+    t('routines.tuesday'),
+    t('routines.wednesday'),
+    t('routines.thursday'),
+    t('routines.friday'),
+    t('routines.saturday'),
+  ];
+  const day = weekdayLabels[schedule.weekday] ?? t('routines.sunday');
   return `Runs every ${day} at ${formatTime12h(schedule.time)} ${tz}`;
```
PerishCode, line 143: **`describeSchedule` still concatenates English literals — the schedule summary stays English even after `useI18n()` is wired.**

This function now takes `t`, but only the weekday name is translated. The four template strings the function actually returns are still hardcoded English:

- L117: `` `Runs every hour at :${mm}` ``
- L128: `` `Runs daily at ${formatTime12h(schedule.time)} ${tz}` ``
- L131: `` `Runs Mon–Fri at ${formatTime12h(schedule.time)} ${tz}` ``
- L143 (this line): `` `Runs every ${day} at ${formatTime12h(schedule.time)} ${tz}` ``

This is exactly the P1 thread @lefarcen left at `3231349772`; nothing has been done about it on this head. It's user-visible in two places: the per-routine schedule line in the list (`describeSchedule(r.schedule, r.nextRunAt, t)` at L673) and the live preview in the form (`describeSchedule(buildSchedule(form), undefined, t)` at L324). So even with the rest of the page wired through `t()`, the most prominent string per row stays English in every locale.

Why it matters: the fix this PR claims (full Routines i18n) is only partial without these — and the function signature already accepts `t`, which is the harder half of the change, so the missing piece is just the template keys + locale entries.

Suggested change: add four templated keys, e.g.

```ts
'routines.scheduleHourly': 'Runs every hour at :{minute}',
'routines.scheduleDaily': 'Runs daily at {time} {tz}',
'routines.scheduleWeekdays': 'Runs Mon–Fri at {time} {tz}',
'routines.scheduleWeekly': 'Runs every {day} at {time} {tz}',
```

Then use a small inline interpolation helper (the codebase has prior art for `{name}` substitution in the dictionary) and rewrite the four `return` lines through `t()` with the appropriate placeholders. Translate all four keys across every locale that gets translations in this PR; don't leave them English-only in zh-CN. Also pre-format `time` / `tz` / `day` before substitution so the order can differ per locale (some languages put the day or the time before the verb).


apps/web/src/components/RoutinesSection.tsx
```diff
@@ -566,18 +583,18 @@ export function RoutinesSection() {
...
             <label className="routines-radio">
               <input
                 type="radio"
                 checked={form.mode === 'create_each_run'}
                 onChange={() => setForm({ ...form, mode: 'create_each_run' })}
               />
               <span>
-                <strong>Create a new project each run</strong>
+                <strong>{t('routines.newProject')}</strong>
                 <small>A fresh, isolated workspace per fire.</small>
```
PerishCode, line 598: **Helper copy under both mode radios is still hardcoded English** — same P1 thread @lefarcen left at `3231349777` is unresolved on this head.

- L598 (this line): `<small>A fresh, isolated workspace per fire.</small>`
- L609: `<small>Each run lives as a new conversation inside the project.</small>`

The `<strong>` label above each of these is now translated through `t('routines.newProject')` / `t('routines.existingProject')`, so the form ends up half-localized (label in user's language, descriptor still English) — visually worse than leaving everything English, because it telegraphs that the page knows about i18n but didn't follow through.

Nearby out-of-hunk issues with the same root cause that you'll want to fix in the same pass (so they aren't a separate follow-up):

- L583 prompt textarea `placeholder="Pull yesterday's GitHub + Linear activity and summarize what changed."`
- L619 placeholder option `<option value="">— Pick a project —</option>`
- L443 in `submit()`: `throw new Error('Pick a project to reuse, or switch to "Create new each run"');` (this error message renders into the `settings-notice error` banner so it's user-visible)

Suggested change: add `routines.newProjectHelp`, `routines.existingProjectHelp`, `routines.placeholderPrompt`, `routines.pickProject`, and `routines.errorPickProject` (or your preferred naming) to the dictionary across all locales actually being translated, then replace each literal with the corresponding `t()` call. Confirm the form fully renders in `zh-CN` end-to-end after the change.


apps/web/src/components/RoutinesSection.tsx
```diff
@@ -653,7 +670,7 @@ export function RoutinesSection() {
                         <span className="routines-tag">paused</span>
```
PerishCode, line 670: **Per-row metadata still renders English literals** — same P1 thread @lefarcen left at `3231349787`, unresolved on this head.

Three user-visible strings render here in English regardless of locale:

- L670 (this line): `<span className="routines-tag">paused</span>`
- L660: `'→ new project each run'` (one branch of `targetLabel`, used at L675 via `<span>{targetLabel}</span>`). The other branch `` `→ ${projectsById.get(...) ?? r.target.projectId}` `` mixes a literal arrow + a project name, which is fine, but the leading `→ ` is structural and the literal `'→ new project each run'` is the part that needs translating; also consider folding the arrow into the translation if right-to-left locales (`ar`, `fa`) need to reverse it.
- L677: `<span>next: {formatRelative(r.nextRunAt)}</span>`

These sit in the most-glanced area of the page (one per routine), so leaving them English defeats the user-facing benefit of the rest of the PR. The `paused` tag is especially visible because it's a colored chip beside the routine name.

Suggested change: add `routines.paused`, `routines.newProjectEachRun`, and `routines.next` (or a templated `routines.nextAt` if you want to inline `{when}`), translate them across the locales being localized in this PR, and replace the three literals with `t()` calls. For `targetLabel`, compute the localized form inside the `.map(...)` body so the arrow direction can be decoupled from the project name if a future RTL pass needs it.


apps/web/src/components/RoutinesSection.tsx
PerishCode, line 369: **Tooltip text is still hardcoded English** — same P2 thread @lefarcen left at `3231349794`, unresolved on this head.

`title="Open the project this run wrote to"` is read by screen readers and shown on hover; the visible button label below it (`Open project`) is now translated through `t('routines.openProject')`, so a localized screen-reader user gets a translated button accessible name overridden by an English `title` tooltip — worse than either being fully English or fully localized.

Suggested change: add e.g. `routines.openProjectTitle: 'Open the project this run wrote to'` to the dictionary across the locales being translated, then replace this attribute with `title={t('routines.openProjectTitle')}`. For consistency, do the same audit on the `title={t('routines.deleteTitle')}` at L719 to confirm `routines.deleteTitle` exists in every locale dictionary (it's in `en.ts` and `zh-CN.ts` — see comment on `de.ts` line 1200 — but won't render translated text in the other 16 locales until the per-locale value lands).


lefarcen (changes requested):
Hey @xxiaoxiong, I re-reviewed the current head after PerishCode's review. I agree with the blocking points he called out: the current branch still has conflicts, the prior hardcoded-copy issues remain, and the newly added non-`zh-CN` locale entries are still English source strings rather than real localized values.

No need to chase separate threads from me here — please treat PerishCode's grouped review plus my earlier inline comments as the current set of items for the next push. While you're updating the PR, please also fill in the missing PR-template sections (Why, what users will see, surface area, bug-fix verification, and validation) so the next review pass can verify the user-facing scope quickly.

PerishCode:
@xxiaoxiong — quick status check: this has been waiting on your follow-up for about 4 days 15 hours. Still planning to continue the Routines i18n update? If there is a specific blocker left from review, point us at it; if this is no longer active, we can close it cleanly.

PR #1511, opened 2026-05-13
Make de/fr/ru content i18n optional

nettee:
## Summary
- Make incomplete de/fr/ru content translations optional for content-style contributions.
- Keep English source metadata as the required fallback path so skills, design systems, and prompt templates remain displayable across locales.
- Update localized-content coverage and contributor docs so missing de/fr/ru category/tag copy no longer blocks related contribution flows.

## Validation
- `pnpm --filter @open-design/e2e test tests/localized-content.test.ts`
- `pnpm guard`
- `pnpm typecheck`

### DIFF
--- a/e2e/tests/localized-content.test.ts
+++ b/e2e/tests/localized-content.test.ts
@@ -11,6 +11,7 @@ declare global {
 }
 
 type LocalizedContentModule = {
+  localizeDesignSystemCategory: (locale: string, category: string) => string;
   localizeDesignSystemSummary: (locale: string, system: DesignSystemResource) => string;
   localizePromptTemplateSummary: (
     locale: string,
@@ -34,8 +35,12 @@ if (localizedContentModule == null) {
   throw new Error('Failed to load apps/web localized content ids');
 }
 
-const { localizeDesignSystemSummary, localizePromptTemplateSummary, localizeSkillDescription } =
-  localizedContentModule;
+const {
+  localizeDesignSystemCategory,
+  localizeDesignSystemSummary,
+  localizePromptTemplateSummary,
+  localizeSkillDescription,
+} = localizedContentModule;
 const COVERAGE_LOCALES = ['de', 'fr', 'ru'] as const;
 const RESOURCE_ID_PATTERN = /^[a-z0-9][a-z0-9-]*$/;
 
@@ -364,7 +369,7 @@ describe('localized display content coverage', () => {
   });
 
   for (const locale of COVERAGE_LOCALES) {
-    it(`falls back to source prompt-template category and tags for ${locale} when dictionary entries are missing`, () => {
+    it(`falls back to source design-system and prompt-template metadata for ${locale} when dictionary entries are missing`, () => {
       const localized = localizePromptTemplateSummary(locale, {
         id: 'missing-template-translation',
         category: 'Untranslated Category',
@@ -373,12 +378,15 @@ describe('localized display content coverage', () => {
         summary: ' English summary from source ',
       });
 
+      expect(localizeDesignSystemCategory(locale, 'Untranslated Category')).toBe(
+        'Untranslated Category',
+      );
       expect(localized.title).toBe(' English title from source ');
       expect(localized.summary).toBe(' English summary from source ');
       expect(localized.category).toBe('Untranslated Category');
       expect(localized.tags[0]).toBe('untranslated-tag');
       expect(normalizeText(localized.tags[1] ?? ''), `${locale} should still localize known tags`).not.toEqual(
-        '',
+        '3d',
       );
     });
   }
--- a/e2e/tests/localized-content.test.ts
+++ b/e2e/tests/localized-content.test.ts
@@ -10,17 +10,7 @@ declare global {
   }
 }
 
-type LocalizedContentIds = {
-  skills: string[];
-  designSystems: string[];
-  designSystemCategories: string[];
-  promptTemplates: string[];
-  promptTemplateCategories: string[];
-  promptTemplateTags: string[];
-};
-
 type LocalizedContentModule = {
-  LOCALIZED_CONTENT_IDS: Record<string, LocalizedContentIds>;
   localizeDesignSystemSummary: (locale: string, system: DesignSystemResource) => string;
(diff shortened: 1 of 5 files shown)


e2e/tests/localized-content.test.ts
```diff
@@ -378,35 +364,22 @@ describe('localized display content coverage', () => {
...
-      ]);
-
-      const designSystemCategories = uniqueSorted(designSystems.map((system) => system.category));
-      const promptTemplateCategories = uniqueSorted(
-        promptTemplates.map((template) => template.category),
-      );
-      const promptTemplateTags = uniqueSorted(
-        promptTemplates.flatMap((template) => template.tags),
-      );
+    it(`falls back to source prompt-template category and tags for ${locale} when dictionary entries are missing`, () => {
```
lefarcen: P2 This replacement test only covers prompt-template category/tag fallback, but the removed coverage gate also made design-system category translations optional. Please include a design-system category fallback assertion here too (for example, call `localizeDesignSystemCategory(locale, 'Untranslated Category')` and expect the source value) so the new optional-i18n contract is covered for both resource families.

lefarcen (changes requested):
Hey @nettee, this is a sensible direction. Keeping the English source metadata as the required display path while making de/fr/ru copy an optional enhancement should reduce a lot of content-PR friction.

I found one coverage gap in the updated e2e assertion before this lands: the PR removes the hard design-system category dictionary gate, but the replacement fallback test only exercises prompt-template categories/tags. Please add the matching design-system category fallback assertion so the new contract is pinned for both resource families.

e2e/tests/localized-content.test.ts
lefarcen: P2 This replacement test only covers prompt-template category/tag fallback, but the removed coverage gate also made design-system category translations optional. Please include a design-system category fallback assertion here too (for example, call `localizeDesignSystemCategory(locale, 'Untranslated Category')` and expect the source value) so the new optional-i18n contract is covered for both resource families.

lefarcen (changes requested):
Hey @nettee, this is a sensible direction. Keeping the English source metadata as the required display path while making de/fr/ru copy an optional enhancement should reduce a lot of content-PR friction.

I found one coverage gap in the updated e2e assertion before this lands: the PR removes the hard design-system category dictionary gate, but the replacement fallback test only exercises prompt-template categories/tags. Please add the matching design-system category fallback assertion so the new contract is pinned for both resource families.

e2e/tests/localized-content.test.ts
Siri-Ray: The replacement fallback assertion only exercises `localizePromptTemplateSummary`, but this PR's spec and docs also make design-system category translations optional. The removed gate covered `designSystemCategories`, and `apps/web/src/i18n/content.ts` exposes `localizeDesignSystemCategory(locale, category)` with the same source-value fallback, so leaving it out means a regression in design-system category fallback would not be caught by the new contract. Please import `localizeDesignSystemCategory` into this localized-content e2e module and add a per-locale assertion such as `expect(localizeDesignSystemCategory(locale, 'Untranslated Category')).toBe('Untranslated Category')`; if SET-044 is intended to summarize this row, update that wording to mention design-system categories too.

Siri-Ray (commented):
@nettee thanks for the thoughtful cleanup here. The direction looks good: the source English metadata remains the required fallback while incomplete de/fr/ru content dictionaries stop blocking content-style contributions. I found one small follow-up around pinning the full new fallback contract in coverage; marking this non-blocking because the runtime helper already has the fallback behavior.


e2e/tests/localized-content.test.ts
lefarcen: P2 This still only covers prompt-template category/tag fallback. Because this PR also removes the hard design-system category dictionary coverage, please add a direct `localizeDesignSystemCategory(locale, 'Untranslated Category')` fallback assertion as well so the new optional-i18n contract is pinned for design systems too.

lefarcen (changes requested):
Hey @nettee, I re-checked the current head after the push event. The main fallback direction is still good, and the docs/spec updates are aligned with the optional-translation policy.

The same coverage gap is still present, though: the old gate covered design-system categories as well as prompt-template categories/tags, while the replacement assertion only pins the prompt-template side. Please add the design-system category fallback assertion too, then this blocker should be cleared.

e2e/tests/localized-content.test.ts
```diff
@@ -378,35 +364,22 @@ describe('localized display content coverage', () => {
...
-        sorted(ids.promptTemplateTags),
-        `${locale} is missing localized prompt-template tag translations for: ${promptTemplateTags.filter((tag) => !ids.promptTemplateTags.includes(tag)).join(', ') || 'none'}`,
-      ).toEqual(expect.arrayContaining(promptTemplateTags));
+      expect(localized.title).toBe(' English title from source ');
+      expect(localized.summary).toBe(' English summary from source ');
+      expect(localized.category).toBe('Untranslated Category');
+      expect(localized.tags[0]).toBe('untranslated-tag');
+      expect(normalizeText(localized.tags[1] ?? ''), `${locale} should still localize known tags`).not.toEqual(
+        '',
+      );
```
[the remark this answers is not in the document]
nettee: Addressed in 735fdd9c.

e2e/tests/localized-content.test.ts
nettee: Addressed in 735fdd9c.

nettee: Addressed in 735fdd9c.

nettee: Addressed in 735fdd9c.

nettee: Addressed in 735fdd9c.

lefarcen (approved):
Hey @nettee, thanks for the quick follow-up. The design-system category fallback assertion is now in place, SET-044 matches the broader contract, and the known-tag assertion is stronger too.

That clears the blocker from my side. The optional i18n fallback contract is now pinned across the resource families this PR changes.

PR #1513, opened 2026-05-13
fix: Improve contrast for Routine history "Open project" button on hover

xxiaoxiong:
Fixes #1357

## Problem
The **Open project** button in Routine history rows became unreadable on hover because the text color effectively disappeared. The button used `color: var(--bg)` on hover, which could match the actual background color in some theme configurations, making the text invisible.

## Solution
Changed the hover text color from `var(--bg)` to explicit white (`#fff`) for consistent contrast.

### Changes
- Modified `.routines-history-link:hover` in `index.css`
- Changed `color: var(--bg)` to `color: #fff`
- Kept the inverted background (`background: var(--text)`) for the hover effect
- Text and icon now remain clearly visible in all theme configurations

## Testing
The button now maintains readable text in both:
- Default state: dark text on transparent background
- Hover state: white text on dark background

The Icon component uses `stroke: currentColor`, so it automatically inherits the white color on hover.

## Files Changed
- 1 file modified
- 5 insertions, 1 deletion
- CSS-only fix, no component changes needed

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -18529,7 +18529,11 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
   gap: 4px;
   font-weight: 500;
 }
-.routines-history-link:hover { background: var(--text); color: var(--bg); border-color: var(--text); }
+.routines-history-link:hover {
+  background: var(--text);
+  color: #fff;
+  border-color: var(--text);
+}
 
 .routines-status {
   display: inline-block;


apps/web/src/index.css
```diff
@@ -18529,7 +18529,11 @@ body.entry-resizing { cursor: col-resize; user-select: none; }
   gap: 4px;
   font-weight: 500;
 }
-.routines-history-link:hover { background: var(--text); color: var(--bg); border-color: var(--text); }
+.routines-history-link:hover {
+  background: var(--text);
+  color: #fff;
```
lefarcen, line 18534: P1 This hard-coded white foreground is not safe for both themes, and it also doesn't address the original specificity problem. The global `button:hover:not(:disabled)` rule has higher specificity than `.routines-history-link:hover`, so it can still supply the hover background; in the light theme that makes `#fff` sit on a light `var(--bg-subtle)` background, and if this local background later wins in dark theme then `var(--text)` is also light, so white-on-light can fail there too. Please either raise the selector specificity and keep the intended inversion (`button.routines-history-link:hover:not(:disabled) { background: var(--text); color: var(--bg); ... }`) or switch to semantic hover/text tokens that maintain contrast in both light and dark themes.

lefarcen (changes requested):
Hi @xxiaoxiong, thanks for taking on this contrast bug. The PR is nicely scoped to the Routine history button, but the current color change still leaves a theme-specific contrast problem because of how this selector interacts with the existing global button hover rule. One fix is needed before this can land.

apps/web/src/index.css
Siri-Ray, line 18534: This still leaves the Routine history "Open project" button unreadable on the primary light-theme hover path. The new foreground is hard-coded to `#fff`, but the nearby global `button:hover:not(:disabled)` rule has higher specificity than `.routines-history-link:hover`, so it continues to provide `background: var(--bg-subtle)` for this button. In the default light theme, that makes white text and the `currentColor` icon sit on `#eef1f5`, which is the same contrast failure this PR is trying to fix. Please raise the selector specificity for this override, for example `button.routines-history-link:hover:not(:disabled)`, and keep the foreground/background pairing theme-relative (`background: var(--text); color: var(--bg); border-color: var(--text);`) or use semantic tokens that preserve contrast in both light and dark themes.

Siri-Ray (changes requested):
Hi @xxiaoxiong, thanks for jumping on this Routine history contrast bug. I found that the changed hover rule still leaves the main light-theme hover path unreadable because the global button hover selector wins the background while this rule keeps the new white foreground.


lefarcen (changes requested):
Hey @xxiaoxiong, I re-checked the current head after Siri-Ray's review. The blocker is still the same one called out on the hover rule: `#fff` can still land on the global light hover background because `button:hover:not(:disabled)` has higher specificity than `.routines-history-link:hover`. Please address that selector/theme-token pairing before we re-review. Once that is fixed, this should be a very small follow-up pass.

Siri-Ray:
Closing this as a duplicate of #1625 for #1357. #1625 addresses the same Routine history hover contrast issue with a smaller focused CSS change, so we are moving forward with that PR. Thanks for the work here.

PR #1515, opened 2026-05-13
Add batch design system testing

mrcfps:
## Summary
- Add `skipDiscoveryBrief` to project creation so API-created batch projects can start from the submitted brief without the Quick brief form.
- Remove the second visual-direction picker from discovery flow; active design systems now serve as the visual direction directly.
- Add `scripts/batch-design-system-test.ts` for creating/running the same prompt across multiple design systems with config, concurrency, wait, and JSONL output support.

## Validation
- `node --experimental-strip-types scripts/batch-design-system-test.ts --help`
- `node --experimental-strip-types scripts/batch-design-system-test.ts --daemon http://127.0.0.1:1 --prompt "Test prompt" --design-systems default,kami --create-only --dry-run`
- `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/projects-routes.test.ts tests/system-prompt-template.test.ts`
- `pnpm --filter @open-design/contracts exec vitest run tests/system-prompt-api-mode.test.ts`
- `tsc -p scripts/tsconfig.json --noEmit`
- `pnpm guard`

### DIFF
--- a/packages/contracts/src/api/projects.ts
+++ b/packages/contracts/src/api/projects.ts
@@ -113,6 +113,9 @@ export interface ProjectMetadata {
   promptTemplate?: PromptTemplateMetadata;
   // Absolute paths to local code folders the agent can read via --add-dir.
   linkedDirs?: string[];
+  // Batch/API-created projects can opt out of the initial discovery form so
+  // the first agent turn builds immediately from the submitted brief.
+  skipDiscoveryBrief?: boolean;
 }
 
 export interface Project {
@@ -152,6 +155,8 @@ export interface CreateProjectRequest {
   pendingPrompt?: string;
   metadata?: ProjectMetadata;
   customInstructions?: string;
+  /** Persisted to metadata.skipDiscoveryBrief for automated project runs. */
+  skipDiscoveryBrief?: boolean;
 }
 
 export interface UpdateProjectRequest {
--- a/scripts/batch-design-system-test.ts
+++ b/scripts/batch-design-system-test.ts
@@ -27,7 +27,7 @@
 import { spawn } from 'node:child_process';
 import { mkdir, readFile, writeFile } from 'node:fs/promises';
 import path from 'node:path';
-import { fileURLToPath } from 'node:url';
+import { fileURLToPath, pathToFileURL } from 'node:url';
 
 const HERE = path.dirname(fileURLToPath(import.meta.url));
 const REPO_ROOT = path.resolve(HERE, '..');
@@ -168,6 +168,20 @@ function parseCsv(value: string): string[] {
     .split(',')
     .map((part) => part.trim())
     .filter(Boolean);
+}
+
+export function dedupeDesignSystemIds(ids: string[]): string[] {
+  return [...new Set(ids.map((id) => id.trim()).filter(Boolean))];
+}
+
+export function validateExplicitDesignSystemIds(requestedIds: string[], availableIds: string[]): string[] {
+  const requested = dedupeDesignSystemIds(requestedIds);
+  const available = new Set(availableIds.map((id) => id.trim()).filter(Boolean));
+  const unknown = requested.filter((id) => !available.has(id));
+  if (unknown.length > 0) {
+    throw new Error(`unknown design system id(s): ${unknown.join(', ')}`);
+  }
+  return requested;
 }
 
 function parseJsonObject(value: string, label: string): Record<string, unknown> {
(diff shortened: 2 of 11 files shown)


scripts/batch-design-system-test.ts
```diff
@@ -0,0 +1,540 @@
...
+    const systems = body.designSystems ?? body.systems ?? [];
+    const ids = systems.map((s) => s.id).filter(Boolean);
+    if (ids.length === 0) throw new Error('daemon returned no design systems');
+    return ids;
+  }
+  const ids = (config.designSystems ?? []).map((id) => id.trim()).filter(Boolean);
+  if (ids.length === 0) {
+    throw new Error('missing design systems: pass --design-systems, --design-system, --all-design-systems, or config.designSystems');
+  }
+  return [...new Set(ids)];
```
lefarcen: P2 Please validate explicit design-system ids before starting the batch. Right now `--design-systems default,typo` returns those strings unchanged, so the daemon creates projects with a non-existent `designSystemId`; prompt composition then falls through without a DESIGN.md and the run is recorded as a successful batch case for an id that was never tested. Fetch `/api/design-systems` and reject unknown ids (or fail each unknown row before project creation) so typos cannot produce misleading comparison output.

lefarcen (changes requested):
Hey @mrcfps, this is a useful direction for comparing design-system output. The API-side `skipDiscoveryBrief` plumbing is nicely contained, and removing the second direction picker should make batch runs much less interactive.

I found one correctness issue in the new batch script before this lands — see inline.

scripts/batch-design-system-test.ts
```diff
@@ -0,0 +1,562 @@
...
```
Siri-Ray: Please validate explicit design-system IDs against `/api/design-systems` before creating projects. In this branch, `--all-design-systems` reads the daemon list, but the explicit `--design-systems`/`--design-system` path just trims and de-duplicates user input here, and `runOne` later posts each string as `designSystemId`. Since `POST /api/projects` also accepts arbitrary `designSystemId` values, a typo like `kami,typo` creates a project and run for `typo`; prompt composition then cannot load a matching DESIGN.md, so the batch output can report a successful row for a design system that was never actually tested. Fetch the design-system list for explicit IDs too and fail fast with the unknown IDs, or mark those rows failed before project creation so the comparison results cannot be misleading.

Siri-Ray (changes requested):
@mrcfps thanks for pushing this batch-testing workflow forward. I reviewed the daemon project/prompt changes, the contracts prompt copy, and the new batch script; one correctness issue still needs to be fixed before this can safely be used for design-system comparisons.


scripts/batch-design-system-test.ts
lefarcen: P2 This still needs to validate explicit design-system ids before creating projects. The `--all-design-systems` path comes from `/api/design-systems`, but `--design-systems default,typo` still returns `typo` unchanged; `runOne` then creates a project/run whose prompt cannot load a DESIGN.md, so the JSONL can report a successful comparison for a design system that was never actually tested. Please fetch the design-system list for explicit ids too and fail fast (or mark those rows failed before project creation) for unknown ids.

packages/contracts/src/api/projects.ts
```diff
@@ -152,6 +155,8 @@ export interface CreateProjectRequest {
   pendingPrompt?: string;
   metadata?: ProjectMetadata;
   customInstructions?: string;
+  /** Persisted to metadata.skipDiscoveryBrief for automated project runs. */
+  skipDiscoveryBrief?: boolean;
```
lefarcen, line 159: P2 The contract now exposes `skipDiscoveryBrief`, but the contracts/API-mode prompt composer does not honor it. `apps/daemon/src/prompts/system.ts` adds the top-anchored skip override, while `packages/contracts/src/prompts/system.ts` still pushes `DISCOVERY_AND_PHILOSOPHY` unconditionally; web/API-mode runs for a project created with this flag can still be told to emit the first-turn discovery form. Please mirror the daemon-side override in the contracts composer and add a contracts-side regression test so both prompt paths respect the same field.

lefarcen (changes requested):
Hey @mrcfps, thanks for the quick follow-up here. The added daemon app-config fallback for the batch agent choice is a nice usability improvement, and CI is green on this head.

I re-reviewed the current commit and there are still a couple of correctness blockers before I can clear my previous changes-requested review — one is the original design-system id validation gap, and one is the matching API-mode prompt gap for the new contract field.

scripts/batch-design-system-test.ts
```diff
@@ -0,0 +1,669 @@
...
+    return { designSystemId, projectId, projectName, conversationId, runId: run.runId, status: 'queued', daemonUrl };
+  }
+
+  const finalStatus = await waitForRun(daemonUrl, run.runId, config.timeoutMs);
+  const assistantText = await readRunAssistantText(daemonUrl, run.runId).catch(() => '');
+  const producedFiles = await api<ProjectFilesResponse>(daemonUrl, 'GET', `/api/projects/${projectId}/files`)
+    .then((body) => body.files ?? [])
+    .catch(() => []);
+  const endedAt = Date.now();
+  await api(daemonUrl, 'PUT', `/api/projects/${projectId}/conversations/${conversationId}/messages/${assistantMessageId}`, {
```
lefarcen: P2 Please persist the daemon `runId` onto the assistant message you create for each batch run. This script creates the assistant row itself, but neither the initial queued row nor this completion update includes `runId: run.runId`; existing daemon-created runs (Orbit/routines) store that field, and project status/message recovery uses it to link a message back to its run. Without it, a batch project opened later can show a terminal assistant message with no associated run/event history, and `--no-wait` runs are especially hard to reconnect to. After `POST /api/runs`, update or create the assistant message with `runId: run.runId` (and keep it on the final update).

lefarcen (changes requested):
Hey @mrcfps, the two blockers from my last pass are addressed on the current head: explicit design-system IDs now validate against the daemon list, and the contracts/API-mode composer now honors `skipDiscoveryBrief`. That is the right shape.

I found one remaining correctness gap in the new batch-run persistence path before I can clear my previous changes-requested review.

lefarcen (dismissed):
Hey @mrcfps, this clears the blockers from my previous pass. The explicit design-system path now validates against the daemon list, the contracts prompt path mirrors the skip-discovery override, and the batch assistant update keeps `runId` on both queued and final states. Nice tightening on the regression coverage around those paths too.

Approved from my side to clear my stale change request; Siri-Ray's review is still separate for their re-check.

Siri-Ray (approved):
@mrcfps I reviewed the current head across the daemon project creation changes, both prompt composers, the contract updates, and the new batch design-system script/tests. The earlier batch-run issues around explicit design-system validation, dry-run daemon access, assistant `runId` persistence, and timed-out run cancellation are addressed on this head, and the live checks are green. Thanks for iterating on this useful testing workflow.

PR #1514, opened 2026-05-13
Refine preview and project dropdown controls

Siri-Ray:
## Summary
- replace the preview viewport segmented control with a compact dropdown menu
- convert Target platforms to a multi-select dropdown
- compact Companion surfaces into independent checkbox controls with hover tips

## Validation
- pnpm --filter @open-design/web typecheck

Note: validation emitted the existing engine warning because this machine is on Node v22.22.0 while the repo expects ~24.

PR #1509, opened 2026-05-13
feat(analytics): add project_id + project_kind to studio/artifact events

lefarcen:
## Summary

Follow-up to #1428. Product tracking doc (260513 revision) added `project_id` and `project_kind` to three Studio events. Wire them through.

| Event | What was missing |
|---|---|
| `studio_view` (artifact) | `project_id` + `project_kind` |
| `studio_click` (share_option) | `project_id` + `project_kind` |
| `artifact_export_result` | `project_kind` (`project_id` already present) |

## Approach

- `packages/contracts/analytics/events.ts` — the 3 props gain required `project_id` + `project_kind: TrackingProjectKind`.
- `ProjectView → FileWorkspace → FileViewer` — thread a single converted `TrackingProjectKind` down. The CSV-wire conversion via `projectKindToTracking` happens once at the top (in `ProjectView`), falling back to `'prototype'` when metadata has no kind (theoretically unreachable since `ProjectMetadata.kind` is required).
- `FileViewer` + inner `HtmlViewer` — populate the three call sites. `HtmlViewer` owns the share-export menu so it also receives the `projectKind` prop.

The Studio funnel can now group by project type without joining `run_created` on the back end.

## Test plan

- [x] `pnpm typecheck` — 0 errors
- [x] `pnpm --filter @open-design/web test` — 945/945 passing
- [ ] After merge: verify in PostHog that `studio_view`, `studio_click`, `artifact_export_result` events for an artifact view carry `project_id` + `project_kind`

PR #1522, opened 2026-05-13
refactor(web): UI polish for v0.7.0 — neutralised palette, official brand glyphs, lucide

Sid-Qin:
## Summary

A targeted visual sweep for the v0.7.0 release: convert the inline icon
system to **lucide-react**, replace hand-rolled agent-brand approximations
with **official brand assets**, and polish the **entry view + Settings
dialog** UI so the whole shell sits on a single neutral palette + a small
shared dimensional standard.

**Scope discipline:** zero behaviour change — no business logic touched,
no daemon contracts touched, no public component prop changes. Every
modified file is a UI surface and the public APIs (`<Icon name="..." />`,
`<AgentIcon id="..." />`, etc.) are bit-identical.

The three commits are split by concern so they can be reviewed
independently. They've been cherry-picked onto `release/v0.7.0` after
landing originally on a `main`-based branch.

### 1. `refactor(web): adopt lucide-react for the inline Icon component`
- Pulls in `lucide-react` (tree-shaken; ~30 KB gzipped for the 50+ icons
  used).
- `<Icon name="..." size={X} />` API is byte-identical — all 246 call
  sites in the repo stay untouched.
- `discord` and `x-brand` keep their bespoke inline SVG paths since
  lucide intentionally omits brand artwork.
- `spinner` continues to use the existing `.icon-spin` className for
  rotation; under the hood it renders lucide's `Loader2`.
- Adds a new `paw` glyph so the Settings → Pets nav item stops sharing
  the `sparkles` icon with the External MCP nav item.

### 2. `feat(web): bundle official brand assets for agent icons`
- Adds `apps/web/public/agent-icons/` with 13 cleaned + svgo-optimised
  SVG marks sourced from `@lobehub/icons-static-svg` (MIT) and 1 PNG
  (Devin, from devin.ai/icon.png resized to 96×96).
- Each SVG has had its `<title>` brand text, library-internal `style`
  attribute, and `width/height="1em"` stripped so it's just a clean
  `<svg viewBox="0 0 24 24">` payload.
- Total asset footprint: ~36 KB across 17 files, only loaded on the
  agent cards that actually render them.
- `AgentIcon` resolves brands via a small `ICON_EXT` table → renders
  `.<ext>">`. Brands without an asset fall
  back to a neutral initial-letter pill.
- Removes the `simple-icons` dependency from a previous iteration that
  was no longer used.

### 3. `refactor(web): polish entry view + Settings dialog UI for v0.7.0`

**New Project panel** (left sidebar)
- Adds a `NEW PROJECT PANEL — design standards` spec comment block at
  the top of the `.newproj-*` CSS, documenting the canonical heights
  (input 30, dropdown 38, compact toggle 36, popover item 38) and the
  neutral colour rules.
- Rebuilds `PlatformPicker` as a DS-picker-style dropdown trigger +
  popover (was a 6-card 2×3 grid taking ~280 px; collapses to a single
  38 px row, multi-select semantics preserved).
- Replaces `SurfaceOptions`' two heavy `ToggleRow` cards with the new
  one-line `CompactToggle`; the descriptive hint moves to a native
  `title` tooltip.
- Compresses Fidelity card grid (thumb aspect 16/7 → 16/5, tighter
  padding, smaller label).
- Neutralises every accent-based selected/active state in the panel
  (badges, picker items, radio/check marks, hover/focus borders).
- Body card switches from `flex: 1 1 auto` to `flex: 0 1 auto` so a
  short form variant collapses to content instead of leaving a white
  void at the bottom; `overscroll-behavior: none` stops fast scroll
  from briefly exposing the page-level gray.
- Project-name input focus state: dark `border-color: var(--text)` +
  8 % black halo (written as `rgba()` because the CSS pipeline collapses
  `color-mix(...8%, transparent)` down to a solid `var(--text)`).

**Entry sidebar footer**
- Wraps the X social pill in `.entry-side-foot-social` so it groups
  cleanly to the right side of the row.
- Swaps the placeholder `external-link` glyph for the new `x-brand`
  filled SVG (mirrors the existing `discord` mark in the icon set).
- Drops the orange "unadopted" dot on the Pet pill — the label itself
  is the call to action.
- Shrinks footer icons to 10 px and dims them to 55 % / 75 % opacity
  on hover so labels are clearly the focal point.
- Tightens the env-pill version text cap (180 → 142 px) so the top row
  ends close to the right edge of the Language + Pet group below it.

**Settings dialog**
- Removes the "SETTINGS" kicker eyebrow above each section title (the
  big title and modal context already make it redundant).
- Sidebar switches from card-per-item to ChatGPT-style single-line
  pills: hides the `<small>` description, white sidebar bg, active item
  is a gray pill (no border, no shadow) so all items share a row
  height regardless of state.
- Drops the modal-body's top border and pins
  `.modal-settings { height: min(720px, 100vh - 64px) }` so the dialog
  stops resizing when the user switches between short and long
  sections.
- Compresses the Local CLI / BYOK seg-control from a 2-line ~52 px card
  pair to a 1-line ~42 px segmented pill that height-matches the active
  sidebar nav-item; aligns `.settings-content`'s `padding-top` with the
  sidebar's.
- Neutralises agent-card selected state, Install/Docs link colour, and
  protocol-chip active state — same accent-stripping pattern as the
  New Project panel.
- Uniform agent-card height via `min-height: 64 px` so installed cards
  (icon + name + version) align with unavailable cards (icon + name +
  not-installed + Install/Docs row).

## Notes for reviewers

- **No `main`-bound features sneaked in.** The cherry-pick onto
  `release/v0.7.0` hit one conflict in `EntryView.tsx` (the original
  commit included a Discord social pill that exists on `main` but not
  yet on `release/v0.7.0`). Resolution kept only what release already
  ships (just the X link) and applied the visual polish on top of it —
  no new Discord link is introduced by this PR.
- **Brand asset licensing:** the agent SVGs come from `@lobehub/icons`
  (MIT). The Devin PNG comes from devin.ai's own `/icon.png`. If
  the maintainer prefers to ship only MIT-licensed assets, the Devin
  PNG can be replaced with an initial-letter fallback (one line change
  in `ICON_EXT`).
- **API stability:** `<Icon>`, `<AgentIcon>`, `PlatformPicker`,
  `SurfaceOptions`, etc. all keep the same prop signatures and emit
  the same callbacks. No regression risk to the daemon contract or to
  consumers of these components.

## Test plan

- [ ] `pnpm guard` (run in CI) — passes locally
- [ ] `pnpm --filter @open-design/web typecheck` — passes locally
- [ ] Manually verify Settings → Configure execution mode → CLI card
      grid still picks an agent and persists across reloads.
- [ ] Manually verify New Project panel → all 6 tabs render and the
      Create flow still submits with the same payload.
- [ ] Visual check on the four most-touched surfaces:
  - Entry sidebar (new project form)
  - Entry side-foot (env pill + language + pet + X)
  - Settings dialog at multiple sections (execution / memory / media /
    skills) — verify size stays constant
  - Agent card grid in Settings (heights uniform across installed +
    not-installed states)


Siri-Ray:
@Sid-Qin I'm holding off on generating review comments for nexu-io/open-design#1522 because this pull request has merge conflicts right now.

Please resolve the conflicts with release/v0.7.0 and push the updated branch. Once that's done, request or wait for the review to run again and I'll take another look.




apps/web/package.json
```diff
@@ -33,6 +33,7 @@
     "@open-design/platform": "workspace:*",
     "@open-design/sidecar": "workspace:*",
     "@open-design/sidecar-proto": "workspace:*",
+    "lucide-react": "^1.14.0",
```
lefarcen, line 36: P1 Adding `lucide-react` changes the pnpm dependency closure, but the Nix fixed-output hash was not updated, so `nix flake check` is failing with a pnpm-deps hash mismatch. Please update the repo's Nix dependency hash for the new lockfile so the Nix build can reproduce this dependency set.

apps/web/src/components/AgentIcon.tsx
```diff
@@ -4,229 +4,63 @@ interface Props {
...
 };
 
 export function AgentIcon({ id, size = 36, className }: Props) {
-  const v = VISUALS[id] ?? FALLBACK;
+  const cls = 'agent-icon' + (className ? ' ' + className : '');
+  const ext = ICON_EXT[id];
+  if (ext) {
+    return (
+      <img
+        src={`/agent-icons/${id}.${ext}`}
```
lefarcen, line 73: P1 The AgentIcon implementation now renders bundled image assets/fallback text, but the existing `apps/web/tests/components/AgentIcon.test.tsx` still asserts the old inline-gradient markup; CI is failing on both AgentIcon tests. Please update the tests to assert the new `` path and the neutral fallback behavior so the web test suite passes again.

lefarcen (changes requested):
Hey @Sid-Qin, this is a nicely scoped visual sweep — the lucide mapping keeps the public Icon API stable, and the brand asset direction is much cleaner than maintaining hand-drawn approximations.

I found a couple of release-blocking validation issues to clear before this can land. Both are tied to the dependency/icon refactor rather than the UI polish itself; once CI is green, the shape looks much easier to review forward.

apps/web/package.json
lefarcen, line 36: P1 This dependency addition still leaves the Nix fixed-output dependency hash stale: the build checks are failing with the pnpm-deps hash mismatch after `lucide-react` was added. Please update the repo's Nix dependency hash for the new lockfile closure so `nix flake check` can reproduce the dependency set.

apps/web/src/components/AgentIcon.tsx
lefarcen, line 73: P1 This implementation now renders image assets/fallback text, but the existing `AgentIcon` tests still assert the old inline-gradient SVG markup, so the web test job remains red. Please update `apps/web/tests/components/AgentIcon.test.tsx` to assert the new asset-backed `` path and neutral fallback behavior.

lefarcen (changes requested):
Hey @Sid-Qin, I rechecked the current head after the conflict notice. The visual direction still looks coherent, but the same validation blockers are still present on this head, and the PR is also still dirty against `release/v0.7.0`.

Please resolve the base-branch conflicts, then clear the CI issues below so the next review pass can focus on the UI polish itself.

Sid-Qin:
Thanks for the thorough review @lefarcen 🙏 Force-pushed three fixups onto the same three commits (no fresh commits — the history stays the original 3-commit shape so you can re-review the same diff).

**Addressed both P1 blockers:**

1. **Nix `pnpm-deps` hash bump** — folded into commit 1 (`refactor(web): adopt lucide-react for the inline Icon component`). Picked up the expected `sha256-/C9tl0CY/vbDr369wVowsrEMxhljX0pnW7kGZbz3Fas=` directly from the failing `nix flake check` log and pinned it in both `nix/package-web.nix` and `nix/package-daemon.nix`.

2. **`AgentIcon` tests** — folded into commit 2 (`feat(web): bundle official brand assets for agent icons`). Rewrote `apps/web/tests/components/AgentIcon.test.tsx` to assert the new contract:
   - `qoder` (and any other lobehub-shipped brand) → `` with `class="agent-icon"`
   - `devin` → `` (PNG-only outlier)
   - unknown id → `.agent-icon-fallback` `<span>` containing the uppercased first letter, no inline gradient
   
   Full local `pnpm --filter @open-design/web test` is green: **102 test files, 948 tests, all passing**.

**Conflict resolution:** rebased onto the latest `release/v0.7.0` (which had picked up #1509 and #1514 in the meantime). The overlap with #1514 was real — both PRs refactored `PlatformPicker` + `SurfaceOptions` independently. I dropped my versions wholesale and adopted #1514's implementation (it has stronger ARIA — `useId`, `aria-controls`, `aria-multiselectable`) so this PR now reads as pure additive polish on top of #1514, no competing implementation. The other UI polish (entry view footer, Settings dialog sidebar/seg-control/agent-cards, BETA badge, fidelity card, newproj-body overscroll) is unchanged from the original review.

Ready for another pass when CI clears.

apps/web/src/components/AgentIcon.tsx
```diff
@@ -4,229 +4,63 @@ interface Props {
...
+  devin: 'png',
 };
 
 export function AgentIcon({ id, size = 36, className }: Props) {
-  const v = VISUALS[id] ?? FALLBACK;
+  const cls = 'agent-icon' + (className ? ' ' + className : '');
+  const ext = ICON_EXT[id];
+  if (ext) {
+    return (
+      <img
```
Siri-Ray, line 72: Several of the bundled SVG assets still use `fill="currentColor"` (for example `cursor-agent.svg`, `opencode.svg`, `hermes.svg`, `mimo.svg`, `pi.svg`, and `kilo.svg`), but rendering them through `` here means they are loaded as separate image documents and do not inherit `.agent-icon { color: var(--text) }` from the app CSS. In dark theme those monochrome assets resolve to the SVG document's default black, so the affected agent icons can disappear against the dark `--bg-panel` cards. Please either normalize those SVGs to explicit theme-safe fills, render/mask the monochrome marks in a way that lets CSS color them, or keep those particular assets inline so `currentColor` is applied by the page.


Siri-Ray (commented):
@Sid-Qin thanks for the careful follow-up on the earlier validation blockers. I re-reviewed the current UI/icon diff and found one merge-safe polish issue around theme-aware agent glyph rendering; the rest of the touched ranges looked consistent with the release UI sweep.




Sid-Qin:
One more force-push: explicitly opting out of #1514's PlatformPicker/SurfaceOptions visual treatment in favour of the neutral design language this PR establishes everywhere else.

**Why** — the rest of this PR strips the orange accent from every selected/active state inside the New Project panel (DS picker, agent cards, protocol chips, fidelity cards, BETA badge, toggle rows…). Letting #1514's accent-tinted `.platform-dropdown-count` chip, `.platform-dropdown-item.active` accent fill, and `.surface-checkbox.active` accent border ride alongside would leave two of the most-visible controls in the panel still calling out in orange while the rest of the panel has gone neutral. The Create CTA is the only intentional orange in the panel.

**What changed in this push** (folded into commit 3, no new commits)
- `PlatformPicker` now renders through the shared `.ds-picker-trigger` / `.ds-picker-popover` / `.ds-picker-item` styles (single neutral row + popover with check marks). It **keeps** #1514's ARIA improvements verbatim: `useId()` for the listbox id, `aria-haspopup`, `aria-expanded`, `aria-controls`, `role="listbox"`, `aria-multiselectable="true"`. The accessibility contract is unchanged.
- `SurfaceOptions` swaps the bordered `.surface-checkbox` cards for the `.compact-toggle` one-line switches the rest of the panel uses (hint goes to a native `title` tooltip). The OS-widget metadata gating change from #1514 is preserved — the prop interface (`{ includeLandingPage, includeOsWidgets, onIncludeLandingPage, onIncludeOsWidgets }`) is untouched.
- CSS: dropped `.platform-dropdown-*` and `.surface-checkbox-*` rules (~140 lines, all accent-tinted), added `.compact-toggle-*` rules back.
- `tests/components/NewProjectPanel.test.tsx`: the OS-widgets metadata test was querying for `role: 'checkbox'` (matching #1514's `<button role="checkbox" aria-checked>`). Updated to `role: 'button'` since `.compact-toggle` renders as `<button aria-pressed>`. Same test, same assertions on the resulting metadata payload — just a different selector for the same control. **All 948 web tests pass locally.**
- Reverted CompactToggle / `.compact-toggle*` CSS that I had dropped during the prior conflict resolution.

Happy to discuss if there's a reason #1514 explicitly wanted the accent treatment that I'm missing. Otherwise the goal here is one consistent neutral palette across the panel + the official lucide / brand-asset visual layer this PR introduces.

apps/web/public/agent-icons/opencode.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" fill="currentColor" fill-rule="evenodd" viewBox="0 0 24 24"><path d="M16 6H8v12h8zm4 16H4V2h16z"/></svg>
```
lefarcen: P2 This asset still relies on `fill="currentColor"`, but `AgentIcon` renders it through ``, so the SVG is loaded as a separate image document and will not inherit `.agent-icon { color: var(--text) }` from the page. In dark theme this can render the monochrome mark as black on the dark agent-card background. Please normalize the monochrome agent SVGs with explicit theme-safe fills, or render/mask these particular marks in a way that lets the page color actually apply.

lefarcen (changes requested):
Hey @Sid-Qin, thanks for the careful fixups here — the Nix pnpm-deps hash and the AgentIcon test contract blockers from my earlier review are addressed, and adopting #1514's stronger picker implementation was the right conflict resolution.

There is one remaining UI correctness issue before I can clear my previous CHANGES_REQUESTED: several monochrome brand SVGs still depend on page-inherited currentColor even though AgentIcon now loads them as external images. See the inline note below. Once that is normalized, I can re-check and unblock my prior review.

Sid-Qin:
Thanks @Siri-Ray — sharp catch on the theme-aware glyph rendering. Pushed a fix folded into commit 2 (the brand-assets commit, since that's where the ``-only rendering originated). The PR head stays at 3 commits.

**What was wrong:** Six bundled SVGs (`cursor-agent.svg`, `opencode.svg`, `hermes.svg`, `mimo.svg`, `pi.svg`, `kilo.svg`) use `fill="currentColor"`. Loaded through `` the browser treats them as separate documents, so they can't inherit the app's `.agent-icon { color: var(--text) }` — under a dark theme they'd collapse to the SVG document default (black) against the dark `--bg-panel`, exactly as you described.

**Fix:** `AgentIcon` now branches on a `MONO_ICONS` set. For those six it renders a CSS-masked `<span>` instead:

```tsx
<span
  className="agent-icon agent-icon-mono"
  style={{ width, height, maskImage: `url("/agent-icons/${id}.svg")` }}
  aria-hidden
/>
```

with the matching CSS

```css
.agent-icon-mono {
  display: inline-block;
  background-color: currentColor;
  mask-size: contain;
  mask-repeat: no-repeat;
  mask-position: center;
  /* + -webkit-* prefixed equivalents for Safari */
}
```

The SVG becomes a mask shape, and `background-color: currentColor` paints it in whatever color the surrounding text resolves to — light text on dark theme, dark text on light theme. Same approach Phosphor/Lucide use for theme-aware monochrome glyphs.

**`qoder.svg` was a different case** (it's a mix — a green `#2adb5c` Q glyph layered on a `fill="currentColor"` dark frame). Painting it via mask would lose the green. Instead I baked the `currentColor` frame to its official Qoder color `#111113`, so the brand mark always renders as "green Q on dark gray" regardless of theme — no `currentColor` dependency, still official, and works as a plain ``.

**No change for the other 10 brand SVGs / Devin PNG** — they ship with explicit baked colors and stay as ``.

**Verification:**
- Added `AgentIcon.test.tsx > renders monochrome SVGs as a CSS-masked <span> so they pick up theme color` — explicitly asserts the mono path renders as a masked `<span>` (not an ``).
- Full web test suite: **102 files, 949 tests, all green** locally.

Both reviews' P1 blockers and the new theme-aware polish are now resolved. Ready for another pass.

Sid-Qin:
Thanks @lefarcen — fair point that the SVG file itself shouldn't carry the assumption. Even though `AgentIcon`'s mono branch went through CSS mask (where the SVG's fill is irrelevant — only the alpha shape matters), the raw asset still read as theme-dependent if anyone bypassed the component.

**Belt + suspenders, folded into commit 2:**

1. **SVG files normalized** — replaced `fill="currentColor"` with `fill="#1c1b1a"` (the app's standard ink color) in all six monochrome assets: `cursor-agent.svg`, `opencode.svg`, `hermes.svg`, `mimo.svg`, `pi.svg`, `kilo.svg`. Loading any of these through a plain `` from a future call site now renders a legible dark mark rather than defaulting to the SVG document's black.

2. **CSS mask stays** — `AgentIcon`'s mono branch still renders these via `<span>` + `mask-image` so `background-color: currentColor` paints them in the surrounding theme color. The mask reads the SVG's alpha shape, not its baked fill, so the normalization is invisible to that path.

3. **Why the dual approach:** if our component is always used → mask gives true theme-awareness (dark text under dark theme, light text under light theme). If someone else ``s the asset → it still renders cleanly as a dark mark. Either way the icon is visible.

4. Updated the `MONO_ICONS` comment in `AgentIcon.tsx` to document the contract.

Tests still green: full web suite 102 files / 949 tests, including the new `cursor-agent` mono branch assertion.

Force-pushed onto the same three commits, fixup folded back into the brand-assets commit. PR head is now `f6271a3d`. Ready for another look whenever you have a moment.

lefarcen (approved):
Hey @Sid-Qin, thanks for the quick and thorough follow-up. The previous Nix hash and AgentIcon test blockers are fixed, and the latest mono-icon path now covers both sides of the theme issue: the component uses a CSS mask for theme-aware rendering, while the raw SVG assets no longer carry a `currentColor` dependency if someone loads them directly.

I also re-checked the PlatformPicker / SurfaceOptions update: keeping the ARIA contract while moving the visual treatment onto the neutral shared picker/toggle language is consistent with the rest of this release polish. From my side, the current-head code review is clear, so I’m clearing my earlier changes-requested review. CI still has the packaged smoke jobs in progress, so final merge timing should wait for those and the maintainer/reviewer call.

PR #1523, opened 2026-05-13
fix(daemon+web): emit conversation-created SSE event when routine run starts

MrRockySL:
## Summary

Fixes #1361. Likely also addresses #1502 (see below).

When a Routine is configured to **Reuse an existing project** and the user clicks **Run now** while the project view is already open, the daemon creates a new conversation in the project's database row, but the open `ProjectView` has no way to learn it: the project events SSE stream only carries `file-changed` and `live_artifact*` events, and `ProjectView` reloads conversations only when `project.id` changes. The "Run now" appears to do nothing until the user exits and re-enters the project, where the missing conversations finally show up.

This PR adds a `conversation-created` event to the existing project events SSE stream and has `ProjectView` refetch its conversation list when one arrives. The active conversation is intentionally **not** changed — per maintainer guidance on #1361, auto-switching is left for a separate UX decision.

## Per-bullet against the maintainer's PR checklist (#1361)

- ✅ **The event only targets the affected project** — the daemon broadcasts through the existing `activeProjectEventSinks` map keyed by `projectId`. Sinks for other projects do not receive it. The web consumer additionally checks `evt.projectId !== project.id` before acting.
- ✅ **The current project view refreshes the conversation list when that event arrives** — `handleProjectEvent` calls `listConversations(project.id)` and `setConversations(list)`.
- ✅ **It does not change the active conversation automatically** — no call to `setActiveConversationId` in the new code path. The user keeps their current context.
- ✅ **A small test / coverage path for the reuse-existing-project case** — `apps/web/tests/providers/project-events.test.ts` adds three tests: single delivery, concurrent-runs delivery (#1502 coverage), and malformed-payload defensive behavior. Manual repro was driven end-to-end via the packaged app; recording attached below.

## What changed

| File | Purpose |
|---|---|
| `apps/daemon/src/server.ts` | Renamed the (already-generic) `emitProjectLiveArtifactEvent` to `emitProjectEvent`; after `insertConversation` in the routine `setRunHandler`, broadcast `{ type: 'conversation-created', projectId, conversationId, title, createdAt }` through it. |
| `apps/web/src/providers/project-events.ts` | Added `ProjectConversationCreatedEvent` type, extended the `ProjectEvent` union, added an `addEventListener('conversation-created', ...)` that mirrors the existing `file-changed` / `live_artifact` defensive parsing. |
| `apps/web/src/components/ProjectView.tsx` | `handleProjectEvent` handles `conversation-created` by refetching `listConversations` and calling `setConversations` (no active-conversation change). `projectEventToAgentEvent` returns null for the new type so it doesn't get routed into the live-artifact path. |
| `apps/web/tests/providers/project-events.test.ts` | Three new tests (see above). |

Diff: 4 files, +161 −5.

## Likely also addresses #1502

#1502 ("Cannot switch between multiple concurrent runs from the same Routine inside a project") is the same root cause from a different angle: routine-created conversations exist in the backend but aren't surfaced in the already-open project conversation list. Because this PR fires a `conversation-created` event for **every** new routine conversation regardless of how many runs are in flight, concurrent runs targeting the same reused project should now all become visible in the sidebar as they're created.

Per the discussion on #1361, this PR uses the wording **"Likely also addresses #1502"** rather than the auto-closing `Fixes #1502` keyword: #1502 also has a "which run did Open project intend to focus?" deep-link/focus angle that may need a separate follow-up after this lands. The concurrent-runs case is covered by a dedicated test in this PR.

## Before / after

**Before** — from the original issue, with the project view open: clicking `Run now` produces a SUCCEEDED entry in the routine history but the project's conversation list stays unchanged until the user exits and re-enters the project.

**After** — local build of this branch, same flow: clicking `Run now` causes the new routine conversation to appear in the project sidebar within roughly a second, without the user leaving the project. Clicking `Run now` twice in quick succession adds two new conversations, both visible.

## Verified

- `pnpm guard` — clean
- `pnpm --filter @open-design/web typecheck` — clean
- `pnpm --filter @open-design/web test` — **1016 tests pass**, including 3 new tests in `tests/providers/project-events.test.ts` (single delivery, concurrent-runs delivery for #1502 coverage, defensive malformed-payload handling).
- Manual end-to-end: built via `pnpm exec tools-pack mac build --to app --portable`, installed locally, reproduced the original bug (with the build immediately before this fix), then verified the same flow succeeds with this fix in place. Both single and concurrent Run-now were exercised against an existing reuse-target project.

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -220,6 +220,7 @@ export function projectSplitClassName(workspaceFocused: boolean): string {
 
 function projectEventToAgentEvent(evt: ProjectEvent): LiveArtifactEventItem['event'] | null {
   if (evt.type === 'file-changed') return null;
+  if (evt.type === 'conversation-created') return null;
   if (evt.type === 'live_artifact') {
     return {
       kind: 'live_artifact',
@@ -750,6 +751,28 @@ export function ProjectView({
       setDesignMdRefreshKey((n) => n + 1);
       return;
     }
+    if (evt.type === 'conversation-created') {
+      // A new conversation was inserted into this project by a path the
+      // open project view can't observe through its own state (currently:
+      // Routines "Run now" in reuse-an-existing-project mode, #1361).
+      // Refetch the conversation list so the new entry becomes visible
+      // without requiring the user to leave and re-enter the project.
+      // Deliberately do NOT change the active conversation here — the
+      // user keeps their current context. Auto-switch is a separate UX
+      // decision tracked in #1361.
+      if (evt.projectId !== project.id) return;
+      void (async () => {
+        try {
+          const list = await listConversations(project.id);
+          setConversations(list);
+        } catch {
+          // Defensive: refresh failed (network blip, daemon gone). The
+          // next project mount or another conversation-created event
+          // will retry; no need to surface an error here.
+        }
+      })();
+      return;
+    }
     const agentEvent = projectEventToAgentEvent(evt);
     if (!agentEvent) return;
     setLiveArtifactEvents((prev) => appendLiveArtifactEventItem(prev, agentEvent));
@@ -758,7 +781,7 @@ export function ProjectView({
     // Live artifact events come from chat-turn-emitted artifacts; they
     // also imply the conversation transcript changed.
     setDesignMdRefreshKey((n) => n + 1);
-  }, [onProjectsRefresh, refreshLiveArtifacts]);
+  }, [onProjectsRefresh, refreshLiveArtifacts, project.id]);
   useProjectFileEvents(project.id, daemonLive, handleProjectEvent);
 
   // When the URL points at a specific file, fire an open request so the
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -419,6 +419,24 @@ export function ProjectView({
   // correctly gate new-conversation creation even during async loads.
   const messagesConversationIdRef = useRef<string | null>(null);
   const creatingConversationRef = useRef(false);
+  // Live mirror of the currently-viewed project id. Used to bail out of
+  // the conversation-created async refresh (#1361) if the user switches
+  // projects while the refetch is in flight — the existing project-load
+  // effects use the same kind of cancellation guard.
+  const projectIdRef = useRef(project.id);
+  useEffect(() => {
+    projectIdRef.current = project.id;
+  }, [project.id]);
+  // Monotonic token bumped on every `conversation-created` refresh dispatch.
+  // Two rapid events (e.g. concurrent routine runs against the same reused
+  // project, #1502) can start overlapping `listConversations` calls; if the
+  // later request resolves first with N+1 conversations and the earlier
+  // request resolves afterwards with only N, an unconditional
+  // `setConversations(list)` would drop the newest conversation. Each
+  // dispatch captures the token at start; only the dispatch whose token
+  // still equals `conversationsRefreshTokenRef.current` at await-return is
+  // allowed to apply its result.
+  const conversationsRefreshTokenRef = useRef(0);
   const [creatingConversation, setCreatingConversation] = useState(false);
   const currentConversationHasActiveRun = useMemo(
     () => messages.some((m) => m.role === 'assistant' && isActiveRunStatus(m.runStatus)),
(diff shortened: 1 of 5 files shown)


apps/web/src/providers/project-events.ts
```diff
@@ -6,9 +6,28 @@ export interface ProjectFileChangeEvent {
...
 
+// Emitted by the daemon when a new conversation is inserted into a project
+// from a path the open project view can't observe through its own state
+// (currently: Routines "Run now" in reuse-an-existing-project mode). The
+// open `ProjectView` listens for these so its conversation list stays
+// aligned with the database without requiring the user to leave and re-enter
+// the project. The active conversation is intentionally NOT changed here —
+// the user keeps their current context; auto-switch is a separate UX
+// decision tracked in #1361.
+export interface ProjectConversationCreatedEvent {
```
lefarcen: P2 This introduces a new daemon↔web SSE payload shape only in the web provider while the producer emits an untyped object from `apps/daemon`. The repo review guide treats SSE event shapes as a contract/protocol seam, so please move/export the `conversation-created` payload type from `packages/contracts` (alongside the existing live-artifact SSE payloads) and consume that shared type here. That keeps producer and consumer from drifting as this stream grows.

apps/web/src/components/ProjectView.tsx
```diff
@@ -750,6 +751,28 @@ export function ProjectView({
...
+      // Refetch the conversation list so the new entry becomes visible
+      // without requiring the user to leave and re-enter the project.
+      // Deliberately do NOT change the active conversation here — the
+      // user keeps their current context. Auto-switch is a separate UX
+      // decision tracked in #1361.
+      if (evt.projectId !== project.id) return;
+      void (async () => {
+        try {
+          const list = await listConversations(project.id);
+          setConversations(list);
```
lefarcen, line 800: P2 Two quick `conversation-created` events can start overlapping `listConversations` requests; if the later request returns first with both conversations and the earlier request returns afterward with only the first, this unconditional `setConversations(list)` overwrites the sidebar with stale data. That leaves #1502 reproducible under response reordering. Please gate refresh results with a monotonically increasing request id, serialize/coalesce the refresh, or merge by conversation id so an older response cannot remove a newer conversation.

lefarcen (changes requested):
Hey @MrRockySL, this is a strong targeted fix — the project-scoped event direction matches the issue discussion, and I appreciate that you kept auto-switching out of scope while adding coverage for the concurrent-run angle.

I found two things we should tighten before this lands: one repo-boundary issue for the new SSE shape, and one race that can still drop a concurrent conversation from the sidebar in unlucky response ordering. Details inline.

PerishCode (dismissed):
@MrRockySL — nice targeted fix. I traced the loop end-to-end: in `apps/daemon/src/server.ts` the new `emitProjectEvent` call fires immediately after `insertConversation` in `setRunHandler` (covering both reuse-an-existing-project and new-project modes), the rename from `emitProjectLiveArtifactEvent` keeps the existing live-artifact callers wired, the web provider in `apps/web/src/providers/project-events.ts` adds an `addEventListener('conversation-created', …)` that mirrors the existing `file-changed` / `live_artifact` defensive parsing, and `ProjectView.handleProjectEvent` refetches `listConversations` while explicitly leaving the active conversation alone — matching the #1361 maintainer guidance on not auto-switching. The three new tests in `apps/web/tests/providers/project-events.test.ts` cover the single-event, concurrent-events (#1502 angle), and malformed-payload paths.

Nothing additional from my pass beyond what the existing review threads already raise. Appreciate the care taken to scope the change tightly, document intent inline, and add the #1502 concurrent-runs coverage alongside the #1361 fix — thanks for shipping this!



MrRockySL:
Thanks @lefarcen and Codex — addressed all three in `86a9d2a3` (pushed just now). Quick summary per point:

**Lefarcen P2 (project-events.ts:17) — contracts boundary**

Moved `ProjectConversationCreatedSsePayload` to `packages/contracts/src/sse/chat.ts` alongside `LiveArtifactSsePayload`. The web provider re-exports it under the local `ProjectConversationCreatedEvent` name so existing import shape stays stable. The daemon emit site picks up the same shape via a JSDoc typedef.

**Codex P1 (ProjectView.tsx:767) — project-switch race**

Captured `project.id` at dispatch and added a live `projectIdRef` to re-check after `await listConversations`. If the user switched projects mid-fetch, the result is dropped (mirrors the existing project-load effect's `cancelled` flag pattern).

**Lefarcen P2 (ProjectView.tsx:767) — concurrent-refresh reordering**

Added a monotonic `conversationsRefreshTokenRef` bumped on every dispatch. Each request captures its token and only applies its result if the live ref still equals it. Two rapid `conversation-created` events from concurrent runs can no longer drop the newest conversation when an earlier request resolves last with a stale shorter list.

Both guards are documented inline with comments referencing the review threads. Existing tests unchanged (the new guards are defensive consumer logic, not new event shapes). `pnpm guard` + typechecks + 1016 web tests all green.

lefarcen (changes requested):
Hey @MrRockySL — @PerishCode's pass is helpful context, and the overall direction still looks well scoped.

I re-checked the current head after that approval, and my two existing blocking threads are still unresolved on this commit: the new `conversation-created` SSE payload still lives only in the web provider instead of the shared contracts package, and the overlapping `listConversations` refresh race can still drop a newer concurrent run from the sidebar.

No new findings from this pass — once those two threads are addressed, this should be in good shape.

lefarcen (approved):
Hey @MrRockySL — thanks for the quick follow-up. I re-checked the new head and the two blockers from my prior review are addressed: the `conversation-created` payload now lives in `packages/contracts` and both the project-switch and concurrent-refresh ordering paths are guarded before applying the refreshed conversation list.

No new blockers from my pass. This clears my previous changes-requested review — nice iteration. 🙌

MrRockySL:
Apologies @lefarcen — accidental re-request right after your approval landed 😅 please ignore. Thanks for the careful review + quick re-check 🙌

MrRockySL:
@lefarcen one tiny favor — looks like fork PR authors can't dismiss requested reviewers from this side (permissions limit). Could you click dismiss on the pending re-request when you get a sec? It's holding the merge state on BLOCKED. Sorry for the extra step 🙏

lefarcen:
Done — I removed the pending re-request. Thanks for flagging the fork permissions issue, and no worries about the extra step. 🙌

leilei926524-tech:
Hi @MrRockySL — quick note from the Open Design team.

What you get there:

PR #1532, opened 2026-05-13
fix(web): hide preview-only toolbar controls in source view

algojogacor:
Fixes #1528

## Summary

When switching an HTML artifact from preview to source view, the toolbar still showed presentation-oriented controls: deck navigation (pagination), Tweaks palette, Draw overlay toggle, and viewport presets. These controls only make sense in preview mode and create visual confusion when viewing source code.

## Root Cause

Four toolbar sections in `HtmlViewer` had no mode-awareness:
1. **Deck nav** (line 4999): checked `effectiveDeck` but not `mode`
2. **Tweaks** (lines 5037–5071): always rendered in `viewer-toolbar-actions`
3. **Draw overlay** (lines 5095–5126): always rendered in toolbar
4. **Viewport controls** (lines 5128–5134): always rendered in toolbar

The `boardAvailable` variable at line 4967 already correctly gates comment/board functionality to preview mode, but the toolbar controls that accompany them were not similarly constrained.

## Fix

Added `mode === 'preview'` guards to each preview-only control block:
- `effectiveDeck ?` → `mode === 'preview' && effectiveDeck ?`
- Wrapped Tweaks `<div>` in `{mode === 'preview' ? ( … ) : null}`
- Wrapped Draw + viewport controls in `{mode === 'preview' ? ( … ) : null}`

This follows the existing pattern at line 4967 (`boardAvailable = mode === 'preview' && source !== null`) for mode-gating preview features.

## Changes

- `apps/web/src/components/FileViewer.tsx`: 4 conditional guards added (`+5 −1` lines)

## Testing

- **HTML artifact → switch to Source view**: Deck nav, Tweaks, Draw, and viewport controls are hidden; only Edit and Zoom remain visible ✅
- **HTML artifact → switch back to Preview**: All controls reappear ✅
- **Non-HTML artifacts (images, PDFs, SVGs)**: Unaffected — these use different viewer components ✅
- **Deck artifacts in preview mode**: Deck navigation still visible when `effectiveDeck` is set ✅

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -4996,7 +4996,7 @@ function HtmlViewer({
               {t('fileViewer.source')}
             </button>
           </div>
-          {effectiveDeck ? (
+          {mode === 'preview' && effectiveDeck ? (
             <span
               className="deck-nav"
               role="group"
@@ -5034,6 +5034,7 @@ function HtmlViewer({
           ) : null}
         </div>
         <div className="viewer-toolbar-actions">
+          {mode === 'preview' ? (
           <div className="palette-tweaks-anchor">
             <button
               type="button"
@@ -5069,6 +5070,7 @@ function HtmlViewer({
               onClose={() => setPalettePopoverOpen(false)}
             />
           </div>
+          ) : null}
           <button
             className={`viewer-action${manualEditMode ? ' active' : ''}`}
             type="button"
@@ -5092,6 +5094,7 @@ function HtmlViewer({
             <Icon name="edit" size={13} />
             <span>{t('fileViewer.edit')}</span>
           </button>
+          {mode === 'preview' ? (
           <button
             className={`viewer-action${drawOverlayOpen ? ' active' : ''}`}
             type="button"
@@ -5131,6 +5134,7 @@ function HtmlViewer({
             t={t}
           />
           <span className="viewer-divider" aria-hidden />
+          ) : null}
           <button
             type="button"
             className="icon-only"


apps/web/src/components/FileViewer.tsx
```diff
@@ -5092,6 +5094,7 @@ function HtmlViewer({
             <Icon name="edit" size={13} />
             <span>{t('fileViewer.edit')}</span>
           </button>
+          {mode === 'preview' ? (
           <button
```
lefarcen, line 5098: P1 This conditional now returns several adjacent JSX siblings (`button`, divider, `PreviewViewportControls`, divider) without a wrapper, so TSX will fail to parse. Please wrap the preview-only block in a fragment (`<>...</>`) or a single container before closing the `mode === 'preview'` branch.

apps/web/src/components/FileViewer.tsx
```diff
@@ -5034,6 +5034,7 @@ function HtmlViewer({
           ) : null}
         </div>
         <div className="viewer-toolbar-actions">
+          {mode === 'preview' ? (
```
lefarcen, line 5037: P2 This hides the Tweaks popover in Source view, but if the user switches modes while it is open, `palettePopoverOpen` stays `true` and the popover will reopen unexpectedly when they return to Preview. Please also close that state when entering Source, e.g. call `setPalettePopoverOpen(false)` in the Source tab handler.

lefarcen (changes requested):
Hi @algojogacor, thanks for jumping on #1528. The direction is right: keeping deck navigation, Tweaks, Draw, and viewport presets tied to Preview mode matches the UI split we want. I found one compile blocker and one small state edge case to clean up before this can land — details inline.

apps/web/src/components/FileViewer.tsx
mrcfps, line 5098: This new `mode === 'preview'` branch now wraps the draw toggle and viewport controls, but in `apps/web/src/components/FileViewer.tsx` it returns four adjacent JSX siblings (`<button>`, divider, `<PreviewViewportControls />`, divider) from a single conditional expression. TSX requires that branch to resolve to one parent node, so the current PR head will fail to parse/build as written. Please wrap this preview-only block in a fragment (`<>...</>`) or another single parent before the `: null` so the source-view cleanup still works without breaking compilation.

mrcfps (changes requested):
@algojogacor Thanks for the quick fix here — I rechecked the updated preview/source guards in `HtmlViewer` and found one blocking issue in the changed toolbar branch that still needs a follow-up before this can land.


lefarcen (changes requested):
Hey @algojogacor, I re-reviewed the current head after @mrcfps's blocking review. The intended preview/source split still looks like the right fix, but the current commit still needs changes before it can build safely: the draw/viewport preview-only branch needs a single JSX wrapper, and the Tweaks popover should be closed when entering Source so it does not reopen unexpectedly when returning to Preview. No new inline comments from me here — the actionable threads already point at the affected lines.

lefarcen:
Hey @algojogacor, thank you again for picking this up and for moving quickly on the toolbar cleanup.

I'm going to close this PR as superseded now because #1556 has landed on `main` and closed #1528 with the same source-view toolbar fix, including the regression test coverage. That merged version keeps Edit and Zoom available in Source view while hiding the preview-only deck nav, Tweaks, Draw, and viewport controls.

Appreciate the contribution — the direction here was useful, and the review thread helped clarify exactly which controls should stay visible in Source mode. 🙌

PR #1534, opened 2026-05-13
feat: Add mouse wheel zoom support to preview

xxiaoxiong:
Fixes #1398

## Problem

Preview zoom currently only supports clicking the zoom controls and does not support mouse-wheel or similar direct zoom interactions. This makes zooming feel **slow and unnatural** during design review workflows.

### Current Behavior

- ❌ Users can only zoom via button clicks
- ❌ No mouse wheel support
- ❌ No trackpad gesture support
- ❌ Slower and less fluid than expected for a modern preview surface

### Expected Behavior

- ✅ Support mouse wheel zoom with modifier key
- ✅ Match standard OS/browser zoom conventions
- ✅ Make zoom a high-frequency action feel natural

---

## Solution

Added **mouse wheel zoom support** with **Ctrl/Cmd modifier key** (standard zoom gesture) to both preview components:

1. **LiveArtifactViewer** (line 773-792)
2. **HtmlViewer** (line 4833-4852)

#### Added `useEffect` hooks

```typescript
useEffect(() => {
  const previewBody = previewBodyRef.current;
  if (!previewBody || mode !== 'preview') return;

  const handleWheel = (e: WheelEvent) => {
    // Only zoom when Ctrl/Cmd is pressed (standard zoom gesture)
    if (!e.ctrlKey && !e.metaKey) return;
    
    e.preventDefault();
    
    // deltaY < 0 means scroll up (zoom in), deltaY > 0 means scroll down (zoom out)
    const delta = e.deltaY < 0 ? 10 : -10;
    bumpZoom(delta);
  };

  previewBody.addEventListener('wheel', handleWheel, { passive: false });
  return () => previewBody.removeEventListener('wheel', handleWheel);
}, [mode]);
```

#### Key Features

- **Modifier key required**: Only zoom when **Ctrl** (Windows/Linux) or **Cmd** (Mac) is pressed
- **Scroll direction**:
  - Scroll **up** (deltaY < 0) → Zoom **in** (+10%)
  - Scroll **down** (deltaY > 0) → Zoom **out** (-10%)
- **Prevent default**: `preventDefault()` to avoid page scroll during zoom
- **Cleanup**: Remove event listener on unmount and when mode changes
- **Mode-aware**: Only active in **preview mode** (not in source/code mode)

---

### Zoom Interaction

| Action | Result |
|--------|--------|
| **Ctrl/Cmd + Scroll Up** | Zoom in (+10%) |
| **Ctrl/Cmd + Scroll Down** | Zoom out (-10%) |
| **Scroll without modifier** | Normal page scroll (unchanged) |
| **Switch to source mode** | Wheel zoom disabled |

### Zoom Range

- **Min**: 25%
- **Max**: 200%
- **Increment**: 10% per scroll tick
- **Same as button controls** (no behavior change)

---

### Files Modified

- **1 file changed**
- **40 insertions**

### Breakdown

**apps/web/src/components/FileViewer.tsx**:
- Line 773-792: Added wheel zoom to `LiveArtifactViewer`
- Line 4833-4852: Added wheel zoom to `HtmlViewer`

---

## Benefits

- ✅ **Natural zoom interaction**: Matches browser/OS conventions (Ctrl/Cmd + Wheel)
- ✅ **Faster design review**: High-frequency action feels fluid
- ✅ **Cross-platform**: Works on Windows/Linux (Ctrl) and Mac (Cmd)
- ✅ **No breaking changes**: Button controls still work exactly as before
- ✅ **Mode-aware**: Only active in preview mode (not in source/code)
- ✅ **Smooth incremental zooming**: 10% per scroll tick

---

### Manual Testing Steps

1. Open a file in preview mode
2. Hold **Ctrl** (or **Cmd** on Mac)
3. **Scroll mouse wheel up** → Verify preview zooms in
4. **Scroll mouse wheel down** → Verify preview zooms out
5. Release Ctrl/Cmd → Verify normal scrolling resumes
6. Switch to **source mode** → Verify wheel zoom is disabled
7. Zoom to **25%** → Verify cannot zoom out further
8. Zoom to **200%** → Verify cannot zoom in further
9. Try without modifier key → Verify normal scroll (no zoom)

### Expected Behavior

- ✅ Ctrl/Cmd + Wheel zooms the preview
- ✅ Zoom increments by 10% per scroll tick
- ✅ Zoom stays within 25%-200% range
- ✅ Normal scroll works without modifier key
- ✅ Wheel zoom disabled in source/code mode
- ✅ Button controls still work
- ✅ No console errors

---

**Priority**: P2 (Usability improvement)
**Impact**: Low-risk enhancement with high UX improvement

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -771,6 +771,26 @@ export function LiveArtifactViewer({
     setZoom((z) => Math.max(25, Math.min(200, z + delta)));
   }
 
+  // Mouse wheel zoom support
+  useEffect(() => {
+    const previewBody = previewBodyRef.current;
+    if (!previewBody || mode !== 'preview') return;
+
+    const handleWheel = (e: WheelEvent) => {
+      // Only zoom when Ctrl/Cmd is pressed (standard zoom gesture)
+      if (!e.ctrlKey && !e.metaKey) return;
+      
+      e.preventDefault();
+      
+      // deltaY < 0 means scroll up (zoom in), deltaY > 0 means scroll down (zoom out)
+      const delta = e.deltaY < 0 ? 10 : -10;
+      bumpZoom(delta);
+    };
+
+    previewBody.addEventListener('wheel', handleWheel, { passive: false });
+    return () => previewBody.removeEventListener('wheel', handleWheel);
+  }, [mode]);
+
   async function handleRefresh() {
     if (refreshing) return;
     setRefreshing(true);
@@ -4810,6 +4830,26 @@ function HtmlViewer({
   function bumpZoom(delta: number) {
     setZoom((z) => Math.max(25, Math.min(200, z + delta)));
   }
+
+  // Mouse wheel zoom support
+  useEffect(() => {
+    const previewBody = previewBodyRef.current;
+    if (!previewBody || mode !== 'preview') return;
+
+    const handleWheel = (e: WheelEvent) => {
+      // Only zoom when Ctrl/Cmd is pressed (standard zoom gesture)
+      if (!e.ctrlKey && !e.metaKey) return;
+      
+      e.preventDefault();
+      
+      // deltaY < 0 means scroll up (zoom in), deltaY > 0 means scroll down (zoom out)
+      const delta = e.deltaY < 0 ? 10 : -10;
+      bumpZoom(delta);
+    };
+
+    previewBody.addEventListener('wheel', handleWheel, { passive: false });
+    return () => previewBody.removeEventListener('wheel', handleWheel);
+  }, [mode]);
 
   function clearBoardComposer() {
     setActiveCommentTarget(null);


apps/web/src/components/FileViewer.tsx
```diff
@@ -771,6 +771,26 @@ export function LiveArtifactViewer({
...
+      if (!e.ctrlKey && !e.metaKey) return;
+      
+      e.preventDefault();
+      
+      // deltaY < 0 means scroll up (zoom in), deltaY > 0 means scroll down (zoom out)
+      const delta = e.deltaY < 0 ? 10 : -10;
+      bumpZoom(delta);
+    };
+
+    previewBody.addEventListener('wheel', handleWheel, { passive: false });
```
lefarcen, line 790: P1 This listener is attached to `.viewer-body`, but the user's wheel event normally targets the sandboxed preview iframe at line 1036. Wheel events inside an iframe do not bubble to the parent document, so Ctrl/Cmd + wheel over the actual live artifact will not reach this handler. Please route the wheel gesture from the iframe/preview bridge back to the host, or add another capture mechanism that works while the cursor is over the iframe content.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4811,6 +4831,26 @@ function HtmlViewer({
...
```
lefarcen, line 4850: P1 Same issue here for the HTML preview: the listener is on the parent `.viewer-body`, while the interactive surface is the iframe rendered below. Scrolling with Ctrl/Cmd over the preview content will be handled inside the iframe instead of this parent handler, leaving the primary use case unchanged. Please wire the gesture through the iframe bridge or another mechanism that captures events from the iframe surface.

lefarcen (changes requested):
Hi @xxiaoxiong! 👋 The feature direction is a good fit for the preview workflow, and reusing the existing zoom state/range keeps the behavior nicely scoped. I found one blocker in the event wiring though: the new listener is attached outside the iframe, so the main interaction area still won't deliver wheel events to it. See the inline notes below.

apps/web/src/components/FileViewer.tsx
nettee, line 790: In `LiveArtifactViewer` this effect attaches the `wheel` listener to `previewBodyRef.current`, and the new `HtmlViewer` effect repeats the same pattern a few thousand lines lower. Both viewers render the interactive preview inside an `iframe`, so wheel events over the actual design stay in that iframe browsing context and do not bubble into the parent document listener added here. That means the primary Ctrl/Cmd + wheel path this PR is trying to add still will not work when the cursor is over the preview content. Please move the zoom handling into the iframe bridge, or forward the iframe wheel gesture back to the host before calling `bumpZoom`, and apply that same fix to both viewer implementations.

nettee (changes requested):
I confirmed one blocking issue in the new wheel-zoom wiring: both added handlers listen on the parent preview container, while the primary preview surface is rendered inside an iframe, so Ctrl/Cmd + wheel over the content still will not reach these listeners. 

lefarcen:
Heads-up: #1812 is now also open in this area. Both PRs touch `apps/web/src/components/FileViewer.tsx` for preview wheel zoom tied to #1398; #1812 also adds editable arbitrary percentages plus i18n copy, while this PR is the narrower wheel-zoom-only implementation.

Sharing so the maintainers and both authors can compare approaches before either one lands.

PR #1538, opened 2026-05-13
[codex] Add Cursor Agent auth diagnostics

### LINKED ISSUE #951: cursor agent can not be used
My cursor agent is been scanned as default cli while open design started. 

When I create a project and want design something, open desing tips me like below message:

"agent exited with code 1 Error: Authentication required. Please run 'agent login' first, or set CURSOR_API_KEY environment variable."

So I carried out agent login and sucess. Then I return to open desing but it is the same out going. I tried to restart my computer ( windows 10). But it's not work.

cursor agent version: 2026.05.07-42ddaca
open code version: [open-design-0.5.0-win-x64-setup.exe]

Please

alchemistklk:
Fixes #951

## Why

Cursor Agent can be installed and detectable while still being unusable because the spawned process is not logged in or cannot see a `CURSOR_API_KEY`. That showed up in #951 as a Windows/Cursor Agent run failure where Open Design only surfaced a generic agent execution error.

This PR makes Cursor auth/session state diagnosable before and during a run without storing Cursor credentials in Open Design. `cursor-agent` remains the source of truth for login/API-key state.

## What users will see

In Settings, Cursor Agent can appear as installed but unauthenticated. The agent card shows localized auth-state labels such as "Authentication required" or "Auth status unknown", and the Settings agent test returns Cursor-specific recovery guidance instead of a generic spawn failure.

During a chat/design run, Cursor auth failures are surfaced as `AGENT_AUTH_REQUIRED` with recovery commands: run `cursor-agent login`, confirm with `cursor-agent status`, or make sure `CURSOR_API_KEY` is available to the Open Design process environment.

## Surface area

- [x] **i18n keys** — added new translation keys (see `TRANSLATIONS.md` for the locale workflow)

## Screenshots

Not attached. The UI change is a localized text status on existing Settings agent cards, and the behavior is covered by focused Settings/component and daemon fake-CLI tests.

## Bug fix verification

- Test path that reproduces the bug: `apps/daemon/tests/runtimes/env-and-detection.test.ts`, `apps/daemon/tests/connection-test.test.ts`, and `apps/daemon/tests/chat-route.test.ts`
- Did the test go red on `main` and green on this branch? No. This PR adds coverage for a new diagnostic path and the follow-up Cursor auth wordings rather than a single pre-existing red spec on `main`.
- If a red spec wasn't cheap to write, explain why and what verification you did instead: the original failure depends on Cursor Agent process auth state on a Windows host. The PR uses fake `cursor-agent` binaries to deterministically simulate `status` and run-time stderr failures including `Authentication required`, `Not logged in`, and `[unauthenticated]` output.

## Validation

- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/runtimes/env-and-detection.test.ts tests/connection-test.test.ts tests/chat-route.test.ts`
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/runtimes/env-and-detection.test.ts`
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/connection-test.test.ts -t "Cursor Agent"`
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts -t "Cursor Agent"`
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts`
- `pnpm --dir apps/web exec vitest run -c vitest.config.ts tests/components/SettingsDialog.test.ts`
- `pnpm guard`
- `pnpm typecheck`


lefarcen (commented):
Hey @alchemistklk, thanks for putting this together. The direction looks useful: separating Cursor binary detection from auth state should make the Windows/Cursor failure mode much easier to explain and recover from.

I noticed this is still marked as a draft, so I’m holding the full line-by-line review until it’s ready for review. I’ve also linked the related issue thread so the reporter has context while we review the implementation path.

Once you mark it ready, we’ll do the deeper pass on the daemon auth probe, error classification, Settings surface, and coverage. 🙌

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -1792,7 +1794,15 @@ export function SettingsDialog({
                             <div className="agent-card-body">
                               <div className="agent-card-name">{a.name}</div>
                               <div className="agent-card-meta">
-                                {a.version ? (
+                                {a.authStatus === 'missing' ? (
+                                  <span title={a.authMessage ?? a.path ?? ''}>
+                                    Authentication required
```
lefarcen: P2 These two new Settings labels are hardcoded English strings in an otherwise localized dialog. In non-English locales this will show mixed-language UI exactly on the new user-visible auth state. Please add i18n keys for the missing/unknown auth labels and render them through `t(...)` here instead of string literals.

lefarcen (changes requested):
Hey @alchemistklk, this is moving in a solid direction. The split between executable availability and Cursor auth state is the right shape, and the new fake-CLI coverage does a nice job pinning both the status-probe path and runtime stderr path.

I found one user-visible issue to fix before this lands — see the inline note. Also, the PR body is still missing the template's **What users will see** and **Surface area** sections; those matter here because this changes Settings copy / diagnostics and the chat-run error surfaced to users.

If you want to use an AI assistant to patch the PR body, this prompt should be enough:

> Read the current PR body for nexu-io/open-design#1538 and the template at
> - **Why**: use case + pain being addressed, keeping the Cursor Agent auth/session failure from #951 in view
> - **What users will see**: user perspective — Settings agent cards/test result and chat-run auth error behavior
> - **Surface area**: check every applicable box; this PR touches user-visible Settings/diagnostics and shared web/daemon API contracts
> - **Screenshots**: only if the UI box is checked and a screenshot is practical
> - **Validation**: keep the commands actually run (`pnpm guard`, `pnpm typecheck`, and the targeted daemon/web Vitest commands)
> Then update via `gh pr edit 1538 --repo nexu-io/open-design --body-file -`.

Once the hardcoded Settings labels are localized and the body is filled in, this should be in good shape for the product review pass.

apps/daemon/src/server.ts
```diff
@@ -4199,6 +4198,18 @@ export async function startServer({
       if (agentStreamError) {
         return design.runs.finish(run, 'failed', code ?? 1, signal ?? null);
       }
+      if (
+        code !== 0 &&
+        !run.cancelRequested &&
+        classifyAgentAuthFailure(agentId, agentStderrTail)?.status === 'missing'
```
lefarcen: P2 This chat-run classifier only sees `agentStderrTail`, so Cursor auth failures emitted on stdout (for example a stream-json/error line containing `[unauthenticated]`) still skip the new `AGENT_AUTH_REQUIRED` guidance and fall through to the generic failure path. Please capture/include Cursor stdout tail or classify the parsed Cursor error payload before this close-path fallback.

lefarcen (changes requested):
Hey @alchemistklk, thanks for the quick turn here. The PR body is now filled in and the Settings card labels are routed through i18n — that addresses my previous blocker cleanly.

One auth-classification edge case still needs fixing before I can clear my prior changes-requested review: Cursor can report auth failures on stdout as structured output, while this chat-run path only classifies stderr. I left the remaining item inline so the next pass is focused.

Once that path is covered, I can re-review the current head and clear my block.

lefarcen (approved):
Hey @alchemistklk, thanks for the follow-up. This current head addresses the items from my last review: the PR body is complete, the Settings auth labels now go through i18n, and the chat-run path now includes Cursor stdout/error payloads in the auth classifier with focused coverage.

Approving to clear my previous changes-requested block. Product review and CI still need to finish before this is merge-ready, but the code path I reviewed looks good from my side.

PR #1540, opened 2026-05-13
[codex] Update release light background

PerishCode:
## Summary
- Change the light-mode app background tokens to `#faf9f7`.
- Keep Settings -> Appearance -> Color scoped to accent/detail colors only.
- Add regression coverage for default light/dark backgrounds and for Appearance color not mutating global background variables.

## Validation
- `pnpm exec vitest run -c vitest.config.ts tests/state/appearance.test.ts tests/styles/default-background.test.ts`
- `pnpm guard`
- `pnpm typecheck`
- `git diff --check`

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -7,8 +7,8 @@
   /* Surface palette — neutral app chrome that does not bias generated artifacts.
      Keep warm/brand colors out of preview backgrounds; generated product UI
      must choose its own palette through the active skill/design brief. */
-  --bg: #f6f7f9;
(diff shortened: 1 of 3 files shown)


lefarcen (commented):
Hey @PerishCode, thanks for opening this. The direction looks focused: keeping the release light background token explicit while preserving Appearance Color as an accent/detail-only control makes sense.

I see this is still marked as a draft, so I’m going to hold the full line-by-line review until it’s ready for review. Once you mark it ready, we’ll take a proper pass over the CSS token change and the regression coverage.

PR #1541, opened 2026-05-13
fix(web): remove duplicate select-all entry in file list toolbar

algojogacor:
## Summary

Remove the duplicate `designFiles.selectPage` button from the file list toolbar in `DesignFilesPanel`. The toolbar had two select-all actions appearing side by side (`全选此页` and `全选`), which felt redundant.

## Root Cause

The `df-select-bar` rendered both a page-level toggle (`toggleSelectPage`) and a global select-all (`selectAllFiles`) button. The page-level select-all is already provided by the tri-state header checkbox in the table's `<thead>`, making the extra button redundant.

## Fix

- Removed the `全选此页` button from the `df-select-bar` in `DesignFilesPanel.tsx`
- Removed the unused `designFiles.selectPage` i18n key from all 18 locale files and the `types.ts` type definition
- Preserved the `toggleSelectPage` function, `allPageSelected`, and `somePageSelected` as they remain used by the header checkbox

## Changes

| File | Change |
|------|--------|
| `apps/web/src/components/DesignFilesPanel.tsx` | Remove the `selectPage` button from `df-select-bar` |
| `apps/web/src/i18n/types.ts` | Remove `designFiles.selectPage` type key |
| `apps/web/src/i18n/locales/*.ts` (18 files) | Remove `designFiles.selectPage` translations |

## Testing

- Verified no remaining references to `designFiles.selectPage` in the codebase
- The header checkbox continues to provide page-level select/deselect

### DIFF
--- a/apps/web/src/components/DesignFilesPanel.tsx
+++ b/apps/web/src/components/DesignFilesPanel.tsx
@@ -720,9 +720,6 @@ export function DesignFilesPanel({
                       {t('designFiles.pageInfo', { start: rangeStart, end: rangeEnd, total: sortedFiles.length })}
                     </span>
                     <div className="df-select-bar">
-                      <button type="button" className="df-select-all" onClick={toggleSelectPage}>
-                        {t('designFiles.selectPage')}
-                      </button>
                       {selected.size < sortedFiles.length ? (
                         <button type="button" className="df-select-all" onClick={selectAllFiles}>
                           {t('designFiles.selectAll', { n: sortedFiles.length })}
--- a/apps/web/src/i18n/locales/ar.ts
+++ b/apps/web/src/i18n/locales/ar.ts
@@ -667,7 +667,6 @@ export const ar: Dict = {
   'designFiles.download': 'تحميل',
   'designFiles.downloadSelected': 'تنزيل {n} كـ ZIP',
   'designFiles.clearSelection': 'مسح التحديد',
-  'designFiles.selectPage': 'تحديد الكل في الصفحة',
   'designFiles.selectAll': 'تحديد الكل',
   'designFiles.dropTitle': '⤓ أسقط الملفات هنا',
   'designFiles.dropDesc':
--- a/apps/web/src/i18n/locales/de.ts
+++ b/apps/web/src/i18n/locales/de.ts
@@ -555,7 +555,6 @@ export const de: Dict = {
   'designFiles.downloadSelected': '{n} als ZIP herunterladen',
   'designFiles.deleteSelected': '{n} löschen',
   'designFiles.clearSelection': 'Auswahl aufheben',
-  'designFiles.selectPage': 'Alle auf dieser Seite auswählen',
   'designFiles.selectAll': 'Alle auswählen',
   'designFiles.dropTitle': '⤓ Dateien hier ablegen',
   'designFiles.dropDesc':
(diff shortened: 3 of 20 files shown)


lefarcen (commented):
Hi @algojogacor! 👋

This is a nice, focused fix for the duplicated selection affordance from #1536. Removing the toolbar-level page-select button while keeping the header checkbox for page-scoped selection preserves the useful behavior and makes the toolbar read much more clearly.

I checked the component change plus the i18n/type cleanup and didn’t find any blocking issues from my pass. I’ve requested the routed pool review as well, so we’ll let that review cycle finish before merge decisions. Thanks for keeping the diff tight. 🙌

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -720,9 +720,6 @@ export function DesignFilesPanel({
                       {t('designFiles.pageInfo', { start: rangeStart, end: rangeEnd, total: sortedFiles.length })}
                     </span>
                     <div className="df-select-bar">
```
PerishCode, line 722: **Non-blocking: page-select control loses its accessible name after this removal.**

With the labeled `selectPage` button gone, the only remaining page-level select control is the table-header span at `apps/web/src/components/DesignFilesPanel.tsx:757-774`:

```tsx
<span
  className="df-row-check"
  onClick={toggleSelectPage}
  role="checkbox"
  aria-checked={allPageSelected}
  tabIndex={0}
  ...
>
  {allPageSelected ? '\u2611' : somePageSelected ? '\u25A3' : '\u2610'}
</span>
```

It has `role="checkbox"` and `aria-checked`, but no accessible name — the only visible content is a `\u2611` / `\u25A3` / `\u2610` glyph, which screen readers won't announce as a meaningful label. Before this PR, sighted-but-screen-reader users still had the `全选此页` button as a discoverable, labeled page-select affordance; after this PR, that affordance disappears entirely. This is a WCAG 2.1 4.1.2 (Name, Role, Value) regression rather than a behavior bug — sighted mouse/keyboard users are unaffected — so it's non-blocking, but worth fixing in the same PR since it's directly caused by this removal.

The pre-existing codex thread on this hunk raised the same concern in general terms; this is a concrete fix.

**Suggested change:** keep the `designFiles.selectPage` i18n key (or rename it to e.g. `designFiles.selectPageAria` to make the new role explicit) and apply it as `aria-label` on the header span, so the page-select tri-state has a localized accessible name:

```tsx
<span
  className="df-row-check"
  onClick={toggleSelectPage}
  role="checkbox"
  aria-checked={allPageSelected}
  aria-label={t('designFiles.selectPage')}
  tabIndex={0}
  ...
>
```

That lets you keep the toolbar deduplication win from this PR while preserving the screen-reader entry point that the removed button used to provide. If you'd rather rename the key, the locale strings already removed by this PR can be re-added under the new name in the same edit pass.

PerishCode (commented):
@algojogacor — the duplicate-button removal is well-scoped: the `df-select-bar` change in `DesignFilesPanel.tsx`, the matching `designFiles.selectPage` deletions across all 18 locale files, and the `types.ts` cleanup are consistent (no stray references remain in `apps/web/src/`), and `toggleSelectPage` / `allPageSelected` / `somePageSelected` are correctly kept for the table-header tri-state checkbox.

One non-blocking accessibility concern about what the header checkbox exposes to assistive tech once the labeled button is gone — see inline.


lefarcen:
Hey @algojogacor, thanks again for taking this on and for keeping the original fix focused.

While this PR was waiting, #1550 landed and was merged with the same toolbar cleanup direction, plus a small regression test around the bulk-select behavior. The linked issue #1536 has also been closed after the 0.7.0 release, so this branch is now superseded by the merged fix rather than needing another review round here.

I’m going to close this PR to avoid keeping a duplicate open. Appreciate the quick contribution — the direction was aligned, and the later PR just happened to land first. 🙌

PR #1542, opened 2026-05-13
fix(web): count nested .slide elements in deck preview bridge

Sid-Qin:
Fixes #1530

## Summary
- The deck preview bridge in `apps/web/src/runtime/srcdoc.ts` counted slides via a narrow selector that only accepted `.slide` as a direct child of `.deck`, `.deck-stage`, `.deck-shell`, or `body`. Generated HTML decks (e.g. `html-ppt-retro-quarterly-review`) commonly nest `.slide` under an extra wrapper, so the bridge reported `count: 0` while the artifact's own keyboard handler still navigated normally — the host preview toolbar then rendered `1 / 0`.
- Kept the structured selectors first so decorative `.slide` markup in non-deck pages (icons, badges, code samples) is not counted as deck slides, and fall back to `document.querySelectorAll('.slide')` only when the structured selector returned zero.
- Added `apps/web/tests/runtime/srcdoc-deck-bridge-nested-slides.test.ts`: extracts the deck bridge script from `buildSrcdoc({ deck: true })`, runs it against a JSDOM with 8 nested `.slide` elements, and pins `count === 8`; a companion case pins the structured-first contract for `.deck > .slide` decks so the fallback is only reached for the nested layout.

## Validation
- `pnpm --filter @open-design/web test tests/runtime/` — 9 files, 87 tests pass (including the new file).
- `pnpm guard` — clean.
- Test was confirmed red before the fix (`expected +0 to be 8`) and green after, so the regression is anchored in observable bridge behavior rather than the source diff.

### DIFF
--- a/apps/web/tests/runtime/srcdoc-deck-bridge-nested-slides.test.ts
+++ b/apps/web/tests/runtime/srcdoc-deck-bridge-nested-slides.test.ts
@@ -80,17 +80,19 @@ describe('deck bridge — nested slide markup (#1530)', () => {
     expect(state.count).toBe(8);
   });
 
-  it('still counts slides under the documented containers as direct children', async () => {
+  it('still counts slides under the documented containers as direct children and ignores decorative .slide markup outside them', async () => {
     // Pin the structured-first contract: direct children of `.deck` /
-    // `.deck-stage` / `.deck-shell` / `body` keep working as before, so
-    // the fallback is only reached for nested decks and decorative
-    // `.slide` markup in non-deck pages does not get accidentally
-    // counted just because it sits somewhere in the document.
+    // `.deck-stage` / `.deck-shell` / `body` keep working as before AND
+    // decorative `.slide` markup placed outside any structured container
+    // (e.g. a utility class on a banner graphic) is not pulled in just
+    // because it shares the class name. Without the structured-first
+    // pass a broad `.slide` selector would count 4 here, so this fixture
+    // pins the precedence directly rather than only by docstring.
     const slides = Array.from({ length: 3 }, (_, i) =>
       `<section class="slide">${i}</section>`,
     ).join('');
     const { win, parentPostMessage } = setupDeckBridge(
-      `<div class="deck">${slides}</div>`,
+      `<header><span class="slide" aria-hidden="true">decoy</span></header><div class="deck">${slides}</div>`,
     );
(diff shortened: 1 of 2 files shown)


apps/web/tests/runtime/srcdoc-deck-bridge-nested-slides.test.ts
```diff
@@ -0,0 +1,100 @@
...
+    await new Promise<void>((resolve) => win.setTimeout(resolve, 350));
+    const state = lastSlideState(parentPostMessage);
+    expect(state).toBeDefined();
+    expect(state.count).toBe(8);
+  });
+
+  it('still counts slides under the documented containers as direct children', async () => {
+    // Pin the structured-first contract: direct children of `.deck` /
+    // `.deck-stage` / `.deck-shell` / `body` keep working as before, so
+    // the fallback is only reached for nested decks and decorative
```
lefarcen: P3 This comment says the test proves decorative `.slide` markup outside the structured deck is not counted, but the fixture only contains the three real slides. Consider adding an extra nested `.slide` elsewhere in the document and still expecting `count === 3`; that would pin the structured-first contract more directly.

lefarcen (commented):
Hey @Sid-Qin, this is a nicely scoped fix. The structured-first selector keeps the existing guardrail against decorative `.slide` markup, and the regression test is tied directly to the `1 / 0` bridge behavior from #1530.

I left one small non-blocking test-strengthening note inline; no correctness blocker from my pass.

apps/web/tests/runtime/srcdoc-deck-bridge-nested-slides.test.ts
Sid-Qin: Good catch — pushed in 650e007d. Added a decoy `<span class="slide">` in a sibling `<header>` outside `.deck`, kept the assertion at `count === 3`. Sanity check: temporarily collapsed `slides()` to a bare `document.querySelectorAll('.slide')` and the case fails with `expected 4 to be 3`, so the precedence is now pinned directly rather than only by docstring.

lefarcen (commented):
Hey @Sid-Qin, thanks for the quick follow-up — the new decoy `.slide` fixture pins exactly the precedence gap from my previous note. I re-read the bridge change and the updated test on head `650e007d`; no remaining blockers from my side. We will still wait for the normal human review path before merge.

alchemistklk (approved):

PR #1543, opened 2026-05-13
fix(daemon): finish live artifact runs

prantikmedhi:
Closes #1451.

## Why
Live artifact runs were staying in `Working` even after the artifact had already been registered. That left Orbit/live-artifact generation looking stuck after the useful work was done.

## What users will see
- Live artifact runs finish as soon as the artifact is registered.
- Orbit and other live-artifact flows stop hanging in `Working` after the deliverable exists.
- The run summary reflects completion instead of lingering on the in-progress state.

## Surface area
- [x] Daemon
- [x] Tests
- [ ] UI
- [ ] API
- [ ] Docs
- [ ] None

## Validation
- `git diff --check`
- Not run in this worktree: `vitest` is not installed locally here.


apps/daemon/src/live-artifact-routes.ts
```diff
@@ -125,6 +136,7 @@ export function registerLiveArtifactRoutes(app: Express, ctx: RegisterLiveArtifa
         provenanceJson,
         createdByRunId: toolGrant.runId,
       });
+      finalizeLiveArtifactRun(design, toolGrant.runId);
```
lefarcen: P1 This finalizes the chat run before `emitLiveArtifactEvent` sends the `live_artifact` agent event. `finish()` emits `end` and closes the run SSE clients, so the UI can miss the artifact handoff even though the artifact was created. Emit the live-artifact event first, then finish the run after the client-visible registration event has been queued.

lefarcen (changes requested):
Hi @prantikmedhi — thanks for jumping on the live-artifact lifecycle bug. The extracted helper and focused daemon unit test are a good direction for closing the Working-state gap from #1451.

I found one ordering issue that needs fixing before this lands: the run is finalized before the live-artifact event is emitted, which can make the chat stream close before the client sees the artifact handoff. See the inline note below.

Once that event ordering is corrected, this should be in much better shape. 🙌

apps/daemon/src/live-artifact-routes.ts
Siri-Ray: This call makes the chat run terminal immediately after `/api/tools/live-artifacts/create` writes the artifact, while the agent child process is still owned by the server-side run lifecycle. In `server.ts`, the tool token is revoked from the child `error`/`close` handlers, and `runs.cancel()` returns without killing the child once the run is terminal. Because `finish()` also no-ops on later terminal runs, a process that hangs or exits non-zero after creating the artifact will be shown as succeeded, cannot be canceled through the normal run path, and can keep using its live tool token until the child eventually exits or the token TTL expires. Please keep terminal run finalization in the child-exit lifecycle, or add an explicit handoff here that first revokes the run's tool token and terminates/awaits the child before marking the run succeeded; the artifact-created event/UI state can still indicate that the deliverable exists without making the run terminal early.

Siri-Ray (changes requested):
@prantikmedhi thanks for taking on the stuck Working-state bug here. I reviewed the changed daemon route, the server run lifecycle, and the small web/UI label updates. The label/CSS changes look fine, but the live-artifact finalization still has a lifecycle blocker: it marks the run terminal from the tool route before the agent process has actually exited, which bypasses the existing child-exit ownership of cancellation, token revocation timing, and exit-status classification.


apps/daemon/src/live-artifact-routes.ts
```diff
@@ -1,10 +1,21 @@
...
+export interface RegisterLiveArtifactRoutesDeps extends RouteDeps<'db' | 'design' | 'http' | 'paths' | 'auth' | 'liveArtifacts' | 'projectStore'> {}
+
+export function finalizeLiveArtifactRun(
+  design: RegisterLiveArtifactRoutesDeps['design'],
+  runId: string | undefined,
+): boolean {
+  if (!runId) return false;
+  const run = design.runs.get(runId);
+  if (!run || design.runs.isTerminal(run.status)) return false;
+  design.runs.finish(run, 'succeeded', 0, null);
```
lefarcen: P1 `finish()` makes the run terminal while the agent child is still owned by the server lifecycle. After this point later child `close`/`error` handlers cannot change the status, `runs.cancel()` treats the run as already terminal, and the tool token/event sink stay active until the child path eventually cleans them up. Please keep terminal completion in the child-exit lifecycle, or add an explicit handoff that revokes the tool token and stops/awaits the child before calling `finish()`. The artifact-created event can still tell the UI the deliverable exists without marking the whole run succeeded early.

lefarcen (changes requested):
Hey @prantikmedhi, I re-reviewed the current head after Siri-Ray's review. The UI copy/test polish still looks fine, but the daemon lifecycle blocker is still present, so we're keeping changes requested for now.

The key issue is that the tool route marks the run succeeded while the agent process is still alive. That bypasses the existing child-exit path that owns token revocation, cancellation, and final exit-status classification. Once the artifact-created handoff is separated from terminal run completion, this should be much safer to land.

prantikmedhi:
Pushed b42d8f15: removed the early live-artifact run finalization so artifact creation no longer ends the chat run before child exit. Daemon typecheck passed.

lefarcen (approved):
Hey @prantikmedhi, thanks for the quick turnaround. I re-reviewed the new head and the blocker from my side is addressed: the tool route now emits the live-artifact event without marking the chat run terminal early, so child exit, token cleanup, cancellation, and final status classification stay with the existing server lifecycle.

This clears my previous changes-requested review. One non-blocking housekeeping note before the maintainer pass: the PR body is still missing the newer template sections (Why, user-visible behavior, surface area, bug-fix verification, and validation), so filling those in will make the final review/release check easier. 🙌

apps/daemon/src/server.ts
```diff
@@ -2854,6 +2854,7 @@ export async function startServer({
   });
   registerLiveArtifactRoutes(app, {
     db,
+    design,
```
Siri-Ray, line 2857: This is the only daemon-side change left in the current combined diff, but `registerLiveArtifactRoutes` no longer declares or reads a `design` dependency after the follow-up revert in `apps/daemon/src/live-artifact-routes.ts`. That means the PR no longer changes the live-artifact run lifecycle described in #1451 or in the PR body (`finish a live artifact run as soon as the artifact is registered`), so a run that has materialized the artifact can still remain `Working` until the existing child/watchdog path resolves it. Please either remove this stray dependency and retarget the PR to the UI copy cleanup, or add a daemon lifecycle fix that updates the run state at the correct ownership point without bypassing child-exit cleanup, with coverage for the artifact-registered-but-agent-still-open case.

Siri-Ray (changes requested):
@prantikmedhi thanks for the quick follow-up here. I reviewed the current head against #1451 and the prior lifecycle discussion; the unsafe early-finalization path is gone, but the daemon fix is gone with it, so the PR no longer changes the live-artifact Working-state behavior it is titled/described to fix. The UI label/test changes look fine, but this needs one more daemon pass before it can close the bug.


lefarcen:
Hey @prantikmedhi! 👋

Checking in on this one — @Siri-Ray's review from yesterday pointed out that while the early-finalization concern is resolved, the daemon side of the fix got lost in the revision, so the PR doesn't currently change the live-artifact Working-state behavior it's meant to address.

Also heads-up: the PR has merge conflicts now and will need a rebase.

Let us know if you'd like any guidance on the daemon pass or if you'd prefer we take it from here. Either way works! ❤️

Siri-Ray:
@prantikmedhi I'm holding off on generating review comments for nexu-io/open-design#1543 because this pull request has merge conflicts right now.

PR #1544, opened 2026-05-13
feat(daemon): make design-system token channel default-on (PR-D)

zoeforfun:
## Why

Agents picking the `default` or `kami` brand have to re-derive the brand's color, typography, and component shapes from prose every turn (`DESIGN.md`). PR-C (#1385) shipped a daemon channel that injects the structured `tokens.css` + `components.html` siblings into the system prompt, gated behind `OD_DESIGN_TOKEN_CHANNEL=1` for smoke testing. Smoke evidence on the two structured brands proved the new channel works:

- **`default`** — treatment used 10/10 brand tokens byte-for-byte; control invented its own colors (0/10 match).
- **`kami`** — treatment recovers the brand name (`Kami · 纸`), the two-tier surface (`--bg` parchment + `--surface` ivory), the `:root[lang="zh-CN"]` font stack override, and the `components.html` card pattern. Control invented `"Replica"` as a brand name and used a single flat surface.

Leaving the channel behind an env var means real users keep getting the prose-only prompt unless they remember to set the flag. This PR flips the default so the new behaviour is what every chat sees by default.

```diff
-      if (process.env.OD_DESIGN_TOKEN_CHANNEL === '1') {
+      if (isDesignTokenChannelEnabled()) {  // env !== '0'
```

## What users will see

Picking `default` or `kami` in the design-system selector and starting a chat:

- The agent now consistently uses the brand's tokens (colours, type scale, surface tiers, accent restraint, CN font stack on zh-CN artifacts) instead of inventing values from prose.
- Component shapes in the artifact match the brand's `components.html` fixture (button structure, card structure, type-scale rhythm).
- No UI affordance changes — there's nothing new to click. The improvement shows up purely as higher-fidelity output for the two brands that ship the structured form today.

For the other 147 brands (which don't ship `tokens.css` / `components.html`), nothing changes — the system prompt is byte-identical to today's, pinned end-to-end by `scripts/check-design-system-flag-parity.ts` (new in this PR, registered in `pnpm guard`).

The kill switch `OD_DESIGN_TOKEN_CHANNEL=0` reverts the daemon to the pre-PR-C DESIGN.md-only path with no migration.

## Surface area

- [ ] **UI** — no
- [ ] **Keyboard shortcut** — no
- [x] **CLI / env var** — `OD_DESIGN_TOKEN_CHANNEL` semantics flip: previously `=1` was the explicit opt-in; now the var is unread unless set to the literal string `0` (kill switch). Documented in the daemon `server.ts` comment block above the gate.
- [ ] **API / contract** — no
- [ ] **Extension point** — no
- [ ] **i18n keys** — no
- [ ] **New top-level dependency** — no
- [x] **Default behavior change** — yes, this is the entire point: brands that ship `tokens.css` + `components.html` (today: `default`, `kami`) now get the structured channel injected by default.
- [ ] **None**

## Screenshots

n/a — no UI surface. Smoke evidence (artifact screenshots for `default` and `kami`, control vs treatment) lives in #1385's description and the comparison comment posted there.

## Validation

- `pnpm guard` — all 13 checks pass, including the new `design system flag parity` check
  - `Design system flag parity passed: 147 prose-only brands produce byte-identical prompts under OD_DESIGN_TOKEN_CHANNEL flag-off vs flag-on; 2 structured brands show the expected token-channel divergence.`
- `pnpm exec tsc -p scripts/tsconfig.json --noEmit` — clean
- `pnpm --filter @open-design/daemon exec vitest run tests/prompts/system.test.ts tests/design-system-assets.test.ts` — 32/32 pass (incl. 6 new tests pinning the `OD_DESIGN_TOKEN_CHANNEL` env-flag predicate)

## Review fixes (round 2 — `f1187cfc`)

Addressed @nettee's two blocking findings on the parity guard:

1. **Inventory readability.** `fileExists()` and the bare `readdir` catch in the guard script swallowed every fs error as "absent", which would let `pnpm guard` pass green after checking 0 brands when `design-systems/` is unreadable. Both paths now only treat `ENOENT` / `ENOTDIR` as absence and rethrow every other failure, mirroring the same fix already applied to `apps/daemon/src/design-systems.ts:readFileOptional` after PR-C review.
2. **Server-layer gate coverage.** The guard exercised `composeSystemPrompt` directly, so a regression that restored `=== '1'`, typo'd the env name, or stopped reading assets when the var is unset would have stayed green. Extracted the predicate into `isDesignTokenChannelEnabled(env)` (re-used by `server.ts` at the gate) and added 6 unit tests in `tests/design-system-assets.test.ts` pinning every input that matters: unset / `'1'` / `'true'` / empty string / `'0'` / whitespace-padded `'0'`. The default-on flip itself is now covered.

## Rollback

Set `OD_DESIGN_TOKEN_CHANNEL=0` on the daemon process. No data migration, no schema change, immediate effect on next chat turn.


lefarcen (changes requested):
Hey @zoeforfun, the default-on token channel direction looks well-contained: the runtime flip stays tiny, the rollback knob is clear, and the new guard checks the prose-only vs structured-brand contract at the right seam. I did not find a code-level blocker in the daemon path or parity script.

One process blocker before this can be merge-ready: because this is a feature PR, please update the PR body to the current template shape so product/release review can see the user-facing scope quickly. The missing pieces are:

- **Why** — keep the smoke evidence, but put it under the template's Why section.
- **What users will see** — explain the end-user/product effect of the token channel being default-on.
- **Surface area** — check the applicable boxes; this looks like default behavior / daemon prompt behavior, and product can confirm whether any UI/API surface should be marked too.

If you want to hand this to an assistant, this prompt should get it there:

> Read the current PR body for nexu-io/open-design#1544 and the template at
> Then update via `gh pr edit 1544 --repo nexu-io/open-design --body-file -`.

Once the body is aligned and product has had a look, this should be straightforward to continue reviewing.

mrcfps (dismissed):

lefarcen (changes requested):
Hey @zoeforfun, re-reviewed the current head after @mrcfps's approval. The daemon/code side still looks clean from my side — the default-on flip is scoped and the parity guard is covering the legacy-brand promise.

I’m keeping my changes-requested review open only for the PR body/template gap called out above: please fill the current template’s **Why**, **What users will see**, and **Surface area** sections so product/release review has the user-facing scope in the standard place. After that update, this blocker should be ready to clear.

scripts/check-design-system-flag-parity.ts
```diff
@@ -0,0 +1,222 @@
...
+  // composeSystemPrompt matches the daemon's runtime title exactly.
+  return raw.replace(/^Design System (Inspired by|for)\s+/i, "").trim();
+}
+
+async function discoverBrandSnapshots(): Promise<BrandSnapshot[]> {
+  let entries;
+  try {
+    entries = await readdir(designSystemsRoot, { withFileTypes: true });
+  } catch {
+    return [];
```
nettee: This guard currently degrades to a vacuous pass when the inventory becomes unreadable. `readdir(designSystemsRoot)` failures here return `[]`, and `fileExists()` above also turns any `stat()` failure into `false`, so a renamed/unreadable `design-systems/` tree or a permission error on a brand `DESIGN.md` would make `pnpm guard` succeed after checking 0 brands. That hides exactly the misconfiguration this rollout check is supposed to catch. Please only treat true absence (`ENOENT` / `ENOTDIR`) as skippable and rethrow every other filesystem error so CI fails loudly when the parity inventory cannot be read.

scripts/check-design-system-flag-parity.ts
```diff
@@ -0,0 +1,222 @@
...
+    // the top of this loop. For prose-only brands that's
+    // `{ tokensCss: undefined, fixtureHtml: undefined }`, which the
+    // composer already skips; for structured brands it's the verbatim
+    // file contents.
+    const flagOn = composeSystemPrompt({
+      designSystemBody: brand.designMd,
+      designSystemTitle: brand.title,
+      designSystemTokensCss: brand.assets.tokensCss,
+      designSystemFixtureHtml: brand.assets.fixtureHtml,
+    });
```
nettee, line 186: This does not actually pin the PR-D behavior that changed in `apps/daemon/src/server.ts`. Both `flagOff` and `flagOn` call `composeSystemPrompt(...)` directly, so the new `process.env.OD_DESIGN_TOKEN_CHANNEL !== '0'` branch is never exercised. If the server regressed back to `=== '1'`, used the wrong env var, or stopped passing assets when the flag is unset, this check would still stay green because the composer divergence is unchanged. Please add at least one test/guard path that runs the same prompt assembly through the server-layer env gate (unset vs `0`) so the default-on flip itself is covered.

nettee (changes requested):
I found two blocking issues in the new parity guard for the default-on rollout. The inline comments cover the places where the current check can still pass without validating the behavior this PR flips by default.


lefarcen (changes requested):
Hey @zoeforfun, I re-read the current head after @nettee's review. I agree with the two blocking guard findings they left: the new parity check should fail loudly on unreadable inventory, and it still needs coverage for the server-layer `OD_DESIGN_TOKEN_CHANNEL` default-on gate itself.

I do not have extra inline comments beyond those threads and the PR-body/template blocker from my earlier review, but my changes-requested state should stay in place until those are addressed. Once the guard covers the runtime gate and the PR body is aligned, I can clear my side on the next update.

apps/daemon/tests/design-system-assets.test.ts
```diff
@@ -112,3 +112,41 @@ describe('readDesignSystemAssets', () => {
...
+
+// Reviewer feedback (nettee, PR-D #1544): the parity guard at
+// `scripts/check-design-system-flag-parity.ts` exercises the prompt
+// composer directly and therefore does NOT cover the server-layer env
+// gate that PR-D actually flipped — a future regression that restored
+// `=== '1'`, used a typo'd env name, or stopped reading assets when
+// the var is unset would still let the guard pass green. These tests
+// pin the predicate that wraps the gate so the default-on flip itself
+// is locked into the test suite.
+describe('isDesignTokenChannelEnabled (PR-D env gate)', () => {
```
lefarcen, line 128: P2 This still only tests the extracted predicate, not the server prompt-assembly path that PR-D changed. If `server.ts` stopped using `isDesignTokenChannelEnabled()` or restored an inline `process.env.OD_DESIGN_TOKEN_CHANNEL === '1'` gate, these tests would stay green because they never exercise the asset-read branch with the env unset. Please add a focused test/helper that runs the same server-layer asset-resolution path for unset vs `'0'`, so the default-on behavior is pinned where the daemon consumes it.

lefarcen (changes requested):
Hey @zoeforfun, thanks for the quick follow-up — the PR body is now in the right template shape, and the inventory-readability fix addresses the vacuous-pass concern cleanly.

I still see one remaining blocker around the default-on coverage. The new predicate tests are useful, but they do not yet exercise the server prompt-assembly branch that actually reads assets for an unset env var. I left one inline comment with the gap. Once that is covered, I should be able to clear my previous changes-requested review on the next push.

lefarcen (dismissed):
Hey @zoeforfun, this addresses my previous blocker. The asset-resolution seam now covers the env gate plus built-in/user-installed fallback against real disk fixtures, and `server.ts` consumes that helper directly, so I’m clearing my changes-requested review from my side.

There are still repo gates outside my review to clear before merge — current CI is not fully green, and product/pool review are still pending — but the specific issues I raised are resolved.

nettee (approved):

Issue #433, opened 2026-05-04
OpenDesign Launcher for Windows


Waxieeee:
Hey! I built a small Windows batch script that solves a daily annoyance with the OpenDesign dev server, every time you start it, it binds to a randomized port, meaning you have to hunt for the URL in the terminal and manually paste it into your browser.
This script starts the server, watches its output, automatically detects whatever port it lands on, and opens Brave Browser directly to the right URL. Zero manual steps after double-clicking.
Might be useful to link from the repo or docs for Windows users getting started.

https://github.com/Waxieeee/Auto-Launch-OpenDesign-in-Windows

lefarcen:
Hi @Waxieeee! 👋 Thanks for sharing this — auto-detecting the dev server port and opening the browser is exactly the kind of small quality-of-life win that makes daily work smoother.

We don't currently have Windows-specific launcher tooling documented in the repo, so this fills a real gap for Windows users getting started. I'll flag this for the maintainers to consider linking from the docs or README.

Appreciate you taking the time to build and share this!

lefarcen:
Hey @Waxieeee — quick follow-up after checking this against the current repo docs.

The launcher idea still fits a real Windows onboarding gap: `tools-dev` uses available ports by default, and the Windows guide currently tells users to read the printed URL manually. We do also document fixed ports via `pnpm tools-dev run web --daemon-port ... --web-port ...`, so the most useful repo-facing version would probably be a short Windows docs note that covers both options:

- use fixed ports when users want a predictable URL;
- optionally use a small launcher script for the “double-click and open the detected URL” workflow.

Would you be interested in opening a small docs PR for this? `docs/windows-troubleshooting.md` is probably the best home, and we can help review it. One thing I’d suggest for the PR: align the example with the current `pnpm tools-dev run web` entry point rather than starting `apps/daemon` and `apps/web` separately, so the docs stay consistent with the supported lifecycle command.

No pressure if you’d rather keep it as an external helper — the issue can stay open as an enhancement note either way. 🙌

lefarcen:
Hey @Waxieeee — a docs PR is now open for this in #1546.

It takes the Windows onboarding angle we discussed here: documenting the supported `pnpm tools-dev run web` entry point, plus a small launcher-style snippet for users who want the browser to open automatically from the printed local URL. We'll review the details in the PR, and this issue should stay open until that lands.

lefarcen:
Hey @Waxieeee — quick update: #1546 has landed.

[...]

PR #1546, opened 2026-05-13
docs: add Windows launcher note

prantikmedhi:
Related to #433.

## Why
Windows users benefit from a single place in the repo that points them to the existing launcher and the supported dev-server flow, instead of hunting through old notes.

## What users will see
- A short Windows onboarding note in README
- A link to the launcher from `docs/windows-troubleshooting.md`
- A clearer path to the supported dev-server startup flow

## Surface area
- [x] Docs
- [ ] UI
- [ ] API
- [ ] Daemon
- [ ] Tests
- [ ] None

## Screenshots
- N/A

## Validation
- `git diff --check`
- Local smoke check still isn’t runnable here because the repo dev deps aren’t installed in this worktree

### DIFF
--- a/docs/windows-troubleshooting.md
+++ b/docs/windows-troubleshooting.md
@@ -203,6 +203,18 @@ python --version   # or py --version
 Get-ExecutionPolicy -List
 ```
 
+## 7. Optional: quick launcher
+
+If you want a double-click entry point on Windows, create a `launch.bat` file in the repo root with:
+
+```bat
+@echo off
+cd /d %~dp0
+corepack pnpm tools-dev run web
+```
+
+That keeps the launcher on the supported `pnpm tools-dev run web` path while still giving you a one-click start.
+
 ---
 
 ## Optional: OpenCode agent CLI on Windows
--- a/README.md
+++ b/README.md
@@ -374,6 +374,8 @@ pnpm tools-dev run web
 
 Environment requirements: Node `~24` and pnpm `10.33.x`. `nvm`/`fnm` are optional helpers only; if you use one, run `nvm install 24 && nvm use 24` or `fnm install 24 && fnm use 24` before `pnpm install`.
 
+Windows users can follow [`docs/windows-troubleshooting.md`](docs/windows-troubleshooting.md) for the native setup path and a tiny double-click launcher.
+
 For desktop/background startup, fixed-port restarts, and media generation dispatcher checks (`OD_BIN`, `OD_DAEMON_URL`, `apps/daemon/dist/cli.js`), see [`QUICKSTART.md`](QUICKSTART.md).
 
 The first load:


docs/windows-troubleshooting.md
```diff
@@ -203,6 +203,18 @@ python --version   # or py --version
...
 ```
 
+## 7. Optional: quick launcher
+
+If you want a double-click entry point on Windows, create a `launch.bat` file in the repo root with:
+
+```bat
+@echo off
+cd /d %~dp0
+corepack pnpm tools-dev run web
```
lefarcen, line 213: P2 This batch file starts `tools-dev`, but it does not detect the printed port or open the browser, which is the core behavior requested in #433. Because the PR says `Closes #433` and the README advertises a launcher, please either add an actual auto-open launcher example (for example, one that waits for the local URL and opens it) or narrow the PR wording/linkage to a simple start shortcut and use a non-closing reference to #433.

lefarcen (changes requested):
Hi @prantikmedhi! 👋 Thanks for picking this up — keeping the docs on the supported `pnpm tools-dev run web` path is the right direction, and the Windows onboarding page is a good place for this guidance.

I found one scope mismatch that we should fix before this lands: the PR currently says it closes #433, but the added launcher is a start shortcut rather than the auto-open/detect-the-port workflow requested there. I also noticed the PR description is missing a few template sections (`Why`, `What users will see`, `Surface area`, and `Validation`); filling those in will make the next pass much faster.

If you are using an AI assistant, you can paste this prompt to update the PR body:

> Read the current PR body for nexu-io/open-design#1546 and the template at
> - **Validation**: what you actually ran (`git diff --check`, plus any repo checks you can run)
> Then update via `gh pr edit 1546 --repo nexu-io/open-design --body-file -`.

Once the launcher wording/scope and PR body are aligned, this should be a small docs-only review from there. 🙌

prantikmedhi:
Updated the PR body to match the template and clarified the scope: this is a docs pointer to the existing Windows launcher, not the launcher implementation itself.

docs/windows-troubleshooting.md
lefarcen, line 213: P2 This still only launches `pnpm tools-dev run web`; it does not watch the output, detect the random local URL, or open Brave. Since the PR body still says `Fixes #433` and describes that auto-open behavior, please either replace this with/link to a launcher that actually does that, or make the body and references clear that this is just a docs pointer/start shortcut rather than the implementation that closes #433.

lefarcen (changes requested):
Hey @prantikmedhi, thanks for updating the body and clarifying the intended scope. The template shape is much easier to review now.

One blocker remains: the PR body still says `Fixes #433` and says the launcher auto-detects the server port and opens Brave, but the diff still only documents a `.bat` file that starts `tools-dev`. That mismatch would leave readers expecting behavior the docs do not actually provide. Please either document/link an actual auto-open launcher, or narrow the PR body/closing keyword to say this is only a docs pointer/start shortcut and does not close #433 yet.

Once that scope is consistent, this should be straightforward to re-check. 🙌

prantikmedhi:
Narrowed the scope so the PR matches the diff: docs pointer only, no issue close claim.

lefarcen (approved):
Hey @prantikmedhi, thanks for tightening the scope. The PR body now matches the actual docs-only diff: it uses a non-closing `Related to #433` reference and no longer claims the launcher auto-opens the random port.

The added README pointer plus the Windows troubleshooting snippet stay on the supported `pnpm tools-dev run web` path, so this is clean from my side now. 🙌

PR #1550, opened 2026-05-13
fix(web): remove redundant bulk-select button

prantikmedhi:
Closes #1536\n\nRemoves the duplicate page-select button from the file list toolbar. The table header checkbox still handles page selection, while the toolbar keeps the all-files action and clear-selection state.

### DIFF
--- a/apps/web/src/components/DesignFilesPanel.tsx
+++ b/apps/web/src/components/DesignFilesPanel.tsx
@@ -720,9 +720,6 @@ export function DesignFilesPanel({
                       {t('designFiles.pageInfo', { start: rangeStart, end: rangeEnd, total: sortedFiles.length })}
                     </span>
                     <div className="df-select-bar">
-                      <button type="button" className="df-select-all" onClick={toggleSelectPage}>
-                        {t('designFiles.selectPage')}
-                      </button>
                       {selected.size < sortedFiles.length ? (
                         <button type="button" className="df-select-all" onClick={selectAllFiles}>
                           {t('designFiles.selectAll', { n: sortedFiles.length })}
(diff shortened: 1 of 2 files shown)


lefarcen (commented):
Hi @prantikmedhi! 👋 The code direction here is clean: removing the toolbar-level page-select button keeps the page selection behavior in the table header checkbox, which matches the issue’s UI clarity goal.

No code blockers from my pass on the two changed files. One process note before the next review iteration: the PR body is missing a few sections from the repository template, which helps reviewers quickly see user impact and validation. Could you fill in these sections?

Missing right now:
- **Why**
- **What users will see**
- **Surface area**
- **Validation**

If you’re using an AI assistant, you can paste this prompt to update the description:

> Read the current PR body for nexu-io/open-design#1550 and the template at
> Then update via `gh pr edit 1550 --repo nexu-io/open-design --body-file -`.

Also noting for context: PR #1541 is already open against the same issue, so maintainers may consolidate between the two. No rush — this just helps the next review iteration go faster 🙏

PerishCode (approved):
@prantikmedhi nice tight fix — dropping the toolbar `Select page` button leaves the table-header checkbox as the single page-select surface and keeps the toolbar focused on `Select everything` and `Clear selection`, which matches what #1536 was asking for. Verified `toggleSelectPage` is still wired to the `df-row-check` header at apps/web/src/components/DesignFilesPanel.tsx:759/766 so the page-toggle behavior is preserved, and the new regression test in apps/web/tests/components/DesignFilesPanel.test.tsx asserts both the positive (`Select everything` present) and negative (`Select all on page` absent) side. Nice touch keeping the change XS — appreciate the focus!

PR #1551, opened 2026-05-13
fix(web): make file grouping by kind work

prantikmedhi:
Fixes #1548.

## Why
The kind grouping control looked selectable, but the file list stayed flat. That made the option feel broken.

## What users will see
- The file list now splits into sections by file kind
- HTML, image, code, and other file types get their own headers
- The default Kind view finally matches the label in the toolbar

## Surface area
- [x] UI
- [x] Tests
- [ ] API
- [ ] Daemon
- [ ] Docs
- [ ] None

## Screenshots
- N/A

## Validation
- `git diff --check`
- Not run in this worktree: `pnpm vitest run apps/web/tests/components/DesignFilesPanel.test.tsx`

### DIFF
--- a/apps/web/src/components/DesignFilesPanel.tsx
+++ b/apps/web/src/components/DesignFilesPanel.tsx
@@ -501,6 +501,29 @@ export function DesignFilesPanel({
     });
   }
 
+  function renderKindSections() {
+    const grouped = new Map<ProjectFileKind, ProjectFile[]>();
+    for (const file of pageFiles) {
+      const next = grouped.get(file.kind) ?? [];
+      next.push(file);
+      grouped.set(file.kind, next);
+    }
+
+    return [...grouped.entries()]
+      .sort(([a], [b]) => kindSortPriority(a) - kindSortPriority(b))
+      .flatMap(([kind, kindFiles]) => [
+        <tr className="df-section-row" key={`${kind}-label`}>
+          <td colSpan={6}>
+            <div className="df-section-label">
+              <span>{kindLabel(kind, t)}</span>
+              <span className="df-section-count">{kindFiles.length}</span>
+            </div>
+          </td>
+        </tr>,
+        ...kindFiles.map(renderFileRow),
+      ]);
+  }
+
   async function handleBatchDownload() {
     const fileList = [...selected];
     if (fileList.length === 0) return;
@@ -807,7 +830,9 @@ export function DesignFilesPanel({
                     <tbody>
                       {groupMode === 'modified'
                         ? renderModifiedSections()
-                        : pageFiles.map(renderFileRow)}
+                        : groupMode === 'kind'
+                          ? renderKindSections()
+                          : pageFiles.map(renderFileRow)}
                     </tbody>
                   </table>
                   <div className="df-pagination df-pagination-center">
--- a/apps/web/tests/components/DesignFilesPanel.test.tsx
+++ b/apps/web/tests/components/DesignFilesPanel.test.tsx
@@ -129,8 +129,11 @@ describe('DesignFilesPanel grouping', () => {
       file({ name: 'chart.png', kind: 'image', mime: 'image/png' }),
     ]);
 
-    expect(screen.getByText('HTML page')).toBeTruthy();
-    expect(screen.getByText('Image')).toBeTruthy();
+    const sectionLabels = Array.from(
+      document.querySelectorAll<HTMLElement>('.df-section-label'),
+    ).map((el) => el.textContent ?? '');
+    expect(sectionLabels.some((text) => text.includes('HTML page'))).toBe(true);
+    expect(sectionLabels.some((text) => text.includes('Image'))).toBe(true);
     expect(screen.getByTestId('design-file-row-page.html')).toBeTruthy();
     expect(screen.getByTestId('design-file-row-chart.png')).toBeTruthy();
     expect(screen.queryByText('Today')).toBeNull();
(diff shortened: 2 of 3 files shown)


lefarcen (commented):
Hi @prantikmedhi! I reviewed the DesignFilesPanel change. The implementation is nicely scoped: it reuses the existing kind labels/sort priority, keeps grouping limited to the current paginated page like the modified-date path, and adds a regression test for the default Kind view. No code blockers from my pass.

One non-blocking PR-body note: before the next pass, please refresh the template details so reviewers can verify the UI change quickly — especially the UI surface/validation bits. If you can, add a small screenshot or GIF of the Kind grouping; and for validation, list the exact web test command once it has run, or simply mark that the web test suite was not run yet.

Leaving this as a comment rather than blocking since the code change itself looks focused. 🙌

prantikmedhi:
Updated the PR body with the missing template bits and added the exact web test command as not run in this worktree. If you want, I can still add a screenshot next.

lefarcen (commented):
Hey @prantikmedhi, thanks for the follow-up push. I re-read the current head and the actual code/test diff still looks clean from my side: the Kind path now renders section headers with the existing kind labels, and the regression test covers the default Kind grouping without disturbing the modified-date grouping path.

On your screenshot question: a small before/after screenshot or GIF would be helpful for the UI history, but I’m not blocking on it. The more important remaining bit is bug-fix verification — once you can run the targeted web test, please update the PR body with the result. CI is currently red in workspace typecheck on an existing FileViewer test fixture (`projectKind` missing), not in this changed file path, so I’m leaving this as another comment rather than requesting changes. 🙌

mrcfps (dismissed):
@prantikmedhi I reviewed the changed ranges in `DesignFilesPanel` and the paired test update. The kind-grouping path now renders section headers with the existing kind labels and priority order, and the regression test covers the default Kind view without changing the modified-date grouping flow. I couldn't rerun the targeted Vitest in this worktree because `node_modules` are missing here, but I didn't find actionable issues in the code changes themselves. Thanks for tightening this up 🙌


lefarcen:
Heads-up: there are two other open PRs targeting the same file-grouping feature in `DesignFilesPanel.tsx`:

- PR #1552 by @xxiaoxiong — adds i18n keys across all 18 locales for the grouping labels
- PR #1561 by @mturac — similar scope to #1552, also touches `MemorySection.tsx`

Yours is the most scoped of the three (just `DesignFilesPanel.tsx` + its test). Sharing so the three authors + maintainer team can coordinate on which approach to land.

lefarcen:
Heads-up: two other open PRs are targeting the same `DesignFilesPanel.tsx` kind-grouping fix for #1548 — PR #1561 by @mturac (which also updates i18n locale files and `i18n/types.ts`) and PR #1632 by @xxiaoxiong (currently blocked on CI: missing i18n keys). You three might want to compare approaches; the maintainer team will pick which one lands. Sharing so no effort gets wasted.

lefarcen:
Hey @prantikmedhi — pushed a small follow-up commit (`a91a7bc8`) to unblock CI on this PR. All three changes are in tests only, no product code touched:

- `FileViewer.test.tsx` line 434 was the typecheck failure — missing `projectKind="prototype"` on the deck-render case. Every other render in that file already passes it; this one slipped through. That was the actual blocker on `Validate workspace`.
- `DesignFilesPanel.test.tsx` "groups files by kind" — `screen.getByText('HTML page')` was hitting both the section header and the per-row kind cell, so the test failed with `multipleElementsFound` once the typecheck didn't gate it anymore. Switched to querying `.df-section-label` elements directly.
- `DesignFilesPanel.test.tsx` "passes every selected file to batch delete" — this was a pre-existing test from `main`, but with kind-grouping now default, `rows[0]` and `rows[1]` are sorted by kind priority (html → sketch → image), so the hard-coded `['file-1.html', 'file-2.png']` no longer matches. Rewrote the assertion to derive expected names from the row testids and use `arrayContaining`, so it stops depending on row order.

Feel free to push a fixup or amend if you'd prefer a different shape — happy to drop my commit if you want to roll your own.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -501,6 +501,29 @@ export function DesignFilesPanel({
...
+  function renderKindSections() {
+    const grouped = new Map<ProjectFileKind, ProjectFile[]>();
+    for (const file of pageFiles) {
+      const next = grouped.get(file.kind) ?? [];
+      next.push(file);
+      grouped.set(file.kind, next);
+    }
+
+    return [...grouped.entries()]
+      .sort(([a], [b]) => kindSortPriority(a) - kindSortPriority(b))
```
mrcfps, line 513: Clicking the Kind column still advertises sortable ascending/descending state (`aria-sort` and the header arrow), but this new `sort(...)` call always forces the section headers into `kindSortPriority` ascending order. In kind-grouped mode that means toggling Kind to descending leaves the sections in the same HTML → image → … order, so the control state and rendered order disagree. The regression comes from these changed lines because `pageFiles` already reflects the active sort before the groups are reordered here. Please apply `sortDir` when ordering the grouped entries (or preserve the insertion order from `pageFiles` when sorting by kind) and add a regression test that clicks the Kind header twice and asserts the section order flips.

mrcfps (commented):
@prantikmedhi Thanks for pushing the follow-up fixes. I re-read the current `DesignFilesPanel`/test diff and verified the kind-grouping path now renders grouped sections and the CI-unblocking test updates line up with that behavior. I could not rerun the web package checks in this worktree because `node_modules` are missing here, but I found one merge-safe behavior issue worth tightening before the next pass.


lefarcen (approved):

PR #1554, opened 2026-05-13
Fix Windows resource cache for Orbit templates

PerishCode:
## Summary
- include `design-templates` in the Windows `win.resource-tree` cache key
- bump the Windows resource-tree cache schema marker so old entries are not reused
- add a regression test proving `design-templates/orbit-general` changes invalidate and rematerialize the Windows resource tree

Fixes #1507

## Why
The packaged resource copy already includes `design-templates`, but the Windows resource-tree cache key did not hash that tree. A stale `win.resource-tree` cache could therefore keep reusing a pre-migration resource tree that lacked Orbit templates, and `win.electron-builder-dir` would inherit that stale key downstream.

## Validation
- red spec before fix: `pnpm -C tools/pack exec vitest run tests/win-resources.test.ts --reporter verbose` failed with stale `version one` content after updating `design-templates/orbit-general/SKILL.md`
- `pnpm -C tools/pack exec vitest run tests/win-resources.test.ts tests/resources.test.ts --reporter verbose`
- `pnpm --filter @open-design/tools-pack test`
- `git diff --check`
- `pnpm guard`
- `pnpm typecheck`

## Packaged artifact proof
- Built a real Windows dir artifact: `pnpm tools-pack win build --to dir --namespace o1507 --dir C:\tmp\od-1507 --json`
- Confirmed `win-unpacked/resources/open-design/design-templates` contains `orbit-general`, `orbit-github`, `orbit-gmail`, `orbit-linear`, and `orbit-notion`
- Confirmed `orbit-general/SKILL.md` exists in the artifact
- Started the packaged dir artifact and verified `document.title` is `Open Design`
- Verified runtime `/api/design-templates` exposes all five Orbit template ids above
- Stopped the packaged process and removed `C:\tmp\od-1507`

## Notes
- PR #1521 separately adds resource-copy coverage for `orbit-general`; this PR fixes the Windows cache invalidation boundary that can keep packaged artifacts stale.
- `pnpm typecheck` completed successfully with existing Astro content warnings and Shiki `urdu` fallback warnings.

### DIFF
--- a/tools/pack/src/win/resources.ts
+++ b/tools/pack/src/win/resources.ts
@@ -6,15 +6,18 @@ import type { ToolPackConfig } from "../config.js";
 import { copyBundledResourceTrees, winResources } from "../resources.js";
 import type { WinPaths, ResourceTreeCacheMetadata } from "./types.js";
 
+const RESOURCE_TREE_CACHE_SCHEMA_VERSION = 2;
+
 async function createResourceTreeCacheKey(config: ToolPackConfig): Promise<string> {
   return hashJson({
     assetsCommunityPets: await hashPath(join(config.workspaceRoot, "assets", "community-pets")),
     assetsFrames: await hashPath(join(config.workspaceRoot, "assets", "frames")),
     craft: await hashPath(join(config.workspaceRoot, "craft")),
     designSystems: await hashPath(join(config.workspaceRoot, "design-systems")),
+    designTemplates: await hashPath(join(config.workspaceRoot, "design-templates")),
     node: "win.resource-tree",
     promptTemplates: await hashPath(join(config.workspaceRoot, "prompt-templates")),
-    schemaVersion: 1,
+    schemaVersion: RESOURCE_TREE_CACHE_SCHEMA_VERSION,
     skills: await hashPath(join(config.workspaceRoot, "skills")),
   });
 }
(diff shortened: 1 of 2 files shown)


lefarcen (commented):
Hey @PerishCode, this is a focused fix for the Windows packaging cache boundary. The source change lines up with #1507: `win.resource-tree` now hashes `design-templates`, and the schema marker bump makes sure old resource-tree entries cannot be reused. The regression test also hits the important stale-cache behavior by changing `design-templates/orbit-general/SKILL.md` and expecting a second cache miss plus rematerialized content.

I don’t see code blockers in this pass. One PR-body housekeeping item before the next review pass: the new template is still missing **What users will see** and **Surface area**. Since this is packaging/runtime behavior rather than UI, Surface area can likely mark the packaged/default-behavior surface or explicitly say there is no direct UI checkbox to select, but reviewers should not have to infer that.

If you want to patch the description with an assistant, this prompt should be enough:

> Read the current PR body for nexu-io/open-design#1554 and the template at
> - **Why**: use case + pain being addressed, not a one-line restatement
> - **Surface area**: check every applicable box; pick "None" if this is not user-facing
> - **Screenshots** if UI is checked: entry-point screenshot, before/after
> Then update via `gh pr edit 1554 --repo nexu-io/open-design --body-file -`.

No rush — the implementation itself looks coherent; this just helps the next reviewer scope the user-visible packaging impact faster.

mrcfps (approved):
@PerishCode I reviewed the Windows resource-tree cache change and the new regression coverage. The cache key now tracks `design-templates`, the schema bump prevents reuse of pre-fix entries, and the added test exercises the stale-cache path by mutating `design-templates/orbit-general/SKILL.md` and verifying the rematerialized resource tree updates as expected.

Thanks for tightening up this packaging edge case 🙌

PR #1428, opened 2026-05-12
feat(analytics): PostHog product analytics (P0 events, consent-gated, packaged)

lefarcen:
## Summary

End-to-end PostHog product-analytics integration. Lands the **17 P0 events** from `埋点文档 1.0` (14 implementable + 3 deferred), gates everything on the existing Langfuse-style consent toggle, and wires PostHog's automatic diagnostics (autocapture, exceptions, web vitals, dead clicks, pageview) through a single privacy-scrubbing audit point.

Validated end-to-end against the live `nexu-io` PostHog project — all 18 daemon-side captures arrive with the correct schema and dedup keys. See the [Verification](#verification) section.

## P0 events shipped (14 of 20)

| # | Event | Where |
|---|---|---|
| 1 | `app_launch` | `App.tsx` first effect tick |
| 2-3 | `home_view` (page + asset_panel) | `EntryView.tsx` |
| 4 | `home_click` create_button | `NewProjectPanel.tsx` |
| 5 | `project_create_result` | `App.tsx::handleCreateProject` |
| 6 | `settings_view` | `SettingsDialog.tsx` (16 sections mapped) |
| 7-9 | `settings_click` (execution_mode_tab / cli_provider_card / byok_field focus) | `SettingsDialog.tsx` |
| 10-11 | `settings_cli_test_result` / `settings_byok_test_result` | `handleTestAgent` / `handleTestProvider` |
| 12 | `studio_view` chat_panel | `ChatComposer.tsx` |
| 13 | `studio_click` chat_composer | `ChatComposer.tsx` send + attach |
| 14-15 | `run_created` / `run_finished` | daemon `chat-routes.ts` POST `/api/runs` + `runs.wait()` |
| 16 | `studio_view` artifact | `FileViewer.tsx` dispatcher |
| 17 | `studio_click` share_option | `FileViewer.tsx` share menu (6 formats) |
| 18 | `artifact_export_result` | `FileViewer.tsx::fireShareExport` |

**Deferred P0** (depend on product surfaces this branch doesn't have): `project_open_result` (no project-load result step), `settings_click byok_provider_option` (CSV enum doesn't map to any UI picker). See `specs/change/20260511-posthog-analytics/tracking-doc-issues.md` for the full 15 BLOCKING/IMPORTANT items the CSV needs to resolve.

## Architecture

```
┌─ Privacy → "Share usage data" toggle (existing) ────────────┐
│   ↓ posthog.opt_out_capturing / opt_in_capturing            │
[...]

PR #1558, opened 2026-05-13
feat(analytics): PostHog + Langfuse instrumentation for assistant feedback

lefarcen:
## Summary

Wires the four P0 events from the `assistant_feedback_*` product-tracking spec into PR #1379's thumbs-up/down UI, and forwards the rating to Langfuse as a `score-create` so evals can filter traces by user feedback.

## Events

PostHog (web → daemon `/api/analytics/config` → posthog-js):

| Event | When | Element / action |
|---|---|---|
| `assistant_feedback_click` | Thumb pressed or cleared | `assistant_feedback_button` / `submit_feedback_rating` \| `clear_feedback_rating` |
| `assistant_feedback_reason_view` | Reason panel rendered | `assistant_feedback_reason_panel` |
| `assistant_feedback_reason_click` | Submit button clicked | `assistant_feedback_reason_submit_button` / `click_submit_feedback_reason` |
| `assistant_feedback_reason_submit` | Reason payload result | `assistant_feedback_reason_submit` / `submit_feedback_reason` |

`reason_click` and `reason_submit` share a `request_id` so PostHog can stitch the click/result pair per the spec.

## Click event state machine (`rating_before` / `rating`)

| Operation | `rating_before` | `rating` | `action` |
|---|---|---|---|
| First 👍 | `none` | `positive` | `submit_feedback_rating` |
| First 👎 | `none` | `negative` | `submit_feedback_rating` |
| 👍 → 👎 | `positive` | `negative` | `submit_feedback_rating` |
| 👍 → cleared | `positive` | `none` | `clear_feedback_rating` |
| 👎 → cleared | `negative` | `none` | `clear_feedback_rating` |

`rating` carries the post-action state. The widened union (`positive | negative | none`) is scoped to click events only; reason_* events still use the strict `positive | negative` since they only fire after a thumb is committed.

## Langfuse score-create (web → daemon → Langfuse)

- `user_rating` (NUMERIC, +1 / -1) — appears in Langfuse's score filters out of the box.
- `user_rating_reason` (CATEGORICAL, one per reason code) — same trace can be filtered by individual reasons.

`POST /api/runs/:id/feedback` is the single network egress; the daemon re-reads \`app-config.json\` and gates on \`telemetry.metrics + telemetry.content\` consent independently of what the browser thinks. Web fires-and-forgets through \`reportChatRunFeedback\`.

## Field threading

`ChatPane` gains `projectKind: TrackingProjectKind`, passes `projectId / projectKind / conversationId` down to `AssistantMessage` so each event carries the funnel keys the product spec requires.

## Product decisions captured in commits

- **2026-05-13** Renamed `previous_rating` → `rating_before` for symmetry with `rating`; clear actions now report `rating='none'` instead of echoing the cleared thumb (the pre/post pair fully describes the transition).
- **2026-05-13** Product confirmed `custom_reason` ships the raw free-text — the earlier length-bucketing approach lost too much signal for analysts. Consent-gated behind `telemetry.content` like other message-content telemetry.

The product-tracking doc currently lists `rating(enum: positive|negative)` and bucketed `custom_reason`; both diverge from the as-shipped wire format. Follow-up with product to align the doc.

## End-to-end verification

Verified locally against the live PostHog project (US Cloud, project 420348):

```
=== assistant_feedback_click ===
  rating: 'positive' | rating_before: 'none' | action: 'submit_feedback_rating'
  page / area / element / project_id / project_kind=prototype / conversation_id /
  assistant_message_id / run_id / has_produced_files: True
=== assistant_feedback_reason_click ↔ reason_submit ===
  reason: ['useful_structure', 'easy_to_continue', 'other']
  reason_count: 3 | has_custom_reason: True | custom_reason: 'test11111'  ← raw text
  request_id stitching: BOTH = '067e4990-e69d-4fcf-aa84-6aff48e6c057'  ← matches spec
```

Langfuse `user_rating` + `user_rating_reason` scores observed on the corresponding trace; full Langfuse end-to-end pending only after merge + CI key wiring.

## Test plan

- [x] `pnpm typecheck` — 0 errors
- [x] `pnpm --filter @open-design/web test` — relevant suites pass (chat-feedback / ChatPane.streaming / FileViewer / NewProjectPanel)
- [x] `pnpm --filter @open-design/daemon exec vitest run tests/langfuse-trace.test.ts tests/langfuse-bridge.test.ts` — 55 / 55 (includes 5 new tests for `buildFeedbackPayload` + `reportRunFeedback`)
- [x] Live PostHog event verification on US Cloud project 420348
- [x] Live Langfuse score verification on US Cloud project (one trace, scores visible in UI)

### DIFF
--- a/apps/daemon/src/chat-routes.ts
+++ b/apps/daemon/src/chat-routes.ts
@@ -12,7 +12,7 @@ import { isSafeId as isSafeProjectId } from './projects.js';
 import { projectKindToTracking } from '@open-design/contracts/analytics';
 import { validateBaseUrlResolved } from './connectionTest.js';
 
-export interface RegisterChatRoutesDeps extends RouteDeps<'db' | 'design' | 'http' | 'chat' | 'agents' | 'critique' | 'validation' | 'lifecycle' | 'paths'> {}
+export interface RegisterChatRoutesDeps extends RouteDeps<'db' | 'design' | 'http' | 'chat' | 'agents' | 'critique' | 'validation' | 'lifecycle' | 'paths' | 'telemetry'> {}
 
 export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
   const { db, design } = ctx;
@@ -120,6 +120,60 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
       return sendApiError(res, 500, 'INTERNAL', `tool result write failed: ${reason}`);
     }
     res.json({ ok: true });
+  });
+
+  // Receives the user's thumbs-up/down (+ reason codes) for an assistant
+  // turn and forwards it to Langfuse as a `score-create`. Web persists the
+  // feedback itself via PUT /messages/:id; this endpoint exists only as a
+  // telemetry side channel — the daemon is the single network egress for
+  // Langfuse and gates on `telemetry.metrics + telemetry.content` consent.
+  // Fire-and-forget: respond 202 once enqueued so the UI isn't blocked.
+  app.post('/api/runs/:id/feedback', (req, res) => {
+    const runId = req.params.id;
+    const body = (req.body ?? {}) as Partial<{
+      projectId: string;
+      conversationId: string;
+      assistantMessageId: string;
+      rating: 'positive' | 'negative';
+      reasonCodes: string[];
+      hasCustomReason: boolean;
+      customReason: string;
+    }>;
+    if (!runId) {
+      return sendApiError(res, 400, 'INVALID_RUN_ID', 'runId missing');
+    }
+    if (body.rating !== 'positive' && body.rating !== 'negative') {
+      return sendApiError(res, 400, 'INVALID_RATING', 'rating must be positive or negative');
+    }
+    const reasonCodes = Array.isArray(body.reasonCodes)
+      ? body.reasonCodes.filter((c): c is string => typeof c === 'string')
+      : [];
+    const customReason = typeof body.customReason === 'string' ? body.customReason : '';
+    const reportFeedback = ctx.telemetry?.reportFeedback;
+    if (!reportFeedback) {
+      res.status(202).json({ status: 'skipped_no_sink' });
+      return;
+    }
+    // Build score metadata bag that lands in the Langfuse score body.
+    // Mirrors the PostHog event so analysts can cross-reference.
+    const scoreMetadata: Record<string, unknown> = {
+      projectId: body.projectId,
+      conversationId: body.conversationId,
+      assistantMessageId: body.assistantMessageId,
+      hasCustomReason: body.hasCustomReason === true,
+      customReason,
+    };
+    // Don't await — the request returns immediately. Errors are logged
+    // inside langfuse-bridge.
+    void reportFeedback({
+      runId,
+      rating: body.rating,
+      reasonCodes,
+      hasCustomReason: body.hasCustomReason === true,
+      customReason,
+      scoreMetadata,
+    });
+    res.status(202).json({ status: 'accepted' });
   });
 
   app.post('/api/chat', (req, res) => {
--- a/apps/daemon/src/chat-routes.ts
+++ b/apps/daemon/src/chat-routes.ts
@@ -12,6 +12,24 @@ import { isSafeId as isSafeProjectId } from './projects.js';
 import { projectKindToTracking } from '@open-design/contracts/analytics';
 import { validateBaseUrlResolved } from './connectionTest.js';
 
+// Allowlist for the `/feedback` route. Mirrors the
+// ChatMessageFeedbackReasonCode union in packages/contracts/src/api/chat.ts.
+// Kept inline (not imported as a runtime value, since the contract type is
+// type-only) so a stale client can't poison Langfuse with unknown categories.
+const FEEDBACK_REASON_ALLOWLIST: ReadonlySet<string> = new Set([
+  'matched_request',
+  'strong_visual',
+  'useful_structure',
+  'easy_to_continue',
+  'followed_design_system',
+  'missed_request',
+  'weak_visual',
+  'incomplete_output',
+  'hard_to_use',
+  'missed_design_system',
+  'other',
+]);
+
 export interface RegisterChatRoutesDeps extends RouteDeps<'db' | 'design' | 'http' | 'chat' | 'agents' | 'critique' | 'validation' | 'lifecycle' | 'paths' | 'telemetry'> {}
 
 export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
@@ -127,8 +145,13 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
   // feedback itself via PUT /messages/:id; this endpoint exists only as a
   // telemetry side channel — the daemon is the single network egress for
   // Langfuse and gates on `telemetry.metrics + telemetry.content` consent.
-  // Fire-and-forget: respond 202 once enqueued so the UI isn't blocked.
-  app.post('/api/runs/:id/feedback', (req, res) => {
+  //
+  // The consent + sink decision is fast (awaits a small file read, no
+  // network); we await it so the response status honestly reflects whether
+  // the score was enqueued, skipped for consent, or skipped because no
+  // Langfuse sink is configured. The actual Langfuse network call happens
+  // as a detached promise inside the bridge.
+  app.post('/api/runs/:id/feedback', async (req, res) => {
     const runId = req.params.id;
     const body = (req.body ?? {}) as Partial<{
       projectId: string;
@@ -145,8 +168,19 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
     if (body.rating !== 'positive' && body.rating !== 'negative') {
       return sendApiError(res, 400, 'INVALID_RATING', 'rating must be positive or negative');
     }
+    // Drop anything outside the contract-side reason allowlist and
+    // deduplicate; otherwise a malformed or replayed client payload could
+    // create unknown Langfuse categories or duplicate score ids in the
+    // same batch.
     const reasonCodes = Array.isArray(body.reasonCodes)
-      ? body.reasonCodes.filter((c): c is string => typeof c === 'string')
+      ? Array.from(
+          new Set(
+            body.reasonCodes.filter(
+              (c): c is string =>
+                typeof c === 'string' && FEEDBACK_REASON_ALLOWLIST.has(c),
+            ),
+          ),
+        )
       : [];
     const customReason = typeof body.customReason === 'string' ? body.customReason : '';
     const reportFeedback = ctx.telemetry?.reportFeedback;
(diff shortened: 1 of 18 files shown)


lefarcen (commented):
Hey @lefarcen — the analytics direction is clear here: the custom-reason length bucketing and daemon-side consent gate are the right privacy shape for feedback instrumentation.

Because this is still marked draft and the diff currently includes the release/v0.7.0 carry-over, I’m leaving this as a draft-state review rather than inline blockers. I did a first pass over the high-signal feedback instrumentation paths and the related lineage (#1428, #1509, #1379); no public blockers from this initial pass.

Once the branch is rebased down to the feedback-only diff and marked ready for review, we should do the full current-head review before merge.

alchemistklk (commented):
English: Findings from review.
Chinese: Review 发现如下。

English: P1: `AssistantMessage` now requires `projectKind` and `conversationId`, but an existing direct test render was not updated, so the web typecheck/test surface will fail once dependencies are installed.
Chinese: P1：`AssistantMessage` 现在要求传入 `projectKind` 和 `conversationId`，但已有的直接测试渲染没有更新，所以安装依赖后 web 的 typecheck/test 会失败。

English: Code evidence: `apps/web/src/components/AssistantMessage.tsx:52` declares `projectKind: TrackingProjectKind`, and `apps/web/src/components/AssistantMessage.tsx:53` declares `conversationId: string | null` as required props.
Chinese: 代码证据：`apps/web/src/components/AssistantMessage.tsx:52` 声明了必填的 `projectKind: TrackingProjectKind`，并且 `apps/web/src/components/AssistantMessage.tsx:53` 声明了必填的 `conversationId: string | null`。

English: Code evidence: `apps/web/tests/components/AssistantMessage.test.tsx:52` renders `<AssistantMessage>` with only `message`, `streaming`, `projectId`, and `onFeedback`, so the new required props are missing.
Chinese: 代码证据：`apps/web/tests/components/AssistantMessage.test.tsx:52` 渲染 `<AssistantMessage>` 时只传入了 `message`、`streaming`、`projectId` 和 `onFeedback`，因此缺少新的必填 props。

English: P1: Langfuse feedback can become stale when a user clears feedback or switches thumbs after previously submitting reasons.
Chinese: P1：当用户提交 reason 后又清除反馈或切换点赞/点踩时，Langfuse 中的反馈可能变成过期数据。

English: Code evidence: `apps/web/src/components/ProjectView.tsx:1010` only calls `reportChatRunFeedback` when `change` is non-null, so clear events are skipped entirely.
Chinese: 代码证据：`apps/web/src/components/ProjectView.tsx:1010` 只有在 `change` 非 null 时才调用 `reportChatRunFeedback`，所以 clear 事件会被完全跳过。

English: Code evidence: `apps/web/src/components/ProjectView.tsx:1017` sends `reasonCodes: change.reasonCodes ?? []`, so switching thumbs without submitting new reasons reports an empty current reason list.
Chinese: 代码证据：`apps/web/src/components/ProjectView.tsx:1017` 发送 `reasonCodes: change.reasonCodes ?? []`，所以切换点赞/点踩但未提交新 reason 时会报告空的当前 reason 列表。

English: Code evidence: `apps/daemon/src/langfuse-trace.ts:724` creates score entries only for the submitted reason list, and `apps/daemon/src/langfuse-trace.ts:731` uses stable IDs like `${traceId}-reason-${code}` without any deletion or tombstone path for previously submitted reason scores.
Chinese: 代码证据：`apps/daemon/src/langfuse-trace.ts:724` 只为提交的 reason 列表创建 score 条目，并且 `apps/daemon/src/langfuse-trace.ts:731` 使用 `${traceId}-reason-${code}` 这样的稳定 ID，但没有为之前提交过的 reason score 提供删除或 tombstone 路径。

apps/web/src/components/AssistantMessage.tsx
```diff
@@ -36,6 +49,8 @@ interface Props {
   message: ChatMessage;
   streaming: boolean;
   projectId: string | null;
+  projectKind: TrackingProjectKind;
```
lefarcen: P1 Making `projectKind` and `conversationId` required is fine for production, but the existing direct `AssistantMessage` tests were not updated with these props. Please update those test renders too; otherwise the web typecheck/test surface fails even though the `ChatPane` caller was updated.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1000,8 +1002,25 @@ export function ProjectView({
               },
         true,
       );
+      // Forward affirmative ratings to the daemon → Langfuse `score-create`.
+      // Clears (change=null) are skipped — Langfuse scores are append-only,
+      // and the rating is also captured by the PostHog event so a clear is
+      // recoverable downstream if we ever need it.
+      const runId = assistantMessage.runId;
+      if (change && runId && activeConversationId) {
```
lefarcen, line 1485: P1 This skips Langfuse reporting when the user clears feedback, so the stable `user_rating` score for the trace can remain positive/negative after the UI has removed the rating. Please add an explicit clear/tombstone path (or another deterministic overwrite) instead of silently skipping `change === null`.

apps/web/src/components/ProjectView.tsx
```diff
@@ -1000,8 +1002,25 @@ export function ProjectView({
...
+        void reportChatRunFeedback({
+          runId,
+          projectId: project.id,
+          conversationId: activeConversationId,
+          assistantMessageId: assistantMessage.id,
+          rating: change.rating,
+          reasonCodes: change.reasonCodes ?? [],
```
lefarcen, line 1492: P1 Sending an empty `reasonCodes` list when the user switches thumbs without submitting new reasons does not remove previously-created Langfuse `user_rating_reason` scores; `buildFeedbackPayload` only creates scores for the current list and has no deletion path for old stable IDs. Please make reason updates replace the previous reason set end-to-end, or avoid emitting stale per-reason scores.

lefarcen (commented):
Hey @lefarcen — the analytics threading is headed in the right direction, especially keeping the browser on the PostHog path and routing Langfuse through the daemon-side consent gate. I agree with the current blocker shape from @alchemistklk, and found the same two must-fix issues on the current head: the direct AssistantMessage tests still miss the new required props, and Langfuse can retain stale feedback after clears / reason changes.

Also, before the next review pass, please fill the PR template sections that are currently missing: Why, What users will see, Surface area, and Validation. That will make the product-review pass much easier to scope.

Once these are tightened up, this should be much easier to review confidently.

lefarcen (commented):
Hey! 👋 Since this is still a draft, I'll hold off on a full review until you mark it ready. But I did notice one thing in the current commit (`6326206c`) that's worth flagging early:

**Privacy claim mismatch** — the PR body states that `custom_reason` is "bucketed into a coarse length band (0 / 1_20 / 21_100 / 101_500 / 501_plus); the raw free-text never leaves the client." However, `langfuse-trace.ts:703-705` shows:

```typescript
customReason: ctx.customReason || undefined,
```

The raw text goes directly to Langfuse (gated by `telemetry.content` consent at line 751, which makes it opt-in). The type comment at `packages/contracts/src/analytics/events.ts:674-676` even notes "Product confirmed on 2026-05-13 that the raw text ships (no length bucketing)."

Two paths forward:
1. Implement the bucketing before Langfuse score-create (if that's still the intended design), or
2. Update the PR body to accurately state "raw text goes to Langfuse when telemetry.content=true"

The consent gate makes this technically correct from a privacy perspective, but the PR body's current claim is misleading. Happy to do a full review once you're out of draft!

apps/web/src/components/AssistantMessage.tsx
```diff
@@ -413,13 +450,67 @@ function AssistantFeedback({
     if (!reasonRating) return;
     reasonsRef.current?.scrollIntoView({ block: "start", behavior: "smooth" });
   }, [reasonRating]);
+  // Fire assistant_feedback_reason_view exactly when the reason panel
+  // becomes visible — once per (message, rating) pair. Re-opening the
+  // panel after a clear+reselect counts as a new view.
+  useEffect(() => {
```
lefarcen: This effect is described as firing `assistant_feedback_reason_view` exactly once when the panel becomes visible, but it is also keyed on `analytics.track` and every `analyticsCtx` field. If the analytics provider resolves or updates while the panel is still open (for example when identity/locale/app-version state changes), React will rerun the effect and emit another `reason_view` for the same `(assistant_message_id, rating)` pair. That inflates the new view funnel this PR is adding and breaks the one-open → one-view invariant the comment documents. Please guard the capture with a ref keyed by message+rating, or move it into the closed→open transition path, and add a regression test that rerenders the component while the panel stays open.

apps/daemon/src/chat-routes.ts
```diff
@@ -219,6 +219,60 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
...
+    // inside langfuse-bridge.
+    void reportFeedback({
+      runId,
+      rating: body.rating,
+      reasonCodes,
+      hasCustomReason: body.hasCustomReason === true,
+      customReason,
+      scoreMetadata,
+    });
+    res.status(202).json({ status: 'accepted' });
```
lefarcen: `ChatRunFeedbackResponse` and `reportRunFeedbackFromDaemon()` both advertise `skipped_consent` / `skipped_no_sink`, but this handler always responds with `{ status: "accepted" }` as soon as `ctx.telemetry.reportFeedback` exists because the promise is fire-and-forgotten. In production `server.ts` always wires `reportFeedback`, so the consent-off branch in `langfuse-bridge.ts` is never observable through the HTTP API. That leaves QA or future callers with a false success signal and a contract the implementation cannot actually satisfy. Please either return the real preflight outcome here (even if the network send remains async), or narrow the response type/docs/comments to `accepted` only if the endpoint is intentionally blind to the final telemetry result.

apps/daemon/src/chat-routes.ts
```diff
@@ -219,6 +219,60 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
...
+      hasCustomReason: boolean;
+      customReason: string;
+    }>;
+    if (!runId) {
+      return sendApiError(res, 400, 'INVALID_RUN_ID', 'runId missing');
+    }
+    if (body.rating !== 'positive' && body.rating !== 'negative') {
+      return sendApiError(res, 400, 'INVALID_RATING', 'rating must be positive or negative');
+    }
+    const reasonCodes = Array.isArray(body.reasonCodes)
```
lefarcen, line 175: `reasonCodes` are only filtered to `string[]` here, then `buildFeedbackPayload()` uses each value verbatim as both the categorical score value and part of the stable id (`${traceId}-reason-${code}`). A malformed or duplicated client payload therefore creates unknown Langfuse categories and can emit duplicate score ids in the same batch. Because this is daemon input, the TypeScript contract on the web side is not enough runtime protection. Please validate against the `ChatMessageFeedbackReasonCode` allowlist and dedupe before forwarding to Langfuse (for example `Array.from(new Set(validCodes))`), then add a daemon test that covers invalid and duplicate codes.

lefarcen (commented):
I found a few follow-up issues in the new feedback analytics path. They do not look merge-blocking, but they are worth tightening before we rely on these signals.


PerishCode (approved):

PR #1552, opened 2026-05-13
fix: Implement file grouping by type in file list

xxiaoxiong:
Fixes #1548

## Problem

Selecting **按类型** (by type) in the file list grouping control did not actually group files by type. The control reflected that type grouping was selected, but the file list remained as a **flat list** without visual separation by file kind.

### Current Behavior

- ❌ Grouping control shows "类型" (Type) as selected
- ❌ File list remains flat (no grouping)
- ❌ Files are not visually separated by type
- ❌ Users cannot organize files efficiently

### Expected Behavior

- ✅ Files should be grouped into clear type-based sections
- ✅ Each group should show file type label and count
- ✅ Groups should be collapsible/expandable
- ✅ Grouping behavior should match the selected mode

---

## Root Cause

The grouping logic only implemented **'modified'** mode with `renderModifiedSections()`. When `groupMode === 'kind'`, the code simply rendered `pageFiles.map(renderFileRow)` without any grouping, resulting in a flat list.

### Code Before

```typescript
<tbody>
  {groupMode === 'modified'
    ? renderModifiedSections()
    : pageFiles.map(renderFileRow)}  // ❌ No grouping for 'kind' mode
</tbody>
```

---

## Solution

Implemented complete **file-type grouping functionality** following the same pattern as `renderModifiedSections()`:

### 1. Added State Management

```typescript
const [collapsedKindSections, setCollapsedKindSections] = useState<
  Set<ProjectFileKind>
>(new Set());
```

### 2. Added Kind Groups

```typescript
const kindGroups = useMemo(() => {
  const groups: Partial<Record<ProjectFileKind, ProjectFile[]>> = {};
  for (const f of pageFiles) {
    if (!groups[f.kind]) groups[f.kind] = [];
    groups[f.kind]!.push(f);
  }
  return groups;
}, [pageFiles]);

const visibleKindSections = (Object.keys(kindGroups) as ProjectFileKind[]).sort(
  (a, b) => kindSortPriority(a) - kindSortPriority(b),
);
```

### 3. Added Toggle Function

```typescript
function toggleKindSection(kind: ProjectFileKind) {
  setCollapsedKindSections((prev) => {
    const next = new Set(prev);
    if (next.has(kind)) {
      next.delete(kind);
    } else {
      next.add(kind);
    }
    return next;
  });
}
```

### 4. Added Render Function

```typescript
function renderKindSections() {
  return visibleKindSections.flatMap((kind) => {
    const sectionFiles = kindGroups[kind] || [];
[...]

PR #1561, opened 2026-05-13
fix(design-files): implement file type grouping in DesignFilesPanel

mturac:
The 'by type' grouping option was not implemented - selecting it showed the active state but rendered files as a flat list.

Added renderKindSections() to group files by their kind (html, image, video, etc.) similar to the existing renderModifiedSections() pattern.

Fixes #1548


apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -501,6 +514,59 @@ export function DesignFilesPanel({
...
+      if (!groups[kind]) groups[kind] = [];
+      groups[kind].push(file);
+    }
+    return groups;
+  }, [pageFiles]);
+
+  const KIND_LABELS: Record<string, string> = {
+    html: t('designFiles.kindHtml'),
+    image: t('designFiles.kindImage'),
+    video: t('designFiles.kindVideo'),
```
lefarcen, line 530: P1 This adds `t('designFiles.kindVideo')` and `t('designFiles.kindAudio')`, but those keys are not present in `Dict` or the locale dictionaries, so the web typecheck will fail. Please add both keys to `apps/web/src/i18n/types.ts` and every locale file, or reuse the existing `kindLabel()` fallback until those labels exist.

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -501,6 +514,59 @@ export function DesignFilesPanel({
...
+    code: t('designFiles.kindCode'),
+    pdf: t('designFiles.kindPdf'),
+    document: t('designFiles.kindDocument'),
+    presentation: t('designFiles.kindPresentation'),
+    spreadsheet: t('designFiles.kindSpreadsheet'),
+    binary: t('designFiles.kindBinary'),
+    other: t('designFiles.kindOther'),
+  };
+
+  function renderKindSections() {
```
lefarcen, line 543: P2 Please add a regression test for Kind grouping in `apps/web/tests/components/DesignFilesPanel.test.tsx`. This PR fixes a user-visible bug from #1548, and the repo review guide asks bug fixes to carry a reproduction/regression signal; a small test that renders mixed kinds, verifies the section headers/rows, and collapses one kind would cover the behavior directly.

apps/web/src/components/MemorySection.tsx
```diff
@@ -916,11 +922,12 @@ export function MemorySection() {
...
             className="ghost"
             onClick={() => void reloadExtractions()}
+            disabled={isRefreshing}
             title={t('settings.memoryExtractionsRefresh')}
           >
-            <Icon name="refresh" size={12} />{' '}
+            <Icon name="refresh" size={12} className={isRefreshing ? 'icon-spin' : ''} />{' '}
             <span style={{ marginLeft: 4 }}>
-              {t('settings.memoryExtractionsRefresh')}
+              {isRefreshing ? t('settings.memoryExtractionsRefreshing') : t('settings.memoryExtractionsRefresh')}
```
lefarcen, line 930: P2 This refresh-state UI change is unrelated to Design Files type grouping and forces a new settings i18n key across every locale. Please split it into a separate PR or remove it here so this bugfix stays scoped to #1548.

lefarcen (changes requested):
Hey @mturac, thanks for jumping on the Design Files grouping bug. The overall direction is right — reusing the same section-row pattern as modified-date grouping fits this panel well.

I found a couple of issues that need fixing before this can land: one compile blocker around the new video/audio labels, and one scope issue from an unrelated Memory settings change. Also, the PR body is missing the template sections **Why**, **What users will see**, **Surface area**, and **Validation**; filling those in will make the next pass quicker.

If you’re using an AI assistant, this prompt should be enough to update the description cleanly:

> Read the current PR body for nexu-io/open-design#1561 and the template at
> Then update via `gh pr edit 1561 --repo nexu-io/open-design --body-file -`.

Once those are addressed, this should be a much tighter fix for #1548. 🙌

apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -501,6 +514,59 @@ export function DesignFilesPanel({
...
+    audio: t('designFiles.kindAudio'),
+    sketch: t('designFiles.kindSketch'),
+    text: t('designFiles.kindText'),
+    code: t('designFiles.kindCode'),
+    pdf: t('designFiles.kindPdf'),
+    document: t('designFiles.kindDocument'),
+    presentation: t('designFiles.kindPresentation'),
+    spreadsheet: t('designFiles.kindSpreadsheet'),
+    binary: t('designFiles.kindBinary'),
+    other: t('designFiles.kindOther'),
```
PerishCode, line 540: **Blocking — `designFiles.kindVideo` / `kindAudio` / `kindOther` are not declared in `Dict`, so the web `tsc -b --noEmit` step fails and `nix flake check`'s `open-design-web` build aborts.**

`t()` is typed as `(key: keyof Dict, vars?) => string` in `apps/web/src/i18n/index.tsx:79`, and `Dict` (`apps/web/src/i18n/types.ts:1038-1048`) only declares `kindHtml`, `kindImage`, `kindSketch`, `kindText`, `kindCode`, `kindPdf`, `kindDocument`, `kindPresentation`, `kindSpreadsheet`, `kindLiveArtifact`, `kindBinary`. The three new keys you reference here — `designFiles.kindVideo` (RIGHT line 530), `designFiles.kindAudio` (RIGHT line 531), `designFiles.kindOther` (RIGHT line 540) — are neither in `Dict` nor in any of the 18 locale files.

`pnpm --filter @open-design/web typecheck` and the PR's failing `nix flake check` run both report:

```
src/components/DesignFilesPanel.tsx(530,14): error TS2345: Argument of type '"designFiles.kindVideo"' is not assignable to parameter of type 'keyof Dict'.
src/components/DesignFilesPanel.tsx(531,14): error TS2345: Argument of type '"designFiles.kindAudio"' is not assignable to parameter of type 'keyof Dict'.
src/components/DesignFilesPanel.tsx(540,14): error TS2345: Argument of type '"designFiles.kindOther"' is not assignable to parameter of type 'keyof Dict'.
```

The reason these keys never existed before is that the existing `kindLabel()` helper at `apps/web/src/components/DesignFilesPanel.tsx:1165-1176` folds `video`, `audio`, and any unknown kind into `kindBinary` rather than giving them their own labels.

**Suggested change — add the three missing keys.** This is the better-UX option because grouping is precisely the place where users want to see `Video` / `Audio` as their own buckets.

In `apps/web/src/i18n/types.ts` next to the existing `kindXxx` entries:

```ts
'designFiles.kindVideo': string;
'designFiles.kindAudio': string;
'designFiles.kindOther': string;
```

Then add the matching translations to every file under `apps/web/src/i18n/locales/` (e.g. `'Video'` / `'Audio'` / `'Other'` for `en`, localized for the rest — same shape you already added for `settings.memoryExtractionsRefreshing` in commits 1eaeb6e and 7c0c98c). Consider also extending `kindLabel()` (DesignFilesPanel.tsx:1165) to use the new `kindVideo` / `kindAudio` so the table's Kind column and the new grouping headers stay in sync.

If you would rather not expand the i18n surface in this PR, the strictly minimal alternative is to reuse `kindBinary` here, but that produces several sections all labelled "Binary" and undoes most of the value of the new grouping.


apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -501,6 +514,59 @@ export function DesignFilesPanel({
...
+              aria-label={`${collapsed ? t('designFiles.expandGroup') : t('designFiles.collapseGroup')} ${label}`}
+              onClick={() => toggleKind(kind)}
+            >
+              <Icon name={collapsed ? 'chevron-right' : 'chevron-down'} size={13} />
+              <span>{label}</span>
+              <span className="df-section-count">{kindFiles.length}</span>
+            </button>
+          </td>
+        </tr>,
+        ...(collapsed ? [] : kindFiles.map(renderFileRow)),
```
PerishCode, line 565: **Blocking — `kindFiles` is inferred as `ProjectFile[] | undefined`, so the web typecheck fails on both usages.**

`kindGroups` is typed `Record<string, ProjectFile[]>` (line 518), and under this repo's `noUncheckedIndexedAccess` setting `kindGroups[kind]` widens to `ProjectFile[] | undefined` even though `kind` came from `Object.keys(kindGroups)`. CI reports:

```
src/components/DesignFilesPanel.tsx(561,51): error TS18048: 'kindFiles' is possibly 'undefined'.
src/components/DesignFilesPanel.tsx(565,30): error TS18048: 'kindFiles' is possibly 'undefined'.
```

`renderModifiedSections` (line 491) avoids this because `modifiedGroups` is a closed `Record<ModifiedSection, ProjectFile[]>` where every key is required, so indexed access does not widen.

**Suggested change — iterate `Object.entries` so the array type is preserved without an extra runtime guard:**

```ts
function renderKindSections() {
  const entries = Object.entries(kindGroups).sort(([a], [b]) => a.localeCompare(b));
  return entries.flatMap(([kind, kindFiles]) => {
    const collapsed = collapsedKinds.has(kind);
    const label = KIND_LABELS[kind] || kind;
    return [
      <tr className="df-section-row" key={`${kind}-label`}>
        <td colSpan={6}>
          <button
            type="button"
            className="df-section-toggle"
            aria-expanded={!collapsed}
            aria-label={`${collapsed ? t('designFiles.expandGroup') : t('designFiles.collapseGroup')} ${label}`}
            onClick={() => toggleKind(kind)}
          >
            <Icon name={collapsed ? 'chevron-right' : 'chevron-down'} size={13} />
            <span>{label}</span>
            <span className="df-section-count">{kindFiles.length}</span>
          </button>
        </td>
      </tr>,
      ...(collapsed ? [] : kindFiles.map(renderFileRow)),
    ];
  });
}
```

`Object.entries(Record<string, T>)` returns `[string, T][]`, so `kindFiles` is `ProjectFile[]` and the existing `.length` / `.map` calls become valid.


PerishCode (commented):
@mturac thanks for picking this up — the "by type" grouping is a real gap and the structure you mirrored from `renderModifiedSections` is the right shape. Two blocking issues from the new `renderKindSections` / `KIND_LABELS` block are currently breaking the web typecheck (and therefore the `nix flake check` build job on this PR). Details inline.


lefarcen (changes requested):
Hey @mturac, quick follow-up after re-reading the current head. The main direction still looks right, but the PR is still blocked on the same compile/scope items from my earlier review. @PerishCode also caught one additional TypeScript blocker in the same `renderKindSections` block: `kindGroups[kind]` is still possibly `undefined` under `noUncheckedIndexedAccess`, so switching that loop to `Object.entries(kindGroups)` (or otherwise narrowing the value) will be needed too.

No need to answer this review separately — the fastest path is to fix the existing inline comments, add the Kind-grouping regression test, remove/split the unrelated Memory refresh change, and fill in the missing bugfix validation/body sections. Once you push an update, we can do a fresh pass on the new head. 🙌

lefarcen:
Heads-up: there are two other open PRs targeting the same file-grouping feature in `DesignFilesPanel.tsx`:

- PR #1551 by @prantikmedhi — narrower (just `DesignFilesPanel.tsx` + test, no i18n)
- PR #1552 by @xxiaoxiong — similar i18n + DesignFilesPanel scope to yours, but doesn't touch `MemorySection.tsx`

Three implementations of essentially the same feature — sharing so you all can coordinate with @prantikmedhi + @xxiaoxiong on which approach lands. Maintainer team will make the final call.

lefarcen:
Heads-up: two other open PRs also fix the same `DesignFilesPanel.tsx` kind-grouping regression for #1548 — PR #1551 by @prantikmedhi and PR #1632 by @xxiaoxiong (currently blocked on CI: missing i18n keys). Your locale changes in this PR already cover exactly that gap in #1632, worth a look for context. The maintainer team will decide which approach lands; sharing so none of the effort gets wasted.

lefarcen:
Hey @mturac — a quick update on where things landed for this feature area.

PR #1551 merged earlier today and closed #1548 — the file-type grouping fix is now in `main`. Your approach here covered the same core path and also added i18n keys across all 20 locales, which was a wider scope than what ended up landing.

Since #1548 is now resolved and #1551 touched the same files in `DesignFilesPanel.tsx`, this PR now has conflicts with main (showing as DIRTY). The maintainer team will decide whether to formally close it.

Thank you for jumping on this so quickly and for the breadth of locale coverage you included — that attention to i18n was a nice touch. 🙌

PR #1566, opened 2026-05-13
fix(web): auto-scroll feedback form

prantikmedhi:
Fixes #1563.

When a user selects feedback, the reasons panel now scrolls into view so the form is immediately visible.

Tested with:
- pnpm exec vitest run -c vitest.config.ts tests/components/chat-feedback.test.tsx

### DIFF
--- a/apps/web/src/components/AssistantMessage.tsx
+++ b/apps/web/src/components/AssistantMessage.tsx
@@ -1,4 +1,4 @@
-import { Fragment, type ReactNode, useEffect, useMemo, useState } from "react";
+import { Fragment, type ReactNode, useEffect, useMemo, useRef, useState } from "react";
 import { ToolCard } from "./ToolCard";
 import { renderMarkdown } from "../runtime/markdown";
 import { projectFileUrl } from "../providers/registry";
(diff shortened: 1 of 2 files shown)


lefarcen (commented):
Hey @prantikmedhi, thanks for picking this up — the implementation is nicely scoped to the feedback panel, and the added component test covers the new scroll behavior directly.

I do not see a code blocker in the diff. One housekeeping ask before the next reviewer pass: the PR body is missing a few sections from this repo's template, which helps us confirm the user-facing surface and validation quickly. Could you fill in **Why**, **What users will see**, **Surface area**, and **Validation**?

If you are using an AI assistant, this prompt should be enough to update the description cleanly:

> Read the current PR body for nexu-io/open-design#1566 and the template at
> Then update via `gh pr edit 1566 --repo nexu-io/open-design --body-file -`.

No rush — this just makes the follow-up review faster. 🙏

Siri-Ray (approved):
@prantikmedhi I reviewed the changed feedback component and the focused component test. The new ref/effect is scoped to the reasons panel and only scrolls after a rating opens that panel, and the test covers the expected `scrollIntoView` call when selecting feedback. I did not find actionable correctness, safety, or maintainability issues in the changed ranges. Thanks for the focused fix.

PR #1568, opened 2026-05-13
refactor(web): rename Execution mode and tighten settings dialog UI

zoeforfun:
## Why

The settings dialog grew section by section and the copy + agent grid got noisy:
- The header was pinned to the legacy "Execution & model" wording, even though the dialog now hosts ~15 sections.
- Every CLI on the support list (~14) showed a full-height "not installed" card with the "Not installed" label, the version row, and stacked action buttons, dominating the grid even after the user had already picked Claude Code months ago.
- The four-step install guide stayed visible whenever any CLI was missing, i.e. for almost everyone.
- Subtitle / hint copy across sections (chat model, CLI proxy, media providers, custom instructions, memory) had drifted into long explanatory paragraphs.

This PR is a polish pass — no protocol changes, no contract changes, no default-behavior changes outside the UI itself.

## What users will see

- **Settings → Execution & model** is now **Settings → Execution mode** everywhere (header, sidebar item, docs, e2e selectors).
- The settings header gains the Settings kicker over the section title and wraps the title row in a `modal-head-line` block so it reads consistently with the welcome variant.
- The agent grid uses `flatMap` to render the **inline test-result row** directly under the selected card (spans both columns) instead of attaching it to the bottom of the section.
- **Unavailable CLI cards** collapse to a single row: icon + name on the left, Docs / Install links on the right, no "Not installed" label and no empty version row. They keep their `aria-label="Name · Not installed"` and reduced opacity for a11y.
- The four-step install guide now renders **only when the user has no working agent picked**.
- Hint copy is trimmed across chat model, CLI proxy ("Advanced: proxy & custom paths"), media providers, custom instructions, and memory descriptions.
- New `info` Icon variant is available for use in the redesigned hints.

The CLI provider card analytics event (`settings_click_cli_provider_card`) still fires from the redesigned card's onClick.

## Surface area

- [x] **UI** — `apps/web` Settings dialog (header, agent grid, install guide, copy in many sections), new `info` Icon variant.
- [ ] **Keyboard shortcut**
- [ ] **CLI / env var**
- [ ] **API / contract**
- [ ] **Extension point**
- [x] **i18n keys** — only re-wording of existing `settings.*` keys; no new keys added.

## Screenshots

_To attach: side-by-side of the redesigned Execution mode section showing the new header, compact unavailable cards with right-aligned links, the inline test-result row under a selected card, and the install guide hidden once a working agent is picked._

## Bug fix verification

Not a bug fix — UI/copy refactor.

## Validation

- \`pnpm install\`
- \`pnpm guard\`
- \`pnpm --filter @open-design/web typecheck\`
- Manual UI walk-through pending; see screenshots section above.
- e2e \`settings-*\` selector updates compiled with their tests; running the full e2e suite locally is recommended before merge.


apps/web/src/i18n/locales/en.ts
```diff
@@ -47,8 +47,8 @@ export const en: Dict = {
   'settings.welcomeSubtitle':
     "Pick how you'd like to run generations. You can change this any time from the Settings button in the top bar.",
   'settings.kicker': 'Settings',
-  'settings.title': 'Execution & model',
-  'settings.subtitle': 'Choose between Local CLI and BYOK. Your API key is stored only in this browser.',
+  'settings.title': 'Execution mode',
```
lefarcen, line 50: P2 The rename only updates the English dictionary, but the PR promises Settings → Execution mode everywhere and this app ships many locale dictionaries. The non-English `settings.title` / `settings.envConfigure` entries still translate the old "Execution & model" / "Configure execution mode" copy, so users in those locales will keep seeing stale navigation and dialog text. Please update the matching keys in the other locale files or explicitly route this copy through a shared fallback if the team wants English-only for now.

apps/web/src/components/MemorySection.tsx
```diff
@@ -305,10 +305,34 @@ export function MemorySection() {
       saved: t('settings.memoryFlashSaved'),
       deleted: t('settings.memoryFlashDeleted'),
       indexSaved: t('settings.memoryFlashIndexSaved'),
+      // Inline English; PR-time translation sweep can hoist this into
+      // the dictionary alongside the other flash labels.
+      pathCopied: 'Path copied',
```
lefarcen: P2 This new user-facing flash string bypasses the i18n dictionary while the surrounding memory UI uses `t(...)`. In non-English locales the path-copy affordance will suddenly show English. Please add a `settings.memoryFlashPathCopied` (or similar) key across the locale dictionaries and read it through `t(...)`; the title/aria-label added for this info button should use dictionary strings too.

apps/web/src/components/SettingsDialog.tsx
```diff
@@ -2042,44 +2110,70 @@ export function SettingsDialog({
...
+                      the agent grid, the Claude Code / Codex card is no
+                      longer in view, so the model picker becomes
+                      ambiguous ("which CLI's model am I editing?").
+                      An explicit "Model for <agent>" line resolves that
+                      without forcing the user to scroll back up. Inline
+                      English; other locales pick this up at PR-time
+                      translation sweep.
+                    */}
+                    <div className="agent-model-row-head">
+                      Model for: <strong>{selected.name}</strong>
```
lefarcen: P2 This attribution label is rendered directly in English inside Settings. Since this row is part of the UI copy polish and the dialog is localized, please move "Model for:" into the settings dictionary and add the corresponding locale entries so non-English users do not get a mixed-language model picker.

lefarcen (changes requested):
Hey @zoeforfun, this is a thoughtful polish pass — the compact unavailable CLI rows and hiding the install guide once a working agent is selected both move the settings dialog in a calmer direction. I found a few i18n regressions that need tightening before this lands; see inline notes.

qiongyu1999 (dismissed):

lefarcen (dismissed):
Hey @zoeforfun — all three i18n findings from my earlier review are cleanly addressed on this head:

- ✅ Non-English locales now carry the translated "Execution mode" copy (`settings.title` updated in ar, de, ja, zh-CN, and the rest).
- ✅ The `pathCopied` flash in MemorySection routes through `t('settings.memoryFlashPathCopied')` with locale entries in en + all translations.
- ✅ No hardcoded English "Model for:" in SettingsDialog — the label is properly dict-routed.

Approving to clear my prior block. One thing still standing before merge: the **Playwright critical** step in "Validate workspace" is red on this head — looks like an e2e selector or daemon-build issue. Worth tracking down before pool review lands.

lefarcen (approved):

PR #1569, opened 2026-05-13
feat(web): add export as image screenshot to share menu

### LINKED ISSUE #1500: Add an option to export the created HTML as image screenshot.

sukumarp2022:
Fixes #1500

## Why

I hit this while using OD to build slide decks and design previews — I wanted a quick way to grab a PNG of the current preview to paste into docs or share on Slack, but the only export options were PDF, HTML, ZIP, and Markdown. Each time I had to resort to OS-level screenshots and manual cropping.

The pain: there is no user-facing image export despite the snapshot bridge already existing internally (used by the annotation/draw-and-send flow). This PR exposes that capability as a first-class export option.

## What users will see

The Share menu in the HTML file viewer gains a new **"Export as image"** option (with an image icon). Clicking it captures the current preview viewport as a PNG and downloads it. The option appears when the preview is rendered in srcdoc mode (decks, inspect mode, palette mode, draw mode, or force-inline) where the snapshot bridge is active.

## Surface area

- [x] **UI** — new menu item in the HTML viewer Share dropdown
- [x] **i18n keys** — added `fileViewer.exportImage` across all 19 locale files with proper translations
- [ ] Keyboard shortcut
- [ ] CLI / env var
- [ ] API / contract
- [ ] Extension point
- [ ] Default behavior change
- [ ] None

## Screenshots

The new "Export as image" entry appears in the Share menu after "Export as Markdown", using the `image` icon. It is conditionally shown only in srcdoc render mode where the snapshot bridge is injected (decks, inspect, palette, draw, force-inline modes).

## Validation

```bash
# Typecheck — no errors
pnpm --filter @open-design/web typecheck
# TypeScript: No errors found

# Unit tests — 38/38 passing (8 new)
pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/runtime/exports.test.ts
# Tests  38 passed (38)

# Guard check — pre-existing failure only (render.cjs), no new issues
pnpm guard
```

New tests cover: null contentWindow, successful bridge response, bridge error, timeout, mismatched id filtering, wrong source window filtering, download trigger with .png filename, and title sanitization.

## Implementation details

- **`requestPreviewSnapshot(iframe, timeout)`** — shared utility in `exports.ts` that sends `od:snapshot` postMessage to the srcdoc iframe and awaits the PNG data URL response. Includes `ev.source` guard for security and timeout cleanup.
- **`exportAsImage(dataUrl, title)`** — converts data URL to Blob (without canvas re-encoding) and triggers download as `.png`. Wrapped in try/catch with user-facing alert on failure.
- **`dataUrlToBlob(dataUrl)`** — validates `data:` prefix before decoding.
- **`PreviewDrawOverlay`** refactored to delegate to the shared `requestPreviewSnapshot()` instead of its own copy, eliminating 17 lines of duplicated logic.
- Menu item gated behind `!useUrlLoadPreview` since the snapshot bridge is only present in srcdoc mode.
- Proper translations provided for all 17 non-English locales.

## Review feedback addressed (v2)

- **i18n**: All non-English locales now have proper translated strings (ar, de, es, fa, fr, hu, id, ja, ko, pl, pt-BR, ru, th, tr, uk, zh-CN, zh-TW)
- **ev.source guard**: `requestPreviewSnapshot` now filters by `ev.source === iframe.contentWindow` matching the established pattern for iframe message handlers
- **User feedback**: alert() shown on snapshot failure (matches existing exportAsPdf pattern)
- **Data URL validation**: `dataUrlToBlob` validates `data:` prefix; `exportAsImage` wraps conversion in try/catch
- **Merge conflict**: Resolved with upstream PreviewVisualMarkKind import

## Bug fix verification

N/A — this is a new feature, not a bug fix.

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -46,6 +46,7 @@ import {
 import type { ProjectFilePreview } from '../providers/registry';
 import {
   exportAsHtml,
+  exportAsImage,
   exportAsJsx,
   exportAsMd,
   exportAsPdf,
@@ -54,6 +55,7 @@ import {
   exportReactComponentAsHtml,
   exportReactComponentAsZip,
   openSandboxedPreviewInNewTab,
+  requestPreviewSnapshot,
 } from '../runtime/exports';
 import { buildReactComponentSrcdoc } from '../runtime/react-component';
 import { buildSrcdoc } from '../runtime/srcdoc';
@@ -5485,6 +5487,33 @@ function HtmlViewer({
                     <span className="share-menu-icon"><Icon name="file" size={14} /></span>
                     <span>{t('fileViewer.exportMd')}</span>
                   </button>
+                  {!useUrlLoadPreview ? (
+                    <button
+                      type="button"
+                      className="share-menu-item"
+                      role="menuitem"
+                      onClick={async () => {
+                        setShareMenuOpen(false);
+                        const iframe = iframeRef.current;
+                        if (!iframe) return;
+                        const snap = await requestPreviewSnapshot(iframe);
+                        try {
+                          if (snap) {
+                            exportAsImage(snap.dataUrl, exportTitle);
+                          } else {
+                            console.warn('[exportAsImage] snapshot capture returned null');
+                            alert(t('fileViewer.exportImageFailed'));
+                          }
+                        } catch (err) {
+                          console.warn('[exportAsImage] failed to convert snapshot:', err);
+                          alert(t('fileViewer.exportImageFailed'));
+                        }
+                      }}
+                    >
+                      <span className="share-menu-icon"><Icon name="image" size={14} /></span>
+                      <span>{t('fileViewer.exportImage')}</span>
+                    </button>
+                  ) : null}
                   <div className="share-menu-divider" />
                   <button
                     type="button"
--- a/apps/web/src/i18n/locales/de.ts
+++ b/apps/web/src/i18n/locales/de.ts
@@ -730,6 +730,8 @@ export const de: Dict = {
   'fileViewer.exportZip': 'Als .zip herunterladen',
   'fileViewer.exportHtml': 'Als eigenständiges HTML exportieren',
   'fileViewer.exportMd': 'Als Markdown exportieren',
+  'fileViewer.exportImage': 'Als Bild exportieren',
+  'fileViewer.exportImageFailed': 'Bildaufnahme fehlgeschlagen. Bitte versuchen Sie es erneut oder verwenden Sie das Screenshot-Tool Ihres Browsers.',
   'fileViewer.exportJsx': 'Als JSX exportieren',
   'fileViewer.exportReactHtml': 'Vorschau als HTML exportieren',
   'fileViewer.saveAsTemplate': 'Als Template speichern…',
--- a/apps/web/src/i18n/locales/fr.ts
+++ b/apps/web/src/i18n/locales/fr.ts
@@ -842,6 +842,8 @@ export const fr: Dict = {
   'fileViewer.exportZip': 'Télécharger en .zip',
   'fileViewer.exportHtml': 'Exporter en HTML autonome',
   'fileViewer.exportMd': 'Exporter en Markdown',
+  'fileViewer.exportImage': 'Exporter en image',
+  'fileViewer.exportImageFailed': 'La capture d\'image a échoué. Veuillez réessayer ou utiliser l\'outil de capture d\'écran de votre navigateur.',
   'fileViewer.exportJsx': 'Exporter en JSX',
   'fileViewer.exportReactHtml': 'Exporter l\'aperçu en HTML',
   'fileViewer.saveAsTemplate': 'Enregistrer comme modèle…',
--- a/apps/web/src/i18n/locales/zh-CN.ts
+++ b/apps/web/src/i18n/locales/zh-CN.ts
@@ -925,6 +925,8 @@ export const zhCN: Dict = {
   'fileViewer.exportZip': '下载为 .zip',
   'fileViewer.exportHtml': '导出为独立 HTML',
   'fileViewer.exportMd': '导出为 Markdown',
+  'fileViewer.exportImage': '导出为图片',
+  'fileViewer.exportImageFailed': '图片捕获失败，请重试或使用浏览器的截图工具。',
   'fileViewer.exportJsx': '导出为 JSX',
   'fileViewer.exportReactHtml': '导出预览 HTML',
   'fileViewer.saveAsTemplate': '保存为模板…',
--- a/apps/web/src/runtime/exports.ts
+++ b/apps/web/src/runtime/exports.ts
@@ -317,6 +317,82 @@ export function exportAsMd(source: string, title: string): void {
   triggerDownload(blob, `${safeFilename(title, 'artifact')}.md`);
 }
 
+// ---------------------------------------------------------------------------
+// Image screenshot export
+// ---------------------------------------------------------------------------
+
+/**
+ * Request a PNG screenshot of the current viewport from the snapshot bridge
+ * injected into a srcdoc preview iframe. Returns null if the bridge is not
+ * present (e.g. URL-load mode) or the capture times out.
+ */
+export function requestPreviewSnapshot(
+  iframe: HTMLIFrameElement,
+  timeout = 2500,
+): Promise<{ dataUrl: string; w: number; h: number } | null> {
+  const win = iframe.contentWindow;
+  if (!win) return Promise.resolve(null);
+  const id = `snap-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
+  return new Promise((resolve) => {
+    let done = false;
+    function onMsg(ev: MessageEvent) {
+      if (ev.source !== win) return;
+      const d = ev.data as {
+        type?: string;
+        id?: string;
+        dataUrl?: string;
+        w?: number;
+        h?: number;
+        error?: string;
+      } | null;
+      if (!d || d.type !== 'od:snapshot:result' || d.id !== id) return;
+      if (done) return;
+      done = true;
+      window.removeEventListener('message', onMsg);
+      if (d.dataUrl && d.w && d.h) resolve({ dataUrl: d.dataUrl, w: d.w, h: d.h });
+      else resolve(null);
+    }
+    window.addEventListener('message', onMsg);
+    try {
+      win.postMessage({ type: 'od:snapshot', id }, '*');
+    } catch {
+      /* sandboxed */
+    }
+    setTimeout(() => {
+      if (!done) {
+        done = true;
+        window.removeEventListener('message', onMsg);
+        resolve(null);
+      }
+    }, timeout);
+  });
+}
+
+/** Convert a data-URL to a Blob without re-encoding through canvas. */
+function dataUrlToBlob(dataUrl: string): Blob {
+  if (!dataUrl.startsWith('data:')) {
+    throw new Error('Invalid data URL');
+  }
+  const [header, base64] = dataUrl.split(',');
+  const mime = header?.match(/:(.*?);/)?.[1] ?? 'image/png';
+  const bytes = atob(base64 ?? '');
+  const arr = new Uint8Array(bytes.length);
+  for (let i = 0; i < bytes.length; i++) arr[i] = bytes.charCodeAt(i);
+  return new Blob([arr], { type: mime });
+}
+
+/** Download a snapshot data-URL as a PNG file. */
+export function exportAsImage(dataUrl: string, title: string): void {
+  try {
+    const blob = dataUrlToBlob(dataUrl);
+    triggerDownload(blob, `${safeFilename(title, 'artifact')}.png`);
+  } catch (err) {
+    console.warn('[exportAsImage] failed to convert snapshot:', err);
+    // Re-throw the error to allow the caller to handle UI feedback
+    throw err;
+  }
+}
+
 export type ProjectPdfExportResult = 'desktop' | 'fallback';
 
 export async function exportProjectAsPdf(opts: {
--- a/apps/web/src/components/PreviewDrawOverlay.tsx
+++ b/apps/web/src/components/PreviewDrawOverlay.tsx
@@ -2,6 +2,7 @@ import { useCallback, useEffect, useRef, useState, type CSSProperties, type Poin
 
 import { Icon } from './Icon';
 import type { PreviewVisualMarkKind } from '../types';
+import { requestPreviewSnapshot } from '../runtime/exports';
 
 export type PreviewDrawMode = 'click' | 'draw';
 
@@ -205,25 +206,9 @@ export function PreviewDrawOverlay({
   }
 
   async function requestSnapshot(): Promise<{ dataUrl: string; w: number; h: number } | null> {
-    const iframe = wrapRef.current?.querySelector('iframe');
-    const win = iframe?.contentWindow;
-    if (!iframe || !win) return null;
-    const id = `snap-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
-    return new Promise((resolve) => {
-      let done = false;
-      function onMsg(ev: MessageEvent) {
-        const d = ev.data as { type?: string; id?: string; dataUrl?: string; w?: number; h?: number; error?: string } | null;
-        if (!d || d.type !== 'od:snapshot:result' || d.id !== id) return;
-        if (done) return;
-        done = true;
-        window.removeEventListener('message', onMsg);
-        if (d.dataUrl && d.w && d.h) resolve({ dataUrl: d.dataUrl, w: d.w, h: d.h });
-        else resolve(null);
-      }
-      window.addEventListener('message', onMsg);
-      try { win.postMessage({ type: 'od:snapshot', id }, '*'); } catch { /* sandboxed */ }
-      setTimeout(() => { if (!done) { done = true; window.removeEventListener('message', onMsg); resolve(null); } }, 2500);
-    });
+    const iframe = wrapRef.current?.querySelector('iframe') as HTMLIFrameElement | null;
+    if (!iframe) return null;
+    return requestPreviewSnapshot(iframe);
   }
 
   function drawCaptureTarget(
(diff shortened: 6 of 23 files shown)


apps/web/src/i18n/locales/de.ts
```diff
@@ -730,6 +730,7 @@ export const de: Dict = {
   'fileViewer.exportZip': 'Als .zip herunterladen',
   'fileViewer.exportHtml': 'Als eigenständiges HTML exportieren',
   'fileViewer.exportMd': 'Als Markdown exportieren',
+  'fileViewer.exportImage': 'Export as image',
```
lefarcen: P2 This new user-facing label is still English in the non-English locale files. Since this PR exposes a new Share-menu action and marks i18n as touched, please localize `fileViewer.exportImage` across the locale files instead of copying `Export as image` everywhere.

lefarcen (changes requested):
Hi @sukumarp2022, thanks for jumping on #1500 so quickly — reusing the existing snapshot bridge is a nice, focused direction for the current-viewport export.

I found one user-visible localization issue that should be fixed before this lands; details inline. While you are updating the PR, please also fill the template gaps for **Screenshots** and **Validation** so product/review can verify the new Share-menu entry quickly.

If you are using an AI assistant, this prompt should get the PR body into shape:

> Read the current PR body for nexu-io/open-design#1569 and the template at
> - **Screenshots**: add the Share menu entry-point screenshot, ideally before/after or showing the new Export as image item
> - **Validation**: list exactly what was run for this branch, for example the relevant web test/typecheck commands
> Then update via `gh pr edit 1569 --repo nexu-io/open-design --body-file -`.

Once the locale strings and PR-body evidence are tightened up, this should be straightforward to re-review. 🙌

apps/web/src/i18n/locales/zh-CN.ts
```diff
@@ -925,6 +925,7 @@ export const zhCN: Dict = {
   'fileViewer.exportZip': '下载为 .zip',
   'fileViewer.exportHtml': '导出为独立 HTML',
   'fileViewer.exportMd': '导出为 Markdown',
+  'fileViewer.exportImage': 'Export as image',
```
PerishCode: **Untranslated i18n value in 18 non-English locale files**

The new `fileViewer.exportImage` key was added with the English string `'Export as image'` in every non-English locale file (`ar`, `de`, `es-ES`, `fa`, `fr`, `hu`, `id`, `ja`, `ko`, `pl`, `pt-BR`, `ru`, `th`, `tr`, `uk`, `zh-CN`, `zh-TW`, and previously `id`), while all neighboring `fileViewer.export*` keys (`exportZip`, `exportHtml`, `exportMd`, `exportJsx`, `exportReactHtml`) are properly localized.

**Why it matters**: users in those locales will see one English item interleaved with otherwise translated menu items in the Share menu, which is visibly inconsistent. The PR description itself frames this as “added `fileViewer.exportImage` across all 19 locale files”, but only the key is added — the values are not localized. The Surface area checklist in the PR body checks **i18n keys**, so this is the right place to honor it.

**Evidence**: e.g. `zh-CN.ts` line 928 shows `'fileViewer.exportImage': 'Export as image',` immediately above `'fileViewer.exportJsx': '导出为 JSX',`. Same pattern repeats in the other 17 non-English diffs.

**Suggested change**: translate the value in each non-English locale file using the same translator convention used for the neighboring `fileViewer.export*` strings — for example:
- `zh-CN.ts`: `'fileViewer.exportImage': '导出为图片',`
- `zh-TW.ts`: `'fileViewer.exportImage': '匯出為圖片',`
- `ja.ts`: `'fileViewer.exportImage': '画像としてエクスポート',`
- `de.ts`: `'fileViewer.exportImage': 'Als Bild exportieren',`
- `fr.ts`: `'fileViewer.exportImage': 'Exporter en image',`
- `es-ES.ts`: `'fileViewer.exportImage': 'Exportar como imagen',`

…and similarly for the remaining locales. If the repo's policy is that English fallback is acceptable until translators land, please mention that in the PR body so reviewers know it's an intentional gap rather than a missed surface; otherwise treat this as an i18n completion task.


apps/web/src/components/FileViewer.tsx
```diff
@@ -5322,6 +5324,27 @@ function HtmlViewer({
...
+                      role="menuitem"
+                      onClick={async () => {
+                        setShareMenuOpen(false);
+                        const iframe = iframeRef.current;
+                        if (!iframe) return;
+                        const snap = await requestPreviewSnapshot(iframe);
+                        if (snap) {
+                          exportAsImage(snap.dataUrl, exportTitle);
+                        } else {
+                          console.warn('[exportAsImage] snapshot capture returned null');
```
PerishCode: **Silent failure on user-initiated export — only `console.warn` is fired**

In the new `Export as image` menu handler, when `requestPreviewSnapshot()` returns `null`, the menu is already closed (line 5333) and the only feedback is a `console.warn` to the devtools. The user clicked the action and observes nothing — no download, no error message, no toast.

`requestPreviewSnapshot()` returns `null` in three distinct, realistic cases (per the function's own docstring and tests):
1. `iframe.contentWindow` is null (sandboxed/cross-origin iframe).
2. The bridge replies with an error (`'snapshot image failed'` is raised by `srcdoc.ts:155` when the SVG image fails to load — e.g. tainted canvas due to a cross-origin `` in the artifact).
3. 2500 ms timeout.

All three are user-visible failures on a button click, not internal best-effort calls.

**Why it matters**: the rest of the Share menu's export actions (`exportAsHtml`, `exportAsMd`, `exportAsJsx`, `exportProjectAsPdf`, etc.) either succeed with a download or surface a visible failure path. Adding a silently-failing item degrades the perceived reliability of the menu and makes the feature look broken in exactly the cases where the underlying bridge is the most likely to fail (large designs, cross-origin assets, slow paint).

**Evidence**:
```tsx
const snap = await requestPreviewSnapshot(iframe);
if (snap) {
  exportAsImage(snap.dataUrl, exportTitle);
} else {
  console.warn('[exportAsImage] snapshot capture returned null');
}
```
No toast, no error UI, no telemetry.

**Suggested change**: surface a user-visible failure path consistent with how this file already reports recoverable errors. Two options, in order of preference:

1. Reuse whatever toast/notice surface this file uses for other export failures (e.g. the same path PDF/zip exports use when they fall back) — search for the existing `setShareError`/`pushToast`/equivalent in `FileViewer.tsx` and wire that into the `else` branch.
2. At minimum, raise the timeout from the default 2500 ms (large designs commonly need longer to paint and serialize) and add a one-line user-facing message — e.g.:
   ```tsx
   const snap = await requestPreviewSnapshot(iframe, 8000);
   if (snap) {
     exportAsImage(snap.dataUrl, exportTitle);
   } else {
     // existing toast/notice mechanism in this file
     showExportError(t('fileViewer.exportImageFailed'));
   }
   ```
   …and add a `fileViewer.exportImageFailed` i18n key alongside `fileViewer.exportImage`.

If user-visible feedback is genuinely out of scope, the `console.warn` should at least mention which of the three failure modes was hit (null contentWindow vs. bridge error vs. timeout) so this is debuggable in the wild.


apps/web/src/runtime/exports.ts
```diff
@@ -317,6 +317,72 @@ export function exportAsMd(source: string, title: string): void {
...
+    function onMsg(ev: MessageEvent) {
+      const d = ev.data as {
+        type?: string;
+        id?: string;
+        dataUrl?: string;
+        w?: number;
+        h?: number;
+        error?: string;
+      } | null;
+      if (!d || d.type !== 'od:snapshot:result' || d.id !== id) return;
```
PerishCode, line 348: **`onMsg` doesn't validate `ev.source` — diverges from the established message-handler pattern in `FileViewer.tsx`**

The new `onMsg` listener accepts any cross-window message whose `type === 'od:snapshot:result'` and whose `id` matches the local id, regardless of which window sent it:
```ts
function onMsg(ev: MessageEvent) {
  const d = ev.data as { /* … */ } | null;
  if (!d || d.type !== 'od:snapshot:result' || d.id !== id) return;
  // …
}
```
Everywhere else `FileViewer.tsx` consumes iframe `postMessage` traffic, the first guard is `ev.source !== iframeRef.current?.contentWindow` (see FileViewer.tsx lines 3779, 3865, 3994, 4100, 4382). The random id mitigates the practical risk, but this is the only listener in the system that diverges from that pattern, and the divergence ships in a *shared utility* that future callers will reuse.

**Why it matters**: any other frame on the page (e.g. an embedded sandboxed preview spawned by `openSandboxedPreviewInNewTab`, an external iframe in a future host integration, or a misbehaving DOM extension) could in principle deliver a forged result by guessing or observing the id. More mundanely: as the snapshot bridge gets reused beyond the single iframe in `HtmlViewer`, the absence of a source check makes it easy to cross-wire two concurrent previews. The fix is cheap and matches the pattern reviewers will scan for.

**Evidence**: the original `PreviewDrawOverlay.requestSnapshot()` (before this PR's refactor) had the same gap, so this is not a regression you introduced — but the patch moves the logic into a shared `runtime/exports.ts` helper, which is the right time to harden it once for every caller. Compare with `apps/web/src/components/FileViewer.tsx:3779`:
```ts
if (ev.source !== iframeRef.current?.contentWindow) return;
```

**Suggested change**: capture the iframe's `contentWindow` once and guard on it inside the listener:
```ts
export function requestPreviewSnapshot(
  iframe: HTMLIFrameElement,
  timeout = 2500,
): Promise<{ dataUrl: string; w: number; h: number } | null> {
  const win = iframe.contentWindow;
  if (!win) return Promise.resolve(null);
  const id = `snap-${Date.now()}-${Math.random().toString(36).slice(2, 8)}`;
  return new Promise((resolve) => {
    let done = false;
    function onMsg(ev: MessageEvent) {
      if (ev.source !== win) return; // <- add this
      const d = ev.data as { /* … */ } | null;
      if (!d || d.type !== 'od:snapshot:result' || d.id !== id) return;
      // …
    }
    // …
  });
}
```
The existing tests will keep passing because their dispatched events don't set `source`; you'll want to extend the success and id-mismatch test cases to pass `source: iframe.contentWindow` and add one new case asserting that a message from a different window is ignored.


PerishCode (commented):
@sukumarp2022 nice reuse of the existing `injectSnapshotBridge` — the refactor that pulls `PreviewDrawOverlay`'s snapshot path into a shared `requestPreviewSnapshot()` cleanly removes the dup, and the unit-test coverage (null contentWindow / success / error / timeout / id-mismatch / filename sanitization) looks good. Three non-blocking comments inline:

- New i18n key landed as an English string in all 18 non-English locale files even though neighboring `fileViewer.export*` keys are translated.
- `Export as image` click silently does nothing when the snapshot bridge returns `null` — only a `console.warn` fires, no user-visible feedback.
- `requestPreviewSnapshot()`'s message listener doesn't filter on `ev.source`, which is the established pattern for `iframeRef` message handlers in `FileViewer.tsx`.

None are merge-blockers, but each is a small win if you'd like to address before merge.



lefarcen (changes requested):
Hey @sukumarp2022, I re-checked the current head after PerishCode's review. The implementation direction is still solid, but the same blocking i18n item from my earlier review remains: `fileViewer.exportImage` is still English in the non-English locale files, so localized users would see one untranslated Share-menu action among translated export items.

No new inline comments from me here — PerishCode's notes cover the other useful polish items, especially visible feedback when snapshot capture returns `null` and matching the existing `ev.source` guard pattern. Once the locale values, screenshots, and validation notes are updated, we can take another pass quickly. 🙌

lefarcen (dismissed):
Hey @sukumarp2022, thanks for the quick iteration. I re-reviewed the new head and the blocker from my earlier review is addressed: the export label is now localized across the non-English locale files, and the PR body now has the screenshots/validation evidence reviewers need.

The source-window guard and user-visible failure path from PerishCode's notes are also covered. No blockers from my side on this head. Product review and CI still need to finish before this is merge-ready. 🙌

apps/web/src/components/FileViewer.tsx
```diff
@@ -5485,6 +5487,30 @@ function HtmlViewer({
...
+                          if (typeof alert !== 'undefined') {
+                            alert('Image capture failed. Please try again or use your browser\'s screenshot tool.');
```
PerishCode: **Failure dialog stays English while the menu label is localized 19 ways**

The new error path on snapshot capture failure surfaces the message via a hardcoded English `alert()`:

```tsx
alert('Image capture failed. Please try again or use your browser\'s screenshot tool.');
```

In a French / Japanese / Arabic / Russian / etc. Share menu, a user clicks the freshly translated `Exporter en image` / `画像としてエクスポート` / `تصدير كصورة` / `Экспорт как изображение` item, hits a failure, and gets a single English modal — visibly inconsistent with the rest of the menu and with the i18n keys this PR explicitly added across 19 locales.

**Why it matters**: i18n completeness was the main blocker on the first review pass, and the resolution was specifically to translate the new key across all locales. Leaving the user-facing failure copy hardcoded undoes that effort in the exact moment the user actually needs to read the message — when the action they just clicked silently doesn't produce a download. Realistic failure modes covered by `requestPreviewSnapshot()` (null `contentWindow` on sandboxed iframes, `'snapshot image failed'` from `srcdoc.ts:155` when an artifact contains a cross-origin `` that taints the canvas, 2500 ms timeout on large designs) will hit this path in production.

**Evidence**:
- `apps/web/src/i18n/types.ts` adds `'fileViewer.exportImage'` only — no `'fileViewer.exportImageFailed'`.
- All 19 locale files translate `fileViewer.exportImage`, but the alert at `apps/web/src/components/FileViewer.tsx:5505` is a string literal.
- `apps/web/src/components/FileWorkspace.tsx:424` is the in-repo precedent for `alert(t(...))`: `alert(t('workspace.deleteSelectedFilesPartial', { n: failed.length }))`.
- The earlier review thread (`#discussion_r3233939907`) proposed exactly this `'fileViewer.exportImageFailed'` key — the localization landed for the menu label but didn't follow through to the failure copy.

**Side note on the guard**: `if (typeof alert !== 'undefined')` is unnecessary here — `window.alert` is always present in the browser runtime where this `onClick` runs, the other 4 in-tree `alert()` callers (`DesignFilesPanel.tsx:301`, `FileWorkspace.tsx:424`, `exports.ts:628`, `exports.ts:653`) do not gate the call, and dropping it keeps the site consistent.

**Suggested change**: add an i18n key for the failure copy, translate it alongside `fileViewer.exportImage` across all 19 locale files, and route the alert through `t()`:

```tsx
// apps/web/src/components/FileViewer.tsx
const snap = await requestPreviewSnapshot(iframe);
if (snap) {
  exportAsImage(snap.dataUrl, exportTitle);
} else {
  console.warn('[exportAsImage] snapshot capture returned null');
  alert(t('fileViewer.exportImageFailed'));
}
```

```ts
// apps/web/src/i18n/types.ts — add next to fileViewer.exportImage
'fileViewer.exportImageFailed': string;

// apps/web/src/i18n/locales/en.ts
'fileViewer.exportImageFailed': 'Image capture failed. Please try again or use your browser\'s screenshot tool.',

// e.g. zh-CN.ts: '图片捕获失败，请重试或使用浏览器的截图工具。'
// e.g. ja.ts:    '画像のキャプチャに失敗しました。再試行するか、ブラウザのスクリーンショット機能をご利用ください。'
// e.g. fr.ts:    'La capture d\'image a échoué. Veuillez réessayer ou utiliser l\'outil de capture d\'écran de votre navigateur.'
// …and similarly for the other 16 locales.
```

Non-blocking — the menu still works and the English fallback is intelligible — but worth folding in alongside the i18n batch you already touched so the locale story stays whole. If you'd rather defer it, a follow-up issue noting the gap is fine; the rest of the PR is in good shape.


PerishCode (commented):
@sukumarp2022 thanks for the quick turnaround — the three notes from the previous pass are wired through cleanly on this head: every non-English locale now carries a translated `fileViewer.exportImage` value, `requestPreviewSnapshot()` guards on `ev.source === iframe.contentWindow` (with a dedicated test for the wrong-source case), and the Share-menu handler no longer fails silently. The `PreviewDrawOverlay` refactor onto the shared helper also reads well and the unit-test matrix (null contentWindow / success / bridge error / timeout / id-mismatch / wrong-source / filename sanitization) covers the interesting branches.

One small non-blocking polish item inline on the new failure path — flagging because it sits next to the i18n surface this PR explicitly opened.



apps/web/src/i18n/locales/fr.ts
```diff
@@ -842,6 +842,9 @@ export const fr: Dict = {
   'fileViewer.exportZip': 'Télécharger en .zip',
   'fileViewer.exportHtml': 'Exporter en HTML autonome',
   'fileViewer.exportMd': 'Exporter en Markdown',
+  'fileViewer.exportImage': 'Exporter en image',
+
+  'fileViewer.exportImageFailed': 'La capture d'image a échoué. Veuillez réessayer ou utiliser l'outil de capture d'écran de votre navigateur.',
```
lefarcen: P1 This French string contains unescaped apostrophes inside a single-quoted TS literal (`d'image`, `l'outil`, `d'écran`), so the locale file will not parse and CI/build fails. Escape those apostrophes (`d\'image`, etc.) or switch this value to a template/string form that preserves the French text without breaking TypeScript.

lefarcen (changes requested):
Hey @sukumarp2022, thanks for folding in the localized failure message too — that closes the UX loop nicely. I found one small but blocking syntax issue in the new French string that lines up with the current failing checks; see the inline note.

Once that quote escaping is fixed, this should be a quick re-check. 🙌

lefarcen (approved):
Hey @sukumarp2022, thanks for the quick fix. I re-reviewed this head and the French locale syntax issue is resolved; CI is green again, and the localized failure message path now looks consistent with the rest of the export UI.

No blockers from my side on this head. Product approval is still pending before this can be merge-ready. 🙌

PerishCode (approved):
@sukumarp2022 nice work on this head — every thread from the prior rounds is fully resolved: the i18n batch now ships translated `fileViewer.exportImage` *and* `fileViewer.exportImageFailed` values across all 19 locales (no English leakage in the menu or the failure modal), `requestPreviewSnapshot()` guards on `ev.source === iframe.contentWindow` with a dedicated test case, the Share-menu handler wraps both the `null` snapshot path and the `dataUrlToBlob` throw path with a localized `alert(t('fileViewer.exportImageFailed'))`, and the French apostrophe-escape issue that broke CI on the previous head is fixed. The `PreviewDrawOverlay` refactor onto the shared helper cleanly removes the duplicated postMessage plumbing, and the unit-test matrix (null contentWindow / success / bridge error / timeout / id mismatch / wrong source / filename sanitization) covers the interesting branches. Verified the typecheck/test/build checks are green on this commit and the menu remains gated behind `!useUrlLoadPreview` so the new entry only surfaces where the snapshot bridge is actually injected. Thanks for the patient iteration through three review rounds — the result reads well and is a clean addition to the export surface. 🙌

Issue #1557, opened 2026-05-13
Orbit summary hides agent data-loading failure reason when no live artifact is registered


shangxinyu1:
## Summary

An Orbit run can complete the agent turn successfully, with the agent clearly explaining why it did not create a live artifact, but the Orbit activity summary only shows the generic message:

```text
Agent succeeded but did not register a live artifact for this Orbit run.
```

This hides the actionable failure reason from the Settings > Orbit UI.

## Observed local run

Observed in the packaged desktop stable namespace.

- conversation: `orbit-conv-5caccbbf-7dce-4c75-808c-d766a7367adb`
- project: `orbit-9aa0d9ce-51c5-44fb-bf03-2301f090c757`
- agent run: `8ebe047b-3a87-402e-a296-a5189ad2285c`
- title: `Orbit · 2026-05-13 16:44`
- started: `2026-05-13T08:44:53.382Z`
- completed: `2026-05-13T08:48:58.573Z`
- template: `orbit-general`
- trigger: manual

The Orbit activity summary records:

```json
{
  "connectorId": "agent-runtime",
  "connectorName": "Orbit Agent",
  "status": "failed",
  "summary": "Agent succeeded but did not register a live artifact for this Orbit run."
}
```

However, the assistant message for the same run contains the useful terminal reason:

```text
Data loading failed, so I did not create a daily digest artifact.

The available connected sources in this run were not usable for a read-only digest: GitHub access could not be read with the current authorization, and the Notion connection has expired. Once one of those sources is readable again, I can generate and register the Orbit live artifact for the 2026-05-12 16:44 to 2026-05-13 16:44 window.
```

The project directory also confirms no live artifact was created; it only contains staged skills and `.daily-digest-tmp/*.json` connector scratch outputs.

## Current behavior

The run appears as a generic Orbit failure with no actionable remediation in the Orbit summary.

Users cannot tell from the summary that the real problem is source auth/expiry:

- GitHub authorization could not read the required data.
- Notion connection had expired.

## Expected behavior

When the agent run succeeds but no live artifact target is registered, Orbit should preserve the agent's final explanation in the run summary, or at least append it to the generic no-artifact message.

For this case, the Orbit summary should show something like:

```text
No live artifact was registered because data loading failed: GitHub access could not be read with the current authorization, and the Notion connection has expired.
```

## Why this matters

[...]

PR #1570, opened 2026-05-13
fix(orbit): surface the failure reason in no-artifact summaries

prantikmedhi:
## Why
Orbit summaries needed to keep a useful final explanation even when no live artifact was registered. Without that, the fallback message was too generic and hid the actual failure context.

## What users will see
- No-artifact runs now keep a compact excerpt from the assistant's final explanation.
- The generic fallback still exists, but it no longer drops the useful context on the floor.
- The summary reads more like the actual run result and less like a generic failure.

## Surface area
- [x] Daemon
- [x] Tests
- [ ] UI
- [ ] API
- [ ] Docs
- [ ] None

## Validation
- `pnpm exec vitest run -c vitest.config.ts tests/orbit-summary.test.ts`
- `pnpm exec tsc -p tsconfig.json --noEmit`

### DIFF
--- a/apps/daemon/src/orbit-summary.ts
+++ b/apps/daemon/src/orbit-summary.ts
@@ -1,0 +1,25 @@
+export function formatOrbitCompletionSummary({
+  status,
+  artifactTitle,
+  assistantMessage,
+}: {
+  status: string;
+  artifactTitle?: string | null;
+  assistantMessage?: string | null;
+}): string {
+  if (artifactTitle) {
+    return `Agent ${status} and registered live artifact ${artifactTitle}.`;
+  }
+
+  const detail = compactSummaryText(assistantMessage);
+  return detail
+    ? `Agent ${status} but did not register a live artifact for this Orbit run: ${detail}`
+    : `Agent ${status} but did not register a live artifact for this Orbit run.`;
+}
+
+function compactSummaryText(text: string | null | undefined, maxLength = 220): string {
+  const firstBlock = typeof text === 'string' ? text.trim().split(/\n\s*\n/, 1)[0]?.trim() ?? '' : '';
+  const singleLine = firstBlock.replace(/\s+/g, ' ');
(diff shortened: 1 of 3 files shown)


lefarcen (commented):
Hey @prantikmedhi, thanks for jumping on this Orbit summary bug. The implementation direction looks good: extracting the no-artifact summary formatting into a focused helper and covering artifact / excerpt / fallback behavior makes the fix easy to reason about.

One process item before the next pass: the PR body is still missing a few sections from the repo template, which helps reviewers and release verification understand user-visible changes quickly. Could you fill in:

- **Why**
- **What users will see**
- **Surface area**
- **Validation** — your test command is useful; please put it in the template’s validation section and include anything else you ran

If you’re using an AI assistant, you can paste this prompt to update the description cleanly:

> Read the current PR body for nexu-io/open-design#1570 and the template at
> Then update via `gh pr edit 1570 --repo nexu-io/open-design --body-file -`.

No rush — this just helps the next review iteration go faster. 🙏

mrcfps (approved):
@prantikmedhi I verified the Orbit summary path here: the completion summary now uses the corrected failed status when no artifact is registered, and it preserves a compact excerpt from the assistant's final explanation instead of dropping back to the generic message. I also reran the focused Orbit summary Vitest coverage and the daemon typecheck locally. Thanks for tightening up this failure path — the extracted helper and focused tests make the behavior much easier to follow.


lefarcen:
Hey @prantikmedhi, thanks again for putting this together.

We ended up merging PR #1576 for the same #1557 Orbit no-artifact summary path. That landed the user-facing behavior we needed — preserving the no-live-artifact failure marker while carrying through the assistant’s actionable explanation — and #1557 is now closed from that merge.

I’m going to close this one as superseded so the queue stays tidy. Appreciate you helping tighten up this failure path. 🙏

PR #1496, opened 2026-05-13
feat(web): render GFM tables in markdown artifact and chat renderers

mutoe:
## Summary

- Both hand-rolled markdown parsers in `apps/web` (`artifacts/markdown.ts` for artifact preview, `runtime/markdown.tsx` for assistant chat) ignored GFM pipe-table syntax, so a `| a | b |` + `|---|---|` block collapsed into a run-on `<p>` of literal pipes.
- This PR teaches both parsers a single new block kind — pipe table — without pulling in a markdown library.
- Closes #1495.

## What changed

- `apps/web/src/artifacts/markdown.ts`: new `parseAlignRow` / `splitCells` / `isTableStart` helpers; emits `<table class="md-table"><thead><th></th></thead><tbody><td></td></tbody></table>` HTML; cell contents go through the existing `formatInline` so inline code, bold, italic, and safe links keep working.
- `apps/web/src/runtime/markdown.tsx`: new `Block` variant `'table'` with parsed `headers`, `aligns`, `rows`; rendered as a React `<table>` with per-column inline `text-align`; cells run through the existing `renderInline` so URLs/code/bold/italic still tokenize correctly.
- Paragraph greedy-loops in both parsers now break at a table start, so `Some paragraph\n| h | h |\n|---|---|` no longer eats the table into the paragraph.
- `apps/web/src/index.css`: small `.md-table` rule under `.prose-block` — collapsed borders, horizontal overflow scroll, header weight bump. No new tokens.
- Tests in `apps/web/tests/artifacts/markdown.test.ts` and `apps/web/tests/runtime/markdown.test.tsx` cover: minimal table, `:---`/`---:`/`:---:` alignment, inline markup in cells, escaped `\|`, padding of short rows, paragraph-to-table boundary, stray-pipe lines not being promoted to tables.

## What's still not supported (on purpose)

- Tables without leading/trailing `|` are allowed (GFM-style), but cells stay single-line — no nested block content.
- No new dependency. Still no `react-markdown` / `remark-gfm`. Cell contents stay sanitized via the existing inline pipeline; `<table>` markup is rendered as React elements (runtime) or string-built HTML with escaped cell content (artifacts).

## Test plan

- [x] `pnpm --filter @open-design/web test --run tests/artifacts/markdown.test.ts tests/runtime/markdown.test.tsx` — 24/24 pass.
- [x] `pnpm guard` — pass.
- [x] `pnpm --filter @open-design/web typecheck` — pass.
[...]

Issue #1495, opened 2026-05-13
feat(web): render GFM tables in markdown artifact and chat renderers


mutoe:
### Problem

Markdown tables (`| col | col |` + `|---|---|` syntax) are not rendered in either the artifact preview or the assistant chat message. The pipe-delimited rows collapse into a single paragraph because the parsers join consecutive non-empty lines with a space, so the visual result is one run-on line of `|` characters and cell contents.

Reproduction: open any artifact or assistant message containing a standard GFM table. Expected: an HTML `<table>`. Actual: a `<p>` paragraph with literal pipes.

### Root cause

There are two hand-rolled markdown parsers in `apps/web`, and neither models a table block:

- `apps/web/src/artifacts/markdown.ts` — artifact preview path. The comment at lines 66–71 explicitly states tables are *not supported on purpose* (alongside nested lists, raw HTML blocks, escaped markdown syntax).
- `apps/web/src/runtime/markdown.tsx` — assistant chat path. The `Block` type union is `'p' | 'h' | 'ul' | 'ol' | 'code' | 'hr'`; no table variant exists.

No GFM-aware dependency (`remark-gfm`, `react-markdown`, `markdown-it`, `marked`) is present in the workspace. Both parsers were intentionally kept small for bundle size and to avoid `dangerouslySetInnerHTML` exposure.

### Proposal

Add minimal GFM pipe-table support to both parsers:

1. Detect a table block when a `|`-containing line is followed by a `|---|---|`-style alignment row.
2. Parse header row, alignment row (`:---`, `---:`, `:---:`), and body rows.
3. Render as `<table><thead><tr><th>` / `<tbody><tr><td>`. Keep `text-align` from the alignment row.
4. Reuse existing inline formatter for cell contents (so `code`, `**bold**`, links inside cells keep working).
5. Add unit tests under `apps/web/tests/artifacts/markdown.test.ts` and `apps/web/tests/runtime/markdown.test.tsx` covering: minimal table, alignment markers, empty cells, escaped pipes (`\|`), table mixed with surrounding paragraphs.

### Out of scope

- Full CommonMark / GFM extension parity (no plan to swap to `react-markdown` here; that would expand the security surface).
- Nested block content inside cells (cells stay single-line inline content, per GFM spec).

### Affected files (anticipated)

- `apps/web/src/artifacts/markdown.ts`
- `apps/web/src/runtime/markdown.tsx`
- `apps/web/tests/artifacts/markdown.test.ts`
[...]

PR #1574, opened 2026-05-13
feat(web): render GFM tables in markdown

prantikmedhi:
## Why
GFM tables were not rendering reliably in the safe HTML preview or the chat markdown renderer. Common cases like intro text before a table and `|` characters inside inline code still broke the layout.

## What users will see
- Markdown tables render as tables in both artifact preview and chat.
- Tables still work when they follow a short intro line without a blank separator.
- Pipes inside inline code spans stay literal instead of creating phantom columns.
- Malformed separator rows no longer get promoted into tables.

## Surface area
- [x] UI
- [x] Tests
- [ ] API
- [ ] Daemon
- [ ] Docs
- [ ] None

## Validation
- `git diff --check`
- `pnpm --dir apps/web exec vitest run tests/artifacts/markdown.test.ts tests/runtime/markdown.test.tsx`


apps/web/src/artifacts/markdown.ts
```diff
@@ -63,12 +63,86 @@ function headingLevel(line: string): number {
...
+  let current = '';
+  for (let i = 0; i < trimmed.length; i += 1) {
+    const ch = trimmed[i] ?? '';
+    const next = trimmed[i + 1] ?? '';
+    if (ch === '\\' && next === '|') {
+      current += '|';
+      i += 1;
+      continue;
+    }
+    if (ch === '|') {
```
lefarcen: P2 This treats every unescaped `|` as a column boundary, including pipes inside inline code spans such as ``| status | `"ready" | "done"` |``. That is a common table-cell pattern for TypeScript unions and it will create phantom columns. Track whether the scanner is inside a backtick code span before splitting, and add an artifact-renderer regression test for that case.

apps/web/src/runtime/markdown.tsx
```diff
@@ -27,9 +27,54 @@ type Block =
...
+  let current = '';
+  for (let i = 0; i < trimmed.length; i += 1) {
+    const ch = trimmed[i] ?? '';
+    const next = trimmed[i + 1] ?? '';
+    if (ch === '\\' && next === '|') {
+      current += '|';
+      i += 1;
+      continue;
+    }
+    if (ch === '|') {
```
lefarcen: P2 Same code-span issue in the chat renderer: a `|` inside backticks is cell content, not a separator. Please make this splitter code-span-aware too, otherwise artifact preview and chat will both misrender union-type cells in tables.

apps/web/src/artifacts/markdown.ts
```diff
@@ -96,6 +170,38 @@ export function renderMarkdownToSafeHtml(markdown: string): string {
       continue;
     }
 
+    if (
```
lefarcen: P1 Table detection only runs when the current top-level line is already the header. If a model emits an intro line immediately followed by a table with no blank line, the paragraph collector below never treats the table header as a block starter and the table still collapses into a paragraph. Add an `isTableStart` check to the paragraph break conditions and cover the no-blank-line intro case.

apps/web/src/runtime/markdown.tsx
```diff
@@ -55,6 +100,38 @@ function parseBlocks(input: string): Block[] {
       out.push({ kind: 'code', lang, body: buf.join('\n') });
       continue;
     }
+    // Table.
```
lefarcen, line 126: P1 The chat parser has the same paragraph-interruption gap: `Intro
| A | B |
|---|---|` is consumed as one paragraph before this table branch can run. Please make the paragraph loop break when the next line starts a table, so common assistant output renders the table without requiring a blank line.

lefarcen (changes requested):
Hi @prantikmedhi! Thanks for jumping on the GFM table gap — keeping this inside the two existing safe markdown renderers is the right direction, and the added styling/tests make the intended UX clear.

I found a couple of correctness blockers where common table output would still render incorrectly in one or both paths; details are inline. One process note too: could you fill the PR template sections that are currently missing — **Why**, **What users will see**, **Surface area**, and **Validation**? That helps reviewers and product understand the user-visible scope quickly. If you use an AI assistant, this prompt should work:

> Read the current PR body for nexu-io/open-design#1574 and the template at `.github/pull_request_template.md`. Rewrite the description filling in:
> - **What users will see**: user perspective — markdown tables now render as tables in artifact preview and chat markdown
> - **Surface area**: check every applicable box; for this PR include the relevant Web/UI surface
> - **Validation**: what you actually ran (`pnpm --dir apps/web exec vitest run tests/artifacts/markdown.test.ts tests/runtime/markdown.test.tsx`)
> Then update via `gh pr edit 1574 --repo nexu-io/open-design --body-file -`.

Once these parser cases are covered, this should be in a much stronger spot. 🙌

apps/web/src/artifacts/markdown.ts
lefarcen: P2 This treats every unescaped `|` as a column boundary, including pipes inside inline code spans such as ``| status | `"ready" | "done"` |``. That is a common table-cell pattern for TypeScript unions and it will create phantom columns. Track whether the scanner is inside a backtick code span before splitting, and add an artifact-renderer regression test for that case.

apps/web/src/runtime/markdown.tsx
lefarcen: P2 Same code-span issue in the chat renderer: a `|` inside backticks is cell content, not a separator. Please make this splitter code-span-aware too, otherwise artifact preview and chat will both misrender union-type cells in tables.

apps/web/src/artifacts/markdown.ts
lefarcen: P1 Table detection only runs when the current top-level line is already the header. If a model emits an intro line immediately followed by a table with no blank line, the paragraph collector below never treats the table header as a block starter and the table still collapses into a paragraph. Add an `isTableStart` check to the paragraph break conditions and cover the no-blank-line intro case.

apps/web/src/runtime/markdown.tsx
lefarcen, line 126: P1 The chat parser has the same paragraph-interruption gap: `Intro\n| A | B |\n|---|---|` is consumed as one paragraph before this table branch can run. Please make the paragraph loop break when the next line starts a table, so common assistant output renders the table without requiring a blank line.

lefarcen (changes requested):
Hi @prantikmedhi! Thanks for jumping on the GFM table gap — keeping this inside the two existing safe markdown renderers is the right direction, and the added styling/tests make the intended UX clear.

I found a couple of correctness blockers where common table output would still render incorrectly in one or both paths; details are inline. One process note too: could you fill the PR template sections that are currently missing — **Why**, **What users will see**, **Surface area**, and **Validation**? That helps reviewers and product understand the user-visible scope quickly. If you use an AI assistant, this prompt should work:

> Read the current PR body for nexu-io/open-design#1574 and the template at `.github/pull_request_template.md`. Rewrite the description filling in:
> - **What users will see**: user perspective — markdown tables now render as tables in artifact preview and chat markdown
> - **Surface area**: check every applicable box; for this PR include the relevant Web/UI surface
> - **Validation**: what you actually ran (`pnpm --dir apps/web exec vitest run tests/artifacts/markdown.test.ts tests/runtime/markdown.test.tsx`)
> Then update via `gh pr edit 1574 --repo nexu-io/open-design --body-file -`.

Once these parser cases are covered, this should be in a much stronger spot. 🙌

prantikmedhi:
Tightened the table parser for the two cases you called out: pipes inside inline code stay literal, and a table now starts cleanly after intro text without needing a blank line. Body is updated too.

apps/web/src/artifacts/markdown.ts
```diff
@@ -63,12 +63,96 @@ function headingLevel(line: string): number {
...
+  return 'left';
+}
+
+function isTableSeparatorLine(line: string): boolean {
+  const cells = splitTableRow(line);
+  return cells.length > 0 && cells.every((cell) => parseTableAlign(cell) !== null);
+}
+
+function isTableStart(line: string, nextLine: string | undefined): boolean {
+  return line.includes('|') && nextLine !== undefined && isTableSeparatorLine(nextLine);
```
lefarcen: P2 This now treats any pipe-containing line followed by a plain `---` line as a table, because the separator row does not need to contain a pipe. For example, `A | B` followed by a horizontal-rule-style `---` becomes a two-column table with only one alignment cell. Require the delimiter row to include a table pipe (or otherwise validate that delimiter/header cell counts match) before `isTableStart` returns true, and add a false-positive regression test.

apps/web/src/runtime/markdown.tsx
```diff
@@ -27,9 +27,64 @@ type Block =
...
+  return 'left';
+}
+
+function isTableSeparatorLine(line: string): boolean {
+  const cells = splitTableRow(line);
+  return cells.length > 0 && cells.every((cell) => parseTableAlign(cell) !== null);
+}
+
+function isTableStart(line: string, nextLine: string | undefined): boolean {
+  return line.includes('|') && nextLine !== undefined && isTableSeparatorLine(nextLine);
```
lefarcen: P2 Same false-positive in the chat renderer: `isTableStart` accepts `nextLine = '---'`, so ordinary pipe text before an hr/setext-style line gets promoted to a table. Please apply the same delimiter-row validation here too so both renderers stay consistent.

lefarcen (changes requested):
Hey @prantikmedhi, thanks for the quick follow-up — the two previous blockers are addressed in the main path now: code-span pipes stay inside the cell, and tables can start after intro text without requiring a blank line.

I found one remaining false-positive parsing case that we should tighten before this lands. It is small, but it affects both renderers because the same table-start predicate is duplicated. Once that is fixed, the parser side looks much closer.

prantikmedhi:
Good catch — I tightened the table-start check so the header and separator widths have to match now, and added a regression for the malformed separator case in both renderers. Validation is updated in the PR body too.

lefarcen (changes requested):
Hey @prantikmedhi, thanks for the quick note. I checked the current head (`844dcdd`) and I do not see the width/false-positive fix in the pushed code yet: both renderers still have `isTableStart(...)` delegating only to `isTableSeparatorLine(...)`, so `A | B` followed by `---` is still accepted as a table start.

Could you push the commit with the header/separator width check and the malformed-separator tests? Once that lands, I’ll re-review the updated head and clear my previous block if it looks good.

apps/web/src/artifacts/markdown.ts
```diff
@@ -84,7 +84,7 @@ function splitCells(line: string): string[] {
   if (line[i] === '|') i++;
   for (; i < line.length; i++) {
     const ch = line[i]!;
-    if (ch === '\\' && line[i + 1] === '|') {
+    if (ch === '\' && line[i + 1] === '|') {
```
lefarcen: P1 This literal is syntactically broken: `ch === '\'` needs to be written as `ch === '\\'` in TypeScript source. The current `ch === '''` form is what CI is reporting as a parse error, and it also prevents escaped pipes from being handled.

apps/web/src/runtime/markdown.tsx
```diff
@@ -121,7 +118,25 @@ function parseBlocks(input: string): Block[] {
       }
       // Skip the closing fence (if present).
       if (i < lines.length) i++;
-      out.push({ kind: 'code', lang, body: buf.join('\n') });
+      out.push({ kind: 'code', lang, body: buf.join('
```
lefarcen: P1 This `buf.join(...)` argument has been split into an actual multiline string literal, which is invalid TypeScript. Restore it to `buf.join('\n')`, and do the same sweep anywhere this push changed `split('\n')`, `join('\n')`, regex `\n`, or escaped-pipe test strings into literal newlines/backslashes.

lefarcen (changes requested):
Hey @prantikmedhi, thanks for pushing the follow-up. The table-start false-positive direction is moving the right way, but this head currently breaks TypeScript parsing — CI is failing before the markdown tests can run.

The main issue is that several escaped string/backslash literals were committed as literal newlines or a bare backslash quote. I left the two highest-signal spots inline; please restore those to normal escaped source literals and sweep the matching test files too.

lefarcen (approved):
Hey @prantikmedhi, nice recovery on this iteration — the escaped backslash/newline corruption is fixed, the header/separator width guard is now in both renderers, and the regression coverage is in place for the malformed separator case.

This clears my previous requested changes from the code-review side. Product review and the remaining CI/check state still need to finish before this can be considered merge-ready, but I do not see another parser blocker on this head. 🙌

lefarcen:
Hey @prantikmedhi — thank you for staying with the review and tightening this up. The final parser shape here is solid, and I appreciate the careful follow-through on the edge cases.

I’m going to close this PR now because the same GFM table-rendering fix already landed in #1496, and the linked issue #1495 has been closed from that merged implementation. Keeping this one open would leave two competing branches for the same artifact/chat markdown table support.

Really appreciate the contribution — the fixes you iterated through here were useful review signal for this area. 🙌

PR #974, opened 2026-05-08
feat(web): add Finalize design package + Continue in CLI buttons (#451)

bankielewicz:
## Summary

This is **PR 3 of #450** and the full implementation of #451 — adds two project-level toolbar buttons:

- **Finalize design package** — POSTs to `/api/projects/:id/finalize/anthropic` (route added in #832, merged 2026-05-08T11:52:12Z by @lefarcen) with a spinner-with-cancel UX during the 60–120 s blocking call. Maps `body.error.code` to user-facing toasts; surfaces `body.error.details` as a secondary toast line so account-level usage-limit responses (Anthropic 400 → daemon's `UPSTREAM_UNAVAILABLE`) show the upstream's own reason instead of just the daemon's category label (committed in [#450 verification reply](https://github.com/nexu-io/open-design/issues/450#issuecomment-4406781315)).
- **Continue in CLI** — three states (disabled / fresh / stale-with-chip) gated on `DESIGN.md` presence + freshness. On click, copies a prepared prompt to the clipboard, then opens the OS file manager at the project's working directory via Electron's `shell.openPath` (desktop) or shows a fallback toast naming the path (browser). Toast text accurately describes file-manager behavior — spawning a terminal is deferred per #451 "Non-goals."
- **⌘+Shift+K (mac) / Ctrl+Shift+K (others)** triggers Continue in CLI. Capture-phase, platform-gated, no-op when DESIGN.md is missing. Doesn't collide with `⌘+P` (Quick Switcher).

Closes #451.

## Locked design decisions (per #451 + #450 thread; ratified by @lefarcen at [#451 comment](https://github.com/nexu-io/open-design/issues/451#issuecomment-4401128215))

| ID | Decision |
|---|---|
| **A** | **Bundled** — both buttons in one PR (shared toolbar / toast / fetch infrastructure). |
| **B** | **Parse `## Provenance` section** of DESIGN.md for staleness inputs (`Generated UTC timestamp`, transcript message count, design system, current artifact). No YAML frontmatter; no daemon follow-up. |
| **C** | **Inline clipboard-prompt template** in the new web component (single source of truth; refactor deferred). |
| **D1** | **Include working directory path** in the fallback toast so users can `cd <path>` manually. |
| **D2** | **`shell.openPath(<projectDir>)`** opens the OS file manager on desktop. Toast text matches the actual behavior. Spawning a terminal application stays deferred per #451's "Non-goals." |
[...]

PR #1584, opened 2026-05-13
fix(web): parse Provenance with Markdown-bold labels

### LINKED ISSUE #1580: bug(web): parseProvenance fails on Markdown-bold labels in finalized DESIGN.md
## Summary

After `POST /api/projects/:id/finalize/anthropic` (added by #832) writes `DESIGN.md`, the Continue in CLI button (added by #974) renders **"Spec freshness unknown — regenerate to refresh signal"** and the Finalize design package button flips to **"Re-finalize (spec is stale)"** — on every successful finalize, with no underlying change. The clipboard prompt the button copies also surfaces corrupted metadata.

Root cause: `apps/web/src/lib/parse-provenance.ts:32-36` regexes fail when `DESIGN.md` uses Markdown bold field labels (`- **Project ID:** ...`), which is Claude's default emit format. The daemon's synthesis prompt at `apps/daemon/src/finalize-design.ts:560-565` lists five required Provenance fields without pinning the field-label syntax, so the parser/spec contract on label syntax is implicit and untested against the production emit.

Discovered post-merge of #974 during routine use; cross-referenced in [#451 follow-up comment](https://github.com/nexu-io/open-design/issues/451#issuecomment-4441091802).

## Primary-source evidence

Verbatim Provenance section from a fresh finalize (Pacman demo project, 2026-05-13):

```
## Provenance

- **Project ID:** `f74648b8-e444-4e06-927c-0ceb18adbb12`
- **Design system:** `default` (Neutral Modern — not applied; wireframe arcade overrides all tokens)
- **Current artifact:** `pacman.html` (single-file, 1,922 lines, 57KB)
- **Transcript message count:** 4
- **Generated UTC timestamp:** 2026-05-13T12:27:21.499Z
```

Regex trace: each parser at `apps/web/src/lib/parse-provenance.ts:32-36` uses a separator class `[:\s]+` that stops at the first non-`:`/non-whitespace char. After `:` in `**Project ID:**`, the next char is `*`, so `\1` slurps `** value` into the capture group.

Empirical `parseProvenance(...)` output against the file above:

```json
{
  "projectId": "** `f74648b8-e444-4e06-927c-0ceb18adbb12`",
  "designSystemId": "** `default` (Neutral Modern — not applied; wireframe arcade overrides all tokens)",
  "currentArtifact": "** `pacman.html` (single-file, 1,922 lines, 57KB)",
  "transcriptMessageCount": null,
  "generatedAt": null
}
```

- `transcriptMessageCount` is null because `/Transcript\s*message\s*count[:\s]+(\d+)/i` requires `\d+` immediately after `[:\s]+`, and the char after `:` is `*`, not a digit.
- `generatedAt` is null because `new Date('** 2026-05-13T12:27:21.499Z')` returns Invalid Date.

DESIGN.md spec compliance (Pacman): all 8 required headings from `finalize-design.ts:550-558` present in order, no preamble before line 1, all 5 Provenance fields listed. The file meets the spec letter; the parser/spec contract on label syntax is the gap.

## User-visible symptoms

1. **Stale chip after every successful finalize.** `apps/web/src/hooks/useDesignMdState.ts:188` returns `{ isStale: true, staleReason: 'unknown-provenance' }` when `generatedMs === null` (the round-7 fail-closed path from commit `f78ea19f`). Chip text "Spec freshness unknown — regenerate to refresh signal" (`apps/web/src/components/ContinueInCliButton.tsx:26`). Button label flips to "Re-finalize (spec is stale)" (`apps/web/src/components/FinalizeDesignButton.tsx:52`).

2. **Corrupted clipboard prompt.** `buildClipboardPrompt` consumes the parsed fields directly (`apps/web/src/lib/build-clipboard-prompt.ts:69-72`). Verbatim "Project context" section from today's Continue in CLI paste:

   ```
   - Design system: ** `default` (Neutral Modern — not applied; wireframe arcade overrides all tokens)
   - Current artifact: ** `pacman.html` (single-file, 1,922 lines, 57KB)
   - Transcript message count when DESIGN.md was generated: unknown
   - DESIGN.md generated at: unknown
   ```

   Three of four metadata lines are broken in the prompt the next `claude` CLI session reads as project context.

## Why CI didn't catch it

Test fixtures at `apps/web/tests/lib/parse-provenance.test.ts:13-17` use unbolded labels (`- Project ID: ...`), so the production emit format has no coverage. Reproduces on two independent finalize runs (different projects, different prompts); Claude's bold-Markdown emit varies between runs (label suffix wording, backticks around values) but the failure mode is consistent.

## Reproduction

1. `pnpm tools-dev`
2. Create or open a project, send a few chat turns.
3. Click "Finalize design package", wait for completion.
4. Observe "Spec freshness unknown" chip and "Re-finalize (spec is stale)" label.
5. Click "Continue in CLI", paste from clipboard into any text field, observe corrupted metadata lines at the bottom of the prompt.

## Scope

In scope:
- Harden `apps/web/src/lib/parse-provenance.ts` regexes to tolerate Markdown emphasis (`**`, `*`, `_`) around the field name and value.
- Update fixtures in `apps/web/tests/lib/parse-provenance.test.ts` (and `apps/web/tests/hooks/useDesignMdState.test.tsx` where relevant) to mirror the production emit format.
- Defense-in-depth (optional): one-line clause in `apps/daemon/src/finalize-design.ts:560-565` instructing Claude to omit Markdown emphasis on Provenance field labels.

Out of scope:
- WSL `shell.openPath` (separate issue).
- Continue in CLI toast wording / clipboard confirmation (separate issue).
- Changing Provenance section detection or field set.

bankielewicz:
Fixes #1580

## Why

I /claim'd #1580 on 2026-05-13: the parser at `apps/web/src/lib/parse-provenance.ts:32-36` was emitted in PR #974 (the Continue in CLI work) with regex separator class `[:\s]+`, which is too tight for the Markdown-bold field labels that Claude actually renders. The synthesis prompt at `apps/daemon/src/finalize-design.ts:560-565` enumerates the five Provenance fields without pinning label syntax, so Claude — following Markdown convention — emits `- **Project ID:** ...` etc. The separator `[:\s]+` stops at the first non-`:`/non-whitespace char, so the captured value starts with `** ` and `transcriptMessageCount` / `generatedAt` fail to parse downstream.

Real-world impact (verified by pasting into Notepad): the Continue in CLI clipboard prompt shows `Design system: ** ...`, `Transcript message count when DESIGN.md was generated: unknown`, `DESIGN.md generated at: unknown`. The hook's round-7 `'unknown-provenance'` chip is also tripped on every finalized DESIGN.md.

## What users will see

The Continue in CLI clipboard prompt now contains the correct Provenance values, and the staleness chip behind the button stops misreporting fresh DESIGN.md files as having unknown provenance.

## Surface area

- [x] **None** — internal bug fix to `apps/web/src/lib/parse-provenance.ts`, the synthesis prompt in `apps/daemon/src/finalize-design.ts`, and matching unit tests. No new UI, CLI, API surface, contracts, extension points, deps, or i18n keys.

## Bug fix verification

Encoded as a red spec before the source change.

- Test path: `apps/web/tests/lib/parse-provenance.test.ts` (3 new cases) + `apps/web/tests/hooks/useDesignMdState.test.tsx` (1 new case)
- Red on parent commit (`d4e44f0e`), green on this branch: yes

**Regex trace for `projectId`:**

Pre-fix regex: `/Project\s*ID[:\s]+([^\n]+)/i`. Input line: `- **Project ID:** \`<project-id>\``.

| Step | Pattern token | Consumed | Remainder |
|---|---|---|---|
| 1 | `Project\s*ID` | `Project ID` | `:** \`<project-id>\`` |
| 2 | `[:\s]+` (greedy, char class is `:` + whitespace only) | `:` | `** \`<project-id>\`` |
| 3 | `([^\n]+)` | `** \`<project-id>\`` | (end of line) |

Pre-fix `projectId` capture value: `\`** \`<project-id>\`\``. Post-fix: helper strips leading `[\s*_]+` → `\`<project-id>\``.

For `transcriptMessageCount`, the pre-fix capture group `(\d+)` required digits immediately after `[:\s]+`, so the regex failed to match entirely on `** 4`. Widening to `([^\n]+)` lets the helper strip step run before `Number.parseInt`.

**Before / after `parseProvenance(...)` against a live finalized DESIGN.md:**

Before (pre-fix `main` @ `d4e44f0e`):

\`\`\`json
{
  "projectId":              "** \`<project-id>\`",
  "designSystemId":         "** \`default\` (...)",
  "currentArtifact":        "** \`<artifact>.html\` (...)",
  "transcriptMessageCount": null,
  "generatedAt":            null
}
\`\`\`

After (this branch):

\`\`\`json
{
  "projectId":              "\`<project-id>\`",
  "designSystemId":         "\`default\` (...)",
  "currentArtifact":        "\`<artifact>.html\` (...)",
  "transcriptMessageCount": 4,
  "generatedAt":            "2026-05-13T12:27:21.499Z"
}
\`\`\`

Backticks remain in the captured value — the issue spec explicitly allows it, and the clipboard prompt reads fine with them. The variant `- **Field**: value` (colon **outside** the emphasis) is not in #1580's enumeration and is intentionally not handled.

**Phase 4 disclosure (daemon synthesis prompt):** I included a one-line addendum to the synthesis prompt at `apps/daemon/src/finalize-design.ts:566` telling Claude to render Provenance bullets without emphasis on the labels — defense-in-depth against future model variants. The parser hardening is the load-bearing fix; the prompt change is not required to close #1580. Daemon restart is needed for the prompt change to take effect locally (no `--watch` on the daemon process).

## Validation

- \`pnpm guard\` — passed
- \`pnpm typecheck\` — passed
- \`pnpm --filter @open-design/web test\` — 1158 / 1158 passed (was 1154 pre-PR; +3 parser tests, +1 hook test)
- Empirical: ran \`parseProvenance(...)\` against a live finalized DESIGN.md from a local project — all five fields returned correctly (see before/after above).

### DIFF
--- a/apps/web/src/lib/parse-provenance.ts
+++ b/apps/web/src/lib/parse-provenance.ts
@@ -41,10 +41,24 @@ export function parseProvenance(designMdText: string): ProvenanceFields | null {
 // (`- **Field:** value`) per Markdown convention. The capture starts
 // just after the label's `:`, so a leading `** ` (and any trailing
 // emphasis if the value itself is wrapped) leaks into the value.
-// Strip both ends; backticks are intentionally kept (the rendered
-// clipboard text reads fine with them).
+//
+// PR #1584 review (lefarcen): narrow the strip to only consume
+// Markdown residue, never literal `*`/`_` characters in the value:
+//   1. Leading `*`/`_` tokens FOLLOWED BY WHITESPACE
+//      (the `** ` left over from `- **Field:** value`).
+//   2. Trailing WHITESPACE followed by `*`/`_` tokens
+//      (mirror of step 1 if Claude closes after the value).
+//   3. A single balanced wrap around the whole remaining value
+//      (`**X**` / `*X*` / `__X__` / `_X_`).
+// Asymmetric literal `*`/`_` without a whitespace separator AND
+// without a balanced closing token are preserved
+// (e.g. `_draft.html`, `build_id_v1_`). Backticks are intentionally
+// kept (the rendered clipboard text reads fine with them).
 function stripMarkdownEmphasis(value: string): string {
-  return value.replace(/^[\s*_]+/, '').replace(/[\s*_]+$/, '');
+  let v = value.replace(/^[*_]+\s+/, '').replace(/\s+[*_]+$/, '');
+  const wrap = v.match(/^(\*\*|__|\*|_)(.+?)\1$/);
+  if (wrap && wrap[2]) v = wrap[2];
+  return v;
 }
 
 function extractRawValue(body: string, re: RegExp): string | null {
--- a/apps/web/src/lib/parse-provenance.ts
+++ b/apps/web/src/lib/parse-provenance.ts
@@ -32,43 +32,49 @@ export function parseProvenance(designMdText: string): ProvenanceFields | null {
     projectId: extractField(body, /Project\s*ID[:\s]+([^\n]+)/i),
     designSystemId: extractFieldOrNone(body, /Design\s*system[^:]*[:\s]+([^\n]+)/i),
     currentArtifact: extractFieldOrNone(body, /Current\s*artifact[^:]*[:\s]+([^\n]+)/i),
-    transcriptMessageCount: extractNumber(body, /Transcript\s*message\s*count[:\s]+(\d+)/i),
-    generatedAt: extractDate(body, /Generated[^:\n]*[:\s]+(\S[^\n]*)/i),
+    transcriptMessageCount: extractNumber(body, /Transcript\s*message\s*count[^:]*[:\s]+([^\n]+)/i),
+    generatedAt: extractDate(body, /Generated[^:\n]*[:\s]+([^\n]+)/i),
   };
 }
 
-function trimBullet(value: string): string {
-  // Lines look like "- Project ID: abc". The regex captures everything
-  // after the colon to the newline; strip incidental trailing whitespace.
-  return value.trim();
+// #1580: Claude renders Provenance fields with Markdown-bold labels
+// (`- **Field:** value`) per Markdown convention. The capture starts
+// just after the label's `:`, so a leading `** ` (and any trailing
+// emphasis if the value itself is wrapped) leaks into the value.
+// Strip both ends; backticks are intentionally kept (the rendered
+// clipboard text reads fine with them).
+function stripMarkdownEmphasis(value: string): string {
+  return value.replace(/^[\s*_]+/, '').replace(/[\s*_]+$/, '');
 }
 
-function extractField(body: string, re: RegExp): string | null {
+function extractRawValue(body: string, re: RegExp): string | null {
   const m = body.match(re);
   if (!m || !m[1]) return null;
-  const value = trimBullet(m[1]);
+  const value = stripMarkdownEmphasis(m[1].trim());
   return value.length > 0 ? value : null;
 }
 
+function extractField(body: string, re: RegExp): string | null {
+  return extractRawValue(body, re);
+}
+
 function extractFieldOrNone(body: string, re: RegExp): string | null {
-  const value = extractField(body, re);
+  const value = extractRawValue(body, re);
   if (value === null) return null;
   if (NONE_SENTINEL.test(value)) return null;
   return value;
 }
 
 function extractNumber(body: string, re: RegExp): number | null {
-  const m = body.match(re);
-  if (!m || !m[1]) return null;
-  const n = Number.parseInt(m[1], 10);
+  const raw = extractRawValue(body, re);
+  if (raw === null) return null;
+  const n = Number.parseInt(raw, 10);
   return Number.isFinite(n) ? n : null;
 }
 
 function extractDate(body: string, re: RegExp): Date | null {
-  const m = body.match(re);
-  if (!m || !m[1]) return null;
-  const raw = trimBullet(m[1]);
-  if (raw.length === 0) return null;
+  const raw = extractRawValue(body, re);
+  if (raw === null) return null;
   const d = new Date(raw);
   return Number.isFinite(d.getTime()) ? d : null;
 }
--- a/apps/daemon/src/finalize-design.ts
+++ b/apps/daemon/src/finalize-design.ts
@@ -564,6 +564,8 @@ The Provenance section MUST list:
 - Transcript message count
 - Generated UTC timestamp
 
+Render Provenance fields as plain Markdown bullets with no emphasis on the field labels, exactly: "- Field name: value". Do not bold, italicize, or otherwise decorate the labels or the colon. Field values may use inline code formatting (backticks) where appropriate.
+
 Output the Markdown body only. No preamble, no chat-style framing, no
 "Here's your DESIGN.md" prefix. Do not invent facts not supported by the
 inputs; if an input is missing or empty, the corresponding section should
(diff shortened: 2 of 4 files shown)


apps/web/src/lib/parse-provenance.ts
```diff
@@ -32,43 +32,49 @@ export function parseProvenance(designMdText: string): ProvenanceFields | null {
...
-  // after the colon to the newline; strip incidental trailing whitespace.
-  return value.trim();
+// #1580: Claude renders Provenance fields with Markdown-bold labels
+// (`- **Field:** value`) per Markdown convention. The capture starts
+// just after the label's `:`, so a leading `** ` (and any trailing
+// emphasis if the value itself is wrapped) leaks into the value.
+// Strip both ends; backticks are intentionally kept (the rendered
+// clipboard text reads fine with them).
+function stripMarkdownEmphasis(value: string): string {
+  return value.replace(/^[\s*_]+/, '').replace(/[\s*_]+$/, '');
```
lefarcen: P1 This now strips every leading/trailing `*` and `_` from all Provenance values, including the existing plain-label path. That can silently corrupt legitimate values such as `- Current artifact: _draft.html` or an id/name with boundary underscores, even when there is no Markdown emphasis to clean up. Please narrow this to the exact Markdown residue we need to remove (for example the leading `** ` left by `- **Field:** value`, and balanced emphasis wrappers) while preserving literal underscores/asterisks in the actual value, and add a regression test for a value like `_draft.html`.

lefarcen (changes requested):
Hey @bankielewicz — this is a well-scoped fix, and the red-spec coverage around the real bold-label Provenance shape is exactly the right direction. I found one regression in the shared value cleanup helper that we should tighten before this lands; details inline.

apps/web/src/lib/parse-provenance.ts
bankielewicz: Narrowed in cb2e2eeb to the residue-only strip we discussed:

1. Leading `*`/`_` tokens **followed by whitespace** — only fires on the `** ` residue from `- **Field:** value`.
2. Trailing **whitespace followed by** `*`/`_` tokens — mirror of (1).
3. A single balanced wrap around the remaining value (`**X**` / `*X*` / `__X__` / `_X_`).

Asymmetric literal `*`/`_` (no whitespace separator, no closing token) is preserved by construction. Added 4 regression tests:
- plain label + `_draft.html` ← your explicit ask
- plain label + `build_id_v1_` (trailing underscore on id)
- bold label + `_draft.html` (residue stripped, literal leading `_` kept)
- plain label + `**wrapped-id**` (balanced wrap stripped)

Three of those four were red against round-1; all 11 parser tests + 1162 web tests are green now. Empirically re-checked against the live finalized DESIGN.md from #1580 — all five fields still parse.

lefarcen (approved):
Hey @bankielewicz — thanks for the quick, precise follow-up. The residue-only cleanup now preserves literal boundary underscores/asterisks while still handling the `** ` leak from bold Provenance labels, and the added regression cases cover the exact failure I called out. This clears my previous changes-requested block from the bot side.

Siri-Ray (approved):
@bankielewicz I reviewed the changed ranges in the Provenance parser, the daemon prompt addendum, and the new parser/hook regression coverage. The follow-up patch addresses the prior underscore-corruption concern by narrowing the cleanup to Markdown residue while preserving the explicit `_draft.html` and `build_id_v1_` cases, and I did not find any remaining actionable correctness issues. I could not rerun the targeted web tests in this worktree because `node_modules` is missing and `vitest` is unavailable locally, but the added fixture coverage lines up with the bug and the reviewed implementation. Thanks for the careful red-spec writeup and the quick fix on the review feedback. 🙂

PR #1585, opened 2026-05-13
fix(daemon): handshake URL param overrides stale __session cookie (Bug 8 + carryforward Bug 7 + Bug 6)

austinmao:
## v7 demo fix + Playwright prod E2E specs

Tracks four daemon patches + production E2E gate against Hetzner prod (`*.opendesign.holalumina.com`).

### Daemon patches (v7)

1. **Bug 9** — expired handshake-URL JWT returns 302 to re-handshake (not 401 dead-end)
2. **Bug 10** — `__session` cookie shadow: iterate cookie list, skip empty values
3. **Handshake URL override** — `__od_handshake` param wins over stale `__session`
4. **Lumina-managed sentinel** — `apiKey=lumina-managed` + `baseUrl=https://lumina-gateway-managed` swap to direct Anthropic with daemon `ANTHROPIC_API_KEY`, preserving `<artifact>` system prompt (bypasses openclaw gateway plugin override)
5. **Vercel env-var fallback** — `readVercelConfig` falls back to `VERCEL_API_TOKEN` / `VERCEL_TEAM_ID` / `VERCEL_TEAM_SLUG` env when `deploy.json` absent
6. **Dockerfile v7** — bundles `apps/web/out` static export into runtime image

### Unit tests (vitest)

- `apps/daemon/tests/tenants/resolver.test.ts` — 24 cases incl. explicit Bug 9 / Bug 10 / handshake URL override regressions
- `apps/daemon/tests/tenants/cross-tenant-isolation.test.ts` — 17 byte-identical 404 vectors
- `apps/daemon/tests/lumina-gateway-override.test.ts` — 10 sentinel-swap tests
- `apps/daemon/tests/deploy.test.ts` — Vercel env fallback coverage
- `apps/daemon/tests/auth/clerk-jwt.test.ts` — 6 JWT verify cases
- 356 pass / 0 regressions on v7 build (1 pre-existing X-Forwarded-Host fail unrelated to these patches)

### Playwright prod E2E (new in this PR)

5 specs against prod ceremonia subdomain, baseline 12 pass / 5 documented skips / 0 fails in 24s:

| Spec | Result | Notes |
|------|--------|-------|
| `handshake.spec.ts` | 3 pass + 1 skip | Real email+password sign-in via Clerk, AgentMail OTP poll for prod 2FA; Bug 9 skipped (covered by resolver.test.ts cases 14-15) |
| `cookie-shadow.spec.ts` | 3 pass + 1 skip | HTTP-boundary Bug 10 probes; browser-level dual-cookie requires CDP host-only flag (Playwright API gap) |
| `lumina-swap.spec.ts` | 3 pass | mode=api + lumina-managed sentinel → 200 from `/api/proxy/stream` → SSE event → `<artifact>` → FileViewer renders |
| `vercel-deploy.spec.ts` | 1 pass + 1 skip | Config-probe asserts daemon `configured:true` + `tokenMask` non-empty; full deploy integration skipped on 401 (storage state refresh edge case) |
| `jwt-expiry-survival.spec.ts` | 1 pass + 2 skip | /api/* 401 case verified; backdate-JWT cases skipped (kind=invalid_signature not expired — `wait` strategy is canonical, see resolver.test.ts cases 14-15) |

### Auth hard rules (enforced in specs)

- Email + password only (NEVER Google SSO; Clerk's `<button>` SSO row breaks storage-state)
- Prod Clerk requires email-OTP at factor-two; specs poll AgentMail inbox + extract code from subject line
- Test user provisioned via Clerk Backend API (prod), ceremonia org membership, `publicMetadata.last_active_organization_id` pinned per skill doc

### Required env (gitignored `e2e/.env.prod.local`)

```
OD_E2E_TENANT={ceremonia|lumina|ericedmeades}
OD_E2E_PROD_EMAIL=...@agentmail.to
OD_E2E_PROD_PASSWORD=...
OD_E2E_AGENTMAIL_INBOX=...@agentmail.to
OD_E2E_AGENTMAIL_API_KEY=...
```

### Run

```
cd e2e
pnpm exec playwright test --config=playwright.prod.config.ts
```

### Rebase note

Branched off `d25a7aaf` (upstream main 2026-04-30). Upstream `open-design-v0.5.0` rewrote `apps/daemon/src/server.ts` (4905-line diff) and `deploy.ts` (394-line diff) since then. A clean rebase requires manual port of the lumina-managed swap + LUMINA_GATEWAY override into the new server.ts structure. **Rebase deferred post-demo (2026-05-15)** to avoid demo-day breakage. Port is scoped + tracked separately; specs in this PR will gate the rebased branch when ready.

### Rollback

Hetzner daemon container `open-design-daemon-v5-backup-164842` parked; rollback = single `docker rename` (~30s).

### DIFF
--- a/apps/daemon/src/tenants/resolver.ts
+++ b/apps/daemon/src/tenants/resolver.ts
@@ -201,6 +201,37 @@ export function tenantResolverMiddleware(deps: TenantResolverDeps): ExpressLikeM
         } catch (err) {
           const kind =
             err instanceof ClerkVerificationError ? err.kind : 'invalid';
+          // Bug 9: when the handshake URL JWT is itself expired (Clerk
+          // JWT TTL = 300s; network/processing latency + user idle can
+          // exceed that between mint at primary and consume at satellite),
+          // bounce back to the handshake endpoint to mint a fresh JWT
+          // instead of 401. Without this, the cascade Bug 6+7+8 is
+          // incomplete: stale-cookie → fresh handshake URL → expired
+          // before consume → 401 → infinite reload. Only `expired`
+          // triggers 302; other kinds (invalid_signature, malformed)
+          // remain 401 since those indicate tampering.
+          if (kind === 'expired') {
+            logResolution({
+              requestId,
+              host: rawHost,
+              subdomain,
+              result: 'redirect',
+              stepFailed: 'jwt_invalid',
+              statusCode: 302,
+              extra: {
+                jwt_failure_kind: kind,
+                source: 'handshake',
+                action: 're_handshake',
+              },
+            });
+            const cleanUrl = stripHandshakeParam(req.url ?? '/');
+            const targetUrl = `https://${rawHost}${cleanUrl}`;
+            respondRedirect(
+              res,
+              `${HANDSHAKE_URL}?target_url=${encodeURIComponent(targetUrl)}`,
+            );
+            return;
+          }
           logResolution({
             requestId,
             host: rawHost,
--- a/apps/daemon/src/tenants/resolver.ts
+++ b/apps/daemon/src/tenants/resolver.ts
@@ -61,6 +61,13 @@ import {
 
 const PLATFORM_DOMAIN_SUFFIX = '.opendesign.holalumina.com';
 const SIGN_IN_URL = 'https://app.holalumina.com/sign-in';
+// Cross-domain handshake endpoint on the primary Clerk-auth host. The primary
+// mints a short-lived JWT bound to the user's active session and 302s back to
+// the satellite subdomain with `?__od_handshake=<jwt>`. The satellite (this
+// middleware) consumes the token, sets the `__session` cookie on its own
+// subdomain, and 302s to the clean URL. See web/src/app/api/od-handshake/.
+const HANDSHAKE_URL = 'https://app.holalumina.com/api/od-handshake';
+const HANDSHAKE_PARAM = '__od_handshake';
 const SUBDOMAIN_PATTERN = /^[a-z0-9-]+$/;
 const NOT_FOUND_BODY = 'Not Found\n';
 const NOT_FOUND_BYTES = Buffer.byteLength(NOT_FOUND_BODY, 'utf8');
@@ -173,11 +180,89 @@ export function tenantResolverMiddleware(deps: TenantResolverDeps): ExpressLikeM
     // 6. Session cookie check.
     const sessionToken = readSessionCookie(req);
     if (!sessionToken) {
+      // 6a. Cross-domain handshake path. The primary auth host (app.holalumina.com)
+      // 302s here with `?__od_handshake=<jwt>` after the user has signed in.
+      // The JWT is identical in shape to what Clerk would have set as
+      // __session, so we run it through the same verifier. On success we set
+      // __session as a Set-Cookie scoped to THIS subdomain (not the apex —
+      // that would leak the cookie across tenants) and 302 to the same URL
+      // with the handshake param stripped so the browser re-enters this
+      // middleware on the next hop with a real cookie.
+      const handshakeToken = readHandshakeToken(req);
+      if (handshakeToken) {
+        void (async () => {
+          let claims;
+          try {
+            claims = await verifyClerkSession(handshakeToken);
+          } catch (err) {
+            const kind =
+              err instanceof ClerkVerificationError ? err.kind : 'invalid';
+            logResolution({
+              requestId,
+              host: rawHost,
+              subdomain,
+              result: 'denied',
+              stepFailed: 'jwt_invalid',
+              statusCode: 401,
+              extra: { jwt_failure_kind: kind, source: 'handshake' },
+            });
+            respondUnauthorized(res);
+            return;
+          }
+          if (claims.o.slg !== subdomain) {
+            logResolution({
+              requestId,
+              host: rawHost,
+              subdomain,
+              result: 'denied',
+              stepFailed: 'org_mismatch',
+              statusCode: 404,
+              tenantResolved: claims.o.slg,
+              userId: claims.sub,
+              auditSeverity: 'high',
+              extra: { source: 'handshake' },
+            });
+            respondNotFound(res);
+            return;
+          }
+          const cleanUrl = stripHandshakeParam(req.url ?? '/');
+          const cookieDomain = `${subdomain}${PLATFORM_DOMAIN_SUFFIX}`;
+          // Cookie TTL matches Clerk default JWT TTL (300s). Cookie outliving
+          // JWT means browser keeps sending stale token, gets 401, never
+          // auto-recovers. Align them so cookie expiry triggers re-handshake
+          // via the no-cookie path at 6b.
+          const cookie =
+            `__session=${handshakeToken}` +
+            `; Domain=${cookieDomain}` +
+            `; Path=/` +
+            `; Max-Age=300` +
+            `; Secure` +
+            `; HttpOnly` +
+            `; SameSite=Lax`;
+          res.setHeader('Set-Cookie', cookie);
+          logResolution({
+            requestId,
+            host: rawHost,
+            subdomain,
+            result: 'redirect',
+            stepFailed: null,
+            statusCode: 302,
+            tenantResolved: claims.o.slg,
+            userId: claims.sub,
+            extra: { source: 'handshake', action: 'cookie_set' },
+          });
+          respondRedirect(res, `https://${cookieDomain}${cleanUrl}`);
+        })();
+        return;
+      }
+
+      // 6b. No cookie, no handshake → bounce to the primary handshake
+      // endpoint. That endpoint reads the user's app.holalumina.com Clerk
+      // session, mints a fresh JWT, and 302s back here with the handshake
+      // param set. If the user is not signed in to the primary, the
+      // handshake endpoint itself 302s to /sign-in first.
       const returnUrl = `https://${subdomain}${PLATFORM_DOMAIN_SUFFIX}${req.url ?? '/'}`;
-      // Use redirect_url (Clerk + app convention) so the sign-in page picks
-      // it up via useSearchParams().get('redirect_url'). Older code shipped
-      // ?return= which the app ignored, dropping the user on /cms.
-      const location = `${SIGN_IN_URL}?redirect_url=${encodeURIComponent(returnUrl)}`;
+      const location = `${HANDSHAKE_URL}?target_url=${encodeURIComponent(returnUrl)}`;
       logResolution({
         requestId,
         host: rawHost,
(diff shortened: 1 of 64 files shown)


apps/daemon/src/tenants/resolver.ts
```diff
@@ -0,0 +1,597 @@
...
+        // handshake re-mint round-trip is fast (~150ms) and silent to the
+        // user.
+        //
+        // EXCEPTION: /api/* requests cannot follow cross-origin 302s while
+        // carrying cookies (browser fetch CORS). Return 401 instead so the
+        // JS layer can surface a refresh prompt rather than silently failing
+        // with "TypeError: Failed to fetch".
+        if (kind === 'expired') {
+          const reqPath = req.url ?? '/';
+          if (reqPath.startsWith('/api/')) {
```
lefarcen, line 365: P1 This keeps the Bug 6 failure mode for active in-app users: after the JWT expires, any existing `/api/*` fetch still gets a bare 401 and the browser never receives a fresh `__session` cookie. The PR text says idle sessions are silently re-minted, but there is no web-side 401 recovery path in this diff, so chat/file/API calls can still fail after the TTL. Please either implement a client recovery flow that navigates through the handshake and retries, or return a same-origin response the web app can use to refresh the cookie before retrying API calls.

apps/web/src/state/config.ts
```diff
@@ -2,15 +2,23 @@ import type { AppConfig, MediaProviderCredentials } from '../types';
...
+// `daemon` mode (CLI subprocess) is unavailable inside the container —
+// no CLI binaries shipped — so leaving that as default surfaces a silent
+// no-op when the user hits Send.
 export const DEFAULT_CONFIG: AppConfig = {
-  mode: 'daemon',
-  apiKey: '',
-  baseUrl: 'https://api.anthropic.com',
-  model: 'claude-sonnet-4-5',
+  mode: 'api',
+  apiKey: 'lumina-managed',
```
lefarcen, line 15: P1 These Lumina-managed defaults are unconditional, so a normal open-design web/desktop build now starts in `api` mode with a fake key/base URL and `onboardingCompleted: true`. That bypasses the BYOK/agent onboarding for upstream users and routes their first send through `/api/proxy/stream` with non-functional credentials. Please gate these defaults behind the Lumina/proxy build flag and keep the existing local defaults for regular builds.

infra/docker/Dockerfile.daemon
```diff
@@ -0,0 +1,104 @@
...
+COPY packages ./packages
+COPY tools ./tools
+# Root postinstall script referenced by package.json.
+COPY scripts ./scripts
+
+# Now copy daemon sources.
+COPY apps/daemon ./apps/daemon
+
+# Install the entire workspace (frozen lockfile) and compile the daemon.
+RUN pnpm install --frozen-lockfile \
```
lefarcen, line 54: P1 The image only builds/deploys `@open-design/daemon`, but `server.ts` serves the web shell from `apps/web/out` plus repo-level resources like `skills/` and `design-systems/`. `pnpm deploy` for the daemon package copies only the daemon package files, so the runtime image will not contain the static web app or those resource directories; authenticated tenant page loads will fall through instead of serving the UI. Please build/copy the web export and required resource folders into the image, or set an explicit resource/static root that matches what the runtime actually contains.

lefarcen (changes requested):
Hey @austinmao — thanks for chasing the handshake loop all the way through. The fresh URL token winning over the stale parent-domain `__session` cookie is the right direction, and the regression coverage around that path is useful.

I found a few blockers that need tightening before this can land, mostly around preserving the idle-session recovery behavior in-app and making sure the daemon container actually serves the surfaces this change relies on. I also noticed the PR description is missing the newer template sections: Why, What users will see, Surface area, and Validation. Since you already have good verification notes, this should mostly be reshaping them so reviewers can scope the user impact quickly.

If you're using an AI assistant, you can paste this to help update the PR body:

> Read the current PR body for nexu-io/open-design#1585 and the template at `.github/pull_request_template.md`. Rewrite the description filling in: Why; What users will see; Surface area; and Validation. Keep the existing handshake-loop, TTL, and live-probe details, but make the user-visible behavior and validation commands explicit. Then update via `gh pr edit 1585 --repo nexu-io/open-design --body-file -`.

Once the blockers below are addressed, this should be much easier to re-review.

nettee:
@austinmao I'm holding off on generating review comments for nexu-io/open-design#1585 because this pull request has merge conflicts right now.

lefarcen (changes requested):
Hey @austinmao — I re-checked the current head after @nettee's conflict notice. The branch is still on the same commit (`091ebf8`), and the three blockers from my earlier review are still present, so my changes-requested review remains in effect.

Please resolve the merge conflicts with `main`, then address the existing inline blockers around `/api/*` session refresh, Lumina-only defaults, and the daemon container packaging. Once you push an updated head, we can do a clean pass on the refreshed diff.

nettee:
@austinmao I'm holding off on generating review comments for nexu-io/open-design#1585 because this pull request has merge conflicts right now.

apps/daemon/src/tenants/resolver.ts
```diff
@@ -0,0 +1,644 @@
...
```
lefarcen, line 365: P1 This still leaves the main in-app idle path without recovery: when a user has an expired JWT and makes any `/api/*` request, the daemon returns a bare 401 and never refreshes the cookie. The PR body says the 6-minute idle flow survives, but the web code still has no 401-to-handshake-and-retry path, so chat/file/project API calls can fail after the JWT TTL. Please add a same-origin recovery mechanism the web client actually uses, or implement the client-side 401 handler that navigates through the handshake and retries the original API action.

apps/web/src/state/config.ts
lefarcen, line 15: P1 These Lumina-managed defaults are still unconditional. In a normal open-design local/desktop build, a fresh user now skips onboarding and sends the first request with `apiKey='lumina-managed'`, which the daemon maps to `ANTHROPIC_API_KEY`; without that Lumina env, the default experience becomes a 502 instead of the existing BYOK/daemon setup flow. Please gate the sentinel defaults behind the Lumina/proxy build mode and preserve the previous defaults for regular builds.

infra/docker/Dockerfile.daemon
```diff
@@ -0,0 +1,117 @@
...
+RUN mkdir -p /data /etc/opendesign \
+    && chown -R node:node /app /data
+
+# Copy the pruned deploy directory from the builder.
+COPY --from=builder --chown=node:node /deploy ./
+
+# v7: bundle the static-exported web frontend so the daemon SPA fallback
+# (server.ts STATIC_DIR = PROJECT_ROOT + apps/web/out) finds index.html.
+# Without this, every non-/api/* request returns 404 "Cannot GET /".
+COPY --from=builder --chown=node:node /workspace/apps/web/out /apps/web/out
```
lefarcen, line 84: P2 Copying `apps/web/out` fixes the shell, but the runtime image still does not include the repo-level resources that `server.ts` resolves from `PROJECT_ROOT`: `/skills`, `/design-systems`, `/assets/frames`, and `/prompt-templates`. In the deployed daemon layout `PROJECT_ROOT` resolves to `/`, so these API surfaces will be empty/missing even though the UI loads. Please either copy those directories into the image at the paths the daemon uses or set `OD_RESOURCE_ROOT` to a copied resource bundle containing them.

lefarcen (changes requested):
Hey @austinmao — thanks for the quick follow-up. The v7 push does move things forward: the web static export is now built into the image, the expired-handshake-token case is covered, and the empty-cookie shadow regression is a good catch.

I re-reviewed the current head (`c5a3671`). A few blockers still need attention before this can land safely: the in-app `/api/*` idle recovery path is still a bare 401, the Lumina-managed defaults still leak into normal builds, and the container still omits the repo-level resources that the daemon resolves at runtime. I’ve kept the review to only the remaining/current-head issues below.

nettee:
@austinmao I'm holding off on generating review comments for nexu-io/open-design#1585 because this pull request has merge conflicts right now.

apps/daemon/src/tenants/resolver.ts
lefarcen, line 365: P1 This branch still returns a bare 401 for expired JWTs on `/api/*`, so the in-app idle path has no actual cookie refresh or retry. Adding E2E coverage does not change the server/client behavior here: chat/file/project API calls after JWT expiry still depend on a web-side recovery path that is not present in this diff. Please implement a same-origin recovery signal plus client retry/navigation, or otherwise make `/api/*` requests refresh the session before failing the user action.

apps/web/src/state/config.ts
lefarcen, line 15: P1 The Lumina-managed sentinel defaults are still unconditional, so regular open-design builds skip onboarding and send fresh users through the daemon's `ANTHROPIC_API_KEY` sentinel path. In non-Lumina local/desktop installs that env is not expected, making the default first run fail instead of offering the existing BYOK/daemon setup. Please gate these defaults on the Lumina/proxy build mode and preserve the previous defaults otherwise.

infra/docker/Dockerfile.daemon
lefarcen, line 84: P2 The image still only copies `apps/web/out`; it does not include `/skills`, `/design-systems`, `/assets/frames`, or `/prompt-templates`, while `server.ts` resolves those from `PROJECT_ROOT` (which is `/` in this deployed layout). The UI shell will load, but the daemon-backed pickers/resources are empty or missing. Please copy those resource directories into the runtime image or set `OD_RESOURCE_ROOT` to a bundled resource root that contains them.

lefarcen (changes requested):
Hey @austinmao — I re-reviewed the new head (`d1ce714`). The added prod E2E coverage is helpful, but the code paths behind the previous blockers are still unchanged on this head, so my changes-requested review still needs to stand.

The remaining items are the same three current-head blockers: `/api/*` idle recovery still dead-ends at 401, the Lumina sentinel defaults still apply to normal builds, and the daemon image still only bundles the web shell without the resource directories the daemon resolves at runtime.

austinmao:
Superseded by fork PR https://github.com/austinmao/open-design/pull/2 — v0.7.0 port includes the same handshake URL param fix (commit `c4ddd545`) plus 25 other lumina deltas re-applied on top of upstream stable v0.7.0. Will close after #2 merges.

lefarcen:
Hey @austinmao — noted, and the v0.7.0 rebase approach makes sense. Porting the lumina deltas onto upstream stable is cleaner than trying to patch the conflicts here given how much `server.ts` diverged.

Once fork PR #2 is settled and you're ready to open the upstream PR, feel free to close this one. We'll do a fresh review pass on the rebased branch when it comes in. ✨

austinmao:
Closing as superseded by #2 (now merged). Same handshake URL param fix lives at commit c4ddd545 inside the v0.7.0 port.

Issue #569, opened 2026-05-05
google-slides daemon: silent private-Drive fallback breaks image insertion under Workspace DLP


Rekez886:
## Symptom

When a Google Workspace user runs the ja-slide skill (or any deck-generation flow that writes outline-referenced images via the Drive→Slides path), and the workspace has a DLP / sharing policy that blocks `drive.permissions.create`, image insertion silently fails on the Slides side.

## Root cause

`apps/daemon/src/google-slides.ts` lines 176–180 — when the `drive.permissions.create` call throws (the documented Workspace DLP scenario), the catch block falls back to using `slidesAccessibleUrl`, which points at a private Drive file.

The Slides API's `createImage` / `replaceImage` then fetches `image.url` as a public URL resource. **It does not reuse the daemon's OAuth token** to read private Drive content. So:

- Public Drive file → works
- Workspace-blocked file with the silent fallback → image fetch fails server-side, slide ends up with a broken / template-default image, no error surfaced to the agent

This regresses for exactly the corp-Workspace use case the catch was meant to support.

## Discovered by

Codex review on #568 (the ja-slide skill iteration PR). Reproduces in the ja-slide deck-generation flow when the Drive permissions step is blocked.

## Suggested fix

In the catch block, do not fall back to `slidesAccessibleUrl`. Either:

1. **Surface the failure** — let the error propagate so the agent can write it into `userActions` (so the user sees "image upload was DLP-blocked, replace manually") instead of silently producing a broken slide
2. **Use a genuinely accessible URL** — e.g., a server-side proxy endpoint on the daemon that streams the bytes with the daemon's OAuth token, then hand Slides a public daemon URL
3. **Pre-flight check** — try `permissions.create` before `Slides.createImage` and short-circuit the whole path with a clear error if it fails

Option 1 is the smallest change and matches the rest of the skill's "fail loudly, surface in userActions" philosophy.

## Out of scope for #568

#568 is a ja-slide skill iteration + slide preview UI PR — it doesn't touch `google-slides.ts`. Filing this as a follow-up so it isn't lost.

lefarcen:
Hi @Rekez886! 🎉

Thank you for catching this — it's a real issue with the current fallback behavior in PR #568.

[...]

PR #1606, opened 2026-05-13
feat(daemon): implement fal.ai renderer for image + video generation

sdanpo:
## Why

Open Design already had fal.ai listed as a media provider and had model catalog entries (`sd-3.5`, `sora-2`, `sora-2-pro`) pointing to the fal provider, but there was no renderer wired up. Any attempt to generate using a fal model fell through to the stub path and returned a 503 error with "provider not configured". This PR wires up a real renderer so those models actually work.

The fal provider is also uniquely valuable because it hosts a wide variety of third-party models (FLUX, Sora, Veo, Wan, Ideogram, Recraft, Kling, Seedance, etc.) under a single API key, making it a high-leverage integration.

## What users will see

- Generating with any fal model (e.g. `flux-pro-ultra`, `sd-3.5`, `sora-2`, `wan-2.1-t2v`) now produces a real image or video instead of a 503 error.
- The fal provider is marked `integrated: true` in the Settings dialog.
- New catalog models appear in the picker: `flux-pro-ultra`, `flux-dev (fal)`, `flux-schnell (fal)`, `ideogram-v3`, `recraft-v3` (images); `veo-3 (fal)`, `veo-2 (fal)`, `wan-2.1-t2v`, `wan-2.1-i2v`, `seedance-1-pro (fal)`, `kling-2.1 (fal)` (video).
- Any `fal-ai/*` path can be passed as `--model` to reach models not in the catalog — no code change required for new fal models.
- When a requested video duration is outside Veo's accepted buckets (4s/6s/8s), it is snapped to the nearest bucket and the providerNote reports both the requested and effective duration.

## What changed

**`apps/daemon/src/media.ts`**
- `falQueueRun` — shared async helper: submits to `queue.fal.run/{endpoint}`, polls `status_url` until `COMPLETED`, fetches `response_url`. Uses `OD_FAL_MAX_POLL_MS` to configure the ceiling (default 5 min image / 10 min video).
- `renderFalImage` — resolves per-endpoint image schema (`aspect_ratio` for flux-pro-ultra vs named `image_size` for dev/schnell/SD), downloads result, sniffs extension from magic bytes.
- `renderFalVideo` — resolves per-endpoint video schema: omits `duration` for Wan models (which reject it), clamps to nearest valid bucket and formats as `"Ns"` string for Veo (4s/6s/8s only), bare integer for others. Tracks effective duration label for the providerNote; emits a snap note when the requested length is adjusted.
- `FAL_ENDPOINTS` map: catalog model IDs → verified fal-ai/* paths. Unlisted models fall through to use their ID as the path directly.
- `FAL_IMAGE_USES_ASPECT_RATIO`, `FAL_VIDEO_NO_DURATION`, `FAL_VIDEO_STRING_DURATION`, `FAL_VEO_DURATION_BUCKETS`: capability sets that drive per-endpoint request body differences.
- `falQueueBase()`: anchored regex replace so custom base URLs aren't mangled.
- `generateMedia`: accepts any `fal-ai/*` model path without a catalog entry (synthetic def, surface from caller).
- Dispatcher: adds `provider === 'fal' && surface === 'image'` and `provider === 'fal' && surface === 'video'` branches.

**`apps/daemon/src/media-models.ts`** / **`apps/web/src/media/models.ts`**
- Fal provider: `integrated: false` → `integrated: true`, adds `supportsCustomModel: true`.
- New image models: `flux-pro-ultra`, `flux-dev-fal`, `flux-schnell-fal`, `ideogram-v3-fal`, `recraft-v3-fal`.
- New video models: `veo-3-fal`, `veo-2-fal`, `wan-2.1-t2v`, `wan-2.1-i2v`, `seedance-1-pro-fal`, `kling-2.1-t2v-fal`.

## Surface area

- [x] **CLI / env var** — new `OD_FAL_MAX_POLL_MS` env var (optional; overrides fal queue poll ceiling; min 30 000 ms; defaults 5 min image / 10 min video)
- [ ] **i18n keys** — added new translation keys
- [ ] **New top-level dependency** — adding any new entry to the root `package.json`; none added here, uses native `fetch`
- [x] **Default behavior change** — fal provider flipped to `integrated: true`; 11 new catalog models appear in the picker by default

## Validation

Tested live against the running daemon with a real fal API key:

- `flux-pro-ultra` (16:9) → `status: done`, 690 KB JPEG, `providerNote: fal/fal-ai/flux-pro/v1.1-ultra · 16:9` — confirms `aspect_ratio` field used correctly
- `fal-ai/flux/schnell` (arbitrary path, not in catalog) → `status: done`, real image — confirms uncatalogued `fal-ai/*` passthrough works
- Endpoint corrections verified: Wan paths updated to `fal-ai/wan-t2v` / `fal-ai/wan-i2v`; Kling updated to `fal-ai/kling-video/v2.1/master/text-to-video`
- Veo duration bucket clamp: requested 5s → effective 4s, providerNote shows `4s (requested 5s → snapped to 4s)`
- Re: the `input` wrapper — tested adding `{ input: { ... } }` wrapping and received a fal 422: `"loc":["body","prompt"],"msg":"Field required"`, confirming the raw queue HTTP API expects a flat body (the `input` wrapper is a JS SDK abstraction only)


apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,221 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+  'flux-pro-ultra':      'fal-ai/flux-pro/v1.1-ultra',
+  'flux-dev-fal':        'fal-ai/flux/dev',
+  'flux-schnell-fal':    'fal-ai/flux/schnell',
+  'ideogram-v3-fal':     'fal-ai/ideogram/v3',
+  'recraft-v3-fal':      'fal-ai/recraft-v3',
+  'sora-2':              'fal-ai/sora',
+  'sora-2-pro':          'fal-ai/sora',
+  'veo-3-fal':           'fal-ai/veo3',
+  'veo-2-fal':           'fal-ai/veo2',
+  'wan-2.1-t2v':         'fal-ai/wan-video/v2.1/text-to-video',
```
lefarcen: P1 Several catalog entries point at fal endpoints that do not match the public fal model IDs, so users selecting these new picker models will get a submit-time failure instead of generation. For example, fal documents Wan 2.1 as `fal-ai/wan-t2v` / `fal-ai/wan-i2v`, and Kling 2.1 Pro is exposed as `fal-ai/kling-video/v2.1/pro/image-to-video` while text-to-video is under the Master endpoint. Please verify each endpoint in this map against the model API pages and adjust the model IDs/caps accordingly.

apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,221 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+  const baseUrl = (credentials.baseUrl || 'https://fal.run').replace(/\/$/, '');
+  const queueBase = baseUrl.includes('queue.fal.run')
+    ? baseUrl
+    : baseUrl.replace('fal.run', 'queue.fal.run');
+  const endpoint = FAL_ENDPOINTS[ctx.model] ?? ctx.model;
+  const imageSize = FAL_IMAGE_SIZES[ctx.aspect ?? '1:1'] ?? 'square_hd';
+
+  const body: Record<string, unknown> = {
+    prompt: ctx.prompt || 'A high-quality image.',
+    image_size: imageSize,
```
lefarcen: P1 This single image payload does not fit all of the catalogued fal image models. `flux-pro-ultra`, for example, expects an `aspect_ratio` field rather than the named `image_size` enum used by FLUX dev/schnell, so the newly added model can fail immediately even though it appears as integrated. Please add per-endpoint request builders or a small capability map that emits the correct schema for each built-in fal model before marking them integrated.

lefarcen (changes requested):
Hey @sdanpo, this is a strong direction — wiring fal through the daemon queue path and keeping the web/daemon registries in lockstep is exactly the right shape for this integration.

I found a couple of correctness blockers that would make some of the newly exposed catalog models fail at runtime, so please address those before this lands. Also, could you update the PR body’s Surface area checklist and Validation section? This touches a new env var/default behavior path and reviewers need to see what was actually exercised. Once the endpoint mapping/request-shape pieces are tightened up, this should be much easier to verify. 🙌

apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,221 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+  'ideogram-v3-fal':     'fal-ai/ideogram/v3',
+  'recraft-v3-fal':      'fal-ai/recraft-v3',
+  'sora-2':              'fal-ai/sora',
+  'sora-2-pro':          'fal-ai/sora',
+  'veo-3-fal':           'fal-ai/veo3',
+  'veo-2-fal':           'fal-ai/veo2',
+  'wan-2.1-t2v':         'fal-ai/wan-video/v2.1/text-to-video',
+  'wan-2.1-i2v':         'fal-ai/wan-video/v2.1/image-to-video',
+  'seedance-1-pro-fal':  'fal-ai/bytedance/seedance-1-pro',
+  'kling-2.1-t2v-fal':   'fal-ai/kling-video/v2.1/pro/text-to-video',
```
mrcfps: Several of these catalog shortcuts still point at fal paths that do not match the published model IDs, so selecting them from the new picker will 4xx even after the renderer is enabled. In the current fal docs, Wan 2.1 is exposed as `fal-ai/wan-t2v` / `fal-ai/wan-i2v`, and Kling 2.1 text-to-video is `fal-ai/kling-video/v2.1/master/text-to-video`, not the `wan-video/...` / `.../pro/text-to-video` paths mapped here. Because this PR also flips the provider to `integrated: true` and advertises these models in both registries, the broken mappings would land as user-facing failures rather than hidden follow-up work. Please re-verify each built-in alias against the current fal model pages and only expose catalog entries whose endpoint IDs are confirmed.<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=reviewer · agent=opencode · An autonomous AI dev team for your GitHub repos.</sub>

apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,221 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+  maxMs: number,
+  onProgress?: ProgressFn,
+  modelLabel?: string,
+): Promise<any> {
+  const authHeader = { 'authorization': `Key ${apiKey}` };
+
+  const submitResp = await fetch(`${queueBase}/${endpoint}`, {
+    method: 'POST',
+    headers: { ...authHeader, 'content-type': 'application/json' },
+    body: JSON.stringify(body),
```
mrcfps: The queue submit payload is missing the top-level `input` wrapper that fal expects, so this code sends `{ prompt, ... }` to `queue.fal.run/...` instead of `{ input: { prompt, ... } }`. The official queue examples for the same endpoints use `fal.queue.submit("<endpoint>", { input: { ... } })`, and they keep `status` / `result` scoped to that endpoint as well. With the current body shape, even the happy-path models like `fal-ai/flux/dev` will fail before polling ever starts. Please wrap the generated payload under `input` (and ideally mirror the documented queue contract end-to-end) before marking fal as integrated.<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=reviewer · agent=opencode · An autonomous AI dev team for your GitHub repos.</sub>

apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,221 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+    : baseUrl.replace('fal.run', 'queue.fal.run');
+  const endpoint = FAL_ENDPOINTS[ctx.model] ?? ctx.model;
+  const aspectRatio = ctx.aspect ?? '16:9';
+  const duration = ctx.length ?? 5;
+
+  const body: Record<string, unknown> = {
+    prompt: ctx.prompt || 'A short cinematic clip.',
+    aspect_ratio: aspectRatio,
+    duration,
+  };
```
mrcfps: These request builders are still treating the exposed fal catalog as if every model shared one image schema and one video schema, but the new built-ins do not. `flux-pro-ultra` expects `aspect_ratio` instead of the `image_size` enum sent here, `wan-t2v` does not accept a `duration` field at all, and Veo 3 documents duration values as suffixed strings (`4s` / `6s` / `8s`) rather than the raw numbers coming from `ctx.length`. That means multiple newly added picker models can fail at submit time even if the endpoint map is corrected. Please add per-endpoint request builders (or a capability table keyed by endpoint) so each built-in fal model emits the documented input shape before it is surfaced as integrated.<sub>🔁 Powered by <a href="https://github.com/nexu-io/looper">Looper</a> · runner=reviewer · agent=opencode · An autonomous AI dev team for your GitHub repos.</sub>

mrcfps (changes requested):
@sdanpo Thanks for wiring fal into the daemon — the queue-based integration direction makes sense, but I hit a few concrete runtime blockers in the changed paths that would still make the new fal surface fail for users. Please fix the queue submit shape and the built-in model wiring before merge.

lefarcen (changes requested):
Hey @sdanpo, I re-checked the current head after @mrcfps’s review. The fal integration direction is still promising, but the blocking issues remain on this commit.

@mrcfps’s review now covers the same endpoint/schema problems plus the missing queue `input` wrapper, so please treat those threads together with my earlier request before the next iteration. I don’t have additional inline notes beyond the existing blocking comments right now. 🙏

apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,251 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+  maxMs: number,
+  onProgress?: ProgressFn,
+  modelLabel?: string,
+): Promise<any> {
+  const authHeader = { 'authorization': `Key ${apiKey}` };
+
+  const submitResp = await fetch(`${queueBase}/${endpoint}`, {
+    method: 'POST',
+    headers: { ...authHeader, 'content-type': 'application/json' },
+    body: JSON.stringify(input),
```
lefarcen, line 2604: P1 This still submits the raw model input as the top-level JSON body, but fal queue submit expects the payload to be wrapped as `{ input: { ... } }`. Renaming the parameter to `input` does not change the wire shape, so even fixed endpoint IDs will still fail at submit time. Please send `JSON.stringify({ input })` here and keep callers passing only the model input object.

lefarcen (changes requested):
Hey @sdanpo, thanks for the quick follow-up here — the endpoint map and per-model request-shape direction moved in the right direction.

One core queue-contract blocker is still present on this head, so my previous block cannot be cleared yet. I also noticed the PR body still has the Surface area checklist unchecked; please update that too so reviewers can scope the user-visible/provider changes correctly. Once the queue submit shape is fixed, this should be ready for another pass. 🙏

apps/daemon/src/media.ts
sdanpo: Thanks for the re-review @lefarcen. We actually tried this exact change and it caused a fal 422 — here's the exact error we got when sending `{ input: { prompt, ... } }`:

```
fal result 422: {"detail":[{"type":"missing","loc":["body","prompt"],"msg":"Field required","input":{"input":{"prompt":"a glowing neon geometric pattern","num_images":1,"aspect_ratio":"16:9"}}}]}
```

The `loc: ["body","prompt"]` means fal received the wrapped `{ input: { prompt } }` body and looked for `prompt` at the top level — it wasn't there. The current flat-body approach produces a real image every time we test.

The `input` wrapper appears in fal's **JS SDK** examples (`fal.queue.submit(endpoint, { input: {...} })`) because the SDK adds it internally, but fal's own **raw HTTP curl examples** send a flat body:
```bash
curl -X POST "https://queue.fal.run/fal-ai/flux/dev"   -H "Authorization: Key $FAL_KEY"   -d '{"prompt": "..."}'
```

We're making raw HTTP calls, not using the SDK, so the flat body is correct. Happy to re-test with the wrapper against a different model if you have a specific fal endpoint where the wrapper is required — but the evidence from `fal-ai/flux-pro/v1.1-ultra` is that it breaks things.

apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,251 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+
+  const input: Record<string, unknown> = {
+    prompt: ctx.prompt || 'A short cinematic clip.',
+    aspect_ratio: aspectRatio,
+  };
+  // Some models (Wan) have no duration parameter; others (Veo) require a
+  // suffixed string ("5s") rather than a bare integer.
+  if (!FAL_VIDEO_NO_DURATION.has(endpoint)) {
+    input.duration = FAL_VIDEO_STRING_DURATION.has(endpoint)
+      ? `${durationSec}s`
```
lefarcen: P1 Veo now gets a suffixed duration, but it can still be an invalid value. The shared video clamp allows/defaults to values like `5`, `10`, `15`, and `30`, so `veo-3-fal` with the default path sends `duration: "5s"` even though the fal Veo endpoints only accept their documented duration buckets such as `4s`, `6s`, or `8s`. Please add a model-specific duration clamp/default for the Veo endpoints (or omit duration when unsupported) so the newly exposed Veo models do not fail on the default generation path.

lefarcen (changes requested):
Hey @sdanpo, thank you for the concrete fal 422 evidence — you’re right on the raw HTTP queue body. I’d mixed the JS SDK wrapper shape with the REST wire shape there, and the flat body is the correct call for this implementation.

The previous queue-wrapper blocker is cleared from my side. I found one remaining runtime blocker in the current head around Veo duration values, so I’m keeping this as changes requested with the single inline note below. The Surface area checklist in the PR body is also still unchecked; please update that before the next review pass. 🙏

apps/daemon/src/media.ts
sdanpo: Fixed in c4af58f0. Added `FAL_VEO_DURATION_BUCKETS = [4, 6, 8]` and a nearest-bucket clamp: when Veo is the target endpoint the raw duration is snapped to whichever of 4, 6, or 8 is closest (e.g. 5 → 4, 7 → 8, 10 → 8), then formatted as `"Ns"`. Wan models still have duration omitted entirely via `FAL_VIDEO_NO_DURATION`.

apps/daemon/src/media.ts
```diff
@@ -1728,6 +1753,259 @@ async function renderFishAudioTTS(ctx: MediaContext, credentials: ProviderConfig
...
+  if (!videoUrl) {
+    throw new Error(`fal video missing video.url: ${truncate(JSON.stringify(result), 200)}`);
+  }
+  const dlResp = await fetch(videoUrl);
+  if (!dlResp.ok) throw new Error(`fal video download ${dlResp.status}`);
+  const bytes = Buffer.from(await dlResp.arrayBuffer());
+
+  return {
+    bytes,
+    providerNote: `fal/${endpoint} · ${aspectRatio} · ${durationSec}s · ${bytes.length} bytes`,
```
lefarcen: P2 After clamping Veo requests to the nearest valid bucket, this note still reports the original `durationSec` rather than the duration actually sent upstream. For example, a default/requested 5s Veo render now sends `4s` but returns a provider note saying `5s`, and 10s reports `10s` even though the request sends `8s`. Please track the effective duration label used in the request and use that in the provider note (and ideally surface a warning when a requested length is snapped).

lefarcen (changes requested):
Hey @sdanpo, nice quick fix on the Veo bucket clamp — the request body now uses a valid `4s` / `6s` / `8s` value, which addresses the submit-time blocker.

There’s one small follow-through issue from that clamp before I can clear my prior changes-requested review: the response metadata still reports the unclamped duration, so users can see a different duration than the one we actually sent to fal. Also, the PR body’s Surface area section is still not using the template checklist boxes; please update that for the next pass too. 🙏

apps/daemon/src/media.ts
sdanpo: Fixed in 6b0fecb1. The function now tracks `effectiveDurationLabel` (the duration actually sent in the request) separately from `durationSec` (the raw requested value). The providerNote now uses the effective label, and when a snap occurred it appends e.g. `(requested 5s → snapped to 4s)`. PR body Surface area section also updated to use the template checklist format with the correct boxes checked.

lefarcen (dismissed):
Hey @sdanpo, this addresses my blocker. The effective Veo duration is now reflected in `providerNote`, the snap is visible when it happens, and the PR body Surface area checklist now matches the template.

Clean from my side on this head. @mrcfps’s review and product review are still the remaining merge gates, but my previous changes-requested review is cleared. Nice iteration here. 🙌

apps/daemon/src/media.ts
mrcfps: Looper checked this thread against head `6b0fecb1cb96b221da2953fab8df03e8846a2e22`. The requested change appears objectively addressed by apps/daemon/src/media.ts:1789-1818 adds endpoint-specific capability tables, renderFalImage switches between aspect_ratio and image_size at 1928-1934, and renderFalVideo omits duration for Wan and sends suffixed Veo durations at 1975-1988., so I’m resolving this thread. Reopen if this still needs discussion.



mrcfps (dismissed):
@sdanpo I re-checked the fal queue contract, the Veo duration handling/providerNote flow, and the mirrored daemon/web model catalog updates on this head, and the runtime blockers from the earlier review pass look addressed. Thanks for iterating on this so quickly — this is in a much better place now. 🙌

mrcfps:
@sdanpo I'm holding off on generating review comments for nexu-io/open-design#1606 because this pull request has merge conflicts right now.

mrcfps:
@sdanpo I'm holding off on generating review comments for nexu-io/open-design#1606 because this pull request has merge conflicts right now.

apps/daemon/src/media.ts
mrcfps: Looper checked this thread against head `aaaae409d3936ff8ed227a7dc00ef02a8aa498fd`. The requested change appears objectively addressed by apps/daemon/src/media.ts:1939-1942 now maps Wan aliases to `fal-ai/wan-t2v` and `fal-ai/wan-i2v`, and Kling 2.1 t2v to `fal-ai/kling-video/v2.1/master/text-to-video`, matching the requested endpoint corrections., so I’m resolving this thread. Reopen if this still needs discussion.



apps/daemon/src/prompts/media-contract.ts
```diff
@@ -276,22 +290,47 @@ substitution. Do not silently fall back.
...
+   - General default when no cue → \`1:1\`
+
+   **Skip the discovery question when all of these are true:**
+   - The subject is described (what to generate)
+   - The style or mood is implied or stated (realistic, cinematic, illustrated, etc.)
+   - Any model/aspect gaps can be filled with the defaults above
+
+   **Do ask** if the output intent is genuinely ambiguous (e.g. "make something cool"
+   with no subject), or the user explicitly requests a model/voice the project
+   metadata doesn't carry.
```
mrcfps, line 395: This workflow block now teaches the media agent to dispatch immediately, but it still never tells the agent that full `fal-ai/*` ids are valid model inputs. A few lines later, the unchanged guard in `apps/daemon/src/prompts/media-contract.ts` says any model not in the registered list should be rejected or substituted instead. That means requests like `fal-ai/flux/schnell` or `fal-ai/wan-i2v` will still be steered back to a catalog model in chat, even though `apps/daemon/src/media.ts` now accepts those paths directly. Please add an explicit `fal-ai/*` exception here (or relax the later registered-id rule for Fal custom models) so the agent forwards uncatalogued Fal ids unchanged.

mrcfps (changes requested):
@sdanpo Thanks for wiring the fal queue renderer through — the daemon integration and catalog sync are much better on this head. I found one remaining blocker in the media-agent contract: the prompt still steers uncatalogued `fal-ai/*` ids away from the new passthrough path, so the custom-model flow advertised in this PR will not work from the normal chat path until that exception is encoded here. 🙏

apps/daemon/src/prompts/system.ts
```diff
@@ -137,6 +137,58 @@ export const SKIP_DISCOVERY_BRIEF_OVERRIDE = `# Automated project mode — skip
...
+  last=\$(printf '%s\\n' "\$out" | tail -1)
+  since=\$(printf '%s\\n' "\$last" | python3 -c "import sys,json; d=json.load(sys.stdin); print(d.get('nextSince',\$since))" 2>/dev/null)
+  since="\${since:-0}"
+done
+printf '%s\\n' "\$last"
+\`\`\`
+
+**Never ask the user for an API key.** The daemon reads provider credentials from its config; keys are never passed through the shell. If the provider returns an auth error, tell the user to open Settings → AI Providers and confirm the key is configured there.
+
+For the best fal image model use \`--model flux-pro-ultra\`. For video use \`--model veo-3-fal\` or \`--model wan-2.1-t2v\`. The dispatcher infers \`--surface\` from the model when unambiguous.`;
```
mrcfps: This helper block is still describing the old CLI contract, and that breaks the exact prototype/deck chat path this PR is trying to repair. In `apps/daemon/src/cli.ts:351-355`, `od media generate` still requires an explicit `--surface`, and in `apps/daemon/src/cli.ts:401-404` / `526-536` it now exits **0** for a queued handoff while returning `{"taskId": ...}` on stdout. The loop here waits only while `$ec == 2` and the footer says the dispatcher can infer `--surface`, so a non-media agent following this prompt will stop after printing the handoff JSON for slow Fal models like `flux-pro-ultra` / `veo-3-fal`, or fail immediately if it omits `--surface`. Please update this hint to (1) keep polling whenever the final JSON contains a `taskId`, regardless of the initial generate exit code, (2) stop only on `media wait` exit 0/5, and (3) remove the `--surface` inference guidance unless the CLI actually supports it.

mrcfps (changes requested):
@sdanpo Thanks for pushing this through — the fal renderer and catalog sync are close, but I found one remaining main-path blocker in the new non-media prompt hint that would stop slow Fal renders from completing in regular chat. Please tighten that helper before merge.

apps/daemon/src/prompts/system.ts
mrcfps: Looper checked this thread against head `b64431c7c3b72272c0e1e703da35244f010b9a65`. The requested change appears objectively addressed by apps/daemon/src/prompts/system.ts:159-194 now says `media generate` exits 0 and may return `{\"taskId\":...}`, loops whenever `taskId` is present, and explicitly requires `--surface`; this matches apps/daemon/src/cli.ts:351-355 and 401-404., so I’m resolving this thread. Reopen if this still needs discussion.



mrcfps (commented):
@sdanpo Thanks for pushing this forward — the Fal renderer and prompt contract are much closer now. I found one follow-up to land with this pass: the new handoff wording in the media-generation contract no longer matches an existing daemon assertion, so the daemon prompt test suite will fail until that expectation is updated. 🙏

Location: apps/daemon/src/prompts/media-contract.ts RIGHT lines 190-241

This handoff block rewrites the generated prompt text, but the daemon prompt regression test still asserts the old copy in `apps/daemon/tests/system-prompt-template.test.ts:331-345` (`media generate` treats the handoff as, `exit `0` so the first dispatch does not look like a failed shell call`, and ``2` from `media wait` is not a failure`). Those strings no longer exist anywhere in the prompt sources on this head, so that test will fail as soon as the daemon suite runs. Please update the expectations to the new handoff wording in the same PR (or keep a compatibility phrase here) so the prompt/CLI contract change lands green.

apps/daemon/src/prompts/system.ts
```diff
@@ -274,14 +331,32 @@ export function composeSystemPrompt({
...
+    metadata?.kind === 'video' ||
+    metadata?.kind === 'audio';
+
+  if (!isMediaSurfaceEarly) {
+    parts.push(DISCOVERY_AND_PHILOSOPHY, '\n\n---\n\n');
+  }
+
   if (metadata?.skipDiscoveryBrief === true) {
     parts.push(SKIP_DISCOVERY_BRIEF_OVERRIDE);
     parts.push('\n\n---\n\n');
```
mrcfps: This still appends `SKIP_DISCOVERY_BRIEF_OVERRIDE` *after* `DISCOVERY_AND_PHILOSOPHY`, so the new `skipDiscoveryBrief: true` path does not actually place the “do NOT emit `<question-form id="discovery">`” instruction ahead of the normal discovery rules. That matters because batch-created prototype/deck runs can still see the first-turn discovery guidance before the override, which is the opposite of what this flag is supposed to suppress. The regression is already visible in the changed daemon suite: `pnpm --filter @open-design/daemon test -- --runInBand apps/daemon/tests/system-prompt-template.test.ts` now fails `pins the API batch-mode discovery skip before the normal discovery rules` because `overrideIdx` ends up after `discoveryIdx`. Please move this override ahead of the discovery block (or skip `DISCOVERY_AND_PHILOSOPHY` entirely when `skipDiscoveryBrief` is true) so the new runtime path and its added test agree.

mrcfps (commented):
@sdanpo Thanks for tightening the Fal renderer and the prompt handoff flow — I rechecked the latest head, reran `pnpm guard`, and reran the daemon suite locally. I found one remaining correctness issue in the new `skipDiscoveryBrief` prompt wiring that still needs a follow-up in this PR.

apps/daemon/src/prompts/system.ts
mrcfps: Looper checked this thread against head `f9bcf3ce2b6b7dcdd04fd8802d5b5bba2a67d507`. The requested change appears objectively addressed by apps/daemon/src/prompts/system.ts now pushes `SKIP_DISCOVERY_BRIEF_OVERRIDE` before `DISCOVERY_AND_PHILOSOPHY` (`parts.push(SKIP_DISCOVERY_BRIEF_OVERRIDE)` at lines 349-352, then discovery at 354-355), and apps/daemon/tests/system-prompt-template.test.ts asserts `overrideIdx < discoveryIdx` for `skipDiscoveryBrief: true`., so I’m resolving this thread. Reopen if this still needs discussion.



e2e/scripts/test-fal-webui.ts
```diff
@@ -0,0 +1,159 @@
+/**
+ * Ad-hoc Playwright script: test fal image generation through the live web UI.
```
mrcfps: `e2e/scripts/test-fal-webui.ts` is repo-owned Playwright UI automation, but this file lives in `e2e/scripts/` and then needs a one-off `scripts/guard.ts` allowlist entry instead of going through the package's normal `e2e/ui/*.test.ts` / `tests` flows. That matters because the script hard-codes a live local port (`WEB_URL`), launches a headed browser, and drives a real Fal-backed prompt, so it can't participate in the e2e package's fixture/reporting/cleanup conventions or provide stable CI signal for future maintainers. Please either convert this into a real `e2e/ui/*.test.ts` / `tests` case with the package harness and a mocked provider boundary, or keep it out of the committed allowlist if it's only ad-hoc local validation.

mrcfps (commented):
@sdanpo Thanks for pushing the Fal renderer and prompt-contract fixes through — I rechecked the changed ranges on this head and found one remaining repo-hygiene follow-up in the new validation script before this is fully tidy. 🙏

mrcfps:
@sdanpo I'm holding off on generating review comments for nexu-io/open-design#1606 because this pull request has merge conflicts right now.

bash.exe.stackdump
```diff
@@ -0,0 +1,30 @@
+Stack trace:
```
mrcfps: This file is a raw MSYS bash crash dump, not a repo-owned source artifact. Keeping it in the PR adds opaque machine-specific noise to the checkout and future diffs, and nothing in the build/test flow consumes it. The evidence is the file content itself: it is only a `bash.exe` stack trace plus loaded DLL list. Please drop `bash.exe.stackdump` from the PR rather than committing it; if you still need it for debugging, attach it to the PR discussion or keep it in local scratch only.

fix_loop.py
```diff
@@ -0,0 +1,17 @@
+import sys
+sys.stdout.reconfigure(encoding='utf-8')
+path = 'F:/open-design/apps/daemon/src/prompts/system.ts'
```
mrcfps: This helper is also a one-off local scratch artifact rather than project code. It hard-codes a Windows-only `F:/open-design/apps/daemon/src/prompts/system.ts` path and only performs a single text replacement against `apps/daemon/src/prompts/system.ts`, so it is not portable, referenced, or maintainable as a committed repo file. Please remove `fix_loop.py` from the PR and keep this kind of one-shot rewrite helper in local scratch outside the repository if it was only used during development.

mrcfps (commented):
@sdanpo Thanks for pushing the Fal renderer and prompt-contract fixes through — I re-checked the current changed ranges on this head and found two small repo-hygiene follow-ups in files added at the repo root. The functional daemon/prompt changes look much closer now; these are the remaining actionable cleanups I’d land with this pass.

sdanpo:
Conflicts resolved and pushed (cf839877). Also removed the ad-hoc `e2e/scripts/test-fal-webui.ts` dev helper per reviewer feedback — it required a live API key and hardcoded port, so it can't be a proper CI test. Guard and daemon typecheck pass clean. Ready for another review pass.

apps/daemon/src/prompts/system.ts
```diff
@@ -142,6 +142,62 @@ export const SKIP_DISCOVERY_BRIEF_OVERRIDE = `# Automated project mode — skip
...
+  elif [ "\$ec" -ne 2 ]; then
+    echo "\$out" >&2; exit "\$ec"
+  fi
+done
+printf '%s\\n' "\$last"
+\`\`\`
+
+**Never ask the user for an API key.** The daemon reads provider credentials from its config; keys are never passed through the shell. If the provider returns an auth error, tell the user to open Settings → AI Providers and confirm the key is configured there.
+
+For the best fal image model use \`--model flux-pro-ultra\`. For video use \`--model veo-3-fal\` or \`--model wan-2.1-t2v\`. Always pass \`--surface\` explicitly (\`image\`, \`video\`, or \`audio\`).`;
```
mrcfps: This new non-media hint still never tells prototype/deck agents that arbitrary `fal-ai/*` ids are valid `--model` values. That matters because the PR body now advertises custom Fal passthrough, but this prompt path only teaches named catalog examples (`flux-pro-ultra`, `veo-3-fal`, `wan-2.1-t2v`), so a user asking for something like `fal-ai/flux/schnell` from a prototype project still has no instruction to forward the raw id unchanged. Please add one sentence here mirroring the `media-contract.ts` exception (for example: any `fal-ai/*` path is valid for image/video and should be passed through as `--model <id>` without substitution), and pin that wording in the prompt test so the media and non-media prompt paths stay aligned.

mrcfps (commented):
@sdanpo Thanks for cleaning up the Fal renderer and follow-up prompt fixes — I rechecked the current changed ranges and found one remaining follow-up in the new non-media media-dispatch hint.

e2e/scripts/test-fal-webui.ts
mrcfps: Looper checked this thread against head `536c889007eb0623a29d987ba8292aa765803814`. The requested change appears objectively addressed by e2e/scripts/test-fal-webui.ts is absent from the worktree, and scripts/guard.ts:16-19 allowlists only e2e/scripts/playwright.ts and e2e/scripts/release-smoke.ts, so the ad-hoc Playwright script is no longer committed/allowlisted., so I’m resolving this thread. Reopen if this still needs discussion.



bash.exe.stackdump
mrcfps: Looper checked this thread against head `536c889007eb0623a29d987ba8292aa765803814`. The requested change appears objectively addressed by bash.exe.stackdump is absent from the current worktree., so I’m resolving this thread. Reopen if this still needs discussion.



fix_loop.py
mrcfps: Looper checked this thread against head `536c889007eb0623a29d987ba8292aa765803814`. The requested change appears objectively addressed by fix_loop.py is absent from the current worktree., so I’m resolving this thread. Reopen if this still needs discussion.



apps/daemon/src/prompts/system.ts
mrcfps: Looper checked this thread against head `536c889007eb0623a29d987ba8292aa765803814`. The requested change appears objectively addressed by apps/daemon/src/prompts/system.ts:199 now states that any fal-ai/* path is a valid --model value and should be passed through unchanged, and apps/daemon/tests/system-prompt-template.test.ts:202-209 pins that wording in a prompt test., so I’m resolving this thread. Reopen if this still needs discussion.



mrcfps (approved):
@sdanpo I rechecked the current changed ranges around the Fal renderer, the media prompt contract, and the mirrored daemon/web model catalog updates on this head. `pnpm guard` passed, and `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/system-prompt-template.test.ts` passed locally for the prompt-path follow-ups. Thanks for working through the review iterations here — this looks ready from my side. 🙌

sdanpo:
@
Still active here — just pushed `99223614`, a clean merge of latest `main` (the branch was ~174 commits behind) so this is back in sync and conflict-free.

State on this head:
- Both code reviews are cleared — @lefarcen dismissed and @mrcfps approved on `536c8890`.
- The merge surfaced no conflicts in the fal renderer files (`media.ts`, `media-models.ts`, `media-contract.ts`, `system.ts`, `models.ts`, `cli.ts`, the prompt test).
- Re-validated locally after `pnpm install`: `pnpm guard` and `pnpm typecheck` pass clean (0 errors), and the daemon prompt suite (`tests/system-prompt-template.test.ts`) is 21/21 green.

Ready for the remaining product-review pass.
@

lefarcen:
Hey @sdanpo — appreciate you keeping this moving on May 20. Since then, main has pushed ahead again and the branch is showing merge conflicts once more.

One more rebase on top of current main would get you back to a clean merge state. Once that's done, mrcfps's approval is already on record and the code review side looks solid — you'd be close to the finish line.

Let us know if you hit anything tricky in the rebase. 👀

mrcfps:
Thanks @sdanpo — this one took real persistence to land and it shows. The core fal renderer is solid work, but what stands out is everything that came after: the Veo duration bucket clamping, the latency fix that took generation from 4m37s down to ~73s by cutting the discovery layer for media surfaces and bumping the wait budget to 120s, the generate→wait loop in the dispatch hint, and the PowerShell env var guard for Windows agents. Each of those was a real edge case that would have quietly broken users in production.

The `fal-ai/*` passthrough is a particularly nice touch — the full fal model library without any catalog maintenance overhead.

Looking forward to seeing what you tackle next. If you're up for it, Replicate would be a natural follow-on given the same queue-based polling pattern you've already proven out here.

leilei926524-tech:
Hi @sdanpo!

You contributed:

Merged PR: #1606 feat(daemon): implement fal.ai renderer for image + video generation
https://github.com/nexu-io/open-design/pull/1606

You worked on the web app UX, so these follow-ups stay close to the same user-facing surface.

1. #569 wix-ja-slide / google-slides daemon: silent private-Drive fallback breaks image insertion under Workspace DLP
https://github.com/nexu-io/open-design/issues/569

This still sits in the web product surface and looks like a manageable next step.

2. #2874 Hide raw local executable paths from Live Artifact running status UI
https://github.com/nexu-io/open-design/issues/2874

This is another focused web UI fix, so the scope should feel familiar.

The Open Design team

PR #1607, opened 2026-05-13
feat(web): add collapsible comment side panel

### LINKED ISSUE #1605: Add a collapsible comments panel in the design preview
### Problem

When comments are open in the desktop design preview, the comment panel takes a large fixed area on the right side and covers the right half of the design file.

This makes it difficult to inspect the full design, especially for layouts where important content or controls are on the right side. After selecting an element/comment, the panel remains visible and the design preview becomes partially obscured.

### Expected behavior

There should be a way to collapse or hide the comments panel while keeping the design preview visible.

A possible interaction:

- Add a collapse/expand button on the comments panel.
- When collapsed, keep a small icon or narrow rail so comments can be reopened quickly.
- The design preview should be able to use the freed horizontal space.

### Actual behavior

The comments panel stays open and blocks a significant part of the design preview. Users have to work with only the remaining visible area, which makes the workspace feel cramped.

### Why this matters

The comments panel is useful, but it should not permanently reduce the usable design canvas. A collapsible comments panel would make reviewing and editing designs much easier on desktop, especially when the viewport is not very wide.

Romantin:
Fixes #1605

## Why

While reviewing generated design files, the comments side panel can permanently cover the right side of the artifact preview. That makes it harder to inspect layouts, especially when the design canvas already needs the full available width. This PR keeps comments available while letting users temporarily reclaim the preview area.

## What users will see

In the HTML preview comments panel, users will see a collapse button in the comments header. Clicking it hides the full comments panel and leaves a narrow comments rail on the right with the comment label and count. Clicking the rail restores the comments panel. The default state is unchanged: the comments panel still opens expanded.

## Surface area

- [x] **UI** - new page / dialog / panel / menu item / setting / empty state in `apps/web` or `apps/desktop` (including Electron menu bar)
- [ ] **Keyboard shortcut** - new or changed
- [ ] **CLI / env var** - new `od` subcommand or flag, new `tools-dev` / `tools-pack` / `tools-pr` flag, or new `OD_*` env var
- [ ] **API / contract** - new `/api/*` endpoint, new SSE event, or changed shape in `packages/contracts`
- [ ] **Extension point** - new entry under `skills/`, `design-systems/`, `design-templates/`, or `craft/`, or change to the skills protocol
- [ ] **i18n keys** - added new translation keys (see `TRANSLATIONS.md` for the locale workflow)
- [ ] **New top-level dependency** - adding any new entry to the **root** `package.json` (`dependencies` or `devDependencies`); workspace-package `package.json` files are out of scope. Include a paragraph on what we get vs. what bytes we ship (see `CONTRIBUTING.md` -> Code style)
- [ ] **Default behavior change** - changes what existing users experience without opting in (default model, default setting, file/SQLite schema, auto-network on startup, auto-install)
- [ ] **None** - internal refactor, docs, tests, or translation update only

## Screenshots

Not attached; this is a small chrome-only change in the existing comments panel. The entry point and collapse/restore behavior are covered by the component test in `apps/web/tests/components/FileViewer.test.tsx`.

## Bug fix verification

- Test path that reproduces the bug: `apps/web/tests/components/FileViewer.test.tsx`
- Did the test go red on `main` and green on this branch? yes

## Validation

- `pnpm --filter @open-design/web test -- FileViewer.test.tsx`
- `pnpm i18n:check`
- `git diff --check`
- `pnpm guard`

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -1991,9 +1991,11 @@ function commentAvatarInitial(comment: PreviewComment): string {
   return seed.charAt(0).toUpperCase();
 }
 
-function CommentSidePanel({
+export function CommentSidePanel({
   comments,
   selectedIds,
+  collapsed,
+  onCollapsedChange,
   onToggleSelect,
   onClearSelection,
   onReply,
@@ -2003,6 +2005,8 @@ function CommentSidePanel({
 }: {
   comments: PreviewComment[];
   selectedIds: Set<string>;
+  collapsed: boolean;
+  onCollapsedChange: (collapsed: boolean) => void;
   onToggleSelect: (commentId: string) => void;
   onClearSelection: () => void;
   onReply: (comment: PreviewComment) => void;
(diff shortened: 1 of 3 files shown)


apps/web/src/components/FileViewer.tsx
```diff
@@ -2013,8 +2017,39 @@ function CommentSidePanel({
   const sorted = [...comments].sort((a, b) => b.createdAt - a.createdAt);
   const visibleSelectedIds = new Set(comments.filter((comment) => selectedIds.has(comment.id)).map((comment) => comment.id));
   const selectedCount = visibleSelectedIds.size;
+  if (collapsed) {
+    return (
+      <button
+        type="button"
+        className="comment-side-rail"
+        data-testid="comment-side-collapsed-rail"
+        aria-label="Show comments"
```
lefarcen: P2 This new control hard-codes English accessible text, and the collapse button below does the same. The visible comment-panel label already goes through `t('chat.tabComments')`, so non-English users would still hear English-only show/collapse labels from assistive tech. Please add localized Dict keys for the show/collapse comment actions in the locale files and read them via `t(...)` here and on the collapse button/title.

lefarcen (changes requested):
Hey @Romantin, nice job turning #1605 into a focused UI change — the component-level collapse/restore path and the regression test are both in the right direction.

I found one localization/accessibility blocker before this can land: the new controls expose English-only accessible text. I also noticed the PR description is missing a few template sections that help reviewers and release checks understand the user-visible surface. Could you fill in:

- **What users will see**
- **Surface area** (this should include UI)
- **Validation**

If you're using an AI assistant, this prompt should be enough to update the body cleanly:

> Read the current PR body for nexu-io/open-design#1607 and the template at
> Then update via `gh pr edit 1607 --repo nexu-io/open-design --body-file -`.

Once the localized labels are wired and the body is filled in, this should be a straightforward next pass. 🙌

Romantin:
Updated the review items:

- Replaced the hardcoded English accessible text with existing localized labels: `preview.showSidebar` / `preview.hideSidebar` using `chat.tabComments` as the label.
- Updated the component test to assert the localized button names.
- Filled the PR body with the template sections for user-visible behavior, surface area, screenshots note, bug verification, and validation.

Validation run:
- `pnpm --filter @open-design/web test -- FileViewer.test.tsx`
- `pnpm i18n:check`
- `git diff --check`
- `pnpm guard`

lefarcen (approved):
Hey @Romantin, thanks for the quick follow-up. The localized show/hide labels now reuse the existing sidebar strings correctly, and the updated test covers the accessible button names as well as the collapse/restore flow.

The PR body also now gives reviewers the user-facing behavior, UI surface, screenshot note, bug verification, and validation details we needed. Approving from my side to clear my previous changes-requested review. Product review is still requested for the UI surface before merge. 🙌

PR #1608, opened 2026-05-13
fix(orbit): avoid sample identity leakage

### LINKED ISSUE #1269: Orbit uses the wrong user identity from connector data and shows Eli instead of the authorized account
## Summary

Orbit is generating output with the wrong user identity. The authorized connector account is not Eli, but the generated Orbit artifact still shows **Eli** in the personalized greeting/content.

Priority: P1
Reason: This is a core data-mapping error that makes Orbit output untrustworthy. The feature still runs, but it binds the result to the wrong identity.

## Current behavior

- The connected / authorized connector account is not Eli.
- Orbit runs successfully.
- The generated Orbit artifact still renders the user identity as **Eli**.

## Expected behavior

Orbit should use the identity from the actually authorized connector account and should not show a stale, default, or unrelated user name.

## Steps to reproduce

1. Connect a connector account that does not belong to Eli.
2. Run Orbit.
3. Open the generated Orbit artifact.
4. Observe that the greeting or personalized content still shows **Eli**.

## Notes

- Screenshots were provided showing the Orbit artifact rendering **Eli** while the connected account is different.
- This looks like stale identity data, incorrect binding, or fallback/default-user leakage into the generated template.
- The issue affects trust in personalized Orbit output.

## Why this matters

If Orbit shows the wrong account identity, users cannot trust that the generated digest belongs to the currently connected source. This is especially problematic for connector-driven summaries where account attribution is part of the core value.

Image

Image

YUHAO-corn:
Fixes #1269

## Why

Orbit briefing templates still carried a concrete sample user identity (`Eli`) in their examples and template guidance. When an agent mirrors those examples too literally, the generated briefing can greet or attribute work to the sample person instead of the authorized connector account, which is a trust issue for a personalized daily briefing.

This PR removes that sample identity leak and adds explicit template guidance so future Orbit generations treat names in `example.html` as mock content, not as the current user.

## What users will see

Users generating Orbit briefings should no longer see the sample name `Eli` copied into generated greetings, titles, mentions, assignee labels, document metadata, or signed-in-user avatars when the current user's identity is not explicitly available.

When connector/request data does not provide an authorized display name, the templates now fall back to neutral wording such as `you`, `your`, `current user`, `我`, or `你`.

## Surface area

- [x] **Extension point** — new entry under `skills/`, `design-systems/`, `design-templates/`, or `craft/`, or change to the skills protocol

## Screenshots

Not applicable: this updates Orbit template examples/instructions rather than an in-app static UI surface.

## Bug fix verification

- Test path that reproduces the bug: no cheap deterministic runtime spec exists for this template-generation path; the issue is example/prompt content being copied into generated artifacts.
- Did the test go red on `main` and green on this branch? no automated red/green spec was added.
- Verification performed instead: searched all Orbit templates for remaining `Eli` / `eli` sample identity leaks and ran the repository guard.

## Validation

- `rg "\\beli\\b|\\bEli\\b|class=\"user-avatar\">E<|class=\"act-av\">E<" design-templates/orbit-* -n`
- `pnpm guard`
- GitHub CI: `Detect packaged smoke changes`, `Validate workspace`, and `nix-check build`


lefarcen (commented):
Hey @YUHAO-corn, thanks for jumping on the Orbit identity leak. The concrete template direction looks right: the hard-coded sample name is gone from the Orbit examples, and the new guard makes it much less likely that mock names get promoted into the generated briefing. I do not see line-level blockers in the template diff.

One housekeeping ask before the next reviewer pass: the PR body is still missing a few sections from our template — **Why**, **What users will see**, **Surface area**, and **Validation**. Since this fixes a P1 trust issue, having the user-facing behavior and exact validation spelled out will make it easier for reviewers to confirm the fix against #1269.

If you are using an AI assistant, you can paste this prompt to fill it in quickly:

> Read the current PR body for nexu-io/open-design#1608 and the template at
> Then update via `gh pr edit 1608 --repo nexu-io/open-design --body-file -`.

No rush — this just helps the next review iteration go faster. 🙏

PerishCode (commented):
@YUHAO-corn — thanks for picking up #1269. The SKILL.md identity-guard paragraph plus the swap of mock identity strings across the five Orbit templates lands the core fix cleanly: the visible `Eli` greetings, `@Eli` mentions, assignee, doc-meta, and email subject are all gone, and the new guard tells future agents not to infer the current user from `example.html` mock content. Nice symmetry with the `orbit-github` reviewer-avatar update (`E` → `Y`, `eli — pending` → `you — pending`) that establishes the rename convention for adjacent initials.

One pattern is partially applied: in two of the touched files the visible `Eli` text was renamed to `You`, but the single-letter avatar initial (`E`, originally Eli's initial) was left in place. The inline comment below on `orbit-linear/example.html` is the resolvable thread for that file; the `orbit-gmail` occurrence is outside the diff hunks so it is captured here as fallback location text.

**Fallback location:** `design-templates/orbit-gmail/example.html:408`

The Gmail header user-avatar (top-right of the chrome) is still rendered as `<div class="user-avatar">E</div>` — `E` was Eli's initial, and that avatar represents the signed-in user. Since the rest of this file's user-facing strings were just neutralized (`Eli` removed from `<title>`, `email-subject`, and `greeting`), the leftover `E` in the signed-in-user slot is the same shape of mock-identity leak the PR set out to eliminate. Suggested change at line 408:

```html
<div class="user-avatar">Y</div>
```

or swap to a generic person glyph (matching the SKILL.md guidance to fall back to neutral wording when no explicit current-user name is available). Either lines up with the `orbit-github/example.html:541` precedent in this same PR.

Non-blocking — happy to see this merge once you decide whether to address these in-PR or in a follow-up.



design-templates/orbit-linear/example.html
```diff
@@ -347,7 +347,7 @@ <h1>Hello, Eli</h1>
             </div>
             <div class="act">
               <div class="act-av">E</div>
```
PerishCode: **Stale `E` avatar initial remains after the `Eli` → `You` rename (same pattern at line 458).**

Problem: this `<div class="act-av">E</div>` on line 349, and the matching `<div class="act-av">E</div>` on line 458, still encode the original mock identity `Eli`. The adjacent `<strong>` label next to each one was correctly switched to `You` in this PR, but the single-letter avatar initial above it was left untouched, so the avatar no longer agrees with the name it sits next to.

Why it matters: the stated goal of this PR (and of #1269) is to remove sample-identity leakage so the generated briefing can't bind output to the wrong user. The visible `Eli` strings are gone, but leaving `E` directly beside `<strong>You</strong>` is exactly the kind of mock-identity remnant the new SKILL.md identity guard tells future agents to ignore — and it diverges from the convention this PR already adopts in `design-templates/orbit-github/example.html:541`, where the reviewer-avatar initial was updated from `E` to `Y` in lockstep with the `eli` → `you` rename. Other activity rows in this file follow `S/Sara`, `M/Marie`, `O/Orbit`, so `E/You` is the only mismatched pair after the change.

Evidence: diff hunks `@@ -347,7 +347,7 @@` and `@@ -449,14 +449,14 @@` both show `<strong>Eli</strong>` → `<strong>You</strong>` with the adjacent `<div class="act-av">E</div>` context line unchanged. Sibling pattern in the same PR: `orbit-github/example.html:541` `title="eli — pending">E` → `title="you — pending">Y`.

Suggested change — apply to both line 349 and line 458 in this file:

```html
<div class="act-av">Y</div>
```

If you'd rather not encode any user initial in mock content, a neutral glyph (e.g. a `·` or a small person SVG) also satisfies the identity guard.


design-templates/orbit-github/example.html
```diff
@@ -538,7 +538,7 @@
                 <div class="reviewer-avatar" style="background:#cf222e;" title="jess — approved">J
                   <span class="reviewer-status approved"><svg viewBox="0 0 16 16"><path d="M13.78 4.22a.75.75 0 0 1 0 1.06l-7.25 7.25a.75.75 0 0 1-1.06 0L2.22 9.28a.751.751 0 0 1 .018-1.042.751.751 0 0 1 1.042-.018L6 10.94l6.72-6.72a.75.75 0 0 1 1.06 0Z"/></svg></span>
                 </div>
-                <div class="reviewer-avatar" style="background:#8250df;" title="eli — pending">E
+                <div class="reviewer-avatar" style="background:#8250df;" title="you — pending">Y
```
lefarcen, line 541: P2 This file still has the signed-in-user header avatar rendered as `E` at line 460 (`<div class="gh-avatar-header">E</div>`), even though this nearby reviewer avatar was correctly changed from `E`/`eli` to `Y`/`you`. That top-right avatar represents the current GitHub user, so leaving `E` keeps the same sample-identity leak shape this PR is removing. Please update the header avatar to `Y` as well, or use a neutral user glyph if we do not want to encode any mock initial.

lefarcen (changes requested):
Hey @YUHAO-corn, thanks for the quick follow-up here. The PR body is much clearer now, and the Gmail + Linear avatar fixes address the stale-initial feedback from the last pass.

I found one remaining identity remnant in the GitHub template before this can land — see the inline note. Once that header avatar is neutralized too, the template set should be consistent with the new identity guard.

lefarcen (approved):
Hey @YUHAO-corn, nice follow-up — the remaining GitHub header avatar is now updated to `Y`, and the previous Gmail/Linear avatar initials are still aligned. I also re-checked the changed Orbit templates for lingering `Eli` / stale `E` user-avatar remnants and don’t see any left.

Clearing my previous change request from the bot side. Product review and the normal human reviewer path still apply before merge, but the identity-leak fix looks consistent from my pass. 🙌

lefarcen:
Hi @qiongyu1999 👋 — this PR was updated since the product review request yesterday. The author addressed the remaining avatar identity remnants from the last review pass. Could you take a quick look at the current head from the product side when you get a chance? Thanks!

alchemistklk (approved):

lefarcen:
Merged! 🎉 Thanks @YUHAO-corn for fixing the Orbit identity leak. The template examples are much cleaner now, and the explicit guidance will prevent future agents from treating mock names as real user identities. Looking forward to your next contribution!

PR #1565, opened 2026-05-13
fix(web): localize accent controls in settings

prantikmedhi:
## Summary
- localize the Settings → Appearance accent color label, swatch group, default swatch, and custom color input
- add a Chinese regression test for the localized copy

## Why
The accent controls in Settings → Appearance still exposed English-only text in an otherwise localized surface. That created a mixed-language experience for translated users and made the controls harder to scan consistently.

## What users will see
- The accent color label, swatch group, default swatch, and custom color input are localized in Settings → Appearance.
- Chinese users see translated copy for the accent controls and matching accessibility labels.
- No behavior changes beyond the copy shown in the UI.

## Surface area
- [x] UI
- [x] i18n keys
- [x] Tests
- [ ] Keyboard shortcut
- [ ] CLI / env var
- [ ] API / contract
- [ ] Extension point
- [ ] New top-level dependency
- [ ] None

- `apps/web/src/components/SettingsDialog.tsx`
- `apps/web/tests/components/SettingsDialog.execution.test.tsx`

## Screenshots
- N/A

## Bug fix verification
- Regression test: `apps/web/tests/components/SettingsDialog.execution.test.tsx` (`localizes the accent color controls in Chinese`)
- Red/green on main: not re-run on `main`; verified on this branch with the new regression test.

## Validation
- `vitest run -c vitest.config.ts tests/components/SettingsDialog.execution.test.tsx -t "localizes the accent color controls in Chinese"`

PR #1623, opened 2026-05-14
fix: Localize Accent color label in Settings

xxiaoxiong:
Fixes #1560

## Problem

When the app language was set to **Chinese** (or any non-English language), the `Accent color` label in Settings → Appearance remained in **English**, creating a **mixed-language experience** in a core settings surface.

### Hardcoded English Strings
- `Accent color` (field label)
- `Accent color` (aria-label)

### Current Behavior
- ❌ Accent color label: English only
- ❌ Mixed-language settings UI
- ❌ Inconsistent with rest of the app

### Expected Behavior
- ✅ Accent color label: Fully localized
- ✅ Consistent language experience
- ✅ All text in the selected language

## Root Cause

The `SettingsDialog.tsx` component contained **hardcoded English strings** instead of using the i18n translation system, even though the translation key `pet.fieldAccent` already existed in all 18 supported languages.

### Updated SettingsDialog Component

**Before:**
```tsx
<span className="field-label">Accent color</span>
<div className="pet-swatches" role="radiogroup" aria-label="Accent color">
```

**After:**
```tsx
<span className="field-label">{t('pet.fieldAccent')}</span>
<div className="pet-swatches" role="radiogroup" aria-label={t('pet.fieldAccent')}>
```

## Testing

- [x] Verified `pet.fieldAccent` exists in all 18 locale files
- [x] Chinese translation: "主题色"
- [x] Applied to both visible label and aria-label for accessibility

## Impact

- **Surface area:** Settings → Appearance
- **Languages affected:** All 18 supported languages
- **Files changed:** 1 (SettingsDialog.tsx)

### DIFF
--- a/apps/web/src/components/SettingsDialog.tsx
+++ b/apps/web/src/components/SettingsDialog.tsx
@@ -4789,8 +4789,8 @@ function AppearanceSection({
         ))}
       </div>
       <div className="field">
-        <span className="field-label">Accent color</span>
-        <div className="pet-swatches" role="radiogroup" aria-label="Accent color">
+        <span className="field-label">{t('pet.fieldAccent')}</span>
+        <div className="pet-swatches" role="radiogroup" aria-label={t('pet.fieldAccent')}>
           {ACCENT_SWATCHES.map((color) => {
             const active = currentAccent === color;
             return (


apps/web/src/components/SettingsDialog.tsx
```diff
@@ -4789,8 +4789,8 @@ function AppearanceSection({
         ))}
       </div>
       <div className="field">
-        <span className="field-label">Accent color</span>
-        <div className="pet-swatches" role="radiogroup" aria-label="Accent color">
+        <span className="field-label">{t('pet.fieldAccent')}</span>
+        <div className="pet-swatches" role="radiogroup" aria-label={t('pet.fieldAccent')}>
```
lefarcen, line 4793: P2 This localizes the group label, but the child controls in this same radiogroup still expose English accessible names: the default swatch uses `Default accent color` and the color input uses `Custom accent color`. Please route those through i18n too (the custom picker can use the existing `pet.fieldAccentCustom`; add/use a locale key for the default swatch) so the whole accent-color control follows the selected app language.

lefarcen (changes requested):
Hey @xxiaoxiong, thanks for picking this up — wiring the visible Settings → Appearance label and radiogroup label through `pet.fieldAccent` is the right direction for #1560.

One related accessibility gap remains in the same accent-color control, so the mixed-language experience is not fully fixed yet for screen-reader users. Please localize that before this lands. While you are touching the PR description, please also fill the template sections that are currently missing: Why, What users will see, Surface area, and Validation.

Once those are covered, this should be a very small and focused bugfix. 🙌

apps/web/src/components/SettingsDialog.tsx
lefarcen, line 4793: P2 This localizes the group label, but the child controls in this same radiogroup still expose English accessible names: the default swatch uses `Default accent color` and the color input uses `Custom accent color`. Please route those through i18n too (the custom picker can use the existing `pet.fieldAccentCustom`; add/use a locale key for the default swatch) so the whole accent-color control follows the selected app language.

lefarcen (changes requested):
(Duplicate of the previous comment 24s earlier — bot trigger fired twice on the same head. Disregard; see the one above. Sorry for the noise.)

lefarcen:
Hey @xxiaoxiong — just a heads-up: the fix for #1560 landed via PR #1565 today (2026-05-15), which covered the full accent-color label + accessibility label localization scope. This PR is now superseded and has merge conflicts against the current `main`.

No further changes are needed here for #1560. If you're interested in follow-up localization work in the same area, a fresh PR off the updated `main` would be the right path. Thanks for jumping in on this one! 🙌

PR #1636, opened 2026-05-14
fix: restore skillIds processing in composeDaemonSystemPrompt

### LINKED ISSUE #1635: @-mention skills not injected into system prompt (regression from PR #955)
# @-mention skills not injected into system prompt (regression from PR #955)

## Problem

The @-mention skill composition feature is completely non-functional. When a user selects a skill via the `@` popover in the chat composer:

1. The frontend correctly sends `skillIds` in the `ChatRequest`
2. The daemon receives `skillIds` but **discards them entirely**
3. The skill content is never injected into the system prompt
4. The LLM has no knowledge of the skill's workflow and generates output independently

## Root Cause

PR #955 (`feat: generic skills + split skills/design-templates + finalize-design API`, commit `b5eb8c16`) introduced a refactor of `composeDaemonSystemPrompt` that added `connectedExternalMcp` to the function signature. During the squash merge conflict resolution, the `skillIds` parameter and its ~30 lines of processing logic were accidentally deleted.

**Timeline:**

```
commit 6c23cf7b — "feat: general-purpose skills with @-mention composition"
  ├── Frontend: ChatComposer.tsx sends skillIds ✅
  ├── Contract: ChatRequest.skillIds defined ✅
  └── Backend: server.ts processes skillIds ✅
         ↓
PR #955 squash merge — commit b5eb8c16
  ├── Frontend: preserved ✅
  ├── Contract: preserved ✅
  └── Backend: skillIds removed during conflict resolution ❌
         ↓
Current state
  ├── Frontend: @ → skill → chip → skillIds sent ✅
  ├── Contract: skillIds?: string[] exists ✅
  └── Backend: daemon receives skillIds but discards ❌
```

## Impact

| Layer | Status | Description |
|-------|--------|-------------|
| UI popover | ✅ | `@` shows skill list |
| Chip display | ✅ | Selected skill shows as removable chip |
| Data transmission | ✅ | `skillIds` sent with `ChatRequest` |
| **Backend processing** | **❌** | **Daemon completely ignores `skillIds`** |

All @-mention functionality is broken across all projects and all tabs.

## Steps to Reproduce

1. Open any project in Open Design
2. In the chat composer, type `@` and select any skill (e.g., `@faq-page`)
3. Send a message (e.g., "fire effect")
4. Observe: the LLM generates output without following the skill's workflow

**Expected:** The skill's SKILL.md content should be injected into the system prompt, and the LLM should follow the skill's workflow.

**Actual:** The LLM ignores the skill and builds from scratch.

## Proposed Fix

Restore the `skillIds` parameter to `composeDaemonSystemPrompt` and the ad-hoc skill processing logic that was deleted in PR #955.

The fix is a single-file change in `apps/daemon/src/server.ts` (+41 lines):

1. Destructure `skillIds` from `chatBody`
2. Add `skillIds` to `composeDaemonSystemPrompt` function signature
3. Restore the ad-hoc skill processing block that concatenates skill bodies into `skillBody`
4. Pass `skillIds` through to the function call

## Verification

Tested with `@faq-page` skill on an "Other" tab project (no project-level skill):

| Condition | Result |
|-----------|--------|
| Without fix | LLM builds FAQ page from scratch, missing category filters, ignores skill workflow |
| With fix | LLM follows faq-page skill workflow, generates complete HTML with search, category filters, accordion, footer CTA |

## Related Files

- `packages/contracts/src/api/chat.ts` — `ChatRequest` interface (skillId + skillIds)
- `apps/web/src/components/ChatComposer.tsx` — @-mention UI
- `apps/daemon/src/server.ts` — `composeDaemonSystemPrompt` (**missing skillIds processing**)

MetaAlms:
Fixes #1635

## Why

The @-mention skill composition feature is completely non-functional. This is a regression introduced during the PR #955 squash merge, where the `skillIds` processing logic was accidentally deleted during conflict resolution.

**Use case**: Users want to compose multiple skills per turn (e.g., `@faq-page` + `@web-search`) without binding the project to one of them. This is a core feature of the skills system.

**Pain**: The frontend correctly sends `skillIds`, but the daemon discards them. Users see the skill chip in the composer, but the skill has zero effect on the output. This affects all @-mention functionality across all projects and tabs.

## What users will see

- **Before fix**: User selects a skill via `@` popover, sees the chip, sends a message — but the LLM ignores the skill entirely and builds from scratch
- **After fix**: The selected skill's workflow is injected into the system prompt, and the LLM follows it exactly (e.g., `@faq-page` generates a complete FAQ page with search, category filters, accordion, and footer CTA)

No UI changes — the fix is entirely in the daemon's prompt composition logic.

## Surface area

- [x] **Extension point** — new entry under `skills/`, `design-systems/`, `design-templates/`, or `craft/`, or change to the skills protocol

## Screenshots

No UI changes — the fix is in daemon prompt composition logic.

## Bug fix verification

- **Test path that reproduces the bug**: Create project with "Other" tab → type `@faq-page fire effect` → send message → observe LLM ignores skill workflow
- **Did the test go red on `main` and green on this branch?** Yes
- **Verification method**: Tested with `@faq-page` skill on "Other" tab project (no project-level skill). Before fix: LLM builds from scratch, missing category filters. After fix: LLM follows skill workflow, generates complete HTML with all required features.

## Validation

- `pnpm guard` — passed
- `pnpm --filter @open-design/daemon typecheck` — passed
- Manual verification: tested `@faq-page` skill on "Other" tab project — without fix, LLM ignores skill workflow; with fix, LLM follows skill workflow and generates complete FAQ page with search, category filters, accordion, and footer CTA

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -4,7 +4,7 @@ import express from 'express';
 import multer from 'multer';
 import JSZip from 'jszip';
 import { execFile, spawn } from 'node:child_process';
-import { randomUUID } from 'node:crypto';
+import { createHash, createHmac, randomUUID, timingSafeEqual } from 'node:crypto';
 import { createRequire } from 'node:module';
 import { fileURLToPath } from 'node:url';
 import path from 'node:path';
@@ -56,7 +56,7 @@ export {
   signDesktopImportToken,
   verifyDesktopImportToken,
 } from './desktop-auth.js';
-import { findSkillById, listSkills, splitDerivedSkillId } from './skills.js';
+import { findSkillById, listSkills, splitDerivedSkillId, withSkillRootPreamble, type SkillInfo } from './skills.js';
 import { validateLinkedDirs } from './linked-dirs.js';
 import { installFromTarget, uninstallById, sanitizeRepoName } from './library-install.js';
 import { buildWindowsFolderDialogCommand, parseFolderDialogStdout } from './native-folder-dialog.js';
@@ -623,11 +623,93 @@ export function validateCodexGeneratedImagesDir(
   }
 }
 
+/**
+ * Resolve ad-hoc (@-mention) skill ids to their on-disk directories.
+ *
+ * Returns an array of `{ id, dir }` pairs, one per unique resolved skill,
+ * deduplicated by resolved directory path so that aliases pointing to the
+ * same folder do not double-stage.
+ *
+ * Exported so daemon tests can exercise the full
+ * skillIds → resolved dirs chain without spinning up an HTTP server.
+ */
+export function collectAdHocSkillDirs(
+  skillIds: unknown,
+  effectiveSkillId: string | null | undefined,
+  allSkills: SkillInfo[],
+): Array<{ id: string; dir: string }> {
+  const ids = Array.isArray(skillIds)
+    ? skillIds
+        .map((s: unknown) => (typeof s === 'string' ? s.trim() : ''))
+        .filter(Boolean)
+        .filter((id: string) => id !== effectiveSkillId)
+    : [];
+  if (ids.length === 0) return [];
+
+  const seenId = new Set(effectiveSkillId ? [effectiveSkillId] : []);
+  // Seed seenDir with the active skill's resolved dir so an alias in
+  // skillIds that resolves to the same directory is also excluded.
+  const seenDir = new Set<string>();
+  if (effectiveSkillId) {
+    const active = findSkillById(allSkills, effectiveSkillId);
+    if (active?.dir) seenDir.add(active.dir);
+  }
+  const result: Array<{ id: string; dir: string }> = [];
+
+  for (const id of ids) {
+    if (seenId.has(id)) continue;
+    seenId.add(id);
+    const skill = findSkillById(allSkills, id);
+    if (!skill?.dir) continue;
+    if (seenDir.has(skill.dir)) continue;
+    seenDir.add(skill.dir);
+    result.push({ id: skill.id, dir: skill.dir });
+  }
+
+  return result;
+}
+
+/**
+ * Assign collision-safe staged aliases to resolved ad-hoc skills.
+ *
+ * Every ad-hoc skill receives a deterministic alias of the form
+ * `<basename>-<hash>` where hash is the first 7 hex chars of the MD5 of the
+ * full dir path. This guarantees:
+ *   1. The same folder always gets the same alias regardless of input order.
+ *   2. Two skills with the same basename but different dirs never collide.
+ *   3. Ad-hoc skills never overwrite the project-level active skill copy.
+ */
+export function resolveAdHocSkillsWithAliases(
+  resolved: Array<{ id: string; dir: string }>,
+): Array<{ id: string; dir: string; alias: string }> {
+  return resolved.map(({ id, dir }) => {
+    const basename = path.basename(dir);
+    // Deterministic hash suffix so the same folder always gets the same alias
+    // regardless of input order or concurrent turns.
+    const hash = createHash('md5').update(dir).digest('hex').slice(0, 7);
+    const alias = `${basename}-${hash}`;
+    return { id, dir, alias };
+  });
+}
+
+/**
+ * Derive `skillMode` from a single ad-hoc @-mention skill when no project
+ * skill is active. Returns `undefined` for multi-skill or empty cases.
+ */
+export function resolveAdHocSkillMode(
+  resolved: Array<{ id: string; dir: string }>,
+  allSkills: SkillInfo[],
+): string | undefined {
+  if (resolved.length !== 1) return undefined;
+  return findSkillById(allSkills, resolved[0].id)?.mode;
+}
+
 export function resolveChatExtraAllowedDirs({
   agentId,
   skillsDir,
   designSystemsDir,
   linkedDirs = [],
+  adHocSkillDirs = [],
   codexGeneratedImagesDir,
   existsSync = fs.existsSync,
 }: {
@@ -635,17 +717,26 @@ export function resolveChatExtraAllowedDirs({
   skillsDir?: string | null;
   designSystemsDir?: string | null;
   linkedDirs?: Array<string | null | undefined>;
+  /**
+   * Ad-hoc (@-mention) skill directories. Always reachable regardless of
+   * agentId because the user explicitly summoned the skill for this turn;
+   * the Codex branch below would otherwise drop them along with the rest
+   * of `linkedDirs`, leaving user-installed ad-hoc skill side files
+   * unreadable on projectless Codex runs.
+   */
+  adHocSkillDirs?: Array<string | null | undefined>;
   codexGeneratedImagesDir?: string | null;
   existsSync?: (path: string) => boolean;
 }): string[] {
   const isCodex =
     typeof agentId === 'string' && agentId.trim().toLowerCase() === 'codex';
   const candidates = isCodex
-    ? [codexGeneratedImagesDir]
+    ? [codexGeneratedImagesDir, ...(Array.isArray(adHocSkillDirs) ? adHocSkillDirs : [])]
     : [
         skillsDir,
         designSystemsDir,
         ...(Array.isArray(linkedDirs) ? linkedDirs : []),
+        ...(Array.isArray(adHocSkillDirs) ? adHocSkillDirs : []),
       ];
   return Array.from(
     new Set(
@@ -4067,7 +4158,7 @@ export async function startServer({
       // Strip full body + on-disk dir from the listing — frontend fetches the
       // body via /api/skills/:id when needed (keeps the listing payload small).
       res.json({
-        skills: skills.map(({ body, dir: _dir, ...rest }) => ({
+        skills: skills.map(({ body, dir: _dir, rawBody: _rawBody, ...rest }) => ({
           ...rest,
           hasBody: typeof body === 'string' && body.length > 0,
         })),
@@ -4082,7 +4173,7 @@ export async function startServer({
       const skills = await listSkills(SKILLS_DIR);
       const skill = findSkillById(skills, req.params.id);
       if (!skill) return res.status(404).json({ error: 'skill not found' });
-      const { dir: _dir, ...serializable } = skill;
+      const { dir: _dir, rawBody: _rawBody, ...serializable } = skill;
       res.json(serializable);
     } catch (err) {
       res.status(500).json({ error: String(err) });
@@ -7580,6 +7671,7 @@ export async function startServer({
     agentId,
     projectId,
     skillId,
+    skillIds,
     designSystemId,
     streamFormat,
     connectedExternalMcp,
@@ -7654,6 +7746,48 @@ export async function startServer({
         console.warn(
           `[plugins] pluginSkillBody load failed: ${err?.message ?? err}`,
         );
+      }
+    }
+
+    // Per-turn skills picked via the composer's @-mention popover. They
+    // never persist on the project — we just append their bodies after the
+    // primary skill so the agent sees one combined block this turn.
+    let adHocSkillDirs: Array<{ id: string; dir: string; alias: string }> = [];
+    if (Array.isArray(skillIds) && skillIds.length > 0) {
+      const allSkills = await listAllSkillLikeEntries();
+      const resolved = collectAdHocSkillDirs(skillIds, effectiveSkillId, allSkills);
+      adHocSkillDirs = resolveAdHocSkillsWithAliases(resolved);
+
+      const blocks = [];
+      const baseBody = skillBody && skillBody.trim().length > 0 ? skillBody : '';
+      for (const { id, dir, alias } of adHocSkillDirs) {
+        const extra = findSkillById(allSkills, id);
+        if (!extra) continue;
+        if (Array.isArray(extra.craftRequires)) {
+          for (const c of extra.craftRequires) {
+            if (!skillCraftRequires.includes(c)) skillCraftRequires.push(c);
+          }
+        }
+        const bodyWithAlias = withSkillRootPreamble(
+          extra.rawBody ?? extra.body,
+          dir,
+          alias,
+        );
+        blocks.push(
+          `\n\n---\n\n## Composed skill — ${extra.name || id}\n\n${bodyWithAlias.trim()}`
+        );
+      }
+      if (blocks.length > 0) {
+        skillBody = baseBody + blocks.join('');
+        if (!skillName) {
+          skillName = adHocSkillDirs.length === 1
+            ? findSkillById(allSkills, adHocSkillDirs[0].id)?.name ?? null
+            : 'composed';
+        }
+        // When no project skill is active, propagate the single ad-hoc skill's mode
+        if (!skillMode) {
+          skillMode = resolveAdHocSkillMode(adHocSkillDirs, allSkills);
+        }
       }
     }
 
@@ -7918,7 +8052,7 @@ export async function startServer({
     // `listSkills()` scan in `startChatRun`. critiqueShouldRun threads
     // the same panel-eligibility decision down to the spawn-path
     // orchestrator gate so prompt and orchestrator stay in lockstep.
-    return { prompt, activeSkillDir, critiqueShouldRun };
+    return { prompt, activeSkillDir, adHocSkillDirs, critiqueShouldRun };
   };
 
   // Plan §3.I1 / §3.D / spec §10.1: fire the pipeline schedule on a
@@ -8008,6 +8142,7 @@ export async function startServer({
       assistantMessageId,
       clientRequestId,
       skillId,
+      skillIds,
       designSystemId,
       attachments = [],
       commentAttachments = [],
@@ -8252,11 +8387,12 @@ export async function startServer({
       .filter((s) => typeof oauthTokensForSpawn[s.id] === 'string')
       .map((s) => ({ id: s.id, label: s.label }));
 
-    const { prompt: daemonSystemPrompt, activeSkillDir, critiqueShouldRun } =
+    const { prompt: daemonSystemPrompt, activeSkillDir, adHocSkillDirs, critiqueShouldRun } =
       await composeDaemonSystemPrompt({
         agentId,
         projectId,
         skillId,
+        skillIds,
         designSystemId,
         streamFormat: def?.streamFormat ?? 'plain',
         connectedExternalMcp,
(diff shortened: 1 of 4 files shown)


lefarcen (commented):
Hi @MetaAlms — thanks for reshaping the body into the new template after my earlier note. The Why / What users will see / Validation / Bug fix verification sections all land cleanly now, and the daemon-side scope reads well from the user-facing description.

Looks ready for pool review on the body side.

apps/daemon/src/server.ts
```diff
@@ -3153,6 +3154,44 @@ export async function startServer({
...
+    // never persist on the project — we just append their bodies after the
+    // primary skill so the agent sees one combined block this turn.
+    const adHocSkillIds = Array.isArray(skillIds)
+      ? skillIds
+          .map((s: unknown) => (typeof s === 'string' ? s.trim() : ''))
+          .filter(Boolean)
+          .filter((id: string) => id !== effectiveSkillId)
+      : [];
+    if (adHocSkillIds.length > 0) {
+      const allSkills = await listAllSkillLikeEntries();
```
nettee, line 7757: Appending each ad-hoc skill body here restores the text prompt, but it does not restore the file reachability those skill bodies assume. Skills that ship `assets/`, `references/`, or `example.html` already advertise `.od-skills/<folder>/...` and absolute fallback paths in their preamble, yet the spawn path still stages only `activeSkillDir` (`apps/daemon/src/server.ts:3692`) and `resolveChatExtraAllowedDirs()` only allowlists the built-in `SKILLS_DIR` for non-Codex agents (`apps/daemon/src/server.ts:638-669, 3725-3731`). That leaves composed side-file skills broken on Codex, and user-installed side-file skills broken on other agents when the model follows the advertised absolute path. Please return all resolved ad-hoc skill dirs from prompt composition, stage/allowlist each unique dir before spawn, and add a daemon test that composes a side-file skill through `skillIds`.

nettee (changes requested):
Found one blocking issue in the daemon-side @-mention restore: extra skills with side files are still not reachable at run time.


MetaAlms:
Thanks! Updated the PR body with all template sections. Ready for review when you have a minute 🙏

lefarcen:
Heads-up: PR #1639 by @xxiaoxiong is also open against the same area — both PRs touch `apps/daemon/src/server.ts` and look to be restoring the `skillIds` passthrough that PR #955's squash dropped. You and @xxiaoxiong might want to compare approaches; the maintainer team will pick which lands. Sharing so neither effort gets wasted.

apps/daemon/src/server.ts
MetaAlms: Addressed. The fix now collects resolved ad-hoc skill dirs via `collectAdHocSkillDirs()`, stages each one before spawn (using `path.basename(dir)` to match `withSkillRootPreamble()`), and appends them to `linkedDirs` for `resolveChatExtraAllowedDirs()`. Added daemon tests covering the full `skillIds → resolved dirs` chain including an alias-to-same-dir mock scenario.

lefarcen:
Thanks for the quick turnaround! 👀 The side-file skill dirs fix + test coverage look substantive. @nettee will re-review the current head when available.

lefarcen:
@nettee — final head is up (`c554765b`). Both commits are in: the original `skillIds` restore + the side-file staging you flagged. The fix:

- Extracts `collectAdHocSkillDirs()` which resolves each @-mention to `{ id, dir }`, deduplicating by both id and resolved path (alias-safe)
- Calls `stageActiveSkill(cwd, path.basename(dir), dir, ...)` per ad-hoc skill so `assets/` / `references/` are CWD-accessible and match the path `withSkillRootPreamble()` advertises to the agent
- Appends ad-hoc dirs to `linkedDirs` for `resolveChatExtraAllowedDirs()`
- 277-line test file covers the full `skillIds → resolved dirs` chain

Ready for your re-review when you have a moment!

apps/daemon/src/server.ts
```diff
@@ -3148,6 +3195,39 @@ export async function startServer({
...
+        blocks.push(
+          `\n\n---\n\n## Composed skill — ${extra.name || id}\n\n${(extra.body || '').trim()}`
+        );
+      }
+      if (blocks.length > 0) {
+        skillBody = baseBody + blocks.join('');
+        if (!skillName) {
+          skillName = resolved.length === 1
+            ? findSkillById(allSkills, resolved[0].id)?.name ?? null
+            : 'composed';
```
nettee, line 7785: When there is no `effectiveSkillId`, this new branch only backfills `skillName`; `skillMode` stays unset even though the downstream prompt path and critique/media gating still key off it (`composeSystemPrompt(... skillMode)` at `apps/daemon/src/server.ts:3400-3405` and `isMediaSurface` at `apps/daemon/src/server.ts:3382-3388`). That means a turn that uses only an ad-hoc `@` skill behaves differently from the same skill bound at project level: a deck skill loses the deck framework injection, and image/video/audio skills skip the media-surface handling entirely. Please derive and carry over the resolved ad-hoc skill's mode here before composing the prompt (for the single-skill case, `resolved[0]` already gives you the source skill), and add a regression test for `no project skill + one ad-hoc skill` so this path stays covered.

nettee (changes requested):
One blocking issue remains in the ad-hoc `skillIds` restore: per-turn skills still lose mode-derived behavior when no project skill is active.


apps/daemon/src/server.ts
MetaAlms: Fixed in 9c5a328d.

The issue was that `skillMode` was never derived from ad-hoc skills when no project skill was active. Here's what changed:

**Code**: Extracted `resolveAdHocSkillMode(resolved, allSkills)` — a pure function that returns the mode for a single resolved ad-hoc skill, or `undefined` for multi-skill/empty cases. Called from the `if (!skillMode)` guard in the skillIds block so that `composeSystemPrompt` receives the correct mode (deck framework injection, `isMediaSurface`, etc.).

**Tests**: 3 cases in `skillids-side-files.test.ts` covering the exact regression scenario — single ad-hoc skill returns mode, multiple skills return undefined, empty list returns undefined.

The image/video/audio surface handling and deck framework injection now work correctly for ad-hoc @-mention skills without a project-level skill.

lefarcen:
@nettee — new head is `9c5a328d`. @MetaAlms addressed the `skillMode` gap: extracted `resolveAdHocSkillMode(resolved, allSkills)` that derives mode from `resolved[0]` for the single-skill case, guarded by `if (!skillMode)` so it only fires when no project skill is active. Three regression tests cover single/multi/empty cases. Ready for your look!

apps/daemon/src/server.ts
```diff
@@ -3660,6 +3758,28 @@ export async function startServer({
...
+    // paths, just like the project-level activeSkillDir above. The folder
+    // name MUST be path.basename(dir) to match the preamble emitted by
+    // withSkillRootPreamble() in skills.ts, which advertises
+    // `.od-skills/${path.basename(dir)}/...` to the agent.
+    if (cwd && adHocSkillDirs.length > 0) {
+      for (const { dir } of adHocSkillDirs) {
+        if (dir === activeSkillDir) continue;
+        const result = await stageActiveSkill(
+          cwd,
+          path.basename(dir),
```
nettee: Blocking: staging by `path.basename(dir)` can collapse two different ad-hoc skills onto the same `.od-skills/<folder>/` slot. `collectAdHocSkillDirs()` only dedupes by full resolved dir (`apps/daemon/src/server.ts:666-683`), but `listAllSkillLikeEntries()` spans multiple roots (`apps/daemon/src/server.ts:1119-1126`) and `withSkillRootPreamble()` also advertises the basename-based alias (`apps/daemon/src/skills.ts:397-432`). If two selected skills live in different roots but share a folder basename, the later `stageActiveSkill()` call overwrites the earlier one and both prompt preambles now point at the same relative path. That can make the agent open the wrong `assets/` or `references/` files for one of the composed skills. Please carry a collision-safe alias per resolved skill (for example the canonical skill id, or `basename + stable hash`) and use that alias consistently in both the preamble and the staging path; add a regression with two temp roots that share a folder basename.

apps/daemon/src/server.ts
```diff
@@ -3684,7 +3804,7 @@ export async function startServer({
       agentId,
       skillsDir: SKILLS_DIR,
       designSystemsDir: DESIGN_SYSTEMS_DIR,
-      linkedDirs,
+      linkedDirs: [...linkedDirs, ...adHocSkillDirs.map((r) => r.dir)],
```
nettee: Blocking: appending `adHocSkillDirs` here still leaves projectless Codex runs unable to read user-installed ad-hoc skill side files. When `projectId` is absent or `ensureProject()` fails, `cwd` stays `null` (`apps/daemon/src/server.ts:3549-3565`), so the staging loop above is skipped and the run falls back to `PROJECT_ROOT` (`apps/daemon/src/server.ts:3790`). `resolveChatExtraAllowedDirs()` drops every linked dir for Codex (`apps/daemon/src/server.ts:716-724`), so a skill under `OD_DATA_DIR/skills` or `OD_DATA_DIR/design-templates` is still outside both the cwd and the allowlist even though its preamble advertises an absolute fallback path. That means this branch still leaves ad-hoc user skills broken on Codex for no-project turns. Please either create a run-scoped cwd/temp staging path for this case, or fail fast before spawn when a selected ad-hoc skill dir would be unreachable.

nettee (changes requested):
Two blocking gaps remain in the ad-hoc skill side-file path: basename-based staging can alias different skills onto the same cwd slot, and projectless Codex runs still cannot reach user-installed ad-hoc skill side files.


lefarcen:
Hey @MetaAlms — pushed a follow-up commit (`bbefda23`) addressing the two remaining blockers in the latest review. Both claims held up against the source, so I went ahead with the fixes; happy to revert or rework if you'd rather take a different shape.

**1. Basename collision in the stage loop**

`collectAdHocSkillDirs()` dedupes by resolved dir, but two ad-hoc skills from different roots (bundled `skills/my-skill` and a user `.od/skills/my-skill`) resolve to distinct dirs while sharing a basename. The stage loop keyed `.od-skills/<folder>/` on `path.basename(dir)`, so a second `stageActiveSkill(cwd, folder, dir)` call with the same `folder` would overwrite the first skill's side files mid-run with no signal.

Tracked `seenBasenames` inside the stage loop (seeded with the active skill's folder so an ad-hoc that aliases it is also caught) and skip + warn on collision. The skipped skill is still reachable via the absolute paths in the preamble — only its `.od-skills/<folder>/` cwd-relative alias is dropped. The warning gives ops/support something to grep when a `@`-mentioned skill's side files appear missing.

**2. Codex + projectless runs lose ad-hoc skill dirs**

`resolveChatExtraAllowedDirs()`'s Codex branch deliberately returns only `[codexGeneratedImagesDir]`, dropping `linkedDirs` (along with `skillsDir`/`designSystemsDir`). The previous fix appended ad-hoc dirs onto `linkedDirs`, so they were dropped with the rest on any Codex run. Combined with the projectless case (where `cwd` is null, the cwd-relative stage loop is skipped, and the agent must read side files via absolute paths under `extraAllowedDirs`), this left user-installed ad-hoc skill side files unreadable on projectless Codex runs.

Added a dedicated `adHocSkillDirs` parameter to `resolveChatExtraAllowedDirs()` that is included in both the Codex and non-Codex candidate sets. Rationale: the user explicitly summoned that skill via `@`-mention this turn, so the Codex narrow-grant rule shouldn't apply to it — the existing Codex restriction on project-metadata `linkedDirs` stays intact for non-mentioned content. Updated the call site to route ad-hoc dirs through the new channel instead of merging them into `linkedDirs`.

**Tests**

Added regression coverage in `apps/daemon/tests/skillids-side-files.test.ts`:

- `resolveChatExtraAllowedDirs ad-hoc skill bypass` — Codex run drops `skillsDir`/`linkedDirs` but keeps ad-hoc dirs; non-Codex keeps everything; ad-hoc dirs that overlap `linkedDirs` are deduped.
- `collectAdHocSkillDirs basename collisions` — builds two skill roots with the same `image-gen` basename, confirms both resolve to distinct dirs that share a basename so the stage-loop guard has a real shape to defend against.

**Validation**

`pnpm --filter @open-design/daemon typecheck` — clean. `pnpm --filter @open-design/daemon exec vitest run tests/skillids-side-files.test.ts tests/chat-route.test.ts` — 42/42 pass.

MetaAlms:
@nettee Both blocking gaps from your latest review have been addressed in `bbefda23`:

1. **Basename collisions**: Added `seenBasenames` guard in the ad-hoc staging loop — skips + warns on duplicate folders, falling back to absolute paths from the preamble.
2. **Codex + projectless runs**: Added dedicated `adHocSkillDirs` parameter to `resolveChatExtraAllowedDirs()` so explicitly @-mentioned skills bypass the Codex `linkedDirs` drop and remain reachable via absolute paths when `cwd` is null.

Regression tests added in `skillids-side-files.test.ts` covering both cases. Could you take another look and approve when ready?

@lefarcen thanks again for pushing the fix directly — lmk if anything else is needed on my end.

nettee:
@MetaAlms I'm holding off on generating review comments for nexu-io/open-design#1636 because this pull request has merge conflicts right now.

apps/daemon/src/server.ts
```diff
@@ -3660,6 +3768,47 @@ export async function startServer({
...
+    if (cwd && adHocSkillDirs.length > 0) {
+      const seenBasenames = new Set<string>();
+      if (activeSkillDir) seenBasenames.add(path.basename(activeSkillDir));
+      for (const { dir, id } of adHocSkillDirs) {
+        if (dir === activeSkillDir) continue;
+        const folder = path.basename(dir);
+        if (seenBasenames.has(folder)) {
+          console.warn(
+            `[od] ad-hoc skill-stage skipped: \`@${id}\` collides with another staged skill at .od-skills/${folder}/; falling back to absolute paths`,
+          );
```
nettee: Skipping the later skill on a basename collision does not actually make the composed prompt safe. `composeDaemonSystemPrompt()` still appends that skill body unchanged above (`blocks.push(...)` at lines 3239-3250), and `withSkillRootPreamble()` always tells the model to prefer `.od-skills/${path.basename(dir)}/...` for side files. In a collision case both selected skills therefore still advertise the same relative folder, but only the first one is staged there; the second skill's preferred relative path now resolves to the first skill's assets/references instead of its own. That can make a composed run read the wrong side files for one of the user-selected skills, so the basename-collision blocker is still present. Please give each resolved skill a collision-safe staged alias and use that alias consistently in both the preamble and `stageActiveSkill`, or fail the run before spawn when you cannot assign unique aliases.

nettee (changes requested):
One blocking gap remains in the basename-collision handling for ad-hoc skill side files.


apps/daemon/src/server.ts
```diff
@@ -695,29 +695,127 @@ export function validateCodexGeneratedImagesDir(
...
+  if (activeSkillDir) seenAliases.add(path.basename(activeSkillDir));
+
+  return resolved.map(({ id, dir }) => {
+    let alias = path.basename(dir);
+    if (seenAliases.has(alias)) {
+      let suffix = 2;
+      while (seenAliases.has(`${alias}-${suffix}`)) {
+        suffix++;
+      }
+      alias = `${alias}-${suffix}`;
```
nettee: Blocking: this alias assignment is only collision-safe within one run; it is not stable across runs for the same project. `stageActiveSkill()` reuses fixed `.od-skills/<alias>/` paths under the shared project cwd and replaces them wholesale on every turn (`apps/daemon/src/cwd-aliases.ts:54-57,86-128`). If one run stages `/built-in/foo` then `/user/foo` and a later run mentions the same two dirs in the opposite order, this helper flips which source owns `foo` vs `foo-2` while the earlier run still has its preamble paths baked from the first ordering. That can make an in-flight run open the other skill's `assets/` / `references/` files instead of its own. Please make the alias deterministic per resolved skill (for example `basename + stable short hash` or `basename + canonical id`) instead of "first collision gets `-2`", and add a regression that feeds the same colliding dirs in reversed order and asserts the aliases stay pinned to the same source dirs.

apps/daemon/src/skills.ts
```diff
@@ -234,6 +238,7 @@ export async function listSkills(
           aggregatesExamples,
           critiquePolicy: normalizeCritiquePolicy(data.od?.critique?.policy),
           body: parentBody,
+          rawBody,
```
nettee, line 241: Non-blocking: wiring `rawBody` onto every `SkillInfo` here now leaks the full SKILL.md content back into `/api/skills`. Both list routes still serialize `skills.map(({ body, dir: _dir, ...rest }) => ({ ...rest, hasBody }))` (`apps/daemon/src/server.ts:4228-4235` and `apps/daemon/src/static-resource-routes.ts:64-71`), so `rawBody` now rides along in `...rest` even though those endpoints are supposed to strip full bodies to keep the payload small. Please keep `rawBody` internal-only for prompt composition or explicitly omit it from the list responses.

nettee (changes requested):
One blocking correctness issue remains in the alias rewrite path, plus one follow-up payload regression in the new `rawBody` plumbing.


MetaAlms:
@nettee Conflicts resolved (rebased onto `main`) and all blocking issues addressed in `622348cc`. Here's the delta since your last review:

**1. Alias determinism (blocking correctness)**
Replaced the order-sensitive `-2`/`-3` suffix allocation with a deterministic hash scheme:
- `resolveAdHocSkillsWithAliases()` now assigns `<basename>-<md5(dir).slice(0,7)>`
- Same folder → same alias regardless of input order or concurrent turns
- Active-skill seeding removed (no longer needed; hash guarantees isolation)

**2. rawBody payload regression**
Explicitly excluded `rawBody` from all 6 API serialization paths:
- `GET /api/skills` (list + detail)
- `GET /api/design-templates` (list + detail)
- `POST /api/skills/import`
- `PUT /api/skills/:id`

**3. Empty-body double-preamble**
Fixed `extra.rawBody || extra.body` → `extra.rawBody ?? extra.body` so an empty-string body doesn't fall back to the already-preambled `extra.body`.

**4. Dead parameter cleanup**
Removed unused `_activeSkillDir` parameter from `resolveAdHocSkillsWithAliases()` and updated call site + tests.

**Validation**
- `pnpm guard` — passed
- `pnpm typecheck` — 0 errors
- `pnpm --filter @open-design/daemon exec vitest run tests/skillids-side-files.test.ts` — 26/26 passed
- Full daemon test suite — only pre-existing Claude-auth / origin-validation failures (unrelated)

Ready for re-review.

qiongyu1999:
@mrcfps Product direction looks good from my side. 已经 a 了；冲突 / CI / code review blockers 解决后可以合并。

MetaAlms:
@nettee All blocking issues from the review rounds have been addressed. Force-pushed after rebasing onto latest `main` (now at `7ebd8155`) — no functional changes since the previous head, just conflict resolution.

**Round 1 — side files not reachable**
- `collectAdHocSkillDirs()` resolves `@`-mention skill IDs to `{ id, dir }` pairs
- Ad-hoc skills are staged via `stageActiveSkill(cwd, alias, dir)` so `assets/` / `references/` are CWD-accessible

**Round 2 — mode-derived behavior lost**
- `resolveAdHocSkillMode()` derives `skillMode` from the single ad-hoc skill when no project skill is active

**Round 3/4 — basename collision + Codex/projectless**
- `resolveAdHocSkillsWithAliases()` assigns collision-safe aliases (`<basename>-<md5(dir).slice(0,7)>`)
- `resolveChatExtraAllowedDirs()` receives ad-hoc dirs via dedicated `adHocSkillDirs` parameter so they survive the Codex narrow-grant branch

**Round 5/6 — alias determinism + rawBody + dead parameter**
- Hash suffix is deterministic per dir regardless of input order
- `extra.rawBody ?? extra.body` avoids double-preamble on empty string
- `rawBody` excluded from all 6 API serialization paths
- Removed unused `_activeSkillDir` parameter from `resolveAdHocSkillsWithAliases()`

**Validation**
- `pnpm guard` ✅
- `pnpm typecheck` ✅
- All CI checks green ✅

Ready for re-review. cc @lefarcen @qiongyu1999

lefarcen:
Heads-up: PR #2552 is now open against the same #1635 regression. Both PRs touch `apps/daemon/src/server.ts` and restore daemon-side handling of `skillIds` for @-mention skill composition; #2552 includes chat-route regressions for prompt injection, deck mode, and alias dedupe.

Sharing so the two approaches can be compared and the maintainer team can pick the one to land.

MetaAlms:
I see that #2552 has been merged and the regression #1635 is resolved. Thanks @nettee and @lefarcen for the thorough reviews on this PR. Closing this now as it's no longer needed. Looking forward to contributing more to the project!

PR #1643, opened 2026-05-14
feat(tweaks): bind toolbar toggle to artifact panel

portseif:
## Why

The viewer toolbar has a **Tweaks** toggle but it wasn't bound to anything — clicking it did nothing for any artifact. Agent-generated artifacts that mount a runtime control panel (e.g. the `.twk-panel` React component pattern emerging in projects in the wild) had no way to be driven from the host. Same for artifacts emitted by the `design-templates/tweaks/` template (`.tw-panel`). The toolbar gave no signal about which artifacts even supported tweaks.

Hit this myself testing the toggle on a real artifact and finding the click was a no-op. Wanted the toolbar to be the canonical surface for opening / closing the panel regardless of which protocol the artifact speaks, and to gracefully grey out when the artifact has no panel.

## What users will see

- On artifacts that ship a tweaks panel: the **Tweaks** toolbar button is enabled. Clicking it opens the panel; clicking it again closes it. Closing the panel locally (`×` inside the panel, or for skill-template artifacts the `T` keyboard shortcut) flips the toolbar toggle back to off — they stay in sync.
- On artifacts with no panel: the **Tweaks** button is disabled with a tooltip that explains why (`No tweaks panel in this artifact`, localized to all 13 supported locales).
- Open/close animates with the artifact's own CSS transition rather than snapping (no more flash).
- Switching between artifacts resets toolbar state correctly — a previous artifact's "tweaks on" state doesn't bleed into the next file.

CleanShot 2026-05-14 at 02 37 38

## Surface area

- [x] **UI** — toolbar toggle behavior in `FileViewer.tsx` (enabled/disabled state + open/close)
- [ ] **Keyboard shortcut**
- [ ] **CLI / env var**
- [x] **API / contract** — documents the host↔iframe postMessage protocol (`__edit_mode_*` for dynamic panels, `od:tweaks-*` for the bridge)
- [x] **Extension point** — new "Host integration contract" section in `design-templates/tweaks/SKILL.md` so future agent-generated tweaks artifacts conform
- [x] **i18n keys** — added `fileViewer.tweaksUnavailable` to all 13 locales
- [ ] **None**

## Screenshots

Tested locally; happy to attach a short GIF of toolbar ↔ panel sync if reviewers want one. The flow is: button greys out → load tweaks artifact → button enables → click → panel slides in → click `×` → button flips off.

## Validation

- [x] `pnpm --filter @open-design/web typecheck` clean
- [x] `pnpm build` clean
- [x] Manually: opened an artifact built with the tweaks skill template (`.tw-panel`) — toggle enabled; click slides panel in/out with transition; `×` and `T` inside the panel sync the toolbar state
- [x] Manually: opened an artifact using the `__edit_mode_*` protocol (`.twk-panel`) — toggle enables on mount; click activates/deactivates the panel; `×` inside the panel flips toolbar off
- [x] Manually: opened a plain HTML artifact with no panel — toggle disabled with explanatory tooltip
- [x] Manually: switched between artifact types — toggle availability updates correctly
- [ ] `pnpm test` — one unrelated pre-existing failure in `apps/web/src/i18n/content.test.ts` (untracked `design-systems/foldar/` directory not represented in German content list, not introduced by this PR)


lefarcen (commented):
Hi @portseif — looking again, the PR body now has the full new-template structure filled in nicely (Why, What users will see, Surface area with UI / API contract / Extension point / i18n keys ticked, Validation with both protocols tested), and the host↔iframe `postMessage` + `od:tweaks-*` bridge contract is well-documented for future tweaks artifacts. Thanks for the iteration.

Since **UI** is ticked, a quick screenshot or GIF of the toolbar↔panel sync would seal the deal for pool review — you already offered one in the description, so feel free to attach when convenient. Otherwise the template's covered.

Siri-Ray:
@portseif I'm holding off on generating review comments for nexu-io/open-design#1643 because this pull request has merge conflicts right now.

portseif:
Thanks both — done on both fronts.

**Template reshape:** Body now follows `pull_request_template.md` — Why / What users will see / Surface area (UI + API/contract + Extension point + i18n keys ticked) / Screenshots / Validation. The original Test plan content is now the Validation checklist.

**Conflicts:** Rebased onto `origin/main` (852a005) and force-pushed. PR now reports `MERGEABLE`. The rebase touched two areas that drifted:

- `apps/web/src/runtime/srcdoc.ts` — adapted to the new `SrcdocOptions` type and the longer `withSelection → withPalette → withEdit → withTweaks → injectSnapshotBridge` chain. The always-injected tweaks bridge slots in just before the snapshot bridge.
- `apps/web/src/components/FileViewer.tsx` — the standalone `handlePreviewFrameLoad` I added is gone; tweaks postMessage now lives inside `syncBridgeModes` so both the url-load and srcdoc iframe `onLoad` handlers carry it. Toolbar toggle button is preserved as-is. Dead `TweaksState` / `DEFAULT_TWEAKS_STATE` removed (never used).
- `SKILL.md` auto-tracked as a rename from `skills/tweaks/` to `design-templates/tweaks/` (upstream moved the directory) — body references updated to match.

Ready for review.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -70,7 +70,12 @@ export function buildSrcdoc(
     ? injectPaletteBridge(withSelection, { initialPalette: options.initialPalette ?? null })
     : withSelection;
   const withEdit = options.editBridge ? injectManualEditBridge(withPalette) : withPalette;
-  return injectSnapshotBridge(withEdit);
+  // The tweaks bridge is always injected — it's a passive listener that
+  // toggles a `.tw-panel`'s visibility in response to host postMessage. Tying
+  // it to a per-call option would force iframe srcdoc regeneration (and a
+  // visible flash) every time the host toggle flips.
+  const withTweaks = injectTweaksBridge(withEdit);
```
Siri-Ray, line 79: This bridge only runs when `HtmlViewer` uses `buildSrcdoc`, but the default preview path for plain HTML still URL-loads the iframe: `FileViewer.tsx` calls `shouldUrlLoadHtmlPreview` without any tweaks-related disqualifier, and `apps/web/tests/components/file-viewer-render-mode.test.ts` currently asserts that a plain HTML preview URL-loads by default. In that URL-load iframe, `injectTweaksBridge` is never inserted, and the class-based tweaks template does not post `od:tweaks-available` itself. The result is that `.tw-panel` artifacts, one of the two protocols this PR says it supports, can leave the new toolbar button disabled with `fileViewer.tweaksUnavailable` unless some unrelated mode such as palette/inspect happens to force srcDoc first. Please make tweaks-template detection participate in the render-mode decision, for example by detecting the fixed `.tw-panel`/`.tw-hidden` template selectors in `source` and passing a `tweaksBridge` flag into `shouldUrlLoadHtmlPreview` so the initial load uses srcDoc, or add an equivalent URL-load bootstrap that can emit `od:tweaks-available` before the button is evaluated. A focused render-mode or `FileViewer` test for a plain `.tw-panel` artifact would cover this path.


Siri-Ray (changes requested):
@portseif thanks for wiring up both tweaks protocols here. I found one blocker in the `.tw-panel` path: the injected bridge only exists in the srcDoc render path, while the default plain-HTML preview path URL-loads the iframe, so tweaks-template artifacts can still leave the new toolbar button disabled on first load. I left the concrete path and suggested fix inline.




portseif:
Thanks for catching this. Fixed in e5420b2.

**Root cause:** Plain HTML previews URL load by default, so `buildSrcdoc` and `injectTweaksBridge` never run for a fresh `.tw-panel` artifact. With no bridge, no `od:tweaks-available` postMessage, and the toolbar toggle stays disabled.

**Fix:**

1. New `hasTweaksTemplate(source)` helper in `apps/web/src/components/file-viewer-render-mode.ts` detects the fixed `.tw-panel` / `.tw-hidden` selectors with a word boundary anchored regex so `twk-panel`, `btw-panel`, etc. are not false positives.
2. `UrlLoadDecision` gains a `tweaksBridge?: boolean` flag; `shouldUrlLoadHtmlPreview` returns `false` when it is set, alongside the existing comment / inspect / palette / draw disqualifiers.
3. `FileViewer` computes `tweaksBridgeRequired = hasTweaksTemplate(source)` and passes it into the decision, so a tweaks template artifact picks the srcDoc render path on first load.

**Tests** in `apps/web/tests/components/file-viewer-render-mode.test.ts` cover:

* `shouldUrlLoadHtmlPreview({ tweaksBridge: true })` returns `false`.
* `hasTweaksTemplate` matches `.tw-panel` and `.tw-hidden`, rejects `twk-panel` / `btw-panel` / `mtw-hidden`, handles null / undefined / empty input.
* `tweaksBridge` combined with `forceInline` still resolves to srcDoc.

19 tests pass; `pnpm --filter @open-design/web typecheck` clean. The `__edit_mode_*` postMessage path is unaffected: those artifacts already announce themselves directly, regardless of render mode.

apps/web/src/runtime/srcdoc.ts
portseif: Fixed in https://github.com/nexu-io/open-design/commit/e5420b218a9f5f28fcc95bef4390b054f5cb2f45

portseif:
Pushed d78ed48 to address three problems that surfaced after the rebase onto v0.7.

* **Naming + icon collision:** v0.7 ships a palette popover button also titled `Tweaks` with the same sliders icon as the new toolbar toggle introduced here. Renamed the palette popover surface to `Themes` (label, dialog title, `aria-label`) and swapped its icon to a new `paint-bucket` glyph. The artifact tweaks toggle keeps the `tweaks` sliders icon. `data-testid="palette-tweaks-toggle"`, CSS classes, and the `PaletteTweaks` component name are kept stable so existing tests and styles still target the same nodes.
* **Auto toggle bug:** opening the Themes popover flipped render mode, which reloaded the iframe, which made the agent protocol artifact re-announce `__edit_mode_available`, which the host had been treating as `user opened the panel`. Result: clicking Themes also flipped Tweaks on. `__edit_mode_available` now only sets `tweaksAvailable`. `syncBridgeModes` pushes the current `tweaksMode` to the artifact on every iframe load so the panel matches the toolbar instead of the other way.
* **Render mode flash:** both URL load and srcDoc iframes are now mounted simultaneously, absolutely positioned and overlapping, with CSS visibility flipping between them. Toggling render mode (Themes open/close, palette select, etc.) no longer reloads the iframe so there is no flash. `isOurIframe(source)` lets six message receive sites accept events from either iframe so startup announcements from the hidden one are not lost. Sends still target `iframeRef.current`, kept aligned with the active iframe via a `useEffect`. A `syncBridgeModesRef` pushes current bridge state to the now visible iframe whenever the render mode flips.

Trade off: per HTML file we now hold two iframes in memory (URL load + srcDoc) instead of one. Acceptable for typical artifacts; could be revisited if it shows up in profiling for heavy multi file prototypes.

Tests in `apps/web/tests/components/FileViewer.test.tsx` updated: render mode assertions use a co attribute regex to confirm the active `data-testid` sits on the expected iframe rather than asserting the other iframe is absent. The Draw bar picking test re-queries the testid across mode toggles via a `getFrame()` helper. 95 tests pass, `pnpm --filter @open-design/web typecheck` clean.

apps/web/src/components/FileViewer.tsx
```diff
@@ -4053,6 +4108,52 @@ function HtmlViewer({
...
+  // (`od:tweaks-*` for skill-template artifacts) and the artifact-native
+  // edit-mode protocol (`__edit_mode_*` for agent-generated artifacts). Either
+  // surface controls toolbar availability and mirrors local close into the
+  // toolbar toggle state.
+  useEffect(() => {
+    function onMessage(ev: MessageEvent) {
+      if (!isOurIframe(ev.source)) return;
+      const data = ev.data as { type?: string; available?: boolean; visible?: boolean } | null;
+      if (!data?.type) return;
+      if (data.type === 'od:tweaks-available') {
```
lefarcen, line 4335: P2 `od:tweaks-available: false` from the hidden srcDoc iframe can race with `__edit_mode_available` and silently disable the Tweaks button for agent-protocol (`.twk-panel`) artifacts.

Both iframes are always mounted. For artifacts where `useUrlLoadPreview = true` (i.e. `.twk-panel` agent artifacts — `hasTweaksTemplate` returns false so `tweaksBridgeRequired = false`), the hidden srcDoc iframe still receives the full artifact source via `buildSrcdoc` including the always-injected tweaks bridge. That bridge fires `od:tweaks-available: false` (no `.tw-panel` in source), and since `isOurIframe` accepts messages from both iframes, this message can land *after* the URL-load iframe's `__edit_mode_available` has already set `tweaksAvailable = true`, overriding it.

Fix: scope `od:tweaks-available` to the active iframe only:
```ts
if (data.type === 'od:tweaks-available') {
  // Only trust od:tweaks-available from the active iframe — the hidden
  // srcDoc iframe's passive tweaks bridge fires available=false for
  // agent-protocol artifacts that have no .tw-panel in their source.
  if (ev.source !== iframeRef.current?.contentWindow) return;
  setTweaksAvailable(!!data.available);
}
```
`__edit_mode_available` and `__edit_mode_dismissed` should stay accepted from both iframes — those are artifact-native signals that carry real intent and need to survive render-mode flips.

lefarcen (changes requested):
Hey @portseif — three rounds of fixes in a single day is solid, focused iteration. The final shape looks good: dual-iframe visibility flip to kill the remount flash, `suppressEcho` via microtask ordering for bidirectional sync without an extra `setTimeout`, `hasTweaksTemplate` with a word-boundary regex that correctly rejects `twk-panel`/`btw-panel`/`mtw-hidden`. The `__edit_mode_available`→availability-only split (no mirroring into `tweaksMode`) is exactly the right call too.

One issue surfaced in the deep review that needs fixing before this lands — see the inline comment below.

portseif:
Thanks, fixed in cf55074.

The hidden srcDoc iframe's tweaks bridge was indeed posting `available: false` for agent protocol artifacts (no `.tw-panel` in source) and overriding the active iframe's `__edit_mode_available`. Scoped that one signal to the active iframe only by re-checking `ev.source === iframeRef.current?.contentWindow` inside the `od:tweaks-available` branch. `__edit_mode_*` stays accepted from either iframe so the artifact's own announcement keeps driving the toolbar toggle across render mode flips, as you noted.

Typecheck clean.

lefarcen (dismissed):
The `ev.source !== iframeRef.current?.contentWindow` guard on `od:tweaks-available` is exactly right — scopes availability signals to the active iframe only while leaving `__edit_mode_*` open to both, since those carry real artifact intent that must survive render-mode flips. The comment in the code explains the reasoning clearly too.

My prior blocking finding is resolved. Clean from my side. ✨

portseif:
@Siri-Ray your CHANGES_REQUESTED block from earlier still gates this PR. The race you flagged was fixed in cf55074 and lefarcen confirmed the resolution. Could you re-review when you have a moment so the PR can unblock?

portseif:
@lefarcen thanks again for the approval. The required `Validate workspace` check is still pending because the workflow runs from my fork are stuck at "Action required" in the Actions tab (default GitHub policy for first time fork contributors). When you have a moment, could you open the Actions tab on `nexu-io/open-design`, find the latest workflow run for this PR, and click **Approve and run workflows**? That should unblock CI and let the merge gate finally go green.

lefarcen:
Hey @portseif! Both pending workflow runs for this PR (`ci` and `nix-check`) are now approved — CI should be running shortly. Keep an eye on the checks; once they go green the remaining gate is Siri-Ray's re-review + product sign-off, and this should be clear to land. 🙌

portseif:
@lefarcen one more nudge: CI is green now, and your APPROVED review is on file, but the merge gate is still stuck on `Siri-Ray`'s CHANGES_REQUESTED from earlier (their flagged race was fixed in cf55074, which you also confirmed). When you have a moment, could you open the Reviewers panel on the PR and click **Dismiss review** on the Siri-Ray entry? That should flip the decision to APPROVED and let the merge button go.

lefarcen:
Noted, @portseif — dismissing another reviewer's CHANGES_REQUESTED is a call the maintainer team makes, so I've flagged this to them now for a decision.

A couple of things worth knowing on the path forward: even once Siri-Ray's review is resolved (dismissed or superseded by a fresh approve), there's still a product review pending on this PR — we need a sign-off from the product lead on the UI/API surface changes before it can land. The team is aware and working through it. Appreciate your patience on this. 🙌

apps/web/src/runtime/srcdoc.ts
```diff
@@ -1563,3 +1568,110 @@ function injectDeckBridge(doc: string, initialSlideIndex = 0): string {
...
+
+  function onReady(){
+    applyClassesToPanel(!document.documentElement.hasAttribute('data-od-tweaks-hidden'));
+    attachObserver();
+    postAvailability();
+    // Post the artifact initial visibility so the toolbar toggle reflects the
+    // default state on mount. Without this the toggle reads OFF while a
+    // default-visible tw-panel artifact clearly shows its panel and the user
+    // would have to click toggle-on then toggle-off to actually hide.
+    postState();
```
Siri-Ray, line 1665: This new initial `postState()` still reports the bridge-forced hidden state, not the artifact's initial panel state. The injected script sets `data-od-tweaks-hidden` before the body parses, and `onReady()` then calls `applyClassesToPanel(!document.documentElement.hasAttribute('data-od-tweaks-hidden'))`; because that attribute is always present at this point, a default-visible `.tw-panel` gets `tw-hidden` added immediately before this line reads `!panel.classList.contains('tw-hidden')`. That means the host receives `visible: false`, so the class-based template path still starts with the toolbar OFF and the panel hidden, which is the mismatch this commit is trying to fix. Capture the panel's authored visibility before applying the host hidden attribute, then use that value to initialize both the document attribute/classes and the posted state, for example `const initialVisible = !!panel && !panel.classList.contains('tw-hidden')` before `applyClassesToPanel(...)`, followed by `document.documentElement.toggleAttribute('data-od-tweaks-hidden', !initialVisible)`, `applyClassesToPanel(initialVisible)`, and `postState()` after the classes reflect that captured value.


Siri-Ray (commented):
@portseif thanks for the continued iteration here. The earlier render-mode and hidden-iframe races look addressed on this head; I found one smaller follow-up in the latest initial-state path for the class-based tweaks bridge.




apps/web/src/components/FileViewer.tsx
```diff
@@ -4246,6 +4307,70 @@ function HtmlViewer({
...
+        // panel visible (the SDK pattern is `useState(true)`), so without
+        // this the toolbar toggle reads as OFF while the panel is clearly
+        // ON — the user has to click toggle-on → toggle-off to hide the
+        // panel. Guarded by `firstEditModeAvailableSeenForFileRef` so a
+        // later iframe remount (Themes popover flipping render mode, etc.)
+        // doesn't snap a user-driven OFF back to ON. `syncBridgeModes` is
+        // still the source of truth on every subsequent load: it pushes the
+        // current `tweaksMode` into the artifact via `__activate_edit_mode`
+        // / `__deactivate_edit_mode` so the artifact tracks the toolbar.
+        if (firstEditModeAvailableSeenForFileRef.current !== file.name) {
```
Siri-Ray, line 4366: This once-per-file guard is inside a message listener installed with `useEffect(..., [])`, so the handler keeps the `file.name` from the first render. After opening one `.twk-panel` artifact, `firstEditModeAvailableSeenForFileRef.current` is set to that first file; when the user switches to another `.twk-panel` file, the reset effect below sets `tweaksMode` back to `false`, but the second file's `__edit_mode_available` still compares against the stale first `file.name` captured by this closure and therefore does not mirror the new artifact's default-open state back to `true`. The result is the toolbar can start OFF while the newly selected artifact's panel is visible, which is exactly the mismatch this guard is trying to prevent on initial load. Please either include `file.name` (and the iframe-source predicate it closes over) in this listener effect's dependency list, or keep the current file key in a ref that is updated on file changes and compare/update the guard against that ref; a FileViewer test that rerenders from one `__edit_mode_available` artifact to another and dispatches the second message would cover the regression.


Siri-Ray (commented):
@portseif thanks for the continued fixes here. The previous class-based initial-state issue looks addressed on this head; I found one remaining file-switch lifecycle issue in the `.twk-panel` path that is worth tightening, but I do not think it needs to block merge.




apps/web/src/i18n/types.ts
```diff
@@ -1216,6 +1216,7 @@ export interface Dict {
   'fileViewer.preview': string;
   'fileViewer.source': string;
   'fileViewer.tweaks': string;
+  'fileViewer.tweaksUnavailable': string;
```
Siri-Ray, line 1219: Adding `fileViewer.tweaksUnavailable` to `Dict` means this user-facing toolbar tooltip should be carried through every active locale dictionary, but the PR only adds explicit translations for the 13 locale files touched in this diff. The repository currently has 19 locale files, and `apps/web/src/i18n/locales/ar.ts`, `fr.ts`, `id.ts`, `it.ts`, `th.ts`, and `uk.ts` still do not define this key locally; they only inherit the English string through the `...en` fallback. That leaves users in those locales seeing the English disabled-tooltip text even though the i18n convention for new keys is to define it in each locale. Please add `fileViewer.tweaksUnavailable` beside `fileViewer.tweaks` in the remaining locale files as well, and update the PR body’s “13 supported locales” note if the supported-locale count has drifted.


Siri-Ray (commented):
@portseif thanks for the continued iteration here. The latest iframe lifecycle fixes and regression coverage look good from this implementation pass; I found one remaining i18n coverage gap that is worth tightening but should not block the main tweaks behavior.




apps/web/src/components/FileViewer.tsx
```diff
@@ -4246,6 +4307,76 @@ function HtmlViewer({
...
+          firstEditModeAvailableSeenForFileRef.current = file.name;
+          setTweaksMode(true);
```
Siri-Ray: `__edit_mode_available` is documented in `design-templates/tweaks/SKILL.md` as supporting panels that may default open or closed, with the host syncing to whichever state the artifact reports, but the actual Protocol A message only carries `{ type: '__edit_mode_available' }` and this branch always mirrors it to `tweaksMode = true`. A default-closed dynamic artifact that follows the new contract will be opened immediately by the host once this state update runs and the effect above sends `__activate_edit_mode`, so the artifact cannot remain closed even though the contract says that is supported. Please either tighten the contract to say Protocol A availability means the panel is currently open / panels should default open, or extend the message shape, for example `{ type: '__edit_mode_available', visible: false }`, and set `tweaksMode` from that value with a FileViewer test for the default-closed announcement path.


Siri-Ray (commented):
@portseif thanks for the careful follow-up fixes here. The previously raised initial-state, iframe-source, file-switch, and locale coverage items look addressed on this head. I found one remaining postMessage contract mismatch that is worth tightening, but it does not need to block the core toggle binding.




Siri-Ray (approved):
@portseif I reviewed the current changed ranges on this head. The earlier issues around class-based `.tw-panel` render-mode selection, hidden-iframe availability races, per-file Protocol A state mirroring, `visible: false` handling, and full locale coverage all look addressed now. `git diff --check origin/main...HEAD` is clean; the focused Vitest command could not start in this prepared worktree because `node_modules` is missing (`vitest` was not found), so I leaned on the live green checks plus the diff review for this pass. Thanks for the careful iteration here.

PR #1649, opened 2026-05-14
fix: Gate Finalize design package when provider config is missing

### LINKED ISSUE #1348: Finalize design package is clickable without provider config and fails with Bad request instead of being gated

xxiaoxiong:
Fixes #1348

## Why

**Use case:** Users working in Local CLI mode or with incomplete BYOK provider configuration see the "Finalize design package" button in the project toolbar.

**Pain being addressed:** The button is currently clickable even when the required provider configuration (API key and model) is missing. Clicking it sends a finalize request with empty config values, which fails with:

This is a **P1 UX issue** because:
- The product should prevent invalid actions rather than surfacing backend errors
- Users experience this as a broken feature
- The failure happens on a key completion action
- The error message doesn't clearly explain what's missing or how to fix it

## What users will see

**Before this fix:**
- "Finalize design package" button is clickable
- User clicks → request sent with empty apiKey/model
- Backend returns 400 Bad Request
- User sees: "Bad request — check the API key and model."
- No clear guidance on what to do next

**After this fix:**
- "Finalize design package" button is disabled when config is incomplete
- Hovering shows tooltip: "Configure API key and model in Settings to enable finalization"
- Button cannot be clicked → no failed request
- Clear, actionable guidance before the user tries to act

**Defensive guard (if triggered programmatically):**
- Toast message: "Cannot finalize: API key and model are required."
- Details: "Configure your provider in Settings before finalizing."

## Surface area

- [x] **UI** — Finalize button disabled state and tooltip
- [ ] **i18n keys** — no changes (tooltip is English-only for now)
- [ ] **Default behavior change** — gates existing action (improvement, not breaking)

## Screenshots

N/A — This is a gating fix. The button now shows disabled state with a tooltip when config is incomplete.

## Bug fix verification

**Test reproduction:**
1. Open a project in Local CLI mode (or with incomplete BYOK config)
2. Ensure `apiKey` or `model` is empty in Settings
3. **Before fix:** "Finalize design package" is clickable → click → 400 error
4. **After fix:** "Finalize design package" is disabled → hover shows tooltip
5. Configure API key and model in Settings
6. **After fix:** Button becomes enabled and clickable

**Why no automated test:**
- This is a UI gating fix that requires visual verification
- The logic is straightforward: `canFinalize = Boolean(apiKey && model)`
- Manual verification confirms the button is disabled when config is incomplete

## Validation

- [x] Code review: verified `canFinalize` check in ProjectView
- [x] Code review: verified disabled state flows through all 3 components
- [x] Code review: verified defensive guard in handleFinalize
- [x] Code review: verified tooltip provides clear guidance
- [x] Logic: `canFinalize = Boolean(config.apiKey && config.model)`
- [x] Defensive: handleFinalize checks config before calling finalize.trigger

**Commands run:**
- Code inspection of all 3 modified files
- Verified prop flow: ProjectView → ProjectActionsToolbar → FinalizeDesignButton
- Confirmed disabled attribute and title are applied correctly

#### 1. `apps/web/src/components/ProjectView.tsx`

**Added config validation (line 2300):**
```typescript
// Check if finalize is available: requires apiKey and model (#1348)
const canFinalize = Boolean(config.apiKey && config.model);
```

**Added defensive guard in handleFinalize (lines 2303-2310):**
```typescript
const handleFinalize = useCallback(() => {
  // Guard against missing config (should be prevented by disabled state,
  // but defend in depth in case the button is triggered programmatically)
  if (!config.apiKey || !config.model) {
    setProjectActionsToast({
      message: 'Cannot finalize: API key and model are required.',
      details: 'Configure your provider in Settings before finalizing.',
    });
    return;
  }
  // ... rest of finalize logic
}, [finalize, config, designMdState]);
```

**Passed canFinalize to toolbar (line 2503):**
```typescript
<ProjectActionsToolbar
  designMdState={designMdState}
  finalizeStatus={finalize.status}
  canFinalize={canFinalize}  // ← new prop
  onFinalize={handleFinalize}
  // ...
/>
```

#### 2. `apps/web/src/components/ProjectActionsToolbar.tsx`

**Added canFinalize prop (line 19):**
```typescript
export interface ProjectActionsToolbarProps {
  designMdState: Pick<DesignMdState, 'exists' | 'isStale' | 'staleReason'>;
  finalizeStatus: FinalizeStatus;
  canFinalize: boolean;  // ← new prop
  // ...
}
```

**Passed disabled state to button (line 45):**
```typescript
<FinalizeDesignButton
  designMdState={designMdState}
  status={finalizeStatus}
  disabled={!canFinalize}  // ← new prop
  onFinalize={onFinalize}
  onCancel={onCancelFinalize}
/>
```

#### 3. `apps/web/src/components/FinalizeDesignButton.tsx`

**Added disabled prop (line 19):**
```typescript
export interface FinalizeDesignButtonProps {
  designMdState: Pick<DesignMdState, 'exists' | 'isStale'>;
  status: FinalizeStatus;
  disabled?: boolean;  // ← new prop
  onFinalize: () => void;
  onCancel: () => void;
}
```

**Applied disabled state to button (lines 66-67):**
```typescript
<button
  type="button"
  className={`project-actions-button ${variantClass}`}
  onClick={onFinalize}
  disabled={disabled}
  title={disabled ? 'Configure API key and model in Settings to enable finalization' : undefined}
>
  {label}
</button>
```

### Why this approach

1. **Preflight validation**: Check config before allowing the action
2. **Clear guidance**: Tooltip explains what's missing and where to fix it
3. **Defense in depth**: Handler also checks config in case button is triggered programmatically
4. **Minimal changes**: Only 3 files, 22 lines added
5. **Type-safe**: Props flow through TypeScript interfaces
6. **Accessible**: Uses native `disabled` attribute and `title` for tooltip

### Validation logic

```typescript
const canFinalize = Boolean(config.apiKey && config.model);
```

- `config.apiKey`: Required for authentication
- `config.model`: Required to specify which model to use
- Both must be truthy (non-empty strings)
- `baseUrl` is optional (has a default)
- `maxTokens` is optional (has a default)

## Impact

- **Surface area**: Finalize button gating logic
- **User experience**: Clear preflight validation instead of backend error
- **Accessibility**: Native disabled state + tooltip
- **Files changed**: 3 (ProjectView, ProjectActionsToolbar, FinalizeDesignButton)
- **Lines changed**: +22
- **Breaking changes**: None (gates existing action, doesn't remove it)

## Related

- #1348 — original issue tracking this UX problem
- #451 — PR that introduced the Finalize design package feature
- #832 — PR that added the /finalize/anthropic endpoint

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -2295,7 +2295,21 @@ export function ProjectView({
   // Continue in CLI / Finalize design package handlers + keyboard
   // shortcut wiring. Close to the JSX so the data flow is easy to
   // trace from the toolbar back to its sources.
+  
+  // Check if finalize is available: requires apiKey and model (#1348)
+  const canFinalize = Boolean(config.apiKey && config.model);
+  
   const handleFinalize = useCallback(() => {
+    // Guard against missing config (should be prevented by disabled state,
+    // but defend in depth in case the button is triggered programmatically)
+    if (!config.apiKey || !config.model) {
+      setProjectActionsToast({
+        message: 'Cannot finalize: API key and model are required.',
+        details: 'Configure your provider in Settings before finalizing.',
+      });
+      return;
+    }
+    
     void finalize.trigger({
       apiKey: config.apiKey,
       baseUrl: config.baseUrl,
@@ -2486,6 +2500,7 @@ export function ProjectView({
       <ProjectActionsToolbar
         designMdState={designMdState}
         finalizeStatus={finalize.status}
+        canFinalize={canFinalize}
         onFinalize={handleFinalize}
         onCancelFinalize={handleCancelFinalize}
         onContinueInCli={handleContinueInCli}
--- a/apps/web/src/components/FinalizeDesignButton.tsx
+++ b/apps/web/src/components/FinalizeDesignButton.tsx
@@ -16,6 +16,7 @@ import type { FinalizeStatus } from '../hooks/useFinalizeProject';
 export interface FinalizeDesignButtonProps {
   designMdState: Pick<DesignMdState, 'exists' | 'isStale'>;
   status: FinalizeStatus;
+  disabled?: boolean;
   onFinalize: () => void;
   onCancel: () => void;
 }
@@ -23,6 +24,7 @@ export interface FinalizeDesignButtonProps {
 export function FinalizeDesignButton({
   designMdState,
   status,
+  disabled = false,
   onFinalize,
   onCancel,
 }: FinalizeDesignButtonProps) {
@@ -61,6 +63,8 @@ export function FinalizeDesignButton({
       type="button"
       className={`project-actions-button ${variantClass}`}
       onClick={onFinalize}
+      disabled={disabled}
+      title={disabled ? 'Configure API key and model in Settings to enable finalization' : undefined}
     >
       {label}
     </button>
--- a/apps/web/src/components/ProjectActionsToolbar.tsx
+++ b/apps/web/src/components/ProjectActionsToolbar.tsx
@@ -16,6 +16,7 @@ import type { FinalizeStatus } from '../hooks/useFinalizeProject';
 export interface ProjectActionsToolbarProps {
   designMdState: Pick<DesignMdState, 'exists' | 'isStale' | 'staleReason'>;
   finalizeStatus: FinalizeStatus;
+  canFinalize: boolean;
   onFinalize: () => void;
   onCancelFinalize: () => void;
   onContinueInCli: () => void | Promise<void>;
@@ -25,6 +26,7 @@ export interface ProjectActionsToolbarProps {
 export function ProjectActionsToolbar({
   designMdState,
   finalizeStatus,
+  canFinalize,
   onFinalize,
   onCancelFinalize,
   onContinueInCli,
@@ -40,6 +42,7 @@ export function ProjectActionsToolbar({
       <FinalizeDesignButton
         designMdState={designMdState}
         status={finalizeStatus}
+        disabled={!canFinalize}
         onFinalize={onFinalize}
         onCancel={onCancelFinalize}
       />


apps/web/src/components/ProjectView.tsx
```diff
@@ -2295,7 +2295,21 @@ export function ProjectView({
   // Continue in CLI / Finalize design package handlers + keyboard
   // shortcut wiring. Close to the JSX so the data flow is easy to
   // trace from the toolbar back to its sources.
+  
+  // Check if finalize is available: requires apiKey and model (#1348)
+  const canFinalize = Boolean(config.apiKey && config.model);
```
Siri-Ray, line 2300: This enables finalize whenever `apiKey` and `model` are truthy, but the daemon route rejects whitespace-only values with `!apiKey.trim()` and `!model.trim()` before running finalize. That means entering spaces in either Settings field still leaves the button enabled, sends the request, and shows the same `BAD_REQUEST` flow this PR is trying to avoid. Please base both `canFinalize` and the defensive guard on trimmed values, and preferably pass the trimmed `apiKey` / `model` into `finalize.trigger` so the UI and route share the same notion of configured input.

Siri-Ray (commented):
@xxiaoxiong thanks for tightening this flow. I found one remaining config-validation gap where the UI still diverges from the daemon's existing validation, so I'm leaving this as non-blocking follow-up feedback.


lefarcen (commented):
Hey @xxiaoxiong 👋 The defense-in-depth shape here is solid — `canFinalize` gate at the view layer, `disabled` prop threaded type-safely down through `ProjectActionsToolbar` → `FinalizeDesignButton`, and the programmatic guard in `handleFinalize` as a backstop. CI green, approach is right. ✨

One thing worth addressing before this lands — @Siri-Ray's inline on line 2300 nails it: whitespace-only values (`"   "`) are truthy in JS so the current check still leaves the button enabled for that edge case, and the daemon rejects them with `!apiKey.trim()` / `!model.trim()`, surfacing the same Bad Request this PR is trying to prevent. The fix is minimal:

```typescript
const canFinalize = Boolean(config.apiKey?.trim() && config.model?.trim());
```

Carry the same trim into the defensive guard (lines 2305–2306) and optionally pass trimmed values into `finalize.trigger` to keep UI and daemon validation fully in sync.

Once that's pushed, this is ready to go.

lefarcen:
Heads-up: PR #1749 has just been opened targeting the same area — both PRs touch `FinalizeDesignButton.tsx` and address the Finalize gating for #1348. PR #1749 adds the `.trim()` handling on `apiKey`/`model` that was flagged in the review discussion here. You and @ecogsandbox may want to compare approaches; the maintainer team will decide which one lands. Sharing so neither effort goes to waste.

PR #1662, opened 2026-05-14
Support folder drops in Design Files

### LINKED ISSUE #776: Fix folder drag-and-drop upload failing in Design Files

mzl2233:
Fixes #776

## Why

Dragging a folder into Design Files currently follows the flat file path and can send the folder placeholder as a single upload item, which fails even though the dropzone invites users to drop folders. I wanted the advertised folder drop path to work for the common case of bringing a local asset folder into a project.

## What users will see

Dropping a folder onto the Design Files dropzone now recursively expands the folder contents and uploads the files it contains. Existing single-file and multi-file drops continue to use the same upload flow.

## Surface area

## Screenshots

No visual change; this updates the existing Design Files dropzone behavior.

## Bug fix verification

- Test path that reproduces the bug: manual browser drag-and-drop behavior depends on `DataTransferItem.webkitGetAsEntry`, so I verified the affected web package with typecheck and the existing Design Files/FileWorkspace test coverage instead of adding a jsdom-only red spec.

## Validation

- `pnpm --filter @open-design/web test -- --runInBand`

### DIFF
--- a/apps/web/src/components/DesignFilesPanel.tsx
+++ b/apps/web/src/components/DesignFilesPanel.tsx
@@ -31,6 +31,18 @@ type DesignFilesGroupMode = 'kind' | 'modified';
 type ModifiedSection = 'today' | 'yesterday' | 'previous7Days' | 'previous30Days' | 'older';
 type SortKey = 'name' | 'kind' | 'mtime';
 type SortDir = 'asc' | 'desc';
+type FileSystemEntryWithReader = FileSystemEntry & {
+  createReader?: () => FileSystemDirectoryReader;
+};
+type FileSystemFileEntryWithFile = FileSystemFileEntry & {
+  file: (
+    successCallback: (file: File) => void,
+    errorCallback?: (error: DOMException) => void,
+  ) => void;
+};
+type DataTransferItemWithEntry = DataTransferItem & {
+  webkitGetAsEntry?: () => FileSystemEntry | null;
+};
 
 const MODIFIED_SECTION_ORDER: ModifiedSection[] = [
   'today',
@@ -538,11 +550,11 @@ export function DesignFilesPanel({
     }
   }
 
-  function handleDrop(ev: React.DragEvent<HTMLDivElement>) {
+  async function handleDrop(ev: React.DragEvent<HTMLDivElement>) {
     ev.preventDefault();
     dragDepthRef.current = 0;
     setDraggingFiles(false);
-    const dropped = Array.from(ev.dataTransfer.files ?? []);
+    const dropped = await filesFromDataTransfer(ev.dataTransfer);
     if (dropped.length > 0) onUploadFiles(dropped);
   }
 
@@ -1076,6 +1088,49 @@ function dateDaysBefore(date: Date, days: number): Date {
   const result = new Date(date);
   result.setDate(result.getDate() - days);
   return result;
+}
+
+async function filesFromDataTransfer(dataTransfer: DataTransfer): Promise<File[]> {
+  const items = Array.from(dataTransfer.items ?? []);
+  if (items.length === 0) return Array.from(dataTransfer.files ?? []);
+
+  const files = await Promise.all(items.map(filesFromDataTransferItem));
+  const flattened = files.flat();
+  return flattened.length > 0 ? flattened : Array.from(dataTransfer.files ?? []);
+}
+
+async function filesFromDataTransferItem(item: DataTransferItem): Promise<File[]> {
+  const entry = (item as DataTransferItemWithEntry).webkitGetAsEntry?.();
+  if (!entry) {
+    const file = item.kind === 'file' ? item.getAsFile() : null;
+    return file ? [file] : [];
+  }
+  return filesFromFileSystemEntry(entry);
+}
+
+async function filesFromFileSystemEntry(entry: FileSystemEntry): Promise<File[]> {
+  if (entry.isFile) return [await fileFromEntry(entry as FileSystemFileEntryWithFile)];
+  if (!entry.isDirectory) return [];
+
+  const reader = (entry as FileSystemEntryWithReader).createReader?.();
+  if (!reader) return [];
+
+  const files: File[] = [];
+  for (;;) {
+    const entries = await readEntryBatch(reader);
+    if (entries.length === 0) break;
+    const nested = await Promise.all(entries.map(filesFromFileSystemEntry));
+    files.push(...nested.flat());
+  }
+  return files;
+}
+
+function fileFromEntry(entry: FileSystemFileEntryWithFile): Promise<File> {
+  return new Promise((resolve, reject) => entry.file(resolve, reject));
+}
+
+function readEntryBatch(reader: FileSystemDirectoryReader): Promise<FileSystemEntry[]> {
+  return new Promise((resolve, reject) => reader.readEntries(resolve, reject));
 }
 
 function kindGlyph(kind: ProjectFileKind): string {


apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -1078,6 +1090,49 @@ function dateDaysBefore(date: Date, days: number): Date {
   return result;
 }
 
+async function filesFromDataTransfer(dataTransfer: DataTransfer): Promise<File[]> {
+  const items = Array.from(dataTransfer.items ?? []);
+  if (items.length === 0) return Array.from(dataTransfer.files ?? []);
+
+  const files = await Promise.all(items.map(filesFromDataTransferItem));
+  const flattened = files.flat();
+  return flattened.length > 0 ? flattened : Array.from(dataTransfer.files ?? []);
```
PerishCode, line 1099: **Empty-folder drop falls back to the very placeholder this PR is trying to avoid.**

Problem: when `items` is non-empty but recursive enumeration yields zero files (for example, the user drops a truly empty directory, or every item is a non-file `DataTransferItem` like a string/URL), `flattened.length === 0` triggers `Array.from(dataTransfer.files ?? [])`. On Chromium, a dropped folder still appears in `dataTransfer.files` as a single zero-byte folder placeholder, so we hand that placeholder to `onUploadFiles` and the upload fails — the exact symptom `#776` describes.

Why it matters: the fallback is meant to handle browsers without `webkitGetAsEntry`, not to recover from "the entries API said this drop has zero files." Conflating the two undoes the fix for the empty-folder case and means the PR description's promise ("recursively expands the folder contents") quietly regresses to the old behavior whenever expansion legitimately produces no files.

Evidence: the changed line is the only thing standing between an empty-folder drop and the original bug — `filesFromDataTransferItem` correctly returns `[]` for an empty directory entry, but `filesFromDataTransfer` then overwrites that empty result with `dataTransfer.files`.

Suggested change: detect API availability up front instead of inferring it from the result. For example:

```ts
async function filesFromDataTransfer(dataTransfer: DataTransfer): Promise<File[]> {
  const items = Array.from(dataTransfer.items ?? []);
  if (items.length === 0) return Array.from(dataTransfer.files ?? []);

  const hasEntryApi = items.some(
    (item) => typeof (item as DataTransferItemWithEntry).webkitGetAsEntry === 'function',
  );
  if (!hasEntryApi) return Array.from(dataTransfer.files ?? []);

  const files = await Promise.all(items.map(filesFromDataTransferItem));
  return files.flat();
}
```

This preserves the legacy-browser fallback while letting an empty / non-file drop legitimately resolve to `[]`, so `handleDrop`'s `dropped.length > 0` guard skips the upload instead of pushing a doomed folder placeholder.


apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -1078,6 +1090,49 @@ function dateDaysBefore(date: Date, days: number): Date {
...
+  const entry = (item as DataTransferItemWithEntry).webkitGetAsEntry?.();
+  if (!entry) {
+    const file = item.kind === 'file' ? item.getAsFile() : null;
+    return file ? [file] : [];
+  }
+  return filesFromFileSystemEntry(entry);
+}
+
+async function filesFromFileSystemEntry(entry: FileSystemEntry): Promise<File[]> {
+  if (entry.isFile) return [await fileFromEntry(entry as FileSystemFileEntryWithFile)];
```
PerishCode, line 1112: **One unreadable file aborts the whole drop, silently.**

Problem: `fileFromEntry` wires `entry.file(resolve, reject)`, so any per-file error (file deleted/moved between drag and drop, OS-level permission failure, sandboxed link target, etc.) rejects the promise. That rejection propagates up through `Promise.all` in both `filesFromFileSystemEntry` and `filesFromDataTransfer`, then through `await filesFromDataTransfer(...)` in `handleDrop` at line 557 — which is now `async` but has no `try/catch`. Net effect: a single unreadable file in a 200-file folder drop causes (a) zero files to be uploaded and (b) an unhandled promise rejection in the console, with no user-visible error from the dropzone.

Why it matters: this is a regression vs. the pre-PR flat-drop behavior, where `dataTransfer.files` would have just omitted unreadable entries and uploaded the rest. Folder drops are exactly the case where partial failure is most likely (large trees, OS junctions, temp files), and they're the path this PR is introducing.

Evidence: `filesFromFileSystemEntry` at line 1112 awaits `fileFromEntry(...)` with no recovery, and `handleDrop` does not catch the rejected promise.

Suggested change: skip individual failures rather than collapsing the batch. The cheapest fix is to swallow per-entry errors at the leaf:

```ts
async function filesFromFileSystemEntry(entry: FileSystemEntry): Promise<File[]> {
  if (entry.isFile) {
    try {
      return [await fileFromEntry(entry as FileSystemFileEntryWithFile)];
    } catch {
      return [];
    }
  }
  // ...rest unchanged
}
```

Optionally also wrap the `await` in `handleDrop` in a `try/catch` so a top-level reader failure surfaces through `uploadError` instead of an unhandled rejection. A short comment on `fileFromEntry` explaining that callers expect failures-as-empty would keep the intent obvious for future readers.


PerishCode (commented):
Folder-drop expansion looks correct for the common case (Chromium with `webkitGetAsEntry`), and the typing of the experimental DataTransferItem / FileSystemEntry surfaces is appropriately narrow. Two non-blocking concerns about edge-case correctness inlined below — both are small follow-ups, not merge blockers.



apps/web/src/components/DesignFilesPanel.tsx
```diff
@@ -538,11 +550,11 @@ export function DesignFilesPanel({
     }
   }
 
-  function handleDrop(ev: React.DragEvent<HTMLDivElement>) {
+  async function handleDrop(ev: React.DragEvent<HTMLDivElement>) {
     ev.preventDefault();
     dragDepthRef.current = 0;
     setDraggingFiles(false);
-    const dropped = Array.from(ev.dataTransfer.files ?? []);
+    const dropped = await filesFromDataTransfer(ev.dataTransfer);
```
lefarcen, line 557: P2 If any `fileFromEntry` or `readEntryBatch` call rejects (permission error, file-removed-between-drag-and-drop, etc.), the rejection propagates through `Promise.all` and out of `filesFromDataTransfer` — and because `handleDrop` has no `try/catch`, that becomes an unhandled Promise rejection. Users see nothing; error monitoring sees an uncaught exception.

Minimal fix that keeps the silent-skip intent PerishCode described:
```ts
const dropped = await filesFromDataTransfer(ev.dataTransfer).catch(() => []);
```
This way a tree-walk failure degrades to "nothing uploaded" rather than a thrown exception.

lefarcen (commented):
Hey @mzl2233! 👋 The recursive `DataTransferItem` enumeration is the right approach here — and narrowing the experimental `FileSystemEntry`/`DataTransferItem` surfaces with your own intersected types instead of casting to `any` keeps the typing honest. Good work on the fallback chain too: missing `webkitGetAsEntry` → `getAsFile()`, empty item list → `dataTransfer.files`, all-empty enumeration → `dataTransfer.files`.

PerishCode's review covers the two main edge cases (empty-folder fallback and per-file error aborting the whole tree). One related thing I'd add below — it's in the same error-handling area, and similarly non-blocking on its own, but worth addressing alongside theirs.

leilei926524-tech:
Hi @mzl2233!

You contributed:

Merged PR: #1662 Support folder drops in Design Files
https://github.com/nexu-io/open-design/pull/1662

1. #1772 Feature request: allow users to configure artifact storage directory
https://github.com/nexu-io/open-design/issues/1772

2. #2180 Conversation button in the chat panel header is too close to the edge
https://github.com/nexu-io/open-design/issues/2180

The Open Design team

leilei926524-tech:
Hi @mzl2233!

You contributed:

Merged PR: #1662 Support folder drops in Design Files
https://github.com/nexu-io/open-design/pull/1662

Your PR touched the web experience around support folder drops in design files, so the follow-up issues below should feel close to the same surface area.

The Open Design team

PR #1666, opened 2026-05-14
fix: Add error feedback for manual folder path import

xxiaoxiong:
Fixes #1408

## Why

**Use case:** Users manually enter a folder path in the "From folder" tab and click "Open folder" to import an existing project.

**Pain being addressed:** When the import fails (invalid path, permission denied, daemon error), the app provides no visible feedback. The button briefly shows "Opening…" then returns to "Open folder" with no error message, leaving users confused about whether the action succeeded, failed, or was ignored.

**Root cause:** The manual path import flow (line 620-628) lacks error handling:
- Does not clear previous errors before attempting import
- Does not catch exceptions from onImportFolder()
- Does not display error messages to the user

The Electron picker flow (line 584-618) already has proper error handling with setImportFolderError(), but the manual path flow was missing this.

## What users will see

**Before this fix:**
- Enter an invalid path → click "Open folder" → button shows "Opening…" briefly → returns to "Open folder" → no error message
- Users cannot tell if the path was invalid, if permissions were denied, or if the daemon is down
- Silent failures make the feature feel broken

**After this fix:**
- Enter an invalid path → click "Open folder" → button shows "Opening…" → error message appears: "Open folder failed: [specific reason]"
- Previous errors are cleared before each attempt
- Success feedback is implicit: the app navigates to the opened project (changing the entire view)

## Surface area

- [x] **UI** — error message display in NewProjectPanel
- [ ] **i18n keys** — no changes (error messages are dynamic)
- [ ] **Default behavior change** — adds error feedback, does not change success path

## Validation

- [x] Code review: error handling matches Electron picker pattern
- [x] Code review: clears previous errors before import
- [x] Code review: catches all errors and displays user-friendly messages
- [x] Logic: success feedback is implicit (navigation to project)

### Changes made

**Before (line 620-628):**
- No error clearing
- No error catching
- Silent failures

**After:**
- Clears previous errors with setImportFolderError(null)
- Catches all errors from onImportFolder()
- Displays user-friendly error messages
- Success feedback is implicit (navigation to opened project)

### Why this approach

1. **Matches existing pattern**: The Electron picker flow already uses setImportFolderError() for error display
2. **Clears previous errors**: Ensures stale error messages don't confuse users
3. **Catches all errors**: Handles invalid path, permission denied, daemon errors
4. **User-friendly messages**: Extracts error message when available
5. **Implicit success feedback**: Navigation provides clear visual feedback

## Impact

- **Surface area**: "From folder" tab error handling
- **User experience**: Clear error feedback for failed imports
- **Files changed**: 1 (NewProjectPanel.tsx)
- **Lines changed**: +8
- **Breaking changes**: None

### DIFF
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -150,7 +150,7 @@ interface Props {
   // input and the renderer POSTs `/api/import/folder` itself. Browser
   // builds have no `shell.openPath` surface, so the renderer naming a
   // path here cannot escalate (PR #974 trust model).
-  onImportFolder?: (baseDir: string) => Promise<void> | void;
+  onImportFolder?: (baseDir: string) => Promise<boolean> | boolean;
   // Electron flow: the desktop main process owns the picker dialog and
   // the import call atomically (`pickAndImport` IPC). The renderer
   // never sees the path or the HMAC token; it only receives the
@@ -623,10 +623,12 @@ export function NewProjectPanel({
     setImportFolderError(null);
     setImportingFolder(true);
     try {
-      await onImportFolder(trimmed);
-      // Success: the parent handler (Home.tsx) navigates to the project,
-      // which provides implicit success feedback by changing the view.
-      // No explicit toast needed here since the navigation is the feedback.
+      const opened = await onImportFolder(trimmed);
+      if (!opened) {
+        setImportFolderError({
+          message: `Open folder failed: ${trimmed}`,
+        });
+      }
     } catch (err) {
       setImportFolderError({
         message: `Open folder failed: ${err instanceof Error ? err.message : 'unknown error'}`,
--- a/apps/web/src/components/NewProjectPanel.tsx
+++ b/apps/web/src/components/NewProjectPanel.tsx
@@ -620,9 +620,17 @@ export function NewProjectPanel({
     if (!onImportFolder) return;
     const trimmed = baseDir.trim();
     if (!trimmed) return;
+    setImportFolderError(null);
     setImportingFolder(true);
     try {
       await onImportFolder(trimmed);
+      // Success: the parent handler (Home.tsx) navigates to the project,
+      // which provides implicit success feedback by changing the view.
+      // No explicit toast needed here since the navigation is the feedback.
+    } catch (err) {
+      setImportFolderError({
+        message: `Open folder failed: ${err instanceof Error ? err.message : 'unknown error'}`,
+      });
     } finally {
       setImportingFolder(false);
     }
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -737,13 +737,14 @@ export function App() {
 
   const handleImportFolder = useCallback(async (baseDir: string) => {
     const result = await importFolderProject({ baseDir });
-    if (!result) return;
+    if (!result) return false;
     setProjects((curr) => [result.project, ...curr.filter((p) => p.id !== result.project.id)]);
     navigate({
       kind: 'project',
       projectId: result.project.id,
       fileName: result.entryFile,
     });
+    return true;
   }, []);
 
   // PR #974: on Electron, the desktop main process owns the picker and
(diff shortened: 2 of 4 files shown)


lefarcen (commented):
Hey @xxiaoxiong 👋 The error handling here is tidy — matches the Electron picker path precisely, clears stale state before each attempt, and moves `setImportingFolder(false)` into `finally` so the button resets correctly even when the import throws. Nice, focused fix.

One thing I noticed in the research: you also have #1487 open against the same area — both PRs touch `NewProjectPanel.tsx` targeting the same missing feedback in the manual-path flow. Looks like you may have created this as a more contained replacement after the review round on #1487. Worth deciding if you'd like to consolidate — if this one supersedes #1487, closing that and proceeding here would help the reviewers focus. Happy to help map the scope either way.

Code-wise, this is clean and ready for the pool reviewer. ✨

apps/web/src/components/NewProjectPanel.tsx
```diff
@@ -620,9 +620,17 @@ export function NewProjectPanel({
     if (!onImportFolder) return;
     const trimmed = baseDir.trim();
     if (!trimmed) return;
+    setImportFolderError(null);
     setImportingFolder(true);
     try {
       await onImportFolder(trimmed);
```
Siri-Ray: The failure path still stays silent here for the main manual import errors. The `onImportFolder` prop is wired to `handleImportFolder` in `App.tsx`, which awaits `importFolderProject({ baseDir })`; that helper returns `null` for non-OK HTTP responses and network/daemon failures instead of throwing, and `handleImportFolder` then just `return`s. That means invalid paths, daemon errors, and offline failures resolve successfully from this component’s point of view, skip this `catch`, do not navigate, and still leave the user with no feedback.

Please make the manual import contract report failure back to this component, for example by having the App-level `handleImportFolder` throw when `importFolderProject` returns `null`, or by changing this prop to return a success/failure result that `NewProjectPanel` can turn into `setImportFolderError(...)`. A focused web component test for a resolved-null import would also lock the bug this PR is trying to fix.

Siri-Ray (changes requested):
@xxiaoxiong Thanks for the focused fix here. I found one remaining issue in the manual folder path: the new error UI still does not fire for the normal failed-import result, because the app-level handler resolves after `importFolderProject` returns `null`.


Siri-Ray:
Maintainer follow-up pushed in e3e2f0ce to address the remaining blocking path.

What changed:
- `onImportFolder` now returns an explicit success boolean for the web manual-path flow.
- `App.handleImportFolder` returns `false` when `importFolderProject()` resolves to `null`, instead of silently returning.
- `NewProjectPanel` now shows the existing folder-import error toast for that resolved-failure case.
- Added a regression test covering a manual folder import that resolves as failed.

Validation run locally:
- `pnpm guard`
- `pnpm typecheck`
- `pnpm --filter @open-design/web test -- NewProjectPanel.test.tsx` (Vitest matched and ran the web suite: 125 files / 1236 tests passed)

Local note: this machine is on Node v22.22.0, so pnpm printed the repo expected Node `~24` engine warning, but the commands completed successfully.

Siri-Ray (approved):
Rechecked after maintainer follow-up e3e2f0ce. The resolved-null manual folder import path now returns an explicit failure result, `NewProjectPanel` surfaces that failure through the existing error toast, and the regression test covers the case that was blocking the earlier review.

CI is green on the updated head, and local validation passed as noted in the comment above.

lefarcen (commented):
Hey @xxiaoxiong, the manual-folder failure case is scoped clearly, and the latest update sounds aligned with the feedback from @Siri-Ray and @qiongyu1999.

Quick PR-body nudge before this goes back through pool/product review: because this changes a UI error surface for a bug fix, could you add (1) a screenshot or GIF of the error message in the From folder tab, and (2) a short bug-fix verification note covering the resolved-null regression test or commands used? That will make release verification much easier.

Thanks — once the body reflects those two bits, the remaining review can stay focused on the fix itself.

PR #1672, opened 2026-05-14
feat(web): enable mid-chat design system switching (#498)

### LINKED ISSUE #498: feat(web): enable mid-chat skill and design system switching via import menu
## Summary

Enable the **"Skills and design systems"** import menu item in `ChatComposer`, allowing users to switch the active design system or skill during a chat session.

## Current state

- The import menu has 6 items; "Skills and design systems" is disabled (SOON badge, no click handler)
- Projects are created with a `skillId` and `designSystemId` — fixed at creation time, no way to change mid-project
- "Link code folder" was recently implemented (#455), proving the pattern for enabling import menu items

## Proposed behavior

Clicking **"Skills and design systems"** opens an inline picker inside the import menu:

### 1. Switch active design system
- Show a compact design system picker (same data source as project creation, grouped by category)
- Selecting a new design system calls `PATCH /api/projects/:id` with the new `designSystemId`
- The next chat message uses the updated `DESIGN.md` in `composeDaemonSystemPrompt`
- A small toast confirms the switch

### 2. Switch active skill
- Show a compact skill picker (grouped by mode)
- Selecting a new skill calls `PATCH /api/projects/:id` with the new `skillId`
- The next chat message uses the updated `SKILL.md` body

**No backend changes needed** — `PATCH /api/projects/:id` already accepts `skillId` and `designSystemId`, and `composeDaemonSystemPrompt` re-reads them on every chat run.

## Technical approach

**Frontend only:**
- `ChatComposer.tsx`: enable the "Skills and design systems" `ImportItem`, add an `onClick` handler
- Inline picker: a two-tab mini-dropdown (Design Systems | Skills) with search/filter
- On select: call `patchProject(id, { designSystemId })` or `patchProject(id, { skillId })`
- Reuse the picker component from the management page proposed in #497

## Open questions

1. **Replace vs. append?** This proposal replaces the active DS/skill. An alternative is keeping the current one as a reference and appending the new one — but that's more complex and increases token usage. Suggest replace-only for v1.
2. **Skill switching mid-project?** Changing skills mid-project can confuse the agent if the previous skill produced artifacts in a different format (e.g., switching from `prototype` to `deck`). Should this show a warning?
3. **Scope of first PR?** Suggest design-system switching only as v1 — it's the most common use case ("I want to see this with a different brand") and has zero risk of format incompatibility.

## Relation

Depends on the management page (#497) for the browse/picker component. Can be implemented as a separate PR after #497 lands.

bulai0408:
## Why

I picked this issue from the `/claim` queue ([open-design#498](https://github.com/nexu-io/open-design/issues/498)). The "Skills and design systems" item in the chat composer's import menu has been disabled with a "Coming soon" badge since launch, even though everything it needs already exists in the daemon: `PATCH /api/projects/:id` accepts `designSystemId` and the daemon re-reads the project's DS on every chat turn. So the only thing in the way of mid-chat DS switching was the front-end picker.

This PR ships the v1 slice the issue author proposed: **design-system switching only**. Skill switching has the cross-mode (prototype ↔ deck) compatibility open question and lands in a follow-up.

## What users will see

- The **Skills and design systems** item in the chat composer Tools popover ("⋮ → Import" section) is no longer disabled.
- Clicking it slides the popover into a scoped picker:
  - Back button + "Choose a design system" title
  - Search box (filters by title / summary / category)
  - **"None — freeform"** entry to clear the project's DS
  - DS list grouped by category, each row shows title, summary, swatches, and an "Active" tag on the currently-bound DS
- Picking an entry PATCHes the project and pops a transient toast ("Design system switched to {title}"). The next chat turn composes with the new `DESIGN.md`.
- If `/api/design-systems` fails to load, the picker shows an explicit `role="alert"` "Couldn't load design systems." banner instead of a misleading empty catalog (the "None — freeform" row stays usable).
- Disabled state is preserved when there is no active project, so the row stays "Coming soon" until a project exists.

## Surface area

- [x] **UI** — new picker view inside the ChatComposer Tools popover
- [ ] Keyboard shortcut
- [ ] CLI / env var
- [ ] API / contract — *reuses existing `PATCH /api/projects/:id`; zero daemon changes*
- [ ] Extension point
- [x] **i18n keys** — adds one new key, `chat.importDesignSystemLoadFailed`, to `i18n/types.ts` + all 18 locale files. The rest of the picker reuses the `chat.importDesignSystem*` vocabulary that was already present but unused.
- [ ] Default behavior change

## Screenshots

Posted in [this comment](https://github.com/nexu-io/open-design/pull/1672#issuecomment-4451094728): import menu entry, picker open, "Active" state, switch, post-switch confirmation.

## Bug fix verification

Not a bug fix.

## Validation

Run on `feat/498-mid-chat-ds-switch` rebased on `upstream/main` @ `d6515643`:

- `pnpm typecheck` (repo-wide) — green
- `pnpm guard` — green (residual JS, test layout, e2e layout, web test layout, tools layout, style policy, design-system token/flag parity all pass)
- `pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/ChatComposer.designSystem.test.tsx` — 5/5 green
- `git diff --check upstream/main...HEAD` — clean

Tests included in this PR:

- `apps/web/tests/components/ChatComposer.designSystem.test.tsx` — 5 cases: PATCH payload shape, success-toast / parent-callback wiring, failure-toast + picker-stays-open path, registry load-failure alert state, disabled state when no project is active, no-op on re-selecting the current system.
- `e2e/ui/chat-design-system-switch.test.ts` — Playwright UI: create prototype project → open import menu → search picker → pick a different DS → re-GET `/api/projects/:id` to confirm the new `designSystemId` landed.


bulai0408:
Closing the test gaps from the original draft body. New commit (`020a94a6`):

- `apps/web/tests/components/ChatComposer.designSystem.test.tsx` — 4 cases. PATCH payload shape, success-toast / parent-callback wiring, failure-toast + picker stays-open path, disabled state when there is no active project, and no-op on a click of the current system.
- `e2e/ui/chat-design-system-switch.test.ts` — Playwright UI: create prototype project → open import menu → search picker → pick a different DS → re-GET `/api/projects/:id` and confirm the new `designSystemId` landed.

Validation run on the new HEAD (`020a94a6`, branched off `upstream/main` @ `53148d52`):

- \`pnpm --filter @open-design/web test -- tests/components/ChatComposer.designSystem.test.tsx\` — 4/4 green
- \`pnpm --filter @open-design/web test -- tests/components/ChatComposer.search.test.tsx tests/components/ChatPane.streaming.test.tsx\` — 13/13 green (no regression in adjacent suites)
- \`pnpm typecheck\` (repo-wide) — green
- \`pnpm guard\` — green
- \`pnpm exec playwright test -c playwright.config.ts --list\` — picks up the new spec under \`e2e/ui/\`

Still on the to-do before flipping ready-for-review: screenshots in the PR body and a manual \`pnpm tools-dev run web\` walk-through. No further source code changes since the v1 feature commit.

bulai0408:
1. new
mac_1778765345193

2. select
mac_1778765364389

3. check
mac_1778765383571

4. change
image

5. check
mac_1778765427878

@qiongyu1999 you can check these screenshots

bulai0408:
Updated to address [Codex's pre-review findings](https://github.com/nexu-io/open-design/issues/1) and the resulting merge conflict against `main`. Branch rebased on `upstream/main` @ `7c9e6202`.

**Conflict resolution**

- `apps/web/src/index.css` had two copies of the `.composer-ds-picker*` rules (one inserted near line 12873 during a polish pass, one appended near the bottom of the file at ~19170 from the original feature commit). The trailing duplicate is dropped; the polished version at line 12873 is the single source of truth now.

**Codex blocker fixes**

- `e2e/ui/chat-design-system-switch.test.ts` — `fetchCurrentProject()` was extracting the **conversation id** out of `/projects/:projectId/conversations/:conversationId` via `url.pathname.split('/').pop()`. The subsequent `/api/projects/:id` GET 404'd, so the spec never actually verified that PATCH landed. Switched to the same `url.pathname.match(/\/projects\/([^/]+)/)` pattern used by `project-management-flows.test.ts` and `entry-configuration-flows.test.ts`. Throws a clear error if the route shape ever drifts again.
- Same file — removed the trailing blank line so `git diff --check origin/main...HEAD` is clean.

**Re-validation on `d41da236`**

- `pnpm guard` — green
- `pnpm typecheck` (repo-wide) — green
- `pnpm --filter @open-design/web exec vitest run -c vitest.config.ts tests/components/ChatComposer.designSystem.test.tsx` — 4/4 green
- `git diff --check` — clean

The branch is now 2 commits ahead of `upstream/main` (feature + tests) with no merge / whitespace / dup issues. Ready for another pass when reviewers have time.

apps/web/src/components/DesignSystemSwitchPicker.tsx
```diff
@@ -0,0 +1,184 @@
...
+}: Props) {
+  const [items, setItems] = useState<DesignSystemSummary[] | null>(null);
+  const [query, setQuery] = useState('');
+  const [pendingId, setPendingId] = useState<string | null | 'none'>(null);
+
+  useEffect(() => {
+    let cancelled = false;
+    void fetchDesignSystems().then((list) => {
+      if (!cancelled) setItems(list);
+    });
```
nettee, line 50: `DesignSystemSwitchPicker` currently can't distinguish `"the registry really has no items"` from `"/api/design-systems failed to load"`. `fetchDesignSystems()` already normalizes non-OK responses to `[]`, and this effect stores that array directly in `items`, so the render path below falls through to the normal picker UI instead of surfacing an error. In the common empty-query case that means the user only sees the `None — freeform` row, which hides a broken integration and makes the catalog look legitimately empty. Please track an explicit load-error state here, render an inline failure/disabled state when the initial fetch fails, and add a regression test that rejects or returns 500 for the initial list load.

nettee (changes requested):
Thanks for pushing this feature forward. I found one blocking issue in the new picker flow that should be fixed before merge.


apps/web/src/components/DesignSystemSwitchPicker.tsx
bulai0408: Fixed in `208476d4`.

Added a discriminated `fetchDesignSystemsResult()` in `providers/registry.ts` that preserves the fetch outcome (`{ ok: true, designSystems }` vs `{ ok: false }`) instead of collapsing failures into `[]`. The picker now tracks an explicit `loadError` state and renders a dedicated `role="alert"` banner — `data-testid="composer-ds-picker-load-error"` with the new localized copy `chat.importDesignSystemLoadFailed` ("Couldn't load design systems.") — in place of the catalog rows when the initial fetch fails. The "None — freeform" row stays clickable in that state so users can still clear their DS while the catalog is unreachable.

The existing `fetchDesignSystems()` is kept as a thin wrapper over the new helper so callers that intentionally want the silent-empty fallback (`NewProjectPanel`, etc.) keep their current contract.

Regression test added as the 5th case in `apps/web/tests/components/ChatComposer.designSystem.test.tsx` — mocks `fetchDesignSystemsResult` to `{ ok: false }`, asserts the alert banner renders with the localized copy and `role="alert"`, asserts the catalog rows are suppressed, and asserts no PATCH / toast / parent-callback fires. The existing four cases were updated to mock `fetchDesignSystemsResult` to match the picker's new dependency.

i18n: new key added to all 18 locale files + `i18n/types.ts` with hand translations.

Re-validation: `pnpm guard`, `pnpm typecheck` (repo-wide), and the 5/5 picker tests all green; `git diff --check` clean.

bulai0408:
@nettee thanks for the catch. Addressed the blocking review feedback in `208476d4` and replied inline at the comment thread — short version:

- `DesignSystemSwitchPicker` no longer accepts the silent-empty fallback from `fetchDesignSystems()`. Picker now consumes a new discriminated `fetchDesignSystemsResult()` helper that returns `{ ok: false }` on non-OK / network failure.
- New `loadError` state renders a `role="alert"` banner (`data-testid="composer-ds-picker-load-error"`) with localized copy `chat.importDesignSystemLoadFailed` in place of the catalog rows. The "None — freeform" row stays clickable so the freeform escape works even when the catalog is unreachable.
- New regression test added: 5th case in `ChatComposer.designSystem.test.tsx` rejects the initial list load and asserts the alert renders + catalog rows suppressed + no PATCH/toast.
- i18n: new key added to `i18n/types.ts` and all 18 locale files with hand translations.

`pnpm guard`, `pnpm typecheck` (repo-wide), `git diff --check`, and the 5/5 picker tests all green on `208476d4`.

Ready for another look when you have a moment.

bulai0408:
@lefarcen @nettee  pr is ready

lefarcen:
Hey @bulai0408! 👋

Thanks for the ping — you've put a lot of careful work into this and it shows.

Two things are still in flight before we can merge:

1. **@nettee re-review** — their earlier review was against the old head (`d41da236`). They need to revisit `208476d4` to confirm the error-state fix looks good. They're already assigned and your earlier @-mention should have reached them.

2. **Product review** — @qiongyu1999 is reviewing the picker interaction (she has the screenshots). Her sign-off is required for UI features before merge.

Nothing blocking on the maintainer coordination side beyond those two. Once both come through, we'll move fast. ✨

apps/web/src/components/DesignSystemSwitchPicker.tsx
```diff
@@ -0,0 +1,203 @@
...
+    const byCategory = new Map<string, DesignSystemSummary[]>();
+    for (const it of matches) {
+      const arr = byCategory.get(it.category) ?? [];
+      arr.push(it);
+      byCategory.set(it.category, arr);
+    }
+    return Array.from(byCategory.entries()).sort(([a], [b]) => a.localeCompare(b));
+  }, [items, query]);
+
+  const showNoneRow = query.trim().length === 0;
```
nettee: When the search box is non-empty, this `showNoneRow` guard hides the `None — freeform` action entirely. That breaks part of the new failure-state contract: the PR body and the resolved review thread both say users should still be able to clear the project's design system while `/api/design-systems` is failing, but after typing any query the `loadError` branch at lines 142-149 is left with no clickable fallback row. Please keep the freeform row visible during `loadError` (or always visible and let it filter independently), and add a regression test that types into the search box before asserting the fallback row is still usable.

nettee (commented):
Thanks for addressing the load-failure case. I found one follow-up in the new picker flow that should be tightened before we rely on this UX contract more broadly.


apps/web/src/components/DesignSystemSwitchPicker.tsx
bulai0408: Fixed in `854e7c20`.

The freeform row is an action (clear the project DS), not a catalog entry, so it no longer hides behind a bare `query.length === 0` guard. New logic in `DesignSystemSwitchPicker.tsx`:

- It filters *independently* — visible whenever the typed query matches its own localized label/summary (`chat.importDesignSystemNone` / `…NoneSub`).
- It is **always** shown while `loadError` is set, so a broken `/api/design-systems` can never strand the user without a reachable fallback, regardless of what they typed.

```ts
const noneRowMatchesQuery = useMemo(() => {
  const needle = query.trim().toLowerCase();
  if (!needle) return true;
  return (
    t('chat.importDesignSystemNone').toLowerCase().includes(needle) ||
    t('chat.importDesignSystemNoneSub').toLowerCase().includes(needle)
  );
}, [query, t]);
const showNoneRow = loadError || noneRowMatchesQuery;
```

Regression test added in `ChatComposer.designSystem.test.tsx` ("keeps the freeform fallback row usable after typing a query when the catalog fails to load"): mounts under `fetchDesignSystemsResult → { ok: false }`, types a non-matching query into the search box, then asserts the `composer-ds-picker-item-none` row is still enabled and still PATCHes `{ designSystemId: null }`.

Validated: `pnpm guard`, `pnpm typecheck` (repo-wide), and `apps/web` picker suite (6/6) all green. Branch rebased onto latest `upstream/main` (`6bf865a4`) before the push.

apps/web/src/components/ChatComposer.tsx
```diff
@@ -820,6 +837,25 @@ export const ChatComposer = forwardRef<ChatComposerHandle, Props>(
...
+      title: string | null,
+    ): Promise<boolean> {
+      if (!projectId) return false;
+      if (designSystemId === currentDesignSystemId) return true;
+      const result = await patchProject(projectId, { designSystemId });
+      if (!result) {
+        onShowToast?.(t('chat.importDesignSystemFailed'));
+        return false;
+      }
+      onActiveDesignSystemChange?.(designSystemId);
```
nettee: `patchProject()` already gives this flow the fully patched `Project`, but the new callback only forwards `designSystemId`, so `ProjectView` reconstructs local state from the stale `project` prop and drops server-owned fields from the PATCH response. In practice the daemon bumps `updatedAt` on project PATCHes, so the client stays immediately stale after a switch and any recent-project / last-updated UI can lag until some later refresh. Please plumb the full patched project back through this callback (or update the parent directly from `result`) instead of rebuilding from `{ ...project, designSystemId }`.

e2e/ui/chat-design-system-switch.test.ts
```diff
@@ -0,0 +1,157 @@
...
+  await page
+    .getByTestId('composer-ds-picker-search')
+    .fill('editorial');
+  await page.getByTestId('composer-ds-picker-item-editorial-noir').click();
+
+  // The composer closes the popover on a successful switch, so the
+  // picker disappears and the project mirrors the new DS.
+  await expect(page.getByTestId('composer-ds-picker')).toHaveCount(0);
+
+  const after = await fetchCurrentProject(page);
```
nettee, line 145: This spec says it verifies that the subsequent chat turn uses the switched design system, but the assertions stop after re-fetching `/api/projects/:id` and never send a turn. That means the test still passes if persistence works but the first post-switch run is composed from stale in-memory state. Please extend the scenario to submit a chat turn and assert the outbound run/request carries `designSystemId: "editorial-noir"` (or otherwise prove the next turn used the switched DS), so the regression boundary matches the behavior the PR body promises.

nettee (commented):
Thanks for the follow-up. I found two merge-safe issues worth tightening before we rely on this flow and its regression coverage more broadly.


apps/web/src/components/ChatComposer.tsx
bulai0408: Fixed in `3fa8f37e`. `onActiveDesignSystemChange` now forwards the full patched `Project` returned by `patchProject()` instead of just `designSystemId`, and `ProjectView` replaces its mirror wholesale (`onProjectChange(updatedProject)`) rather than rebuilding `{ ...project, designSystemId }`. Server-owned fields like `updatedAt` are preserved now. Threaded through ChatComposer → ChatPane → ProjectView; the `ChatComposer.designSystem.test.tsx` assertions moved to `expect.objectContaining({ ..., updatedAt })` so an id-only regression is caught.

e2e/ui/chat-design-system-switch.test.ts
bulai0408: Fixed in `3fa8f37e`. The spec now sends a chat turn after the switch: it stubs `POST /api/runs` (+ the run SSE), submits a prompt via the composer, waits for the run-create request, and asserts the captured body carries `designSystemId: "editorial-noir"`. If the composer ever composes the next turn from stale in-memory state, that assertion fails.

bulai0408:
@nettee @lefarcen fixed

bulai0408:
@qiongyu1999 — product context below. (Screenshots / screen recording still need a GUI capture pass on the maintainer side; the written walkthrough here should unblock the product-direction review in the meantime.)

## Intended user experience

A user is mid-conversation in a project — they've already built several iterations of a prototype or deck with the agent. Today the design system (the brand `DESIGN.md`) is locked at project-creation time; there is no in-session way to change it.

With this PR:

1. In the chat composer they open the **Tools** popover (the control left of the input) → **Import** tab.
2. The **"Skills and design systems"** row — previously disabled with a "Soon" badge — is now active. Clicking it slides the same popover over to an inline design-system picker: no modal, no navigation away from chat.
3. The picker shows a back affordance, a search box, and the workspace's design systems grouped by category — the **same catalogue and data source as the New Project panel**, so it's already familiar. The project's current DS carries an "Active" tag; a **"None — freeform"** row lets the user drop the DS entirely.
4. Picking a row PATCHes the project and pops a transient toast ("Design system switched to …"); the popover closes.
5. **The next chat turn is composed with the new `DESIGN.md`** — so the user can say "now show me this in the Editorial brand" → switch → "redo the hero", and the agent rebuilds against the new brand in the same conversation, keeping all prior context.

Edge states are deliberate: no active project keeps the row disabled; a failed `/api/design-systems` load shows an explicit error with the "None — freeform" clear still reachable; a failed PATCH keeps the picker open with an error toast for retry.

## Product rationale

The friction this removes: **`designSystemId` is fixed at project creation.** The single most common reason a user wants to change it — *"I like this layout, what does it look like in our other brand?"* (issue #498 names this as the primary use case) — currently forces them to **start a brand-new project and abandon the conversation history and iterations** they just built. That is a real break in the core "explore a design" loop.

Everything needed already exists server-side: `PATCH /api/projects/:id` accepts `designSystemId`, and `composeDaemonSystemPrompt` re-reads it on every chat run. The only missing piece was the front-end entry point — the menu item shipped disabled as a placeholder. So this is a low-risk unlock of latent capability, not new system behavior.

Scope is deliberately **design-system switching only** for v1. Skill switching is held back because changing a skill mid-project can strand artifacts produced in an incompatible format (e.g. prototype → deck); that needs a compatibility warning and is tracked as v2. DS switching has no such format-incompatibility risk — it only swaps brand tokens / prose — so it is the safe, high-value first slice.

## Expected impact

- **Who benefits:** anyone iterating on a prototype or deck who wants to compare brands or correct a wrong DS choice at creation — i.e. the central design-exploration audience, not a niche.
- **Behavior change:** purely additive. A previously-disabled menu item becomes functional; nothing existing changes. Projects that never open the picker behave identically.
- **Surface / risk:** front-end only — zero daemon or contract changes, zero new dependencies. The PATCH path is the same one project rename and custom-instructions already use.
- **Reversibility:** fully reversible in-session — re-open the picker and switch back, or pick "None — freeform". No destructive action; existing artifacts are untouched, since the new DS only affects subsequent turns.

Happy to adjust scope or framing if the product call lands differently.

bulai0408:
@nettee could you please refresh your review when you have a moment? After the 2026-05-19 rebase the PR head is now `b20dbdaa`; all CI checks are green, and `qiongyu1999` has approved from the product side. Your 2026-05-17 inline review was against the now-superseded head `854e7c20`, so the PR is still showing `CHANGES_REQUESTED` even though the branch has moved on.

bulai0408:
Rebased onto `upstream/main` @ `590cfcb4` (`fix(daemon): run Trae CLI ACP with yolo (#2856)`).

### Conflict resolution

Three conflicts, all in the prop-plumbing for the new mid-chat DS switcher (commits 1 and 5). All resolved by taking both sides; no behavior or contract changes.

- `apps/web/src/components/ChatComposer.tsx` x2 — upstream added `footerAccessory` to the same props interface where this PR added `currentDesignSystemId` / `onActiveDesignSystemChange` / `onShowToast`. Kept both prop sets in the interface declaration and in the destructured signature.
- `apps/web/src/components/ChatPane.tsx` — same shape (`composerFooterAccessory` from upstream alongside the DS-switcher trio), plus the `../types` import line which now imports both `DesignSystemSummary` (HEAD) and `Project` (this PR's commit `b20dbdaa`).

`ProjectView.tsx` and all 18 locale files + `i18n/types.ts` auto-merged cleanly.

### Review threads

All four review threads from @nettee were already addressed pre-rebase. The fix SHAs have been preserved through the rebase as their new equivalents:

- Distinguish empty registry from `/api/design-systems` load failure (`208476d4` → `3154a291`).
- Keep "None — freeform" reachable when registry load fails and a search query is typed (`854e7c20` → `a75bb373`).
- Forward full patched `Project` from `patchProject()` (`3fa8f37e` → `7f9e1620`).
- e2e spec submits a chat turn and asserts `POST /api/runs` body carries the switched `designSystemId` (also `3fa8f37e` → `7f9e1620`).

### Validation (post-rebase, on `7f9e1620`)

- `pnpm install` — green
- `pnpm guard` — green
- `pnpm typecheck` (repo-wide) — green
- `pnpm --filter @open-design/web test` — **225 files / 2108 tests passed**. First run flaked on an unrelated unhandled rAF callback in `SettingsDialog.execution.test.tsx` → `DesignSystemsSection.tsx#scrollIntoView`; the file itself passes 83/83 in isolation and a clean second run was 2108/2108 green with zero errors. Pre-existing upstream flake, not introduced by this PR.
- `pnpm --filter @open-design/daemon test` — **268 files / 3184 passed | 1 skipped | 4 todo**.
- PR-specific `tests/components/ChatComposer.designSystem.test.tsx` — 6/6 green.

@nettee please take another look.

leilei926524-tech:
Hi @bulai0408!

This PR looks quiet for a bit, so I'm checking in to make sure it doesn't get stuck.

If you are waiting on review, feedback, or a specific unblock from the team, feel free to say so here. If faster coordination helps, you can also drop into our contributors channel in Discord:
https://discord.gg/3C6EWXbdQQ

We want to help move this forward instead of letting it stall.

lefarcen:
Hey @bulai0408 — quick heads-up on where things stand: GitHub is currently showing the branch as **conflicting with `main`** (`mergeable: false`). The conflict likely emerged after your May 25 rebase as new commits landed upstream — which is why @nettee's re-review hasn't kicked off yet (reviewers typically wait for a clean diff).

Once you rebase onto the latest `main` and push, @nettee and @Eli-tangerine can do their final pass on a stable diff. Everything else is in good shape — CI is green and the feature logic is solid. 🙌

bulai0408:
Rebased onto latest upstream/main and pushed head e471c0448. Merge conflicts are gone, the PR is mergeable again, and the CI/visual checks on the new head are green. @nettee @Eli-tangerine please take another look when you have a moment.

bulai0408:
Rebased onto the latest `nexu-io/main` @ `9f09d1b64` and pushed head `02a589154e`.

Current GitHub state after recompute:
- `mergeable=MERGEABLE`
- `mergeStateStatus=BLOCKED` (review gate only)
- `reviewDecision=CHANGES_REQUESTED`
- Checks: all terminal checks are `SUCCESS` or `SKIPPED` (`Browser tests`, `Build workspaces`, `Web workspace tests`, `Workspace unit tests`, `Validate Nix flake`, `Preflight`, `Validate workspace`, visual capture/verify all succeeded; daemon/runtime/placeholder jobs skipped as expected).

Review threads: the old nettee review threads are all marked resolved on GitHub. Review requests are active for @nettee, @Eli-tangerine, and @elihahah666, so the remaining blocker is fresh reviewer action on the clean, mergeable head.

nettee (approved):
@bulai0408 I re-reviewed the final head and verified the picker flow now covers the earlier gaps: the load-failure path stays observable and still leaves the freeform fallback reachable, the successful PATCH propagates the full updated project back up to the parent mirror, and the UI regression test now proves the next run request carries the switched design system. The focused unit coverage and end-to-end scenario line up with the behavior this PR promises. Nice cleanup on the follow-up fixes.

PR #1675, opened 2026-05-14
fix: gate chat token params by model family

### LINKED ISSUE #361: OpenAI-compatible provider option missing; custom OpenAI endpoint is routed through Anthropic flow

drjaat:
Fixes #361

## Why

Azure chat-completions requests can use arbitrary deployment aliases in both the default `/openai/deployments/<deployment>` path and Azure OpenAI-compatible `/openai/v1`, so deriving `max_tokens` vs `max_completion_tokens` from the configured `model` string is unreliable. A GPT-5/o-series deployment named `prod` needs to recover based on Azure actual unsupported-parameter response.

## What users will see

Azure BYOK chat and Settings connection tests first try the legacy `max_tokens` field, preserving older deployment compatibility. When Azure rejects `max_tokens` for GPT-5/o-series deployments, Open Design retries once with `max_completion_tokens`, including for `/openai/v1` deployment aliases. Settings connection-test latency reflects the final provider response after that retry, including failed retry responses.

## Surface area

- [ ] **New top-level dependency** — adding any new entry to the **root** `package.json` (`dependencies` or `devDependencies`); workspace-package `package.json` files are out of scope. Include a paragraph on what we get vs. what bytes we ship (see `CONTRIBUTING.md` -> Code style)

## Screenshots

N/A

## Bug fix verification

- Test path that reproduces the bug: `apps/daemon/tests/proxy-routes.test.ts` and `apps/daemon/tests/connection-test.test.ts`
- Red/green check: the Azure `/openai/v1` alias retry specs using `model: "prod"` went red before the fix and green after the retry fallback was applied.
- Latency follow-up: the Azure retry latency regression returned `25ms` before the fix and `100ms` after `latencyMs` was refreshed after the retry fetch.
- Failed-retry latency follow-up: the new Azure failed-retry regression covers a 25ms first response plus 75ms retry response that returns 500 and asserts the returned `latencyMs` is `100`.
- Rebase follow-up: rebased onto `origin/main` at `b8cddc42` and refreshed workspace links with `pnpm install --config.confirmModulesPurge=false`.
- Additional regressions keep generic OpenAI-compatible hosts like DeepSeek on `max_tokens`, keep Azure `gpt-4o` deployment paths on `max_tokens`, and keep deployment-mode aliases on the same unsupported-parameter retry fallback.

## Validation

- `pnpm exec vitest run -c vitest.config.ts tests/connection-test.test.ts` (71 passed)
- `pnpm exec vitest run -c vitest.config.ts tests/proxy-routes.test.ts tests/connection-test.test.ts` (147 passed)
- `pnpm guard`
- `pnpm typecheck`

### DIFF
--- a/apps/daemon/src/chat-routes.ts
+++ b/apps/daemon/src/chat-routes.ts
@@ -1,6 +1,7 @@
 import type { Express } from 'express';
 import type { RouteDeps } from './server-context.js';
 import { seedProviderIfMissing } from './media-config.js';
+import { buildOpenAIChatTokenParam } from './openai-chat-token-params.js';
 import {
   BYOK_SENSEAUDIO_TOOLS,
   executeGenerateImage,
@@ -756,8 +757,10 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
     const payload: any = {
       model,
       messages: payloadMessages,
-      max_tokens:
+      ...buildOpenAIChatTokenParam(
+        model,
         typeof maxTokens === 'number' && maxTokens > 0 ? maxTokens : 8192,
+      ),
       stream: true,
     };
 
@@ -869,8 +872,10 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
     const payload = {
       ...(usesVersionedOpenAIPath ? { model } : {}),
       messages: payloadMessages,
-      max_tokens:
+      ...buildOpenAIChatTokenParam(
+        model,
         typeof maxTokens === 'number' && maxTokens > 0 ? maxTokens : 8192,
+      ),
       stream: true,
     };
 
--- a/apps/daemon/src/chat-routes.ts
+++ b/apps/daemon/src/chat-routes.ts
@@ -1,7 +1,12 @@
 import type { Express } from 'express';
 import type { RouteDeps } from './server-context.js';
 import { seedProviderIfMissing } from './media-config.js';
-import { buildOpenAIChatTokenParam } from './openai-chat-token-params.js';
+import {
+  buildLegacyMaxTokensParam,
+  buildMaxCompletionTokensParam,
+  buildOpenAIChatTokenParam,
+  isUnsupportedMaxTokensError,
+} from './openai-chat-token-params.js';
 import {
   BYOK_SENSEAUDIO_TOOLS,
   executeGenerateImage,
@@ -869,40 +874,67 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
       payloadMessages.unshift({ role: 'system', content: systemPrompt });
     }
 
+    const effectiveMaxTokens =
+      typeof maxTokens === 'number' && maxTokens > 0 ? maxTokens : 8192;
     const payload = {
       ...(usesVersionedOpenAIPath ? { model } : {}),
       messages: payloadMessages,
-      ...buildOpenAIChatTokenParam(
-        model,
-        typeof maxTokens === 'number' && maxTokens > 0 ? maxTokens : 8192,
-      ),
+      ...buildLegacyMaxTokensParam(effectiveMaxTokens),
+      stream: true,
+    };
+    const retryPayload = {
+      ...(usesVersionedOpenAIPath ? { model } : {}),
+      messages: payloadMessages,
+      ...buildMaxCompletionTokensParam(effectiveMaxTokens),
       stream: true,
     };
 
     const sse = createSseResponse(res);
     sse.send('start', { model });
     try {
-      const response = await fetch(url, {
+      const requestInit = {
         method: 'POST',
         headers: {
           'Content-Type': 'application/json',
           'api-key': apiKey,
         },
+        redirect: 'error' as const,
+      };
+      let response = await fetch(url, {
+        ...requestInit,
         body: JSON.stringify(payload),
-        redirect: 'error',
       });
 
       if (!response.ok) {
-        const errorText = await response.text();
-        console.error(
-          `[proxy:azure] upstream error: ${response.status} ${redactAuthTokens(errorText)}`,
-        );
-        sendProxyError(sse, `Upstream error: ${response.status}`, {
-          code: proxyErrorCode(response.status),
-          details: errorText,
-          retryable: response.status === 429 || response.status >= 500,
-        });
-        return sse.end();
+        let errorText = await response.text();
+        if (
+          response.status === 400 &&
+          isUnsupportedMaxTokensError(errorText)
+        ) {
+          console.warn(
+            `[proxy:azure] retrying request with max_completion_tokens deployment=${model}`,
+          );
+          response = await fetch(url, {
+            ...requestInit,
+            body: JSON.stringify(retryPayload),
+          });
+          if (response.ok) {
+            errorText = '';
+          } else {
+            errorText = await response.text();
+          }
+        }
+        if (!response.ok) {
+          console.error(
+            `[proxy:azure] upstream error: ${response.status} ${redactAuthTokens(errorText)}`,
+          );
+          sendProxyError(sse, `Upstream error: ${response.status}`, {
+            code: proxyErrorCode(response.status),
+            details: errorText,
+            retryable: response.status === 429 || response.status >= 500,
+          });
+          return sse.end();
+        }
       }
 
       let ended = false;
--- a/apps/daemon/src/connectionTest.ts
+++ b/apps/daemon/src/connectionTest.ts
@@ -40,7 +40,12 @@ import {
   cursorAuthGuidance,
   probeAgentAuthStatus,
 } from './runtimes/auth.js';
-import { buildOpenAIChatTokenParam } from './openai-chat-token-params.js';
+import {
+  buildLegacyMaxTokensParam,
+  buildMaxCompletionTokensParam,
+  buildOpenAIChatTokenParam,
+  isUnsupportedMaxTokensError,
+} from './openai-chat-token-params.js';
 import type { AgentCliEnvPrefs } from './app-config.js';
 import type { RuntimeAgentDef } from './runtimes/types.js';
 import {
@@ -516,6 +521,7 @@ interface ProviderCallShape {
   headers: Record<string, string>;
   body: unknown;
   extractText: (data: unknown) => string;
+  retryBodyOnUnsupportedMaxTokens?: unknown;
 }
 
 function buildProviderCall(input: ProviderTestRequest): ProviderCallShape {
@@ -601,10 +607,19 @@ function buildProviderCall(input: ProviderTestRequest): ProviderCallShape {
         },
         body: {
           ...(usesVersionedOpenAIPath ? { model } : {}),
-          ...buildOpenAIChatTokenParam(model, PROVIDER_MAX_TOKENS),
+          ...(usesVersionedOpenAIPath
+            ? buildOpenAIChatTokenParam(model, PROVIDER_MAX_TOKENS)
+            : buildLegacyMaxTokensParam(PROVIDER_MAX_TOKENS)),
           messages: [{ role: 'user', content: SMOKE_PROMPT }],
           stream: false,
         },
+        retryBodyOnUnsupportedMaxTokens: usesVersionedOpenAIPath
+          ? undefined
+          : {
+              messages: [{ role: 'user', content: SMOKE_PROMPT }],
+              stream: false,
+              ...buildMaxCompletionTokensParam(PROVIDER_MAX_TOKENS),
+            },
         extractText: extractOpenAIMessageText,
       };
     }
@@ -727,14 +742,58 @@ export async function testProviderConnection(
     );
     if (modelError) return modelError;
 
-    const response = await fetch(call.url, {
+    const requestInit = {
       method: 'POST',
       headers: call.headers,
-      body: JSON.stringify(call.body),
       signal: controller.signal,
-      redirect: 'error',
+      redirect: 'error' as const,
+    };
+    let response = await fetch(call.url, {
+      ...requestInit,
+      body: JSON.stringify(call.body),
     });
     const latencyMs = Date.now() - start;
+    if (
+      !response.ok &&
+      call.retryBodyOnUnsupportedMaxTokens !== undefined
+    ) {
+      let detailText = '';
+      try {
+        detailText = await response.text();
+      } catch {
+        detailText = '';
+      }
+      if (response.status === 400 && isUnsupportedMaxTokensError(detailText)) {
+        console.warn(
+          `[test:provider] ${input.protocol} ${validated.parsed.hostname} model=${input.model} → retrying with max_completion_tokens`,
+        );
+        response = await fetch(call.url, {
+          ...requestInit,
+          body: JSON.stringify(call.retryBodyOnUnsupportedMaxTokens),
+        });
+      } else {
+        const redactedDetail = redactSecrets(detailText.slice(0, 240), [
+          input.apiKey,
+        ]);
+        const kind = statusToKind(response.status, redactedDetail);
+        const detail =
+          redactedDetail ||
+          (response.status === 404
+            ? 'HTTP 404 from provider; check the Base URL path.'
+            : '');
+        console.warn(
+          `[test:provider] ${input.protocol} ${validated.parsed.hostname} model=${input.model} → ${response.status} in ${latencyMs}ms (${kind})${detail ? ` ${detail}` : ''}`,
+        );
+        return {
+          ok: false,
+          kind,
+          latencyMs,
+          model,
+          status: response.status,
+          detail,
+        };
+      }
+    }
     if (response.ok) {
       let data: unknown;
       let rawText = '';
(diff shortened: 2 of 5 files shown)


lefarcen:
Heads-up: PR #1090 is also open in this area. Both PRs touch `apps/daemon/src/connectionTest.ts` and are aimed at the Azure/OpenAI-compatible `max_tokens` vs `max_completion_tokens` issue from #361; this PR appears to take the more targeted model-family approach and adds DeepSeek-compatible regression coverage.

Sharing the link so the reviewer team can compare/consolidate the two approaches instead of reviewing them in isolation.

lefarcen (commented):
Hi @drjaat! Thanks for tightening this fix around the actual model-family behavior — the Summary and Verification sections make the intent and regression coverage easy to follow.

One quick PR-body cleanup before review: could you add the new `## Surface area` checklist and tick the applicable backend/API-contract area for the chat-completions payload change? The existing Summary already covers the why/what, and Verification is covered; this just helps reviewers and release validation see exactly which surface changed.

Appreciate the focused follow-up here — the DeepSeek-compatible angle is useful context. 🙌

apps/daemon/src/openai-chat-token-params.ts
```diff
@@ -0,0 +1,22 @@
+export function usesMaxCompletionTokens(model: string): boolean {
+  const normalized = model.trim().toLowerCase();
+  // Keep the generic OpenAI-compatible path on max_tokens by default. Only
+  // switch the newer GPT-5/o-series chat-completions families that reject the
+  // legacy field.
+  return (
+    /^gpt-5(?:[.-]|$)/.test(normalized) ||
+    /^o1(?:[.-]|$)/.test(normalized) ||
+    /^o3(?:[.-]|$)/.test(normalized) ||
+    /^o4(?:[.-]|$)/.test(normalized)
```
mrcfps, line 10: `usesMaxCompletionTokens()` is correct for OpenAI-style requests where `model` is the real model id, but it breaks the default Azure `/openai/deployments/<model>` path because `model` is the user’s deployment name there, not the underlying family. A GPT-5 deployment named `prod` or `chat` will still fall through to `max_tokens`, so the main Azure bug this PR is trying to fix remains reproducible for the common aliased-deployment flow. The rest of the repo already treats this field as a deployment name (`apps/web/src/i18n/locales/en.ts` and `apps/daemon/src/memory-llm.ts`), and the new Azure regressions only pass because they use `model: 'gpt-5.4'` as if the deployment name matched the family. Please stop deriving the token field from deployment-mode Azure `model` strings; either thread the actual model family through explicitly for Azure requests, or add a deployment-mode fallback/retry that swaps to `max_completion_tokens` when Azure rejects `max_tokens`.

mrcfps (changes requested):
@drjaat Thanks for tightening the OpenAI-compatible token-param split and for adding targeted regression coverage here. I found one blocking gap in the Azure deployment-mode path: the new helper now decides between `max_tokens` and `max_completion_tokens` from the deployment name, so aliased GPT-5/o-series deployments still send the legacy field and the main Azure fix stays broken.


apps/daemon/src/openai-chat-token-params.ts
drjaat: Implemented a deployment-mode fallback instead of inferring the token field from Azure deployment aliases. `/openai/deployments/<name>` now sends `max_tokens` first, retries once with `max_completion_tokens` only on the unsupported-parameter response, and keeps the model-family heuristic only for Azure OpenAI-compatible `/openai/v1` requests. Added focused regressions for the retry path plus the older Azure `gpt-4o` and DeepSeek-compatible cases, and re-ran `pnpm exec vitest run -c vitest.config.ts tests/proxy-routes.test.ts tests/connection-test.test.ts` (106 passed).

apps/daemon/src/chat-routes.ts
```diff
@@ -871,26 +879,64 @@ export function registerChatRoutes(app: Express, ctx: RegisterChatRoutesDeps) {
...
+          )
+        : buildLegacyMaxTokensParam(
+            typeof maxTokens === 'number' && maxTokens > 0 ? maxTokens : 8192,
+          )),
       stream: true,
     };
+    const retryPayload = usesVersionedOpenAIPath
+      ? null
+      : {
+          ...(usesVersionedOpenAIPath ? { model } : {}),
```
lefarcen: P3 Dead spread: `usesVersionedOpenAIPath` is `false` in this `else` branch, so `...(usesVersionedOpenAIPath ? { model } : {})` always expands to `{}`. Can simplify to just omit the spread entirely — `retryPayload` for deployment-mode paths doesn't carry a `model` key anyway.

lefarcen (commented):
Hey @drjaat, the retry-based approach in this revision is the right call — starting with `max_tokens` and only upgrading to `max_completion_tokens` when Azure explicitly rejects it via the 400 payload works cleanly for any deployment alias, which was exactly the gap @mrcfps flagged on the previous head. The connection-test path in `connectionTest.ts` is wired up consistently via `retryBodyOnUnsupportedMaxTokens`, so the Settings dialog keeps the same behavior. One P3 nit below.

@mrcfps — the aliased-deployment concern looks addressed here; the code no longer infers the token param from the deployment name at all. Happy for you to re-check when you get a moment.

apps/daemon/src/openai-chat-token-params.ts
```diff
@@ -0,0 +1,43 @@
...
```
mrcfps, line 10: Azure `/openai/v1` still treats `model` as the deployment name, not the underlying model family. Microsoft's own v1 examples use `model="YOUR-DEPLOYMENT-NAME"`, so an aliased GPT-5/o-series deployment like `prod` will fail every regex in this helper and keep sending `max_tokens`. That means both the Azure streaming path (`apps/daemon/src/chat-routes.ts`) and the Settings connection-test path (`apps/daemon/src/connectionTest.ts`) can still reproduce the original 400 whenever the deployment name does not encode `gpt-5`/`o*`; the new regressions only stay green because they pass `model: "gpt-5.4"` as if the deployment name matched the family. Please stop inferring the token field from Azure v1 `model` values—either apply the same unsupported-parameter retry to `usesVersionedOpenAIPath` requests, or thread the real model family/param choice in separately, and add alias-based coverage in both new test blocks.

mrcfps (changes requested):
@drjaat Thanks for iterating on the Azure retry path and for adding focused regression coverage here. I found one remaining blocking gap in the Azure `/openai/v1` path, so I've left the concrete fix suggestion inline below.


apps/daemon/src/openai-chat-token-params.ts
drjaat: Fixed in 4cb41103.

Azure `/openai/v1` now uses the same unsupported-parameter fallback as deployment-mode Azure requests: it sends `max_tokens` first with the deployment alias left in `model`, then retries once with `max_completion_tokens` only when Azure returns the unsupported-parameter 400 for `max_tokens`.

Added alias-based coverage in both paths using `model: "prod"`:
- proxy streaming verifies the first Azure v1 request carries `max_tokens`, then the retry carries `model` plus `max_completion_tokens`.
- Settings connection tests verify the same Azure v1 alias retry behavior.

Validation: `pnpm exec vitest run -c vitest.config.ts tests/proxy-routes.test.ts tests/connection-test.test.ts` (106 passed) and `git diff --check`.

apps/daemon/src/chat-routes.ts
drjaat: Fixed in 4cb41103.

The Azure retry payload no longer has the dead conditional spread from the deployment-mode branch. The retry body is now explicit: it includes `model` only for Azure `/openai/v1` requests and omits it for deployment-mode paths.

Validation: `pnpm exec vitest run -c vitest.config.ts tests/proxy-routes.test.ts tests/connection-test.test.ts` (106 passed) and `git diff --check`.

apps/daemon/src/connectionTest.ts
```diff
@@ -599,14 +612,58 @@ export async function testProviderConnection(
...
-      body: JSON.stringify(call.body),
       signal: controller.signal,
-      redirect: 'error',
+      redirect: 'error' as const,
+    };
+    let response = await fetch(call.url, {
+      ...requestInit,
+      body: JSON.stringify(call.body),
     });
     const latencyMs = Date.now() - start;
```
mrcfps: `latencyMs` is captured before the optional Azure retry, but the rest of this function reuses that same value even after `fetch(call.url, { ...requestInit, body: JSON.stringify(call.retryBodyOnUnsupportedMaxTokens) })` runs. When the new fallback succeeds, the Settings connection test will under-report the real round-trip time by roughly one full upstream request, which makes the retry path harder to observe and can skew timeout/SLO debugging from the UI response and logs. Please recompute `latencyMs` after the retry finishes (or derive it immediately before each return/log) so the reported latency reflects the final response the user actually saw.

mrcfps (commented):
@drjaat Thanks for extending the Azure retry path to both deployment-mode and `/openai/v1` aliases — the new alias regressions cover the main bug well. I found one merge-safe follow-up in `connectionTest.ts` around how the retry path reports latency.


apps/daemon/src/connectionTest.ts
drjaat: Fixed in 00f86502.

`testProviderConnection` now refreshes `latencyMs` after the optional Azure unsupported-`max_tokens` retry fetch completes, so the Settings response and success log use the elapsed time for the final provider response instead of only the first failed attempt.

Added a regression that mocks the retry path with a 25ms first response plus a 75ms retry response and asserts the returned `latencyMs` is 100ms. Before the fix, that test returned 25ms.

Validation:
- `pnpm exec vitest run -c vitest.config.ts tests/connection-test.test.ts` (70 passed)
- `pnpm exec vitest run -c vitest.config.ts tests/proxy-routes.test.ts tests/connection-test.test.ts` (107 passed)
- `pnpm guard`
- `pnpm typecheck`
- `git diff --check`

apps/daemon/src/connectionTest.ts
```diff
@@ -599,14 +612,59 @@ export async function testProviderConnection(
...
+      }
+      if (response.status === 400 && isUnsupportedMaxTokensError(detailText)) {
+        console.warn(
+          `[test:provider] ${input.protocol} ${validated.parsed.hostname} model=${input.model} → retrying with max_completion_tokens`,
+        );
+        response = await fetch(call.url, {
+          ...requestInit,
+          body: JSON.stringify(call.retryBodyOnUnsupportedMaxTokens),
+        });
+        latencyMs = Date.now() - start;
```
mrcfps, line 771: `latencyMs` is only refreshed on the successful retry path here. If the second Azure request also comes back non-2xx, the code falls through to the shared failure handling below with the stale value from the first attempt, so Settings under-reports how long the user actually waited by roughly one upstream round-trip. That makes retry failures harder to debug from the returned latency and the warning log. Please recompute `latencyMs` immediately after the retry `fetch(...)` returns, before branching on `response.ok`, so both successful and failed retries report the final end-to-end timing. I’d also extend the new latency regression with a retry-fails case to keep that behavior covered.

mrcfps (commented):
@drjaat Thanks for extending the Azure unsupported-parameter retry across both chat proxying and the Settings smoke test paths — the alias-based regressions cover the main bug well. I found one merge-safe follow-up in `apps/daemon/src/connectionTest.ts` around how the retry path reports latency.


lefarcen:
Hey @drjaat — good news: mrcfps's latest review on this head is non-blocking (the retry approach across both the chat proxy and Settings connection test is accepted). Two things standing between this and merge:

1. **Merge conflict** — the PR is currently DIRTY against `main` and needs a rebase before it can land.
2. **Latency follow-up** — mrcfps's inline note on `connectionTest.ts`: `latencyMs` isn't refreshed after the retry `fetch()` returns, so a failed retry would under-report the real round-trip time in logs and the returned response. A one-liner recompute after the retry call + a retry-fails test case would close that out (non-blocking, but worth doing while you're in there).

A rebase + those two small fixes should get this across the line. Let us know if anything's unclear — happy to help scope the latency change.

mrcfps (approved):
@drjaat I re-checked the Azure retry flow on this head: both deployment-mode and `/openai/v1` alias requests now start with `max_tokens`, retry once on the Azure unsupported-parameter 400, and the new regression coverage includes both successful and failed retry latency reporting. Thanks for sticking with the follow-ups here — this looks ready to merge from my side. 🙌


drjaat:
Rebased onto current `origin/main` (`b8cddc42`) and force-with-lease pushed the PR branch at `756482d7`.

Also added the retry-fails latency coverage requested inline. The existing implementation refreshes `latencyMs` immediately after the retry fetch returns, and the new regression covers the failed-retry path by asserting a 25ms first response plus 75ms failed retry reports `latencyMs: 100`.

Validation:
- `pnpm install --config.confirmModulesPurge=false`
- `pnpm exec vitest run -c vitest.config.ts tests/proxy-routes.test.ts tests/connection-test.test.ts` (147 passed)
- `pnpm guard`
- `pnpm typecheck`
- `git diff --check`

apps/daemon/src/connectionTest.ts
drjaat: Fixed in 756482d7.

The implementation already refreshes `latencyMs` immediately after the retry `fetch(...)` returns, before the shared success/failure handling. I added the missing failed-retry regression to pin that behavior: the test simulates a 25ms unsupported-`max_tokens` response followed by a 75ms retry response that returns 500, and asserts the final result reports `latencyMs: 100`.

Validation:
- `pnpm exec vitest run -c vitest.config.ts tests/connection-test.test.ts` (71 passed)
- `pnpm exec vitest run -c vitest.config.ts tests/proxy-routes.test.ts tests/connection-test.test.ts` (147 passed)
- `pnpm guard`
- `pnpm typecheck`
- `git diff --check`

lefarcen:
Hey @drjaat — the CI failure showing up on your latest push isn't a test failure. It's the **fork PR workflow approval gate** that requires a maintainer to authorize CI runs for PRs from external forks before the actual test suite kicks off.

Nothing wrong with the code — a maintainer just needs to click "approve and run" on the Actions page for this workflow run. I'll flag it for the team so it doesn't block you.

leilei926524-tech:
Thanks for the model-family token param fix in #1675. Two follow-up issues that seem related:
- #1709 finish_reason / max_tokens continuation bug
- #1928 Could not start Claude Code: exit 1 / not logged in
#1709 is the closer fit on model/runtime response behavior.

lefarcen:
Thanks for connecting these, @leilei926524-tech. 👀

I agree that #1709 is the closer follow-up to this PR: #1675 fixed which token-limit parameter we send to OpenAI-compatible/Azure hosts, while #1709 is about what we do after a response is actually truncated (`finish_reason: "length"` / continuation). #1928 looks separate to me — that thread is about Claude Code startup/authentication rather than model runtime token handling.

Since #1675 has already merged, keeping the next debugging/implementation work in #1709 should give that follow-up the cleanest context.

leilei926524-tech:
Hi @drjaat — welcome to Open Design contributors. 🎉

Your first merged PR (#1675) just landed — that puts you in a small group of people who've actually shipped code into OD. We have a Discord **#contributors** channel where the rest of the crew hangs out, and we'd love for you to drop in:

Why it's worth joining:

PR #1676, opened 2026-05-14
Reject reserved project paths in artifact lint

### LINKED ISSUE #1665: Orbit follow-up preview artifacts can save invalid reserved-path references
## Summary

A follow-up artifact generated inside an Orbit project can be saved successfully and marked complete, but opening it fails with 400 Bad Request.

In the reproduced case, the artifact is a wrapper HTML page that embeds the real live artifact via an iframe. The generated iframe src points into `.live-artifacts/...`, but the project file API explicitly blocks that directory as a reserved internal path.

So the saved artifact is invalid by construction: the system accepts it, but its preview target is guaranteed to fail.

## Why this matters

This should not rely on the agent "guessing" the right internal path shape.

Internal storage layout such as `.live-artifacts` is an implementation detail. Agents should not need to know it, and user-facing artifacts should not be allowed to reference it directly.

## Repro

1. Run an Orbit digest that successfully registers a live artifact.
2. In the same Orbit project, ask the agent to fix a preview/encoding issue and generate a follow-up preview artifact.
3. Let it create a wrapper HTML artifact such as `orbit-github-daily-digest-preview.html`.
4. Open that follow-up artifact in Preview.

## Actual result

Preview fails with:

`400 Bad Request`
`{"error":{"code":"BAD_REQUEST","message":"Error: reserved project path"}}`

The generated wrapper HTML contains an iframe like:

`<iframe src=".live-artifacts/<artifact-id>/index.html"></iframe>`

Directly requesting that target through the project file route reproduces the failure:

`GET /api/projects/:projectId/files/.live-artifacts/:artifactId/index.html`
`-> 400 Bad Request`
`-> Error: reserved project path`

## Expected result

One of these should happen instead:

- the artifact opens successfully via a supported preview path, or
- Open Design rejects the artifact before saving it, with a clear validation error

## Root cause

The project file layer explicitly reserves `.live-artifacts` and blocks reads through the normal project file API.

Relevant code:

- `apps/daemon/src/projects.ts`
  - `RESERVED_PROJECT_FILE_SEGMENTS = new Set(['.live-artifacts'])`
- `apps/daemon/src/project-routes.ts`
  - `GET /api/projects/:id/files/*`

At the same time, the daemon already exposes a proper live artifact preview route:

- `/api/live-artifacts/:artifactId/preview`

So the generated follow-up artifact is using the wrong access path.

## Why this is a product bug, not only an agent mistake

The issue is not just that the agent generated a bad reference.

The system currently allows a user-facing artifact to be saved even though it references a path that the server is guaranteed to reject. That means Open Design is leaking internal storage structure into agent-authored output without enforcing a safe boundary.

At least one layer should prevent this class of failure.

## Suggested fix

Open Design should prevent this at the product level rather than depending on the agent to infer correct internal paths.

1. Reject saved artifacts that reference reserved internal project paths such as `.live-artifacts`, `.od`, `.tmp`, or other non-user-facing storage directories.
2. Update agent guidance so follow-up preview artifacts never reference internal storage paths directly.
3. Provide a canonical supported embed/preview mechanism for existing live artifacts, using the existing `/api/live-artifacts/:artifactId/preview` route or an explicit artifact-embed helper.
4. Optionally surface a clearer validation error in the UI when an artifact contains blocked internal references.

## Additional note

The underlying Orbit run itself succeeded and registered a valid live artifact. The failure is specific to the later follow-up preview artifact, not the original Orbit digest.

Image

mzl2233:
This adds artifact lint coverage for HTML that links directly into reserved project storage paths such as `.live-artifacts`, `.od`, or `.tmp`. Those references now produce a P0 finding before save/preview, while the supported `/api/live-artifacts/:id/preview` route remains allowed. Fixes #1665.

## Why
Orbit follow-up preview artifacts could save successfully while embedding `.live-artifacts/...`, then fail at preview time because that path is intentionally blocked by the project file layer.

## What users will see
Artifacts that reference internal storage paths are flagged with a clear lint finding instead of being accepted silently and failing later.

## Validation
- `npx tsc -p apps/daemon/tsconfig.tests.json --noEmit --pretty false` passed.
- `pnpm --filter @open-design/daemon test -- apps/daemon/tests/lint-artifact.test.ts` could not run because `vitest` was not installed in the checkout after dependency installation timed out in this environment.

## Surface area
- [x] Daemon / API
- [x] Tests
- [ ] Web UI
- [ ] Desktop
- [ ] CLI flags
- [ ] Environment variables
- [ ] i18n keys
- [ ] Root package dependencies

### DIFF
--- a/apps/daemon/src/lint-artifact.ts
+++ b/apps/daemon/src/lint-artifact.ts
@@ -127,6 +127,17 @@ export function lintArtifact(rawHtml: unknown): LintFinding[] {
   const html = rawHtml.replace(/<!--[\s\S]*?-->/g, '');
   const lower = html.toLowerCase();
 
+  const reservedPath = detectReservedProjectPathReference(html);
+  if (reservedPath) {
+    out.push({
+      severity: 'P0',
+      id: 'reserved-project-path',
+      message: `Artifact references reserved internal project path "${reservedPath.segment}".`,
+      fix: 'Use a public preview/embed route instead of linking to internal project storage paths such as .live-artifacts, .od, or .tmp.',
+      snippet: clip(reservedPath.snippet),
+    });
+  }
+
   // ── P0-1: purple gradient backgrounds ─────────────────────────────
   for (const hex of PURPLE_HEXES) {
     const re = new RegExp(
@@ -544,6 +555,35 @@ function clip(s: string): string {
   if (!s) return '';
   const trimmed = s.replace(/\s+/g, ' ').trim();
   return trimmed.length > 200 ? trimmed.slice(0, 197) + '…' : trimmed;
+}
+
+const RESERVED_PROJECT_PATH_SEGMENTS = ['.live-artifacts', '.od', '.tmp'];
+
+function detectReservedProjectPathReference(html: string): { segment: string; snippet: string } | undefined {
+  const attrRe = /\b(?:src|href)\s*=\s*([\"'])(.*?)\1/gi;
+  let match: RegExpExecArray | null;
+  while ((match = attrRe.exec(html)) !== null) {
+    const value = match[2] || '';
+    for (const segment of RESERVED_PROJECT_PATH_SEGMENTS) {
+      if (referencesReservedProjectPath(value, segment)) {
+        return { segment, snippet: match[0] };
+      }
+    }
+  }
+  return undefined;
+}
+
+function referencesReservedProjectPath(value: string, segment: string): boolean {
+  const decoded = safeDecodeURIComponent(value);
+  return decoded === segment || decoded.startsWith(`${segment}/`) || decoded.includes(`/${segment}/`);
+}
+
+function safeDecodeURIComponent(value: string): string {
+  try {
+    return decodeURIComponent(value);
+  } catch {
+    return value;
+  }
 }
 
 function escapeRe(s: string): string {
--- a/apps/daemon/tests/lint-artifact.test.ts
+++ b/apps/daemon/tests/lint-artifact.test.ts
@@ -8,6 +8,26 @@ function requiredFinding(findings: LintFinding[], id: string): LintFinding {
   return hit;
 }
 
+describe('reserved-project-path', () => {
+  it('flags artifacts that iframe live artifact storage directly', () => {
+    const html = `<iframe src=".live-artifacts/demo/index.html"></iframe>`;
+    const findings = lintArtifact(html);
+    const hit = requiredFinding(findings, 'reserved-project-path');
+    expect(hit.severity).toBe('P0');
+  });
+
+  it('flags other reserved internal project path references', () => {
+    const findings = lintArtifact(`<a href="assets/.tmp/render.html">Preview</a>`);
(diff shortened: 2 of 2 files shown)


Siri-Ray (commented):
@mzl2233 thanks for tightening the artifact lint around reserved project storage paths. I found one merge-safe gap in the new matcher: valid unquoted `src`/`href` attributes can still point at those reserved paths and bypass the new P0. Details inline.


apps/daemon/src/lint-artifact.ts
```diff
@@ -546,6 +557,35 @@ function clip(s: string): string {
   return trimmed.length > 200 ? trimmed.slice(0, 197) + '…' : trimmed;
 }
 
+const RESERVED_PROJECT_PATH_SEGMENTS = ['.live-artifacts', '.od', '.tmp'];
+
+function detectReservedProjectPathReference(html: string): { segment: string; snippet: string } | undefined {
+  const attrRe = /\b(?:src|href)\s*=\s*([\"'])(.*?)\1/gi;
```
Siri-Ray, line 563: This matcher only accepts quoted `src`/`href` values, but HTML permits unquoted attribute values and browsers will still load examples like `<iframe src=.live-artifacts/demo/index.html>` or `<a href=.tmp/render.html>`. That means a syntactically valid artifact can still reference the same reserved project storage paths and avoid the new `reserved-project-path` P0; the added tests cover only quoted values. Please handle the unquoted form too, for example by matching `src|href` followed by either a quoted capture or an unquoted token up to whitespace/`>`, then passing whichever capture is present into `referencesReservedProjectPath`, and add a fixture for the unquoted iframe case.

lefarcen (commented):
Hey @mzl2233, the issue framing is clear and the Surface area section is already scoped to Daemon/API + Tests. One body-only cleanup before merge review wraps: since this is labeled as a bugfix, could you update Validation with the final regression verification once the targeted lint-artifact test or CI run is available? The current note says the targeted test could not run, so reviewers do not yet have the red-to-green check for the `.live-artifacts` failure case. 🙌

lefarcen:
Hey @mzl2233 👋 — 24h follow-up since two review threads are still open:

**1. Regex gap (Siri-Ray's inline comment on `lint-artifact.ts`)**
The current matcher at line 563 requires quoted attribute values — `([\"'])(.*?)\1` — so an unquoted form like `<iframe src=.live-artifacts/demo/index.html>` or `<a href=.tmp/render.html>` isn't caught by the new P0. HTML spec allows unquoted attributes and browsers render them fine. Fixing this (e.g. a two-branch alternate that captures either a quoted group or an unquoted token up to whitespace/`>`) would close that bypass, and a fixture for the unquoted iframe case would keep the regression surface covered.

**2. Validation update**
The PR body still notes that the targeted lint-artifact test couldn't run. Once you get a clean pass (e.g. `pnpm --filter @open-design/daemon test -- apps/daemon/tests/lint-artifact.test.ts`), adding the result to the Validation section gives reviewers the red→green confirmation for the `.live-artifacts` failure case.

Happy to clarify anything — let us know when you get a chance!

PR #1655, opened 2026-05-14
fix: Correct user tab count in Skills page by defaulting undefined source to built-in

xxiaoxiong:
Fixes #1374

## Why

**Use case:** Users open Settings → Skills to browse and manage their skill catalog.

**Pain being addressed:** The `user` tab shows a count of `0` even when user skills exist. This makes the source filter look broken or misleading, and users cannot tell whether user skills are supported, missing, or simply miscounted.

**Root cause:** Skills with an `undefined` `source` field were not being counted in either the `user` or `built-in` tabs. According to the contract in `packages/contracts/src/api/registry.ts`, the `source` field is optional (`source?: SkillSource`), and the comment states:

> Origin of the skill: 'built-in' lives under the repo's `skills/` directory and cannot be deleted from the UI; 'user' lives under `<runtimeData>/user-skills/` and is fully owned by the user.

When `source` is `undefined`, it should default to `'built-in'` since that's the expected behavior for skills without explicit source attribution.

## What users will see

**Before this fix:**
- `user 0` (always shows 0, even if user skills exist)
- `built-in 106` (only counts skills with explicit `source: 'built-in'`)
- Skills with `undefined` source are not counted in any tab
- Clicking `user` or `built-in` tabs may show incorrect results

**After this fix:**
- `user X` (shows correct count of skills with `source: 'user'`)
- `built-in Y` (includes skills with `source: 'built-in'` OR `undefined`)
- All skills are counted in exactly one tab
- Clicking tabs filters correctly

## Surface area

- [x] **UI** — Skills page source filter tabs
- [ ] **Keyboard shortcut** — no changes
- [ ] **CLI / env var** — no changes
- [ ] **API / contract** — no changes (respects existing optional field)
- [ ] **Extension point** — no changes
- [ ] **i18n keys** — no changes
- [ ] **New top-level dependency** — no changes
- [ ] **Default behavior change** — clarifies undefined source as built-in
- [ ] **None** — not applicable

## Screenshots

N/A — This is a counting/filtering fix. The tabs will now show accurate counts.

## Bug fix verification

**Test reproduction:**
1. Open Settings → Skills
2. **Before fix:** `user` tab shows `0`
3. **After fix:** `user` tab shows correct count (likely still `0` if no user skills exist, but will count correctly when they do)
4. **Before fix:** Skills with `undefined` source are not counted
5. **After fix:** Skills with `undefined` source are counted as `built-in`

**Why no automated test:**
- This is a UI counting fix that depends on the actual skill catalog
[...]

PR #1677, opened 2026-05-14
fix: Make Skills category filter counts respect active mode and source filters

xxiaoxiong:
Fixes #1376

## Why

**Use case:** Users browse the Skills catalog using multiple filters (mode, source, category) to find relevant skills.

**Pain being addressed:** Category filter counts show global totals instead of context-aware counts. When users select "image" mode and then see "Documents 5", they expect 5 image-related document skills. Instead, they get zero results because the 5 documents are in other modes.

**Root cause:** The `categoryOptions` computation counted all skills globally without applying the current `modeFilter` and `sourceFilter` first.

## What users will see

**Before this fix:**
- Select "image" mode (25 skills)
- Category filter shows "Documents 5"
- Click "Documents" → empty result list
- UI says "没有匹配的项目。" (No matching items)
- Filter counts promise items that don't exist in the current context

**After this fix:**
- Select "image" mode (25 skills)
- Category filter shows only categories that exist in image mode
- Counts reflect actual available results
- No more misleading empty result lists
- Filter counts are accurate promises

## Surface area

- [x] **UI** — category filter counts in Skills page
- [ ] **i18n keys** — no changes
- [ ] **Default behavior change** — counts now context-aware

## Screenshots

N/A — This fixes the count calculation logic. The UI appearance is unchanged, but the counts now match the filtered results.

## Bug fix verification

**Test reproduction:**
1. Open Settings → Skills page
2. Select "image" mode filter
3. Observe category filter counts
4. **Before fix:** Shows "Documents 5" (global count)
5. Click "Documents" → empty result list
6. **After fix:** Only shows categories that exist in image mode with accurate counts

**Why no automated test:**
- This fixes count calculation logic
- The filtering logic is already tested
- Manual verification confirms accurate counts

## Validation

- [x] Code review: applies modeFilter and sourceFilter before counting
- [x] Code review: adds filters to useMemo dependencies
- [x] Code review: uses same source defaulting as PR #1655
- [x] Consistency: matches the filtering logic in filteredSkills

**Commands run:**
- Code inspection of count calculation
- Verified filter application order
- Confirmed dependency array

#### `apps/web/src/components/SkillsSection.tsx`

**Before (line 135-143):**
```typescript
const categoryOptions = useMemo(() => {
  const counts = new Map<string, number>();
  for (const s of skills) {
    const cat = s.category;
    if (typeof cat !== 'string' || !cat) continue;
    counts.set(cat, (counts.get(cat) ?? 0) + 1);
  }
  return Array.from(counts.entries()).sort((a, b) => a[0].localeCompare(b[0]));
}, [skills]);
```

**After (line 135-147):**
```typescript
const categoryOptions = useMemo(() => {
  const counts = new Map<string, number>();
  for (const s of skills) {
    // Apply current mode and source filters before counting categories
    if (modeFilter !== 'all' && s.mode !== modeFilter) continue;
    if (sourceFilter !== 'all' && (s.source ?? 'built-in') !== sourceFilter) continue;
    const cat = s.category;
    if (typeof cat !== 'string' || !cat) continue;
    counts.set(cat, (counts.get(cat) ?? 0) + 1);
  }
  return Array.from(counts.entries()).sort((a, b) => a[0].localeCompare(b[0]));
}, [skills, modeFilter, sourceFilter]);
```

### Why this approach

1. **Context-aware counts**: Category counts now reflect the current filter state
2. **Consistent logic**: Uses the same filter logic as `filteredSkills`
3. **Source defaulting**: Uses `s.source ?? 'built-in'` (same as PR #1655)
4. **Reactive updates**: Added `modeFilter` and `sourceFilter` to dependencies
5. **No breaking changes**: Only fixes the count calculation

### Filter application order

The fix applies filters in the same order as `filteredSkills`:
1. **Mode filter** (image, text, audio, etc.)
2. **Source filter** (user, built-in)
3. **Category filter** (Documents, Media, etc.)
4. **Search query**

Category counts are computed after mode and source filters but before category and search filters, so they show "how many skills in each category are available given the current mode and source selection."

## Impact

- **Surface area**: Skills page category filter counts
- **User experience**: Accurate filter counts, no more misleading empty results
- **Accessibility**: No changes (count calculation only)
- **Files changed**: 1 (SkillsSection.tsx)
- **Lines changed**: +5 -1
- **Breaking changes**: None (fixes existing behavior)

## Related

- #1376 — original issue tracking the count mismatch
- #1655 — PR that fixed user tab count (same source defaulting pattern)
- Line 148-150 — filteredSkills uses the same filter logic

### DIFF
--- a/apps/web/src/components/SkillsSection.tsx
+++ b/apps/web/src/components/SkillsSection.tsx
@@ -146,6 +146,14 @@ export function SkillsSection({ cfg, setCfg }: Props) {
     return Array.from(counts.entries()).sort((a, b) => a[0].localeCompare(b[0]));
   }, [skills, modeFilter, sourceFilter]);
 
+  // Reset category filter to 'all' when the selected category is no longer
+  // available for the current mode/source filters
+  useEffect(() => {
+    if (categoryFilter === 'all') return;
+    const stillValid = categoryOptions.some(([cat]) => cat === categoryFilter);
+    if (!stillValid) setCategoryFilter('all');
+  }, [categoryFilter, categoryOptions]);
+
   const filteredSkills = useMemo(() => {
     const q = search.toLowerCase().trim();
     return skills.filter((s) => {
--- a/apps/web/src/components/SkillsSection.tsx
+++ b/apps/web/src/components/SkillsSection.tsx
@@ -132,15 +132,19 @@ export function SkillsSection({ cfg, setCfg }: Props) {
   // SKILL.md frontmatter). The pill row only renders when at least one
   // skill in the listing carries one, so a project that ships only the
   // baseline functional skills doesn't see an empty filter row.
+  // Counts are computed based on the current mode and source filters.
   const categoryOptions = useMemo(() => {
     const counts = new Map<string, number>();
     for (const s of skills) {
+      // Apply current mode and source filters before counting categories
+      if (modeFilter !== 'all' && s.mode !== modeFilter) continue;
+      if (sourceFilter !== 'all' && (s.source ?? 'built-in') !== sourceFilter) continue;
       const cat = s.category;
       if (typeof cat !== 'string' || !cat) continue;
       counts.set(cat, (counts.get(cat) ?? 0) + 1);
     }
     return Array.from(counts.entries()).sort((a, b) => a[0].localeCompare(b[0]));
-  }, [skills]);
+  }, [skills, modeFilter, sourceFilter]);
 
   const filteredSkills = useMemo(() => {
     const q = search.toLowerCase().trim();


apps/web/src/components/SkillsSection.tsx
```diff
@@ -132,15 +132,19 @@ export function SkillsSection({ cfg, setCfg }: Props) {
   // SKILL.md frontmatter). The pill row only renders when at least one
   // skill in the listing carries one, so a project that ships only the
   // baseline functional skills doesn't see an empty filter row.
+  // Counts are computed based on the current mode and source filters.
   const categoryOptions = useMemo(() => {
     const counts = new Map<string, number>();
     for (const s of skills) {
+      // Apply current mode and source filters before counting categories
+      if (modeFilter !== 'all' && s.mode !== modeFilter) continue;
+      if (sourceFilter !== 'all' && (s.source ?? 'built-in') !== sourceFilter) continue;
```
nettee, line 141: When `categoryOptions` starts filtering by `modeFilter` and `sourceFilter`, the selected `categoryFilter` can become stale. Repro from this branch: select a category, then switch mode or source so that category no longer appears in `categoryOptions`. `filteredSkills` still applies the old `categoryFilter` at lines 154-155, but the pill row at lines 405-426 no longer renders that category (or disappears entirely when `categoryOptions.length === 0`), so the list shows no matches without exposing the active filter to clear.

That regresses the flow this PR is trying to fix: users can still land on an empty list from the filter chips, just with the active category hidden instead of miscounted. Please either reset `categoryFilter` back to `all` whenever the recomputed `categoryOptions` no longer contains the selected category, or keep rendering the selected category pill with a zero count / disabled state so the user can see and clear it. A regression test should cover "select category → change mode/source so that category disappears".

nettee (changes requested):
This fixes the displayed category counts, but there is still one blocker in the new mode/source-aware counting path: changing mode or source can now hide an active category filter while the stale category filter keeps the list empty.


lefarcen:
Hey @xxiaoxiong! 👋

@qiongyu1999's point makes sense — here's the concrete implementation pattern:

Add a `useEffect` that resets `categoryFilter` to `'all'` when the current value is no longer in the available categories:

```tsx
useEffect(() => {
  if (categoryFilter === 'all') return;
  const stillValid = categoryOptions.some(([cat]) => cat === categoryFilter);
  if (!stillValid) setCategoryFilter('all');
}, [categoryFilter, categoryOptions]);
```

For the regression test, you could check:
1. Start with a category selected (e.g., 'filesystem')
2. Change mode/source such that 'filesystem' is no longer available
3. Verify `categoryFilter` resets to 'all' and the result list shows all (filtered by mode/source)

Let me know if this makes sense!

xxiaoxiong:
Thanks for the feedback! I've added the requested feature:

✅ Added `useEffect` that resets `categoryFilter` to `'all'` when the selected category is no longer available in the current mode/source filters

This prevents the UX issue where users could have an empty result list with no visible way to clear the filter.

The implementation follows the pattern you suggested:
```tsx
useEffect(() => {
  if (categoryFilter === 'all') return;
  const stillValid = categoryOptions.some(([cat]) => cat === categoryFilter);
  if (!stillValid) setCategoryFilter('all');
}, [categoryFilter, categoryOptions]);
```

apps/web/src/components/SkillsSection.tsx
```diff
@@ -132,15 +132,27 @@ export function SkillsSection({ cfg, setCfg }: Props) {
...
-  }, [skills]);
+  }, [skills, modeFilter, sourceFilter]);
+
+  // Reset category filter to 'all' when the selected category is no longer
+  // available for the current mode/source filters
+  useEffect(() => {
+    if (categoryFilter === 'all') return;
+    const stillValid = categoryOptions.some(([cat]) => cat === categoryFilter);
+    if (!stillValid) setCategoryFilter('all');
+  }, [categoryFilter, categoryOptions]);
```
nettee, line 155: `categoryOptions` and this reset effect are the entire bugfix path for #1376 now, but the PR still leaves them untested. `apps/web/tests/components/SkillsSection.test.tsx` only exercises delete actions today (lines 77-110), so there is no automated coverage proving that mode/source changes recompute the category pills correctly or that an invalid selected category resets back to `all` instead of leaving the list empty again. That matters here because this interaction already needed a follow-up commit once, and another dependency/order regression in these changed lines would silently bring back the same empty-results UX. Please add a component regression test in `apps/web/tests/components/SkillsSection.test.tsx` that (1) switches mode/source and asserts the category counts update for the visible context, and (2) selects a category, changes mode/source so that category disappears, and verifies the filter resets to `all` and the list repopulates.

nettee (changes requested):
This fixes the stale hidden-category state from the previous round, but one blocking gap remains: the new mode/source-aware category behavior still ships without regression coverage.

PR #1695, opened 2026-05-14
Limit DOM fallback comment targets

mzl2233:
## Summary
Tightens the comment picker DOM fallback so previews without explicit anchors only target interactive, semantic, media, heading, list/table, ARIA-labelled, or text-leaf elements instead of any visible wrapper.

## Why
This came up while checking comment picker behavior on generated previews that do not include `data-od-id` or `data-screen-label` anchors. In that fallback path, the picker could attach comments to layout wrappers after viewport reflow, which makes anchors and labels confusing instead of pointing at the interface element the user clicked.

## What users will see
When using the comment picker on unannotated generated previews, clicks are less likely to select broad layout containers such as shells, grids, and cards. The picker will prefer visible UI/content elements, while empty or structural surfaces can still fall through to the existing free-pin behavior.

## Surface area
- [x] Web app / UI runtime
- [ ] Daemon / API
- [ ] Desktop / packaged runtime
- [ ] CLI flags or environment variables
- [ ] Skills / design systems / design templates / craft
- [ ] Documentation

## Validation
- `git diff --check`
- `pnpm --filter @open-design/web typecheck` could not run locally because this checkout has no `node_modules` installed; pnpm reported `tsc: not found`.

Closes #1692

### DIFF
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -724,10 +724,22 @@ function injectSelectionBridge(
     }
     return true;
   }
+  function meaningfulDomFallbackTarget(el){
+    if (!visibleTarget(el)) return false;
+    var tag = el.tagName ? el.tagName.toLowerCase() : '';
+    if (/^(a|button|input|textarea|select|label|img|video|canvas|svg|h1|h2|h3|h4|h5|h6|p|li|td|th)$/.test(tag)) return true;
+    if (el.getAttribute && (el.getAttribute('role') || el.getAttribute('aria-label') || el.getAttribute('title'))) return true;
+    var text = (el.textContent || '').replace(/\s+/g, ' ').trim();
+    if (!text) return false;
+    for (var child = el.firstElementChild; child; child = child.nextElementSibling) {
+      if ((child.textContent || '').replace(/\s+/g, ' ').trim()) return false;
+    }
+    return true;
+  }
   function targetFrom(el, allowDomFallback){
     var id = el.getAttribute('data-od-id') || el.getAttribute('data-screen-label');
     var selector = annotatedSelectorFor(el);
-    if (!id && allowDomFallback && visibleTarget(el)) {
+    if (!id && allowDomFallback && meaningfulDomFallbackTarget(el)) {
       selector = domSelectorFor(el);
       if (selector) id = 'dom:' + selector;
     }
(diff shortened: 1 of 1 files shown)


apps/web/src/runtime/srcdoc.ts
```diff
@@ -724,10 +724,22 @@ function injectSelectionBridge(
...
+  function meaningfulDomFallbackTarget(el){
+    if (!visibleTarget(el)) return false;
+    var tag = el.tagName ? el.tagName.toLowerCase() : '';
+    if (/^(a|button|input|textarea|select|label|img|video|canvas|svg|h1|h2|h3|h4|h5|h6|p|li|td|th)$/.test(tag)) return true;
+    if (el.getAttribute && (el.getAttribute('role') || el.getAttribute('aria-label') || el.getAttribute('title'))) return true;
+    var text = (el.textContent || '').replace(/\s+/g, ' ').trim();
+    if (!text) return false;
+    for (var child = el.firstElementChild; child; child = child.nextElementSibling) {
+      if ((child.textContent || '').replace(/\s+/g, ' ').trim()) return false;
+    }
```
nettee, line 736: This new child-text check makes any semantic container with nested text ineligible for DOM fallback, so a click on `<section id="card"><h2>Card</h2></section>` now falls through to free-pin instead of selecting the section itself. That breaks the existing contract already covered in `apps/web/tests/runtime/srcdoc-bridge-empty-targets.test.ts:239`, which expects comment-mode fallback targets to include that section for Pods hit-testing on unannotated pages. Please keep visible semantic containers targetable here (for example by preferring better descendants during ancestor selection, rather than rejecting every element whose child subtree has text) so container-level clicks and the existing test behavior still work.

nettee (changes requested):
Found one regression in the DOM fallback filtering; details inline.


lefarcen (commented):
Hey @mzl2233! 👋 The problem statement here is clear and the PR body is well-structured — tightening the fallback filter to semantic/interactive targets rather than any visible node is the right direction.

A couple of things need attention before this can land:

**Blocking regression (nettee's inline comment)** — the new child-text check at line 736 makes any semantic container with nested text ineligible for DOM fallback, so a click on `<section><h2>Card</h2></section>` now falls through to free-pin instead of selecting the section. That breaks the existing coverage at `apps/web/tests/runtime/srcdoc-bridge-empty-targets.test.ts:239`. nettee's inline comment suggests preferring better descendants during ancestor selection rather than rejecting every element whose subtree has text — that's the direction to explore so container-level clicks still work.

**CI — "Validate workspace" is failing** — [details here](https://github.com/nexu-io/open-design/actions/runs/25853004552/job/75987771382); likely connected to the logic change.

**Related open PR** — @Hetsavani also has PR #1706 open touching the same file with a different approach to the same fallback issue — both close #1692. The maintainer team will decide which lands; just flagging it so the efforts don't go uncoordinated.

Once the regression is addressed and CI is green, happy to move this forward! ❤️

lefarcen:
Hey @mzl2233 — quick lifecycle update: PR #1706 has now merged and closed #1692 with the same DOM fallback comment-picker fix, so this PR is superseded.

I’m closing this one to keep the queue tidy, especially since the current branch is also conflicted and still has the earlier blocking review/CI failure. Really appreciate you jumping on this and writing up the user-facing behavior so clearly — it helped validate the direction. ❤️

PR #1718, opened 2026-05-14
feat: add POST /api/projects/:id/handoff endpoint

bankielewicz:
## Why

I filed nexu-io/open-design#462 myself on 2026-05-04 and authored the
prerequisite stack (transcript export #493 + finalize endpoint #832 +
"Continue in CLI" handoff button #974). On 2026-05-12 @lefarcen
directly invited me to implement #462 (issue comment 4435528822). This
PR ships the de-risked sub-slice he named verbatim: *"start with the
explicit user-triggered slice: add the POST /api/projects/:id/handoff
synthesis endpoint, then wire a separate 'Resume conversation in new
chat' action."* The web button + wiring will follow as a separate PR
once Q1 from my /claim comment resolves.

The pain: in long multi-turn design sessions, useful context
accumulates beyond what fits cleanly in the next conversation, and the
manual workaround today is the user typing out a handoff prompt by
hand. This is the server-side primitive that lets a future button
close that gap.

Refs nexu-io/open-design#462 (partial — endpoint only).

## What users will see

Nothing visible in this PR alone — the "Resume conversation in new
chat" button + composer-seed wiring is the follow-up PR. This one adds
the daemon endpoint the button will call.

API-wise, callers can now `POST /api/projects/:id/handoff` with
`{ apiKey, model, baseUrl?, maxTokens? }` (BYOK, mirrors
`/finalize/anthropic`) and get back
`{ prompt, model, inputTokens, outputTokens, transcriptMessageCount }`.
The `prompt` is a synthesized Markdown body the next conversation
sends as its first user message (Context / Decisions made / Open
questions / Current focus / Provenance — system prompt verbatim in
`apps/daemon/src/handoff-design.ts`).

## Surface area

- [ ] **UI** — none in this PR
- [x] **API / contract** — new `POST /api/projects/:id/handoff` (daemon)
      and new `@open-design/contracts/api/handoff` subpath (with
      `HandoffRequest` / `HandoffResponse` / `HANDOFF_SCHEMA_VERSION`)
- [ ] **Extension point** — none
- [ ] **i18n keys** — none (deferred to the web-button follow-up PR)
- [ ] **New top-level dependency** — none
- [ ] **Default behavior change** — none (new opt-in endpoint, no
      callers ship in this PR)
- [ ] **None**

### Architecture notes

Mirrors `finalize-design.ts` deliberately so reviewer cognitive load
matches the prior PR (#832):
- Reuses `exportProjectTranscript` (#493 primitive),
  `truncateTranscriptForPrompt`, `callAnthropicWithRetry`,
  `extractDesignMd`, `FinalizeUpstreamError`, and the newly-exported
  `DEFAULT_TIMEOUT_MS` (one-line `export` added at
  `apps/daemon/src/finalize-design.ts:61`; called out so reviewers
  don't ask).
- **No lockfile** (handoff is read-only; concurrent calls are safe →
  no `FinalizePackageLockedError` / 409 CONFLICT branch).
- **No disk write** (the caller seeds the composer; persistence is
  the next conversation's responsibility).
- **No design-system / artifact loaders.** This implements Q2's
  "transcript-only" default per the issue body's *"Implementation
  hints: ... LLM pass against transcript"*. If you'd prefer the full
  finalize mirror (design system + current artifact), it's an
  additive change to `synthesizeHandoffPrompt` — happy to follow up
  on review.

The system prompt's Provenance section asks for plain
`- Field: value` bullets with no Markdown emphasis on labels — this
is a deliberate pre-emption of the round-2 strip narrowing fixed in
PR #1584's parser.

## Bug fix verification

Not a bug fix — feature work.

## Validation

- `pnpm guard` ✓
- `pnpm typecheck` ✓ (root; covers contracts + daemon + web + desktop
  + tools + packaged)
- `pnpm --filter @open-design/contracts test` — 6 files, 68 tests
  green (includes 4 new handoff contract tests + `package-runtime`
  validating the new `./api/handoff` subpath emits a real `.mjs` per
  the NodeNext-published-submodule gotcha)
- `pnpm --filter @open-design/contracts build` — `dist/api/handoff.{mjs,d.ts}` emitted
- `pnpm --filter @open-design/daemon test` — 136 files, 2008 tests
  green (4 todo, 0 fail). Handoff coverage:
  - `handoff-design.test.ts` (10) — `buildHandoffPrompt` shape +
    `synthesizeHandoffPrompt` pipeline (happy path, default
    baseUrl, network-error → 502, non-JSON → 502, AbortSignal,
    internal `DEFAULT_TIMEOUT_MS` timeout, 384 KiB truncation).
  - `handoff-route.test.ts` (12) — HTTP-layer validation (400 missing
    apiKey/model, 400 invalid baseUrl, 403 private-IP baseUrl, 400
    unsafe id, 404 unknown project), 200 happy path, upstream-error
    mapping (401/429/502 generic / 502 non-JSON), api-key redaction.
  - `handoff-route-abort.test.ts` (1) — client-disconnect aborts the
    daemon-side `synthesizeHandoffPrompt` controller via
    `res.on('close')`.

**TDD disclosure (one spot).** Steps 1-4 followed red → green
strictly (test fails on missing module / 404 endpoint, then source
turns it green). Step 5 (`handoff-route-abort.test.ts`) was written
*after* the abort wiring was already in `handoff-routes.ts`, so the
red phase wasn't observed for that specific file. The assertion is
load-bearing (`capture.signal.aborted` only flips through the route's
`res.on('close')` wiring), but I want to flag the gap rather than
imply uniformly clean TDD.

### Follow-up PR

Web wiring ("Resume conversation in new chat" button next to "New
conversation", `synthesizeHandoff` state helper, composer-seed) is
queued and depends on Q1 resolution from the /claim comment thread on
#462.


lefarcen:
Hey @bankielewicz! 👋

The prerequisite write-up here is exceptional — threading the lineage from #493 → #832 → #974 through to this endpoint, calling out the intentional lockfile omission, explaining the Provenance bullet format against the #1584 round-2 narrowing, and flagging the one TDD gap honestly rather than glossing over it. That level of reviewer briefing makes the pool's job a lot easier.

Routing:
- **Code review** → @nettee assigned (pool rotation)
- **Product review** → @qiongyu1999 needs to sign off before merge (new API/contract surface)

The issue pointer on #462 is updated with what this PR delivers vs. what the follow-up PR handles.

apps/daemon/src/handoff-design.ts
```diff
@@ -0,0 +1,221 @@
...
+  const project = getProject(db, projectId);
+  if (!project) {
+    throw new Error(`project not found: ${projectId}`);
+  }
+
+  const now = options.now ?? (() => new Date());
+  const baseUrl = options.baseUrl ?? DEFAULT_BASE_URL;
+  const maxTokens = options.maxTokens ?? DEFAULT_MAX_TOKENS;
+
+  const transcriptResult = exportProjectTranscript(db, projectsRoot, projectId, { now });
```
nettee: `exportProjectTranscript()` still acquires `.transcript.lock`, so this path is not actually concurrency-safe yet. A second handoff request (or a handoff that overlaps `/api/projects/:id/finalize/anthropic`) will throw `TranscriptExportLockedError` before the Anthropic call, and `handoff-routes.ts` currently has no special handling for that exception, so it falls through to the generic 500/`INTERNAL_ERROR` branch. That contradicts the new module/route comments and turns a realistic double-submit race into an internal error for callers. Please either avoid the shared transcript-export lock in the handoff flow, or catch `TranscriptExportLockedError` here and return an intentional retryable response (for example `409 CONFLICT`) with a regression test that covers overlapping handoff/transcript-export requests.

nettee (changes requested):
Found one blocking issue in the new handoff path: concurrent handoff/finalize requests can still trip the shared transcript-export lock and surface as a 500, so the endpoint is not actually safe to double-submit or run alongside other transcript consumers.


apps/daemon/src/handoff-design.ts
bankielewicz: Verified the lock contention against `transcript-export.ts:49-63,131-163` — confirmed. Going with option B from your two paths: catch `TranscriptExportLockedError` in `handoff-routes.ts` → `409 CONFLICT` (mirrors `FinalizePackageLockedError → 409` at `import-export-routes.ts:603-605`), plus a regression test in `handoff-route.test.ts` that pre-acquires `.transcript.lock` on disk before firing the request. Will also correct the misleading "no lockfile (read-only)" comment block in `handoff-design.ts:8-12` — handoff doesn't add its own lock but transitively acquires `.transcript.lock` via the transcript-export call.

Adjacent finding: finalize has the same gap (`finalize-design.ts:267` calls `exportProjectTranscript`; `import-export-routes.ts` only catches `FinalizePackageLockedError`, not `TranscriptExportLockedError`). Leaving that out of this PR; happy to file a follow-up issue if you want it tracked.

bankielewicz:
Heads-up for reviewers: I merged current `main` into the branch to clear a merge conflict — the PR was `CONFLICTING` after `main` advanced ~250 commits.

- **Not a rebase, no force-push.** The head moved by exactly one merge commit (`a2917796`) plus a one-line conflict resolution. The four original commits (`c49a119b`, `380d6475`, `ffdc0d21`, `cb74f347`) are untouched ancestors.
- **Sole conflict:** `packages/contracts/src/index.ts`. `main` migrated the contracts barrel to NodeNext `.js`-extension exports; this branch had added `export * from './api/handoff'`. Resolved by taking main's version and re-applying the handoff export as `export * from './api/handoff.js'`.
- **Verified post-merge:** `pnpm guard`, `pnpm typecheck`, and the contracts + daemon test/build suites pass. All three handoff suites green. (Two `project-watchers.test.ts` failures are pre-existing WSL2 chokidar flakiness — that module is byte-identical to `origin/main` and untouched by this branch.)

PR is back to `MERGEABLE`. Ready for the re-review against `a2917796`.

apps/daemon/src/handoff-design.ts
```diff
@@ -0,0 +1,225 @@
...
+      const errName =
+        err && typeof err === 'object' && 'name' in err
+          ? (err as { name?: unknown }).name
+          : '';
+      if (errName === 'AbortError') throw err;
+      const message = err instanceof Error ? err.message : String(err);
+      throw new FinalizeUpstreamError(502, '', `upstream network error: ${message}`);
+    }
+  } finally {
+    clearTimeout(timeoutId);
```
nettee, line 267: `clearTimeout(timeoutId)` runs before `await response.json()`, but `fetch()` resolves as soon as the upstream sends headers. That means a response that sends headers and then stalls its body will bypass the 120s guard completely: `callAnthropicWithRetry()` returns, the timeout is cleared here, and the subsequent body read can hang forever. On this route that turns a slow/broken upstream into an unbounded daemon request instead of the intended timeout failure. Please keep the timeout signal alive until the body has been fully consumed/parsing has finished (for example by moving the `clearTimeout(timeoutId)` until after `response.json()` succeeds or fails), and add a regression test that returns a `Response` whose body never closes after headers so this path cannot regress.

nettee (changes requested):
Found one blocking issue in the new handoff synthesis path: the route-level timeout stops as soon as Anthropic sends response headers, so a half-open body can leave `/api/projects/:id/handoff` hanging indefinitely.


apps/daemon/src/handoff-design.ts
bankielewicz: Fixed in `d0ae1211`.

`clearTimeout(timeoutId)` now lives in a single outer `finally` spanning both the upstream call and `response.json()`, so the timeout stays armed through the body read — a response that stalls its body after headers now aborts instead of hanging. Regression test added in `handoff-design.test.ts`: a `fetchImpl` whose `Response` body never enqueues or closes after headers, raced against a 500ms deadline, asserts the call aborts (not hangs) and rejects with `AbortError`.

One change beyond the literal ask, flagging it explicitly: the body-parse `catch` now re-throws `AbortError` as-is, mirroring the call-phase `catch`. Without that, a body-phase timeout would surface as `502` "non-JSON body" — technically a failure, but not the timeout failure. Re-throwing lets the route map it to the intended `503` "handoff timed out" (`handoff-routes.ts:122-124`), consistent with how a headers-phase timeout already behaves. Happy to drop that half if you'd prefer the strictly minimal change.

apps/daemon/src/handoff-routes.ts
```diff
@@ -0,0 +1,131 @@
...
+      if (err instanceof FinalizeUpstreamError) {
+        const safeDetails = redactSecrets(err.rawText || '', [apiKey]);
+        const init = safeDetails ? { details: safeDetails } : {};
+        if (err.status === 401) {
+          return sendApiError(res, 401, 'UNAUTHORIZED', err.message, init);
+        }
+        if (err.status === 429) {
+          return sendApiError(res, 429, 'RATE_LIMITED', err.message, init);
+        }
+        return sendApiError(res, 502, 'UPSTREAM_UNAVAILABLE', err.message, init);
```
nettee, line 162: `FinalizeUpstreamError` can still represent caller-side request errors here. `callAnthropicWithRetry()` preserves any non-retryable upstream status, so if Anthropic rejects an otherwise well-typed request body (for example an unknown `model` or invalid `max_tokens`) with HTTP 400, this branch remaps that deterministic input error to `502 UPSTREAM_UNAVAILABLE`. That makes the endpoint report bad caller input as a transient server outage, which can trigger pointless retries and hides which field needs fixing. Please special-case upstream 400s here and add a route test in `handoff-route.test.ts` that drives an Anthropic `400 invalid_request_error`, so request-shape failures come back as `400 BAD_REQUEST` with redacted details instead of 502.

packages/contracts/tests/handoff-contract.test.ts
```diff
@@ -0,0 +1,67 @@
+import { describe, expect, it } from 'vitest';
+
+import {
+  HANDOFF_SCHEMA_VERSION,
+  type HandoffRequest,
+  type HandoffResponse,
+} from '../src/api/handoff';
+
+describe('handoff contract', () => {
+  it('exports a runtime schema-version marker so esbuild emits a .mjs (NodeNext requires it)', () => {
```
nettee, line 11: This test only imports `../src/api/handoff`, so it never exercises the publish-time surfaces this PR actually changes: the esbuild entrypoint, the package `exports` map, and the root `index.ts` re-export. The new `HANDOFF_SCHEMA_VERSION` exists specifically so NodeNext consumers can resolve a real emitted `.mjs`, but this assertion would still pass if `@open-design/contracts/api/handoff` were missing from the built package. Please extend `package-runtime.test.ts` (or equivalent runtime/export coverage) to import the new subpath export directly, and optionally the root package export too, so the published contract surface added here cannot regress unnoticed.

nettee (changes requested):
Found one blocking issue in the new handoff route, plus one follow-up test-coverage gap.


apps/daemon/src/handoff-routes.ts
bankielewicz: Fixed in `6b976441`.

`handoff-routes.ts` now special-cases `err.status === 400` → `400 BAD_REQUEST`, ahead of the generic `502` fallback, reusing the shared `redactSecrets(init)` path the `401`/`429` branches already use — so the upstream `invalid_request_error` detail is surfaced (caller learns which field is wrong) while the inbound key stays redacted. Route test added in `handoff-route.test.ts`: drives an Anthropic `400 invalid_request_error` whose body echoes the api key, asserts `400 BAD_REQUEST`, the detail is surfaced, and the key is redacted out of the response.

Scope note: the sibling `/finalize` route at `import-export-routes.ts:612-622` has the identical `400 → 502` mapping. Same latent issue — I've kept it out of this PR as a follow-up rather than widening scope, but flagging it so it isn't invisible.

packages/contracts/tests/handoff-contract.test.ts
bankielewicz: Fixed in `6b976441`.

Extended `packages/contracts/tests/package-runtime.test.ts` rather than `handoff-contract.test.ts`, so the assertions run against the *built* package:

- the "exports built JavaScript" test now asserts the `./api/handoff` `exports`-map entry (`./dist/api/handoff.mjs` + `.d.ts`);
- the "makes runtime exports importable" test now `import`s `@open-design/contracts/api/handoff` through the package `exports` map and asserts `HANDOFF_SCHEMA_VERSION === 1`, plus the root re-export via `contracts.HANDOFF_SCHEMA_VERSION` (the optional half — included to close the surface fully).

That exercises the esbuild entrypoint, the `exports` map, and the root `index.ts` re-export — the publish surfaces a source-only `import '../src/api/handoff'` would miss. `handoff-contract.test.ts` stays as the source-level DTO-shape coverage.

nettee (dismissed):
@bankielewicz Re-checked the current head's handoff route, synthesis pipeline, and contracts export wiring against the changed-range feedback. The timeout/body-read fix, transcript-lock 409 mapping, upstream 400 handling, and package export coverage now line up cleanly on this head. Nice job tightening this up and carrying the regressions forward.


bankielewicz:
Thanks @qiongyu1999 — evaluating the full resume-conversation flow before merge makes sense for a product call, happy to do it that way. I'll put the web entry point up as a companion PR so #1718 and the UI can be reviewed together (keeping them as separate PRs per @lefarcen's earlier framing — just reviewed as a pair rather than sequentially).

One product decision I need to build the UI to the flow you'll be evaluating. When the user clicks "Resume conversation in new chat", the daemon returns the synthesized handoff prompt; the new conversation can either:

- **(a) auto-send** it as the first message — one click and the resumed chat is already moving; or
- **(b) seed the composer** with it unsent — the user can review/trim the synthesized prompt before spending tokens on it.

The branch's current plan defaults to **(a) auto-send**. (b) trades a click for a review checkpoint. Since you'll be evaluating this exact flow, I'd rather build it to your call than guess — which would you prefer? Once you confirm, I'll get the companion UI PR up for the joint review.

bankielewicz:
@lefarcen — quick re-review request. `main` advanced ~71 commits and #1718 went `CONFLICTING`; I've merged latest main and resolved it — head is now `e45fbb0f` and the PR is back to `MERGEABLE`.

The only conflict was in `finalize-design.ts`: a one-line union — the branch's `export` on `DEFAULT_TIMEOUT_MS` plus main's new `FINALIZE_PROVIDER_PROTOCOLS` / `isFinalizeProviderProtocol` / `defaultBaseUrlForFinalizeProtocol`. No logic change — it's a conflict-resolution merge only (66 handoff + finalize tests, daemon typecheck, and `guard` all green locally).

That push re-staled @nettee's and @qiongyu1999's approvals (`reviewDecision` is now `REVIEW_REQUIRED`). Both had already approved `806db576`; `e45fbb0f` differs only by the merge + that one-line resolution. Could you re-request them on the new head?

lefarcen:
Done — re-requested @nettee and @qiongyu1999 on `e45fbb0f`. Thanks for spelling out the conflict scope clearly.

apps/daemon/src/route-context-contract.ts
```diff
@@ -16,6 +17,7 @@ type AllRegisteredRouteDeps =
   & RegisterDeployRoutesDeps
   & RegisterDeploymentCheckRoutesDeps
   & RegisterFinalizeRoutesDeps
+  & RegisterHandoffRoutesDeps
```
nettee, line 20: AllRegisteredRouteDeps now includes RegisterHandoffRoutesDeps, but the assertServerContextSatisfiesRoutes(...) object in apps/daemon/src/server.ts:9863 still omits the new handoff field. That leaves the compile-time coverage check out of sync with the route registrar set: once dependencies are installed, this object literal is missing a required ServerContext.handoff property and @open-design/daemon stops type-checking. Please thread handoff: handoffDeps into that assertion object, and any other ServerContext literals, so the new route dependency is actually covered.

nettee (changes requested):
Found one blocking issue in the new handoff route wiring: the compile-time route-context coverage check now requires handoff, but the server assertion object was not updated to provide it.


apps/daemon/src/route-context-contract.ts
bankielewicz: Fixed in `951dc598` — added `handoff: handoffDeps` to the `assertServerContextSatisfiesRoutes({...})` call in `server.ts` so it enumerates the handoff deps alongside every other route registrar's.

One correction for the record, though: the daemon was **not** failing type-check. `tsc -p tsconfig.json --noEmit` is clean (before and after this commit), and CI `Validate workspace` + `build` are green on `e45fbb0f` — as they were on `806db576`, which carried this exact same literal without `handoff`. The compile-time route-dep coverage guarantee is the `Assert<ServerContext extends AllRegisteredRouteDeps>` type at `route-context-contract.ts:34`, and `AllRegisteredRouteDeps` already includes `RegisterHandoffRoutesDeps` (line 20) — so handoff coverage was asserted regardless of this call-site literal. The literal simply wasn't listing `handoff`; now it does, which is the right consistency fix.

lefarcen:
API endpoint by itself is invisible to users — what's the end-user-visible flow that calls `/handoff`? A short clip or screenshot of the handoff UX would help me route this for product review.

bankielewicz:
Good point — agreed, the endpoint alone has no user-visible surface, so there's nothing to product-review in isolation.

The intended flow: a **"Resume conversation in new chat"** control next to the existing "New conversation" action in the chat header. On click, the web app calls `POST /api/projects/:id/handoff`; the daemon synthesizes a handoff prompt from the current transcript; the web app then opens a fresh conversation with that prompt as its first user message — so the new chat resumes the work without the user replaying context by hand. #1718 is the daemon half of that.

I'm building the web entry point as the companion UI PR now — that's the screenshot-able artifact. I'll post it with a short clip of the click → resumed-chat flow and link it here, so you and @qiongyu1999 can evaluate #1718 and the UI together (as qiongyu1999 asked). I'll also bring #1718 current with `main` — it's drifted into a conflict again — when that companion PR goes up, rather than re-merging it in isolation now.

nettee:
@bankielewicz I'm holding off on generating review comments for nexu-io/open-design#1718 because this pull request has merge conflicts right now.

apps/daemon/src/handoff-design.ts
```diff
@@ -0,0 +1,241 @@
...
+
+  const transcriptResult = exportProjectTranscript(db, projectsRoot, projectId, { now });
+  const transcriptJsonl = fs.readFileSync(transcriptResult.path, 'utf8');
+  const truncatedJsonl = truncateTranscriptForPrompt(transcriptJsonl);
+
+  const { systemPrompt, userPrompt } = buildHandoffPrompt({
+    projectId,
+    transcriptJsonl: truncatedJsonl,
+    transcriptMessageCount: transcriptResult.messageCount,
+    now: now(),
```
nettee: `exportProjectTranscript()` can return a header-only transcript with `messageCount === 0`, and this new path still proceeds to call Anthropic and synthesize a resume prompt from that empty session. That means `POST /api/projects/:id/handoff` will spend BYOK tokens and can fabricate a handoff for a project that has no conversation history, instead of failing fast on missing input. Please reject the empty-transcript case immediately after export (for example with a typed `EmptyTranscriptError` or equivalent), map it in `handoff-routes.ts` to a 400-style response, and add route/pipeline coverage for an empty project or zero-message conversation.

nettee (changes requested):
Found one blocking issue in the new handoff synthesis path.


apps/daemon/src/handoff-design.ts
bankielewicz: Confirmed — a zero-message export should fail fast, not spend BYOK tokens fabricating a handoff. Proposed:

- After the transcript export in `synthesizeHandoffPrompt`, `messageCount === 0` → throw a typed `EmptyTranscriptError`.
- `handoff-routes.ts` maps it to `400 BAD_REQUEST` (distinct `EMPTY_TRANSCRIPT` code), ahead of the generic 500 branch.
- Coverage: a pipeline test (`synthesizeHandoffPrompt` on a zero-message project rejects) and a route test (`400`).

Heads-up on an incoming contract change: a review on the companion UI PR (#2264) is making this endpoint **conversation-scoped** — `HandoffRequest` gains a required `conversationId` and the export is filtered to that one conversation. The empty-transcript guard then sits after the scoped export, so "empty" becomes "the selected conversation has no messages". Both land together in the next #1718 push — flagging so the contract delta isn't a surprise.

Adjacent (out of scope, flagged so it's not invisible): `/finalize` calls the same `exportProjectTranscript` with no empty-transcript guard either — same latent token-burn. Happy to file a follow-up issue if you'd like it tracked.

apps/daemon/src/handoff-routes.ts
```diff
@@ -0,0 +1,176 @@
...
+
+      // Handoff is conversation-scoped — the conversation must exist AND
+      // belong to this project, otherwise the synthesized transcript would
+      // either be empty or (worse) summarize an unrelated project's chat.
+      const conversation = getConversation(db, conversationId);
+      if (!conversation || conversation.projectId !== req.params.id) {
+        return sendApiError(
+          res,
+          404,
+          'CONVERSATION_NOT_FOUND',
```
nettee, line 102: This branch adds two new daemon error codes here and below (`CONVERSATION_NOT_FOUND` here, `EMPTY_TRANSCRIPT` at line 142), but `packages/contracts/src/errors.ts` still does not include either value in `API_ERROR_CODES` / `ApiErrorCode`. That leaves the live `/api/projects/:id/handoff` response surface ahead of the shared contract: typed callers cannot represent these payloads, contract-driven tests/helpers reject them, and the next UI/CLI consumer has no canonical enum member for the new failures. Please add both codes to the contracts error union (or reuse an existing shared code if you want to keep the public surface smaller) and extend the contract coverage so the daemon route and `@open-design/contracts` cannot drift again.

nettee (changes requested):
Found one blocking issue in the new conversation-scoped handoff route: it now emits error codes that are not part of the shared API contract.


apps/daemon/src/handoff-design.ts
bankielewicz: Fixed in `42fd54a8` (just pushed to `feat/handoff-endpoint`).

- `synthesizeHandoffPrompt` now throws a typed `EmptyTranscriptError` when the exported transcript has `messageCount === 0`; `handoff-routes.ts` maps it to `400 EMPTY_TRANSCRIPT` ahead of the generic 500. The guard fires before the Anthropic call — a pipeline test asserts the upstream `fetchImpl` is never invoked.
- Coverage: a pipeline test (`synthesizeHandoffPrompt` on a zero-message conversation rejects with `EmptyTranscriptError`, no upstream call) and a route test (`400 EMPTY_TRANSCRIPT` on an empty conversation).

The conversation-scoping change from the #2264 review landed in the same commit: `HandoffRequest` gained a required `conversationId`, the route 404s if it is not in the project, and the transcript export is filtered to that one conversation. `HANDOFF_SCHEMA_VERSION` is bumped to 2.

Daemon typecheck + the 56 handoff + 26 transcript-export tests + `guard` are green. (3 unrelated `project-watchers.test.ts` failures are pre-existing WSL2 chokidar flakiness — that module is byte-identical to `origin/main` and untouched here.)

apps/daemon/src/handoff-routes.ts
bankielewicz: Fixed in `aa1d9b81` (pushed to `feat/handoff-endpoint`).

`CONVERSATION_NOT_FOUND` and `EMPTY_TRANSCRIPT` are now registered in `packages/contracts/src/errors.ts` `API_ERROR_CODES` (placed next to `PROJECT_NOT_FOUND`), so they are canonical `ApiErrorCode` members. A contract test in `handoff-contract.test.ts` asserts both are present — paired with the existing `handoff-route.test.ts` cases that assert the route emits them, the daemon route and `@open-design/contracts` can't drift apart again.

Contracts build + 90 contract tests, daemon typecheck, and the 62 handoff/CLI tests are green.

apps/daemon/src/handoff-cli.ts
```diff
@@ -0,0 +1,154 @@
...
+    }
+    const result = payload as HandoffResponse;
+    if (options.json) {
+      writeJson(result);
+    } else {
+      // Default output is the synthesized prompt itself, so it pipes
+      // cleanly into a file or another command.
+      process.stdout.write(`${result.prompt}\n`);
+    }
+    return { exitCode: 0 };
```
nettee: `runProjectHandoff()` currently treats any 2xx response as success even when the body is not valid `HandoffResponse` JSON. `resp.json().catch(() => ({}))` swallows parse failures, then the success path casts whatever came back to `HandoffResponse` and prints `result.prompt` / exits 0 without checking that `prompt`, `model`, or the token fields are actually present. That means a broken daemon/proxy response can make `od project handoff` print `undefined` (or `{}` under `--json`) while scripts still see a successful exit code, which violates the fail-fast contract for required API responses. Please fail this path when JSON parsing fails or the parsed object is missing the required handoff fields, and add a regression test for a 200 response with malformed or shape-invalid JSON.

nettee (changes requested):
Found one blocking issue in the new handoff CLI surface.


apps/daemon/src/handoff-cli.ts
bankielewicz: Fixed in `3809db22`.

`runProjectHandoff` now validates the 2xx body with an `isHandoffResponse` type guard (checks `prompt`/`model` are strings and the three token fields are numbers). A 200 whose body is unparseable or shape-invalid is now `fail()`-ed (`exitCode 1`, `malformed handoff response`) instead of printing `undefined` / `{}` and exiting 0 — `resp.json()` parse failures route through the same guard via `undefined`.

Regression tests added to `handoff-cli.test.ts`: a 200 with a shape-invalid JSON body (asserts no `undefined` in stdout) and a 200 with an unparseable body — both assert a non-zero exit. Daemon typecheck + the 7 CLI tests + `guard` are green.

apps/daemon/src/cli.ts
```diff
@@ -3774,6 +3779,13 @@ Common options:
       console.log(`[project] deleted ${id}`);
       return;
     }
+    case 'handoff': {
+      // Conversation-scoped resume handoff. runProjectHandoff self-parses
+      // `rest` (projectId positional + flags) and resolves the daemon URL.
+      const { exitCode } = await runProjectHandoff(rest);
```
nettee: `runProject()` says this branch lets `runProjectHandoff` "self-parse" `rest`, but the shared `parseFlags(rest, PROJECT_…)` at the top of `runProject()` has already run before control reaches this case. That means malformed invocations like `od project handoff <id> --unknown` or `--max-tokens` without a value never hit the structured `fail(...)` path in `handoff-cli.ts`; they throw out of the generic project parser instead. For an agent-facing CLI surface that is supposed to fail fast with machine-readable errors, that makes the top-level `od project handoff` entrypoint behave differently from the helper and from the new tests, which only exercise `runProjectHandoff()` directly. Please short-circuit `sub === 'handoff'` before the generic `parseFlags` / `projectDaemonUrl` work (or catch and remap those parser errors here), and add one integration test through `runProject()` so malformed handoff flags stay covered at the real entrypoint.

nettee (commented):
Found one non-blocking follow-up in the new `od project handoff` CLI wiring.


apps/daemon/src/cli.ts
bankielewicz: Fixed in `e76e4450`.

`runProject` now short-circuits `sub === 'handoff'` immediately after computing `sub`/`rest` — before the shared `parseFlags(PROJECT_*)` and `projectDaemonUrl` calls. So `od project handoff` runs exactly `runProjectHandoff(rest)` with no intervening parsing; a malformed invocation (`--unknown`, `--max-tokens` with no value) now hits `handoff-cli.ts`'s structured `fail()` instead of throwing out of the generic project parser. The old `case 'handoff'` is removed from the switch and the handoff-only flags dropped from `PROJECT_STRING_FLAGS`.

On the integration test: `runProject` lives in `cli.ts`, which runs its `SUBCOMMAND_MAP` dispatch at import — the reason CLI logic lives in sibling modules (`artifacts-cli.ts`, `tools-connectors-cli.ts`, ...) is precisely so tests don't import `cli.ts`. Rather than make handoff the first test to break that pattern, I added the malformed-flag cases (`--unknown`, `--max-tokens` with no value) to `handoff-cli.test.ts`. Post-short-circuit there is no logic between `runProject` and `runProjectHandoff` for the handoff sub, so the helper's tests are the entrypoint's behavior. Happy to add an exported-`runProject` test if you'd still want the literal form despite the import-dispatch cost.

apps/daemon/src/cli.ts
```diff
@@ -3686,6 +3690,16 @@ Common options:
   }
   const sub = args[0];
   const rest = args.slice(1);
+  // Handoff owns its own flag parsing, daemon-URL resolution, and
+  // structured fail() output. Dispatch it before the generic project
+  // parser below so a malformed `od project handoff` invocation
+  // (`--unknown`, `--max-tokens` with no value) hits handoff-cli's
+  // machine-readable fail() path instead of throwing out of parseFlags.
+  if (sub === 'handoff') {
```
nettee: `runProjectHandoff()` now owns malformed-flag parsing, but `od project handoff --help` still never reaches that helper because `runProject()` exits on any `args.includes("--help")` before the new `sub === "handoff"` branch runs. That means the top-level entrypoint only prints the generic `od project` usage and hides the dedicated handoff flags/description you added in `handoff-cli.ts`, so the shipped command and the unit-tested helper still disagree. Consider moving the `sub === "handoff"` short-circuit above the generic help gate (or special-casing handoff in that gate), and add a regression test for `od project handoff --help` through the real `runProject` entrypoint.

nettee (commented):
Found one non-blocking follow-up in the new `od project handoff` dispatch path.


apps/daemon/src/cli.ts
bankielewicz: Fixed in `bd8b9d3c`.

`runProject` now short-circuits `args[0] === 'handoff'` **above** the generic `--help` gate, so `od project handoff --help` reaches `runProjectHandoff`'s own USAGE (the one that lists `--conversation` / `--api-key` / `--model` / `--base-url` / `--max-tokens` / `--json`) instead of falling into the generic od-project help. The earlier short-circuit moved up; the entrypoint and helper now print identical help.

`handoff-cli.test.ts` gains a `--help` case that asserts the dedicated flag names appear in stdout and the daemon is never called. Same architectural note as before re: the literal `runProject` integration test — `cli.ts` runs its `SUBCOMMAND_MAP` dispatch at import, the reason CLI logic lives in sibling modules — but post-short-circuit `runProject` runs exactly `runProjectHandoff(args.slice(1))` for the handoff sub with no intervening logic, so the helper's `--help` test is the entrypoint's behavior.

Daemon typecheck + 11 handoff/CLI tests + `guard` green.

(Not restacking #2264 here — mrcfps's clean approval is on `9874889d` and a restack would re-stale it; #2264 will pick up these changes naturally when #1718 merges to main.)

apps/daemon/src/handoff-cli.ts
```diff
@@ -0,0 +1,176 @@
+// `od project handoff` — the CLI surface for the resume-conversation
+// handoff capability. Per AGENTS.md "Capability exposure (UI/CLI
+// dual-track)", every user-facing capability must be reachable through
+// the `od` CLI as well as the web UI; both drive the same
+// `POST /api/projects/:id/handoff` endpoint.
```
nettee, line 5: Blocking: this PR introduces a new user-facing handoff capability through both `POST /api/projects/:id/handoff` and `od project handoff`, but the diff never lands the matching `apps/web` surface. `AGENTS.md:90-95` makes the UI/CLI dual-track rule explicit: new capabilities have to ship as HTTP + web UI + CLI in the same PR, and the current changed-file list only touches `apps/daemon/**` and `packages/contracts/**`. Merging this partial slice would leave the main product UI unable to discover or invoke the feature while the CLI/API already expose it. Suggested fix: land the web provider/UI entry point in this PR, or keep the endpoint/CLI internal until the web surface ships in the same change.

apps/daemon/src/handoff-cli.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+const USAGE = `Usage:
+  od project handoff <projectId> --conversation <id> --api-key <key> --model <model>
+                     [--base-url <url>] [--max-tokens <n>] [--daemon-url <url>] [--json]
+
+Synthesizes a "resume conversation" handoff prompt from one conversation's
+transcript via the local daemon. Prints the prompt to stdout; --json emits
+the full response (prompt + model + token usage).
+`;
```
nettee, line 44: Blocking: this usage text, the parser below, and the `handoff requires --api-key <key>` guard at lines 129-132 make a raw argv secret the only supported way to call the new CLI. That leaks BYOK credentials into shell history, `ps` output, CI logs, and any wrapper that records command lines, which is a concrete secret-handling regression for a user-facing surface. Suggested fix: add a non-argv transport such as `--api-key-file <path|->` and/or an environment-variable path, make that the documented flow here, and cover it in `apps/daemon/tests/handoff-cli.test.ts` so the safer path stays wired.

nettee (changes requested):
Found two blocking issues in the new handoff surface.


bankielewicz:
@nettee — the argv-secret finding (your inline at `apps/daemon/src/handoff-cli.ts:44`) is addressed in PR #2454, off `main`.

This branch (`feat/handoff-endpoint @ bd8b9d3c`) is structurally unmergeable post-#2264 squash: its content shipped to main as commit `c530d163` (a single squash commit), so every file on this branch now conflicts with itself against main. Any further push here is orphaned.

The follow-up PR adds:

- `--api-key-file <path|->` (file path or stdin), mirroring the `--prompt-file` / `--body-file` precedent in `apps/daemon/src/cli.ts:4817-4828` (cited in AGENTS.md "Capability exposure").
- `ANTHROPIC_API_KEY` env-var fallback (standard upstream convention, already used in `apps/daemon/src/memory-llm.ts:136`).
- `--api-key <key>` still works but emits a one-line stderr deprecation warning so the safer paths stay discoverable.

Precedence (highest wins): `--api-key` (warned) → `--api-key-file` → `ANTHROPIC_API_KEY`. Mirrors the layered approach in `memory-llm.ts:135-156` and standard CLI conventions.

Branch-complete test coverage in `apps/daemon/tests/handoff-cli.test.ts` — 10 new tests, all red against the existing source (parser rejects `--api-key-file` as `unknown option`) and green on the follow-up branch. The 9 existing tests still pass.

Closing this PR now since the daemon + CLI content is already on main via #2264 and there's no merge path remaining. Review focus moves to #2454.

PR #1726, opened 2026-05-14
Prevent imported Claude canvases from zooming on scroll

leessju:
## Why

This PR fixes a bug found while importing and working with a Claude Design export inside Open Design.

### Problem

Imported Claude Design canvases could zoom while the user was only doing a normal two-finger vertical scroll. The symptom was confusing because Open Design's toolbar still showed `100%`, but the design inside the iframe visually scaled and jumped.

### Cause

The imported `design-canvas.jsx` used this heuristic to distinguish mouse-wheel zoom from trackpad pan:

```js
const isMouseWheel = (e) =>
  e.deltaMode !== 0 ||
  (e.deltaX === 0 && Number.isInteger(e.deltaY) && Math.abs(e.deltaY) >= 40);
```

On macOS, ordinary two-finger vertical scroll can arrive as a large integer `deltaY` with `deltaX === 0`. That made the artifact classify regular scrolling as mouse-wheel zoom and call:

```js
zoomAt(e.clientX, e.clientY, Math.exp(-Math.sign(e.deltaY) * 0.18));
```

Captured evidence from the repro showed the host toolbar stayed at `100%`, while the imported iframe canvas itself scaled:

```json
{
  "toolbarZoom": "100%",
  "innerTransform": "matrix(1.13589, 0, 0, 1.13589, -2972.1, -4263.43)",
  "scale": 1.13589
}
```

### Resolution

Normalize imported Claude Design `design-canvas.jsx` files during zip import:

- Plain wheel input now pans the infinite canvas.
- `Cmd + wheel` remains the explicit zoom gesture.
- Safari/native `gesture*` events are ignored by the imported canvas so ordinary scrolling cannot accidentally scale it.
- Host toolbar zoom still uses the existing `__dc_set_zoom` path.

This keeps the fix at the import boundary where the artifact code enters Open Design, instead of trying to intercept iframe wheel events from the host after the artifact has already handled them.

### Process

1. Reproduced the symptom in a headed Playwright browser.
2. Added temporary host/iframe logging locally to distinguish Open Design host zoom from imported iframe canvas zoom.
3. Confirmed the active broken project was an imported Claude Design project, not the earlier restored standalone artifact.
4. Patched the project-local `design-canvas.jsx` and verified:
   - normal two-finger scroll pans,
   - `Cmd + scroll` zooms,
   - the iframe canvas scale no longer changes on plain scroll.
5. Moved the durable fix into `importClaudeDesignZip` so future imports are normalized automatically.
6. Added a daemon import regression test that proves the old heuristic is removed and the new pan/Cmd-zoom behavior is written into imported `design-canvas.jsx` files.

## What users will see

Users importing Claude Design ZIP exports can scroll around the imported infinite canvas with a normal two-finger vertical scroll without the canvas unexpectedly zooming. They can still intentionally zoom the imported canvas with `Cmd + scroll` or the existing toolbar zoom control.

## Surface area

- [x] **Keyboard shortcut** — new or changed
- [ ] **Extension point** — new entry under `skills/`, `design-systems/`, `design-templates`, or `craft`, or change to the skills protocol

## Screenshots

No new UI surface. The visible behavior change is in imported artifact canvas input handling.

Manual evidence from the project-local repro before moving the fix into the importer:

```json
{
  "before": { "scale": 1, "y": 0 },
  "plainWheel": { "scale": 1, "y": -72 },
  "cmdWheelZoomIn": { "scale": 3.32012 },
  "cmdWheelZoomOut": { "scale": 1 }
}
```

## Bug fix verification

- Test path that reproduces the bug: `apps/daemon/tests/claude-design-import.test.ts`
- Did the test go red on `main` and green on this branch? Not run on `main`; the test encodes the imported `design-canvas.jsx` heuristic and asserts the importer rewrites it away on this branch.
- Additional manual verification: headed Playwright/browser repro confirmed the failing path was iframe-local `design-canvas.jsx` scale changing while the host toolbar stayed at `100%`.

## Validation

- `pnpm --filter @open-design/daemon exec vitest run tests/claude-design-import.test.ts`
- `pnpm guard`


apps/daemon/src/claude-design-import.ts
```diff
@@ -80,6 +80,58 @@ export async function importClaudeDesignZip(zipPath: string, projectDir: string)
...
+      tf.current.y -= dy;
+      apply();
+    };
+
+    const onWheel = (e) => {
+      e.preventDefault();
+      e.stopPropagation();
+      if (isGesturing) return;
+      if (e.metaKey) {
+        zoomAt(e.clientX, e.clientY, Math.exp(-e.deltaY * 0.01));
```
PerishCode, line 116: **Cmd+wheel zoom step is too aggressive for a physical mouse wheel after this collapse.**

The original heuristic this rewriter replaced split Cmd+wheel into two paths — smooth (`Math.exp(-e.deltaY * 0.01)`) for trackpad pinch where `deltaY` is small/fractional, and notched (`Math.exp(-Math.sign(e.deltaY) * 0.18)`) for a physical wheel where `deltaY` arrives in chunks of ~100/120 per click. The new code routes both through the smooth formula.

For a real mouse wheel, `Math.exp(-100 * 0.01) ≈ 0.367` — a single notch click shrinks the imported canvas by ~63%. The old notched branch was ~0.835 (≈17% per click), so Cmd + mouse-wheel users will overshoot dramatically on every click. Trackpad pinch is unaffected because its `deltaY` is small.

This is a deliberate-looking simplification (the PR body says "Cmd + wheel remains the explicit zoom gesture"), but it isn't called out as a UX change for the mouse-wheel case, and the daemon test doesn't cover it.

Suggested change — keep the notched branch alive inside the Cmd path:

```js
if (e.metaKey) {
  const isNotched =
    e.deltaMode !== 0 ||
    (e.deltaX === 0 && Number.isInteger(e.deltaY) && Math.abs(e.deltaY) >= 40);
  const factor = isNotched
    ? Math.exp(-Math.sign(e.deltaY) * 0.18)
    : Math.exp(-e.deltaY * 0.01);
  zoomAt(e.clientX, e.clientY, factor);
  return;
}
```

Note this is the same heuristic the rewriter is removing from the import path, but it's safe to reuse here because the gating is now "Cmd held" rather than "any wheel," so a plain two-finger scroll without Cmd can never enter it. If the call is intentional (i.e., "we only support trackpad-style Cmd zoom for imported canvases"), worth documenting that in the new comment block so reviewers and future-you can find the decision.


apps/daemon/src/claude-design-import.ts
```diff
@@ -80,6 +80,58 @@ export async function importClaudeDesignZip(zipPath: string, projectDir: string)
...
+        return;
+      }
+      panByWheel(e);
+    };
+`);
+  const gestureBlock = /    \/\/ Safari sends native gesture\* events for trackpad pinch with a smooth\n[\s\S]*?    const onGestureEnd = \(e\) => \{ e\.preventDefault\(\); isGesturing = false; \};/;
```
PerishCode, line 122: **Both rewrite regexes fail silently on Claude Design format drift, with no signal to anyone.**

Both `wheelBlock` (line 91) and `gestureBlock` (line 122) match on exact 4-space indentation, exact comment prose ("Mouse-wheel vs trackpad-scroll heuristic." / "Safari sends native gesture* events for trackpad pinch with a smooth"), and exact arrow-function/closing-brace shape. If a future Claude Design export switches to tabs, rephrases the comment, or breaks `onGestureEnd` across multiple lines, `wheelBlock.test(source)` returns false, `normalizeDesignCanvasWheelHandling` returns the original `source`, and the importer writes the original buggy `design-canvas.jsx` verbatim. The user sees the same zoom-on-scroll bug, with no log line, no warning, and no failing test — only the regression report.

The daemon test at `apps/daemon/tests/claude-design-import.test.ts` only verifies the happy path on the current fixture, so it can't catch format drift either.

Two suggested mitigations, ideally both:

1. **Warn when normalization is skipped on a `design-canvas.jsx`.** The basename check at line 84 already proves this file came from a Claude Design canvas; if neither block matches, that's a strong signal the heuristic needs an update:

   ```ts
   function normalizeImportedClaudeDesignFile(relPath: string, body: Buffer): Buffer {
     if (path.basename(relPath) !== 'design-canvas.jsx') return body;
     const source = body.toString('utf8');
     const normalized = normalizeDesignCanvasWheelHandling(source);
     if (normalized === source) {
       console.warn('[claude-design-import] design-canvas.jsx found but wheel-handler regex did not match; imported canvas may zoom on scroll');
     }
     return normalized === source ? body : Buffer.from(normalized, 'utf8');
   }
   ```

   Even a single `console.warn` in the daemon log gives ops/support something to grep when the same bug reappears.

2. **Add a negative test in the fixture matrix.** Right now there's one positive fixture; add a 2-space-indent or tab-indent variant and assert the normalizer logs / no-ops as expected so changes to the regexes are forced through a deliberate update.

This is the kind of fix that's silently going to bit-rot the next time Anthropic touches the canvas export template — a 5-line warning today saves the next debugging round.


PerishCode (commented):
Hi @leessju — read through the import normalizer and the preview-state bridge changes. The Claude Design canvas fix lands at the right boundary (importClaudeDesignZip), and the daemon test does pin the new shape. A few non-blocking concerns inline: the collapsed Cmd-zoom formula regresses physical-mouse-wheel UX, and the regex-based rewriter has no signal when Claude Design's export format drifts. Also flagging at the top level that the diff bundles a second, larger "preserve HTML preview state across mode toggles" change (urlModeBridge gating, filesRefreshKey wiring, new od:preview-scroll / __dc_viewport postMessage protocols, comment+edit toolbar behavior) that the PR body doesn't describe — readers reviewing against the title will miss it. Worth either splitting it out or updating the Why / Surface area sections to cover both. None of this is a merge blocker, just things to weigh before landing.



lefarcen:
Hey @leessju — pushed a merge commit (`87a564df`) to clear the conflicts with `main`. Two text-level conflicts, both clean:

- `apps/web/src/components/FileViewer.tsx` (line 62) and the matching `tests/components/file-viewer-render-mode.test.ts` import: your branch added `hasUrlModeBridge` to the `file-viewer-render-mode` imports while `main` independently added `htmlNeedsSandboxShim`. Both functions are still exported from `file-viewer-render-mode.ts`, so the resolution was just to keep both names in the import lists. `FileViewer.tsx` already calls both (line 4022 `hasUrlModeBridge(source)`, line 4034 `htmlNeedsSandboxShim(source)`), so nothing else needed touching.

Verified locally with `pnpm --filter @open-design/web typecheck` (clean) and `pnpm exec vitest run tests/components/file-viewer-render-mode.test.ts tests/runtime/srcdoc.test.ts` (40/40 pass).

Feel free to drop or rebase out the merge commit if you'd rather rebase the branch onto main yourself.

lefarcen (approved):

PR #1733, opened 2026-05-14
feat(odml): seed dialect, tokens, skill prompt — chala-ai-mobile + od-dialect foundation

Zacc-r:
## Why

**Use case:** I'm building Chala.AI (an iOS fitness app, the first of N apps in my workspace) and want Open Design to be the canonical source of truth for the design system. Today the daemon serves DESIGN.md as raw markdown (`GET /api/design-systems/:id`) and `chala-ai-mobile`-style skill output is freeform HTML — useful for prototyping but neither structured enough to push the same tokens to multiple consumer apps (iOS, web, React Native) without per-output reinterpretation.

**Pain being addressed:** Token drift between consumer apps. Each consumer hand-mirrors hex/spacing values from DESIGN.md; nothing enforces parity. A tokens-only fix isn't enough — even with a `tokens.json` file, skill output that contains `<div class="..." style="color: #fff">` still pulls every consumer back into per-platform reinterpretation. The leverage point is a restricted dialect (ODML) where token names are the only syntactically legal way to specify a styling value.

I considered opening a discussion first per CONTRIBUTING. Choosing PR-first because the scope is contained: this lands the foundation only (AST types + tokens + the skill prompt), not the cross-cutting daemon route or the SwiftUI translator. Those are designed but deferred to follow-up PRs.

## What users will see

**Skill author / prompt engineer:** a new `chala-ai-mobile` skill exists with a system prompt that constrains output to ODML — a restricted `<od-*>` XML dialect, 20-component vocabulary v1. The skill's iframe preview renders ODML via Web Component shims so designers still see the layout.

**Design-system curator:** `design-systems/chala-ai/` now ships a DTCG-compliant `tokens.json` (machine-readable source of truth) and `tokens-audit.md` (one-time reconciliation against the prose) alongside the existing `DESIGN.md`.

**Other API consumers:** nothing — no daemon routes added or changed.

## Surface area

- [ ] **UI** — none
- [ ] **API / contract** — none in this PR (a future `GET /api/design-systems/:id/tokens.json` route is planned but not included here)
- [x] **Extension point** — new `skills/chala-ai-mobile/`, new `design-systems/chala-ai/`, plus a new `packages/od-dialect/` workspace package
- [ ] **i18n keys** — none
- [ ] **New top-level dependency** — none (root `package.json` unchanged)
- [ ] **Default behavior change** — none (the `chala-ai-mobile` skill is brand new, not a modification of an existing skill)

## Screenshots

N/A — no UI surface in this PR.

## Bug fix verification

N/A — feature contribution, not a bug fix.

## Validation

Ran from the repo root against commit `030e383b`:

- `pnpm --filter @open-design/od-dialect typecheck` → passes
- `pnpm guard` → passes (after adding `skills/chala-ai-mobile/assets/od-elements.js` to the residual-JS allowlist with the documented "no-build browser context" reason)

I did not run the full test suite because the changes are additive (new package + new skill + new design system + one allowlist entry); none of the existing apps/packages import from the new package yet.

## Notes for reviewers

1. **ODML AST is intentionally fleshed out before the parser/validator land.** It defines 20 element kinds + a violation discriminated union but no parser yet. The next two pieces — a Layer 1 `validate.ts` for the skill and a Layer 3 SwiftUI translator — both import from this AST; defining it up front prevents the two implementations drifting on tree shape.
2. **`od-elements.js` is declared throwaway preview code.** It exists so the iframe preview renders something visual; any future per-platform translator never reads it. Hard-rule discipline lives in the skill prompt + (forthcoming) validator, not in the preview shims.
3. **Token authoring is hand-seeded v1.** The plan was to extract tokens via Claude from DESIGN.md at build time; the extractor is deferred to a follow-up because the seed needed human review against the consumer iOS app's existing `Theme.swift` constants (audit in `tokens-audit.md`).
4. **No daemon changes in this PR.** A future PR will add `GET /api/design-systems/:id/tokens.json` so consumers (iOS + web) can fetch tokens at build time, plus a `skill-runner` integration that invokes the validator after each ODML render and re-prompts the LLM on violations.

### DIFF
--- a/design-systems/chala-ai/tokens.json
+++ b/design-systems/chala-ai/tokens.json
@@ -1,0 +1,130 @@
+{
+  "$schema": "https://design-tokens.github.io/community-group/format/",
+  "$description": "Chala.AI V2 design tokens (DTCG). Hand-authored seed v1; reconciled against design-systems/chala-ai/DESIGN.md and projects/Gym.AI/Chala.AI/apps/ios/ChalaAI/ChalaAI/Views/Components/Theme.swift on 2026-05-14. See tokens-audit.md for the reconciliation table. Future updates flow via tools/tokens-build/extract.ts (LLM extraction + human PR review).",
+  "color": {
+    "bg": {
+      "$value": "#050505",
+      "$type": "color",
+      "$description": "Absolute black canvas. Primary screen background."
+    },
+    "bg-elevated": {
+      "$value": "#0B0B0B",
+      "$type": "color",
+      "$description": "Slightly elevated surface — cards on cards, sheets, second tier."
+    },
+    "bg-overlay": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "Opacity-only overlay surface. Use with opacity in the renderer — designers wanting a visible overlay should layer with a fill ~10% opacity.",
+      "$extensions": {
+        "com.open-design.opacity": 0.06
+      }
+    },
+    "fg": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "Primary text and active-state foreground."
+    },
+    "fg-muted": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "Body copy, secondary headings (55% opacity on bg).",
+      "$extensions": {
+        "com.open-design.opacity": 0.55
+      }
+    },
+    "fg-subtle": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "Labels, metadata, timestamps (32% opacity on bg).",
+      "$extensions": {
+        "com.open-design.opacity": 0.32
+      }
+    },
+    "accent": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "System is monochrome — accent is pure white. Used for primary fill backgrounds and active emphasis."
+    },
+    "accent-fg": {
+      "$value": "#0B0B0B",
+      "$type": "color",
+      "$description": "Foreground text on accent surfaces (e.g. primary button labels)."
+    },
+    "success": {
+      "$value": "#34C759",
+      "$type": "color",
+      "$description": "Legacy iOS green. Avoid in new V2 surfaces; reserved for status badges."
+    },
+    "warning": {
+      "$value": "#FF9500",
+      "$type": "color",
+      "$description": "Legacy iOS orange. Avoid in new V2 surfaces; reserved for status badges."
+    },
+    "danger": {
+      "$value": "#FF3B30",
+      "$type": "color",
+      "$description": "Legacy iOS red. Used for destructive button variant and danger badges."
+    },
+    "border": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "Default hairline border (8% opacity on bg). All 1px borders use this.",
+      "$extensions": {
+        "com.open-design.opacity": 0.08
+      }
+    },
+    "border-strong": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "Elevated border (16% opacity on bg) — interactive element rings.",
+      "$extensions": {
+        "com.open-design.opacity": 0.16
+      }
+    }
+  },
+  "spacing": {
+    "none": { "$value": "0", "$type": "dimension" },
+    "xs":   { "$value": "4",  "$type": "dimension" },
+    "sm":   { "$value": "8",  "$type": "dimension" },
+    "md":   { "$value": "16", "$type": "dimension" },
+    "lg":   { "$value": "24", "$type": "dimension" },
+    "xl":   { "$value": "32", "$type": "dimension" },
+    "2xl":  { "$value": "48", "$type": "dimension" },
+    "3xl":  { "$value": "64", "$type": "dimension" }
+  },
+  "radius": {
+    "none": { "$value": "0",    "$type": "dimension" },
+    "sm":   { "$value": "6",    "$type": "dimension" },
+    "md":   { "$value": "8",    "$type": "dimension" },
+    "lg":   { "$value": "15",   "$type": "dimension" },
+    "xl":   { "$value": "24",   "$type": "dimension" },
+    "full": { "$value": "9999", "$type": "dimension" }
+  },
+  "typography": {
+    "font-family": {
+      "sans": { "$value": "Geist", "$type": "fontFamily", "$description": "Primary UI font. 400/500/600/700 weights." },
+      "mono": { "$value": "GeistMono", "$type": "fontFamily", "$description": "All UI chrome — labels, timestamps, stats, tab bar." }
+    },
+    "style": {
+      "display":  { "$value": { "fontFamily": "{typography.font-family.sans}", "fontSize": "42", "fontWeight": "500", "letterSpacing": "-0.02em", "lineHeight": "1.1"  }, "$type": "typography", "$description": "Calorie counts, hero numerics." },
+      "title":    { "$value": { "fontFamily": "{typography.font-family.sans}", "fontSize": "36", "fontWeight": "500", "letterSpacing": "-0.02em", "lineHeight": "1.15" }, "$type": "typography", "$description": "Screen headings." },
+      "headline": { "$value": { "fontFamily": "{typography.font-family.sans}", "fontSize": "22", "fontWeight": "500", "letterSpacing": "-0.02em", "lineHeight": "1.2"  }, "$type": "typography", "$description": "Card headings, subheadings." },
+      "body":     { "$value": { "fontFamily": "{typography.font-family.sans}", "fontSize": "15", "fontWeight": "400", "letterSpacing": "0",       "lineHeight": "1.4"  }, "$type": "typography", "$description": "Descriptions, chat messages." },
+      "caption":  { "$value": { "fontFamily": "{typography.font-family.mono}", "fontSize": "10", "fontWeight": "400", "letterSpacing": "0.18em",  "lineHeight": "1.2",  "textCase": "uppercase" }, "$type": "typography", "$description": "Uppercase mono labels — UI chrome." },
+      "mono":     { "$value": { "fontFamily": "{typography.font-family.mono}", "fontSize": "14", "fontWeight": "400", "letterSpacing": "0",       "lineHeight": "1.3"  }, "$type": "typography", "$description": "Tabular numerics — stats, counts." }
+    }
+  },
+  "icon-size": {
+    "xs": { "$value": "12", "$type": "dimension" },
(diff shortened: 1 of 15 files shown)


lefarcen (commented):
Hey @Zacc-r! 👋 Thanks for this well-scoped contribution — the problem statement is clear and I appreciate that you thought carefully about the rollout order (AST-first, parser/validator deferred, daemon route follow-up). The ODML idea of enforcing token-name-only styling at the syntax level is a solid architectural choice for preventing token drift.

I've assigned @mrcfps for code review and added @qiongyu1999 for product sign-off (standard flow for new Extension point features). CI will run shortly. In the meantime feel free to address any questions that come up from the reviewer.

skills/chala-ai-mobile/assets/od-elements.js
```diff
@@ -0,0 +1,562 @@
...
+const ICON_SIZE = { xs: 12, sm: 16, md: 20, lg: 28, xl: 36 };
+const AVATAR_SIZE = { sm: 32, md: 48, lg: 72 };
+const BUTTON_SIZE_PADDING = {
+  sm: '6px 12px',
+  md: '10px 16px',
+  lg: '14px 20px',
+};
+
+function tokenStyle(prefix, name) {
+  return name ? `var(--od-${prefix}-${name})` : null;
```
mrcfps: The helper on this line produces `var(--od--fg)` when callers pass an empty prefix, but `TOKEN_CSS` only defines names like `--od-fg` and `--od-accent-fg`. Both `od-text` (line 259) and `od-icon` (line 284) call `tokenStyle('', ...)`, so valid ODML such as `<od-text color="fg-muted">` or the `accent-fg` button-label example in `SKILL.md` will not preview with the requested token color. Please special-case the empty-prefix path (or pass the full custom-property name through) so the preview resolves the same color tokens the prompt allows.

skills/chala-ai-mobile/assets/od-elements.js
```diff
@@ -0,0 +1,562 @@
...
+    const radius = this.getAttribute('radius') || 'md';
+    const bg = this.getAttribute('background');
+    const shadow = this.attachShadow({ mode: 'open' });
+    shadow.innerHTML = `
+      <style>${TOKEN_CSS}
+        :host {
+          display: block;
+          padding: var(--od-space-${padding});
+          border-radius: var(--od-radius-${radius});
+          background: ${bg ? tokenStyle('bg', bg === 'bg-elevated' ? 'elevated' : bg === 'bg-overlay' ? 'overlay' : '') : 'rgba(255,255,255,0.04)'};
```
mrcfps: The prompt and AST both allow any `ODColor` token for `background`, but this branch only handles `bg-elevated` and `bg-overlay`. For any other valid token, `tokenStyle('bg', ...)` receives an empty name, which leaves `od-screen` falling back to `var(--od-bg)` and makes `od-card` emit `background: null;`. That means valid ODML like `<od-card background="accent">` or `<od-screen background="danger">` silently previews with the wrong surface. Please either tighten the ODML contract to the surface-only tokens or map all allowed background tokens directly instead of hard-coding two special cases.

skills/chala-ai-mobile/assets/od-elements.js
```diff
@@ -0,0 +1,562 @@
...
+    shadow.innerHTML = `
+      <style>${TOKEN_CSS}
+        :host {
+          display: inline-block;
+          ${TEXT_STYLES[style] || TEXT_STYLES.body}
+          color: ${tokenStyle('', color === 'fg-muted' ? 'fg-muted' : color === 'fg-subtle' ? 'fg-subtle' : color)};
+          text-align: ${align === 'leading' ? 'left' : align === 'trailing' ? 'right' : 'center'};
+        }
+      </style>
+      <span>${text}</span>
```
mrcfps: This preview is rendering model-authored and user-authored strings by interpolating them directly into `shadow.innerHTML`. Because skill previews run in `sandbox="allow-scripts"` iframes, a value like `text="&lt;img src=x onerror=alert(1)&gt;"` becomes live DOM here instead of inert text, and the same pattern repeats later for `placeholder`, `label`, `title`, and tab labels in this file. That turns untrusted ODML content into scriptable markup inside the preview surface. Please build these nodes with `textContent` / DOM attribute setters (or escape every interpolated string before concatenation) so the shim treats ODML content as data, not HTML.

mrcfps (commented):
@Zacc-r Thanks for the ODML foundation work — the AST-first split makes sense, and I re-ran `pnpm --filter @open-design/od-dialect typecheck` plus `pnpm guard` on this head after installing the workspace dependencies. I left a few non-blocking comments on the new preview shim because there are a couple of spots where valid ODML either renders incorrectly or gets inserted unsafely in the sandbox preview.


Zacc-r:
**Self-review follow-up — added `tokens.css` + `components.html` pair (commit `92053727`)**

While planning follow-on work I read upstream's design-system loader (`apps/daemon/src/design-systems.ts`) and the system-prompt composer (`apps/daemon/src/prompts/system.ts` lines 166–330) more carefully and found a concrete gap in this PR as originally drafted: the structured token channel (`designSystemTokensCss` + `designSystemFixtureHtml`) keys off `tokens.css` + `components.html` being present alongside `DESIGN.md`. Without them, chala-ai falls into the prose-only cohort and the existing injection blocks skip silently for this brand.

This commit adds:

- `design-systems/chala-ai/tokens.css` — standard-schema CSS-variable contract per `craft/color.md`. Mirrors `DESIGN.md` values into the shared `--bg` / `--surface` / `--fg` / `--muted` / `--meta` / `--border` / `--accent` / `--accent-on` / `--warn` / `--danger` / `--font-display` / `--elev-ring` / etc. vocabulary that the daemon's lint and prompt-injection consume.
- `design-systems/chala-ai/components.html` — reference fixture demonstrating worked button / card / form / hero-numeric / status-pill shapes wired through the tokens. First `:root` block byte-equivalent to `tokens.css` per the guard contract.
- `tokens-audit.md` — revised provenance section. `DESIGN.md` is the source of intent; `tokens.css` and the existing `tokens.json` are coexisting derived forms with distinct consumers (upstream HTML pipeline vs ODML multi-platform pipeline). Added an ODML → standard-schema name cross-walk table.

`pnpm guard` passes: chala-ai is now the 3rd structured brand (after `default` + `kami`), all A1/A2/B-slot required tokens declared, no unknown tokens, fixture sync OK.

Happy to split this into a follow-up PR if reviewers prefer to keep this one scoped to the originally-drafted seed; flag it and I'll move it. Otherwise leaving it here so chala-ai actually exercises the existing injection pipeline on the same merge.

Zacc-r:
**Review feedback addressed — `79504d2a`**

Thanks @mrcfps for the careful read on `skills/chala-ai-mobile/assets/od-elements.js`. All three issues fixed in `79504d2a` (the inline comments got auto-outdated by the line-number shift; replying to them via the API returned 404 so consolidating here):

**1. `tokenStyle('', name)` empty-prefix double-dash bug** — replaced `tokenStyle` with a `colorVar(token)` helper that validates against the full `OD_COLORS` set (matching `SKILL.md` and `packages/od-dialect/src/ast.ts`), returns `var(--od-${token})` for valid tokens, and returns `null` for anything else so callers decide the fallback explicitly. Both `od-text` and `od-icon` now route through it — `<od-text color="fg-muted">` previews with the correct token color.

**2. `background` only handled `bg-elevated` / `bg-overlay`** — both `od-screen` and `od-card` now route the `background` attribute through the same `colorVar` validator. Any valid ODColor resolves correctly; invalid tokens fall back (`var(--od-bg)` for od-screen, the default 4%-white card fill for od-card) instead of silently emitting `background: null` or `var(--od-bg-undefined)`. The fallback path is intentional — preview should differ visibly from intended rendering when the ODML attribute is off-spec, since that's exactly what the future validator session will reject.

**3. XSS via `shadow.innerHTML` interpolation of user strings** — every user-supplied string in the shim is now set via `textContent` or `setAttribute`, never concatenated into a template literal. Components affected and fix shape:

| Component | User-supplied string | New handling |
|---|---|---|
| `od-text` | content (textContent / bind) | `shadow.querySelector('span').textContent = text` |
| `od-icon` | `name` attribute (used as title) | `span.setAttribute('title', name)` |
| `od-badge` | `text` attribute | `span.textContent = text` |
| `od-input` | `placeholder` attribute | `input.setAttribute('placeholder', placeholder)` |
| `od-toggle` | `label` attribute | `labelEl.textContent = label` |
| `od-nav-bar` | `title` attribute | `span.textContent = title` |
| `od-tab-bar` | comma-split tab labels | rewritten with `createElement` + `textContent` per tab so each name lives in its own span |

Static CSS continues to use `innerHTML` since no user data interpolates there. Good catch — the file's throwaway preview code but it runs inside a script-enabled iframe, same standard applies as anywhere else.

While touching the file I also tightened two micro-hardening issues in the same bug family:
- `od-grid columns` is now clamped to `1..12` and integer-cast (was passing arbitrary attribute values into `repeat(N, ...)`).
- `od-progress value` is NaN-guarded and clamped at parse-time (was relying on the renderer to handle out-of-range values).

`pnpm guard` and `node --check` both clean. Happy to break any of the three into a separate commit if reviewers prefer atomic commits over a combined "review-feedback" commit.

design-systems/chala-ai/tokens.json
```diff
@@ -0,0 +1,130 @@
...
+    "fg-subtle": {
+      "$value": "#FFFFFF",
+      "$type": "color",
+      "$description": "Labels, metadata, timestamps (32% opacity on bg).",
+      "$extensions": {
+        "com.open-design.opacity": 0.32
+      }
+    },
+    "accent": {
+      "$value": "#FFFFFF",
```
mrcfps, line 45: `design-systems/chala-ai/tokens.json` is still drifting from the brand values exported by the other structured files. The concrete example here is `accent`: this file keeps it at `#FFFFFF`, while `design-systems/chala-ai/tokens.css:106`, `design-systems/chala-ai/components.html:39`, and the cross-walk in `design-systems/chala-ai/tokens-audit.md:30-35,46-55` all map the same token to `#E0E0E0`. That breaks the PR's stated goal of using ODML as the canonical cross-platform source of truth, because primary CTAs and any other `accent` usage will render one color in ODML/SwiftUI consumers and a different color in the existing HTML/daemon pipeline. Please pick one authoritative value for the ODML ↔ standard-schema mapping and update `tokens.json`, `od-elements.js`, and the audit text together so every shared token stays identical across the two structured contracts.

mrcfps (changes requested):
@Zacc-r Thanks for iterating on the preview fixes — I re-checked the current head and found one remaining blocker in the new token foundation. Right now the ODML token file no longer matches the values exposed by the standard design-system files, so the same Chala token name would render differently across ODML consumers versus the existing HTML/daemon pipeline.

PR #1737, opened 2026-05-14
feat(daemon): write-lock — serialize concurrent artifact writes

Zacc-r:
## Why

**Use case:** I've been hitting an intermittent race when two streaming agent turns emit the same artifact identifier in quick succession — two writers race on both the \`.html\` bytes and the \`.artifact.json\` sidecar, occasionally producing torn output (sidecar JSON references a content hash that doesn't match what landed on disk).

**Pain being addressed:** Concurrent writes to the same \`(project, file)\` aren't serialized today. The bug surface is small (you need two streams targeting the same canonical filename, which the current versioning scheme mostly avoids by suffixing), but the fix is a pure correctness improvement and is a prerequisite for any feature that writes-in-place (e.g. canonical-filename overwrite-in-place on regeneration).

This PR ships only the utility. A follow-up will wire it into \`projects.ts\` write paths and depends on this merging.

## What users will see

Nothing visible. Internal correctness fix — a utility module that will be consumed by upcoming work.

## Surface area

- [ ] **UI** — none
- [ ] **API / contract** — none
- [ ] **Extension point** — none
- [ ] **i18n keys** — none
- [ ] **New top-level dependency** — none
- [ ] **Default behavior change** — none (utility module; no consumers in this PR)
- [x] **None** — internal-only utility, no behavior change

## Screenshots

N/A.

## Bug fix verification

This is a preventative correctness utility rather than a fix for a tracked bug. The race it serializes is observable in principle but rare in practice (depends on identifier collision across concurrent streams). No red-then-green test in this PR; the follow-up that wires the lock into project writes will include the integration test.

## Validation

Ran from the repo root against the PR's HEAD commit:

- \`pnpm --filter @open-design/daemon typecheck\` → passes
- \`pnpm guard\` → passes
- Full test suite intentionally not run — this PR adds one standalone utility file with no consumers in the diff; nothing else is affected.

## Notes for reviewers

1. **Minimal scope is intentional.** The wired-up consumers (projects.ts write-paths, collapse-versions race window) live in a follow-up PR. Reviewing them bundled would have mixed pure-correctness with feature work; this split keeps each reviewable on its own merits.
2. **No exported types beyond \`withWriteLock\`.** The internal queue map is module-private; nothing leaks past the function signature.
3. **Per-key serialization, not global.** Two unrelated writes proceed in parallel; only same-key writes serialize. Key shape is consumer's choice (the planned follow-up uses \`\${projectId}:\${safeName}\`).

### DIFF
--- a/apps/daemon/src/write-lock.ts
+++ b/apps/daemon/src/write-lock.ts
@@ -16,6 +16,17 @@
 
 const writeLocks = new Map<string, Promise<unknown>>();
 
+/**
+ * Test-only accessor: exposes the active lock count so a unit test can
+ * assert the cleanup branch fires (the leak fixed in this module was
+ * exactly the kind of bug that's invisible in production but trivially
+ * detectable with `size === 0` after a settle). Do not consume from
+ * production code paths.
+ */
+export function __writeLocksSize(): number {
+  return writeLocks.size;
+}
+
 export async function withWriteLock<T>(
   key: string,
   fn: () => Promise<T>,
@@ -24,18 +35,21 @@ export async function withWriteLock<T>(
   const next = prev.then(fn, fn);
   // Stored as a settled promise so a rejected `next` doesn't block the
   // queue — the next caller chains off the settled state and runs.
-  writeLocks.set(
-    key,
-    next.then(
-      () => undefined,
-      () => undefined,
-    ),
+  // Capture the sentinel by reference: `next.then(...)` returns a NEW
+  // promise instance distinct from `next`, so comparing the map entry
+  // against `next` itself in the finally below was structurally always
+  // false and the cleanup branch never fired. Reusing this binding
+  // closes that leak.
+  const sentinel: Promise<void> = next.then(
+    () => undefined,
+    () => undefined,
   );
+  writeLocks.set(key, sentinel);
   try {
     return await next;
   } finally {
     // Best-effort cleanup: if no one chained on top of us, drop the
     // entry so the map doesn't grow unbounded over a long-lived daemon.
-    if (writeLocks.get(key) === next) writeLocks.delete(key);
+    if (writeLocks.get(key) === sentinel) writeLocks.delete(key);
   }
 }
--- a/apps/daemon/src/write-lock.ts
+++ b/apps/daemon/src/write-lock.ts
@@ -1,0 +1,41 @@
+// In-process serialization for concurrent writes to the same key.
+//
+// Two streams that finish "redesign sign-in page" at roughly the same
+// moment can both compute the same canonical filename and race
+// `writeFile` calls. Without serialization this is last-write-wins on
+// the inode (probably fine — `fs.promises.writeFile` is atomic at the
+// filesystem level on macOS/Linux for reasonable sizes), but it's also
+// last-write-wins on the manifest sidecar, where the order is
+// observable and confusing. Serializing per (projectId, fileName) makes
+// ordering deterministic.
+//
+// Limitation: in-process only. If multiple daemon instances ever share
+// the same .od/projects/, this guard does not extend across processes —
+// fall back to filesystem-level locking at that point. Today the
+// daemon is single-process per machine, so we're fine.
+
+const writeLocks = new Map<string, Promise<unknown>>();
+
+export async function withWriteLock<T>(
+  key: string,
+  fn: () => Promise<T>,
+): Promise<T> {
+  const prev = writeLocks.get(key) ?? Promise.resolve();
+  const next = prev.then(fn, fn);
+  // Stored as a settled promise so a rejected `next` doesn't block the
+  // queue — the next caller chains off the settled state and runs.
+  writeLocks.set(
+    key,
+    next.then(
+      () => undefined,
+      () => undefined,
+    ),
+  );
+  try {
+    return await next;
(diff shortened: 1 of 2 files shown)


apps/daemon/src/write-lock.ts
```diff
@@ -0,0 +1,41 @@
...
+      () => undefined,
+      () => undefined,
+    ),
+  );
+  try {
+    return await next;
+  } finally {
+    // Best-effort cleanup: if no one chained on top of us, drop the
+    // entry so the map doesn't grow unbounded over a long-lived daemon.
+    if (writeLocks.get(key) === next) writeLocks.delete(key);
```
PerishCode: **Cleanup branch never fires — `writeLocks.get(key) === next` is structurally always false.**

The map is populated with `next.then(() => undefined, () => undefined)` (lines 27–33), which is a *new* Promise instance distinct from `next`. The finally check on line 39 then compares the stored sentinel against `next` itself, so the identity check can never succeed and `writeLocks.delete(key)` is unreachable in practice.

**Why it matters.** The inline comment on lines 37–38 documents the intent: "drop the entry so the map doesn't grow unbounded over a long-lived daemon." As written, that invariant doesn't hold — every distinct key (planned shape: `${projectId}:${safeName}`) leaves a permanent settled-promise entry in `writeLocks`. For a long-lived daemon serving many projects/artifacts, the map monotonically grows with unique-key cardinality; it can't shrink even when the queue for a key has fully drained. The follow-up wiring this into `projects.ts` write-paths will inherit that leak.

**Evidence.** Lines 27–33 construct and store one promise; line 24 holds a different one (`next`); line 39 compares them. In single-threaded JS those references are independent objects, so `===` is always false here.

**Suggested change.** Capture the sentinel and compare against it (and reuse it as the stored value):

```ts
const sentinel: Promise<void> = next.then(
  () => undefined,
  () => undefined,
);
writeLocks.set(key, sentinel);
try {
  return await next;
} finally {
  if (writeLocks.get(key) === sentinel) writeLocks.delete(key);
}
```

A tiny unit test would have caught this and would be cheap to add here rather than waiting for the consumer PR — e.g. run one `withWriteLock('k', async () => {})`, await it, then assert the module-private map is empty (either via a test-only `__size` accessor or by exporting a `clear`/`size` helper guarded for tests). Up to you whether to land the test in this PR or fold it into the follow-up, but the one-line cleanup fix is worth doing in this PR so the documented invariant is true on day one.


PerishCode (commented):
@Zacc-r — overall the `withWriteLock` shape (per-key serialization, swallowing rejections on the chained sentinel so a failure doesn't poison the queue, await on the live `next` so callers still see the error) is exactly right for the documented use case. One concrete issue with the best-effort cleanup branch left inline; everything else looks good.



apps/daemon/src/write-lock.ts
Zacc-r: Good catch, fixed in `842a8899`. Captured the sentinel into a local binding so the `===` check in `finally` compares against the same instance that was stored in the map, instead of against `next` (which was always a different chained promise). Cleanup branch now reachable.

Also took your suggestion to land a unit test in this PR rather than waiting for the consumer. Added `apps/daemon/tests/write-lock.test.ts` with six cases — serialization order, value pass-through, rejection propagation, parallel-on-different-keys, and the two cleanup invariants (success path and rejection path). Both cleanup tests assert `__writeLocksSize()` returns to baseline after `await` + a couple microtasks, so if a future change re-introduces the same identity-mismatch the test goes red immediately.

The test seam is exported as `__writeLocksSize` (underscore prefix + comment marking it test-only) to keep production callers from accidentally consuming it; happy to swap to a `clear()` / `size()` pair guarded behind `NODE_ENV` instead if that fits your conventions better.

lefarcen (commented):
Hey @qiongyu1999 — quick review ping on this one when you have a chance. PR #1737 is a daemon write-lock utility; CI is green, and @Zacc-r already addressed the stale cleanup-thread finding in `842a8899` with tests. The remaining blocker appears to be product review on the feature-labeled PR.

PerishCode (approved):
@Zacc-r — the sentinel capture is exactly the right fix. `next.then(...)` always returned a fresh promise instance, so the identity check in `finally` could never match `next` itself; storing the chained-and-swallowed promise into a `sentinel` binding and comparing the map entry against that same reference makes the cleanup branch reachable and the documented "map doesn't grow unbounded" invariant actually true.

The new `tests/write-lock.test.ts` is also well-shaped for the change: the two `drains-*` cases assert `__writeLocksSize()` returns to baseline after a settle on both the resolved and rejected paths, so a future regression that re-breaks the identity comparison flips the suite red immediately. The serialization-order / value pass-through / rejection-continuation / parallel-keys cases round out the contract nicely, and the `__`-prefixed + docblocked test seam keeps the production surface clean. Nice turnaround on the follow-up — appreciate the per-test unique-key choice too, that keeps the cleanup assertions robust against test ordering.



lefarcen (commented):
Hey @mrcfps — quick reviewer ping here. @PerishCode has approved the current head (`842a8899`), @qiongyu1999 has product-approved it, and CI is green. From the current state, this looks like it is waiting on your requested review / merge-blocker check rather than more author changes.

PR #1703, opened 2026-05-14
fix: Restore dropdown chevron in page-size selector for dark mode

xxiaoxiong:
## Summary

Fixes the page-size selector rendering issue in dark mode where the dropdown chevron icon was missing.

## Why

The `.df-pagination select` CSS rule was using `background: var(--bg)` which is a shorthand property that resets all background-related properties, including `background-image`. This caused the dropdown chevron icon (defined in the global `select` styles) to disappear in dark mode.

## What users will see

- ✅ Dropdown chevron icon now visible in dark mode
- ✅ Proper contrast and readability
- ✅ Consistent styling with other select elements

## Surface area

- **Changed**: `apps/web/src/index.css` - `.df-pagination select` background property
- **Impact**: All pagination controls with page-size selectors

## Validation

1. Enable dark mode
2. Open any file list view with pagination controls
3. Verify the page-size selector (e.g., "30" dropdown) displays the chevron icon
4. Verify the icon has proper contrast against the dark background
5. Test both explicit `[data-theme="dark"]` and system `prefers-color-scheme: dark`

## Technical details

Changed from:
```css
background: var(--bg);
```

To:
```css
background-color: var(--bg);
```

This preserves the global `select` background-image (chevron icon) while still setting the background color.

Fixes #1697

PR #1739, opened 2026-05-14
Fix Design Files dark-mode page-size selector

prantikmedhi:
Fixes #1697

## Why

Design Files' page-size select in dark mode lost its chevron styling because the control was getting a shorthand background override. That made the pagination control look broken and harder to scan.

## What users will see

The page-size selector keeps its chevron in dark mode and still matches the shared select styling.

## Surface area

- [x] **UI**

## Validation

- `pnpm --filter @open-design/web test -- tests/components/DesignFilesPanel.test.tsx`
- `pnpm guard`

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -8371,10 +8371,10 @@ button.connector-action.is-loading {
 }
 .df-pagination select {
   font-size: 12px;
-  padding: 2px 6px;
+  padding: 2px 24px 2px 6px;
   border: 1px solid var(--border);
   border-radius: 4px;
-  background: var(--bg);
+  background-color: var(--bg-panel);
   color: var(--text);
   cursor: pointer;
 }


lefarcen (commented):
Hey @prantikmedhi! 👋

The root cause is spot-on — `background` shorthand resetting `background-image` is a well-known CSS footgun, and your fix goes a step further than the minimal rename: switching to `--bg-panel` and adding explicit right padding for the chevron space is a thoughtful touch.

Before review picks this up, a quick template ask — a few sections are missing that help reviewers scope quickly:

- **Why**: the user-visible symptom (dark mode → chevron disappears from the Design Files pagination dropdown) and what caused it
- **What users will see**: e.g. "page-size selector chevron reappears in dark mode in the Design Files panel"
- **Surface area**: please tick **UI** in the checklist — this changes a visible control
- **Validation**: your `pnpm test` run covers the component, but a one-liner manual step (enable dark mode → open a file list with pagination → confirm chevron shows) would make spot-check at merge time easy

Once those are in, review can move fast given the tiny diff. ✨

---

**Related**: #1703 by @xxiaoxiong is also open targeting the same `.df-pagination select` background bug — both touch `apps/web/src/index.css` for the same dark-mode chevron issue (#1697). The approaches differ slightly (this PR uses `--bg-panel` + padding adjustment; #1703 does the minimal `background` → `background-color` rename). The maintainer team will pick which one lands — sharing so neither effort goes unnoticed.

xxiaoxiong:
Thanks for working on this! This issue was already fixed in PR #1703 (submitted earlier today). You might want to close this PR to avoid duplication.

prantikmedhi:
Thanks — noted. I see #1703 already covers the same bug, so I’ll leave this one for maintainers to resolve.

PR #1746, opened 2026-05-14
feat(creative-memory-system): add validated subsystem prototype

itsmeved24:
## Why

**Use case.** I've been building a persistent creative-preference memory subsystem for Open Design — something that remembers, across sessions and projects, what styles a user consistently accepts, refines, or rejects, and folds that back into the next generation prompt as a small `[MEMORY CONTEXT]` block. This PR is the integration-staged drop of that subsystem.

**Pain addressed.** Today, two consecutive generations don't know about each other. A user can spend a session refining toward "airy spacing, serif headlines" and the next session starts cold. The subsystem captures that signal locally, deterministically, and bounded — without embeddings, vector DBs, ML rankers, or cloud sync — and gives the host pipeline one function to call at generation time.

This is an RFC-shape PR. The subsystem is heavily simulated as a standalone module (238/238 assertions across four suites), but the pipeline hookpoints in `extractionAdapter.js` are stubs awaiting the pipeline team's confirmation on event timing and shape. I'm raising it now so reviewers can shape the integration surface and placement before I do the TypeScript port and wire-up. The remaining work is mine; the rest of the integration will be handled by the team as documented in `docs/open-questions.md`.

## What users will see

Nothing yet, by design. This PR adds an internal subsystem with no caller. Users will see no behavior change until follow-up work wires the adapter handlers into real generation events and concatenates the prompt block into outgoing prompts. The intent is to merge the subsystem and its simulations as a reviewable artifact first, then land the wire-up in a focused follow-up PR.

## Surface area

- [ ] **UI** — none
- [ ] **CLI / env var** — `MEMORY_STORAGE_ROOT` env var (optional override of the subsystem's local JSON storage root). No `od` subcommand or `tools-*` flag changes.
- [ ] **API / contract** — none. No `/api/*` endpoint, no SSE event, no `packages/contracts` change.
- [ ] **Extension point** — none.
- [ ] **i18n keys** — none.
- [ ] **New top-level dependency** — none. Subsystem is dependency-free (Node `fs`/`path` only).
- [ ] **Default behavior change** — none. Subsystem is dormant until a host wires it up.
- [ ] **None** — internal subsystem + docs only.

## Screenshots

N/A — no UI surface.

## Validation

What I ran from the subsystem directory:

```
node creative-memory-system/sims/testHarness.js          # 64 / 64 ✓
node creative-memory-system/sims/retrievalQualitySim.js  # 36 / 36 ✓
node creative-memory-system/sims/retrievalAdvancedSim.js # 91 / 91 ✓
node creative-memory-system/sims/retrievalLongRunSim.js  # 47 / 47 ✓
node creative-memory-system/extractionAdapter.js         # 12 / 12 ✓
```

Total: **238/238 engine assertions** + **12/12 adapter assertions** passing.

The simulations cover: lifecycle (CRUD, ingestion accumulation, confidence ladder), reversal logic (noise guard → graduated reduction → under_review + shadow records → shadow promotion + original archived), decay (90-day strength × 0.70, 180-day archive), retrieval quality (ranking monotonicity, hysteresis stability, flicker detection, prompt pollution, threshold gating), constraint interaction (token budget, negative priority multiplier, project-override conflict diagnostics, polarity diversity ceiling + backfill, category diversity ceiling + backfill, MIN_NEG_FLOOR), and long-running temporal scenarios (20+ session evolution, 50-pattern stress, cross-project isolation, polarity entropy, category dominance over 25–30 sessions).

## Reviewer notes

A few deliberate decisions reviewers should be aware of upfront. All are documented in `creative-memory-system/docs/open-questions.md`:

1. **Placement.** The subsystem sits at `<repo-root>/creative-memory-system/` rather than under `packages/` or `apps/daemon/src/`. It's a standalone module with its own simulation harness; once the pipeline contracts are confirmed and a TypeScript port lands, migration to `packages/creative-memory-system/` is the natural next step. Open question #8.
2. **CommonJS / `.js`.** The validated baseline is CommonJS. The repo is TypeScript-first per `AGENTS.md`. A local `creative-memory-system/package.json` declares `"type": "commonjs"` so the module ignores the root's `"type": "module"`. The TypeScript port is the planned follow-up; doing it before the pipeline contracts are confirmed risks introducing bugs the simulations wouldn't catch. Open question #9.
3. **`extractionAdapter.js` handlers are stubs.** The handler signatures and signal mapping are fully specified, but the pipeline hookpoints (where these are called from inside the generation pipeline) are TBD pending pipeline-team confirmation. Open questions #1 (event timing), #2 (refinement diff shape), #3 (debounce/coalesce), #4 (attachment point).
4. **No new top-level dependencies.** The subsystem uses only Node `fs` and `path`.
5. **Storage location.** Default is `<module dir>/memory/<userId>/preferences.json`, overridable via `MEMORY_STORAGE_ROOT`. Per-user JSON, schema-versioned, human-readable. Gitignored.

## Where to start reading

1. `creative-memory-system/README.md` — overview, philosophy, integration surface.
2. `creative-memory-system/docs/architecture.md` — preference lifecycle, reversal handling, shadow records, balancing philosophy.
3. `creative-memory-system/docs/retrieval-pipeline.md` — the 11 retrieval stages, in order, with rationale for ordering.
4. `creative-memory-system/docs/diagnostics.md` — every diagnostic event the engine can emit.
5. `creative-memory-system/docs/validation.md` — what each simulation suite proves.
6. `creative-memory-system/docs/open-questions.md` — the questions I'd most like reviewer input on.

The `_PROJECT_context.md` is the original working-context document carried over from the prototype phase; it overlaps with the curated docs and is preserved as historical reference.


lefarcen (commented):
Hey @itsmeved24! 👋 The RFC-shape framing here is really thoughtful — the 238/238 simulation coverage across lifecycle, reversal logic, decay, and long-running temporal scenarios is an unusually solid foundation for a subsystem prototype, and the curated doc set (`architecture.md`, `retrieval-pipeline.md`, `open-questions.md`) makes the design legible at multiple depths. One small thing before pool review digs in:

**Surface area checkboxes:** the `CLI / env var` row mentions `MEMORY_STORAGE_ROOT` but that box is left unchecked, while `[x] None` is checked — those two are in slight tension. If the env var is intentionally in-scope for this PR (even as a dormant default), the `CLI / env var` box should be checked so future release-verification passes can scope it correctly. If it's truly a no-op until the subsystem is wired and you consider it out of scope for this PR, a short note clarifying that would resolve the ambiguity.

Assigned @nettee for pool review. ✨

itsmeved24:
Converting to draft. CI is failing on `pnpm guard` exactly as `AGENTS.md` warns:

> Residual project-owned JavaScript files found: …  Convert these files to TypeScript or add a documented generated/vendor/output allowlist entry.

This is expected — the subsystem was validated as CommonJS `.js` (238/238 simulation assertions) and I dropped the validated baseline in unchanged for reviewability. The TypeScript port is the natural next step; logic doesn't change, the four simulation suites become a Vitest suite under the package's `tests/` directory, and the residual-JS guard goes green.

I'll force-push the TS port to this branch and flip back to ready-for-review once `pnpm guard` and `pnpm typecheck` pass locally. The structural decisions (placement, adapter stubbing, etc.) are tracked in `creative-memory-system/docs/open-questions.md` and would still benefit from maintainer input before I commit to a final shape — happy to take feedback in the meantime.

itsmeved24:
TypeScript port pushed. Subsystem now lives at `packages/creative-memory-system/` as `@open-design/creative-memory-system`, matching the workspace conventions used by `@open-design/contracts` / `@open-design/platform`.

**What's preserved.** All logic, all constants, all balancing semantics, all diagnostic events, all 238 invariants from the original simulation runners. The Vitest suites under `tests/` reproduce the same scenarios as the original `sims/*.js` files (lifecycle, retrievalQuality, retrievalAdvanced, retrievalLongRun, extractionAdapter).

**What changed structurally.** Module exports are named ESM. Per-handler boilerplate in `extractionAdapter.ts` is collapsed into a single `dispatchSignals` helper without behavioral change. `STORAGE_ROOT` resolution is now lazy via `getStorageRoot()` so tests can override `MEMORY_STORAGE_ROOT` after the engine is imported (ESM hoisting would otherwise freeze the wrong path).

**One repo-level change.** `scripts/guard.ts` adds `packages/creative-memory-system/esbuild.config.mjs` to the residual-JS allowlist alongside the existing entries for contracts/platform/sidecar/sidecar-proto. Same justification: esbuild config entrypoints execute directly by Node before package dist output exists.

**Validation.**

```
pnpm guard                                                # passes
pnpm typecheck                                            # 0 errors
pnpm --filter @open-design/creative-memory-system test    # 81/81
pnpm --filter @open-design/creative-memory-system build   # passes
```

CI is now all green (6/6). Flipping back to ready-for-review.

The structural decisions and the `extractionAdapter` stub hookpoints are still tracked in `packages/creative-memory-system/docs/open-questions.md` and would benefit from maintainer input — happy to take feedback there.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,958 @@
...
+        if (!remainingCats[t]) underRepTypes.add(t);
+      }
+
+      if (overflowPrefs.length > 0 && underRepTypes.size > 0) {
+        const catBackfill = overflowPrefs
+          .filter((p) => underRepTypes.has(p.preference_type))
+          .slice(0, catTrimmed.length);
+        if (catBackfill.length > 0) {
+          prefs.push(...catBackfill);
+          prefs.sort((a, b) => (b._effective_priority ?? 0) - (a._effective_priority ?? 0));
```
nettee: Stage 10 can re-expand a category past `MAX_PER_CATEGORY`. `underRepTypes` is computed once from `remainingCats`, but the `overflowPrefs.filter(...).slice(...)` backfill never tracks how many slots each type still has left, so multiple overflow entries from the same type can all be re-added here. That means a type that was just trimmed to 3 can end up back at 4+ after backfill. The current `ADV-23` coverage only checks that minority types are present, so this regression still passes. Please backfill against per-type remaining capacity instead of set membership alone, and add a regression that asserts every returned type count stays `<= MAX_PER_CATEGORY` after category backfill.

packages/creative-memory-system/src/extractionAdapter.ts
```diff
@@ -0,0 +1,201 @@
...
+ * accepting a layout also signals typography if both changed.
+ *
+ * STUB: In production, artifactMeta is populated by the generation pipeline.
+ * The classifier here is a pass-through. The actual extraction of (type,
+ * pattern) from rendered artifacts is upstream of this module. Doing it here
+ * would violate the local-first / no-ML constraint (open question #4).
+ */
+export function classifyArtifact(artifactMeta: ArtifactMeta | null | undefined): ArtifactSignal[] {
+  if (!artifactMeta || !artifactMeta.signals) return [];
+  return artifactMeta.signals;
```
nettee: `classifyArtifact()` currently turns missing `artifact_meta` / `signals` into `[]`, which means a broken pipeline payload silently disables learning for that event. In this repo we usually fail fast on missing required integration input, and `BaseEvent.artifact_meta` is typed as required, so this fallback hides exactly the wiring bug the adapter should surface. The new test in `packages/creative-memory-system/tests/extractionAdapter.test.ts` also blesses that hidden-failure path. Please throw a clear error here (and update the test) so a missing classifier payload is observable instead of being treated as a no-op.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,958 @@
...
+    .filter((p): p is Preference => p.polarity === "positive")
+    .sort((a, b) => b.signal_strength - a.signal_strength);
+
+  const negatives = budgeted
+    .filter((p): p is Preference => p.polarity === "negative")
+    .sort((a, b) => b.signal_strength - a.signal_strength);
+
+  const overrides = projectPrefs.filter(
+    (p) => p.signal_strength >= INJECTION_THRESHOLD && p.polarity_status === "stable",
+  );
```
nettee: `projectOverrides` ignores `context.preference_types`, even though Stage 3 applies that filter to the main injection set. When a caller retrieves a type-scoped subset and then calls `buildPromptBlock()`, the `Project override:` line can still include unrelated categories from the same project because this branch only checks strength/status. That leaks filtered-out context back into the prompt. Please apply the same `preference_types` predicate to `projectPrefs` before returning `projectOverrides`, and add a regression that exercises `retrieveForInjection(..., { project_id, preference_types })` plus `buildPromptBlock()`.

nettee (changes requested):
I found two blocking issues in the retrieval/adapter paths that should be fixed before this package is merged; I also left one scoped follow-up on the filtered project-override output.


itsmeved24:
Thanks for the careful read. All three are fixed in `c2b186eb`.

**1. Stage 10 — category backfill re-expansion past `MAX_PER_CATEGORY`.** Confirmed reproduces under "one trimmed type + one large overflow pool of another type". The previous flat `underRepTypes` set membership check let multiple overflow entries from the same type re-enter. Replaced with a per-type remaining-capacity counter (`MAX_PER_CATEGORY - current count`) that decrements as records are admitted, walking overflow in priority order. New regression test `ADV-27` asserts every type's count stays `<= MAX_PER_CATEGORY` in the final injection set under the bug-triggering conditions; without the fix it fails as expected.

**2. `classifyArtifact` silently disabling learning on broken payloads.** Agreed — the fallback contradicted `BaseEvent`'s typed contract and hid exactly the wiring bug the adapter should surface. Now throws clear errors on `null`/`undefined` payload and on a non-array `signals` field. The adapter test that previously blessed the no-op path is replaced with a `toThrow` assertion across all four malformed-payload shapes (`null`, `undefined`, `signals: undefined`, `signals: "not-an-array"`).

**3. `projectOverrides` ignoring `context.preference_types`.** Good catch — Stage 3 applied the predicate but the project-override branch at the bottom of `retrieveForInjection` did not, so the prompt block's `Project override:` line could leak unrelated categories. Now applies the same `preference_type.startsWith(t)` predicate before returning. New regression test `ADV-28` exercises `retrieveForInjection({ project_id, preference_types: ["layout"] })` against a project that overrides both a `layout` and a `color` pattern, then verifies the prompt block contains the in-scope pattern and does not contain the out-of-scope one.

Validation:

```
pnpm --filter @open-design/creative-memory-system test       # 83/83 (12 + 18 + 12 + 15 + 26)
pnpm --filter @open-design/creative-memory-system typecheck  # 0 errors
pnpm guard                                                    # passes
```

Worth flagging: bugs 1 and 3 existed in the validated `.js` baseline too. The original simulation suites covered "trimming fires" and "diversity ceiling fires" at the diagnostic level but didn't pressure-test the post-backfill invariants or the project-override + type-filter intersection. The Vitest port now closes both gaps.

Re-requesting review.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,978 @@
...
+
+      if (daysSinceDecay >= DECAY_DAYS) {
+        // 180+ days since last signal — archive
+        pref.polarity_status = "archived";
+        archived++;
+      } else if (daysSinceDecay >= 0) {
+        // 90–180 days — apply decay
+        pref.signal_strength = clamp(pref.signal_strength * 0.70, 0, 1);
+        pref.confidence = dropConfidence(pref.confidence);
+        pref.decay_at = addDays(n, DECAY_DAYS);
```
nettee: `runDecay()` currently measures both the 90-day decay threshold and the 180-day archive threshold from `decay_at`, but the decay branch also resets `decay_at = now + DECAY_DAYS`. That means a routine decay pass at ~day 90 pushes the archive decision out by another 90 days, so a preference can survive until roughly day 270 instead of archiving at 180 days since `last_seen` as the docs and comments describe. The existing tests only backdate `decay_at` directly into the archive band, so they miss the real scheduled-flow regression. Please base archival on `last_seen` age (or store a separate fixed archive deadline that is not rewritten by the decay branch), and add a regression that runs decay once around day 90 and again around day 180.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,978 @@
...
+
+    if (!pref.sources.includes(signal.signal_type)) pref.sources.push(signal.signal_type);
+    if (signal.tag_text && !pref.explicit_tags.includes(signal.tag_text)) {
+      pref.explicit_tags.push(signal.tag_text);
+    }
+    if (signal.polarity === "positive") pref.accept_count++;
+    else pref.reject_count++;
+
+    pref.last_seen = t;
+    pref.decay_at = addDays(t, DECAY_DAYS);
```
nettee, line 569: `findExistingPref()` here will happily reuse an archived record, and the matching-polarity update path below only increases strength / confidence and refreshes timestamps. It never clears `polarity_status`, so once a preference is archived, later same-polarity signals keep mutating an archived record that retrieval will continue to filter out forever. In practice that means a user cannot re-establish a previously archived preference without manually deleting the old record. Please either ignore archived matches and create a fresh active record, or explicitly unarchive/reset the record when reinforcing same-polarity input arrives, and cover it with a regression for `archive -> new same-polarity signal -> injectable again`.

packages/creative-memory-system/src/extractionAdapter.ts
```diff
@@ -0,0 +1,219 @@
...
+      pattern,
+      preference_type,
+      polarity,
+      tag_text,
+      scope,
+      project_id: event.project_id || null,
+      artifact_id: event.artifact_id,
+      session_id: event.session_id,
+      timestamp: event.timestamp,
+    });
```
nettee, line 194: The new fail-fast check only verifies that `artifact_meta.signals` is an array; it still trusts every array member here. A payload like `{ signals: [{}] }` now passes `classifyArtifact()`, then this loop forwards `undefined` `pattern` / `preference_type` into `ingestSignal()`, which persists a corrupt preference record instead of surfacing the malformed pipeline payload. That still violates the repo's fast-fail contract and turns an integration bug into bad stored state. Please validate each signal entry before dispatching (for example, require non-empty string `pattern` and `preference_type` in `classifyArtifact()`), and add a regression that asserts malformed array members throw instead of creating records.

nettee (changes requested):
I found three blocking correctness issues that still affect decay lifecycle, archived-record recovery, and adapter payload validation. Details are inline.


itsmeved24:
All three fixed in `e9bab788`.

**1. Decay lifecycle — archive based on `last_seen` age, not `decay_at`.**

Confirmed: the decay branch resetting `decay_at = now + DECAY_DAYS` pushed the archive decision forward indefinitely (~270 days instead of the documented 180). The fix separates the two checks:

- **Archive:** `daysBetween(pref.last_seen, now) >= ARCHIVE_DAYS` — fixed deadline from last user interaction, never rewritten by decay passes.
- **Decay:** `daysBetween(pref.decay_at, now) >= 0` — fires when the preference has been untouched for DECAY_DAYS since its last signal or last decay pass; resets `decay_at` forward for the next cycle.

New regression: "archives at 180 days since last_seen even after intermediate decay passes" — runs decay at day ~91 (fires, resets `decay_at`), then again at day ~181 (archives based on `last_seen` age, not the reset `decay_at`).

**2. Archived-record recovery on same-polarity signal.**

Confirmed: `findExistingPref()` matched archived records, and the accumulation path only refreshed timestamps without clearing `polarity_status`. A user could never re-establish a previously archived preference.

The fix detects an archived match at the top of the accumulation path and resets `polarity_status` to `"stable"` + clears `reversal_signals` / `reversal_first_seen` before accumulating. This lets the record re-enter the injectable set naturally as strength builds.

New regression: "archived record is reactivated by new same-polarity signal" — archives a record via `last_seen` backdating, then sends a new same-polarity signal and asserts the record is `stable` and appears in `retrieveForInjection`.

**3. Signal entry validation in `classifyArtifact`.**

Confirmed: `{ signals: [{}] }` passed the array check and forwarded `undefined` pattern/preference_type into `ingestSignal`, persisting corrupt records. The fix validates each entry: both `pattern` and `preference_type` must be non-empty strings. Throws with the index of the malformed entry so the pipeline team can trace which upstream classifier produced the bad payload.

New regression: "classifyArtifact throws on malformed signal entries" — asserts empty-string, missing-field, and mixed-valid/invalid arrays all throw with the correct index.

**Validation:**

```
pnpm --filter @open-design/creative-memory-system test       # 86/86
pnpm --filter @open-design/creative-memory-system typecheck  # 0 errors
pnpm guard                                                    # passes
```

All three bugs existed in the original validated `.js` baseline. The simulation suites tested decay and ingestion at the "does it fire" level but not at the "does the scheduled flow produce the documented lifecycle" level. The Vitest port now has post-condition invariants for all three paths.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1003 @@
...
+  const fp = userFilePath(userId);
+  if (!fs.existsSync(fp)) return null;
+  return JSON.parse(fs.readFileSync(fp, "utf8")) as PreferenceStoreFile;
+}
+
+function saveFile(userId: string, data: PreferenceStoreFile): void {
+  const fp = userFilePath(userId);
+  fs.mkdirSync(path.dirname(fp), { recursive: true });
+  data.last_updated = now();
+  fs.writeFileSync(fp, JSON.stringify(data, null, 2), "utf8");
```
nettee: `userId` is interpolated into the on-disk path without any containment check. Right now `resetMemory('../escape')` creates a `preferences.json` file outside `MEMORY_STORAGE_ROOT`, because `path.join(resolveStorageRoot(), userId, "preferences.json")` accepts `..` segments and the same helper is used for both reads and writes. That turns any bad or untrusted caller input into arbitrary file access under the process's permissions. Please either reject user IDs that are not a safe basename (for example `/^[A-Za-z0-9_-]+$/`) or resolve the final path and assert it still stays under the resolved storage root before calling `existsSync`, `mkdirSync`, or `writeFileSync`, then add a regression for `../` input.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1003 @@
...
+          override_polarity: override ? override.polarity : "unknown",
+          override_strength: override ? override.signal_strength : 0,
+          project_id,
+          trace: `${project_id} override ${gp.pattern} active — global ${gp.pattern} (${gp.polarity}, ${gp.signal_strength.toFixed(2)}) suppressed for this generation`,
+        });
+      }
+    }
+
+    prefs = prefs.filter((p) => !projectPatterns.has(p.pattern));
+    prefs = [...prefs, ...projectPrefs];
```
nettee: This override merge keys suppression only by `pattern`, even though the store identity everywhere else is `preference_type + pattern + polarity`. If a project override defines `pattern: "shared"` for one category and the global set contains `pattern: "shared"` for another, line 684 drops the unrelated global record too. I reproduced this on the built package with a global positive `layout_density/shared` and a project negative `color/shared`: retrieval for that project returns only the override and suppresses the global layout preference. Please key `projectPatterns`, the diagnostic lookup, and the final filter on at least `preference_type + pattern` (and keep polarity handling explicit), then add a regression with the same pattern text in two categories.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1003 @@
...
+        const slotBudget = catTrimmed.length;
+        for (const p of overflowPrefs) {
+          if (catBackfill.length >= slotBudget) break;
+          const cap = remainingCapacity[p.preference_type] ?? 0;
+          if (cap <= 0) continue;
+          catBackfill.push(p);
+          remainingCapacity[p.preference_type] = cap - 1;
+        }
+        if (catBackfill.length > 0) {
+          prefs.push(...catBackfill);
```
nettee, line 1048: Stage 8 promotes entries out of `positiveOverflow`, but Stage 10 later iterates the original `overflowPrefs` again without removing the ids that were already admitted here. That lets the same preference be injected twice when both backfill stages run in one retrieval. I reproduced it against the built package with 11 negatives, 4 `A` positives, 4 `B` positives, and 2 `C` positives: the final result contains `C:c_1` twice and both `diversity_backfill` and `category_backfill` diagnostics fire. Please remove Stage-8-admitted ids from the overflow pool before Stage 10 (or skip any overflow entry already present in `prefs`) and add a regression that asserts final injected ids are unique when polarity backfill and category backfill both execute together.

nettee (changes requested):
I found three blocking issues that still affect storage safety and retrieval correctness on the current head. I also re-ran `pnpm --filter @open-design/creative-memory-system test` and `pnpm --filter @open-design/creative-memory-system build`; both pass locally, but the cases below still need fixes before merge.


itsmeved24:
All three fixed in `d20c91c8`.

**1. `userId` path traversal.**

`userFilePath()` now rejects IDs that are not safe basenames (`/^[A-Za-z0-9_-]+$/`). Previously `path.join(storageRoot, userId, "preferences.json")` accepted `..` segments and could write outside the storage root. New regression `ADV-29` asserts `../escape`, `../../etc`, and `foo/bar` all throw with a clear error.

**2. Project-override merge keyed on `preference_type + pattern`.**

Confirmed: the previous `Set` was keyed on `pattern` alone, so a project override for `color/shared` would suppress an unrelated global `layout_density/shared`. The fix uses a composite key (`preference_type + NUL + pattern`) for the Set, the diagnostic lookup, and the filter. New regression `ADV-30` builds a global `layout_density/shared` and a project `color/shared`, then asserts both survive in the injection set for that project.

**3. Overflow pool deduplication between Stage 8 and Stage 10.**

Confirmed: Stage 8 promoted entries from `positiveOverflow` (a subset of `overflowPrefs`) into `prefs`, but Stage 10 iterated the original `overflowPrefs` without excluding already-admitted IDs. The fix removes Stage-8-admitted IDs from `overflowPrefs` immediately after the polarity backfill, before Stage 10 runs. New regression `ADV-31` builds a profile that triggers both polarity and category backfill in one retrieval (11 negatives + 4 type-A + 4 type-B + 2 type-C positives) and asserts all IDs in the final injection set are unique.

**Validation:**

```
pnpm --filter @open-design/creative-memory-system test       # 89/89
pnpm --filter @open-design/creative-memory-system typecheck  # 0 errors
pnpm guard                                                    # passes
```

All three bugs existed in the original validated baseline. The review is surfacing real correctness issues that the original simulation suites never pressure-tested — the Vitest port now has post-condition invariants for each.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1031 @@
...
+}
+
+function getPrefArray(
+  data: PreferenceStoreFile,
+  scope: "global" | "project",
+  projectId: string | null,
+): Preference[] {
+  if (scope === "global") return data.global_preferences;
+  if (scope === "project" && projectId) {
+    if (!data.project_overrides[projectId]) data.project_overrides[projectId] = [];
```
nettee: `project_id` is still used as a raw key on a plain `{}` here. Meta-keys like `__proto__` and `constructor` do not create a normal bucket; `data.project_overrides[projectId] = []` mutates object semantics instead, so a project-scoped event can corrupt override storage or make that project's prefs unretrievable later. Because `project_id` comes from runtime event payloads and retrieval context and is unvalidated in this PR, this is a real correctness/safety bug in the changed range. Please either reject unsafe project IDs (similar to the new `userId` validation) or store `project_overrides` in `Object.create(null)`/`Map` consistently across init/read/reset/retrieval, then add a regression for `"__proto__"`.

packages/creative-memory-system/src/extractionAdapter.ts
```diff
@@ -0,0 +1,238 @@
...
+  }
+}
+
+/**
+ * Fired when the user applies an inline tag during or after generation.
+ * Tags are the strongest signal type — explicit user intent.
+ */
+export function onExplicitTagApplied(event: ExplicitTagAppliedEvent): void {
+  const tagLower = event.tag_text.toLowerCase().trim();
+  const polarity: Polarity = TAG_POLARITY[tagLower] ?? "positive";
```
nettee: Unknown explicit tags default to `"positive"` here. Since `explicit_tag` is the strongest signal weight in this package, a new negative phrase like `too busy` or `don't use this` will be learned as a positive preference instead of surfacing the missing contract. That silently inverts user intent and pollutes future prompt injection. Please fail fast on unknown tags (or require the caller to provide polarity explicitly) and add a regression asserting that an unrecognized tag cannot create a positive record.

packages/creative-memory-system/src/extractionAdapter.ts
```diff
@@ -0,0 +1,238 @@
...
+  dispatchSignals(event, "explicit_tag", polarity, event.tag_text);
+}
+
+/**
+ * Fired when the user gives a thumbs up or thumbs down on a generation.
+ * Weaker than explicit tags — captures sentiment without specificity.
+ */
+export function onThumbsRated(event: ThumbsRatedEvent): void {
+  const signal_type: SignalType = event.rating === "up" ? "thumbs_up" : "thumbs_down";
```
nettee, line 263: Anything other than exact `"up"` is treated as `thumbs_down` here. The TypeScript union only protects compile-time callers; at runtime a malformed payload such as `"UP"`, `null`, or a renamed enum value will silently create negative memory instead of surfacing the wiring bug. That contradicts the fail-fast validation you already added for `artifact_meta`. Please validate `event.rating` at runtime and throw unless it is exactly `"up"` or `"down"`, with a regression covering an invalid value.

nettee (changes requested):
I found three blocking correctness issues in the project-override storage path and adapter runtime validation. Details are inline.


itsmeved24:
All three fixed in `00c821a6`, plus a comprehensive input-validation sweep to close the class of issues rather than playing whack-a-mole.

**Reviewer-requested fixes:**

**1. `project_id` prototype pollution.** `getPrefArray`, `retrieveForInjection`, `listPreferences`, and `resetMemory` now reject `__proto__`, `constructor`, and `prototype` via a shared `validateProjectId()` helper that checks both an explicit `UNSAFE_OBJECT_KEYS` set and the safe-basename regex. Regression: adapter test asserts `__proto__` and `constructor` both throw.

**2. Unknown explicit tags.** `onExplicitTagApplied` now throws on tags not in `TAG_POLARITY` instead of defaulting to positive. Prevents silent intent inversion for negative phrases not yet registered. Regression asserts unrecognized tags throw and create no records.

**3. Thumbs rating.** `onThumbsRated` now validates `event.rating` is exactly `"up"` or `"down"` at runtime. Regression covers `"UP"`, `null`, and empty string.

**Proactive hardening (comprehensive sweep):**

Rather than wait for the next round, I audited every public function and every external input path. Here's what was added beyond the three requested fixes:

| Input | Validation added | Rejects |
|---|---|---|
| `signal.polarity` | `ingestSignal()` | Anything other than `"positive"` / `"negative"` |
| `signal.pattern` | `ingestSignal()` | Empty string, non-string |
| `signal.preference_type` | `ingestSignal()` | Empty string, non-string |
| `event.user_id` | `dispatchSignals()` | Empty string, non-string |
| `event.timestamp` | `dispatchSignals()` | Empty string, unparseable ISO date |
| Corrupt JSON file | `loadFile()` | Wraps `JSON.parse` in try/catch with clear error message |
| Write atomicity | `saveFile()` | Atomic write-then-rename with Windows EPERM fallback |

**What's intentionally left unvalidated** (metadata strings that don't affect correctness or serve as keys/paths): `event.artifact_id`, `event.session_id`, `signal.tag_text`, `entry.diff` in `logRefinement`.

**Validation:**

```
pnpm --filter @open-design/creative-memory-system test       # 94/94
pnpm --filter @open-design/creative-memory-system typecheck  # 0 errors
pnpm guard                                                    # passes
```

The subsystem now validates every input that could corrupt stored state, escape the storage root, pollute object prototypes, or silently invert user intent. The remaining unvalidated fields are pure metadata that is stored as-is and never interpreted by the engine.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1100 @@
...
+}
+
+/** Keys that would corrupt plain-object semantics if used as property names. */
+const UNSAFE_OBJECT_KEYS = new Set(["__proto__", "constructor", "prototype"]);
+
+/**
+ * Validate a project_id for use as a key on project_overrides (a plain object).
+ * Rejects prototype-polluting keys and non-safe-basename strings.
+ */
+function validateProjectId(projectId: string): void {
```
nettee, line 153: This validation still leaves other inherited object keys such as `toString` and `valueOf` usable as `project_id`s even though `project_overrides` is stored as a plain `{}`. In the changed code, `initFile()` creates `project_overrides: {}`, `getPrefArray()` checks `if (!data.project_overrides[projectId])`, and then returns `data.project_overrides[projectId]!` as if it were a real `Preference[]`. For `project_id = "toString"`, that lookup resolves to the inherited function, so the next project-scoped ingest crashes on `array.push(...)` instead of creating an isolated bucket. Because `project_id` comes from runtime payloads and retrieval context, this is still a correctness/safety bug in the current head. Please either normalize `project_overrides` to a null-prototype object (or `Map`) everywhere it is created/read/reset, plus use `Object.hasOwn(...)` for existence checks, or expand validation to reject every inherited key rather than only `__proto__`, `constructor`, and `prototype`.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1100 @@
...
+  fs.mkdirSync(path.dirname(fp), { recursive: true });
+  data.last_updated = now();
+  const content = JSON.stringify(data, null, 2);
+  // Attempt atomic write (write-to-temp + rename). On platforms where rename
+  // fails (Windows EPERM due to file locking), fall back to direct write.
+  // The atomic path prevents corruption from mid-write crashes on POSIX;
+  // the fallback ensures the subsystem still works on Windows where rename
+  // semantics are less reliable under concurrent access.
+  const tmpFp = `${fp}.${process.pid}.tmp`;
+  try {
```
nettee, line 209: This broad fallback turns any temp-write or rename failure into a best-effort direct write. That means an unrelated required-step failure in the atomic path is silently downgraded instead of surfacing, which conflicts with the fail-fast/error-visibility rules this repo uses for local tooling and storage code. In the changed lines, the catch block does not inspect the error at all, so ENOSPC/permission/path problems are handled the same way as the intended Windows rename edge case, and cleanup failures are ignored too. Please catch the actual error object, only fall back for the specific recoverable rename case you want to support, and rethrow everything else so required I/O failures remain observable.

packages/creative-memory-system/src/extractionAdapter.ts
```diff
@@ -0,0 +1,274 @@
...
+  }
+  // Validate timestamp is a parseable date (catches "not-a-date" etc.)
+  if (Number.isNaN(new Date(event.timestamp).getTime())) {
+    throw new Error(
+      `dispatchSignals: event.timestamp "${event.timestamp}" is not a valid ISO date string.`,
+    );
+  }
+
+  const signals = classifyArtifact(event.artifact_meta);
+  const scope: "global" | "project" = event.project_id ? "project" : "global";
```
nettee: `project_id` is still interpreted by truthiness here, so an empty-string payload is silently treated as global scope instead of being rejected. On the changed lines, `const scope = event.project_id ? "project" : "global"` and `project_id: event.project_id || null` mean a malformed project-scoped event such as `{ project_id: "" }` is persisted as global memory. That leaks project-only preference signals across projects and hides the upstream wiring bug rather than surfacing it. Please validate that `project_id` is either `null` or a safe non-empty ID before dispatch, and derive scope with an explicit null check (`event.project_id === null`) instead of truthiness. Add a regression that proves `""` cannot fall through to global storage.

nettee (changes requested):
I found three blocking issues that still affect project override key safety, adapter scope handling, and required-file write semantics on the current head. Details are inline.


itsmeved24:
All three fixed in `244107f7`.

**1. Inherited keys on `project_overrides`.** Right call — `validateProjectId` was only blocking `__proto__`/`constructor`/`prototype`, not `toString`, `valueOf`, `hasOwnProperty`, etc. Two-layer fix:

- `project_overrides` is now a null-prototype object (`Object.create(null)`) in `initFile`, restored on every `loadFile` (since `JSON.parse` reintroduces `Object.prototype`), and reset to null-prototype in `resetMemory({ scope: "all" })`.
- All existence checks now use `Object.hasOwn(data.project_overrides, projectId)` instead of truthiness on the lookup, in `getPrefArray`, `retrieveForInjection`, and `listPreferences`.

`validateProjectId` is still there as defense-in-depth but is no longer the only line of defense — even if validation were bypassed, the null-prototype map plus `Object.hasOwn` would prevent inherited keys from resolving to functions.

**2. `saveFile` fallback is now narrow.** Agreed — the broad `catch {}` swallowed everything. The catch is now restricted to:

```ts
const isWindowsRenameRace =
  process.platform === "win32" && (code === "EPERM" || code === "EBUSY");
```

Anything else (`ENOSPC`, `EACCES`, write failures on the temp file, etc.) is rethrown after best-effort cleanup of the temp file. The Windows EPERM/EBUSY path stays because that's a real concurrent-access race, not an integration bug to surface.

**3. `event.project_id` empty-string truthiness.** Confirmed: `event.project_id ? "project" : "global"` would silently treat `""` as global, persisting project-scoped signals into global memory. The fix:

- Validate `event.project_id` is either `null` or a non-empty string. Empty string and other falsy non-null values now throw with a clear error.
- Derive scope via explicit null check: `event.project_id === null ? "global" : "project"`. No truthiness anywhere.
- Pass `event.project_id` directly (no `|| null` coercion) since it's now guaranteed valid.

New regression asserts `{ project_id: "" }` throws and creates no records.

**Validation:**

```
pnpm --filter @open-design/creative-memory-system test       # 96/96
pnpm --filter @open-design/creative-memory-system typecheck  # 0 errors
pnpm guard                                                    # passes
```

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1135 @@
...
+  }
+
+  const { project_id, preference_types } = context;
+  const diagnostics: Diagnostic[] = [];
+
+  // Stage 2 — Project-override merge
+  let prefs: RankedPreference[] = [...data.global_preferences];
+  let projectPrefs: Preference[] = [];
+
+  if (project_id && Object.hasOwn(data.project_overrides, project_id)) {
```
nettee, line 811: `retrieveForInjection()` still only validates `project_id` after this `Object.hasOwn(...)` guard passes. That means malformed runtime values such as `""`, `"foo/bar"`, or `"__proto__"` fall through this branch entirely and the function quietly injects global memory instead of surfacing the bad caller input. The current tests only cover the adapter ingestion path (`project_id: ""` now throws there), so this retrieval path can still leak project-scoped intent across generations while hiding the wiring bug. Please validate any provided `context.project_id` before the lookup using an explicit null/undefined check rather than truthiness, then keep the `hasOwn` test just for bucket existence. Add a regression around `retrieveForInjection(USER, { project_id: "" })` plus one unsafe key so the read path fails fast too.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1135 @@
...
+    );
+  }
+
+  const weight = SIGNAL_WEIGHTS[signal.signal_type];
+  if (weight === undefined) {
+    throw new Error(`Unknown signal_type: ${signal.signal_type}`);
+  }
+
+  const array = getPrefArray(data, signal.scope, signal.project_id);
+  const t = signal.timestamp || now();
```
nettee: The new fail-fast validation sweep still leaves one direct-store escape hatch here: `signal.timestamp || now()` fabricates a current timestamp when the caller passes `""` or another falsy value. `Signal.timestamp` is a required field, and the adapter side now rejects malformed timestamps, so silently replacing a bad timestamp here hides upstream bugs and changes decay/ordering behavior by inventing event time. Please require `signal.timestamp` to be a non-empty, parseable ISO string inside `ingestSignal()` as well, and add a regression for direct calls with `""` / `"not-a-date"` so the store matches the adapter's fail-fast contract.

nettee (changes requested):
I found two blocking correctness gaps that still affect direct store callers on this head; details are inline.


itsmeved24:
Both fixed in `2c53a9eb`.

**1. `retrieveForInjection` validates `context.project_id` before the lookup.**

Confirmed: `if (project_id && Object.hasOwn(...))` let `""`, `"foo/bar"`, `"__proto__"` etc. fall through the truthiness check entirely, silently downgrading to global-only retrieval. The fix moves validation *before* the lookup:

- If `project_id` is `null` or `undefined`, retrieval proceeds global-only (valid case).
- If `project_id` is defined, it must be a non-empty string and must pass `validateProjectId()` (basename + UNSAFE_OBJECT_KEYS).
- The `Object.hasOwn(data.project_overrides, project_id)` test stays, but now purely for bucket existence, not as a sneaky validation pass.

New regression `ADV-32` covers `""`, `"foo/bar"`, `"__proto__"` (all throw) and `null` / missing context (both proceed cleanly).

**2. `ingestSignal` requires a valid timestamp.**

Confirmed: `signal.timestamp || now()` fabricated event time for empty/falsy values, which silently changed decay and ordering. The store boundary now enforces the same fail-fast contract as the adapter:

```ts
if (typeof signal.timestamp !== "string" || signal.timestamp.length === 0) throw …;
if (Number.isNaN(new Date(signal.timestamp).getTime())) throw …;
```

`const t = signal.timestamp` (no `|| now()` fallback). Direct store callers cannot bypass this even if they skip the adapter.

New regression `ADV-33` covers `""` and `"not-a-date"` from direct `store.ingestSignal()` calls.

**Validation:**

```
pnpm --filter @open-design/creative-memory-system test       # 98/98
pnpm --filter @open-design/creative-memory-system typecheck  # 0 errors
pnpm guard                                                    # passes
```

Per @lefarcen's earlier guidance, this is review-feedback-only — no speculative expansion beyond what was flagged.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1166 @@
...
+
+      const excess = catPrefs.slice(MAX_PER_CATEGORY);
+      catTrimmed.push(...excess);
+    }
+
+    if (catTrimmed.length > 0) {
+      const trimmedIds = new Set(catTrimmed.map((p) => p.id));
+      prefs = prefs.filter((p) => !trimmedIds.has(p.id));
+
+      // Stage 10 — Category backfill from overflow, preferring under-represented types
```
nettee, line 1010: After Stage 9 trims an over-represented positive category, this function never re-runs the Stage 7 negative-budget check, so the final injection set can violate `NEGATIVE_BUDGET_RATIO` even though the earlier pass succeeded. I reproduced this from the built package with 6 `layout` positives and 5 negatives: `retrieveForInjection()` returns 3 positives and 5 negatives, and the only diagnostic is `category_ceiling_applied`. That means a profile that is supposed to be capped at 50% negatives can still emit a 62.5% negative prompt after category balancing, which changes generation behavior and breaks the documented invariant. Please run one final polarity-balancing pass after category trim/backfill (or move the ratio solver to the end of retrieval) and add a regression that asserts the ratio still holds after category trimming.

packages/creative-memory-system/src/preferenceStore.ts
```diff
@@ -0,0 +1,1166 @@
...
+  const medPos = positives.filter((p) => p.confidence === "medium").map((p) => p.pattern);
+  const highNeg = negatives.filter((p) => p.confidence === "high").map((p) => p.pattern);
+  const medNeg = negatives.filter((p) => p.confidence === "medium").map((p) => p.pattern);
+
+  if (highPos.length) lines.push(`Prefer (high):    ${highPos.join(" · ")}`);
+  if (medPos.length) lines.push(`Prefer (medium):  ${medPos.join(" · ")}`);
+  if (highNeg.length) lines.push(`Avoid (high):     ${highNeg.join(" · ")}`);
+  if (medNeg.length) lines.push(`Avoid (medium):   ${medNeg.join(" · ")}`);
+
+  if (projectOverrides.length && projectId) {
```
nettee, line 1235: `buildPromptBlock()` appends every `projectOverride` pattern here, but Stage 11 only budgets the `positives`/`negatives` arrays above. In practice that means the emitted block can exceed `TOKEN_BUDGET` and also mention override patterns that were already dropped from the ranked set. I reproduced this from the built package with 9 globals plus 6 project overrides: Stage 11 stops at an estimated 198 tokens, but the final block is about 216 tokens and the `Project override:` line still lists `override_2` through `override_5`, which never made it into the budgeted retrieval result. Please derive the override summary from the final budgeted set (or budget that line as part of Stage 11) and add an end-to-end regression that checks project-scoped prompt blocks stay within the intended token budget.

nettee (changes requested):
I found two blocking retrieval correctness issues on the current head; details are inline.


itsmeved24:
Both fixed in `a1fef63b`, plus a proactive sweep for analogous patterns.

**1. Polarity ratio violated after category trim.**

Confirmed: Stage 9 trims over-represented positive categories, which lowers the positive count without revisiting the negative budget. The fix adds **Stage 10b**, a polarity re-check that runs after the category ceiling/backfill and trims weakest negatives back to the limit when needed. `MIN_NEG_FLOOR` still applies.

The reproduction case in your comment (6 positives + 5 negatives → 3 pos + 5 neg = 62.5%) now lands at 3 pos + 3 neg = 50%, with the new diagnostic trace `Diversity re-check after category trim` so the second pass is observable.

**2. `buildPromptBlock` exceeds budget and lists dropped overrides.**

Two coordinated fixes for what is really one symptom of the same shape: Stage 11 not knowing about the override line.

- **Reserve the override-line cost up-front in Stage 11.** The reservation accounts for wrapper text (`Project override: … active (id)`) plus the per-pattern tokens for *each* override pattern, since override patterns appear twice in the rendered block (once in their `Prefer/Avoid` line, once in the override line). Previously only the first occurrence was counted.
- **Derive `projectOverrides` from the budgeted set.** Instead of a fresh filter on `projectPrefs` that ignored hard cap / diversity / budget, `projectOverrides` is now `projectPrefs.filter(p => budgetedIds.has(p.id))`. Patterns that were dropped from the ranked set can no longer reappear in the override line.

**Proactive sweep — analogous "earlier stage's invariant violated by later stage" issues:**

I audited every stage transition for the same pattern and found one more genuine case worth fixing while in the area:

- **Stage 11 drops by token cost without polarity awareness**, so it can drop a positive and leave the ratio violated even after Stage 10b passed. Added **Stage 11b**, a final polarity check against the post-budget set. Same `MIN_NEG_FLOOR`-respecting logic, no backfill (overflow is exhausted by Stage 11).

What I checked and decided was already safe:

- Stage 8 backfill respects Stage 6 (only fills slots freed by Stage 7's trim → total ≤ cap). ✓
- Stage 8 backfill before Stage 9 → category quota catches any per-type over-representation downstream. ✓
- Stage 11's drops cannot violate Stage 6 (only removes records). ✓
- Stage 11's drops cannot violate Stage 9 (per-type counts only decrease). ✓

**Regressions added:**

- `ADV-34`: 6 layout positives + 5 distinct-category negatives → asserts the polarity ratio holds in the final result (was 62.5% before, now ≤ 50% with floor).
- `ADV-35`: 9 long-pattern globals + 6 project overrides → asserts the rendered block stays under `TOKEN_BUDGET` and every `projectOverride` pattern is also in the budgeted set.

**Validation:**

```
pnpm --filter @open-design/creative-memory-system test       # 100/100
pnpm --filter @open-design/creative-memory-system typecheck  # 0 errors
pnpm guard                                                    # passes
```

itsmeved24:
Thanks @qiongyu1999  fair ask. Honest framing first: this PR is the memory engine package (packages/creative-memory-system/), not a host-integrated UI. By design (it's a documented non-goal in the README) the package has no user-facing surface of its own. The actual UX affordances you flagged earlier : inspect / edit / disable / reset, live at the host-app layer and depend on how apps/daemon and apps/web choose to surface them, which is one of the integration-boundary questions in open-questions.md that the team's internal work would resolve.

What I can demo end-to-end at this layer is the engine flow itself: a terminal walkthrough showing signals in → JSON state evolving → decay/reversal lifecycle → retrieval → prompt block in the next generation, with the diagnostic events that make each balancing decision visible. That's the level the package operates at and the level a product call on the engine shape can be made on.

Want me to record that, or would a written walkthrough with the actual JSON state at each step be more useful for your re-review? Both are quick.

itsmeved24:
@qiongyu1999 @lefarcen — added the end-to-end demo in the latest commit. Two artifacts under [`packages/creative-memory-system/demo/`](https://github.com/nexu-io/open-design/tree/add/creative-memory-system/packages/creative-memory-system/demo):

**Runnable script** (`demo/walkthrough.ts`):

```bash
pnpm --filter @open-design/creative-memory-system demo
```

Spins up an isolated scratch directory, walks through nine numbered scenarios, prints state snapshots / retrieval results / actual prompt blocks at each step, and cleans up on exit. Never touches real user data.

**Narrative transcript** (`demo/walkthrough.md`) — same nine scenarios as a markdown read-through, with each step explaining *why this matters* in user-experience terms:

1. Fresh user — empty injection set, no surprise context
2. First positive signal arrives — recorded, but not yet injectable
3. Confidence ladder climbs — preference crosses the threshold and appears in the prompt
4. Negative preference — rejection memory becomes a first-class "Avoid" line
5. Reversal lifecycle — graduated ladder + shadow record creation
6. Shadow promotion — genuine taste shift, original archived, no flicker
7. Decay and archive — `last_seen`-based archival
8. Realistic profile — full retrieval pipeline with category ceiling firing and a structured diagnostic
9. Project override — same pattern, different context, different prompt

**Honest framing on scope.** The package has no UI surface of its own (a documented non-goal in the README). The demo is at the engine layer because that's the level the package operates at. What the demo *intentionally* does not show:

- User-facing controls (inspect / edit / disable / reset). The package exposes everything a host needs to build that UI (`listPreferences`, `updatePreference`, `deletePreference`, `resetMemory`, the `memory_enabled` flag) but does not impose a UI shape — that lives in `apps/daemon` / `apps/web` and depends on the integration-boundary questions in [`docs/open-questions.md`](https://github.com/nexu-io/open-design/blob/add/creative-memory-system/packages/creative-memory-system/docs/open-questions.md).
- Wired pipeline integration. The adapter handlers in `src/extractionAdapter.ts` are stubs awaiting pipeline-team confirmation on event timing and attachment points.

@qiongyu1999 — happy to extend in either direction once you've had a chance to look. If you want a screen recording in addition to the runnable transcript I can do that too, or if seeing this clarifies that you'd rather I stage a host-integrated prototype, that's a different conversation about scope and we can plan it.

Validation: 100/100 vitest specs, `pnpm guard`, `pnpm typecheck` all pass.

lefarcen:
Could you share a short clip of the creative-memory subsystem in action — what users see, when memory surfaces, what triggers a recall? It's hard to gauge UX impact from the diff alone.

itsmeved24:
@lefarcen — moved the derivation-lifecycle / raw-events conversation into [`packages/creative-memory-system/docs/open-questions.md`](https://github.com/nexu-io/open-design/blob/add/creative-memory-system/packages/creative-memory-system/docs/open-questions.md) as Section 11 ("Derivation lifecycle and raw-events contract"). Synthesizes the @Ilya0527 ↔ @lefarcen exchange into a single explicit contract that future implementation work can build to.

What the section captures:

- **The attribution problem.** Unary signals can't isolate which dimension the user actually objected to. Without contrastive context, an extractor's salience hardens into phantom preferences after enough events.
- **Forward-looking contract.** Raw events canonical, derived features cacheable, pairwise > unary for attribution where the host UX supports it.
- **Cache-invalidation policy.** Event-count primary + 30d time fallback, lazy re-derive on read for MVP with background-worker upgrade path documented, store both `last_derived_event_seq` and `source_event_count`, version-tagged derivations.
- **Explicit pairwise/unary weighting.** Visible on each derived record so the policy is auditable rather than buried in code.

What the section is honest about:

- The current engine is **write-time-derivation only** — `ingestSignal` mutates preference records directly, there is no raw-events log, the `refinement_log[]` is dormant. Section 11 documents the boundary the agreement says we should hold, not what the code does today.
- This is a meaningful restructuring (new raw-events store, derivation layer, pairwise event type, staleness fields, lazy re-derive paths) and should drive the next iteration rather than be retrofitted into this PR.

Per the earlier "no large new code expansion" guidance from review, no engine changes here — just the documented contract. If/when the foundation-vs-reference call lands and someone (this PR's evolution or the OpenHuman track) starts on the next iteration, Section 11 is the shape to build to.

lefarcen:
Hey @itsmeved24 — quick status note: @nettee's last review was against `2c53a9eb` (May 17), but you've pushed several rounds of fixes since then and the current head is considerably ahead of that. @nettee, could you take a look at the current head when you get a chance? The author has been very responsive across each round and the diff has moved substantially since your last pass.

@itsmeved24 — appreciate the patience through all the iterations. Nothing blocking on our side right now; just waiting on the re-review to complete.

itsmeved24:
Thanks for the sync, @lefarcen. Quick correction on the two retrieval items — both were addressed on May 16 in commit [`a1fef63b`](https://github.com/nexu-io/open-design/pull/1746/commits/a1fef63b):

- **Stage 9 → polarity ratio.** Stage 10b runs after category trimming and re-trims negatives to satisfy `NEGATIVE_BUDGET_RATIO` against the post-trim positive count, with `MIN_NEG_FLOOR` respected. Covered by regression `ADV-34` ("polarity ratio holds after category trimming").
- **`buildPromptBlock` token budget.** Stage 11 reserves the override-line cost up-front (wrapper text + duplicated pattern tokens). `projectOverrides` is now derived from the budgeted set, so dropped patterns can't reappear in the override line. Covered by regression `ADV-35` ("project-scoped prompt block stays within token budget").

Stage 11b was added in the same commit as a proactive sweep — final polarity check against the post-budget set. Test count went 98 → 100 in that change, currently 100/100.

The line 1010 / 1235 references look like they're against an older revision. On the current head (`3ac4f9b9`) the Stage 10b block is in `preferenceStore.ts` inside the category-ceiling section, and the override-line reservation is in the `Stage 11` comment block above it. Happy to walk through either if I've misread which lines you meant.

On the merge state: agreed, the branch needs a rebase. I had been deferring it on the read of your earlier note ("for now, keeping the red state and the Node/toolchain explanation visible on PR #1746 is enough") plus the PR being parked on `needs-product-direction`. Happy to do the rebase + Node-24 lockfile regen now if that's the right next step — let me know.

lefarcen:
Hey @itsmeved24 — bringing the product-direction sync from #1637 back onto this PR so the next step is clear.

@Eli-tangerine's read is that PR #1746 is valuable as a concrete RFC/prototype/reference for the memory roadmap, but we should not treat it as the final foundation to merge unchanged yet. If you want to keep this branch alive, yes: the next useful step is to rebase against `main` and get CI green again. After that, pool/product can evaluate whether this lands as a reference package or feeds a follow-up pipeline/product integration plan.

No need to expand the engine scope while doing that — the integration plan remains the separate decision point.

itsmeved24:
Quick status update — branch is back in sync with `main` and CI is fully green again, per @lefarcen's earlier note that "the next useful PR-side step is the practical one Eli called out: rebase against `main` and get CI green again, while keeping the package focused and reviewable" ([comment](https://github.com/nexu-io/open-design/issues/1637#issuecomment-4535240444)).

What landed (commits `19e1aec0` → `78df59e8`):

- Merged `upstream/main` into the branch (no rebase — preserves the existing review threads on this PR; same end state for CI).
- Regenerated `pnpm-lock.yaml` under Node 24.16.0 + pnpm 10.33.2 (the previous Node 22 ↔ Node 24 lockfile drift I'd flagged is gone).
- Refreshed `nix/pnpm-deps.nix` with the new `daemonHash` / `webHash` values that Linux Nix CI computes against the regenerated lockfile.
- No source changes inside `packages/creative-memory-system/` — engine surface is unchanged from `3ac4f9b9`.

Current check state:

- ✅ `Validate Nix flake`
- ✅ `Validate workspace` (Required)
- ✅ `Workspace unit tests` / `Daemon workspace tests` / `Web workspace tests` / `Browser tests` / `Build workspaces`
- ✅ `Validate landing page`, visual capture/verify, fork-PR approval

The "Changes requested" banner is the May-17 review against `c53e9eb` (per the [thread sync here](https://github.com/nexu-io/open-design/issues/1637#issuecomment-4524282250)). Both retrieval items it called out — Stage 9 polarity ratio after category trimming, and `buildPromptBlock` token-budget accounting — were addressed in `a1fef63b` on May 16, covered by regressions `ADV-34` and `ADV-35`. @lefarcen already noted on the issue that "those comments were against stale context" ([comment](https://github.com/nexu-io/open-design/issues/1637#issuecomment-4535240444)). No code changes needed for that review pass; happy to walk through either fix on the current head if it's useful before re-review.

PR remains parked per @Eli-tangerine's product call: this is valuable as a concrete RFC/prototype and input into the memory roadmap, not assumed to be the final foundation by itself. Not pushing for a merge here — just keeping the artifact green and reviewable while that direction settles. The integration-boundary discussion (signal capture points, retrieval into generation/critique, user controls) stays the substantive next conversation, on the issue rather than the PR.

itsmeved24:
One personal note alongside the status update: I'd like to keep contributing on the integration-boundary side (signal capture, retrieval into generation/critique, user controls) regardless of whether this specific package ends up being the foundation. The doc trail in `open-questions.md` — especially Section 11's raw-events contract and the @Ilya0527 content-addressed-derivations refinement — is the part I'd most like to see carried forward when the memory roadmap moves. Happy to write up the integration-shape proposal as a separate doc-only PR or issue thread if that would be useful @lefarcen .

The last ten or so days of solo work on the engine, simulation suite (238 invariants), TypeScript port, Vitest specs, docs, demo, and the CI green-up today have been a fair amount of time, and I'd rather invest the next stretch in something that plugs into the team's plan than keep iterating on this branch in isolation.

lefarcen:
Hey @itsmeved24 — appreciate you saying this, and the amount of careful work here is very visible.

I agree the most useful next contribution is not more isolated engine churn on this branch. A focused integration-shape write-up would be useful, but let's keep it anchored on #1637 / the existing `open-questions.md` thread rather than broadening this PR. My suggestion: either add a concise proposal there, or open a doc-only PR that covers the decision points product/pipeline need to react to:

- signal capture points
- retrieval insertion into generation / critique
- user controls and escape hatches
- how Section 11's raw-events + content-addressed derivations shape the contract

That keeps the path concrete without assuming this package is the final foundation. Thanks again for keeping the artifact green and reviewable while the roadmap direction settles.

itsmeved24:
Integration-shape follow-up landed at [#3021](https://github.com/nexu-io/open-design/pull/3021) per @lefarcen's routing on [#1637](https://github.com/nexu-io/open-design/issues/1637#issuecomment-4537045117). This PR stays parked as RFC input; engine is no longer the active gate. Will keep the branch green if `main` moves underneath it.

PR #1749, opened 2026-05-15
fix(web): gate Finalize design package + trim handling reviewers asked for on #1649

### LINKED ISSUE #1348: Finalize design package is clickable without provider config and fails with Bad request instead of being gated

ecogsandbox:
## Why

**Use case:** Mirror PR #1649's gating fix for #1348, but include the one outstanding piece of reviewer feedback that PR #1649 has not yet incorporated: trim handling on `apiKey` / `model`.

**Pain being addressed:** Without trim handling, whitespace-only values (`\"   \"`) still pass the JS truthy check while the daemon route rejects them (`!apiKey.trim()` at `apps/daemon/src/import-export-routes.ts:551`). The user sees the same backend `Bad request` toast this fix is meant to prevent. This is the exact `.trim()` contract mismatch that closed PR #1367 (per @PerishCode review feedback) and that @Siri-Ray, @lefarcen, and the Codex bot are blocking #1649 on:

- https://github.com/nexu-io/open-design/pull/1649#discussion_r3239468985
- https://github.com/nexu-io/open-design/pull/1649#discussion_r3239444172

Happy to close this in favor of #1649 once @xxiaoxiong pushes the trim addition. Opening so reviewers can pull a verified branch in the meantime.

## What users will see

- \"Finalize design package\" button is disabled when \`apiKey\` or \`model\` is empty or whitespace-only; tooltip reads \"Configure API key and model in Settings to enable finalization\".
- Defensive toast in \`handleFinalize\` covers programmatic triggers: \"Cannot finalize: API key and model are required.\"
- Trimmed values are passed into \`finalize.trigger\` so the UI gate matches the daemon's \`apiKey.trim()\` / \`model.trim()\` validation.

## Surface area

- [x] UI — Finalize button disabled state, tooltip, defensive toast
- [ ] Keyboard shortcut — no changes
- [ ] CLI / env var — no changes
- [ ] API / contract — no changes
- [ ] Extension point — no changes
- [ ] i18n keys — no changes (English-only tooltip, matches existing strings)
- [ ] New top-level dependency — no changes
- [ ] Default behavior change — gates an existing action

## Screenshots

N/A — gating change. Disabled button + tooltip render via the native \`disabled\` attribute and \`title\`.

## Bug fix verification

1. Open a project with provider config empty (Local CLI mode or unconfigured BYOK).
2. Before fix: Finalize is clickable; click → backend 400 → toast \"Bad request — check the API key and model.\"
3. After fix: Finalize is disabled; tooltip explains the prerequisite.
4. Set apiKey + model in Settings; button re-enables; click finalizes normally.
5. Edge case (this PR vs. #1649): set apiKey to \`\"   \"\` (whitespace only). Before this PR's trim handling, button is enabled and click still 400s; with trim handling, button stays disabled.

## Validation

- \`pnpm guard\` — green
- \`pnpm typecheck\` — green
- \`pnpm --filter @open-design/web test\` — 1189/1189 passed

Fixes #1348
References #1649, #1367

---

Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/apps/web/src/components/ProjectView.tsx
+++ b/apps/web/src/components/ProjectView.tsx
@@ -2294,11 +2294,25 @@ export function ProjectView({
   // Continue in CLI / Finalize design package handlers + keyboard
   // shortcut wiring. Close to the JSX so the data flow is easy to
   // trace from the toolbar back to its sources.
+  
+  // Check if finalize is available: requires apiKey and model (#1348)
+  const canFinalize = Boolean(config.apiKey?.trim() && config.model?.trim());
+  
   const handleFinalize = useCallback(() => {
+    // Guard against missing config (should be prevented by disabled state,
+    // but defend in depth in case the button is triggered programmatically)
+    if (!config.apiKey?.trim() || !config.model?.trim()) {
+      setProjectActionsToast({
+        message: 'Cannot finalize: API key and model are required.',
+        details: 'Configure your provider in Settings before finalizing.',
+      });
+      return;
+    }
+    
     void finalize.trigger({
-      apiKey: config.apiKey,
+      apiKey: config.apiKey.trim(),
       baseUrl: config.baseUrl,
-      model: config.model,
+      model: config.model.trim(),
       maxTokens: effectiveMaxTokens(config),
     }).then((result) => {
       if (result) void designMdState.refresh();
@@ -2500,6 +2514,7 @@ export function ProjectView({
       <ProjectActionsToolbar
         designMdState={designMdState}
         finalizeStatus={finalize.status}
+        canFinalize={canFinalize}
         onFinalize={handleFinalize}
         onCancelFinalize={handleCancelFinalize}
         onContinueInCli={handleContinueInCli}
--- a/apps/web/src/components/FinalizeDesignButton.tsx
+++ b/apps/web/src/components/FinalizeDesignButton.tsx
@@ -16,6 +16,7 @@ import type { FinalizeStatus } from '../hooks/useFinalizeProject';
 export interface FinalizeDesignButtonProps {
   designMdState: Pick<DesignMdState, 'exists' | 'isStale'>;
   status: FinalizeStatus;
+  disabled?: boolean;
   onFinalize: () => void;
   onCancel: () => void;
 }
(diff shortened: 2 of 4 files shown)


lefarcen (commented):
Hey @ecogsandbox! 👋 The trim-contract analysis here is precise — the mismatch between the JS truthy check and the daemon's `apiKey.trim()` at `import-export-routes.ts:551` is exactly the gap reviewers flagged on #1649, and your bug-fix verification steps (especially edge case 5) show it end-to-end.

Heads-up: PR #1649 by @xxiaoxiong is also open against this area — both PRs touch `FinalizeDesignButton.tsx` and address the Finalize gating for #1348. This PR adds the `.trim()` handling that #1649 hasn't yet incorporated. The maintainer team will decide which one lands; sharing so neither effort goes to waste.

@mrcfps has been assigned to review. ✨

apps/web/src/components/ProjectView.tsx
```diff
@@ -2294,11 +2294,25 @@ export function ProjectView({
...
+      return;
+    }
+    
     void finalize.trigger({
-      apiKey: config.apiKey,
+      apiKey: config.apiKey.trim(),
       baseUrl: config.baseUrl,
-      model: config.model,
+      model: config.model.trim(),
       maxTokens: effectiveMaxTokens(config),
```
mrcfps, line 2316: `apps/web/src/components/ProjectView.tsx:2316` now trims `config.model` before calling `finalize.trigger`, but `maxTokens` is still computed from the untrimmed `config` object. When a saved model has leading or trailing whitespace and there is no explicit max-token override, `effectiveMaxTokens(config)` falls through to `8192` because `apps/web/src/state/maxTokens.ts:96` looks up the raw model string, while the request is sent with the trimmed model id. That means a padded value like ` claude-sonnet-4-5 ` can still finalize with the fallback token cap and get unnecessarily truncated even though this PR is trying to normalize whitespace-only config issues. Please derive the trimmed values once, pass the trimmed model into `effectiveMaxTokens`, and add a regression in `apps/web/tests/state/maxTokens.test.ts` for a whitespace-padded known model id.

mrcfps (commented):
Thanks for the #1348 follow-up, @ecogsandbox — the main gate looks good, but I found one trim-path mismatch that still needs a small follow-up before the behavior fully lines up with the daemon validation.

PR #1776, opened 2026-05-15
fix: block internal storage refs in html artifacts

### LINKED ISSUE #1665: Orbit follow-up preview artifacts can save invalid reserved-path references

prantikmedhi:
Fixes #1665

## Summary
Reject AI-emitted HTML artifacts that reference internal storage paths like `.live-artifacts`, `.od`, or `.tmp` before they are saved.

## Why
Orbit can generate a follow-up artifact that embeds `.live-artifacts/` in an iframe or link. That content may save successfully, but the resulting file later opens with a 400 because the reserved path is blocked by the file API.

## What users will see
Instead of silently saving a broken artifact, the save is rejected with a clear validation error.

## Surface area
- [x] Default behavior change

## Validation
- `pnpm --dir apps/web exec vitest run -c vitest.config.ts tests/artifacts/validate.test.ts`

### DIFF
--- a/apps/web/src/artifacts/validate.ts
+++ b/apps/web/src/artifacts/validate.ts
@@ -14,6 +14,8 @@
  * - the *first* non-whitespace token is `<!doctype html>` or `<html`
  *   (anchored at the start; mid-string mentions of these tags do NOT count —
  *   AI prose like "Updated the <html lang> attribute…" must be rejected)
+ * - does not reference internal storage paths such as `.live-artifacts/`,
+ *   `.od/`, or `.tmp/`
  *
  * What this gate is NOT:
  * - It is **not** an HTML linter or validator. Malformed but recognizably
@@ -36,6 +38,7 @@
 
 const MIN_HTML_LENGTH = 64;
 const STARTS_WITH_DOCUMENT_RE = /^(?:<!doctype\s+html\b|<html\b)/i;
+const INTERNAL_STORAGE_PATH_RE = /(?:^|[./"'`(=\s])(?:\.live-artifacts|\.od|\.tmp)(?:[/?#]|$)/i;
 
 export type HtmlArtifactValidationResult =
   | { ok: true }
@@ -51,6 +54,9 @@ export function validateHtmlArtifact(content: string): HtmlArtifactValidationRes
   }
   if (!STARTS_WITH_DOCUMENT_RE.test(trimmed)) {
     return { ok: false, reason: 'content does not start with <!doctype html> or <html — looks like prose, not a complete HTML document' };
+  }
+  if (INTERNAL_STORAGE_PATH_RE.test(trimmed)) {
+    return { ok: false, reason: 'content references an internal storage path such as .live-artifacts, .od, or .tmp' };
   }
   return { ok: true };
 }
--- a/apps/web/src/artifacts/validate.ts
+++ b/apps/web/src/artifacts/validate.ts
@@ -38,7 +38,7 @@
 
 const MIN_HTML_LENGTH = 64;
 const STARTS_WITH_DOCUMENT_RE = /^(?:<!doctype\s+html\b|<html\b)/i;
-const INTERNAL_STORAGE_PATH_RE = /(?:^|[./"'`(=\s])(?:\.live-artifacts|\.od|\.tmp)(?:[/?#]|$)/i;
+const INTERNAL_STORAGE_PATH_RE = /(?:^|[./"'`(=\s])(?:\.live-artifacts|\.od|\.tmp)(?:[/?#"'`\s>)]|$)/i;
 
 export type HtmlArtifactValidationResult =
   | { ok: true }
--- a/apps/web/tests/artifacts/validate.test.ts
+++ b/apps/web/tests/artifacts/validate.test.ts
@@ -52,6 +52,13 @@ describe('validateHtmlArtifact', () => {
     expect(result.ok).toBe(false);
   });
 
+  it('rejects an otherwise valid document that points at an internal storage path', () => {
+    const html = '<!doctype html><html><body><a href="/.live-artifacts/artifact-1/preview.html">preview</a><p>Enough content to look like a real document.</p></body></html>';
+    const result = validateHtmlArtifact(html);
+    expect(result.ok).toBe(false);
+    if (!result.ok) expect(result.reason).toMatch(/internal storage path/i);
+  });
+
   it('accepts a complete <!doctype html> document', () => {
     const html = '<!doctype html><html><head><title>x</title></head><body><h1>hello</h1></body></html>';
     const result = validateHtmlArtifact(html);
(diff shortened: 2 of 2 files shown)


lefarcen (commented):
Hey @prantikmedhi! The regex-gate approach in `validateHtmlArtifact` is exactly the right layer to catch these reserved-path references early — rejecting before save is cleaner than failing on open.

We have a [PR template](https://github.com/nexu-io/open-design/blob/main/.github/pull_request_template.md) that helps pool reviewers scope changes quickly. A few additions would help:

- **Why** — a short user-visible description of the failure scenario (Orbit follow-up artifact embeds `.live-artifacts/` in an iframe, saves fine, but opens with a 400 because the file API blocks that reserved path)
- **What users will see** — what changes after this lands (artifact save is rejected with a clear validation error rather than silently succeeding)
- **Surface area** — please tick the relevant box(es) in the 9-box checklist — this looks like at least **Default behavior change**
- **Validation** — the `pnpm vitest run` command is there, but it would help to note which test(s) were red before and are green now

Once those are in, the reviewer can scope this without opening the diff twice. Thanks for the contribution! 🙌

apps/web/src/artifacts/validate.ts
```diff
@@ -36,6 +38,7 @@
 
 const MIN_HTML_LENGTH = 64;
 const STARTS_WITH_DOCUMENT_RE = /^(?:<!doctype\s+html\b|<html\b)/i;
+const INTERNAL_STORAGE_PATH_RE = /(?:^|[./"'`(=\s])(?:\.live-artifacts|\.od|\.tmp)(?:[/?#]|$)/i;
```
mrcfps: This regex still misses the exact root-directory links the PR is trying to block. The suffix only allows slash, question mark, hash, or end-of-string after the reserved segment, so <a href="/.live-artifacts"> (and the same shape for /.od and /.tmp) does not match because the next character is the closing quote. I verified that the regex returns false for that HTML, which means the artifact would still save and later fail when opened. Please expand the trailing delimiter check to treat common HTML terminators like double quotes, single quotes, whitespace, or > as valid segment boundaries, and add a regression test for a reserved-path link without a trailing slash.

mrcfps (changes requested):
@prantikmedhi Thanks for tightening this guard — I found one regex edge case that still lets the broken reserved-path artifact through, so I’m asking for one follow-up before merge.


lefarcen (commented):
Hey @prantikmedhi, thanks for updating the PR body — the failure scenario and user-facing behavior are much easier to scope now. One remaining template detail for this bugfix: could you add the red→green verification note for the regression seam (which reserved-path case failed before this fix and now passes, or note that the new test covers that gap)? The test command is useful; that extra line helps reviewers quickly confirm the exact bugfix coverage.

apps/web/src/artifacts/validate.ts
```diff
@@ -36,6 +38,7 @@
 
 const MIN_HTML_LENGTH = 64;
 const STARTS_WITH_DOCUMENT_RE = /^(?:<!doctype\s+html\b|<html\b)/i;
+const INTERNAL_STORAGE_PATH_RE = /(?:^|[./"'`(=\s])(?:\.live-artifacts|\.od|\.tmp)(?:[/?#"'`\s>)]|$)/i;
```
mrcfps, line 41: This regex still matches plain text mentions of these names, not just path references. Because the leading character class includes whitespace and the trailing class includes whitespace, a valid document like `<p>The .tmp directory stores metadata.</p>` or `<p>The .od folder is reserved.</p>` now trips this check even though nothing in that HTML would ever hit the reserved-path file route. I verified `INTERNAL_STORAGE_PATH_RE.test(...)` returns `true` for those text-only examples, so this would reject legitimate documentation/tutorial artifacts rather than only broken embeds. Please narrow the matcher to actual URL/path contexts (for example `href=`, `src=`, `url(...)`, or a required `/` / `./` segment boundary) and add a regression test showing that a document which merely mentions `.od` or `.tmp` in text still saves successfully.

mrcfps (changes requested):
@prantikmedhi Thanks for tightening this guard and for following up on the earlier regex gap. I found one remaining false-positive in the new internal-path check that can reject valid HTML saves, so I’m asking for one more follow-up before merge.

PR #1796, opened 2026-05-15
fix(web): back chevron honors history stack instead of pushing home

abhid-007:
Fixes #1789

## Why

Hit this myself on the 0.8.0 preview: enter a project from `/projects`, hit the in-app Back chevron, and you land on `/` (the Home tab) instead of returning to the Projects list with state intact. Same flow from `/design-systems` and `/tasks` also drops to Home. The bug feels worse than it sounds because the *browser* back button is still pointing at the project you just left — `handleBack` was pushing a fresh entry instead of popping, so the history stack quietly grew on every Back click.

This PR pays back the small bit of design debt in `handleBack` and pins the behavior with tests so it can't quietly regress.

## What users will see

- The in-app Back chevron on a project now returns the user to whatever surface they came from — Projects, Design Systems, Tasks, Plugins, Integrations, or Home — instead of always dropping to Home.
- Scroll position and active tab on the prior surface are preserved (because we now pop the real history entry rather than mount a fresh `EntryView`).
- Deep-linking directly to `/projects/<id>` in a fresh tab still falls back to Home on Back, since there is no prior in-app entry to return to.

## Surface area

- [x] **UI** — behavior of the in-app Back chevron in `EntryView` / project route
- [ ] Keyboard shortcut
- [ ] CLI / env var
- [ ] API / contract
- [ ] Extension point
- [ ] i18n keys
- [ ] Default behavior change — the *target* of Back changes for users who entered a project from a non-Home tab. Calling that out explicitly because it is a visible behavior shift, but it strictly matches the pre-existing browser-Back expectation, so there is no new opt-in to add.

## Screenshots

This is a navigation-stack change with no rendered surface diff — the same `EntryView` mounts, it just mounts in the tab the user came from. I can capture a short GIF showing the Projects → project → Back loop if it'd help review; flagging here rather than padding the PR with a screenshot of an unchanged page.

## Bug fix verification

- Test path that reproduces the bug: `apps/web/tests/App.test.ts` → `describe('resolveBackNavigation (issue #1789)', ...)` (4 cases).
- Did the test go red on `preview/v0.8.0` and green on this branch? **Yes.** The 4 new cases fail with `TypeError: resolveBackNavigation is not a function` against the unmodified `App.tsx`, and pass after the fix.
- Approach: factored the decision out of `handleBack` into a pure `resolveBackNavigation({ historyState, historyLength })` helper, matching the existing `resolveSettingsCloseConfig` / `buildPersistedConfig` pattern in this file. That lets us test the decision logic under vitest's `node` environment without standing up jsdom. `handleBack` then either calls `window.history.back()` or falls through to `navigate({ kind: 'home', view: 'home' })`.

A small note on the issue thread for context: the analysis comment described option 1 as "use `window.history.back()`" and option 2 as "add a `projects` route", but the router already has `view: 'projects' | 'tasks' | 'plugins' | 'design-systems' | 'integrations'` under `kind: 'home'` (see `apps/web/src/router.ts`). The real defect is that `handleBack` hardcoded `view: 'home'` regardless of which view the user came from. Popping history fixes that by construction — it returns to the actual prior `(kind, view)` pair without `handleBack` needing to know about it — which is why I went with option 1 as written and didn't widen scope to teach `handleBack` about per-view targets.

## Validation

- `pnpm guard` — passed.
- `pnpm typecheck` — passed.
- `pnpm --filter @open-design/web test` — **1456 / 1456 passing** (the 4 new `resolveBackNavigation` cases plus the existing suite, no regressions).
- Manual: Node 24.15 (per `engines: { node: '~24' }`), pnpm 10.33.2.

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -33,7 +33,7 @@ import {
   fetchSkills,
   uploadProjectFiles,
 } from './providers/registry';
-import { navigate, useRoute } from './router';
+import { navigate, useRoute, type Route } from './router';
 import {
   fetchDaemonConfig,
   DEFAULT_PET,
@@ -154,6 +154,31 @@ export function resolveSettingsCloseConfig(
 ): AppConfig {
   const base = latestPersisted === rendered ? rendered : latestPersisted;
   return base.onboardingCompleted ? base : { ...base, onboardingCompleted: true };
+}
+
+// Decision helper for the in-app Back chevron (issue #1789). The previous
+// implementation always pushed `{ kind: 'home', view: 'home' }`, which
+// (a) added a new entry instead of popping and (b) hardcoded the Home
+// tab regardless of which surface the user came from (e.g. /projects,
+// /design-systems). Honoring the real history stack covers every entry
+// surface for free; the fallback only fires for true deep-links where
+// there is nothing to pop back to.
+export type BackNavigationDecision =
+  | { action: 'pop' }
+  | { action: 'navigate'; route: Route };
+
+export function resolveBackNavigation(input: {
+  historyState: unknown;
+  historyLength: number;
+}): BackNavigationDecision {
+  // `history.state` is non-null whenever the SPA itself called
+  // pushState/replaceState — i.e. there is a real prior in-app entry to
+  // pop back to. `length > 1` covers the rarer case of a same-tab
+  // navigation that did not set state (e.g. external → in-app link).
+  if (input.historyState !== null || input.historyLength > 1) {
+    return { action: 'pop' };
+  }
+  return { action: 'navigate', route: { kind: 'home', view: 'home' } };
 }
 
 export function App() {
@@ -942,7 +967,15 @@ export function App() {
   }, []);
 
   const handleBack = useCallback(() => {
-    navigate({ kind: 'home', view: 'home' });
+    const decision = resolveBackNavigation({
+      historyState: window.history.state,
+      historyLength: window.history.length,
+    });
+    if (decision.action === 'pop') {
+      window.history.back();
+      return;
+    }
+    navigate(decision.route);
   }, []);
 
   const handleClearPendingPrompt = useCallback(() => {
--- a/apps/web/tests/App.test.ts
+++ b/apps/web/tests/App.test.ts
@@ -4,6 +4,7 @@ import {
   buildPersistedConfig,
   isAutosaveDraftOnlyChange,
   persistComposioConfigChange,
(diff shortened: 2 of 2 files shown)


apps/web/src/App.tsx
```diff
@@ -156,6 +156,31 @@ export function resolveSettingsCloseConfig(
...
+
+export function resolveBackNavigation(input: {
+  historyState: unknown;
+  historyLength: number;
+}): BackNavigationDecision {
+  // `history.state` is non-null whenever the SPA itself called
+  // pushState/replaceState — i.e. there is a real prior in-app entry to
+  // pop back to. `length > 1` covers the rarer case of a same-tab
+  // navigation that did not set state (e.g. external → in-app link).
+  if (input.historyState !== null || input.historyLength > 1) {
```
mrcfps, line 178: `historyLength > 1` is too broad a proxy for "there is a prior in-app entry to pop" here. Our router still writes `pushState(null, '', target)` / `replaceState(null, '', target)`, so Open Design entries never leave a non-null marker in `window.history.state`; in the real app this branch will therefore pop for *any* same-tab deep link with pre-existing browser history, including `/projects/:id` opened from an external page. That sends the chevron out of Open Design instead of taking the fallback-to-home path described in this PR. Can we store an app-owned marker in `navigate()` (for example `{ odRoute: true }`) and check for that specific marker before calling `window.history.back()` here, then add a regression case for the same-tab deep-link flow?

mrcfps (commented):
@abhid-007 Thanks for tightening up the back-chevron behavior and covering the happy path with focused tests. I found one follow-up in the new history heuristic that looks worth fixing before we depend on it more broadly.


lefarcen:
👋 Heads up — this PR's base was originally `preview/v0.8.0`. We've now merged `preview/v0.8.0` into `main` (see #1832), so going forward all PRs target `main` as usual. I've retargeted this one for you, and it auto-merges cleanly against the new base.

Next step is on the team side: review + CI. No action needed from you unless you want to rebase on the latest `main` proactively.

Thanks for the contribution! 🙌

---

👋 这个 PR 原本的目标分支是 `preview/v0.8.0`。我们已经把 `preview/v0.8.0` 合进了 `main`(详见 #1832),后续所有 PR 都跟以前一样提到 `main` 即可。我已经帮你把 base 改到 `main` 了,自动合并没冲突。

接下来由团队这边走 review + CI 流程,你这边不需要做什么。如果想主动跟一下 `main` 最新代码也欢迎。

感谢贡献! 🙌

PR #1806, opened 2026-05-15
fix(web): demote Plugins and Integrations to nav rail footer

zoeforfun:
## Why

The left nav rail currently places six items at equal visual weight:
Home, Projects, Automations, Plugins, Design Systems, Integrations.
After reviewing the product information architecture, Plugins and
Integrations don't belong in the primary navigation tier:

- **Plugins** is a platform extension mechanism (discover/install),
  not a daily-use destination.
- **Integrations** is a "use Open Design on other platforms" entry
  point — a one-time configuration action, not a recurring workspace.

Putting them at the same level as Projects, Automations, and Design
Systems dilutes the primary tier and signals to users that they are
equally important daily destinations, which they are not.

## What users will see

The primary nav group now shows four items only:
**Home · Projects · Automations · Design Systems**

Plugins and Integrations move to the bottom footer section of the
rail (alongside the existing help launcher), separated from primary
items by a thin 1 px divider. Both remain fully accessible via their
existing icons and tooltips — they're de-emphasised, not hidden.

## Surface area

- [x] Web UI (`apps/web`)
- [ ] Daemon / API
- [ ] Contracts / DTOs
- [ ] CLI flags or env vars
- [ ] i18n keys
- [ ] Design systems / skills / craft
- [ ] New root `package.json` dependencies

## Test plan

- [ ] Start `pnpm tools-dev` and confirm primary rail shows Home,
      Projects, Automations, Design Systems only
- [ ] Scroll/look at footer: Plugins and Integrations appear below
      the divider, above the help icon
- [ ] Click Plugins → Plugins view opens correctly
- [ ] Click Integrations → Integrations view opens correctly
- [ ] Active state (highlight) still works for both footer items
- [ ] Hover tooltips still appear for both footer items


lefarcen (commented):
Hey @zoeforfun, the IA rationale for moving Plugins and Integrations out of the primary rail is clear, and the manual checks you listed are useful. Before pool review picks this up, could you adjust the PR body so it matches the current template fields: use the current `## Surface area` checklist format with the applicable Web UI box checked, and add/rename the verification section to `## Validation` with the commands or checks you ran. That will make release scoping much easier for this preview-branch UI change.

lefarcen (commented):
Hey @zoeforfun, quick follow-up on the PR description after this push: the body still reads clearly on the nav-rail IA change, but the current template info is still not quite complete. Could you update it with the current `## Surface area` checklist format, a `## Bug fix verification` note for the red → green behavior this fixes, and a `## Validation` section for the checks you ran? Also, since the latest diff now includes the help menu social links, please reflect that user-visible scope in the description too so reviewers can evaluate the full change set.

apps/web/src/components/EntryShell.tsx
```diff
@@ -412,251 +333,17 @@ export function EntryShell({
...
-            </span>
-            <span>{t('avatar.settings')}</span>
-          </button>
-        </div>
-      ) : null}
-    </div>
+    <button
+      type="button"
+      className="settings-icon-btn"
+      onClick={() => onOpenSettings()}
```
mrcfps, line 381: Problem: this turns the cog into a direct `onOpenSettings()` button, but the responsive layout still hides `.entry-main__topbar-chips` below 900px and relied on the avatar popover's compact-only rows to keep the hidden actions reachable. Why it matters: at compact widths the execution/model switcher and the `Use everywhere` entrypoint disappear entirely, so smaller-window users lose those flows. Evidence: `entry-layout.css` still says "Collapse the three topbar chips into the settings dropdown" and sets `.entry-main__topbar-chips { display: none; }` inside the `@media (max-width: 900px)` block, while this hunk removes the dropdown rows that used to mirror those controls. Suggested change: either keep an actual popover for the compact-only actions, stop hiding the topbar chips at that breakpoint, or add another equivalent compact-width surface before merging.

apps/web/src/components/HomeHero.tsx
```diff
@@ -1353,6 +1352,52 @@ function getPluginQueryPreview(plugin: InstalledPluginRecord): string {
...
+}
+
+function TypeTabBar({
+  activeChipId,
+  pendingChipId,
+  pendingPluginId,
+  pluginsLoading,
+  onPickChip,
+}: TypeTabBarProps) {
+  const chips = useMemo(() => chipsForGroup('create'), []);
```
mrcfps, line 1433: Problem: moving the create chips into `TypeTabBar` changes both their DOM placement and their semantics (`role="tab"` / `aria-selected` instead of toolbar buttons with `aria-pressed`), but the existing rail spec still asserts the old structure. Why it matters: the PR is currently red in CI because `apps/web/tests/components/HomeHero.rail.test.tsx` still expects every chip to live under `home-hero-rail` and checks `aria-pressed` on the active create chip. Evidence: the failing `Validate workspace` run reports `HomeHero.rail.test.tsx:56` (`rail.contains(node)`) and `:70` (`aria-pressed`) against this new tab markup. Suggested change: update that spec in the same PR to split the output-type tabs from the lower starter-shortcut rail and assert `role="tab"` / `aria-selected` for the create group.

apps/web/src/components/HomeView.tsx
```diff
@@ -609,11 +609,21 @@ export function HomeView({
...
           chipId: chip.id,
           inputs: chip.action.inputs,
-        });
+        };
+        // Output-type tabs (create group) are mode-selection gestures:
+        // switching between them should never prompt for confirmation,
+        // even when the input already has template text from a previous
+        // tab. Migrate-group chips (From Figma, etc.) still go through
+        // the replacement guard because they carry a meaningful prompt.
+        if (chip.group === 'create') {
```
mrcfps, line 808: Problem: this intentionally bypasses `requestActivePlugin()` for create-group chips, so the replace-current-prompt confirmation is no longer shown there, but the prefill spec still asserts the old dialog flow. Why it matters: the web test suite is red on the current head, and this behavior change lands without matching coverage for the new contract. Evidence: `gh pr checks` shows `Validate workspace` failing at `apps/web/tests/components/HomeView.prefill.test.tsx:518` because the test still waits for `role="dialog"` / `name=/replace current prompt/i` after clicking `home-hero-rail-prototype`. Suggested change: update that test in this PR to assert the new immediate-apply behavior for create tabs, and keep a separate migrate-chip assertion around `requestActivePlugin()` so the confirmation path stays covered where it still applies.

mrcfps (changes requested):
@zoeforfun Thanks for the thoughtful IA cleanup here — I found a couple of blocking regressions to address before this can merge: the compact-width settings-cog path now drops existing actions, and the home-hero changes leave the web test suite red in CI.


mrcfps (approved):

PR #1814, opened 2026-05-15
fix(claude-design-import): restore notched wheel zoom + warn on regex drift

lefarcen:
Follow-up to #1726 picking up two review notes ([1](https://github.com/nexu-io/open-design/pull/1726#discussion_r4458348800) / [2](https://github.com/nexu-io/open-design/pull/1726#discussion_r4458348801)) that landed after merge.

### 1. Cmd+wheel zoom over-shrinks for physical mouse wheels

The original Claude Design canvas export split Cmd+wheel into two ratios:

- smooth `Math.exp(-deltaY * 0.01)` for trackpad pinch (small/fractional `deltaY`)
- notched `Math.exp(-Math.sign(deltaY) * 0.18)` for physical wheels (~100/120 px per click)

#1726's normalizer collapsed both onto the smooth formula. For a real mouse, a single notch (`deltaY ≈ 100`) now scales by `exp(-1.0) ≈ 0.367` — a **~63% shrink per click**. The old notched ratio was `exp(-0.18) ≈ 0.835`, i.e. **~17% per click**. So Cmd + mouse-wheel users overshoot dramatically on every click.

### 2. Both rewrite regexes fail silently on Claude Design template drift

`wheelBlock` (`claude-design-import.ts:91`) and `gestureBlock` (`:122`) both match on exact 4-space indentation, exact comment prose, and exact arrow-function/closing-brace shape. If a future Claude Design export switches to tabs, rephrases a comment, or breaks `onGestureEnd` across multiple lines, `wheelBlock.test(source)` returns `false`, `normalizeDesignCanvasWheelHandling` returns the original `source`, and the importer writes the unpatched `design-canvas.jsx` verbatim. The user sees the zoom-on-scroll bug back with no log line, no warning, and no failing test — only a regression report.

## What changed

- `apps/daemon/src/claude-design-import.ts`: the rewritten Cmd+wheel path now detects notched wheels via `deltaMode !== 0 || (deltaX === 0 && Number.isInteger(deltaY) && |deltaY| >= 40)` and uses the original `~17%` ratio for those, while trackpad pinch keeps the `~1%` smooth ratio. The non-Cmd path (`panByWheel`) is untouched, so a plain two-finger scroll without Cmd still pans rather than zooms — the original #1726 fix.
- `apps/daemon/src/claude-design-import.ts`: `normalizeImportedClaudeDesignFile` now emits a `console.warn` when the basename is `design-canvas.jsx` but neither rewrite regex matched. Gives ops/support something to grep when the symptom reappears.
- `apps/daemon/tests/claude-design-import.test.ts`: existing positive test now asserts the new notched-zoom shape (`isNotchedWheel`, both `Math.exp` branches present). New negative test feeds a tab-indented / rephrased-comment `design-canvas.jsx` and asserts the importer leaves the source byte-for-byte intact **and** logs the warning.

## Surface area

- [ ] **UI** — new page / dialog / panel / menu item / setting / empty state in `apps/web` or `apps/desktop`
- [ ] **i18n keys** — added new translation keys
- [ ] **New top-level dependency** — adding any new entry to the **root** `package.json`
- [x] **Default behavior change** — Cmd + mouse wheel on imported Claude Design canvases now scales ~17% per notch again instead of ~63%. Trackpad pinch and non-Cmd scroll are unchanged.

## Validation

- `pnpm --filter @open-design/daemon exec vitest run tests/claude-design-import.test.ts` — 6/6 pass (5 pre-existing + 1 new drift-fixture test)

## Test plan

- [ ] Reimport a Claude Design canvas and confirm Cmd + physical mouse-wheel zoom feels like the original Claude export (~17% per click), not the collapsed 63%
- [ ] Confirm Cmd + trackpad pinch behavior is unchanged
- [ ] Confirm a plain two-finger trackpad scroll (no Cmd) still pans, doesn't zoom — i.e. #1726's original fix is still in effect
- [ ] Confirm the new negative-fixture test fails closed if the regex drifts further

### DIFF
--- a/apps/daemon/src/claude-design-import.ts
+++ b/apps/daemon/src/claude-design-import.ts
@@ -84,7 +84,21 @@ function normalizeImportedClaudeDesignFile(relPath: string, body: Buffer): Buffe
   if (path.basename(relPath) !== 'design-canvas.jsx') return body;
   const source = body.toString('utf8');
   const normalized = normalizeDesignCanvasWheelHandling(source);
-  return normalized === source ? body : Buffer.from(normalized, 'utf8');
+  if (normalized === source) {
+    // We already confirmed via basename that this came from a Claude Design
+    // canvas export, so reaching this branch means neither rewrite regex
+    // matched. Almost always that is because Anthropic touched the canvas
+    // template (whitespace, comment prose, statement layout); the importer
+    // will then silently write the unpatched source and the imported canvas
+    // will reproduce the zoom-on-scroll bug this normalizer exists to
+    // prevent. Log loudly so operators have something to grep before the
+    // bug report comes back in.
+    console.warn(
+      '[claude-design-import] design-canvas.jsx found but wheel-handler regex did not match; imported canvas may zoom on scroll. Update normalizeDesignCanvasWheelHandling to match the new template.',
+    );
+    return body;
+  }
+  return Buffer.from(normalized, 'utf8');
 }
 
 function normalizeDesignCanvasWheelHandling(source: string): string {
@@ -108,12 +122,24 @@ function normalizeDesignCanvasWheelHandling(source: string): string {
       apply();
     };
 
+    // Cmd+wheel still zooms, but we have to split notched mouse wheels from
+    // smooth trackpad pinch deltas inside the Cmd branch: a single mouse
+    // notch arrives as deltaY≈100, and Math.exp(-100*0.01)≈0.367 would shrink
+    // the canvas by ~63% per click. The notched ratio Math.exp(-sign*0.18)
+    // gives ~17% per click — the same feel the original Claude export had
+    // before this normalizer collapsed both paths.
+    const isNotchedWheel = (e) =>
+      e.deltaMode !== 0 ||
+      (e.deltaX === 0 && Number.isInteger(e.deltaY) && Math.abs(e.deltaY) >= 40);
     const onWheel = (e) => {
       e.preventDefault();
       e.stopPropagation();
       if (isGesturing) return;
       if (e.metaKey) {
-        zoomAt(e.clientX, e.clientY, Math.exp(-e.deltaY * 0.01));
+        const factor = isNotchedWheel(e)
+          ? Math.exp(-Math.sign(e.deltaY) * 0.18)
+          : Math.exp(-e.deltaY * 0.01);
+        zoomAt(e.clientX, e.clientY, factor);
         return;
       }
       panByWheel(e);
--- a/apps/daemon/src/claude-design-import.ts
+++ b/apps/daemon/src/claude-design-import.ts
@@ -83,27 +83,42 @@ export async function importClaudeDesignZip(zipPath: string, projectDir: string)
 function normalizeImportedClaudeDesignFile(relPath: string, body: Buffer): Buffer {
   if (path.basename(relPath) !== 'design-canvas.jsx') return body;
   const source = body.toString('utf8');
-  const normalized = normalizeDesignCanvasWheelHandling(source);
-  if (normalized === source) {
-    // We already confirmed via basename that this came from a Claude Design
-    // canvas export, so reaching this branch means neither rewrite regex
-    // matched. Almost always that is because Anthropic touched the canvas
-    // template (whitespace, comment prose, statement layout); the importer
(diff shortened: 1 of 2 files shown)


lefarcen (commented):
Hey @lefarcen — the numeric regression details and validation are clear. Before pool/product review, could you add a short `## What users will see` section that states the user-facing experience in one place (Cmd + mouse-wheel zoom feels like the original Claude export again; trackpad pinch and non-Cmd scroll stay unchanged)? That will help release verification distinguish the behavior change from the implementation notes.

apps/daemon/src/claude-design-import.ts
```diff
@@ -108,12 +122,24 @@ function normalizeDesignCanvasWheelHandling(source: string): string {
...
       if (isGesturing) return;
       if (e.metaKey) {
-        zoomAt(e.clientX, e.clientY, Math.exp(-e.deltaY * 0.01));
+        const factor = isNotchedWheel(e)
+          ? Math.exp(-Math.sign(e.deltaY) * 0.18)
+          : Math.exp(-e.deltaY * 0.01);
+        zoomAt(e.clientX, e.clientY, factor);
         return;
       }
       panByWheel(e);
```
mrcfps: Blocking: this branch now zooms only on `e.metaKey`, but the handler we are rewriting treats `(e.ctrlKey || e.metaKey)` as zoom intent because browsers set `ctrlKey` during trackpad pinch. After this rewrite, smooth pinch events on Chromium/Firefox fall through to `panByWheel(e)` instead of zooming, and the gesture* block below has also been reduced to no-ops, so there is no remaining fallback to preserve the old pinch behavior. That breaks the `Cmd + trackpad pinch behavior is unchanged` guarantee in the PR body.

Suggested change: keep the old `ctrlKey` fallback in the zoom branch (for example, treat `e.ctrlKey || e.metaKey` as zoom intent while still using `isNotchedWheel(e)` to choose the factor) and add a regression test that exercises a smooth `ctrlKey` wheel event so this does not slip again.

apps/daemon/src/claude-design-import.ts
```diff
@@ -84,7 +84,21 @@ function normalizeImportedClaudeDesignFile(relPath: string, body: Buffer): Buffe
...
+    // matched. Almost always that is because Anthropic touched the canvas
+    // template (whitespace, comment prose, statement layout); the importer
+    // will then silently write the unpatched source and the imported canvas
+    // will reproduce the zoom-on-scroll bug this normalizer exists to
+    // prevent. Log loudly so operators have something to grep before the
+    // bug report comes back in.
+    console.warn(
+      '[claude-design-import] design-canvas.jsx found but wheel-handler regex did not match; imported canvas may zoom on scroll. Update normalizeDesignCanvasWheelHandling to match the new template.',
+    );
+    return body;
```
mrcfps: Non-blocking: this warning only fires when `normalizeDesignCanvasWheelHandling(source) === source`, which means the wheel rewrite missed entirely. If `wheelBlock` still matches but `gestureBlock` drifts, `normalizedWheel.replace(gestureBlock, ...)` becomes a no-op, `normalized !== source`, and we silently ship a half-rewritten canvas with the old gesture handler still active. That leaves one of the two template-drift cases described in this PR unreported.

Suggested change: return match metadata from `normalizeDesignCanvasWheelHandling` (or split the two replacements here) and warn whenever either expected block fails to match, then add a fixture where only the gesture block drifts.

mrcfps (changes requested):
@lefarcen Thanks for the fast follow-up here. I found one user-facing regression in the new wheel normalization, plus one smaller drift-detection gap worth tightening before this lands.


mrcfps (approved):
@lefarcen I re-reviewed the updated daemon import normalization and verified the follow-up fixes cover the two regressions called out on the prior round: the rewritten wheel handler now preserves Chromium/Firefox ctrlKey pinch zoom, and the importer now warns on both full and gesture-only regex drift. I also reran `pnpm --filter @open-design/daemon exec vitest run tests/claude-design-import.test.ts` and `pnpm --filter @open-design/daemon typecheck` locally on this head — both passed. Thanks for turning the follow-up quickly and for tightening the drift coverage. 🙏

PR #1820, opened 2026-05-15
fix(web): keep picker hint clear of comments panel

mturac:
## Summary

Fixes #1816.

Keeps the Inspect/Picker hint banner out of the open Comments side panel footprint in comment mode, so the hint and its close action stay usable. When the panel is collapsed, the hint returns to its normal placement.

## Surface area

- [x] UI
- [ ] Daemon/API
- [ ] CLI
- [ ] Tests only
- [ ] Docs only

## Testing

- `git diff --check`
- `corepack pnpm --filter @open-design/web test -- FileViewer.test.tsx` *(blocked locally: this checkout has no `node_modules`, so `vitest` is not installed; local Node is `v25.9.0` while the repo expects `~24`)*

## Screenshots

Not attached: local app dependencies are not installed in this checkout, so I could not boot the UI for a reliable screenshot.

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -6214,7 +6214,12 @@ function HtmlViewer({
               && openHintBox
               && !activeInspectTarget
               && !activeCommentTarget ? (
-              <div className="inspect-empty-hint-container">
+              <div
+                className={`inspect-empty-hint-container${
+                  boardMode && !commentSidePanelCollapsed ? ' comment-side-panel-open' : ''
+                }`}
+                data-testid="inspect-empty-hint-container"
+              >
                 {liveCommentTargets.size === 0 ? (
                   <div
                     className="inspect-empty-hint"
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -10714,6 +10714,20 @@ button.ghost.mcp-copy-btn:hover:not(:disabled) {
   pointer-events: none;
 }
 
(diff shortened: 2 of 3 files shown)


lefarcen (commented):
Hey @mturac, thanks for tying this directly back to #1816 — the current/expected behavior is easy to understand, and the Testing notes are specific. One quick PR-body cleanup before pool review: could you add the new `## Surface area` checklist and tick **UI**? This changes the picker hint placement in the editor, so that surface marker will help release review catch the user-visible scope.

PerishCode (approved):
@mturac thanks for the focused fix on #1816 — the change is tight and the rationale reads cleanly.

What I verified:
- `FileViewer.tsx` adds the `comment-side-panel-open` class only when `boardMode && !commentSidePanelCollapsed`, which is the exact window where the right-edge comment side panel is on screen and would otherwise sit under the hint.
- The new CSS shifts the hint to `right: 340px` (320px panel + 8px panel offset + 12px gap) and caps the max-width with `calc(100% - 368px)`, leaving symmetric breathing room from the left edge.
- The `<= 760px` breakpoint re-anchors the hint to the top with `left/right: 14px` so it doesn't get pushed off-screen on narrow viewports.
- The Vitest case toggles board mode, asserts the class is present while the panel is open, hides the panel via the existing button, and asserts the class is removed — exactly the state machine the fix introduces.

Nice, surgical change with the right amount of test coverage for the surface area touched. Appreciate the inline comment context and the explicit `data-testid` on the hint container — it makes this easy to keep honest going forward.

PR #1835, opened 2026-05-15
fix: format smoke prompt as JSONL for claude-stream-json input format

### LINKED ISSUE #1834: Bug: Claude Code smoke test fails with --input-format stream-json
### What happened?

When using Claude Code 2.1.142 as the local CLI, the connection test always fails with:

"Could not start Claude Code: exit 1 · stderr: Error parsing streaming input line: Reply with only: ok: SyntaxError: JSON Parse error: Unexpected identifier "Reply"."

### Steps to reproduce

1. Install open-design from source (git clone + pnpm install)
2. Have Claude Code 2.1.142 installed and authenticated via OAuth (Pro plan)
3. Run: pnpm tools-dev run web
4. Open Settings → Configure execution mode
5. Claude Code is detected but the connection test fails immediately

### Expected behavior

The connection test should pass and Claude Code should be usable as the local CLI engine. After manually fixing connectionTest.ts (see below), the test passes with "Claude Code replied in 4511 ms — 'ok'".

Root cause: connectionTest.ts sends the smoke prompt as plain text to stdin, but Claude Code is spawned with --input-format stream-json which requires JSONL-formatted input. The fix is:

child.stdin.end(
  def.streamFormat === 'claude-stream-json'
    ? JSON.stringify({ type: 'user', message: { role: 'user', content: SMOKE_PROMPT } }) + '\n'
    : SMOKE_PROMPT,
  'utf8'
);

### Open Design version

0.7.0

### Platform

macOS (Apple Silicon)

```shell

```

### Additional context

_No response_

jannismarvin:
[connectionTest.ts was sending the smoke prompt as plain text, but Claude Code is spawned with --input-format stream-json which requires JSONL input.

Fixes #

## Why

The connection test in connectionTest.ts sends the smoke prompt as plain text to stdin. However, Claude Code is spawned with --input-format stream-json, which requires JSONL-formatted input. This causes the smoke test to always fail with:

"Could not start Claude Code: exit 1 · stderr: Error parsing streaming input line: Reply with only: ok: SyntaxError: JSON Parse error: Unexpected identifier "Reply"."

The main run path in server.ts already handles this correctly (line 4630+).

Fixes #1834

## What users will see

The Claude Code connection test passes instead of failing. The green confirmation "Claude Code replied in Xms — 'ok'" appears in Settings → Configure execution mode.

## Surface area

## Bug fix verification

-

## Validation

-
](fix: format smoke prompt as JSONL for claude-stream-json input format)

### DIFF
--- a/apps/daemon/src/connectionTest.ts
+++ b/apps/daemon/src/connectionTest.ts
@@ -1333,7 +1333,12 @@ async function testAgentConnectionInternal(
           });
         }
       });
-      child.stdin.end(SMOKE_PROMPT, 'utf8');
+      child.stdin.end(
+  def.streamFormat === 'claude-stream-json'
+    ? JSON.stringify({ type: 'user', message: { role: 'user', content: SMOKE_PROMPT } }) + '\n'
+    : SMOKE_PROMPT,
+  'utf8'
+);
     }
     const cancellationPromise = new Promise<{ kind: 'timeout' } | { kind: 'aborted' }>((resolve) => {
       timer = setTimeout(() => resolve({ kind: 'timeout' }), agentTimeoutMs());


lefarcen (commented):
Hi @jannismarvin! The failure mode is clearly described, and tying it to Claude Code's `stream-json` input is helpful for reviewers. Before pool review, could you fill in the empty `## Why` / `## What users will see` sections and tick the relevant `## Surface area` box? For this one, also worth linking it to #1834 directly (for example `Fixes #1834`) since that's the issue carrying the repro/claim context.

Related: #1834 tracks the same smoke-test failure and currently has an active claim, so please coordinate there to avoid duplicate work.

lefarcen:
Heads-up: PR #1844 is also open against this same Claude Code `stream-json` smoke-test failure. Both PRs touch `apps/daemon/src/connectionTest.ts`; #1844 also adds regression coverage in `apps/daemon/tests/connection-test.test.ts`.

@jannismarvin and @mturac may want to compare approaches; the maintainer team will pick which lands. Sharing so neither effort gets wasted.

mrcfps:
Thanks so much for the contribution here 💙 PR #1844 has already addressed this issue, and it comes with a more complete regression test as well, so I’m going to close this one for now. Really appreciate you taking the time to put this together — looking forward to your next PR! ✨

PR #1794, opened 2026-05-15
feat(design-systems): add structured tokens for notion, linear, github, figma, slack, discord, openai, shopify, spotify, uber

zoeforfun:
## Why

This is the second batch of hand-authored oracle fixtures for the Design System Token Channel project (follow-on to #1652). The first batch provided 5 brands (cursor, apple, stripe, airbnb, vercel) as proof-of-concept oracles for the future bulk-derivation script (PR-B). This PR adds 10 more well-documented brands that agents and the derive script can validate against.

User pain addressed: agents currently invent brand-variant token names like `--notion-cream` or pick wrong weight values because the only source of truth is prose in `DESIGN.md`. Hand-authored `tokens.css` + `components.html` fixtures lock in the correct translations once, letting agents paste the `:root` block verbatim.

## What users will see

No visible product change. Design panel agents get a richer oracle set — 12 structured brands instead of 2 — so generated artifacts using notion, linear, github, figma, slack, discord, openai, shopify, spotify, or uber tokens will match the official brand specification more accurately.

## Changes

Added `tokens.css` and `components.html` for 10 brands:

| Brand | Visual theme | Accent | Notable schema decision |
|-------|-------------|--------|------------------------|
| notion | Warm white, near-black text | `#0075de` | `--radius-sm: 4px`; multi-layer shadow (max opacity 0.04) |
| linear-app | Dark #08090a, achromatic | `#5e6ad2` | weight-510; semi-transparent white borders |
| github | True white, hairline borders | `#0969da` | `--text-base: 14px` (density identity) |
| figma | Strict B&W, figmaSans variable | `#000000` | `--radius-sm: 50px` (pill); dashed focus |
| slack | White + aubergine sidebar | `#4a154b` | `--radius-sm: 4px`; logo palette = semantic only |
| discord | Dark #313338, 3-step depth | `#5865f2` | luminance stepping; gg sans weight 510 |
| openai | White, Signifier serif display | `#10a37f` | two font voices; `--radius-sm: 12px` |
| shopify | Void black, NeueHaasGrotesk | `#36f4a4` | `--radius-sm: 9999px`; ultra-light weight 330 |
| spotify | Near-black #121212 | `#1ed760` | `--radius-sm: 9999px`; uppercase button labels |
| uber | Pure B&W, UberMove bold | `#000000` | `--radius-sm: 999px`; whisper shadow 0.12 |

## Validation

```
pnpm guard
```

[...]

PR #1841, opened 2026-05-15
feat(design-systems): add tokens.css + components.html for 9 AI product brands

zoeforfun:
## Why

The design-system library ships `tokens.css` + `components.html` as the two agent-consumable artifacts that let an AI produce on-brand HTML without needing to reason from prose. This batch covers the 9 AI-product brands that were most requested for the repo but had only `DESIGN.md` with no compiled artifacts. (`openai` was compiled and merged separately before this PR.)

The concrete user pain: when an agent is asked to generate UI in the style of, e.g., ElevenLabs or NVIDIA, it currently has no structured token reference to draw from, so it produces generic output. These files close that gap.

## What users will see

No product UI change. Design-system consumers (skills, agents, external tooling) gain structured token files for:

- **claude** — warm parchment canvas, terracotta accent, Anthropic Serif display font
- **cohere** — pure white, 22 px signature radius, CohereText + Unica77 type pairing
- **elevenlabs** — achromatic + warm stone undertones, Waldenburg weight 300, pill-shaped elements
- **huggingface** — IBM Plex Mono for all headings/tags, HF Yellow (#ffd21e) accent, Source Sans Pro body
- **mistral-ai** — golden amber palette, Arial weight 400 across all tiers, near-zero radius
- **nvidia** — true black canvas, NVIDIA Green (#76b900) as border/outline signal only (never fill), teal (#1eaedb) hover
- **opencode-ai** — Berkeley Mono as the sole typeface, flat depth (ring-only elevation), Apple HIG semantic colors, 880 px narrow column
- **replicate** — pill-shaped everything (border-radius 9999 px), rb-freigeist-neue heavy, Replicate Red accent, dotted underlines
- **together-ai** — "The Future" font with −0.03 em tracking, pure-white / midnight-blue (#010120) dual-world sections, lavender soft accent

Each `components.html` embeds the full `:root` token block inline so agents can copy-paste it into any `<style>` tag with zero external dependency.

## Surface area

- [x] `design-systems/` — 9 new brand pairs (tokens + components)
- [ ] `apps/`
- [ ] `packages/`
- [ ] `tools/`
- [ ] `e2e/`
- [ ] Skills / design-templates / craft
- [ ] CLI flags / env vars / i18n keys
- [ ] New root `package.json` dependencies

### DIFF
--- a/design-systems/opencode-ai/components.html
+++ b/design-systems/opencode-ai/components.html
@@ -1,0 +1,389 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <title>OpenCode AI — reference components</title>
+    <meta
+      name="description"
+      content="Reference fixture for design-systems/opencode-ai. Terminal-native dark:
+        warm near-black canvas, Berkeley Mono sole typeface, flat depth (no shadows),
+        Apple HIG semantic colors, 880px narrow reading column."
+    />
+
+    <style>
+      :root {
+        --bg: #201d1d;
+        --surface: #302c2c;
+        --surface-warm: var(--surface);
+
+        --fg: #fdfcfc;
+        --fg-2: #c8c6c4;
+        --muted: #9a9898;
+        --meta: #6e6e73;
+
+        --border: #464343;
+        --border-soft: #302c2c;
+
+        --accent: #007aff;
+        --accent-on: #ffffff;
+        --accent-hover: #0056b3;
+        --accent-active: #004085;
+
+        --success: #30d158;
+        --warn: #ff9f0a;
+        --danger: #ff3b30;
+
+        /* Berkeley Mono is the ONLY font — no typographic variation */
+        --font-display: "Berkeley Mono", "IBM Plex Mono", ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
+        --font-body: "Berkeley Mono", "IBM Plex Mono", ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
+        --font-mono: "Berkeley Mono", "IBM Plex Mono", ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
+
+        --text-xs: 12px;
+        --text-sm: 14px;
+        --text-base: 16px;
+        --text-lg: 18px;
+        --text-xl: 22px;
+        --text-2xl: 28px;
+        --text-3xl: 38px;
+        --text-4xl: 48px;
+
+        --leading-body: 1.5;
+        --leading-tight: 1.0;
+        --tracking-display: 0em;
+
+        --space-1: 4px;
+        --space-2: 8px;
+        --space-3: 12px;
+        --space-4: 16px;
+        --space-5: 20px;
+        --space-6: 24px;
+        --space-8: 32px;
+        --space-12: 48px;
+
+        --section-y-desktop: 80px;
+        --section-y-tablet: 64px;
+        --section-y-phone: 48px;
+
+        --radius-sm: 4px;
+        --radius-md: 6px;
+        --radius-lg: 8px;
+        --radius-pill: 9999px;
+
+        --elev-flat: none;
+        --elev-ring: 0 0 0 1px var(--border);
+        --elev-raised: 0 0 0 2px var(--border);
+
+        --focus-ring: 0 0 0 2px var(--accent);
+
+        --motion-fast: 100ms;
+        --motion-base: 150ms;
+        --ease-standard: cubic-bezier(0.2, 0, 0, 1);
+
+        --container-max: 880px;   /* Narrow reading column — "everything is code" */
+        --container-gutter-desktop: 24px;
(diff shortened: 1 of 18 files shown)


lefarcen (commented):
Hey @zoeforfun, the brand coverage is nicely scoped and the per-brand user-facing notes make it easy to understand what this batch adds. One PR-body cleanup before pool review: could you add a `## Validation` section with what you checked, and align the `## Surface area` checklist with the repo template by ticking the closest existing bucket for these design-system artifacts (or noting why only `design-systems/` applies)?

Related: #1794 is the prior merged structured-token batch that included OpenAI; this PR looks like the follow-up brand batch rather than an overlapping open duplicate.

lefarcen (commented):
Duplicate of #pullrequestreview-4298088465 — please use the first body-lint note above.

lefarcen (commented):
Duplicate of #pullrequestreview-4298088465 — please use the first body-lint note above.

design-systems/opencode-ai/components.html
```diff
@@ -0,0 +1,389 @@
...
+        --accent-active: #004085;
+
+        --success: #30d158;
+        --warn: #ff9f0a;
+        --danger: #ff3b30;
+
+        /* Berkeley Mono is the ONLY font — no typographic variation */
+        --font-display: "Berkeley Mono", "IBM Plex Mono", ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
+        --font-body: "Berkeley Mono", "IBM Plex Mono", ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
+        --font-mono: "Berkeley Mono", "IBM Plex Mono", ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, monospace;
```
mrcfps: This `:root` copy has drifted from the canonical token file: `design-systems/opencode-ai/tokens.css:58-60` includes `"Liberation Mono"` in all three Berkeley Mono fallback stacks, but this fixture copy drops it. That matters because the sync guard compares the first `:root` blocks byte-for-byte and already fails on this head, and it also means agents copying `tokens.css` would get a slightly different fallback stack than the reference fixture demonstrates. Please re-paste the font declarations from `tokens.css` into this file so the three font tokens stay identical.

design-systems/together-ai/components.html
```diff
@@ -0,0 +1,425 @@
...
+        --border: #ebebeb;
+        --border-soft: var(--border);
+
+        --accent: #010120;
+        --accent-on: #ffffff;
+        --accent-hover: color-mix(in oklab, var(--accent), white 8%);
+        --accent-active: color-mix(in oklab, var(--accent), white 16%);
+
+        /* Lavender — soft secondary UI tint (not illustration-only) */
+        --lavender: #bdbbff;
```
mrcfps: This inline `:root` introduces `--lavender`, but `design-systems/together-ai/tokens.css` does not declare that token at all. That split-brains the brand contract: the fixture later relies on `var(--lavender)`, while the canonical token channel agents copy from `tokens.css` would never see it, and the sync guard is already failing on this exact drift. Please either promote `--lavender` into `tokens.css` (and add it to `design-systems/_schema/tokens.schema.ts`'s `BRAND_EXTENSIONS` allowlist if it stays Together-specific) or replace the component-only uses with literal color values, then re-paste the canonical `:root` block here.

mrcfps (changes requested):
Thanks for the big brand-artifact batch — I found two concrete token/fixture sync regressions that are breaking `pnpm guard` on this head, so I'm requesting changes until the canonical `tokens.css` blocks and inline fixture `:root` blocks are back in sync.


design-systems/together-ai/components.html
```diff
@@ -0,0 +1,422 @@
...
+        border-color: rgba(255, 255, 255, 0.1);
+      }
+      .section-dark .eyebrow { color: rgba(255, 255, 255, 0.5); }
+      .section-dark p { color: rgba(255, 255, 255, 0.7); }
+      .section-dark .card {
+        background: rgba(255, 255, 255, 0.05);
+        border-color: rgba(255, 255, 255, 0.1);
+      }
+      .section-dark .btn-primary {
+        background: var(--lavender);
```
mrcfps, line 230: `var(--lavender)` is used here, but this file's pasted `:root` block never declares `--lavender`, and `design-systems/together-ai/tokens.css` does not declare it either on this head. That matters because an undefined custom property invalidates the whole declaration: in the dark section, `.btn-primary` falls back to its earlier `background: var(--accent)`, which is the same `#010120` as the section background, so the shipped reference CTA loses the lavender contrast the fixture text promises. The same missing token also drops the secondary hover border and the badge color later in this file. Please either add `--lavender: #bdbbff` to the canonical token set (and allowlist it if it stays Together-specific), then re-paste the synced `:root` here, or replace these `var(--lavender)` call sites with the literal lavender value already used elsewhere in the fixture.

mrcfps (changes requested):
Thanks for pushing the nine-brand artifact batch — I found one concrete Together AI token-contract break that leaves the shipped fixture demonstrating the wrong dark-section CTA styling, so I'm requesting one fix before merge.


mrcfps (approved):

PR #1844, opened 2026-05-15
fix(daemon): wrap Claude smoke test stdin as stream-json

### LINKED ISSUE #1834: Bug: Claude Code smoke test fails with --input-format stream-json

mturac:
Fixes #1834

## Why

I picked this up after the Claude Code connection smoke test was reported failing when the adapter launches with `--input-format stream-json`. The Settings connection test was still writing the smoke prompt as plain text, so Claude tried to parse `Reply with only: ok` as JSONL and exited before users could confirm their local CLI setup.

## What users will see

Settings connection tests for Claude Code can now complete when Claude is configured with stream-json stdin. A working local Claude Code install should report the smoke-test success instead of failing with a JSON parse error.

## Surface area

- [ ] **Extension point** — new entry under `skills/`, `design-systems`, `design-templates`, or `craft/`, or change to the skills protocol

## Screenshots

Not applicable; this fixes daemon-side connection-test behavior without changing UI layout.

## Bug fix verification

- Test path that reproduces the bug: `apps/daemon/tests/connection-test.test.ts`
- Did the test go red on `main` and green on this branch? The new regression encodes the failing stdin shape from #1834: fake Claude parses stdin as stream-json JSONL and would fail if plain text is written. It passes on this branch after the daemon wraps the smoke prompt according to `promptInputFormat`.

## Validation

- `pnpm guard`
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/connection-test.test.ts`
- `pnpm --dir apps/daemon exec vitest run -c vitest.config.ts tests/connection-test.test.ts -t 'wraps Claude connection smoke prompts as stream-json stdin'`

### DIFF
--- a/apps/daemon/src/connectionTest.ts
+++ b/apps/daemon/src/connectionTest.ts
@@ -40,6 +40,7 @@ import {
   probeAgentAuthStatus,
 } from './runtimes/auth.js';
 import type { AgentCliEnvPrefs } from './app-config.js';
+import type { RuntimeAgentDef } from './runtimes/types.js';
 import {
   isLoopbackApiHost,
   validateBaseUrl,
@@ -110,6 +111,23 @@ const SAMPLE_MAX_CHARS = 120;
 // before producing a visible `ok`.
 const PROVIDER_MAX_TOKENS = 100;
 const SMOKE_PROMPT = 'Reply with only: ok';
+
+function formatPromptForAgentStdin(
+  def: Pick<RuntimeAgentDef, 'promptInputFormat'>,
+  prompt: string,
+): string {
+  const promptInputFormat = def.promptInputFormat ?? 'text';
+  if (promptInputFormat === 'stream-json') {
+    return `${JSON.stringify({
+      type: 'user',
+      message: {
+        role: 'user',
+        content: [{ type: 'text', text: prompt }],
+      },
+    })}\n`;
+  }
+  return prompt;
+}
 
 function codexExecutableGuidance(
   agentId: string,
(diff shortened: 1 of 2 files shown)


lefarcen:
Heads-up: PR #1835 is also open against this same Claude Code `stream-json` smoke-test failure. Both PRs touch `apps/daemon/src/connectionTest.ts`; this one also adds regression coverage in `apps/daemon/tests/connection-test.test.ts`.

@jannismarvin and @mturac may want to compare approaches; the maintainer team will pick which lands. Sharing so neither effort gets wasted.

mturac:
Thanks for the heads-up and for linking the duplicate context 🙏 I’ll keep an eye on this and am happy to adjust or close this one if maintainers prefer the other direction.

Issue #1777, opened 2026-05-15
Left and right arrow keys do not turn pages in fullscreen mode in 0.8.0


shangxinyu1:
## Summary

In fullscreen presentation mode, the left and right arrow keys do not turn pages in Open Design 0.8.0.

## Version

Observed on 0.8.0.

## Environment

- Platform: macOS
- Surface: Fullscreen presentation / slide preview mode
- Build: 0.8.0

## Current behavior

After entering fullscreen mode, pressing the left and right arrow keys does not move between pages.

## Expected behavior

In fullscreen mode, the left and right arrow keys should navigate to the previous and next page.

## Steps to reproduce

1. Open Open Design 0.8.0.
2. Open a multi-page deck or presentation preview.
3. Enter fullscreen mode.
4. Press the left and right arrow keys.
5. Observe that page navigation does not happen.

## Regression?

Unknown. Needs confirmation against 0.7 stable or an earlier 0.8.0 build.

## Notes

- This appears to be a functional issue, not just a UX polish issue.
- Keyboard navigation is especially important in fullscreen presentation scenarios.
- Screenshot shows fullscreen deck view where page turning is expected to work via keyboard.

<img width="3024" height="1890" alt="Image" src="https://github.com/user-attachments/assets/af2be278-d9ee-44dd-b35e-4a041646f2af" />

https://github.com/user-attachments/assets/3df279f8-96a6-4a2f-a945-2171609c8ad4

lefarcen:
Claimed by @prantikmedhi. Thanks for picking this up! 🎉

**Quick question to time check-ins right** — roughly how long do you think this will take?

- A few hours / today
- 1-3 days
- ~ 1 week
- Longer / unsure

Reply with whatever fits. If you don't reply, I'll default to checking in every ~12h. Either way, any comment, linked PR, or progress note keeps the claim active.

prantikmedhi:
@lefarcen i need a few hours i guess

shangxinyu1:
@prantikmedhi thanks for picking this up.

One important branch note for this fix: please open the PR against `preview/v0.8.0`, not `main`.

Target branch:
https://github.com/nexu-io/open-design/tree/preview/v0.8.0

Since this issue is tracked under the `preview/v0.8.0` line, it would be best to land the implementation and PR there first.

lefarcen:
Hey @prantikmedhi, sounds good — here's what I dug up in the source so you can hit the ground running.

**Working hypothesis — root cause**

[...]

PR #1858, opened 2026-05-15
fix(web): restore fullscreen deck navigation

prantikmedhi:
Addresses #1777.

- add a capture-phase key handler inside the deck iframe so ArrowLeft/Right, PageUp/Down, Home, and End work in fullscreen mode
- keep the iframe bridge from double-firing when the deck has its own keyboard handler
- add regression coverage for arrow-key navigation inside the iframe bridge

### DIFF
--- a/apps/web/src/runtime/srcdoc.ts
+++ b/apps/web/src/runtime/srcdoc.ts
@@ -1306,6 +1306,28 @@ function injectDeckBridge(doc: string, initialSlideIndex = 0): string {
   }
   ownDeckButton('deck-prev', 'prev');
   ownDeckButton('deck-next', 'next');
+  document.addEventListener('keydown', function(e){
+    var target = e.target;
+    var tag = target && target.tagName;
+    if (tag === 'INPUT' || tag === 'TEXTAREA' || (target && target.isContentEditable)) return;
+    if (e.key === 'ArrowRight' || e.key === 'PageDown') {
+      e.preventDefault();
+      e.stopImmediatePropagation();
+      go('next');
+    } else if (e.key === 'ArrowLeft' || e.key === 'PageUp') {
+      e.preventDefault();
+      e.stopImmediatePropagation();
+      go('prev');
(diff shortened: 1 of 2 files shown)


lefarcen (commented):
Hey @prantikmedhi, thanks for picking up #1777 and writing the implementation notes so clearly. Before this goes into pool review, could you fill in the PR template context: **Why**, **What users will see**, **Surface area**, and **Validation**? Also, #1777 is labeled `preview/v0.8.0` and the issue thread asked for that branch first, so please retarget or confirm the base branch if `main` is intentional. I requested @nettee for review; once the description/base are sorted, that should make the review path much smoother.

apps/web/src/runtime/srcdoc.ts
```diff
@@ -1306,6 +1306,28 @@ function injectDeckBridge(doc: string, initialSlideIndex = 0): string {
...
+      go('prev');
+    } else if (e.key === 'Home') {
+      e.preventDefault();
+      e.stopImmediatePropagation();
+      go('first');
+    } else if (e.key === 'End') {
+      e.preventDefault();
+      e.stopImmediatePropagation();
+      go('last');
+    }
```
nettee, line 1329: This capture-phase handler recurses through the bridge's own fallback path for decks that do **not** expose mutable `.slide` state. In that case `go('next')` falls back to `dispatchKey('ArrowRight')`, and `dispatchKey()` emits another `keydown` on `document`, which re-enters this same listener before the deck's own keyboard handler can see it. The result is an infinite loop / stack overflow instead of forwarding the key in fullscreen mode, and the new test does not cover that shape because its fixture already has an `active` class so `setActive()` returns early. Please ignore the bridge's synthetic events here (for example `if (!e.isTrusted) return` or a local reentrancy flag around `dispatchKey()`), then add a regression test that exercises the `dispatchKey()` fallback path.

nettee (changes requested):
I found one blocking issue in the new fullscreen keyboard bridge.

PR #1861, opened 2026-05-15
fix(prompts): stabilize discovery brand answers

### LINKED ISSUE #426: i18n: question-form options stay English when other fields localize
## Summary

Smart agents (Claude / Codex / etc.) **already do** emergent localization on `<question-form>` blocks when the user is conversing in a non-English locale — most labels and option chips arrive translated automatically. The exception is the `brand` field, whose option strings stay English even when every other field flips. This isn't agent laziness — those English literals also appear later in the **same prompt** as branch-matching keys, and a careful agent (correctly) chooses prompt-rule integrity over UX consistency.

The result is a confusing, half-translated form (screenshot below). The deeper observation: today's prompt design forces a tradeoff between localization and branch correctness on the agent's plate — a tradeoff the prompt schema can resolve at zero translation cost.

## Policy alignment

I read [`CONTRIBUTING.md` § Localization maintenance](https://github.com/nexu-io/open-design/blob/main/CONTRIBUTING.md) before writing this. The policy is clear:

> Locale PRs should translate UI chrome, core docs, and display-only gallery metadata in `apps/web/src/i18n/content.ts`, but should not translate `skills/`, `design-systems/`, or prompt bodies that agents execute. Those source prompts are maintained as workflow inputs, and keeping one source language avoids multiplying prompt QA across locales. (…) Daemon errors, export filenames, and **agent-generated artifact text are known limitations unless a PR explicitly scopes them**.

This issue is meant to be the *"PR that explicitly scopes them"* the policy anticipates. The two approaches I'm proposing below are deliberately split along that policy line:

- **Approach 1 (option `id`s) is policy-compliant on its face.** It does *not* translate any prompt body. It changes the `FormQuestion.options` schema so labels and branch keys can finally be different things. The English source prompts stay English; emergent agent localization on the *display* side simply stops being trapped.
- **Approach 2 (locale plumbing) is the explicitly-scoped extension.** It does ask the agent to render visible labels in the user's locale, but **does not translate the source prompts** — only adds one runtime hint. Prompt QA stays single-language. Translation quality is offloaded to the model, not the maintainers.

If the call is "let's not do Approach 2, only Approach 1", that already removes the half-translated-form symptom this issue opens with — Approach 1 alone is a meaningful win.

## Reproduction

1. `pnpm tools-dev` (or a packaged build ≥ `v0.3.0`).
2. Settings → Language → `繁體中文` (or `日本語`, `한국어`, `العربية`, …).
3. Start a new project with a brief in that language — e.g. *"製作一個課程宣傳動畫片段"* (make a course promo animation clip).
4. Wait for the agent's first `<question-form id="discovery">`.

**Expected:** Every form field (including `brand`) renders options in the user's locale.

**Actual:** Most fields are translated by the agent's own emergent localization, but the `brand` options stay English (`Pick a direction for me` / `I have a brand spec — I'll share it` / `Match a reference site / screenshot — I'll attach it`).

> 📎 Screenshot attached. The project title (top, "製作一個課程宣傳動畫片段"), the field labels (`主要觀眾是誰?` / `動畫長度` / `視覺語氣` / `品牌方向` / `最想主打什麼?`), and most option chips (`15 秒短版`, `專業 / 企業感`, `科技 / 工具感`, …) are all in Traditional Chinese — produced by the agent's emergent translation. The `品牌方向 *` field below them shows three options that stayed English, highlighted with a red rectangle.

## Why this is happening (root cause, with evidence)

The agent isn't being lazy. **The prompt actively forbids translating the `brand` options** by using their English literals as branch-matching keys later in the same prompt:

```ts
// packages/contracts/src/prompts/discovery.ts — form definition
{ "id": "brand", "label": "Brand context", "type": "radio",
  "options": [
    "Pick a direction for me",
    "I have a brand spec — I'll share it",
    "Match a reference site / screenshot — I'll attach it"
  ] }

// …later, in the SAME prompt:
### Branch A — `brand: "Pick a direction for me"`
### Branch B — `brand: "I have a brand spec — I'll share it"` or `"Match a reference site / screenshot"`
```

Because option strings double as **(a) display labels (b) the value submitted back (c) the key the prompt branches on**, a localizing agent faces a forced choice:

- **Translate the option →** user-friendly text, but Turn 2 branch matching breaks (the user's Chinese answer no longer matches the English literal in the rule).
- **Keep English →** branch matching works, but UX inconsistency (the form is half-translated, exactly what the screenshot shows).

The agent is solving this by sacrificing UX consistency on the one field where the prompt makes branch keys load-bearing. Other fields (`tone`, `output`, etc.) have **no branch matching downstream**, so the agent freely localizes them.

There is no `locale` parameter anywhere in the prompt builders today (`grep -r 'locale\|userLanguage\|preferredLocale' apps/daemon/src/prompts packages/contracts/src/prompts` returns nothing).

## The same shape recurs across the prompt corpus

I audited all six files under `packages/contracts/src/prompts/` (and their `apps/daemon/src/prompts/` mirrors). Affected user-visible surfaces, by structural shape:

### A. Structured strings — fixable by adding stable `id`s (Approach 1)

| Surface | Where | Risk |
| --- | --- | --- |
| `<question-form>` `options[]` (esp. `brand`) | `discovery.ts` | branch keys force English (this issue's primary symptom) |
| `direction-cards` `label` / `mood` / `references` | `directions.ts` (5 directions × 4 user-facing fields each, ~50–100 chars per blurb) | label doubles as id-ish lookup into `DESIGN_DIRECTIONS`; mood/refs are pure prose |
| TodoWrite item descriptions | implicit (agent-generated, follows English templates from prompts) | shown live in the chat as a "Todos" card |

### B. Semi-structured — needs both id-stabilization and locale plumbing

| Surface | Where | Risk |
| --- | --- | --- |
| `<artifact title="...">` titles | implicit | rendered as artifact card title |
| `brand-spec.md` extraction prose | `discovery.ts` Branch B | persisted file content; agent also narrates it in chat |
| Deck section names (Hero / Problem / Solution / …) | `deck-framework.ts` | agent uses these as slide titles in non-English projects |

### C. Free-form prose narration — only fixable by Approach 2

| Surface | Where | Note |
| --- | --- | --- |
| Turn 1 preamble pattern (*"Got it — pitch deck for a SaaS product, B2B audience. Tell me the rest:"*) | `discovery.ts` line ~36 region | agent imitates the English voicing |
| End-of-turn summary (*"what's there, what's still open, what you'd suggest next"*) | `official-system.ts` line ~29 | every artifact-emitting turn closes with this |
| Image/video/audio kind preamble (*"This is an **image** project. Plan the prompt carefully…"*) | `system.ts` lines ~183 / 201 / 226 | leaks into agent's reasoning the user can see |
| Brand-spec vocalization sentence | `discovery.ts` Branch B step 5 | one summary sentence per project |
| Identity / workflow voice | `official-system.ts` lines 11–22 | the **root cause for tone** — sets the entire voice in English |

### D. Out of scope (worth their own follow-ups)

- Localization of the `<artifact>` HTML body itself (the deliverable's button text, headings).
- Skill-level prompts in `skills/*/SKILL.md` — different system, different issue. (Note: `CONTRIBUTING.md` already states `skills/` should not be translated; consistent with this issue's stance of leaving prompt bodies English.)
- Tool-call summaries (e.g. Bash command echoes) — convention is English, fine to leave.

## Why naive fixes don't work

- **Adding zh-TW / ja / ar entries to `apps/web/src/i18n/locales/`** — the strings aren't there. They live in the prompts.
- **Translating only the prompt's `options[]`** — would violate the `CONTRIBUTING.md` "don't translate prompt bodies" policy *and* break Turn 2 branch matching for the next user (the prompt's `### Branch A — brand: "<localized>"` references go stale instantly; you'd have to ship one prompt-fork per locale).
- **Asking the agent "respond in user's language"** via a single prompt addition — works for capable models on free-text but **doesn't help with branch-matching strings**, because the agent already does this for non-load-bearing fields and is correctly avoiding it for `brand`. It also varies in quality across model strengths (`gemini-flash` and older Codex tend to drift back to English mid-conversation).

## Proposed approaches

These are **complementary, not alternatives.** Different categories above need different tools.

### Approach 1 — Promote `FormQuestion.options` from `string[]` to `Array<{ id; label }>`

**Policy stance:** Does not translate any prompt body. Source prompts stay 100% English. Approach 1 alone resolves Category A.

Schema:

```ts
{ "id": "brand", "label": "Brand context", "type": "radio",
  "options": [
    { "id": "pick",  "label": "Pick a direction for me" },
    { "id": "spec",  "label": "I have a brand spec — I'll share it" },
    { "id": "match", "label": "Match a reference site / screenshot — I'll attach it" }
  ] }
```

Touch points:

- `apps/web/src/artifacts/question-form.ts` — parser; accept both shapes for one minor (legacy `string[]` is treated as `[{ id: opt, label: opt }]`).
- `apps/web/src/components/QuestionForm.tsx` — submit `{ brand: "pick" }` (id), not the label.
- `packages/contracts/src/prompts/discovery.ts` (and `apps/daemon/src/prompts/discovery.ts` mirror) — keep options English, but switch Branch A/B prose to match on `brand === "pick"` (id) instead of label.
- `packages/contracts/src/prompts/directions.ts` — already has stable per-direction `id`s; just confirm `discovery.ts` Branch A's lookup uses the id, not the label.

What this unlocks:

- The agent can safely translate the visible `label` strings on its own — without breaking the contract.
- Future per-locale `label` overrides (an explicit `apps/web/src/i18n/...` lookup keyed by `id`) become possible without re-touching prompts.

### Approach 2 — Add a single `locale` hint through the prompt builder chain

**Policy stance:** Does not translate any prompt body. Adds one runtime parameter and one short instruction. Per-locale prompt QA still does not exist; translation quality lives in the model.

- Web → daemon `/api/agents` request gains a `locale` field (the user's UI locale).
- Prompt builders accept `locale` and weave it into the system prompt:
  > *"The user's UI locale is `{locale}`. Render labels, option `label`s, placeholders, your turn-1 preamble, your end-of-turn summary, and TodoWrite items in `{locale}`. Branch keys (option `id`s) stay English."*
- Plumbing pattern matches `feat(version): add app version awareness` (#204), which already moves new daemon/web data through `/api/*`.

What this unlocks: Category C (free-form narration) follows the user's locale on capable models. Category A's display labels also start localizing more confidently because the agent is now told *to* (rather than guessing).

### Approach 3 — (Stretch / later) Per-locale label table for `DESIGN_DIRECTIONS`

`directions.ts` already has stable `id`s per direction. Each direction also carries ~100 chars of English `label` + `mood` + `references` — quality of agent on-the-fly translation may be uneven for evocative copy like *"Print-magazine feel. Generous whitespace, large serif headlines…"*.

Pragmatic v1: keep the English content as today, rely on Approach 2's runtime hint. Revisit only if real user reports show the on-the-fly translations drifting.

## Acceptance check

A user opens the app in `Deutsch`, starts a project, and:

1. The first `<question-form>` shows **every** label and option in German (including `Markenkontext` options).
2. Submits the form. The Turn-2 branch logic correctly routes to direction-cards / brand-spec extraction / TodoWrite without language mismatches.
3. The agent's preamble (*"Verstanden — Pitch-Deck für ein SaaS-Produkt…"*) and end-of-turn summary are also in German.
4. TodoWrite items render in German.
5. Repeat in `ar` (RTL): chips and labels lay out right-to-left.
6. (Regression check) Switch the same project back to English — the prompt's branch logic still works because keys are language-stable `id`s, not labels.

## Related

- Recently merged: #316 Arabic + RTL; #376 French; #259 dark mode (UI-chrome only — none touched the agent surface).
- Recently merged: #204 `feat(version): add app version awareness` — already plumbs new daemon/web data through `/api/*` endpoints; a `locale` parameter could ride a similar plumbing pattern.
- Existing locale issues asking for new languages — #236 Spanish, #252 Korean — would benefit from Approach 1 the moment new options are added (they'd be id-keyed from day one).

## Offer

I'm happy to take a stab at Approach 1 (the `options` schema migration) if maintainers agree on the direction first. Approach 2 (locale plumbing) deserves a follow-up issue / RFC if Approach 1 lands cleanly. Pinging because the prompt schema change is the kind of thing better aligned before code lands, and because Approach 1 touches `packages/contracts` (the contract layer) where coordination matters most.

---

## Reporter context

- **Environment:** Windows 11 Pro, packaged Open Design `v0.3.0` desktop app and `pnpm tools-dev` against current `main` — same effect on both.
- **Agent CLI:** Codex CLI on Windows. (No CLI-specific behavior; Claude Code BYOK shows the same partial-localization pattern in spot checks.)
- **Project kind:** `blog-post` with a video brief (*"製作一個課程宣傳動畫片段"*).
- **Skill / design system:** No active skill / design system pair on this run (default discovery flow).
- **Daemon stderr:** No errors related to this issue — the form rendering is from agent output, not a daemon error path. Discovery-form emission proceeds normally; only the visible English option chips on `brand` are surprising.
- **Full assistant message:** The agent's discovery-form turn is what the screenshot captures end-to-end. The form JSON inside `<question-form id="discovery">` matches the `discovery.ts` template structure with field-level localization applied to all fields except `brand` (and `brand` matches the verbatim English in the prompt).

Image

YUHAO-corn:
Fixes #426.

## Why

Discovery currently asks the user to choose between using brand source material, matching references, or picking a direction from scratch. The visible answer labels may eventually be localized, but the prompt branch logic was tied to those visible labels. This makes the flow fragile: changing the display copy can change which discovery branch the agent takes.

This PR keeps the user-facing labels separate from stable internal answer values so the branch logic can remain deterministic while labels stay free to localize.

## What users will see

No visible UI change. Users still see the same brand-choice question, but the follow-up discovery behavior stays correct even if the displayed options are translated or reworded.

## Surface area

## Screenshots

N/A — no UI surface changes.

## Bug fix verification

- Test path that reproduces the bug: `apps/daemon/tests/prompts/system.test.ts`, `apps/daemon/tests/system-prompt-template.test.ts`, and `packages/contracts/tests/system-prompt-api-mode.test.ts` cover the generated discovery prompt, stable `brand` values, skip-form/source handling, and active design-system branch behavior.
- Did the test go red on `main` and green on this branch? Not separately run on `main`; the assertions target prompt text and branch rules introduced by this fix and pass on this branch.
- If a red spec wasn't cheap to write, explain why and what verification you did instead: focused prompt regression tests were added for the stable value contract and branch behavior, plus the existing web question-form tests were run to confirm `{ label, value }` options are already submitted as stable values.

## Validation

- `pnpm --filter @open-design/contracts test -- system-prompt-api-mode.test.ts`
- `pnpm --filter @open-design/daemon exec vitest run tests/prompts/system.test.ts tests/system-prompt-template.test.ts -c vitest.config.ts`
- `pnpm --filter @open-design/web exec vitest run tests/artifacts/question-form.test.ts tests/components/QuestionForm.test.tsx -c vitest.config.ts`
- `git diff --check`

## Review notes

This intentionally does not add locale plumbing. It relies on the existing question-form object-option support from #1293: `label` remains display copy, while `value` is the stable answer key used by the next prompt turn.


lefarcen (commented):
Hey @YUHAO-corn, thanks for picking up the prompt-side half of #426 — the Summary makes the stable label/value direction easy to follow.

Before pool review scopes this, could you fill the two PR-template pieces still missing: add the `Surface area` checklist with the relevant prompt/contract/daemon surface marked, and add a `Validation` section? The commands under `Tests` are useful; you can reuse those there so reviewers and release checks can scan the proof quickly.

I’ve routed this to @mrcfps for pool review.

mrcfps (approved):
@YUHAO-corn Thanks for the prompt-side fix here. I checked the discovery-branch updates in both prompt copies and the added regression coverage around stable `brand` values, skip-form handling, and active-design-system behavior; the new instructions line up with the existing question-form `[value: ...]` contract and should keep the daemon/contracts prompts from drifting on this flow.

Issue #1855, opened 2026-05-15
OpenCode Not working


AnesuPhiri521:
### What happened?

When I use opencode. it used to work but not it not working after and update. now im getting these issues:

'Agent completed without producing any output. The model or provider may have returned an empty response — check the agent logs for upstream errors.'

I thought it was me only but it seems there are many people facing these issues.

### Steps to reproduce

1. setup opencode to be used for the Local CLI model
2. try any thing on any model

### Expected behavior

to design the requested

### Open Design version

0.6.0

### Platform

macOS (Apple Silicon)

### Logs / screenshots (optional)

```shell

```

lefarcen:
Hi @AnesuPhiri521 — thanks for writing this up, and sorry you hit the OpenCode path right after an update. The exact banner you pasted is useful because it comes from Open Design's empty-output guard: the daemon started OpenCode, but no visible `text`, tool, or error event reached the app.

This closely overlaps with #517, where we're already collecting Local CLI/OpenCode evidence, and it is related to the earlier #691 fix that made silent OpenCode failures show a visible error instead of looking successful. Since this report is on Open Design `0.6.0`, please first update to the latest desktop release (`0.7.0`) and retry **Settings → Local CLI → OpenCode → Test**.

If it still fails on `0.7.0`, could you share the smallest diagnostic bundle below?

```bash
opencode --version
printf 'Design a simple card\n' | opencode run --format json --dangerously-skip-permissions
```

Also please include the selected OpenCode model/provider and the daemon log lines around the failed Open Design run. That will tell us whether OpenCode itself is returning an empty/no-text response, or whether Open Design's adapter is missing a newer OpenCode event shape. We'll keep this linked with #517 so the evidence does not get split across threads.

CamiloFernando:
Found two root causes for the "Agent completed without producing any output" error with OpenCode after the v1.15.0 update:

**1. Outdated adapter since OpenCode v1.15.0**

The adapter in `apps/daemon/src/agents.ts` was written for pre-v1.15.0 OpenCode. Two issues:
- **Primary binary**: Uses `opencode-cli` which no longer exists. v1.15.0 unified the binary — `opencode` handles both GUI and CLI. If `opencode-cli` isn't on PATH, the fallback to `opencode` may fail silently depending on environment.
- **Deleted flag**: Passes `--dangerously-skip-permissions` which was removed in v1.15.0.

**2. Connection test uses temp dir for CWD**

[...]

PR #1864, opened 2026-05-15
fix(opencode): accept file-only project output

SirEdvin:
## Why

Project-scoped OpenCode runs can complete successfully by writing files into the OD project directory while emitting no user-visible JSON stream text. Today the daemon treats that clean `0` exit as an empty-output failure, so users see a failed run even though the artifact was created.

The repro that motivated this was an OpenCode project run that wrote `index.html` but ended with `AGENT_EXECUTION_FAILED` / “Agent completed without producing any output.” A related project-cwd issue also showed OpenCode consulting `PWD`, so spawned agents need `PWD` aligned with the actual project cwd.

## What users will see

When OpenCode updates project files without streaming text, the run now succeeds and emits a short assistant message such as:

```text
Updated project files: index.html
```

If no files changed and no output was produced, the existing empty-output failure path remains unchanged.

## Surface area

- [ ] **Extension point** — new entry under `skills/`, `design-systems`, `design-templates`, or `craft/`, or change to the skills protocol

## Screenshots

N/A — daemon/SSE behavior only.

## Bug fix verification

- Test path that reproduces the bug: `apps/daemon/tests/chat-route.test.ts` → `treats file-only OpenCode project runs as successful output`
- Did the test go red on `main` and green on this branch? Not rerun on `main` during this cleanup pass; the new regression encodes the prior empty-output failure for a file-only OpenCode project run and passes on this branch.
- Additional regression: `sets PWD to the project cwd for spawned project agents`, covering project agents receiving matching `process.cwd()` and `process.env.PWD`.

## Validation

- `corepack pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts` — 27 passed
- `corepack pnpm --filter @open-design/daemon typecheck` — passed
- `git diff --check origin/main...HEAD` — passed


lefarcen (commented):
Hi @SirEdvin — the OpenCode file-only run case is clearly described, and the verification notes are helpful. One small PR-body housekeeping ask before pool review: could you add the new `## Surface area` checklist and tick the relevant box for this daemon/runtime behavior change? Your Summary already covers the why/user-visible behavior and Verification covers validation, so no need to rename those sections. 🙌

SirEdvin:
Follow-up after reproducing the meal planner failure:

Root cause was not just file-only output. Project-scoped OpenCode launches inherited the daemon process `PWD` (`/home/siredvin/open-design`) even though `spawn({ cwd })` pointed at the `.od/projects/<id>` artifact directory. OpenCode consults `PWD` while resolving its project/session root, so the meal planner run was treated as launched from the Open Design checkout and the structured stdout events never reached OD as expected.

This branch now also sets `PWD` to the effective project cwd for spawned agents and adds a regression test that checks project agents receive matching `process.cwd()` / `process.env.PWD`.

Additional verification:
- `corepack pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts` — 26 passed
- `corepack pnpm --filter @open-design/daemon typecheck` — passed
- Replayed the latest meal planner prompt against a fresh daemon on port 7466: run succeeded and emitted OpenCode `agent` events instead of the empty-output failure.

lefarcen (commented):
Hey @SirEdvin, saw the follow-up on the meal-planner repro — that `PWD` detail is useful context. Since this is now explicitly a bugfix, could you copy the red→green verification from your comment into the PR body and also add the `## Surface area` checklist? That keeps the review trail self-contained even after the comment thread grows.

lefarcen:
Thanks for the update @SirEdvin — I see this is out of draft now and the branch was refreshed.

@nettee, could you take a look when you have a chance? You’re the routed reviewer for this one.

apps/daemon/src/server.ts
```diff
@@ -3143,6 +3143,58 @@ function resolveAcpStageTimeoutMs(): number | undefined {
...
+        if (snapshot.size >= PROJECT_OUTPUT_SNAPSHOT_LIMIT) return;
+        if (
+          entry.name === '.git'
+          || entry.name === 'node_modules'
+          || entry.name === '.od-skills'
+        ) {
+          continue;
+        }
+        const abs = path.join(dir, entry.name);
+        const rel = path.relative(root, abs) || entry.name;
```
lefarcen: P3 The `|| entry.name` fallback is dead code. `path.relative(root, abs)` returns an empty string only when `abs === root`, but `abs` here is always `path.join(dir, entry.name)` with a non-empty directory entry — so it can never equal `root`. Safe to drop to avoid confusing future readers.

lefarcen (commented):
Hey @SirEdvin — read through the implementation properly this time. The snapshot approach is well-constructed: the 500-file cap, the skip list for `.git`/`node_modules`/`.od-skills`, and the outer `try/catch` returning `null` on any I/O error all add up to a safe, bounded design. The PWD injection is the right fix (aligning `process.env.PWD` with the actual spawn cwd before the agent resolves its project root), and the test exercises the full path — file written → SSE `text_delta` → run status `succeeded`.

One P3 nit inline; nothing blocking from a logic standpoint. When @nettee gets a chance to review the API/contract surface, this looks ready. 🙌

SirEdvin:
Addressed the inline P3 nit in 5951da9: removed the unreachable `|| entry.name` fallback from `snapshotProjectOutputFiles`.

Verification:
- `corepack pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/chat-route.test.ts` — 36 passed
- `corepack pnpm --filter @open-design/daemon typecheck` — passed
- `corepack pnpm guard` — passed

lefarcen:
Thanks for addressing the follow-up @SirEdvin — this is back to waiting on routed review now.

@nettee, could you take another look when you have a chance? The latest head is `5951da9`.

lefarcen:
Thanks for the refresh @SirEdvin — the template/body updates are in place and CI is green now.

@nettee, could you take another look at the latest head when you have a chance? It’s `3ffec77`.

lefarcen:
Heads-up @SirEdvin / @nettee: #1855 has a new repro from @McFredward saying current main and this branch still hit an OpenCode failure while writing `index.html`, with the stream stopping mid-sentence.

I don’t think we can call that confirmed against this PR yet — the issue thread is now asking for the exact OD build, provider/model, disk state, and daemon logs — but it does mean we should treat #1864 as covering the file-only/no-visible-text path, not necessarily the whole #1855 failure family, until that new evidence comes back.

SirEdvin:
@lefarcen it seems issue described by  @McFredward is not related, so you can merge it.

lefarcen:
Thanks for the scoping clarity, @SirEdvin — the new repro from @McFredward in #1855 (stream stopping mid-sentence with text output already present) does look like a different failure class from the file-only/no-visible-text path this PR addresses. Your read makes sense.

The code is in good shape from my side — the snapshot + PWD injection logic is solid, tests cover the full success path, and the P3 nit was addressed cleanly. We're just waiting on a pool reviewer's approval before merge. @nettee has been queued for a while, so I'm escalating to the on-call to either expedite the review or approve directly. Should have a decision shortly.

lefarcen:
One more heads-up, @SirEdvin — the branch just developed a conflict with main (merge state flipped to DIRTY). A rebase on main should clear it, and CI will re-run on the fresh head. Worth doing before the approval lands so there's no delay on the merge side.

SirEdvin:
I fixed this again, can we merge or reject it already, please?

lefarcen:
Hey @SirEdvin, understood. The latest head is back to CI-green and no longer conflicted, so we’re not waiting on another author change from your side. What’s still missing is a human approval/merge call on this head; I’ve re-requested routed review from @nettee, and there’s already an escalation out from earlier today, so this is pending maintainer decision rather than blocked on you.

PR #1865, opened 2026-05-15
fix(web): surface zip import failures

### LINKED ISSUE #1862: Unable to import a zip file into a new project
### What happened?

When I try to create a new project by clicking import then selecting a zip file that is intended for claude design, it doesn't initiate the creation of a new project using that zip file.

### Steps to reproduce

1. Open open design.
2. Create new
3. Upload zip file
Nothing happens after selecting file from finder

### Expected behavior

In Claude Design, a zip file exported from Relume can be imported as such (see 3:05 mark) https://www.loom.com/share/68fee82cd9244c319d7ffb0db838342e

I don't see a way to similarly import the same file into open design :(

### Open Design version

0.7.0

### Platform

macOS (Intel)

```shell

```

### Additional context

_No response_

mturac:
Fixes #1862

## Why

I checked the ZIP import path from the issue and found the web side was swallowing daemon rejections. When the daemon rejected a Claude Design/Relume-like ZIP, the file picker closed and the New Project panel had no visible recovery signal.

## What users will see

Uploading a ZIP that cannot be imported now shows an error toast with the daemon rejection reason instead of doing nothing.

## Surface area

## Screenshots

Not attached; this is a transient error toast and is covered by the component regression test below.

## Bug fix verification

- Test path that reproduces the bug: `apps/web/tests/components/NewProjectPanel.test.tsx`
- Did the test go red on `main` and green on this branch? Not rerun on `main`; the assertion targets the panel behavior that previously ignored failed ZIP imports.
- The regression verifies that a failed Claude Design ZIP import now renders an error toast.

## Validation

- `pnpm guard`
- `pnpm --dir apps/web exec vitest run tests/components/NewProjectPanel.test.tsx -t 'Claude Design zip import'`
- `pnpm --dir apps/web exec vitest run tests/components/NewProjectPanel.test.tsx`
- `git diff --check`


PerishCode:
Hi @mturac, this is approved and all reported checks are green, but GitHub now marks the branch as DIRTY against main. Could you rebase or merge main into the branch when you get a chance so it can move forward?

PerishCode (changes requested):
@mturac Reviewed head 825f70f1.

I cannot approve this yet because the priority:p1 regression coverage still stops one layer short of the actual failure path. apps/web/tests/components/NewProjectPanel.test.tsx now proves that `NewProjectPanel` shows a toast when its `onImportClaudeDesign` callback resolves `{ ok: false }`, but the bug in #1862 was the web import path swallowing a daemon rejection from `/api/import/claude-design`. If apps/web/src/state/projects.ts or the App-level handler regressed back to returning `null`/`void` on a non-2xx import response, the new component test could still pass while users would again see nothing after selecting the ZIP.

Please add targeted coverage for that boundary, either in apps/web/tests/state/projects.test.ts for `importClaudeDesignZip` throwing/preserving the daemon error on non-2xx responses, or an App-level import test that stubs the endpoint rejection and observes the New Project toast. Once that is covered and the branch is rebased out of the current dirty state, the UI direction looks good ^_^

mturac:
Updated this branch with the requested boundary coverage and merged current main.

What changed:
- Added a state-level regression test for `importClaudeDesignZip` so non-2xx `/api/import/claude-design` responses preserve and throw the daemon error message.
- Kept the existing New Project panel failure-toast coverage.
- Merged `origin/main` into the branch to clear the dirty branch state.

Validation run:
- `pnpm --dir apps/web exec vitest run -c vitest.config.ts tests/state/projects.test.ts tests/components/NewProjectPanel.test.tsx`

lefarcen:
Hey @PerishCode, @mturac has pushed the requested state-level boundary coverage and merged current `main`; CI is green on the new head. Could you take another look when you have a chance?

mturac:
Thanks! I merged current `main` into this branch after the latest merge.

I'm going to spend a little time with family and will come back to this after that. This is genuinely one of my favorite projects :)

lefarcen:
Hey @mturac, enjoy the family time — and thank you for the kind words about the project. We’ll keep this parked for @PerishCode’s re-review; no rush from our side.

mturac:
Thank you, I really appreciate it.

Very suddenly and unexpectedly, we lost my wife's father. The work you are doing here is genuinely valuable, and although we have not really had the chance to talk much yet, I already have a lot of respect and affection for this team.

I just did not want to leave any open work or loose ends behind without saying something. I will take a little time, and then I will be happy to come back and continue with you.

You are doing wonderful work, and I am proud to have been a small part of it.

lefarcen:
Hey @mturac, I’m so sorry for your family’s loss.

Please take all the time you need — there are no loose ends to worry about here. We’ll be here whenever you’re ready, and we’re grateful you’ve been part of this with us.

apps/web/src/state/projects.ts
```diff
@@ -112,23 +112,28 @@ export async function importFolderProject(
...
+    body: form,
+  });
+  if (!resp.ok) {
+    const payload = await resp.json().catch(() => null);
+    const message =
+      payload != null &&
+      typeof payload === 'object' &&
+      typeof (payload as { error?: unknown }).error === 'string'
+        ? (payload as { error: string }).error
+        : `Import failed (${resp.status})`;
```
PerishCode, line 129: **Non-blocking:** the fallback message here will render with a doubled prefix.

`NewProjectPanel.handleImportPicked` prepends `Import failed: ` to whatever message it receives, and `App.handleImportClaudeDesign` passes the thrown `Error.message` straight through. When the daemon's documented `{ error }` envelope is present the toast reads cleanly (`Import failed: <daemon reason>`). But this fallback branch fires when the response body is *not* the `{ error }` shape — proxy/infra failures such as a 413 on a large ZIP or a 502/504 with a non-JSON body — and then the toast renders `Import failed: Import failed (413)`.

Why it matters: surfacing those failures legibly is the entire point of this PR, and a large-ZIP size-limit rejection is a plausible primary case for it, so the doubled wording reads as a bug in exactly the path being fixed.

Suggested change: drop the redundant `Import failed` from this fallback so the single UI-layer prefix owns it, e.g. `` `Daemon returned HTTP ${resp.status}` `` or `` `Unexpected response (HTTP ${resp.status})` ``. This fallback branch is also currently uncovered — worth a matching `projects.test.ts` case that returns a non-JSON (or no-`error`) non-2xx body and asserts the thrown message.


apps/web/tests/components/NewProjectPanel.test.tsx
```diff
@@ -635,6 +635,34 @@ describe('NewProjectPanel design system defaults', () => {
 });
 
 describe('NewProjectPanel folder import feedback', () => {
+  it('shows an error when Claude Design zip import resolves as failed', async () => {
```
PerishCode, line 638: **Nit:** this new test covers ZIP import, but it lives inside `describe('NewProjectPanel folder import feedback', ...)` — a folder-import group.

Why it matters: someone scanning the suite for ZIP-import regression coverage (the behavior #1862 is about) would not expect to find it under a folder-import describe, and the mis-grouping gets harder to untangle as the file grows.

Suggested change: move this case into its own `describe('NewProjectPanel zip import feedback', ...)` block (or rename the existing block to cover both surfaces), keeping the folder and zip cases grouped by the feature they exercise.


PerishCode (commented):
@mturac Thanks for circling back with the state-level `importClaudeDesignZip` coverage I asked for — `apps/web/tests/state/projects.test.ts` now exercises the actual non-2xx boundary from #1862, and routing the daemon rejection through the new `ImportClaudeDesignOutcome` shape into a New Project toast is the right structure. Two small follow-ups inline, neither blocking: a doubled "Import failed" prefix in one fallback branch, and a ZIP-import test placed under the folder-import describe block. Everything else looks solid.



lefarcen:
Thanks @PerishCode — noted that both follow-ups are non-blocking. One GitHub-state note: the earlier `CHANGES_REQUESTED` review still leaves the PR blocked, so if you’re comfortable with this head aside from those non-blocking items, could you flip the review state when you have a moment?

PR #1869, opened 2026-05-16
fix: use SKILL_ROOTS in /api/skills endpoints for correct source attribution

chu0119:
## Summary
- Fixed the `GET /api/skills` and `GET /api/skills/:id` daemon endpoints to use `SKILL_ROOTS = [USER_SKILLS_DIR, SKILLS_DIR]` instead of `SKILLS_DIR` alone
- This correctly separates user-installed skills (source: "user") from built-in skills (source: "built-in")
- Fixes incorrect source filter counts on the Skills settings page

## Why
`listSkills` determines a skill's source based on its root index: `rootIdx === 0 ? "user" : "built-in"`. When only `SKILLS_DIR` (built-in) was passed, ALL skills were attributed as "user" since that single root was at index 0. Using `SKILL_ROOTS` restores the correct two-root ordering where user skills shadow built-in skills.

This also aligns the `server.ts` handlers with `static-resource-routes.ts` which already correctly uses `listAllSkills()` → `listSkills(SKILL_ROOTS)`.

## What changed
- `apps/daemon/src/server.ts`: Changed `listSkills(SKILLS_DIR)` → `listSkills(SKILL_ROOTS)` in both `/api/skills` and `/api/skills/:id` handlers

## Test plan
- [ ] Verify `/api/skills` returns skills with correct `source` field ("user" for user-installed, "built-in" for built-in)
- [ ] Confirm Skills settings page shows accurate source filter counts

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -4127,7 +4127,7 @@ export async function startServer({
 
   app.get('/api/skills', async (_req, res) => {
     try {
-      const skills = await listSkills(SKILLS_DIR);
+      const skills = await listSkills(SKILL_ROOTS);
       // Strip full body + on-disk dir from the listing — frontend fetches the
       // body via /api/skills/:id when needed (keeps the listing payload small).
       res.json({
@@ -4143,7 +4143,7 @@ export async function startServer({
 
   app.get('/api/skills/:id', async (req, res) => {
     try {
-      const skills = await listSkills(SKILLS_DIR);
+      const skills = await listSkills(SKILL_ROOTS);
       const skill = findSkillById(skills, req.params.id);
       if (!skill) return res.status(404).json({ error: 'skill not found' });
       const { dir: _dir, ...serializable } = skill;


lefarcen (commented):
Hi @chu0119! The daemon/source-attribution write-up is clear — the Summary and Why sections make the root of the count bug easy to follow.

One quick PR-body cleanup before pool review: could you add the missing `## Surface area` checklist and a `## Validation` section with what was actually verified (or mark the current checks as pending if they still need to be run)? Since this touches the daemon API behind the Skills settings page, the surface area note helps reviewers scope the UI impact quickly.

Related: #1374 tracks the user/built-in source-count confusion, and #1655 is an open UI-side approach for the same symptom. Worth keeping those in view so we land the fix in the right layer.

lefarcen:
Hey @PerishCode, this one is still waiting on pool review when you have a chance.

Context: it’s a small daemon-side Skills source-attribution fix (`/api/skills` now uses `SKILL_ROOTS`), with #1374 and #1655 as related source-count context.

PerishCode (changes requested):
@chu0119 Thanks for digging into the Skills source attribution path. I cannot approve this as a bug fix yet because the changed handler appears to be behind an earlier duplicate route, so the PR likely does not change the endpoint users hit.

apps/daemon/src/server.ts calls `registerStaticResourceRoutes(app, ...)` before the edited `/api/skills` handlers. That module already registers GET `/api/skills` and GET `/api/skills/:id` in apps/daemon/src/static-resource-routes.ts, and those active handlers call `listAllSkills()` rather than `listSkills(SKILLS_DIR)`. Since Express stops once the earlier handler sends the response, the later server.ts handlers changed by this PR are not the ones serving the Skills settings page.

That also matches the #1374 screenshot/context: the page showed `user 0` and `built-in 106`, not all built-ins mislabeled as `user`. Please either move the fix to the active route path with a route-level regression test, or remove/untangle the stale duplicate handlers and update the PR scope if this is cleanup rather than the source-count fix. A test that starts the daemon with one user skill and one built-in skill, then asserts `/api/skills` and `/api/skills/:id` return the expected `source`, would make the authority clear. ^_^

PR #1813, opened 2026-05-15
fix(web): distinguish expanded memory preview action

mturac:
Fixes #1810

## Why

I picked this up from the Settings > Memory preview bug report. Expanded Project memory rows currently use the same close-shaped icon for the preview toggle and the delete action, which makes the row look like it has two destructive controls.

## What users will see

When a memory entry preview is expanded, the preview toggle now shows a downward chevron instead of a close icon. The real delete action remains the only close-shaped/destructive-looking control in the row.

## Surface area

- [x] **UI** — new page / dialog / panel / menu item / setting / empty state in `apps/web` or `apps/desktop` (including Electron menu bar)
- [ ] **Keyboard shortcut** — new or changed
- [ ] **CLI / env var** — new `od` subcommand or flag, new `tools-dev` / `tools-pack` / `tools-pr` flag, or new `OD_*` env var
- [ ] **API / contract** — new `/api/*` endpoint, new SSE event, or changed shape in `packages/contracts`
- [ ] **Extension point** — new entry under `skills/`, `design-systems`, `design-templates`, or `craft/`, or change to the skills protocol
- [ ] **i18n keys** — added new translation keys (see `TRANSLATIONS.md` for the locale workflow)
- [ ] **New top-level dependency** — adding any new entry to the **root** `package.json` (`dependencies` or `devDependencies`); workspace-package `package.json` files are out of scope. Include a paragraph on what we get vs. what bytes we ship (see `CONTRIBUTING.md` → Code style)
- [ ] **Default behavior change** — changes what existing users experience without opting in (default model, default setting, file/SQLite schema, auto-network on startup, auto-install)
- [ ] **None** — internal refactor, docs, tests, or translation update only

## Screenshots

Not attached from this fresh local clone because the workspace is not bootstrapped. The UI change is limited to the expanded memory preview affordance: close icon -> downward chevron.

## Bug fix verification

- Test path that reproduces the bug: `apps/web/tests/components/MemorySection.test.tsx`
- Did the test go red on `main` and green on this branch? no; local test execution is blocked because this fresh clone has no `node_modules` and `vitest` is not installed.
- The added regression asserts that an expanded memory card has only one close-shaped icon path, leaving the delete action as the only close-shaped control.

## Validation

- `git diff --check`
- `corepack pnpm --filter @open-design/web test -- MemorySection.test.tsx` (blocked: node_modules missing / vitest not found; local Node is v25.9.0 while the repo expects ~24)

PR #1871, opened 2026-05-16
fix(contracts): tighten LiveArtifactSsePayload.refreshStatus to LiveArtifactRefreshStatus

leessju:
## Why

I was auditing the open-source codebase for type-safety gaps between `packages/contracts` and its consumers when I noticed that `LiveArtifactSsePayload.refreshStatus` was typed as `string | undefined`, while every consumer and the REST counterpart treat it as the canonical `LiveArtifactRefreshStatus` enum (`'never' | 'idle' | 'running' | 'succeeded' | 'failed'`).

The drift is silent and easy to miss:

- The daemon emitter at `apps/daemon/src/server.ts` already passes `artifact.refreshStatus`, which is typed as the enum on `LiveArtifact`.
- Web consumers in `apps/web/src/components/{DesignsTab,FileViewer,LiveArtifactBadges}.tsx` `switch` on the enum members.
- But the SSE contract typed it as a bare `string`, so a daemon adding a new refresh state for REST could ship it out via SSE as an opaque string with no compile-time signal to the consumer that a `switch` is now non-exhaustive.

This PR aligns the SSE payload type with the canonical enum so the two transports cannot drift unnoticed. Runtime behavior is unchanged — this is a type-only fix.

## What users will see

Nothing user-facing changes. The wire format is identical; only TypeScript types are tightened.

## Surface area

- [x] **API / contract** — changed shape in `packages/contracts` (type tightening only; no field added/removed)

## Bug fix verification

A red spec wasn't cheap here because the bug is a *future-drift hazard* rather than a current incorrect-runtime-behavior bug — at the moment, every consumer happens to compare against valid enum members, so no red runtime test exists today. The verification I ran is type-level:

- Before this change: a hypothetical daemon emit of `refreshStatus: 'totally-bogus'` would type-check on the daemon side and on the consumer side (because the consumer accepts `string`).
- After this change: the same emit would fail `pnpm typecheck` on both ends, because the SSE payload narrows to the enum.

I confirmed this by hand-testing both directions in a scratch file.

## Validation

- `pnpm --filter @open-design/contracts build` — green
- `pnpm --filter @open-design/contracts test` — 78 tests passed
- `pnpm --filter @open-design/web typecheck` — green
- `pnpm --filter @open-design/daemon typecheck` — green
- `pnpm guard` — green
- No `apps/web` runtime changes, no SSE wire format change, so no Playwright regression run is necessary. The two consumers that read this field (`AssistantMessage`/streaming flow, `DesignsTab` cards) keep the same JavaScript output before/after.

### DIFF
--- a/packages/contracts/src/sse/chat.ts
+++ b/packages/contracts/src/sse/chat.ts
@@ -1,3 +1,4 @@
+import type { LiveArtifactRefreshStatus } from '../api/live-artifacts.js';
 import type { SseErrorPayload } from '../errors.js';
 import type { SseTransportEvent } from './common.js';
 
@@ -10,7 +11,14 @@ export interface LiveArtifactSsePayload {
   projectId: string;
   artifactId: string;
   title: string;
-  refreshStatus?: string;
+  /**
+   * Refresh lifecycle state of the artifact at emit time. Typed against the
+   * canonical `LiveArtifactRefreshStatus` enum used by the REST API so that
+   * SSE consumers (web, CLI) can switch on the same union members without
+   * widening to `string`. Optional because the daemon may omit the field on
+   * legacy events; consumers must still null-check before narrowing.
+   */
+  refreshStatus?: LiveArtifactRefreshStatus;
 }
 
 export interface LiveArtifactRefreshSsePayload {


Siri-Ray (approved):
@leessju I reviewed the contract-only change in `packages/contracts/src/sse/chat.ts` and checked the daemon producer/web consumer references around `refreshStatus`. The SSE payload now points at the same `LiveArtifactRefreshStatus` union as the REST live-artifact contract while keeping the field optional, so the wire shape stays compatible and the type tightening looks consistent with the existing emitter. Thanks for tightening this boundary and documenting the intent clearly.


leessju:
Heads-up: the `Validate workspace > App workspace web tests` failure here is **not caused by this PR**. `main` itself has been red on the same `MemorySection > keeps the expanded preview control visually distinct from delete` assertion since #1813 merged — the assertion in #1813's new test points at the Lucide-default close-icon path (`M18 6 6 18`) while this repo's `Icon name="close"` deliberately renders `M20 4 4 20`. I opened #1877 to retarget the assertion at the path the close glyph actually emits, which restores `main` (and this PR) to green. Reproducible: check out `origin/main` at `1896c699` and run `pnpm --filter @open-design/web test -- tests/components/MemorySection.test.tsx`.

Issue #1892, opened 2026-05-16
Dependency security audit: 4 HIGH advisories with proposed pnpm.overrides + Next.js patch + Astro 6 path


leessju:
## Summary

`pnpm audit --prod --json` at the repo root reports **11 advisories** on `main`: 4 HIGH, 5 MODERATE, 2 LOW (zero critical). The four HIGH advisories are reachable from `apps/web` (Next.js Middleware bypass) and `apps/daemon` (transitive via `@modelcontextprotocol/sdk` → `ajv` → `fast-uri`) and from `apps/landing-page` (Astro `devalue` DoS).

All vulnerable transitive versions are pinned in `pnpm-lock.yaml`, so there is no version drift — the fix is a deterministic `pnpm.overrides` block (for transitive deps) plus two direct upgrades (Next.js + Astro).

## HIGH advisories

| # | Advisory | Package@Version | Path | CVE / GHSA | Reachability |
|---|---|---|---|---|---|
| 1 | Next.js Middleware/Proxy bypass via segment-prefix | `next@16.2.5` | `apps/web` direct | GHSA-pending | Active — `apps/web` ships middleware |
| 2 | `fast-uri@3.1.0` path traversal via `%2e%2e` in `normalize()`/`equal()` | transitive | `apps/daemon` → `@modelcontextprotocol/sdk` → `ajv` → `fast-uri` | CVE-2026-6321, CVSS 7.5 | Active — daemon validates inbound JSON via ajv schemas |
| 3 | `fast-uri@3.1.0` host confusion via `%40`/`%3A` (SSRF/allowlist bypass relevance) | same transitive path | same | CVE-2026-6322, CVSS 7.5 | Active — daemon already does its own hostname resolution (PR #1176) but defense-in-depth still matters |
| 4 | Svelte `devalue@5.8.0` DoS via sparse arrays | transitive via Astro | `apps/landing-page` → `astro` → `devalue` | GHSA-pending | Static-site build path; runtime exposure ≈ 0 but build-time DoS is still real |

## MODERATE advisories (4)

- Astro 5.18 XSS in `define:vars` (CVE-2026-41067). Static landing page doesn't use `define:vars` with user input, so practical exposure is low; resolved by Astro 6.1.10.
- Hono cache/CSS-injection (`hono@4.12.16`, patched `>=4.12.18`). Daemon does not use Hono JSX or Hono cache directly; practical exposure low.
[...]

PR #1895, opened 2026-05-16
fix(security): pin fast-uri and devalue via pnpm.overrides to resolve 3 HIGH advisories

leessju:
Refs #1892

## Why

`pnpm audit --prod --json` at the repo root currently surfaces **4 HIGH-severity advisories** on `main`:

```
high    1117870  fast-uri  <=3.1.0   path traversal via percent-encoded dots
high    1117884  fast-uri  <=3.1.1   host confusion via percent-encoded authority
high    1118938  next      >=16.0.0 <16.2.6   Middleware/Proxy bypass in App Router
high    1119005  devalue   >=5.6.3 <=5.8.0    Svelte devalue: DoS via sparse array deserialization
```

This PR resolves **three of the four** via the most surgical change possible — a `pnpm.overrides` block in the root `package.json`. All three vulnerable packages (`fast-uri` x2 + `devalue` x1) are pinned transitively (`fast-uri` reaches `apps/daemon` via `@modelcontextprotocol/sdk` → `ajv`; `devalue` reaches `apps/landing-page` via `astro`), and none has a direct dependency in any workspace manifest, so a single overrides block is the right tool: it forces the patched version into the resolved tree without bumping any workspace's declared dependency surface.

Verified with `pnpm audit --prod --json` after `pnpm install`: the three resolved HIGH advisories drop off (`fast-uri@<=3.1.0`, `fast-uri@<=3.1.1`, `devalue@>=5.6.3 <=5.8.0`). Only `next` HIGH remains.

The remaining HIGH advisory — `next@^16.2.5` (direct in `apps/web`) — is intentionally held back to keep this PR easy to revert if anything downstream catches a transitive resolution change. See #1892 for the proposed split; the Astro 5 → 6 major bump path noted in that issue is also out of scope here because the `devalue` override in this PR already neutralizes that advisory without needing the major bump.

## What users will see

Nothing user-facing changes. Wire formats, APIs, UI, and shipped JavaScript are unaffected; only the resolved transitive versions of `fast-uri` and `devalue` move up to the patched lines.

## Surface area

- [x] **None** — internal refactor, docs, tests, or translation update only (root `package.json` + `pnpm-lock.yaml` only)

## Validation

- `pnpm install` regenerates the lockfile with `fast-uri@>=3.1.2` and `devalue@>=5.8.1` resolved across every transitive consumer.
- `pnpm audit --prod` after the change: the two `fast-uri` HIGH advisories and the `devalue` HIGH advisory drop off. `next@^16.2.5` remains until the follow-up PR.
- `pnpm typecheck` — green across every workspace.
- `pnpm guard` — green.
- Each touched-workspace test suite I could run locally — green (contracts, sidecar, platform, sidecar-proto, registry-protocol).

## Why I'm holding the other two HIGH advisories back

- **`next@^16.2.5` → `next@^16.2.6`**: a Next.js patch bump is mechanical, but it touches `apps/web` directly, so I'd rather land it as its own PR with a focused web-test run. Filed as a follow-up in #1892.
- **`astro@^5.15.4` → `astro@^6.x`**: that's a major-version bump for `apps/landing-page`, with a real risk of breaking-change surface in content collections, view transitions, etc. The landing page deploys to Cloudflare Pages on every push to `main` (`.github/workflows/landing-page-deploy.yml`), so I don't want to land an Astro 6 path here. The `devalue` override in this PR resolves the `devalue` DoS without needing the Astro bump — it forces the patched `devalue@>=5.8.1` into Astro 5's resolved tree.

## Open question for reviewers

If you'd rather see all four advisories handled in a single consolidated PR, I'll fold the `next` patch + Astro 6 path in. The reason this PR is scoped narrowly is risk surface: the overrides change is mechanically safe (transitive resolution only), where the direct-dep bumps need more attention.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,10 @@
       "better-sqlite3",
       "electron",
       "esbuild"
-    ]
+    ],
+    "overrides": {
+      "fast-uri": "3.1.2",
+      "devalue": "5.8.1"
+    }
   }
 }
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -4,6 +4,10 @@ settings:
   autoInstallPeers: true
   excludeLinksFromLockfile: false
 
+overrides:
+  fast-uri: 3.1.2
+  devalue: 5.8.1
+
 importers:
 
   .:
@@ -2592,8 +2596,8 @@ packages:
     resolution: {integrity: sha512-KxektNH63SrbfUyDiwXqRb1rLwKt33AmMv+5Nhsw1kqZ13SJBRTgZHtGbE+hH3a1mVW1cz+4pqSWVPAtLVXTzQ==}
     engines: {node: '>=18'}
 
-  devalue@5.8.0:
-    resolution: {integrity: sha512-2zA9pFEsnp7vWBZbXF5JAgAq0fsUIt/1XPbRiAmRV3lp/2C3upzH+sADiyy66aFCihoLEsrQHxNM5w1gIDfsBg==}
+  devalue@5.8.1:
+    resolution: {integrity: sha512-4CXDYRBGqN+57wVJkuXBYmpAVUSg3L6JAQa/DFqm238G73E1wuyc/JhGQJzN7vUf/CMphYau2zXbfWzDR5aTEw==}
 
   devlop@1.1.0:
     resolution: {integrity: sha512-RWmIqhcFf1lRYBvNmr7qTNuyCt/7/ns2jbpp1+PalgE/rDQcBT0fioSMUpJ93irlUhC5hrg4cYqe6U+0ImW0rA==}
@@ -2839,8 +2843,8 @@ packages:
   fast-json-stable-stringify@2.1.0:
     resolution: {integrity: sha512-lhd/wF+Lk98HZoTCtlVraHtfh5XYijIjalXck7saUtuanSDyLMxnHhSXEDJqHxD7msR8D0uCmqlkwjCV8xvwHw==}
 
-  fast-uri@3.1.0:
-    resolution: {integrity: sha512-iPeeDKJSWf4IEOasVVrknXpaBV0IApz/gp7S2bb7Z4Lljbl2MGJRqInZiUrQwV16cpzw/D3S5j5Julj/gT52AA==}
+  fast-uri@3.1.2:
+    resolution: {integrity: sha512-rVjf7ArG3LTk+FS6Yw81V1DLuZl1bRbNrev6Tmd/9RaroeeRRJhAt7jg/6YFxbvAQXUCavSoZhPPj6oOx+5KjQ==}
 
   fast-xml-builder@1.2.0:
     resolution: {integrity: sha512-00aAWieqff+ZJhsXA4g1g7M8k+7AYoMUUHF+/zFb5U6Uv/P0Vl4QZo84/IcufzYalLuEj9928bXN9PbbFzMF0Q==}
@@ -6448,7 +6452,7 @@ snapshots:
   ajv@8.20.0:
     dependencies:
       fast-deep-equal: 3.1.3
-      fast-uri: 3.1.0
+      fast-uri: 3.1.2
       json-schema-traverse: 1.0.0
       require-from-string: 2.0.2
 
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -39,6 +39,10 @@
       "better-sqlite3",
       "electron",
       "esbuild"
-    ]
+    ],
+    "overrides": {
+      "fast-uri": ">=3.1.2",
+      "devalue": ">=5.8.1"
```
nettee: These `pnpm.overrides` entries should be exact pins (`"3.1.2"` / `"5.8.1"`), not lower-bounded ranges. Because `overrides` rewrites dependency specs globally, `">=3.1.2"` and `">=5.8.1"` let a future `pnpm install` on this same commit silently float to newer minor or major releases that the upstream dependents never declared compatibility with. That breaks the PR's goal of being a surgical, low-risk remediation and weakens reproducibility even though the current lockfile happens to resolve to 3.1.2 and 5.8.1. Please pin the override values to the exact audited versions here, then regenerate the lockfile so the contract and the resolved tree stay aligned.

nettee (changes requested):
I found one blocking issue in the new override contract: the security overrides are written as lower-bounded ranges instead of exact pins, which makes this change float to future unreviewed transitive upgrades on the next install.


lefarcen (commented):
Hey @leessju, the split from #1892 is easy to follow, and the PR body does a good job explaining why the transitive override path is lower-risk than mixing in the Next.js / Astro follow-ups.

One template note before this goes back through review: since this is labeled as a bug/security fix, could you add a short **Bug fix verification** section that spells out the before → after check explicitly? For this PR, something like “before: root `pnpm audit --prod` reports the two `fast-uri` HIGH advisories plus the `devalue` HIGH; after: those three are gone and only the intentionally deferred `next` advisory remains” would give reviewers the red-to-green seam at a glance.

@nettee has already covered the current blocking code-review item separately, so this is just the PR-body bookkeeping piece.

leessju:
Pushed fixup addressing both review notes:

**@nettee** — Tightened both overrides to exact pins:
```json
"overrides": {
  "fast-uri": "3.1.2",
  "devalue": "5.8.1"
}
```
The next install no longer floats transitive consumers onto unreviewed future versions. Verified that `pnpm install` resolves cleanly with the tighter constraint.

**@lefarcen** — Added an explicit before → after Bug fix verification block in the updated commit message:
```
Before (origin/main):
  high 1117870 fast-uri
  high 1117884 fast-uri
  high 1118938 next
  high 1119005 devalue

After (this patch):
  high 1118938 next
```
(reproduced locally via `pnpm audit --prod --json | jq …` against both commits)

The remaining HIGH (`next@^16.2.5 -> ^16.2.6`) stays scoped out per #1892's split rationale.

**On the failing `build` job**: it surfaces a Nix `pnpmDepsHash` mismatch — the override bump moved `pnpm-lock.yaml`, so the pinned hash in `nix/package-{daemon,web}.nix` (`sha256-lROdH5HgKFf3R7DYGbc8n/GrmINwLbfVwC4Xp7SrHN4=`) is now stale. I don't have a Nix toolchain locally so I can't regenerate it from here; the CI build log will print the expected hash, and bumping the pinned value follows the same pattern as #1765 (`Nix flake refresh pnpmDepsHash after merging main`). Happy to push the regen as a follow-up commit once the expected hash is shared.

lefarcen (commented):
FYI @leessju: there are several empty Looper-generated PRs for the same issue (#2540, #2545, #2548, #2555 — all 0 changed files). None of them overlap with your work here; the Looper runner opened them but left the branches unchanged because it kept hitting detection false positives. Your PR is the only substantive implementation for these advisories.

Just flagging so the duplicate PR list doesn't cause confusion during review.

lefarcen:
Hey @nettee — this one may have slipped through. @leessju pushed a fixup on May 16 addressing your blocking point directly: both overrides are now exact pins (`fast-uri: "3.1.2"` and `devalue: "5.8.1"`) rather than floating ranges. Could you take another look when you get a chance?

One remaining infra item: the `nix-check` CI is failing because `pnpmDepsHash` in `nix/package-{daemon,web}.nix` went stale after the lockfile regeneration. The CI log prints the expected hash; same update pattern as #1765. @leessju mentioned they don't have a Nix toolchain locally, so a maintainer push of the updated hash would clear that final check.

leessju:
Closing as obsolete: the `fast-uri@3.1.2` and `devalue@5.8.1` pnpm overrides this PR adds are already present in `main`'s `package.json` `pnpm.overrides` (verified against the latest `main`), so the HIGH advisories are already pinned. Rebasing would leave an empty diff. Closing to keep the queue clean.

Issue #1886, opened 2026-05-16
Lint craft section references to surface typos without losing the forward-reference allowance


leessju:
## What I noticed

`apps/daemon/src/craft.ts` resolves the slugs a skill lists under `od.craft.requires` into `craft/<slug>.md` snippets that get spliced into the system prompt:

```ts
try {
  const filePath = path.join(craftDir, `${slug}.md`);
  const text = await readFile(filePath, "utf8");
  // ...
} catch {
  // File doesn't exist or unreadable — skip silently. Skills can
  // forward-reference future craft sections without breaking.
}
```

The silent drop is deliberate (the comment is explicit) and the forward-reference behavior is worth preserving — a skill that lists `motion` before `craft/motion.md` ships should not break.

The cost of that design is that **typos drop silently too**. A skill that asks for `typograpy` instead of `typography` quietly loses that knowledge slice at runtime, with no signal at PR time, no signal at boot, and no signal in `pnpm guard`.

I noticed this while auditing `craft/` (12 sections shipped today) against `skills/*/SKILL.md`. The forward-reference allowance and a typo are indistinguishable from inside `loadCraftSections`, but they're easily distinguishable at the repo level: a typo will _never_ resolve across the whole repo, while a real forward-reference is a slug that doesn't exist *yet* but is documented somewhere as planned.

## Proposed lightweight lint (no behavior change)

Add a `scripts/lint-craft-references.ts` invoked from `pnpm guard` (or a new `pnpm lint:craft` if `guard` should stay hard-fail only). The script:

1. Reads every `skills/*/SKILL.md` (+ `design-templates/*/SKILL.md`) and collects every slug under `od.craft.requires`.
2. Reads `craft/*.md` to learn which slugs *do* exist today.
3. Reads a tiny `craft/FUTURE_SECTIONS.md` (or a single `# Future sections` list in `craft/README.md`) for known-planned slugs that we're intentionally forward-referencing.
4. Reports any slug that's referenced by a skill but **neither** present in `craft/` **nor** in the future list. That's the typo class.

The runtime `loadCraftSections` keeps its current behavior — silent drop — so forward-references and lints never interfere with the actual prompt assembly path.

## Why bother

- One-line skill manifests are easy to typo (`typograpy`, `accessibility-baselines`, …) and the only signal today is "the agent's output feels generically less-disciplined than expected", which is hard to attribute back to the missing craft slice.
[...]

PR #1897, opened 2026-05-16
feat(scripts): add informational craft.requires reference linter (pnpm lint:craft)

leessju:
Refs #1886

## Why

`apps/daemon/src/craft.ts` resolves each slug under a skill's `od.craft.requires` into `craft/<slug>.md` and splices the snippet into the system prompt. The loader silently drops slugs whose matching file is missing — that's the documented forward-reference allowance:

```ts
} catch {
  // File doesn't exist or unreadable — skip silently. Skills can
  // forward-reference future craft sections without breaking.
}
```

The forward-reference behavior is worth preserving. The cost is that **typos drop silently too**. A skill asking for `typograpy` instead of `typography` quietly loses that knowledge slice at runtime with no signal at PR time, at boot, or in `pnpm guard`. The forward-reference allowance and a typo look identical from inside `loadCraftSections`.

This PR adds `scripts/lint-craft-references.ts` (run via `pnpm lint:craft`) that makes the typo class visible **without** removing the forward-reference behavior, by maintaining the distinction externally:

- Slugs intentionally forward-referenced go in `craft/FUTURE_SECTIONS.md` (optional file; absence is fine).
- Slugs that match neither `craft/<slug>.md` nor the future list are flagged as "likely typos or missing-from-future-list".

The runtime `loadCraftSections` keeps its current behavior — silent drop — so nothing in the production prompt path changes. This is purely a lint signal.

## Result on `main` today

Running `pnpm lint:craft` against `main` surfaces 3 unresolved slugs:

```
craft references: 51 total across 19 manifests
craft sections present: 11 (accessibility-baseline, animation-discipline,
                            anti-ai-slop, color, form-validation, laws-of-ux,
                            rtl-and-bidi, state-coverage, typography,
                            typography-hierarchy, typography-hierarchy-editorial)
craft sections marked future-only: 0 (no craft/FUTURE_SECTIONS.md present)

Unresolved craft slugs (likely typos or missing-from-future-list): 3
  'motion-discipline' — referenced by 1 manifest(s):
    - design-templates/live-dashboard/SKILL.md
  'pixel-discipline' — referenced by 4 manifest(s):
    - design-templates/kami-deck/SKILL.md
    - design-templates/kami-landing/SKILL.md
    - design-templates/open-design-landing/SKILL.md
    - design-templates/open-design-landing-deck/SKILL.md
  'typographic-rhythm' — referenced by 4 manifest(s):
    - design-templates/kami-deck/SKILL.md
    - design-templates/kami-landing/SKILL.md
    - design-templates/open-design-landing/SKILL.md
    - design-templates/open-design-landing-deck/SKILL.md
```

Whether those three slugs are typos, intentional forward-references for craft sections that are about to land, or genuine missing-from-future-list entries — that's a maintainer call. This PR doesn't try to answer it; it just makes the question surface explicitly so the answer can be made deliberately.

## What users will see

Nothing user-facing. The runtime craft loader behavior is unchanged. The new `pnpm lint:craft` command is opt-in for contributors and maintainers.

## Surface area

- [x] **None** — internal refactor, docs, tests, or translation update only (new informational script under `scripts/`)

## Validation

- `pnpm lint:craft` runs end-to-end against `main` content, exit 0.
- `pnpm lint:craft --strict` would exit 1 on the current `main` state (3 unresolved slugs). Not wired into CI by this PR.
- `pnpm guard` green.
- `pnpm exec tsc -p scripts/tsconfig.json --noEmit` green.

## Open question for reviewers

1. Are the three flagged slugs (`motion-discipline`, `pixel-discipline`, `typographic-rhythm`) typos or intentional forward-references? If the latter, dropping a `craft/FUTURE_SECTIONS.md` with those three slugs would silence the linter without changing runtime behavior. I didn't include that file in this PR because it's a content decision (it ships a maintainer commitment that those sections will eventually land), not a tooling decision.

2. Should `pnpm guard` invoke this script in `--strict` mode? Issue #1886 frames this as an RFC question — leaving the policy hook out of this PR so the script can land before the decision is made.

## Why I'm not folding the answer into this PR

Both questions touch maintainer-side decisions:
- The future-list contents commit the project to landing those craft sections.
- Wiring `--strict` into `guard` is a CI-policy change that will fail every PR (or every contributor's local `pnpm guard`) until the three unresolved slugs are reconciled.

Each can be a follow-up PR once #1886 has direction. The script in this PR is the substrate, not the policy.

### DIFF
--- a/scripts/lint-craft-references.ts
+++ b/scripts/lint-craft-references.ts
@@ -1,0 +1,244 @@
+// Lint craft section references across skills and design-templates.
+//
+// Background: `apps/daemon/src/craft.ts` resolves each slug under a
+// skill's `od.craft.requires` into `craft/<slug>.md` and splices the
+// snippet into the system prompt. The loader silently drops slugs
+// that don't have a matching file, by deliberate design — skills are
+// allowed to forward-reference future craft sections without
+// breaking. (See the inline comment in `craft.ts`.)
+//
+// The downside is that typos drop silently too. A skill that asks
+// for `typograpy` instead of `typography` quietly loses that
+// knowledge slice at runtime, with no signal at PR time, no signal
+// at boot, and no signal in `pnpm guard`.
+//
+// This script lints for that typo class without taking away the
+// forward-reference allowance. It:
+//
+//   1. Reads every `skills/*/SKILL.md` and `design-templates/*/SKILL.md`,
+//      extracts the `craft.requires` slug list.
+//   2. Reads `craft/*.md` to learn which slugs do exist today.
+//   3. Reads `craft/FUTURE_SECTIONS.md` (if present) for slugs the
+//      project is intentionally forward-referencing. Falls back to an
+//      empty list if absent — the policy decision on whether to keep
+//      a future-section allowlist is tracked in #1886.
+//   4. Reports any slug referenced by a skill but neither present in
+//      `craft/` nor in the future list. That's the typo class.
+//
+// Exit code stays 0 by default — this is informational, matching the
+// scope of #1886 ("RFC before code, please"). To make it blocking,
+// pass `--strict` or wire it into `pnpm guard`; the runtime loader
+// behavior in `craft.ts` is unchanged either way.
+
+import { readFile, readdir } from "node:fs/promises";
+import path from "node:path";
+
+const repoRoot = path.resolve(import.meta.dirname, "..");
+const craftDirectory = path.join(repoRoot, "craft");
+const skillsDirectory = path.join(repoRoot, "skills");
+const designTemplatesDirectory = path.join(repoRoot, "design-templates");
+const futureSectionsFile = path.join(craftDirectory, "FUTURE_SECTIONS.md");
+
+const SLUG_RE = /^[a-z0-9][a-z0-9-]*$/;
+
+type Reference = {
+  manifestPath: string;
+  slug: string;
+};
+
+async function readExistingCraftSlugs(): Promise<Set<string>> {
+  const entries = await readdir(craftDirectory, { withFileTypes: true });
+  const slugs = new Set<string>();
+  for (const entry of entries) {
+    if (!entry.isFile()) continue;
+    if (!entry.name.endsWith(".md")) continue;
+    if (entry.name === "README.md") continue;
+    const slug = entry.name.slice(0, -".md".length);
+    if (SLUG_RE.test(slug)) slugs.add(slug);
+  }
+  return slugs;
+}
+
+async function readFutureSlugs(): Promise<Set<string>> {
+  const slugs = new Set<string>();
+  try {
+    const text = await readFile(futureSectionsFile, "utf8");
+    // Pick up slugs in any of three flavors. A `FUTURE_SECTIONS.md`
+    // can use any of them:
+    //   1. bullet-list items:        `- slug` or `* slug`
+    //   2. inline-code spans:        `` `slug` ``
+    //   3. bare slugs on their own line: `slug`
+    // (1) and (2) cover slugs that sit inside prose paragraphs;
+    // (3) covers the simplest flat list. The bare-line pattern is
+    // anchored to the full line so prose like `Implement pixel-discipline`
+    // doesn't grab `pixel-discipline` accidentally — only a line that
+    // contains exactly the slug (and optional whitespace) counts.
+    const patterns = [
+      /(?:^[*\-]\s*|`)([a-z0-9][a-z0-9-]*)(?:`|$)/gm,
+      /^[ \t]*([a-z0-9][a-z0-9-]*)[ \t]*$/gm,
+    ];
+    for (const pattern of patterns) {
+      pattern.lastIndex = 0;
+      let match: RegExpExecArray | null;
+      while ((match = pattern.exec(text)) !== null) {
+        const slug = match[1];
+        if (slug && SLUG_RE.test(slug)) slugs.add(slug);
+      }
+    }
+  } catch (err) {
+    if (
+      err &&
+      typeof err === "object" &&
+      "code" in err &&
+      (err as { code?: string }).code === "ENOENT"
+    ) {
+      return slugs;
+    }
+    throw err;
+  }
+  return slugs;
+}
+
+function extractCraftRequiresSlugs(source: string): string[] {
+  // SKILL.md frontmatter shape (yaml-flavored):
+  //
+  //   craft:
+  //     requires: [typography, accessibility-baseline, state-coverage]
+  //
+  // or:
+  //
+  //   craft:
+  //     requires:
+  //       - typography
+  //       - accessibility-baseline
+  //
+  const inlineMatch = /^\s*craft:\s*\n\s*requires:\s*\[([^\]\n]*)\]/m.exec(source);
+  if (inlineMatch) {
+    const list = inlineMatch[1];
+    if (!list) return [];
+    return list
+      .split(",")
+      .map((part) => part.trim())
+      .filter((part) => part.length > 0)
+      .map((part) => part.replace(/^['"]|['"]$/g, ""));
+  }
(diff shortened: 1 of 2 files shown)


mrcfps:
@leessju I'm holding off on generating review comments for nexu-io/open-design#1897 because this pull request has merge conflicts right now.

scripts/lint-craft-references.ts
```diff
@@ -0,0 +1,231 @@
...
+}
+
+async function readFutureSlugs(): Promise<Set<string>> {
+  const slugs = new Set<string>();
+  try {
+    const text = await readFile(futureSectionsFile, "utf8");
+    // Pick out fenced slugs or list-item slugs. A `FUTURE_SECTIONS.md`
+    // can use any flavor; we accept `- slug` bullet lines, ``code``
+    // spans, and bare slug tokens.
+    const pattern = /(?:^[*\-]\s*|`)([a-z0-9][a-z0-9-]*)(?:`|$)/gm;
```
mrcfps: `readFutureSlugs()` advertises support for “bare slug tokens”, but the regex here only matches list items (`- slug`) or backticked slugs. For example, `pixel-discipline` on its own line produces no match, so a maintainer could add `craft/FUTURE_SECTIONS.md` exactly as described and still keep getting unresolved warnings. That matters because the allowlist is the main escape hatch this PR introduces for intentional forward references. Please either teach the parser to accept plain `slug` lines as well (for example by also matching `^\s*([a-z0-9][a-z0-9-]*)\s*$`) or narrow the documented file format so the code and guidance stay in sync.\n\n\n<sub>🔁 Powered by <a href=\"https://github.com/nexu-io/looper\">Looper</a> · runner=reviewer · agent=opencode · An autonomous AI dev team for your GitHub repos.</sub>

mrcfps (commented):
@leessju Thanks for shipping this lint pass — the overall approach looks good, but I found one follow-up in the `FUTURE_SECTIONS.md` parser before maintainers start relying on that allowlist.



leessju:
Thanks @mrcfps — you're right, the parser advertised 'bare slug tokens' but the regex only matched list items and backticked tokens. Just pushed a fix.

The new shape uses two patterns instead of one, so each slug shape is matched explicitly:

1. `(?:^[*\-]\s*|\`)([a-z0-9][a-z0-9-]*)(?:\`|$)/gm` — list items and backticked spans (unchanged)
2. `^[ \t]*([a-z0-9][a-z0-9-]*)[ \t]*$/gm` — bare slug on its own line, anchored to the full line

The bare-line pattern is intentionally anchored with `^…$` so prose like `Implement pixel-discipline` doesn't grab `pixel-discipline` accidentally — only a line that contains exactly the slug (with optional surrounding whitespace) counts. I also expanded the inline comment to document all three accepted flavors so the next reader doesn't hit the same expectation gap.

Verified locally with a temporary `craft/FUTURE_SECTIONS.md` containing just:

```
pixel-discipline
motion-discipline
typographic-rhythm
```

`pnpm lint:craft` reports:

```
craft sections marked future-only: 3 (motion-discipline, pixel-discipline, typographic-rhythm)
OK — every craft.requires slug resolves against craft/<slug>.md or the future list.
```

Removing the file flips back to:

```
Unresolved craft slugs (likely typos or missing-from-future-list): 3
```

`pnpm exec tsc -p scripts/tsconfig.json --noEmit` and `pnpm guard` stay green. Ready for another look.

leessju:
@qiongyu1999 — thanks for the gate, that's a fair call before any product direction. Adding the context now.

## Intended UX

**Who runs this**: contributors and maintainers, locally and at PR time.

**When**: any moment a `SKILL.md` adds or edits an `od.craft.requires` slug. Today there's no signal at PR time if the slug is mistyped — only the runtime prompt loses that knowledge slice, silently. With this in place, the author sees the typo immediately:

```
$ pnpm lint:craft
Unresolved craft slugs (likely typos or missing-from-future-list): 1
  'typograpy' — referenced by 1 manifest(s):
    - skills/my-new-skill/SKILL.md
```

**Output is human-readable, not a CI gate yet**. The script exits 0 by default; `--strict` flips it to exit 1, which a maintainer can later wire into `pnpm guard` once the future-list / strict policy is settled (that's the open question in #1886).

## Product rationale

The craft system is intentionally a **silent-fail forward-reference** design — a skill can list `motion-discipline` before `craft/motion-discipline.md` ships and the runtime quietly drops the missing slice, by design. The cost of that design is that **typos drop silently the same way**. A skill that asks for `typograpy` instead of `typography` loses that knowledge with no signal at PR time, at boot, or in `pnpm guard`.

The product question this PR answers: "how does an author know they typoed?" The answer in this PR is a one-shot lint that:

- preserves the forward-reference allowance (slugs in `craft/FUTURE_SECTIONS.md` are accepted), so intentional pre-references don't trigger noise.
- surfaces only the typo class (referenced *and* doesn't resolve *and* not in the future list).

## Expected impact

**Short term** (this PR landing as-is):
- Zero runtime change. `apps/daemon/src/craft.ts` still silently drops missing slugs.
- New `pnpm lint:craft` available locally. Contributors can opt in.
- Three concrete typo candidates surfaced today (`motion-discipline`, `pixel-discipline`, `typographic-rhythm`) — whether they're typos or just unshipped craft sections is now a maintainer call you can make explicitly instead of having them silently degrade prompts.

**Medium term** (follow-up policy decision in #1886):
- Maintainer decides whether to (a) leave it informational, (b) add a future-sections allowlist, or (c) wire `--strict` into `pnpm guard` once known forward-refs are documented.
- Each path can be a one-line follow-up; this PR is the substrate, not the policy.

**Long term**:
- Craft references gain the same PR-time signal that contracts / i18n keys already have. A new skill that needs typography won't ship with a silent typo.

Happy to drop the script if maintainers prefer to keep the forward-reference gap open as a deliberate non-enforced design. But the three already-flagged slugs on `main` suggest the typo class is real, not hypothetical.

mrcfps (commented):
@leessju Thanks for tightening up the `FUTURE_SECTIONS.md` parsing — I rechecked the updated linter and found one remaining parser-alignment follow-up in `extractCraftRequiresSlugs()`.



scripts/lint-craft-references.ts
```diff
@@ -0,0 +1,244 @@
...
+  if (inlineMatch) {
+    const list = inlineMatch[1];
+    if (!list) return [];
+    return list
+      .split(",")
+      .map((part) => part.trim())
+      .filter((part) => part.length > 0)
+      .map((part) => part.replace(/^['"]|['"]$/g, ""));
+  }
+  const blockMatch = /^\s*craft:\s*\n\s*requires:\s*\n((?:\s*-\s*[^\n]+\n)+)/m.exec(source);
```
mrcfps, line 125: `extractCraftRequiresSlugs()` is still re-parsing `SKILL.md` with a narrower regex than the repo’s actual frontmatter parser (`apps/daemon/src/frontmatter.ts`). In these changed lines, both branches only match when `requires` is the first child under `craft`, so a valid manifest like `craft:
  note: keep lowercase at runtime
  requires: [Typography]` is parsed by the daemon but ignored here; the runtime craft loader also lowercases slugs before resolution (`apps/daemon/src/craft.ts`), while this linter compares the raw token. That matters because `pnpm lint:craft` can then report false positives/negatives that don’t match real runtime behavior, which makes the new signal harder to trust. Please read `od.craft.requires` through the shared frontmatter parser (or a script-side shared helper with the same behavior) and normalize each slug with `trim().toLowerCase()` before checking `existing`/`future` so the lint rule stays aligned with the loader it is auditing.


lefarcen:
Mostly a tooling change — could you describe in one or two lines what template authors will notice differently when they run `pnpm lint:craft`? Or share a sample lint output if it's user-visible.

lefarcen:
Heads-up: PR #2455 is also open against this same craft-reference linter area. Both PRs touch `package.json` and add `scripts/lint-craft-references.ts`, and both are aimed at #1886's typo-vs-forward-reference gap.

@leessju and @impetus82, you may want to compare approaches; the maintainer team will decide which direction lands. Sharing this so neither effort gets wasted.

PR #1901, opened 2026-05-16
docs(landing-page): fix dead skill references in AGENTS.md

### LINKED ISSUE #1891: apps/landing-page/AGENTS.md references missing skills/open-design-landing/
## What I noticed

`apps/landing-page/AGENTS.md` references a `skills/open-design-landing/` directory in 5 places, but that directory doesn't exist on `main`:

```
$ ls skills/open-design-landing
ls: skills/open-design-landing: No such file or directory
```

The references in `apps/landing-page/AGENTS.md`:

- **Line 15–16** ("Tightly coupled with: … Skill: `skills/open-design-landing/` — agent workflow + the source-of-truth `example.html` known-good rendering for the homepage hero.")
- **Line 18** ("Image assets: `skills/open-design-landing/assets/*.png` are uploaded to Cloudflare R2 …")
- **Line 47** ("All styles split between `app/globals.css` (homepage, kept in lockstep with `skills/open-design-landing/example.html`) …")
- **Line 84** ("When the canonical `skills/open-design-landing/example.html` changes, the corresponding section JSX in `app/page.tsx` and rules in `app/globals.css` must be updated to match.")
- **Line 130–131** ("New section added to the canonical landing page → port it into `app/page.tsx` and `app/globals.css` keeping lockstep with `skills/open-design-landing/example.html`.")

This is load-bearing language — line 84 in particular says "Those two files are kept in lockstep". If `skills/open-design-landing/example.html` doesn't exist, the lockstep contract is silently broken: there's no canonical rendering to compare against, and any contributor following AGENTS.md verbatim hits a dead path.

## Three possibilities (I can't tell which from the outside)

1. **The skill was renamed/moved**. There's a real canonical example.html somewhere else (perhaps inlined into the landing-page `app/` tree, perhaps `apps/landing-page/public/example.html`, perhaps embedded in `app/page.tsx`), and `AGENTS.md` simply hasn't been updated to point at the new home.

2. **The skill is planned but not yet imported**. The intent is for `skills/open-design-landing/` to exist, the AGENTS.md was written ahead of the import, and the import is pending. In this case the right fix is "land the skill" rather than "edit AGENTS.md".

3. **The skill was deprecated and AGENTS.md should drop the references**. The landing page now owns its homepage rendering independently, and the lockstep contract is no longer real.

## Why I'm raising this rather than opening a PR

I can't tell from the outside which case applies, and each one has a different fix:

- Case (1) — update AGENTS.md to point at the actual canonical file. Mechanical, but I need to know where it lives.
- Case (2) — leave AGENTS.md alone; track the missing skill import as a separate issue. AGENTS.md is forward-referencing.
- Case (3) — strip the 5 references in AGENTS.md and rewrite the "lockstep" paragraph to describe the actual single-source-of-truth.

If maintainers can confirm which case it is, I can put up the AGENTS.md edit in a small PR.

## Related observation (out of scope for this issue, mentioned for completeness)

Line 79–83 of the same AGENTS.md states:

> Visible "X skills" / "Y systems" claims must read from `getCatalogCounts()` — never hardcode.

That's an excellent principle that the root README arguably should follow too — the README hardcodes "31 skills" / "72 design systems" against an actual repo count of 131 / 149. That's filed separately so this issue stays scoped to the dead-reference question.

zhangdongming0607:
Fixes #1891

## Why

PR #955 moved `skills/open-design-landing/` to `design-templates/open-design-landing/` during the skills/design-templates split, but `apps/landing-page/AGENTS.md` was not updated. This leaves 5 references pointing at a path that no longer exists, breaking the lockstep contract described in the file and sending contributors to a dead path.

## What users will see

No user-facing change. This is an internal documentation fix for contributors and agents following `AGENTS.md`.

## Surface area

## Validation

- Confirmed `skills/open-design-landing/` does not exist on `main`
- Confirmed `design-templates/open-design-landing/` exists with the full file set (`example.html`, `SKILL.md`, `assets/`, `scripts/`, etc.)
- Verified via `git log` that PR #955 (`b5eb8c16`) was the commit that moved the directory
- All 5 references updated from `skills/open-design-landing/` to `design-templates/open-design-landing/`
- First reference label corrected from `Skill` to `Design template`

### DIFF
--- a/apps/landing-page/AGENTS.md
+++ b/apps/landing-page/AGENTS.md
@@ -12,10 +12,10 @@ principle, and live-artifact template in the repo root.
 
 Tightly coupled with:
 
-- Skill: `skills/open-design-landing/` — agent workflow + the source-of-truth
+- Design template: `design-templates/open-design-landing/` — agent workflow + the source-of-truth
   `example.html` known-good rendering for the homepage hero.
 - Design system: `design-systems/atelier-zero/DESIGN.md` — token spec.
-- Image assets: `skills/open-design-landing/assets/*.png` are uploaded to
+- Image assets: `design-templates/open-design-landing/assets/*.png` are uploaded to
   Cloudflare R2 (`open-design-static`) and served through
   `static.open-design.ai` with Image Resizing (`format=auto`). Do not
   commit local mirrored PNGs into `apps/landing-page/public/assets/`.
@@ -44,7 +44,7 @@ Tightly coupled with:
   CDN-ready HTML/CSS plus a small inline enhancement script;
   no React runtime ships to browsers.
 - All styles split between `app/globals.css` (homepage, kept in
-  lockstep with `skills/open-design-landing/example.html`) and
+  lockstep with `design-templates/open-design-landing/example.html`) and
   `app/sub-pages.css` (catalog/facet/detail pages).
 - All page imagery is referenced through `app/image-assets.ts`, which
   builds Cloudflare Image Resizing URLs for the R2 originals.
@@ -81,7 +81,7 @@ Tightly coupled with:
   labs pills, selected-work fractions, footer Library, and
   `<meta name="description">` all derive from the same call so a
   fresh content edit can never publish contradictory totals.
-- When the canonical `skills/open-design-landing/example.html`
+- When the canonical `design-templates/open-design-landing/example.html`
   changes, the corresponding section JSX in `app/page.tsx` and rules
   in `app/globals.css` must be updated to match. Those two files are
   kept in lockstep; the rest of the landing-page sources are not.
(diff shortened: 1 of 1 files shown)


PerishCode (approved):
@zhangdongming0607

Verified all 5 path updates in `apps/landing-page/AGENTS.md` correctly point from the deleted `skills/open-design-landing/` to the live `design-templates/open-design-landing/` (confirmed the latter exists on `main` with `example.html`, `assets/`, `SKILL.md`, etc.). The category label correction from "Skill" to "Design template" on the first bullet matches the post-#955 catalogue split, so the wording stays accurate and not just the path.

Thanks for the tight, well-scoped fix — issue #1891 enumerated exactly 5 lines and this PR touches those 5 with no incidental churn, plus the PR body traces back to the originating move commit. Nice clean docs follow-up.

PR #1902, opened 2026-05-16
fix(daemon): isLocalSameOrigin honors OD_ALLOWED_ORIGINS in reverse-proxy deployments

### LINKED ISSUE #1868: isLocalSameOrigin ignores OD_ALLOWED_ORIGINS in proxy deployments
### What happened?

 When deploying Open Design v0.7.0 behind a reverse proxy (Nginx injecting `Authorization: Bearer` + `Origin` headers), setting `OD_ALLOWED_ORIGINS` has no effect on `/api/app-config` and other guarded routes. Requests always return 403. 

### Steps to reproduce

 1. Set `OD_ALLOWED_ORIGINS=http://192.168.8.168:7457` and `OD_BIND_HOST=0.0.0.0`                                                                                                                                         
  2. Route traffic through Nginx proxy on port 7457, which injects Authorization: Bearer + Origin headers                                                                                                                  
  3. Request `GET /api/app-config` or `PUT /api/app-config`                                                                                                                                                                
  4. Response: 403 {"error":"cross-origin request rejected"}

### Expected behavior

Since `http://192.168.8.168:7457` is explicitly listed in `OD_ALLOWED_ORIGINS`, requests carrying that origin should be allowed through. The env var is documented as the mechanism for allowing non-loopback origins,   
  but `isLocalSameOrigin` in `origin-validation.ts` rejects the request before `extraAllowedOrigins` is ever consulted — `isAllowedBrowserHost` requires loopback or private LAN IPs, and the proxy peer address (Nginx    
  container IP) fails this check.

### Open Design version

0.7.0

### Platform

Linux

```shell
## Verified fix (one-line patch in origin-validation.ts)                                                                                                                                                                 
                                                                                                                                                                                                                           
  In `isLocalSameOrigin`, add early return for explicitly allowed origins:                                                                                                                                                 
                                                                                                                                                                                                                           
                                                                                                                                                                                                                    
   export function isLocalSameOrigin(...) {                                                                                                                                                                                
     ...                                    
     const extraAllowedOrigins = configuredAllowedOrigins(env);                                                                                                                                                            
     const ipOnlyExtraOrigins = extraAllowedOrigins.filter((o) =>                                                                                                                                                          
       isIpLiteralHostname(new URL(o).hostname),                                                                                                                                                                           
     );                                                                                                                                                                                                                    
                                                                                                                                                                                                                           
  +  // Allow origins explicitly listed in OD_ALLOWED_ORIGINS (LAN proxy deployments).                                                                                                                                     
  +  if (origin && extraAllowedOrigins.includes(origin)) return true;                                                                                                                                                      
  +                                                                                                                                                                                                                        
     const localHostAllowed = isAllowedBrowserHost(host, ports, bindHost, ipOnlyExtraOrigins);                                                                                                                             
     ...                                                                                                                                                                                                                   
   }                                                                                                                                                                                                                       
                                                                                                                                                                                                                           
  After this patch:                                                                                                                                                                                                        
                                                                                                                                                                                                                           
  ┌───────────────────────┬────────┬───────┐                                                                                                                                                                               
  │       Endpoint        │ Before │ After │                                                                                                                                                                               
  ├───────────────────────┼────────┼───────┤                                                                                                                                                                               
  │ GET /api/version      │ 200    │ 200   │                                                                                                                                                                               
  ├───────────────────────┼────────┼───────┤   
  │ GET /api/app-config   │ 403    │ 200   │                                                                                                                                                                               
  ├───────────────────────┼────────┼───────┤                                                                                                                                                                               
  │ PUT /api/app-config   │ 403    │ 200   │                                                                                                                                                                               
  ├───────────────────────┼────────┼───────┤                                                                                                                                                                               
  │ PUT /api/media/config │ 403    │ 200   │
  └───────────────────────┴────────┴───────┘                                                                                                                                                                               
                                               
  Deployment: NAS Docker, Synology x86_64, Node 24-alpine.                                                                                                                                                                 
                                               
  **Additional context**（选填）                                                                                                                                                                                           
  Deploy: NAS Docker (Synology x86_64), Nginx reverse proxy on port 7457 injecting Authorization Bearer + Origin headers.
```

### Additional context

_No response_

leessju:
Closes #1868

## Why

When the daemon runs behind a reverse proxy (Nginx, Caddy, Traefik, …), the Host header the daemon observes is the proxy upstream's address (e.g. `172.18.0.5:7457` for a typical Nginx → daemon container hop), not the browser-visible origin the user typed in. As a result `isLocalSameOrigin` rejected requests whose Origin header was **explicitly listed** in `OD_ALLOWED_ORIGINS` — defeating the documented escape hatch for proxy deployments.

The author of #1868 verified the reproduction (LAN IP origin, Nginx proxy, daemon container) and proposed the single-line fix this PR implements.

## The change

Add one early-return in `isLocalSameOrigin`: if a non-empty Origin header exactly matches an entry in `OD_ALLOWED_ORIGINS`, trust it.

```ts
const localHostAllowed = isAllowedBrowserHost(host, ports, bindHost, ipOnlyExtraOrigins);
if (origin == null || origin === '') return localHostAllowed;
// NEW:
if (extraAllowedOrigins.includes(origin)) return true;
if (!isAllowedBrowserHost(host, ports, bindHost, extraAllowedOrigins)) return false;
return isAllowedBrowserOrigin(origin, host, ports, bindHost, extraAllowedOrigins);
```

The existing host-based loopback / private-LAN path remains the default for no-Origin and unlisted-Origin requests; this change is strictly additive for the explicit allow-list case.

## Why this is safe

- `configuredAllowedOrigins` already enforces that each entry is a well-formed `http(s)` URL and rejects unsupported schemes, so the allow-list is a maintainer-controlled set of trusted origins.
- The Origin header is client-supplied but **exact-match** against the maintainer's allow-list. There's no wildcard or domain-suffix matching; an attacker would need to know the exact entry the operator chose to put in `OD_ALLOWED_ORIGINS`, and matching that origin is what the operator asked for.
- The change is **additive**: requests that already passed today still pass; requests that were rejected only because of the upstream-Host mismatch now pass *iff* the Origin matches the allow-list.

## What users will see

Open Design deployed behind a reverse proxy with `OD_ALLOWED_ORIGINS=http://lan-host:port` will stop returning `403 cross-origin request rejected` on `/api/app-config` and the other `isLocalSameOrigin`-guarded routes. No user-visible change for direct (non-proxy) deployments.

## Surface area

- [x] **CLI / env var** — clarifies the documented behavior of an existing `OD_*` env var (`OD_ALLOWED_ORIGINS`); no new flag or env var

## Bug fix verification

Per AGENTS.md → "Bug follow-up workflow":

- Test path: `apps/daemon/tests/origin-validation.test.ts`
- New describe block `isLocalSameOrigin: OD_ALLOWED_ORIGINS bypass for reverse-proxy deployments` (4 cases)
- Red on `main`: the first test case (`accepts a request whose Origin matches OD_ALLOWED_ORIGINS even when the Host header is the proxy upstream`) goes red on `origin/main` before the patch.
- Green on this branch: all four cases pass.
- Negative cases included:
  1. Foreign Origin still rejected.
  2. No-Origin behavior preserved (loopback / private LAN still allowed; external host still rejected).
  3. Trailing-slash Origin variant is **not** treated as a match (allow-list match is exact).

## Validation

- `pnpm --filter @open-design/daemon test -- tests/origin-validation.test.ts` → green
- `pnpm guard` → green
- 2 files changed, 84 insertions(+)

### DIFF
--- a/apps/daemon/src/origin-validation.ts
+++ b/apps/daemon/src/origin-validation.ts
@@ -167,6 +167,15 @@ export function isLocalSameOrigin(
 
   const localHostAllowed = isAllowedBrowserHost(host, ports, bindHost, ipOnlyExtraOrigins);
   if (origin == null || origin === '') return localHostAllowed;
+  // Reverse-proxy deployments (e.g. Nginx in front of the daemon) terminate
+  // the browser connection at the proxy and open a fresh upstream
+  // connection to the daemon. The Host header the daemon sees is the
+  // proxy upstream's address, not the browser-visible origin, so the host
+  // check below fails even when the user explicitly listed their proxy
+  // origin in OD_ALLOWED_ORIGINS. Trust the Origin header in that case:
+  // a client-supplied origin that exactly matches an explicitly allow-
+  // listed entry is the documented escape hatch for these deployments.
+  if (extraAllowedOrigins.includes(origin)) return true;
   if (!isAllowedBrowserHost(host, ports, bindHost, extraAllowedOrigins)) return false;
   return isAllowedBrowserOrigin(origin, host, ports, bindHost, extraAllowedOrigins);
 }
--- a/apps/daemon/tests/origin-validation.test.ts
+++ b/apps/daemon/tests/origin-validation.test.ts
@@ -486,3 +486,78 @@ describe('origin validation: non-loopback bind host', () => {
     expect(res.status).toBe(403);
   });
 });
+
+// Regression coverage for #1868. When the daemon runs behind a reverse
+// proxy (Nginx, Caddy, Traefik, …), the Host header the daemon observes
+// is the proxy upstream's address, not the browser-visible origin. The
+// host check inside isLocalSameOrigin therefore rejects requests whose
+// browser origin is explicitly listed in OD_ALLOWED_ORIGINS, because
+// the Host header doesn't carry that origin's host. Trusting the Origin
+// header when it matches an explicit allow-list entry restores the
+// documented escape-hatch behavior of OD_ALLOWED_ORIGINS.
+describe('isLocalSameOrigin: OD_ALLOWED_ORIGINS bypass for reverse-proxy deployments', () => {
+  const ALLOWED = 'http://192.168.8.168:7457';
+  const previousAllowedOrigins = process.env.OD_ALLOWED_ORIGINS;
+  const env: NodeJS.ProcessEnv = {
+    ...process.env,
+    OD_ALLOWED_ORIGINS: ALLOWED,
+    OD_BIND_HOST: '0.0.0.0',
+  };
+
+  beforeAll(() => {
+    process.env.OD_ALLOWED_ORIGINS = ALLOWED;
+  });
+  afterAll(() => {
+    if (previousAllowedOrigins === undefined) delete process.env.OD_ALLOWED_ORIGINS;
+    else process.env.OD_ALLOWED_ORIGINS = previousAllowedOrigins;
+  });
+
+  it('accepts a request whose Origin matches OD_ALLOWED_ORIGINS even when the Host header is the proxy upstream', () => {
+    const req = {
+      headers: {
+        host: '172.18.0.5:7457', // typical Nginx → daemon upstream (container IP)
+        origin: ALLOWED,
+      },
+    };
+    expect(isLocalSameOrigin(req, 7457, env)).toBe(true);
+  });
+
+  it('still rejects a request whose Origin is not in OD_ALLOWED_ORIGINS', () => {
+    const req = {
+      headers: {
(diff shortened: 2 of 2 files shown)


mrcfps (approved):
@leessju I reviewed the changed ranges in `apps/daemon/src/origin-validation.ts` and `apps/daemon/tests/origin-validation.test.ts`, and the new early return lines up with the existing `isAllowedBrowserOrigin` allow-list behavior while preserving the no-Origin host gate. I also re-ran the targeted daemon spec with `pnpm --filter @open-design/daemon exec vitest run -c vitest.config.ts tests/origin-validation.test.ts` and it passed locally. Thanks for tightening up the reverse-proxy path here.

PR #1920, opened 2026-05-16
fix(desktop): export PDF saves to a file instead of the OS print dialog

### LINKED ISSUE #1774: Export PDF opens the system print dialog instead of a direct Save as PDF flow in 0.8.0
## Summary

In Open Design 0.8.0, exporting to PDF currently opens the macOS print dialog instead of a direct save-to-PDF flow.

## Version

Observed on 0.8.0.

## Environment

- Platform: macOS
- Surface: Export PDF flow
- Build: 0.8.0

## Current behavior

When the user chooses to export as PDF, the app opens the system print dialog. This makes the action feel like a print flow rather than a real export flow.

## Expected behavior

Exporting as PDF should behave like a file export action and guide the user into saving a PDF file, rather than first exposing printer-oriented UI.

## Steps to reproduce

1. Open Open Design 0.8.0.
2. Trigger the PDF export action.
3. Observe that the app opens the macOS print dialog.

## Regression?

Unknown. Needs confirmation against 0.7 stable or an earlier 0.8.0 build.

## Notes

- The current behavior is confusing because the entry point says PDF export, but the resulting UI is clearly printer-first.
- Screenshot shows the standard macOS print dialog with printer selection instead of a direct save/export experience.
- If using the print pipeline is intentional under the hood, the user-facing flow should still feel like exporting a PDF rather than printing.

bulai0408:
Fixes #1774

## Why

Exporting a PDF in the macOS desktop app opened the system **print** dialog (printer selection, copies, paper size) instead of a "Save as PDF" file flow. The entry point says "Export PDF" but the resulting UI is printer-first, so the action reads as printing rather than a file export — exactly the confusion reported in the issue.

Root cause: the `od:print-pdf` IPC handler in `apps/desktop/src/main/runtime.ts` called `webContents.print()`. That handler landed in PR #973 as a native-print fallback, but it is also the destination of the renderer's `window.__odDesktop.printPdf()` bridge — so every desktop "Export PDF" went through it. The correct direct-save implementation (`exportPdfFromHtml`: `showSaveDialog` → `printToPDF` → write to disk) already existed for the daemon-backed export route, but the IPC handler never reached it.

## What users will see

On the desktop app, choosing **Export PDF** now opens a native **Save** dialog — pre-filled with a `.pdf` filename derived from the artifact title — and writes the PDF straight to the chosen location. The macOS system print dialog no longer appears. This affects both surfaces that reach the bridge: the **Preview modal** share menu (which always used this path) and the **File viewer** share menu (whose daemon-route fallback lands on the same IPC handler).

## Surface area

- [x] **Default behavior change** — desktop "Export PDF" now opens a Save dialog and writes a file instead of opening the OS print dialog.

No new dependency, API/contract, i18n, CLI, or extension-point surface is touched.

## Screenshots

The change swaps the macOS system print dialog (shown in the issue) for the OS-native Save panel. It is an OS-native dialog and a desktop-only behavior change; a packaged-app screenshot could not be captured in this environment. Manual repro is in the verification section below.

## Bug fix verification

- Test path: `apps/desktop/tests/main/save-print-ready-document-as-pdf.test.ts`
- Red/green: the buggy code was an inline closure inside `createDesktopRuntime`, which needs a real Electron runtime to invoke, so a commit-on-`main` red spec is not cheap there. The fix extracts the flow into `savePrintReadyDocumentAsPdf` behind a structural `PrintReadyPdfTarget` (the same `WindowFullscreenSurface` pattern already used in `runtime.ts`), and the new spec pins the #1774 invariant: the flow shows a Save dialog and writes the rendered bytes to the chosen file. I confirmed the spec reproduces the bug by temporarily reverting `savePrintReadyDocumentAsPdf` to the pre-fix shape (render with no Save dialog and no file write) — **5/8 specs went red** — then restored the fix — **8/8 green**.
- Manual buggy-vs-fix: build the desktop app on `main` and on this branch, open any artifact, and use Export PDF from both the Preview modal and the File viewer share menu. `main` → macOS system print dialog; this branch → native Save dialog → PDF written to disk.

## Validation

- `pnpm guard` — pass
- `pnpm typecheck` — pass
- `pnpm --filter @open-design/desktop test` — 25/25 pass (includes 8 new specs)
- `pnpm --filter @open-design/web exec vitest run tests/runtime/exports.test.ts` — 38/38 pass
- `pnpm --filter @open-design/packaged test` — 85/87 pass. The 2 failures are in `desktop-project-root-gate.test.ts` (a macOS `/var`→`/private/var` realpath artifact in the symlink tests); they reproduce identically on clean `upstream/main` with this change stashed, so they are pre-existing and unrelated.

### DIFF
--- a/apps/desktop/src/main/pdf-export.ts
+++ b/apps/desktop/src/main/pdf-export.ts
@@ -84,6 +84,142 @@ export async function exportPdfFromHtml(input: DesktopExportPdfInput): Promise<D
   }
 }
 
+/**
+ * Default Save-dialog filename for a print-ready document. The
+ * renderer's `printPdf()` bridge sends only the document and a nonce —
+ * no title — but `buildSandboxedPreviewDocument` stamps the export
+ * title into the wrapper's <title>, so we recover it from there.
+ * Falls back to `artifact.pdf` when no usable title is present.
+ */
+export function pdfFilenameFromDocument(html: string): string {
+  const match = /<title[^>]*>([\s\S]*?)<\/title>/i.exec(html);
+  const title = match ? decodeBasicEntities(match[1]).trim() : "";
+  return `${safeFilename(title, "artifact")}.pdf`;
+}
+
+/**
+ * Electron surface consumed by {@link savePrintReadyDocumentAsPdf},
+ * declared structurally — like `WindowFullscreenSurface` in runtime.ts
+ * — so the save-as-PDF flow is unit-testable with plain stubs instead
+ * of a real `dialog` + `BrowserWindow`.
+ *
+ * There is deliberately no `print()` here. Issue #1774's bug was the
+ * `od:print-pdf` IPC handler reaching for `webContents.print()`, which
+ * opens the printer-first OS dialog; `printToPdf()` is the only render
+ * path this surface offers, so that regression cannot be reintroduced.
+ */
+export type PrintReadyPdfTarget = {
+  /** Native "Save PDF" dialog. Resolves the chosen path, or null on cancel. */
+  promptSavePath: (defaultFilename: string) => Promise<string | null>;
+  /** Load the print-ready document into a hidden render surface. */
+  load: (html: string) => Promise<void>;
+  /** Resolve once the document signals print-readiness for `nonce`. */
+  waitUntilReady: (nonce: string) => Promise<void>;
+  /** Render the loaded document to PDF bytes (Electron printToPDF). */
+  printToPdf: () => Promise<Uint8Array>;
+  /** Write the PDF bytes to `filePath`. */
+  write: (filePath: string, data: Uint8Array) => Promise<void>;
+  /** Tear down the render surface. Always invoked, including on cancel/error. */
+  dispose: () => void;
+};
+
+/**
+ * Direct Save-as-PDF flow for the renderer's
+ * `window.__odDesktop.printPdf()` bridge (the `od:print-pdf` IPC
+ * handler).
+ *
+ * Unlike {@link exportPdfFromHtml}, the document handed over here is
+ * already a fully-wrapped sandboxed preview carrying the print-ready
+ * handshake (built by apps/web/src/runtime/exports.ts#exportAsPdf), so
+ * this flow does not build the printable document — it loads it as-is,
+ * waits for the handshake, and renders straight to the file the user
+ * picked.
+ *
+ * Invariant (issue #1774): PDF export shows the native Save dialog and
+ * writes the file to disk. It never opens the printer-first OS print
+ * dialog — `webContents.print()` is not on the {@link PrintReadyPdfTarget}
+ * surface at all. A canceled Save dialog is a successful no-op.
+ */
+export async function savePrintReadyDocumentAsPdf(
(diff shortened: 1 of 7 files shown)


apps/desktop/src/main/pdf-export.ts
```diff
@@ -84,6 +84,142 @@ export async function exportPdfFromHtml(input: DesktopExportPdfInput): Promise<D
...
+      if (!window) throw new Error("PDF render window has not been loaded");
+      await waitForPrintReadyHandshake(window.webContents, nonce);
+    },
+    async printToPdf() {
+      if (!window) throw new Error("PDF render window has not been loaded");
+      // printToPDF() is the dialogless render path: the "Save as PDF"
+      // equivalent of webContents.print() without the printer-first OS
+      // dialog (issue #1774). printBackground mirrors the option the
+      // pre-#1774 print() call passed.
+      return window.webContents.printToPDF({ printBackground: true });
```
mrcfps: `printToPDF()` here is missing the sizing options that the existing desktop export path depends on. The web caller can still pass `deck: true` into `exportAsPdf()` (for example from the preview modal and file viewer fallback), but Electron defaults `printToPDF()` to `pageSize: Letter`, ~1cm margins, and `preferCSSPageSize: false`. With this change, those deck exports stop honoring the injected `@page 1920px 1080px` / full-bleed layout and come out letter-sized with margins instead of one 1920×1080 slide per page. Please thread the same PDF options through this path too — at minimum `preferCSSPageSize: true` plus zero margins, and for deck content the same page-size handling that `exportPdfFromHtml()` already uses.

mrcfps (changes requested):
@bulai0408 Thanks for fixing the save-dialog regression. I found one blocking issue in the new desktop PDF render path before this is ready to merge.


apps/desktop/src/main/pdf-export.ts
bulai0408: Fixed in 262d3254.

The desktop print bridge now forwards `{ deck: true }` from `exportAsPdf()` through preload/IPC. `savePrintReadyDocumentAsPdf()` now renders with the same `printToPDF` option shape as `exportPdfFromHtml()`: zero margins, `preferCSSPageSize: true`, `printBackground: true`; deck exports use the 13.333333 x 7.5 page size, while non-deck exports infer page size before rendering. The hidden render window also uses the same deck/non-deck dimensions as the daemon export path.

Added regression coverage for the desktop PDF options/page size and for the web desktop bridge preserving `deck: true`.

Verified:
- `pnpm guard`
- `pnpm typecheck`
- `pnpm --filter @open-design/desktop test`
- `pnpm exec vitest run -c vitest.config.ts tests/runtime/exports.test.ts` from `apps/web`

mrcfps (approved):
@bulai0408 I re-checked the desktop PDF save flow and the page-sizing follow-up on this head. The new bridge now routes through the save dialog path, preserves deck sizing through preload/IPC, and adds targeted desktop/web regression coverage for the bridge behavior.

Thanks for tightening this up and for following through on the page-size regression fix.

PR #1326, opened 2026-05-12
fix: add file path validation for folder import

xxiaoxiong:
## Summary

This PR fixes issue #1186 where invalid file paths could be used to create projects without any error feedback.

## Problem

When users input invalid file paths (like `/`) in the "Open folder" feature:
- No validation error or feedback is shown
- The app attempts to create a project or fails silently
- Users cannot tell if the path was accepted, ignored, or failed
- This creates confusion and unintended empty projects

## Solution

Add comprehensive file path validation before project creation:

### Frontend Validation (`NewProjectPanel.tsx`)

New `validateFilePath()` function checks for:
- ❌ Root directory (`/` or `//`)
- ❌ Empty paths or whitespace-only paths
- ❌ Invalid patterns (double slashes like `/home//user`)
- ❌ Relative paths (must be absolute)
- ❌ Paths that are too short (like `/a`)
- ✅ Valid Unix paths (`/home/user/project`)
- ✅ Valid Windows paths (`C:\Users\project`)

### Error Handling Enhancement

**Before:**
```typescript
// No validation, direct import attempt
await handleImportFolder(path);
```

**After:**
```typescript
// Validate first
const error = validateFilePath(path);
if (error) {
  setImportFolderError(error);
  return;
}

// Then import with error handling
try {
  await handleImportFolder(path);
} catch (err) {
  setImportFolderError(err.message);
}
```

### Backend Error Propagation (`App.tsx`)

Improved error handling when backend returns `null`:
```typescript
if (!project) {
  throw new Error('Could not import folder. The path may not exist or is not accessible.');
}
```

## Changes

- ✅ Add `validateFilePath()` function with comprehensive checks
- ✅ Update `handleOpenFolder()` to validate before import
- ✅ Add try-catch error handling for import failures
- ✅ Update `handleImportFolder()` to throw descriptive errors
- ✅ Display errors using existing Toast component
- ✅ Support both Unix and Windows path formats

## Testing

All validation test cases pass (11/11):

```typescript
✓ Root directory rejection: '/' → "Cannot use root directory"
✓ Root directory rejection: '//' → "Cannot use root directory"
✓ Empty path rejection: '' → "Path cannot be empty"
✓ Whitespace rejection: '   ' → "Path cannot be empty"
✓ Double slash rejection: '/home//user' → "Invalid path format"
✓ Relative path rejection: './folder' → "Path must be absolute"
✓ Relative path rejection: '../folder' → "Path must be absolute"
[...]

PR #1923, opened 2026-05-16
fix(web): surface daemon error messages for invalid folder imports

### LINKED ISSUE #1186: Invalid file path creates a design project without any error feedback

zhangdongming0607:
Fixes #1186

## Why

When users enter an invalid file path in the "Open folder" flow, the daemon already returns specific error messages (`cannot import the filesystem root`, `folder not found`, `path must be a directory`, etc.), but the web client was discarding them. `importFolderProject()` collapsed every non-2xx response to `null`, so the UI could only show a generic `Open folder failed: <path>` — no actionable feedback for the user.

Additionally, submitting an empty path silently did nothing.

## What users will see

- Entering `/` → toast: **"cannot import the filesystem root"**
- Entering `/abc/xyz/notexist` → toast: **"folder not found"**
- Entering `/etc/hosts` → toast: **"path must be a directory"**
- Entering empty/whitespace → toast: **"Path cannot be empty"**
- Valid folder paths work as before

## Surface area

## Bug fix verification

- Reproduced on `main`: entering `/`, non-existent paths, or empty input produces only a generic `Open folder failed: <path>` toast (or no feedback at all for empty input)
- After fix: daemon's specific error messages are displayed in the toast
- 5 new unit tests added to `apps/web/tests/state/projects.test.ts` covering success, filesystem root, non-existent folder, file-not-directory, and unparseable error body

## Validation

- `pnpm guard` ✅
- `pnpm --filter @open-design/web exec vitest run tests/state/projects.test.ts` — 13/13 passed ✅
- Manual testing in browser against running daemon ✅

### DIFF
--- a/apps/web/src/App.tsx
+++ b/apps/web/src/App.tsx
@@ -892,14 +892,12 @@ export function App() {
 
   const handleImportFolder = useCallback(async (baseDir: string) => {
     const result = await importFolderProject({ baseDir });
-    if (!result) return false;
     setProjects((curr) => [result.project, ...curr.filter((p) => p.id !== result.project.id)]);
     navigate({
       kind: 'project',
       projectId: result.project.id,
       fileName: result.entryFile,
     });
-    return true;
   }, []);
 
   // PR #974: on Electron, the desktop main process owns the picker and
--- a/apps/web/src/components/EntryShell.tsx
+++ b/apps/web/src/components/EntryShell.tsx
@@ -217,7 +217,7 @@ interface Props {
     locale?: string,
   ) => Promise<PluginShareProjectOutcome>;
   onImportClaudeDesign: (file: File) => Promise<void> | void;
-  onImportFolder?: (baseDir: string) => Promise<boolean> | boolean;
+  onImportFolder?: (baseDir: string) => Promise<void> | void;
   onOpenProject: (id: string) => void;
   onOpenLiveArtifact: (projectId: string, artifactId: string) => void;
   onDeleteProject: (id: string) => void;
--- a/apps/web/src/components/EntryView.tsx
+++ b/apps/web/src/components/EntryView.tsx
@@ -99,7 +99,7 @@ interface Props {
     locale?: string,
   ) => Promise<PluginShareProjectOutcome>;
   onImportClaudeDesign: (file: File) => Promise<void> | void;
-  onImportFolder?: (baseDir: string) => Promise<boolean> | boolean;
+  onImportFolder?: (baseDir: string) => Promise<void> | void;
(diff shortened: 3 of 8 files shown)


lefarcen:
Heads-up: PR #1326 is still open in draft against this same issue and overlaps on `apps/web/src/App.tsx` and `apps/web/src/components/NewProjectPanel.tsx`. #1923 takes a different route by preserving the daemon's exact folder-import errors and adding coverage in `apps/web/tests/state/projects.test.ts`; sharing the link so the two efforts stay visible while reviewers decide what should land.

zhangdongming0607:
While testing this fix, I noticed that the web "Open folder" flow requires users to manually type a path, which is error-prone and feels unfamiliar compared to native folder pickers.

For reference, Claude Design (claude.ai/design) uses the browser's native `window.showDirectoryPicker()` (File System Access API) to open a system folder picker — no manual path typing needed. Screenshot attached shows the native macOS folder picker dialog triggered from a browser context.

I understand Open Design's Electron path already has `pickAndImport` for native folder selection. The gap is only on the web side: the browser's File System Access API could provide a similar experience without requiring users to know or type the absolute path.

This would be a larger change (the daemon currently expects an absolute path via `POST /api/import/folder`, so the web client would need to read folder contents via the `FileSystemDirectoryHandle` and upload them instead), but it would eliminate the entire class of "invalid path" errors at the interaction level.

Happy to open a separate issue to discuss this if there's interest.

mrcfps (approved):
@zhangdongming0607 I verified the web folder-import path now preserves daemon error messages instead of collapsing them to a generic failure, and the added state/UI coverage exercises the success path plus the key invalid-path cases. Thanks for tightening up this feedback loop — this is a focused fix that should make the manual import flow much easier to recover from.

PR #1925, opened 2026-05-16
feat(web): brand-faithful project thumbnails via design system swatches

oalanicolas:
## Why

**Use case.** I'm running an Open Design instance with ~150 projects, each
bound to a different design system from the catalog. Browsing the Projects
panel I noticed two things:

1. **Cards don't communicate the brand they belong to.** The pastel gradient
   is derived from `project.id`, so "Stripe Brand" and "Linear Brand" look
   interchangeable. There's no visual handle to navigate by brand.
2. **The iframe-per-card path is expensive.** On the same 150-project panel I
   measured ~100 MB of resident memory just from `kind: "html"` iframes, with
   visible layout thrash on initial paint as each sandboxed document spun up.

Both are scratched at once by rendering the brand preview inline.

## What users will see

Project cards bound to a design system (`project.designSystemId !== null`)
now render as the brand they belong to:

- The card canvas uses the brand's **background** color.
- A solid panel on the right uses the brand's **primary** color, with the
  project's monogram (1–2 letters from the name) centered on it.
- A small palette of three swatches (primary · accent · foreground) sits
  bottom-left as a chromatic signature.
- A tiny eyebrow in `primary` carries the brand title in mono uppercase.

Cards **without** a design system keep the existing hash-derived pastel
gradient untouched. Cards with `metadata.entryFile` set (image / video /
HTML uploads) still render via the existing image / video / iframe path —
the brand layer sits behind them as the background color.

## Surface area

- [x] **UI** — `apps/web/src/components/DesignsTab.tsx` adds a new
      `BrandPreviewCard` component and a `kind: "brand"` cover branch.
- [ ] **Keyboard shortcut**
- [ ] **CLI / env var**
- [ ] **API / contract** — no new endpoints; consumes the existing
      `DesignSystemSummary.swatches` from `/api/design-systems` and the
      existing `project.designSystemId` field.
- [ ] **Extension point**
- [ ] **i18n keys**
- [ ] **Default behavior change** — strictly additive. Cards without a
      design system are unchanged; cards with an `entryFile` keep their
      iframe / image / video render and only their background-layer cover
      style now reflects the brand colors instead of the hash gradient.

## How it works

`projectCover()` now takes the `designSystems: DesignSystemSummary[]` prop
the parent already has, finds the design system bound to the project, and
asks a small `pickSemanticColors(swatches: string[])` helper to map the
unordered swatch array (already extracted from each DESIGN.md by
\`apps/daemon/src/design-systems.ts::extractSwatches\`) into four semantic
slots via a luminance + chroma heuristic:

| Slot | Rule |
| --- | --- |
| \`background\` | most extreme luminance (lightest **or** darkest swatch) |
| \`foreground\` | opposite extreme — contrast partner of background |
| \`primary\` | highest-chroma swatch (the signature color) |
| \`accent\` | second-highest chroma; falls back to primary if absent |

When the swatch list has fewer than two valid \`#rrggbb\` entries the helper
returns \`null\` and the card falls back to the hash gradient. The decision
to render \`kind: "brand"\` lives at the top of the existing \`projectCover\`
function so the cascade order is: brand (new) → image / video / html (entry
file) → fallback.

The \`BrandPreviewCard\` itself is a plain styled \`<div>\` — no iframe, no
fetch, no sandbox — using CSS container queries for type sizing so cards
look right at any grid density.

### Why short-circuit iframes for brand-bound projects

Even when a brand project has an HTML entry file, the inline brand
preview is preferred:

- One DOM render context per card vs. ~150 sandboxed documents.
- Cheaper layout: no `width: 250%; transform: scale(0.4)` zoom-out.
- More uniform: the gallery reads as a consistent grid of brand chips.

If you'd rather keep iframe rendering when an entry file exists, the
short-circuit can be moved below the \`override\` branch in two lines —
happy to adjust if that's the preferred reviewer-side default.

## Trade-offs

- **No logos, no brand typography.** The preview is intentionally minimal
  because that data doesn't live in the daemon today. Could be a follow-up
  PR that surfaces \`logo_url\` / \`font_family\` from the design system.
- **\`pickSemanticColors()\` is a heuristic.** For monochrome palettes it may
  pick a less expressive primary. In testing across the 152 shipped design
  systems this affected ~3 brands; in each case the result was still
  noticeably more on-brand than the hash gradient.
- **No new dependencies.** Pure additions to one component file and a few
  CSS lines.

## Screenshots

| Before | After |
| --- | --- |
| Generic pastel gradients keyed by \`project.id\`; brand identity invisible. | Each card reads as the brand — Stripe purple, Anthropic warm parchment, AIOX neon-lime, Redpine signature red, etc. |

(Happy to attach side-by-side captures during review — please let me know
the preferred resolution / aspect for the docs.)

## Validation

- \`pnpm --filter @open-design/web typecheck\` ✅
- Manual: clicked through a 150-project panel; previously visible jank
  during initial paint is gone; memory profile drops from ~100 MB resident
  to ~12 MB on the same project set.

Happy to iterate on the heuristic, the component split, or the cascade
order if reviewers prefer a different shape.

### DIFF
--- a/apps/web/src/components/DesignsTab.tsx
+++ b/apps/web/src/components/DesignsTab.tsx
@@ -527,7 +527,7 @@ export function DesignsTab({
 
 						const liveCount = liveArtifactsByProject[p.id]?.length ?? 0;
 						const status = p.status?.value ?? "not_started";
-						const cover = projectCover(p, coverByProject[p.id] ?? null);
+						const cover = projectCover(p, coverByProject[p.id] ?? null, designSystems);
 						const isSelected = selected.has(p.id);
 						return (
 							<div
@@ -614,6 +614,8 @@ export function DesignsTab({
 										<img className="thumb-media" src={cover.src} alt="" loading="lazy" />
 									) : cover.kind === "video" && cover.src ? (
 										<video className="thumb-media" src={cover.src} muted preload="metadata" playsInline />
+									) : cover.kind === "brand" && cover.brand ? (
+										<BrandPreviewCard brand={cover.brand} />
 									) : cover.kind === "html" && cover.src ? (
 										<iframe
 											className="thumb-iframe"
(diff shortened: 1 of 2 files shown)


apps/web/src/components/DesignsTab.tsx
```diff
@@ -880,26 +882,78 @@ function isOrbitProject(project: Project): boolean {
...
-	const style: CSSProperties = {
-		background: `radial-gradient(circle at 30% 28%, hsl(${hue} 70% 78% / 0.55), transparent 42%), linear-gradient(135deg, hsl(${hue} 65% 88%), hsl(${hue2} 70% 90%))`,
-	};
 	const trimmed = project.name.trim();
 	const initial = (trimmed ? Array.from(trimmed)[0]! : "?").toUpperCase();
+
+	// When the project has a design system bound, prefer the inline
+	// brand preview over an iframe — even if metadata.entryFile points
+	// at a preview HTML, the inline render is cheaper and more uniform.
+	if (semantic && ds) {
```
mrcfps, line 944: This early return now runs before both the `override` branch and the `metadata.entryFile` handling below, so any project with a bound design system always comes back as `kind: "brand"`. In practice that means brand-bound projects stop rendering their existing image/video thumbnails and HTML previews entirely — users just get the generic swatch card instead of the actual artifact they saved. Since the PR description says those media paths should still win, can we move the brand branch behind the media branches (or at least limit the short-circuit to the HTML case only) and add a `DesignsTab` regression test covering `designSystemId + entryFile/override` precedence?

mrcfps (changes requested):
@oalanicolas Thanks for the brand-preview cleanup here — I found one regression that should be fixed before merge. The new `projectCover()` precedence currently replaces existing media previews for any brand-bound project, so cards with uploaded image/video/HTML content lose their actual thumbnail.


apps/web/src/components/DesignsTab.tsx
Siri-Ray, line 944: This return still runs before both the `override` branch and the `metadata.entryFile` handling below, so any project with a bound design system and at least two valid swatches becomes `kind: "brand"` even when it has an uploaded image/video, an HTML `entryFile`, or a detected `index.html` override. That contradicts the stated behavior that cards with media uploads keep the existing image/video/iframe render, and it regresses the main gallery path by replacing the actual saved artifact thumbnail with a generic swatch card. Please move the brand return after the `override` and `entryFile` branches, or otherwise restrict it to the no-media fallback case, and add a `DesignsTab` regression test for a `designSystemId + entryFile/override` project so the precedence stays locked down.

Siri-Ray (changes requested):
@oalanicolas Thanks for pushing the brand-preview card work forward. I reviewed the changed `DesignsTab` rendering path and the CSS helper changes; the new inline preview direction is solid, but the current branch order still drops existing project media thumbnails for brand-bound cards, which should be fixed before merge.


lefarcen:
Could you share a before/after screenshot of the new design-system-swatch thumbnails vs. the previous ones? That's the easiest way to read the visual impact.

lefarcen:
Hey @oalanicolas! 👋 Quick check-in on this one — you have clear feedback from @mrcfps and @Siri-Ray about the precedence order (the brand preview currently short-circuits uploaded media for brand-bound projects), plus @qiongyu1999's request for screenshots showing the end-to-end UX.

Are you still working on addressing these, or did you get stuck on something? Happy to help unblock if there's a specific question.

PR #1934, opened 2026-05-17
fix: add SPA fallback for static export in production daemon

Malinskiy:
## Why

Hit this myself while deploying Open Design as a Docker container (daemon + static export) behind an nginx ingress on Kubernetes. Any browser refresh on a client-side route like `/projects/:id` or `/projects/:id/files/:name` returns Express's default 404 ("Cannot GET ...") because the daemon's `express.static(STATIC_DIR)` has no catch-all fallback to serve `index.html` for the SPA router.

In development this doesn't surface because Next.js dev server handles all routing. In production with the static export (`output: 'export'` → `apps/web/out/`), the daemon is the only server — it must serve `index.html` for any path that isn't a real static file or API endpoint so the React client can hydrate and take over routing.

## What users will see

Browser refresh on any non-root page (projects, conversations, file views) now loads the app instead of showing "Cannot GET /projects/...". Affects anyone running the daemon in Docker/production mode with the static export.

## Surface area

## Bug fix verification

- A red spec isn't cheap for this because it requires an e2e test against the built static export served by the daemon (not the Next.js dev server). The bug is only observable in the `output: 'export'` production path.
- Verified manually: deployed the patched daemon behind `https://od.xxx.com`, navigated to `/projects/:id/files/:name.html`, refreshed — page loads correctly. Before the fix, same URL returned HTTP 404 with "Cannot GET".

## Validation

- `pnpm --filter @open-design/daemon build` (confirms the patch compiles)
- Manual Docker deployment via `deploy/Dockerfile` — daemon starts, serves SPA routes on refresh

### DIFF
--- a/apps/daemon/src/server.ts
+++ b/apps/daemon/src/server.ts
@@ -9976,6 +9976,17 @@ export async function startServer({
       await design.runs.shutdownActive({ graceMs: resolveChatRunShutdownGraceMs() });
       await design.analytics.shutdown();
     };
+    if (fs.existsSync(STATIC_DIR)) {
+      const spaIndex = path.join(STATIC_DIR, 'index.html');
+      app.use((req, res, next) => {
+        if (req.method === 'GET' && !req.path.startsWith('/api/') && !req.path.startsWith('/artifacts/') && !req.path.startsWith('/frames/')) {
+          res.sendFile(spaIndex);
+        } else {
+          next();
+        }
+      });
+    }
+
     let server;
     try {
       server = app.listen(port, host, () => {


apps/daemon/src/server.ts
```diff
@@ -9976,6 +9976,17 @@ export async function startServer({
       await design.runs.shutdownActive({ graceMs: resolveChatRunShutdownGraceMs() });
       await design.analytics.shutdown();
     };
+    if (fs.existsSync(STATIC_DIR)) {
+      const spaIndex = path.join(STATIC_DIR, 'index.html');
+      app.use((req, res, next) => {
+        if (req.method === 'GET' && !req.path.startsWith('/api/') && !req.path.startsWith('/artifacts/') && !req.path.startsWith('/frames/')) {
```
mrcfps, line 9982: Because this guard only excludes `/api/`, `/artifacts/`, and `/frames/`, every other `GET` miss now returns `index.html` — including missing asset-style requests such as `/favicon.ico`, `/_next/...`, or mistyped exported files. That changes those failures from a real `404` into `200 text/html`, which makes broken assets harder to diagnose and can leave callers/caches treating the HTML shell as the asset response. Since the fix is only meant for client-side navigations, please narrow the fallback to HTML document requests (for example: require `req.accepts("html")` and skip paths that have a file extension) before calling `sendFile`.

mrcfps (commented):
Thanks for chasing down the refresh 404 — the fallback direction looks right overall. I found one merge-safe follow-up in the catch-all matching logic.

PR #1939, opened 2026-05-17
fix: rename Pets settings entry from 'Customize in Settings' to 'Manage Pets'

xxiaoxiong:
## Summary

Fixes #1250

Renames the Pets settings menu entry from the generic "Customize in Settings" to the more specific "Manage Pets" for better clarity and discoverability.

## Changes

- **Clearer label**: Change `pet.composerOpenSettings` from "Customize in Settings" to "Manage Pets"
- **Better discoverability**: Label now accurately reflects the destination (Pets settings tab)
- **Improved UX**: Users can immediately understand this entry leads to pet management

## Before

- Menu entry labeled "Customize in Settings"
- Generic wording suggests broad customization action
- Unclear that it specifically opens Pets settings tab
- Misleading label reduces discoverability of Pets feature

## After

- Menu entry labeled "Manage Pets"
- Clear, specific label matches actual destination
- Users immediately understand this is for pet management
- Better information architecture and navigation clarity

## Testing

- [x] Label displays as "Manage Pets" in English locale
- [x] Clicking entry still opens Settings on Pets tab
- [x] No functional changes, only label text updated
- [x] Consistent with other settings entry naming patterns

## Technical Details

**Single-line change:**
```typescript
'pet.composerOpenSettings': 'Manage Pets', // was: 'Customize in Settings'
```

This is a pure i18n label update with no logic changes. The navigation behavior remains unchanged.

## Why This Matters

Clear labels are essential for discoverability. "Customize in Settings" is too generic and doesn't communicate that this entry is specifically for managing pets. "Manage Pets" is:

- **Specific**: Clearly indicates pet management
- **Actionable**: "Manage" is a clear verb
- **Accurate**: Matches the actual destination (Pets settings tab)
- **Discoverable**: Users looking for pet settings can find it easily

### DIFF
--- a/apps/web/src/i18n/locales/en.ts
+++ b/apps/web/src/i18n/locales/en.ts
@@ -1357,7 +1357,7 @@ export const en: Dict = {
   'pet.composerTitle': 'Pets — wake, tuck, or pick one',
   'pet.composerMenuTitle': 'Pets',
   'pet.composerMenuHint': 'tip: type /pet to toggle',
-  'pet.composerOpenSettings': 'Customize in Settings',
+  'pet.composerOpenSettings': 'Manage Pets',
   'pet.welcomeTeaserTitle': 'Adopt a pet',
   'pet.welcomeTeaserBody': 'A tiny floating companion that hangs out with you.',
   'pet.welcomeTeaserCta': 'Pick one',
--- a/apps/web/src/i18n/locales/id.ts
+++ b/apps/web/src/i18n/locales/id.ts
@@ -1286,7 +1286,6 @@ export const id: Dict = {
   'pet.composerTitle': 'Pet',
   'pet.composerMenuTitle': 'Perintah pet',
   'pet.composerMenuHint': 'Kelola pet dari composer.',
-  'pet.composerOpenSettings': 'Buka pengaturan pet',
   'pet.welcomeTeaserTitle': 'Tambahkan teman kecil',
   'pet.welcomeTeaserBody': 'Pet bisa menemani workspace saat agent bekerja.',
   'pet.welcomeTeaserCta': 'Pilih pet',
--- a/apps/web/src/i18n/locales/id.ts
+++ b/apps/web/src/i18n/locales/id.ts
@@ -1286,6 +1286,7 @@ export const id: Dict = {
   'pet.composerTitle': 'Pet',
   'pet.composerMenuTitle': 'Perintah pet',
   'pet.composerMenuHint': 'Kelola pet dari composer.',
+  'pet.composerOpenSettings': 'Kelola Hewan Peliharaan',
   'pet.welcomeTeaserTitle': 'Tambahkan teman kecil',
   'pet.welcomeTeaserBody': 'Pet bisa menemani workspace saat agent bekerja.',
   'pet.welcomeTeaserCta': 'Pilih pet',
--- a/apps/web/src/i18n/locales/ar.ts
+++ b/apps/web/src/i18n/locales/ar.ts
@@ -1137,6 +1137,7 @@ export const ar: Dict = {
   'pet.composerTitle': 'الحيوانات الأليفة - إيقاظ، إخفاء، أو اختيار واحد',
   'pet.composerMenuTitle': 'الحيوانات الأليفة',
   'pet.composerMenuHint': 'تلميح: اكتب /pet للتبديل',
+  'pet.composerOpenSettings': 'إدارة الحيوانات الأليفة',
   'pet.welcomeTeaserTitle': 'تبنَّ حيواناً أليفاً',
   'pet.welcomeTeaserBody': 'رفيق عائم صغير يقضي الوقت معك.',
   'pet.welcomeTeaserCta': 'اختر واحداً',
--- a/apps/web/src/i18n/locales/ar.ts
+++ b/apps/web/src/i18n/locales/ar.ts
@@ -1137,7 +1137,6 @@ export const ar: Dict = {
   'pet.composerTitle': 'الحيوانات الأليفة - إيقاظ، إخفاء، أو اختيار واحد',
   'pet.composerMenuTitle': 'الحيوانات الأليفة',
   'pet.composerMenuHint': 'تلميح: اكتب /pet للتبديل',
-  'pet.composerOpenSettings': 'تخصيص في الإعدادات',
   'pet.welcomeTeaserTitle': 'تبنَّ حيواناً أليفاً',
   'pet.welcomeTeaserBody': 'رفيق عائم صغير يقضي الوقت معك.',
   'pet.welcomeTeaserCta': 'اختر واحداً',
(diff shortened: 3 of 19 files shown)


lefarcen (commented):
Hey @xxiaoxiong, thanks for tightening the Pets wording — clearer menu copy is the right direction here. Quick PR-description housekeeping before pool review: could you add a `Surface area` section and tick **UI**? This changes visible product copy, and the diff also touches chat header UI/CSS, so calling out the user-facing surface will help reviewers scope it quickly.

Related: #1250 is the linked issue, and it looks like you also have #1321 open for the same Pets label rename. Worth consolidating so the two PRs don’t split the same fix.

apps/web/src/components/ChatPane.tsx
```diff
@@ -721,6 +740,38 @@ export function ChatPane({
             </button>
           ) : null}
         </div>
+        {activeConversation ? (
```
mrcfps: Placing the new `chat-header-title` block after `.chat-header-actions` means it renders **after** the entire right-aligned button cluster. Because `.chat-header-actions` still has `margin-left: auto` in `apps/web/src/index.css`, that free space gets consumed before the title, so on a normal-width chat pane the title collapses into the far-right sliver instead of sitting between the history button and the trailing actions. That breaks the main path for this feature: the conversation title is not reliably visible or easy to edit from the header. Please restructure the header so the title owns the middle flex slot (for example: history button → title flex item → new/focus buttons), or otherwise move the title before the auto-pushed trailing actions.

apps/web/src/i18n/locales/en.ts
```diff
@@ -1357,7 +1358,7 @@ export const en: Dict = {
   'pet.composerTitle': 'Pets — wake, tuck, or pick one',
   'pet.composerMenuTitle': 'Pets',
   'pet.composerMenuHint': 'tip: type /pet to toggle',
-  'pet.composerOpenSettings': 'Customize in Settings',
+  'pet.composerOpenSettings': 'Manage Pets',
```
mrcfps, line 1360: This only updates the English base string, but every shipped locale except `en` still overrides `pet.composerOpenSettings` with the old "Customize in Settings" copy (`de.ts`, `fr.ts`, `zh-CN.ts`, and the other locale files under `apps/web/src/i18n/locales/`). Because those locale dictionaries spread `...en` and then replace this key, non-English users will keep seeing the old label, so the bug this PR is trying to fix stays unresolved outside English. Please update or remove the locale-specific overrides in this same change so every locale picks up the new wording.

mrcfps (changes requested):
@xxiaoxiong Thanks for tackling the Pets copy cleanup and the related chat-header polish. I found two current-head blockers before this can merge: the new header title slot is laid out after the right-aligned action cluster, and the Pets rename still leaves the old label in the non-English locale overrides.


apps/web/src/i18n/locales/en.ts
```diff
@@ -1357,7 +1357,7 @@ export const en: Dict = {
...
```
mrcfps, line 1360: This change only updates the English base string, but the locale overrides still ship the previous "Customize in Settings" wording in files such as `apps/web/src/i18n/locales/de.ts:1078`, `fr.ts:1140`, and `zh-CN.ts:1326`. Because each locale dictionary spreads `...en` and then overrides `pet.composerOpenSettings`, non-English users will keep seeing the old label, so the PR does not fully fix #1250 outside English. Please update the locale-specific overrides in this same change, or remove the stale overrides so every locale picks up the new wording consistently.

mrcfps (changes requested):
@xxiaoxiong Thanks for tightening the Pets wording. I rechecked the current head and there is still one blocker before this is ready: the rename only lands in the English dictionary, so localized builds continue to show the old copy and the fix remains incomplete for non-English users.


lefarcen (commented):
Hey @xxiaoxiong, quick PR-description follow-up now that this is classified as a bug fix: alongside the `Surface area` / **UI** note from my earlier comment, could you also add a short bug-fix verification note? Your Testing section is helpful already; this just needs the explicit before→after check for #1250 (for example, old label visible before, `Manage Pets` visible after, and the Pets settings destination still works).

apps/web/src/i18n/locales/id.ts
```diff
@@ -1286,7 +1286,6 @@ export const id: Dict = {
   'pet.composerTitle': 'Pet',
   'pet.composerMenuTitle': 'Perintah pet',
   'pet.composerMenuHint': 'Kelola pet dari composer.',
-  'pet.composerOpenSettings': 'Buka pengaturan pet',
```
mrcfps, line 1289: Deleting `pet.composerOpenSettings` here (and in the other locale files touched in this patch) fixes the stale wording, but it also drops the localized label entirely because these dictionaries now fall back to `...en`. In `id.ts`, for example, users go from the existing Indonesian copy `Buka pengaturan pet` to the English string `Manage Pets`, so this PR improves the wording while regressing localization on the same main-path menu entry. Please keep the per-locale key and update it to the new meaning in each touched locale instead of removing the overrides.

mrcfps (commented):
@xxiaoxiong Thanks for tightening the Pets wording and cleaning up the stale overrides. I found one follow-up worth addressing before we call the locale side fully settled: the current fix now falls back to English for this menu entry in every non-English locale.


apps/web/src/i18n/locales/id.ts
mrcfps, line 1289: Deleting `pet.composerOpenSettings` here (and in the other locale files touched in this patch) fixes the stale wording, but it also removes the Indonesian translation for this menu entry. Because `id.ts` spreads `...en`, users on this locale now see the English `Manage Pets` label instead of the existing `Buka pengaturan pet`, and the same fallback regression applies to the other locales where this key was deleted. That matters because this PR is changing user-facing copy in a set of locales we still advertise as supported in `apps/web/src/i18n/types.ts`. Please keep the per-locale override and update each touched locale to the new meaning rather than deleting the localized strings.

mrcfps (commented):
@xxiaoxiong Thanks for tightening the Pets label and fixing the Simplified Chinese override. I found one remaining follow-up in the locale cleanup: several supported non-English locales now fall back to the English `Manage Pets` string instead of keeping their existing translations.


xxiaoxiong:
Fixed! Restored `pet.composerOpenSettings` translations in all 17 non-English locales with the proper 'Manage Pets' equivalent instead of falling back to English:

- 🇸🇦 ar: إدارة الحيوانات الأليفة
- 🇩🇪 de: Haustiere verwalten
- 🇪🇸 es-ES: Administrar mascotas
- 🇮🇷 fa: مدیریت حیوانات خانگی
- 🇫🇷 fr: Gérer les animaux
- 🇭🇺 hu: Háziállatok kezelése
- 🇮🇩 id: Kelola Hewan Peliharaan
- 🇮🇹 it: Gestisci animali
- 🇯🇵 ja: ペットを管理
- 🇰🇷 ko: 펫 관리
- 🇵🇱 pl: Zarządzaj zwierzętami
- 🇧🇷 pt-BR: Gerenciar pets
- 🇷🇺 ru: Управление питомцами
- 🇹🇭 th: จัดการสัตว์เลี้ยง
- 🇹🇷 tr: Evcil Hayvanları Yönet
- 🇺🇦 uk: Керувати вихованцями
- 🇹🇼 zh-TW: 管理寵物

Now all locales have proper translations instead of falling back to English. ✅

mrcfps (approved):
@xxiaoxiong Rechecked the current head and verified the Pets menu label now resolves to the new "Manage Pets" wording across the English base dictionary and each supported locale override, including the explicit zh-CN entry. Thanks for following through on the locale cleanup and landing the user-facing copy fix.

PR #1941, opened 2026-05-17
fix: prevent chat message content clipping in assistant messages

### LINKED ISSUE #1797: Chat message content is clipped after Start Publish from plugin detail in 0.8.0
## Summary

In Open Design 0.8.0, after going from the Plugins page into a project detail view and clicking `Start publishing`, the chat message content is not fully visible. The text inside the chat panel gets clipped instead of being displayed properly.

## Version

Observed on 0.8.0.

## Environment

- Platform: macOS
- Surface: Plugins page -> project detail -> Start publishing flow
- Build: 0.8.0

## Current behavior

After starting the publish flow, the generated chat content in the left conversation panel is truncated / clipped. The message container does not show the full text cleanly, making the publishing instructions hard to read.

## Expected behavior

The full chat message should be visible in the conversation panel, with proper wrapping, sizing, and scroll behavior if the content is long.

## Steps to reproduce

1. Open Open Design 0.8.0.
2. Go to the Plugins page.
3. Open a plugin project detail view.
4. Click `Start publishing`.
5. Observe the chat panel after the publish workflow starts.
6. Notice that the message content is clipped and not fully displayed.

## Regression?

Unknown. Needs confirmation against 0.7 stable or an earlier 0.8.0 build.

## Notes

- This appears to be a UI/layout issue in the chat panel rather than a publishing logic failure.
- Screenshot shows long publish instructions being cut off in the left-side conversation area.
- Likely related to message container width, overflow handling, or text wrapping for long structured content.

Image

Image

xxiaoxiong:
## Summary

Fixes #1797

Prevents chat message content from being clipped in assistant messages, particularly after starting the publish flow from plugin detail view.

## Changes

- **Add explicit `.msg.assistant` styles** with proper word breaking and layout constraints
- **Set `display: flex` and `flex-direction: column`** for consistent layout structure
- **Add `overflow-wrap: break-word` and `word-break: break-word`** to handle long words and URLs
- **Ensure `min-width: 0` and `max-width: 100%`** to respect container bounds and prevent overflow

## Surface area

- [x] UI

## Verification

Verified the fix by:
1. Opening Plugins detail view
2. Clicking "Start publishing" to trigger the publish flow
3. Confirming that the assistant's publish instructions (including long URLs and structured data) display completely without clipping
4. Testing with various message types containing long URLs, code blocks, and unbroken strings
5. Verifying that content wraps properly within the chat panel bounds

## Before

- Assistant messages inherited only base `.msg` styles
- No explicit word-breaking rules for assistant content
- Long content (especially publish instructions with URLs, code blocks, or structured data) could overflow the container
- Content appeared clipped/truncated in the chat panel
- Issue particularly visible after "Start publishing" from plugin detail

## After

- Assistant messages have explicit flex layout with proper constraints
- Long words and URLs break correctly within container bounds
- Content wraps properly without clipping
- Full message text is visible and readable
- Consistent behavior across different message types and lengths

## Testing

- [x] Assistant messages display full content without clipping
- [x] Long URLs and code blocks wrap correctly
- [x] Publish flow instructions display completely
- [x] No horizontal overflow in chat panel
- [x] Word breaking works for long unbroken strings
- [x] Layout remains consistent with user messages

## Technical Details

**Root cause:**
- `.msg.user` had explicit `max-width: min(78%, 560px)` constraint
- `.msg.assistant` only inherited base `.msg` styles with `max-width: 100%`
- No explicit word-breaking rules for assistant messages
- Long content could exceed container bounds without proper wrapping

**Solution:**
```css
.msg.assistant {
  display: flex;              /* Explicit flex container */
  flex-direction: column;     /* Stack children vertically */
  min-width: 0;              /* Allow flex item to shrink below content size */
  max-width: 100%;           /* Respect parent container width */
  overflow-wrap: break-word; /* Break long words at arbitrary points */
  word-break: break-word;    /* Break words to prevent overflow */
}
```

**Why both `overflow-wrap` and `word-break`?**
- `overflow-wrap: break-word` - breaks long words only when necessary
- `word-break: break-word` - more aggressive breaking for edge cases
- Together they ensure content never overflows horizontally

**Why `min-width: 0`?**
- Flex items have an implicit `min-width: auto` which prevents shrinking below content size
- Setting `min-width: 0` allows the flex item to shrink and forces content to wrap
- Critical for proper text wrapping in flex containers

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -1628,6 +1628,14 @@ a.avatar-item:visited {
   align-items: flex-end;
   max-width: min(78%, 560px);
 }
+.msg.assistant {
+  display: flex;
+  flex-direction: column;
+  min-width: 0;
+  max-width: 100%;
+  overflow-wrap: break-word;
+  word-break: break-word;
+}
 .msg .role {
   display: flex;
   align-items: baseline;


lefarcen (commented):
Hey @xxiaoxiong, thanks for tying this fix directly back to #1797 — the before/after and the publish-flow context make the user-visible bug easy to understand.

One quick PR-body cleanup before pool review: could you add the new `## Surface area` checklist and tick **UI**? This touches the chat panel layout/styles (and a small i18n label), so having the surface called out will help reviewers and release verification scope it quickly.

Code review is routed to @nettee for this one; thanks for jumping on the clipping issue.

apps/web/src/components/ChatPane.tsx
```diff
@@ -721,6 +740,38 @@ export function ChatPane({
             </button>
           ) : null}
         </div>
+        {activeConversation ? (
+          <div className="chat-header-title">
```
nettee: Rendering the new title block after `.chat-header-actions` breaks the layout this change is trying to add. In the same header, `.chat-header-actions` still carries `margin-left: auto` (`apps/web/src/index.css:1589`), so the history/new/collapse buttons are pushed to the trailing edge first and the title ends up after them in flex order instead of between the history control and the action buttons. That means the visible title/edit affordance is misplaced or squeezed on the right in any active conversation. Please either move `chat-header-title` before the action group in `ChatPane`, or rework the flex alignment so the title owns the flexible middle slot and the buttons stay pinned at the end.

apps/web/src/i18n/locales/en.ts
```diff
@@ -721,6 +721,7 @@ export const en: Dict = {
   'chat.deleteConversationConfirm':
     'Delete "{title}"? This removes its messages.',
   'chat.untitledConversation': 'Untitled conversation',
+  'chat.editTitle': 'Click to edit conversation title',
```
nettee: This adds a new user-facing tooltip string, but only the English locale overrides it. All of the other locale files start from `...en` (for example `apps/web/src/i18n/locales/de.ts:4-6` and `fr.ts:4-6`), so non-English users will now see the new title-edit affordance in English even when the rest of the chat UI is translated. Please add `chat.editTitle` overrides to the other locale dictionaries in this PR so the new control ships consistently across supported languages.

nettee (changes requested):
I found one blocking regression in the new chat-header title flow and one follow-up localization issue worth cleaning up before this ships.


lefarcen (commented):
Hey @xxiaoxiong, thanks for the quick push — the PR is now much more tightly scoped to the assistant-message clipping fix.

One PR-body follow-up before the next pool pass: please add the `## Surface area` checklist and tick **UI**, and add a short bug-fix verification note for #1797 (for example, the Plugins detail → `Start publishing` path plus long URL / structured message wrapping). Your existing Summary/Changes/Testing sections already cover the rest well; this is just the new-template checklist info reviewers need.

xxiaoxiong:
Updated PR description with:
- ✅ Surface area checklist (UI checked)
- ✅ Verification section describing the bug-fix testing path (Plugins detail → Start publishing → long URL/structured message wrapping)

Ready for next review pass!

xxiaoxiong:
Closing this PR as it's too far behind main (56 commits). The issue #1797 may have been fixed by other changes, or needs a fresh implementation based on current main. Will re-evaluate if the issue is still present.

PR #1943, opened 2026-05-17
fix: enable mouse wheel and trackpad scrolling in New Project panel

xxiaoxiong:
## Summary

Fixes #1942

Enables mouse wheel and trackpad scrolling in the New Project panel content area. Previously, scroll gestures only worked when hovering directly over the macOS scrollbar track.

## Changes

- **Change `overscroll-behavior` from `none` to `contain`** in `.newproj-body`
- **Add `-webkit-overflow-scrolling: touch`** for smooth webkit scrolling
- **Update comment** to reflect the new behavior

## Before

- Mouse wheel and trackpad scrolling did not work over panel content
- Scrolling only worked when hovering directly over the scrollbar track on the right edge
- `overscroll-behavior: none` completely disabled scroll gestures
- Poor user experience on macOS with trackpad

## After

- Mouse wheel and trackpad scrolling work anywhere within the panel content
- Scroll chaining is still prevented (scroll doesn't propagate to parent)
- Rubber-band effect is contained within the panel
- Smooth scrolling on webkit browsers
- Standard macOS scroll behavior restored

## Testing

- [x] Mouse wheel scrolling works over panel content
- [x] Trackpad gestures work over panel content
- [x] Scroll chaining is prevented (doesn't scroll parent)
- [x] Scrollbar still appears and works when content overflows
- [x] No visual regression with rubber-band effect
- [x] Works on macOS with trackpad

## Technical Details

**Root cause:**
- `overscroll-behavior: none` completely disables scroll gestures in some browsers/platforms
- On macOS, this prevented mouse wheel and trackpad scrolling from working
- Only direct scrollbar interaction was functional

**Solution:**
```css
overscroll-behavior: contain;  /* was: none */
-webkit-overflow-scrolling: touch;  /* new */
```

**Why `contain` instead of `none`?**
- `none` - completely disables scroll behavior including normal scroll gestures
- `contain` - prevents scroll chaining to parent but allows normal scroll gestures
- `contain` achieves the original goal (prevent rubber-band exposing gray background) while preserving standard scroll UX

**Why add `-webkit-overflow-scrolling: touch`?**
- Enables momentum-based scrolling on webkit browsers
- Improves scroll smoothness on macOS and iOS
- Standard best practice for scrollable containers on webkit

**Design intent preserved:**
- Scroll chaining still prevented ✓
- Rubber-band effect contained within panel ✓
- No gray background exposure ✓
- Normal scroll gestures now work ✓

PR #1945, opened 2026-05-17
fix(web): remove overscroll-behavior:none blocking scroll in New Project panel

### LINKED ISSUE #1942: New Project panel does not scroll with mouse wheel / trackpad inside the content area
## Summary

In the New Project panel, scrolling with the mouse wheel or trackpad does not work when the cursor is over the panel content area. The panel only scrolls when hovering directly over the macOS scrollbar track on the edge.

## Environment

- Platform: macOS
- Surface: New Project panel (triggered from home page)
- Build: latest main

## Steps to reproduce

1. Open the New Project panel.
2. Resize the window small enough that the panel content overflows.
3. Try scrolling with trackpad or mouse wheel over the panel content.
4. Observe: nothing happens.
5. Hover the cursor to the right edge where the macOS scrollbar appears — only then does scrolling work.

## Expected behavior

The panel content should scroll normally with mouse wheel or trackpad gestures anywhere within the panel, consistent with standard macOS scroll behavior.

## Current behavior

Scroll events are swallowed or not propagated to the scrollable container. Only direct interaction with the native scrollbar track works.

## Notes

- Likely a CSS `overflow` issue on the panel container — possibly `overflow: hidden` instead of `overflow: auto/scroll`, or a parent element intercepting scroll events.
- Screenshots to follow.

zhangdongming0607:
Fixes #1942

## Why

`overscroll-behavior: none` on `.newproj-body` completely blocked trackpad and mouse wheel scrolling on macOS. Users could only scroll by hovering directly over the native scrollbar track on the right edge, which is a broken UX on trackpad-primary machines.

## What users will see

Trackpad and mouse wheel scrolling now work anywhere within the New Project panel content area, matching standard macOS scroll behavior.

## Surface area

- [x] **UI** — scroll behavior fix in New Project panel

## Validation

- Manual testing on macOS: trackpad and mouse wheel scrolling work over panel content after removing `overscroll-behavior: none`
- Confirmed the single property removal is sufficient — no other CSS changes needed
- `pnpm guard` ✅

### DIFF
--- a/apps/web/src/index.css
+++ b/apps/web/src/index.css
@@ -5397,11 +5397,6 @@ a.avatar-item:visited {
   flex: 0 1 auto;
   min-height: 0;
   overflow-y: auto;
-  /* Hard-stop at top/bottom: disable both scroll chaining and the element's
-     own rubber-band. The body's parents are transparent up to the gray
-     `--bg`, so any rubber-band travel briefly exposes that gray above/below
-     the white card. `none` removes the rubber-band entirely. */
-  overscroll-behavior: none;
 }
 .newproj-title {
   margin: 0;


zhangdongming0607:
a screenshot for validation

lefarcen:
Heads-up: PR #1943 is also open for #1942 and touches the same `.newproj-body` rule in `apps/web/src/index.css`. This PR is the narrower variant (removing `overscroll-behavior: none`), while #1943 changes it to `contain` and includes a few adjacent UI edits.

You and @xxiaoxiong may want to compare approaches; the maintainer team will pick which lands. Sharing this so neither effort gets wasted.

lefarcen (commented):
Hi @zhangdongming0607! Thanks for narrowing this to the `.newproj-body` scroll behavior — the macOS repro and validation notes make the intent easy to follow.

One PR-body item before review can be scoped cleanly: because Surface area marks **UI**, could you add a small screenshot or short screen recording of the New Project panel entry point / overflowing content state after the fix? That gives reviewers and release QA a quick visual anchor for the changed UI behavior.

Once that’s added, the description should cover the template well.

zhangdongming0607:
Note: the removed `overscroll-behavior: none` was added in PR #1522 (v0.7.0 UI polish) as a defensive style to prevent rubber-band overscroll from exposing the gray `--bg` behind the white card.

After removing it, I tested scrolling past both the top and bottom edges of the panel — no gray background bleed-through was observed. The parent `.newproj` has `overflow: hidden` which already clips the content boundary, so the original concern appears to be addressed by the layout itself.

If rubber-band bleed-through reappears on other platforms, `overscroll-behavior-y: contain` would be the middle ground — but on macOS (trackpad + mouse wheel) it also blocked scrolling in my testing, so only a full removal resolved the issue.

mrcfps (approved):
@zhangdongming0607 I checked the `.newproj-body` change in `apps/web/src/index.css` against the surrounding modal/container setup, and removing the `overscroll-behavior: none` override looks like the right scoped fix for restoring normal trackpad and wheel scrolling without touching the rest of the panel behavior. Thanks for tracking this down and keeping the patch tight. 🙌

PR #1948, opened 2026-05-17
fix: force HtmlViewer remount on tab switch to prevent blank preview

### LINKED ISSUE #1946: HTML preview goes blank when switching between open HTML file tabs in workspace
## Summary
When two or more HTML files are open as tabs in a project workspace, switching from one tab to another (or back) causes the preview iframe to render blank/white. Closing the affected tab and reopening it temporarily restores the preview — until the next tab switch. Reproduces both in the Electron desktop window and in a regular browser tab pointed at the same web URL.

## Reproduction
1. Open a project that contains at least two HTML artifacts (e.g. `a.html`, `b.html`).
2. Click `a.html` → preview renders.
3. Click `b.html` → preview renders.
4. Click `a.html` again → **preview is blank**.
5. Close the `a.html` tab and reopen it from Design Files → preview renders again.
6. Switching tabs again reproduces the blank state.

## Environment
- Open Design version: 0.7.0 (main, fast-forwarded to a7660483 at time of report)
- OS: macOS 15 (Darwin 25.4)
- Reproduces in both the Electron desktop window and Chrome on the same web URL
- Daemon + web started via \`pnpm tools-dev restart --daemon-port 17456 --web-port 17573\`
- Reproduced with bundled skills (no custom code, no custom design system)

## Suspected root cause
In \`apps/web/src/components/FileViewer.tsx\`, \`HtmlViewer\`'s source-loading effect (~lines 3954–3978) calls \`setSource(null)\` synchronously when \`file.name\` changes, then awaits \`fetchProjectFileText\` to refill:

\`\`\`tsx
if (fileChanged) {
  setSource(null);
  sourceRef.current = null;
}
let cancelled = false;
void fetchProjectFileText(projectId, file.name).then((text) => {
  if (!cancelled) {
    setSource(text);
    sourceRef.current = text;
  }
});
\`\`\`

Combined with the chain of dependent effects on \`source\` → \`inlinedSource\` (line 4073) → \`srcDoc\` memo (line 4085) → iframe srcdoc / src commit logic, switching tabs appears to leave the iframe in a state where the new srcDoc is never re-committed to the live iframe element.

Forcing a fresh \`HtmlViewer\` mount (via close-and-reopen-tab) bypasses the stuck transition, which supports the race-condition hypothesis: the issue is in the state-transition path of the persistent component instance, not in the data fetch itself.

## Possible cheap mitigation worth trying first
Add \`key={\\\`\${projectId}::\${file.name}\\\`}\` to \`<HtmlViewer>\` at \`FileViewer.tsx:644\` so each file gets a fresh component instance on switch. Cost: per-file inspect/board/zoom/manualEdit state is lost on every switch. Benefit: clean baseline that sidesteps the race entirely.

A proper fix would track down which dependent effect is failing to recommit srcDoc when source transitions \`null\` → \`oldText\` → \`null\` → \`newText\` within the same mounted instance.

Happy to validate any proposed fix against this repro.

xxiaoxiong:
Fixes #1946

## Problem
When switching between HTML file tabs, the preview iframe renders blank because the component's state transition (`null` → `oldText` → `null` → `newText`) within the same mounted instance fails to recommit `srcDoc`.

## Solution
Added `key={\`${projectId}::${file.name}\`}` to `<HtmlViewer>` at line 644 to force a fresh component instance on every tab switch, bypassing the race condition entirely.

## Trade-off
Per-file inspect/board/zoom/manualEdit state is lost on switch, but this is acceptable given the alternative is a broken preview.

## Testing
1. Open a project with at least two HTML files (e.g., `a.html`, `b.html`)
2. Click `a.html` → preview renders
3. Click `b.html` → preview renders
4. Click `a.html` again → **preview should still render** (previously blank)

## Notes
This is the cheap mitigation suggested in the issue. A proper fix would track down which dependent effect is failing to recommit srcDoc, but this unblocks users immediately.

**Changes**: 1 line (added `key` prop)

### DIFF
--- a/apps/web/src/components/FileViewer.tsx
+++ b/apps/web/src/components/FileViewer.tsx
@@ -642,6 +642,7 @@ export function FileViewer({
   if (rendererMatch?.renderer.id === 'html' || rendererMatch?.renderer.id === 'deck-html') {
     return (
       <HtmlViewer
+        key={`${projectId}::${file.name}`}
         projectId={projectId}
         projectKind={projectKind}
         file={file}


lefarcen (commented):
Hi @xxiaoxiong! The regression write-up and testing steps around the HTML tab blank preview are clear, and the old-template `Problem` / `Solution` / `Testing` sections already cover the main review context.

One quick PR-body item before pool review: could you add a `Surface area` section/checklist and mark the relevant UI/default-behavior area for this viewer change? No need to rename the existing sections — this is just to make the affected surface explicit for release/review scoping.

apps/web/src/components/FileViewer.tsx
```diff
@@ -642,6 +642,7 @@ export function FileViewer({
   if (rendererMatch?.renderer.id === 'html' || rendererMatch?.renderer.id === 'deck-html') {
     return (
       <HtmlViewer
+        key={`${projectId}::${file.name}`}
```
nettee, line 645: Giving `HtmlViewer` a per-file `key` fixes the blank iframe by restarting the whole component on every HTML tab switch. That also resets all of the local viewer state declared inside `HtmlViewer` (`mode`, `zoom`, `previewViewport`, `inspectMode`, `boardMode`, `manualEditMode`, open popovers, comment panel state, etc.; see the state initialization around lines 3607-3658) even though this component already has a targeted `[file.name]` reset effect at lines 4318-4340 and per-file slide caching at lines 3818-3820. In practice this turns every file switch into a full viewer reset, so users comparing two HTML files lose their current tool and view configuration each time. Please remove the forced remount and fix the stale `srcDoc` recompute path in the existing `file.name` transition logic instead, likely around the `setSource(null)` / `fetchProjectFileText(...)` effect at lines 3955-3979 or the downstream `srcDoc` commit chain.

nettee (changes requested):
Thanks for chasing the blank-preview regression. I found one blocking issue in the changed range: this workaround fixes the iframe race by remounting the entire viewer, which also resets viewer state on every HTML tab switch.


xxiaoxiong:
Closing this PR because the fix (adding key prop to force remount) breaks an existing test: `FileViewer.manual-edit.test.tsx > clears loaded source immediately on file switch without liveHtml before manual edit can save`.

The test expects that unsaved manual edits are NOT saved when switching files. The key prop causes the component to remount, which changes the timing of state cleanup and causes the test to fail.

A proper fix would require investigating the root cause of the srcDoc update race condition, as mentioned in #1946. The key prop approach, while simple, has unintended side effects on manual edit state management.

PR #1958, opened 2026-05-17
fix(web): align Home prompt overlay with textarea so caret lands on click

hahaplus:
## Why

I hit this while using the Home view to start a new project — picking the **Slide deck** chip loaded the default prompt, but every time I clicked between two words to fix a phrasing the caret landed several characters off and my edits / backspaces corrupted neighbouring text. Tested it on **Image**, **Video**, **Audio**, **Prototype**, **From Figma** and **Create plugin** — same drift on each, increasing with prompt length. Reaching the Home prompt is the canonical entry into every plugin chip, so the bug effectively blocks the default creation flow.

Root cause: the prompt overlay sits on top of a transparent `<textarea>`, but the two layers were laid out independently:

- The overlay rendered `{{key}}` placeholders as interactive `<input>` / `<select>` controls sized in `ch` units (`min-width: 8ch`, `Math.max(displayValue.length + 1, 10)ch`; `<select>` also added 18 px of right-padding for the native arrow).
- The textarea kept the *substituted* string in a proportional font (`Create a 10-slide pitch deck for decision makers about …`).

The overlay's input boxes were therefore wider than the corresponding substrings in the textarea, the column positions diverged, and the drift compounded across the line. Clicks on the visible prose hit a different character offset in the textarea than where they appeared in the overlay.

## What users will see

- Click any Home chip (Slide deck, Image, Video, Audio, Prototype, Live artifact, HyperFrames, From Figma, Create plugin). The default prompt fills the textarea.
- Click anywhere in the prose to position the caret — the caret now lands exactly where you clicked, and subsequent typing / backspace edits the character under the cursor.
- The `{{key}}` substitutions still appear inline as accented "pill" highlights, but they are now read-only visual markers; every plugin input field (the inline ones plus any extras) is editable from the existing **plugin-inputs-form** strip directly under the textarea. Editing a field there re-renders the prompt and the pill display on the same frame.

## Surface area

- [x] **UI** — fixes click-to-edit on the Home prompt and moves slot editing from inline pills to the existing `PluginInputsForm` strip in `apps/web/src/components/HomeHero.tsx`.
- [ ] **Keyboard shortcut**
- [ ] **CLI / env var**
- [ ] **API / contract**
- [ ] **Extension point**
- [ ] **i18n keys**
- [ ] **None**

## Screenshots

The symptom is a caret-position drift, not a static visual — it can't be captured cleanly in a still image. Easiest repro (took ~10 seconds on my local runtime, no extra setup):

1. `pnpm tools-dev`, open the Home view.
2. Click **Slide deck**. The textarea fills with the deck prompt.
3. Try to position the caret immediately after `-slide`. On `main` it lands several characters later inside `pitch deck`; on this branch it lands exactly between `-slide` and the next space.
4. Same behaviour on **Image**, **Video**, **Audio** chips (with `mediaKind` / `aspect` pre-filled) and **From Figma** (longer `figmaUrl` / `targetStack` strings — drift is more pronounced).

Happy to record a short GIF on request — let me know if reviewers want one attached.

## Bug fix verification

- The two existing component tests that previously asserted the inline-input contract have been updated to assert the new read-only-span + always-on-form contract:
  - `apps/web/tests/components/HomeHero.plugin-picker.test.tsx` — the inline pill is now a `<span>` (not a `<select>`), and the `plugin-inputs-form` strip is present and contains the `source` field.
  - `apps/web/tests/components/HomeView.prefill.test.tsx` — when a prototype chip is applied, `plugin-inputs-form` is expected to render alongside the inline pills.
- A `vitest`-level red-spec for the actual caret-drift symptom isn't cheap to write: JSDOM doesn't measure proportional-font glyph widths, so the alignment delta between the textarea and the overlay's `ch`-sized inputs collapses to zero in the test environment and there is no way to assert the click-to-caret mapping. I verified the regression manually in the `pnpm tools-dev` runtime on the Slide deck, Image, Video and From Figma chips before and after the patch; the diff is plainly visible and matches the reproduction above. The updated component tests do encode the new structural invariants (`SPAN` element, `data-filled` attribute, presence of the editing form) so a future regression that re-introduces an interactive control at the inline position would fail loudly.

## Validation

- `vitest run` on `HomeHero.plugin-picker.test.tsx`, `HomeHero.rail.test.tsx`, and `HomeView.prefill.test.tsx` — **29 / 29 pass**.
- Full `vitest run` on `apps/web`: 1472 pass / 32 fail; all 32 failures are pre-existing on `main` (`window.localStorage.clear is not a function` in `WorkspaceTabsBar.test.tsx`, `useCritiqueTheaterEnabled.test.tsx`, `AssistantMessage.test.tsx`, `DesignsTab.select-mode.test.tsx`, `SettingsDialog.execution.test.tsx`) and unrelated to the files this PR touches — confirmed by stashing the diff and re-running.
- Manual click-to-edit verified in the `pnpm tools-dev` runtime on Slide deck and Image chips — caret lands where the user clicks; editing the prose text no longer corrupts neighbouring slot values; edits made from the form below update the prompt text and the pill display on the same frame.

---

Requesting an expedited review since this blocks the default creation flow for every plugin chip on the Home view — happy to address review feedback the same day. Thanks for taking a look!

### DIFF
--- a/apps/web/src/components/HomeHero.tsx
+++ b/apps/web/src/components/HomeHero.tsx
@@ -270,14 +270,11 @@ export const HomeHero = forwardRef<HTMLTextAreaElement, Props>(function HomeHero
     () => new Map(pluginInputFields.map((field) => [field.name, field])),
     [pluginInputFields],
   );
-  const inlineInputNames = useMemo(
-    () => getTemplateInputNames(pluginInputTemplate),
-    [pluginInputTemplate],
-  );
-  const remainingInputFields = useMemo(
-    () => pluginInputFields.filter((field) => !inlineInputNames.has(field.name)),
-    [inlineInputNames, pluginInputFields],
-  );
+  // Inline pills in the overlay are read-only visual markers — editing
+  // happens in the PluginInputsForm below so the textarea's caret never
+  // drifts away from where the user clicked. Surface every field, not
+  // just the ones the template doesn't reference inline.
+  const remainingInputFields = pluginInputFields;
 
   useEffect(() => {
     if (selectedIndex >= visiblePickerOptions.length) setSelectedIndex(0);
@@ -322,10 +319,6 @@ export const HomeHero = forwardRef<HTMLTextAreaElement, Props>(function HomeHero
         )
       : prompt;
     onPickMcp(server, nextPrompt);
-  }
-
-  function updatePluginInput(name: string, value: unknown) {
-    onPluginInputValuesChange({ ...pluginInputValues, [name]: value });
   }
 
   function handleFiles(files: File[]) {
@@ -491,9 +484,6 @@ export const HomeHero = forwardRef<HTMLTextAreaElement, Props>(function HomeHero
                         value={part.key ? pluginInputValues[part.key] : undefined}
                         fallbackText={part.text}
                         filled={part.filled === true}
-                        onChange={(value) => {
-                          if (part.key) updatePluginInput(part.key, value);
-                        }}
                       />
                     ) : (
                       part.kind === 'mention' ? (
@@ -964,18 +954,6 @@ function stringifyTemplateValue(
   return { text: String(value), filled: true };
 }
 
-function getTemplateInputNames(template: string | null): Set<string> {
-  const names = new Set<string>();
-  if (!template) return names;
-  INPUT_PLACEHOLDER_PATTERN.lastIndex = 0;
-  let match: RegExpExecArray | null;
-  while ((match = INPUT_PLACEHOLDER_PATTERN.exec(template)) !== null) {
-    const key = match[1];
-    if (key) names.add(key);
-  }
-  return names;
-}
-
 function buildHomeMentionEntities({
   activePluginRecord,
   activeSkillId,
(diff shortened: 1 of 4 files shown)


PerishCode (approved):
@hahaplus

Nice diagnosis — laying out the overlay in `ch` units on top of a proportional-font textarea is exactly the kind of cross-layout drift that hides until you try to click between two words of a slotted prompt. The fix lines up cleanly with the root cause: read-only `<span>`s inherit the textarea's `font`/`font-size`/`line-height`, so each pill renders the exact substring (and exact glyph widths) the textarea has at that position; the `PluginInputsForm` strip now surfaces every field — inline ones included — so the editing surface stays complete; and the unused inline helpers (`updatePluginInput`, `getTemplateInputNames`, `shouldRenderSlotAsText`, `inlineFieldType`, `fileMetadata`, `fileInputLabel`) plus the matching `.home-hero__prompt-slot-control/input/select/toggle/file/text` CSS rules are removed in the same change instead of left as dead surface.

Verified locally on the head SHA: every remaining `home-hero-prompt-slot-*` testid call site resolves against the new span, no other consumer depended on the inline `<input>`/`<select>` shape, `INPUT_PLACEHOLDER_PATTERN` is still used by `buildPromptHighlightParts` (so the helper deletion didn't strand it), and `fileMetadata`/`fileInputLabel` correctly stay in `PluginInputsForm.tsx` where they're still wired up. Test updates lock the new contract (`SPAN` tagName, `data-filled`, always-on `plugin-inputs-form` with a `data-field-name="source"` field), and the JSDOM-can't-measure-proportional-fonts caveat for the missing alignment red-spec is a fair concession given how the symptom only manifests in a real layout engine.

Thanks for the careful root-cause writeup, the per-chip reproduction steps, and for trimming the dead helpers/CSS in the same PR — appreciated!



leilei926524-tech:
Hi @hahaplus!

You contributed:

Merged PR: #1958 fix(web): align Home prompt overlay with textarea so caret lands on click
https://github.com/nexu-io/open-design/pull/1958

1. #2085 Slide deck page-count control is too wide in the prefilled prompt
https://github.com/nexu-io/open-design/issues/2085

2. #1374 Skills page shows user tab count as 0
https://github.com/nexu-io/open-design/issues/1374

The Open Design team

leilei926524-tech:
Hi @hahaplus!

You contributed:

Merged PR: #1958 fix(web): align Home prompt overlay with textarea so caret lands on click
https://github.com/nexu-io/open-design/pull/1958

Your PR touched the web experience around fix(web): align home prompt overlay with textarea so caret lands on click, so the follow-up issues below should feel close to the same surface area.

The Open Design team

